PR #1583, opened 2022-06-13
Add system patch to core 5e types

arbron:
This adds a patch to `ActiveEffect5e`, `Actor5e`, `Item5e`, and `TokenDocument5e` which allows `data.data` to be accessed from `system` when running v9. This should enable the work of converting old system data references to v10 style without breaking compatibility with v9.


arbron:
Closed because we aren't doing joint compatibility

PR #1584, opened 2022-06-13
 [#1441] Move migration version constants into system.json flags

### LINKED ISSUE #1441: Move migration version numbers into system.json flags
Currently the `NEEDS_MIGRATION_VERSION` and `COMPATIBLE_MIGRATION_VERSION` are constants hardcoded in the ready hook. It would be better if these pieces of data were moved to `system.json` where the rest of the system's version numbers are defined:

```json
"flags": {
  "needsMigrationVersion": "1.5.6",
  "compatibleMigrationVersion": "0.8"
}
```

This would keep the `needsMigrationVersion` value directly adjacent to the current `version` number to remind us to check if it needs to be increased whenever we are preparing new releases of the system. It would also allow that data to be read by modules if they wanted to act upon it in some way.

arbron:
Moves `NEEDS_MIGRATION_VERSION` and `COMPATIBLE_MIGRATION_VERSION` into flags defined in `system.json` to avoid having to change `dnd5e.js` every time they need to be updated. Keeping these flags adjacent to the current system version should also help ensure we don't forget to update them.

Resolves #1441


aaclayton:
I implemented these changes in https://github.com/foundryvtt/dnd5e/commit/4b57267bb0303997f6f87462ff5abf906f8a1340

PR #1585, opened 2022-06-13
 [#1249, #1461] Add physical damage bypasses to actor configuration

### LINKED ISSUE #1249: Add adamantine & silvered physical damage types to immunities, resistances, and vulnerabilities
Currently the system support `"Non-Magical Physical"` in damage resistances and related fields to cover the resistance to `"Bludgeoning, Piercing, and Slashing from Nonmagical Attacks"` commonly found on monsters. Another common variation found quite frequently is `"Bludgeoning, Piercing, and Slashing from Nonmagical Attacks that aren't Silvered"` and `"Bludgeoning, Piercing, and Slashing from Nonmagical Attacks that aren't Adamantine"`.

Both of these two variations are common enough that they should be added to the list of options. This would not only improve the data stored on characters, but also give role automation systems the data they need to properly apply damage by checking if the attacking weapon is silvered or adamantine.

### LINKED ISSUE #1461: Monsters (SRD) with Resistance/Immunity to bludgeoning, piercing, and slashing from nonmagical attacks
Originally in GitLab by @Rabulias

Currently this Resistance/Immunity is noted in the Special free text field under Resistances or Immunities. There is the checkbox for "Non-Magical Physical" which I think would be a better setting, and one that automation modules could take advantage of.

For examples, see Ghost (Damage Resistance) and Kraken (Damage Immunity).

This may touch on Issue 1249 (Add adamantine & silvered physical damage types to immunities, resistances, and vulnerabilities (https://gitlab.com/foundrynet/dnd5e/-/issues/1249 )) with some monsters like Devils and Lycanthropes that are Resistant/Immune to "bludgeoning, piercing, and slashing from non-magical attacks that aren't silvered".

arbron:
This commit adds a new method for configuring creatures that have resistance or immunity to physical damage except from magical, silvered, or adamantine weapons.

The current setup only has a single resistance value for `"Non-Magical Physical"`, but this is limiting for several reasons. First, it doesn't account for creatures such as various types of devils or werewolves that have their resistance bypassed by silvered weapons in addition to magical ones. It also doesn't handle a creature like the Xorn from the SRD which has resistance to only piercing and slashing damage from non-magical attacks, but not bludgeoning. Currently all of these exceptions are implemented using the custom field which isn't easily automated.

This proposal replaces the simple `physical` value from `DND5E.damageResistanceTypes` with a more expressive system of system of physical resistance bypasses. In the actor's data, this means a new `bypasses` array has been added to the damage immunity, resistance, and vulnerability traits. This contains an array of weapon property keys that can bypass any resistance to physical damage types. So the Xorn example above would be configured like so:

```json
"traits": {
  "dr": {
    "value": ["piercing", "slashing"],
    "bypasses": ["ada", "mgc"],
    "custom": ""
  }
}
```

Indicating that it has resistance to `"Piercing"` and `"Slashing"` damage, but that can be bypassed by any weapon with the `"Adamantine"` or `"Magical"` properties.

This is handled by the splitting off two configuration values. `DND5E.physicalWeaponProperties` contains a subset of weapon properties that can bypass physical resistance and currently includes adamantine, magical, and silvered. `DND5E.physicalDamageTypes` contains a subset of damage types that are capable of being bypassed and includes bludgeoning, piercing, and slashing.

### Configuration
A new version of `TraitSelector` is used to configure these damage types which includes a bypasses section:

Configuration

### Sheet
The character sheet then takes those values and formats them in a style similar to how they are displayed in the SRD:

Barbed Devil Comparison

The left is how it displays in the current SRD entry for `"Barbed Devil"`, and the right is the new version using this system.

Resolves #1249
Resolves #1461


Fyorl (changes requested):
Good stuff, some minor suggestions

arbron:
@Fyorl Feel free to make adjustments according to your comments

PR #1588, opened 2022-06-13
 [#1553] Add ability to duplicate advancements

### LINKED ISSUE #1553: [Feature Request] Allow Duplication of Advancement Items
Originally in GitLab by @garrysmod367

It would be very helpful to have the ability to duplicate an advancement feature so that features that have repeating choices from one level to the next can be implemented more quickly. Current fastest workflow I can find for this is to open up the previous advancement feature, create the new feature and drag the repeating choices from one to the other and then make the necessary modifications. This is rather slow and most other aspects of Foundry have duplication.

arbron:
Adds a context menu to the advancement list with options for editing, duplicating, and deleting advancements. The new `Item5e#duplicateAdvancement` method creates a copy of the advancement but resets its `value`.

Also fixes some advancement issues introduced by the v10 transition.

Resolves #1553


aaclayton (approved):
Looks good, thanks @arbron!

PR #1589, opened 2022-06-13
Replace text inputs that use Number dtype with proper number inputs

arbron:


aaclayton (approved):
Nice changes, thanks @arbron

PR #1590, opened 2022-06-13
[#406, #866] Convert max HP to calculated value on characters & add hit points configuration app

### LINKED ISSUE #406: Compute correct hit points for a given character using their chosen class levels, ability scores, with optional support for toughness or other modifiers.
Originally in GitLab by @gatesvp

Now that we have explicit class levels, we should be able to directly calculate a PCs HP from that data. This would match up with Adventurers League.

Notes:
 - For multiclassing, we clearly need to know which class was "first". Though we need this if we ever want to track proficiencies as well, so this is reasonable to have.
 - Not clear how we hook up modifiers from things like Dwarves or Feats like Tough that grant extra HP at every level.
 - We likely want a manual override here, though we may be a little tight on space for that.

### LINKED ISSUE #866: Create a "Hit Points Configuration" sub-application which is used to configure level-up HP values, monster HP formula, or other hit point properties.
Imported from Gitlab

arbron:
## Convert HP to Calculated Value

Max HP on characters is now the result of any hit points stored in `HitPointsAdvancement`s on classes, the player's constitution modifier, and a set of new per-level and overall HP bonus formulas. There is also a new `override` field to replace any calculated max HP.

```json
"hp": {
  "value": 0,
  "min": 0,
  "temp": 0,
  "tempmax": 0,
  "override": null,
  "bonuses": {
    "level": "",
    "overall": ""
  }
}
```

## Configuration App

A new hit points configuration app allows configuration of HP bonuses and override on characters, and HP max and roll formula on NPCs.

Character Config

NPC Config

Rolling hit points for NPCs has been moved into the configuration app to avoid accidental re-rolls, and a new `Actor5e#rollNPCHitPoints` method has been added to standardize that rolling.

### Todo

- [x] Create migration to transition existing characters smoothly
- [ ] Ensure `Starter Heroes` compendium is migrated

Resolves #406 
Resolves #866


aaclayton (changes requested):
Looking good @arbron - thank you. The big missing piece is the migration. In addition to migrating existing data stored in `hp.max` to instead be `hp.override` (only if needed) we also need to consider adding a shim so that modules and macros which currently make updates to `hp.max` would be redirected to make those updates to `hp.override` instead.

I also think it might be useful to have some visual indicator on the character sheet to show when max HP is overridden instead of automatically computed.

module/applications/actor/hit-points-config.mjs
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Handle rolling NPC health values using the provided formula.
+   * @param {Event} event  The original click event.
+   * @protected
+   */
+  async _onRollHPFormula(event) {
+    event.preventDefault();
+    try {
+      const roll = await this.clone.rollNPCHitPoints();
```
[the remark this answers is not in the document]
arbron: Well this was to distinguish from the existing `rollClassHitPoints` on actor, which `HitPointsAdvancement` uses.

module/data/actor/character.mjs
```diff
@@ -70,11 +73,15 @@ export default class CharacterData extends CreatureTemplate {
...
-          max: new foundry.data.fields.NumberField({
-            nullable: false, integer: true, min: 0, initial: 0, label: "DND5E.HitPointsMax"
+          override: new foundry.data.fields.NumberField({
+            nullable: true, integer: true, min: 0, initial: null, label: "DND5E.HitPointsOverride"
           }),
           temp: new foundry.data.fields.NumberField({integer: true, initial: 0, min: 0, label: "DND5E.HitPointsTemp"}),
-          tempmax: new foundry.data.fields.NumberField({integer: true, initial: 0, label: "DND5E.HitPointsTempMax"})
+          tempmax: new foundry.data.fields.NumberField({integer: true, initial: 0, label: "DND5E.HitPointsTempMax"}),
+          bonuses: new foundry.data.fields.SchemaField({
+            level: new FormulaField({label: "DND5E.HitPointsBonusLevel"}),
```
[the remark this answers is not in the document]
arbron: It does need to be deterministic because we don't have any place here to store a rolled value (that would have to be added to `HitPointsAdvancement`). I don't think I've seen anything that adds an extra rolled value per level, only flat values like Tough.

arbron:
@aaclayton I set it up so the tooltip changes depending on whether or not max HP is overridden. If you think we need more, I was thinking making it italic would be a way to do that but not be too obtrusive for people who want to leave it with the override.

templates/actors/character-sheet.hbs
```diff
@@ -52,15 +52,16 @@
             {{!-- Header Attributes --}}
             <ul class="attributes flexrow">
                 <li class="attribute health">
-                    <h4 class="attribute-name box-title">{{ localize "DND5E.HitPoints" }}</h4>
+                    <h4 class="attribute-name box-title">{{localize "DND5E.HitPoints"}}</h4>
+                    <a class="config-button" data-action="hit-points"
+                        data-tooltip="{{localize 'DND5E.HitPointsConfig'}}"><i class="fas fa-cog"></i></a>
                     <div class="attribute-value multiple">
                         <input name="system.attributes.hp.value" type="text"
```
aaclayton, line 59: Would be good to convert `<input type="text" data-dtype="Number">` into `<input type="number">`

templates/actors/character-sheet.hbs
```diff
@@ -52,15 +52,16 @@
...
                             value="{{system.attributes.hp.value}}" placeholder="10"
-                            title="{{localize 'DND5E.HitPointsCurrent'}}" data-dtype="Number">
+                            data-tooltip="{{localize 'DND5E.HitPointsCurrent'}}" data-dtype="Number">
                         <span class="sep"> / </span>
-                        <input name="system.attributes.hp.max" type="text"
-                            value="{{system.attributes.hp.max}}" placeholder="10"
-                            title="{{localize 'DND5E.HitPointsMax'}}" data-dtype="Number">
+                        <div data-tooltip="{{#if source.attributes.hp.max}}{{localize 'DND5E.HitPointsOverride'}}
```
aaclayton, line 63: Could we use a `<span>` for this instead of a `<div>`? that would be more consistent with how we handle non-editable labels elsewhere in the HTML structure.

templates/actors/npc-sheet.hbs
```diff
@@ -59,15 +62,14 @@
                             data-dtype="Number">
                     </div>
                     <footer class="attribute-footer flexrow">
-                        <input name="system.attributes.hp.formula" class="hpformula" type="text"
-                            placeholder="{{ localize 'DND5E.HPFormula' }}"
-                            value="{{system.attributes.hp.formula}}"
-                            title="{{localize 'DND5E.HPFormula'}}">
                         <input name="system.attributes.hp.temp" type="text" class="temphp"
```
aaclayton, line 65: Should this be `type="number"`?

templates/actors/npc-sheet.hbs
```diff
@@ -59,15 +62,14 @@
...
-                            placeholder="+{{localize 'DND5E.Temp'}}"
-                            value="{{system.attributes.hp.temp}}"
-                            title="{{localize 'DND5E.HitPointsTemp'}}"
-                            data-dtype="Number">
+                               placeholder="+{{ localize 'DND5E.Temp' }}"
+                               value="{{system.attributes.hp.temp}}"
+                               title="{{localize 'DND5E.HitPointsTemp'}}">
+                        <input name="system.attributes.hp.tempmax" type="text" class="temphp"
```
aaclayton, line 69: Should this be `type="number"`?

templates/actors/character-sheet.hbs
arbron: These were converted back to `text` because a bug in Chrome prevents players from entering delta values (`+10` to add 10 rather then set to 10).

templates/actors/character-sheet.hbs
arbron: Hmm, I used `<div>` here specifically because I need to add some flex styling to match the spacing of the `<input>` without affecting the spacer in between. I could do this by specifically excluding the separator:

Current:
```css
&.multiple input, &.multiple div {
```

New:
```css
&.multiple input, &.multiple span:not(.sep) {
```

module/applications/advancement/hit-points-flow.mjs
```diff
@@ -18,6 +18,7 @@ export default class HitPointsFlow extends AdvancementFlow {
   getData() {
     const source = this.retainedData ?? this.advancement.value;
     const value = source[this.level];
+    // TODO: Also preview CON mod + per-level bonuses
```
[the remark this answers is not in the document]
arbron: I have this finished, but I think it should be handled as a separate small PR.

module/documents/advancement/hit-points.mjs
```diff
@@ -112,6 +112,19 @@ export default class HitPointsAdvancement extends Advancement {
...
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * Total hit points taking the provided ability modifier into account, with a minimum of 1 per level.
+   * @param {number} mod  Modifier to add per level.
+   * @returns {number}    Total hit points plus modifier.
+   */
+  adjustedTotal(mod) {
```
aaclayton, line 122: Would prefer this with a verb-style method name like `getAdjustedTotal(mod)`

module/documents/advancement/hit-points.mjs
```diff
@@ -125,14 +138,27 @@ export default class HitPointsAdvancement extends Advancement {
   /*  Application Methods                         */
   /* -------------------------------------------- */
 
+  /**
+   * Add the ability modifier and any bonuses to the provided hit points value to get the number to apply.
+   * @param {number} value  Hit points taken at a given level.
+   * @returns {number}      Hit points adjusted with ability modifier and per-level bonuses.
+   */
+  #applicableValue(value) {
```
aaclayton, line 146: Similar comment, `#getApplicableValue()`

aaclayton (approved):
Looks good to me. I have 2 remaining comments on code style but I think this is ready to merge in.

PR #1591, opened 2022-06-13
[#1431] Add keybindings for fast forwarding rolls

### LINKED ISSUE #1431: Add support for custom keybindings
With the new custom keybindings system built in, 5e should support user customization for things like roll fast-forwards.

arbron:
An attempt to introduce customizable keybindings for fast forwarding d20 and damage rolls. The current keybindings setup doesn't seem to be built to handle keys that a pressed while clicking on something, so I created a utility function (currently within `dice.js`, but it might be valuable to move this up to `utils.js` to make it more widely available) that checks if the provided keybinding is satisfied by the currently pressed keys.

Resolves #1431


module/dice/dice.mjs
```diff
@@ -112,12 +112,15 @@ export async function d20Roll({
  * @returns {{isFF: boolean, advantageMode: number}}  Whether the roll is fast-forward, and its advantage mode.
  */
 function _determineAdvantageMode({event, advantage=false, disadvantage=false, fastForward=false}={}) {
-  const isFF = fastForward || (event && (event.shiftKey || event.altKey || event.ctrlKey || event.metaKey));
+  const keys = {
+    normal: event ? _areKeysPressed(event, "d20RollFastForwardNormal") : false,
+    advantage: event ? _areKeysPressed(event, "d20RollFastForwardAdvantage") : false,
+    disadvantage: event ? _areKeysPressed(event, "d20RollFastForwardDisadvantage") : false
+  };
+  const isFF = fastForward || Object.values(keys).some(k => k);
```
kaelad02, line 120: My module, Advantage Reminder, does this same kind of check to see if the user is fast-forwarding rolls. I basically copied the original way `isFF` was calculated. It's a bit more complicated now, which is to be expected. I'm wondering for my use case if I'd be better off just calling `_determineAdvantageMode` myself but my only hesitation is that it's marked as a private function (starting with `_`).

You alluded to potentially moving some of this into `util.js`. If that makes it more reusable by modules, like mine, then consider this a +1 to doing that.

hakr14:
I'd argue this (and other similar "rebinding" issues and PRs) should be labeled https://github.com/foundryvtt/dnd5e/labels/accessibility

module/utils.mjs
```diff
@@ -227,6 +227,33 @@ export function staticID(id) {
...
 
+/* -------------------------------------------- */
+/*  Keybindings Helper                          */
+/* -------------------------------------------- */
+
+/**
+ * Based on the provided event, determine if the keys are pressed to fulfill the specified keybinding.
+ * @param {Event} event    Triggering event.
+ * @param {string} action  Keybinding action within the `dnd5e` namespace.
+ * @returns {boolean}      Is the keybinding is triggered?
```
Fyorl, line 238:
```suggestion
 * @returns {boolean}      Is the keybinding triggered?
```

module/settings.mjs
```diff
@@ -1,6 +1,28 @@
...
+  });
+
+  game.keybindings.register("dnd5e", "skipDialogAdvantage", {
+    name: "KEYBINDINGS.DND5E.SkipDialogAdvantage",
+    editable: [{ key: "AltLeft" }, { key: "AltRight" }]
+  });
+
+  game.keybindings.register("dnd5e", "skipDialogDisadvantage", {
+    name: "KEYBINDINGS.DND5E.SkipDialogDisadvantage",
+    editable: [{ key: "CtrlLeft" }, { key: "CtrlRight" }, { key: "OsLeft" }, { key: "OsRight" }]
```
Fyorl, line 20:
```suggestion
    editable: [{ key: "ControlLeft" }, { key: "ControlRight" }, { key: "OsLeft" }, { key: "OsRight" }]
```

Fyorl (approved):

Issue #1559, opened 2022-05-21
Add titles to fields in item usage section


arbron:
There are a number of fields in the item activation section of various item templates that don't have titles. Things like the three different inputs after `"Target"` give no indication of what they are used for without poking around. These inputs should all have titles for accessibility and general usability.

Issue #1558, opened 2022-05-21
Split non-movement units from distanceUnits into separate config enum


arbron:
Currently the config for `distanceUnits` includes a combination of `movementUnits` and a number of other, special units:

```javascript
DND5E.distanceUnits = {
  none: "DND5E.None",
  self: "DND5E.DistSelf",
  touch: "DND5E.DistTouch",
  spec: "DND5E.Special",
  any: "DND5E.DistAny",
  ...DND5E.movementUnits
};
```

Splitting off these special unit types into a separate config object before combining them would make it possible to display them distinctly in the UI and act upon them more precisely in code.

Issue #1557, opened 2022-05-21
Add distinction between area target types and individual target types


arbron:
Currently all target types for items are contained in one large list. Splitting these between targeting modes that effect distinct targets (`"Self"`, `"Ally"`, `"Enemy"`, etc.) versus ones that effect an area (`"Radius"`, `"Sphere`", `"Line"`, etc.) would be useful both to clarify the UI and the API.

PR #1592, opened 2022-06-13
[#1419, #1557, #1558, #1559] Split out distanceUnits & targetTypes config entries to allow for more information & add accessibility titles to fields on item sheet

### LINKED ISSUE #1419: Cover missing from "activatedEffect" item template
Items with the `"Crew Action"` activation type have a field named `"Cover"` that can be set but this is not included anywhere in `template.json`. This should be added to the `"activatedEffect"` template so it is properly set up on all items.

Also, the cover dropdown currently allows an empty value as an input, but I'm not sure how that is different from the `"None"` cover type. It seems the empty value should be removed and the default value should be `0` to represent no cover.

arbron:
This MR accomplishes two tasks, improving configuration data for ranges and targets, and improving the item activation template.

## Config Adjustments

A continuation of the work @zeel02 began a year ago with !182, this splits up some config objects in order to allow more information to be gathered from the selected values.

First, this adds a new `DND5E.rangeTypes` config object that contains entries like `"self"`, `"touch"`, and more that were previously included directly in `distanceUnits`. This will help with some of the UI improvements Josh was working on in his old MR.

Second, it splits `targetTypes` into separate `individualTargetTypes` and `areaTargetTypes` to indicate which are which (and then merges them together into `targetTypes` to avoid breaking any existing behavior). The `individualTargetTypes` is a simple string enum, but `areaTargetTypes` now contains objects with the type label and a mapping to the `MeasuredTemplate` type. This is the only potential breaking change if any modules area using the old `areaTargetTypes` to look up templates.

## Template Adjustments

I've replaced all of the instances of `{{#select}}` with `{{selectOptions}}` to clean up the template a bit.

I also added titles to a bunch of the fields to improve clarity and accessibility.

Finally, I grouped the options within target types into individual and area sections to make it easier to find that target type you are searching for:

Area : Individual Units

## Notes

- This also adds the missing `"cover"` value to the activated effect template to resolve #1419

- Both the target and range fields have a value of `"none"` in addition to the blank value. As far as I can tell there is no difference between none and an empty field, so it seems like we should remove one or the other.


Fyorl:
I think there might be a small semantic difference between blank vs 'none', where blank can mean 'not applicable', and 'none' can mean 'explicitly incapable of targeting anything'. I think we do occasionally make use of the blank value to check if some logic should apply or some extra fields should be visible, but I don't think that's the case for any of these target types. I can't think of an occasion where 'none' would be a useful distinction over blank for these targets.

I'd be in favour of removing 'none' from the target types then, but not in this PR since it would likely require a migration.

Fyorl (approved):
Thanks Jeff, looks good, just needs a rebase. Tagged it as breaking since, even with the `toString` patch, a caller might still get back an unexpected object value, but I think it's pretty minor.

PR #1597, opened 2022-06-16
[#1470] Allow migration version to be set via world manifest flag

### LINKED ISSUE #1470: Unnecessary System Data Migration if world has no settings (should use `systemVersion` from manifest)
Originally in GitLab by @arcanistzed

I'm working on a world package for distribution and I am not including the `settings.db` with my package. I have noticed that there is always a migration whenever I install it despite it not needing one:
```js
const NEEDS_MIGRATION_VERSION = "1.5.6";
const currentVersion = "";
isNewerVersion(NEEDS_MIGRATION_VERSION, currentVersion) // returns true
```
A solution for this would be to provide a fallback for the current version from `game.world.data.systemVersion` so that it isn't an empty string.

arcanistzed:
Closes #1470

This allows world distributors to set a manifest flag to let the dnd5e system know that this world has already been migrated to a certain version. Normally, the `settings.db` serves this purpose, but sharing that is often undesirable (e.g. in a git repo) since it may contain sensitive info such as API keys for certain modules.

For example, you may find this in a world's manifest file to indicate that the world has been migrated to v1.6.3 of dnd5e:
```json
"flags": {
    "dnd5e": {
        "version": "1.6.3"
     }
}
```

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -189,7 +189,7 @@ Hooks.once("ready", function() {
 
   // Determine whether a system migration is required and feasible
   if ( !game.user.isGM ) return;
-  const cv = game.settings.get("dnd5e", "systemMigrationVersion");
+  const cv = game.settings.get("dnd5e", "systemMigrationVersion") || game.world.data.flags.dnd5e?.version;
   const totalDocuments = game.actors.size + game.scenes.size + game.items.size;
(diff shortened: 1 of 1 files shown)


arbron (changes requested):
This will need to be rebased atop the new `v10-dev` branch for the upcoming version.

arbron (approved):

aaclayton (approved):

PR #1599, opened 2022-06-17
[#1350] Minimum Roll for skills, abilities and attacks

### LINKED ISSUE #1350: Support minimum roll values on skill checks

The Bard College of Eloquence has the feature Silver Tongue which allows:

> When you make a Charisma (Persuasion) or Charisma (Deception) check, you can treat a d20 roll of 9 or lower as a 10.

I would like to be able to set a "minimum value" for individual skills, possibly as part of the skill schema.

szefo09:
Allows the global attack, ability or skills to have minimum roll modifier defined and applied.

https://user-images.githubusercontent.com/19979839/174321154-5a558619-a0b6-41bd-ac28-d1521ae580aa.mp4

Resolves #1350


arbron (approved):

arbron (changes requested):
Could you rebase this atop the `v10-dev` branch and fix the merge issues? I might attempt to get this into `2.0`.

szefo09:
Haven't tested those changes in latest Foundry, but I did my best to fix the merge issues.

arbron (changes requested):
Be sure you document `options.minimum` in the JSDoc comment at the top of `D20Roll`.

arbron:
One thing to consider, and I'm not sure of a good solution at the moment, is if you have a ability check minimum defined (e.g. minimum 5 on dex checks), and you roll a skill that is normally rolled with that ability, but you change the ability in the configuration dialog. So if you roll a charisma-based Sleight of Hand check, right now it still applies the dex minimum, not whatever minimum charisma ability checks have.

Not sure the best way to fix that at the moment, it might be okay to just leave it to a later patch unless you have an idea.

module/documents/item.mjs
```diff
@@ -619,6 +619,22 @@ export default class Item5e extends Item {
...
+   * @returns {number|null}  The minimum roll value that should be added to the attack roll formula.
+   */
+   getMinimumRoll() {
+    const itemData = this.data.data;
+    const actorFlags = this.actor.data.flags.dnd5e || {};
+    if ( !this.hasAttack || !itemData ) return;
+
+    // Return the minimum from the flag
+    return actorFlags?.minimumAttackRoll;
+  }
```
arbron, line 634: There is a warning here for accessing the deprecated `data` object.

Also, we can probably avoid most of the complexity used in `getCriticalThreshold` because we only care about one flag.

```suggestion
  /**
   * Retrieve an item's minimum attack roll.
   * @returns {number|null}  The minimum roll value that should be added to the attack roll formula.
   */
  getMinimumRoll() {
    if ( !this.hasAttack ) return null;
    return this.actor.flags.dnd5e?.minimumAttackRoll ?? null;
  }
```

module/documents/item.mjs
```diff
@@ -1292,6 +1311,7 @@ export default class Item5e extends Item {
       flavor: title,
       elvenAccuracy,
       halflingLucky: flags.halflingLucky,
+      minimum: minimum,
```
arbron, line 1314:
```suggestion
      minimum,
```

module/documents/item.mjs
```diff
@@ -1683,6 +1706,7 @@ export default class Item5e extends Item {
       chooseModifier: true,
       halflingLucky: this.actor.getFlag("dnd5e", "halflingLucky" ),
       reliableTalent: (this.system.proficient >= 1) && this.actor.getFlag("dnd5e", "reliableTalent"),
+      minimum: minimum,
```
arbron, line 1709:
```suggestion
      minimum,
```

arbron (changes requested):

szefo09:
> One thing to consider, and I'm not sure of a good solution at the moment, is if you have a ability check minimum defined (e.g. minimum 5 on dex checks), and you roll a skill that is normally rolled with that ability, but you change the ability in the configuration dialog. So if you roll a charisma-based Sleight of Hand check, right now it still applies the dex minimum, not whatever minimum charisma ability checks have.
> 
> Not sure the best way to fix that at the moment, it might be okay to just leave it to a later patch unless you have an idea.

Honestly I don't have a good fix for this currently, as the Formula would have to update after changing the Ability Modifier, to represent the Minimum defined for each of them.  
From the look of it, it would have to be done in `_onDialogSubmit` method in `d20-roll.mjs`. However I don't think I can manage that. 😕

arbron (approved):

szefo09:
I found a possible solution for the issue. Please let me know what you think about it!

Fyorl:
@szefo09 has created a rebased version at #1760

PR #1600, opened 2022-06-17
[#1506] Add a copy button next to Scalar variable name 

### LINKED ISSUE #1506: [Feature Request] Add a copy button next to Scalar variable name
Originally in GitLab by @sharkbruhaha

It would be a nice add to have a Copy button next to the name of the identifier for the Scale Value to help save some typing time, and prevent fat fingering. Or at least make the text able to be highlighted and copied.

Kulkodar:
Added a simple icon to copy the whole scalar variable.
copy-icon
ksnip_20220829-171713
ksnip_20220829-171840

~~There is currently no indication that the info is copied to the clipboard, this could be fixed using a notification. Should this also be added?~~

closes #1506

### DIFF
--- a/module/advancement/types/scale-value.mjs
+++ b/module/advancement/types/scale-value.mjs
@@ -262,7 +262,7 @@ export class ScaleValueConfig extends AdvancementConfig {
    */
   _onIdentifierHintCopy(event) {
     let data = this.getData();
-    navigator.clipboard.writeText(`@scale.${data.classIdentifier}.${data.previewIdentifier}`);
+    game.clipboard.copyPlainText(`@scale.${data.classIdentifier}.${data.previewIdentifier}`);
     game.tooltip.activate(event.target, {text: game.i18n.localize("DND5E.IdentifierCopied"), direction: "UP"});
   }
 
--- a/module/advancement/types/scale-value.mjs
+++ b/module/advancement/types/scale-value.mjs
@@ -252,7 +252,7 @@ export class ScaleValueConfig extends AdvancementConfig {
   activateListeners(html) {
     super.activateListeners(html);
     this.form.querySelector("input[name='data.title']").addEventListener("input", this._onChangeTitle.bind(this));
-    this.form.querySelector("i[name='identifier-hint-copy']").addEventListener("click", this._onIdentifyerHintCopy.bind(this));
+    this.form.querySelector(".identifier-hint-copy").addEventListener("click", this._onIdentifierHintCopy.bind(this));
   }
 
   /* -------------------------------------------- */
@@ -260,9 +260,10 @@ export class ScaleValueConfig extends AdvancementConfig {
   /**
    * Copies the full scale identifier hint to the clipboard.
    */
-  _onIdentifyerHintCopy() {
+  _onIdentifierHintCopy(event) {
     let data = this.getData();
     navigator.clipboard.writeText(`@scale.${data.classIdentifier}.${data.previewIdentifier}`);
+    game.tooltip.activate(event.target, {text: game.i18n.localize("DND5E.IdentifierCopied"), direction: "UP"});
   }
 
   /* -------------------------------------------- */
--- a/module/advancement/types/scale-value.mjs
+++ b/module/advancement/types/scale-value.mjs
@@ -261,7 +261,7 @@ export class ScaleValueConfig extends AdvancementConfig {
    * Copies the full scale identifier hint to the clipboard.
    */
   _onIdentifierHintCopy(event) {
-    let data = this.getData();
+    const data = this.getData();
     game.clipboard.copyPlainText(`@scale.${data.classIdentifier}.${data.previewIdentifier}`);
     game.tooltip.activate(event.target, {text: game.i18n.localize("DND5E.IdentifierCopied"), direction: "UP"});
   }
--- a/module/advancement/types/scale-value.mjs
+++ b/module/advancement/types/scale-value.mjs
@@ -259,6 +259,7 @@ export class ScaleValueConfig extends AdvancementConfig {
 
   /**
    * Copies the full scale identifier hint to the clipboard.
+   * @protected
    */
   _onIdentifierHintCopy(event) {
     const data = this.getData();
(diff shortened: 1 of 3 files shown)


sharkbruhaha:
thanks all!

module/advancement/scale-value.js
```diff
@@ -257,6 +257,17 @@ export class ScaleValueConfig extends AdvancementConfig {
...
     this.form.querySelector("input[name='data.title']").addEventListener("input", this._onChangeTitle.bind(this));
+    this.form.querySelector("i[name='identifier-hint-copy']").addEventListener("click", this._onIdentifyerHintCopy.bind(this));
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Copies the full scale identifier hint to the clipboard.
+   */
+  _onIdentifyerHintCopy() {
```
arbron, line 268:
```suggestion
  _onIdentifierHintCopy() {
```

module/advancement/scale-value.js
```diff
@@ -257,6 +257,17 @@ export class ScaleValueConfig extends AdvancementConfig {
   activateListeners(html) {
     super.activateListeners(html);
     this.form.querySelector("input[name='data.title']").addEventListener("input", this._onChangeTitle.bind(this));
+    this.form.querySelector("i[name='identifier-hint-copy']").addEventListener("click", this._onIdentifyerHintCopy.bind(this));
```
arbron, line 260:
```suggestion
    this.form.querySelector(".identifier-hint-copy").addEventListener("click", this._onIdentifierHintCopy.bind(this));
```

templates/advancement/scale-value-config.html
```diff
@@ -26,6 +26,7 @@
       </div>
       <p class="hint identifier-hint">
         {{{localize "DND5E.AdvancementScaleValueIdentifierHint" class=classIdentifier identifier=previewIdentifier}}}
+        <i name="identifier-hint-copy" class="far fa-clipboard"></i>
```
arbron, line 29: This should be a link so it properly highlights for the users:

```suggestion
        <a class="identifier-hint-copy" title="{{localize 'DND5E.IdentifierCopy'}}">
          <i class="far fa-clipboard"></i>
        </a>
```

I also added a `title`:

```
"DND5E.IdentifierCopy": "Copy Identifier",
```

module/advancement/scale-value.js
```diff
@@ -257,6 +257,17 @@ export class ScaleValueConfig extends AdvancementConfig {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Copies the full scale identifier hint to the clipboard.
+   */
+  _onIdentifyerHintCopy() {
+    let data = this.getData();
+    navigator.clipboard.writeText(`@scale.${data.classIdentifier}.${data.previewIdentifier}`);
```
arbron, line 270: Showing a notification when the copy occurs here would be helpful:

```suggestion
    const identifier = `@scale.${data.classIdentifier}.${data.previewIdentifier}`;
    navigator.clipboard.writeText(identifier);
    ui.notifications.info(game.i18n.format("DND5E.IdentifierCopied", { identifier }));
```

```
"DND5E.IdentifierCopied": "Identifier {identifier} copied to clipboard",
```

arbron (changes requested):

arbron:
Any make sure you rebase atop the current `v10-dev` branch.

module/advancement/scale-value.js
Varriount: Hm, rather than an entire notification for what is a relatively small operation, might something similar to what GitHub does be more appropriate?
image

I think just changing the icon temporarily to a check-mark would be sufficient. Plus, it requires less translation.

arbron: That would work.

You could to try and use the new [`TooltipManager`](https://foundryvtt.com/api/v10/classes/client.TooltipManager.html) in V10 to easily show a little balloon.

Kulkodar:
oh damit, i didn't notice. Thanks for the review, i will resolve the requested changes later this day.

module/advancement/scale-value.js
Fyorl: Core foundry uses a `ui.notifications.info` for its clipboard copied message for document IDs so I think it's ok (and more consistent) to do the same here.

module/advancement/types/scale-value.mjs
```diff
@@ -252,6 +252,18 @@ export class ScaleValueConfig extends AdvancementConfig {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Copies the full scale identifier hint to the clipboard.
+   */
+  _onIdentifierHintCopy(event) {
+    let data = this.getData();
+    navigator.clipboard.writeText(`@scale.${data.classIdentifier}.${data.previewIdentifier}`);
```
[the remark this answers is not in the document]
Kulkodar: good to know, i will fix this.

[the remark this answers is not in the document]
Fyorl: With the latest foundry release, there is a utility method `game.clipboard.copyPlainText` which can be used to perform this copy in a cross-browser compatible way that will also work in insecure contexts.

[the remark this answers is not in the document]
Kulkodar: ok i will change to code to use the utility method.

module/advancement/types/scale-value.mjs
```diff
@@ -252,6 +252,18 @@ export class ScaleValueConfig extends AdvancementConfig {
...
+    this.form.querySelector(".identifier-hint-copy").addEventListener("click", this._onIdentifierHintCopy.bind(this));
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Copies the full scale identifier hint to the clipboard.
+   */
+  _onIdentifierHintCopy(event) {
+    let data = this.getData();
```
Fyorl, line 264:
```suggestion
    const data = this.getData();
```

module/advancement/types/scale-value.mjs
```diff
@@ -252,6 +252,18 @@ export class ScaleValueConfig extends AdvancementConfig {
...
     super.activateListeners(html);
     this.form.querySelector("input[name='data.title']").addEventListener("input", this._onChangeTitle.bind(this));
+    this.form.querySelector(".identifier-hint-copy").addEventListener("click", this._onIdentifierHintCopy.bind(this));
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Copies the full scale identifier hint to the clipboard.
+   */
```
Fyorl, line 262:
```suggestion
  /**
   * Copies the full scale identifier hint to the clipboard.
   * @protected
   */
```

Fyorl (changes requested):

Fyorl (approved):

PR #1602, opened 2022-06-19
ci: add Github Actions to replace Gitlab-CI

akrigline:
This action setup has full parity with the previous gitlab ci configuration:

- Verifies that the `system.json` `version` and `download` fields are correct for the tag that triggers the build.
- Installs `npm` deps, using node 16
- Runs `npm run build` to build everything at once
- Zips up the same file list as is present in `gitlab-ci.yml`
- Attaches that zip file and the `system.json` file to a release which is generated for this tag

Simplifications from the Gitlab CI:
- We no longer need a hacky `/latest` workaround using a "generic package", as github provides a permalink to `foundryvtt/dnd5e/releases/latest` out of the box.

I tested this in my own fork:
- Failure because of `download` url: https://github.com/ElfFriend-DnD/dnd5e/actions/runs/2522296308
- Failure because of `version` mismatch: https://github.com/ElfFriend-DnD/dnd5e/actions/runs/2522300430
- Success: https://github.com/ElfFriend-DnD/dnd5e/actions/runs/2522327283
  - Created Release:  https://github.com/ElfFriend-DnD/dnd5e/releases/tag/release-408

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,102 @@
+name: Release Creation
+
+on:
+  release:
+    types: [published]
+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [12.x, 14.x, 16.x]
+    
+    steps:
+    - uses: actions/checkout@v3
+
+
+    # Set up our some variables for future use
+    # Adapted from https://github.community/t/how-to-get-just-the-tag-name/16241/7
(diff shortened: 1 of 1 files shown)


arbron (approved):

aaclayton:
Is the triggering condition for this pipeline pushing a tag? Or somthing else?

arbron:
It looks like it is pushing a tag in the format `release-*`.

Issue #1574, opened 2022-06-01
Transform and revert breaks token flags


aaclayton:
Originally in GitLab by @DJ.Addi

**Description**

When transforming a token, the flags are entirely ignored. This causes missing module settings for the token.

When reverting the transformation, the flags are restored using a recursive, non-differential update. This causes odd states where any flags modified on the transformed token are now merged with the flags of the original token.

**Environment**

- System version: 1.6.1
- FoundryVTT version: 9.269 on Electron for Windows
- Modules enabled: Bar Brawl 1.6.0

**Expected behavior**

What I would expect is for the transformation to simply copy (and restore) the flags from the prototype non-recursively. The revert transformation method already does this for unlinked tokens, so I don't see why linked token data is merged rather than overwritten.

Alternatively, a hook for the revert could be introduced to allow modules to customize the behavior (which is already possible for the transformation itself).

**Steps to reproduce**

1. Install the Bar Brawl module.
2. Create two actors (one original, one for the transformation).
3. Open the Prototype Token configuration for both, go to the Resources tab and change add resources to make sure that they have different amounts of resources. Save both prototype tokens.
4. Drop the token of the original actor onto the scene.
5. Drop the transformation actor onto the original actor sheet, check "Transform all linked tokens?" and click "Wild Shape". Observe how the displayed bars don't match either configuration (there may be bars depending on the defaults).
6. Revert the transformation. Observe how the restored token now has a mix of the original data and the transformed token data.

Fyorl:
Need to check if this was resolved as part of #1840

Issue #1139, opened 2021-06-04
Add preserve passive/inactive/temporary effects toggle when using polymorph/wild-shape


aaclayton:
Originally in GitLab by @juraj.rarita

Feature request for polymorph functionality:
When polymorphing add additional toggles that will let the user select which Active Effects(passive/inactive/temporary) would be transferred to the polymorphed form.

Usage:
Used when token polymorphs to another creature that can not use the Active Effects that original token had.

A specific example from DnD5e:
Druid player wearing Cloak of protection that gives +1 to AC via Passive Effect(no duration) that merges with druid new form upon wild-shape and as such should not affect new form. At the same time, the druid has a Bless or Bane spell cast on himself that adds Temporary Effect (10 round duration) that should still affect the new form.

In this example, you can see why it is needed to have all 3 Effect types to be toggleable separately.

I would also add that it would be great if users could select on specific effect bases but I am not sure if that would not be overkill. The reason being specifically Druid/Monk or Druid/Barbarian multiclass. Both add Unarmored defense, Barbarian adds Fast movement also - both very effectively added by Passive Effects and also affect the wild-shaped form. So in the example case described above, it would be needed to not inherit AC bonus from the cloak but inherit Unarmored Defense and Fast movement, all Passive Effects.

Another possible solution how to solve the situation with multiclass would be a toggle that would say: "Keep only Active effects which origin is not equipment". That would exclude only the cloak.

Found suggestion for disguise self here that also seems to request additional functionality to polymorph so maybe it can be merged in one update? :) https://gitlab.com/foundrynet/dnd5e/-/issues/1100

Issue #1100, opened 2021-05-04
Transformation Tool - No Stats option


aaclayton:
Originally in GitLab by @ESDictor

Would it be possible to add an option to the Transformation Tool that didn't alter any stats at all? It would just change the Token Name, Portrait Image, and Token Image. This would be for Alter Self, Disguise, etc.

Issue #767, opened 2020-11-08
Improve Wild Shape transformations by keeping the biography / equipment / features / spells of the original actor on the polymorphed character sheet, even if it is not customary to use those while in the transformed form.


aaclayton:
Originally in GitLab by @claytonhenry92

Currently, the Wild Shape preset of transformation options doesn't seem to include keeping biography, equipment, features, and spells. Not all of these things are always kept, but from my experience with a player who is a druid it seems like it would be more useful to have these things kept by default, and simply make it known when they don't apply to the beast form they chose. Particularly in regards to things like race and class features, and spells with continuous effects that were cast before transforming.

PR #1604, opened 2022-06-19
Added some feature to the polymorph mechanism

p4535992:
1) Add two new feature to polymorph settings:

- Keep Self : just changing image and size for spell like Alter Self, Disguise Self, ecc. (this should be the fix to #1139)
- Not Keep AE: do not transfer any active effect during the polymorph action
- Keep AE Only Origin Not Equipment : Keep only Active effects which origin is not equipment, "you want to be a a Bear? sure! but you lose all your equipment benefit" (this should be the fix to #767)

Issue Related: #1139 and #767

2) Add new button for Alter Self/Disguise Self on the Polymorph panel (Together with WildShape and Polymorph)

Issue related: #1100

3) Strange bug... possibly connected to this issue #1391 not sure if my solution is the correct one.
On this [line](https://gitlab.com/p4535992/dnd5e/-/blob/1.6.x/module/actor/entity.js#L1787) i made a backup of the token information because on the line of the constructor

`const newActor = await this.constructor.create(d, {renderSheet: true});`

I lose my width and height information and maybe others as well.
My solution is spartan, but it seems to work.

Issue related: #1391

Additional tickets to be addressed, (no solution for this for now):
https://github.com/foundryvtt/dnd5e/issues/1574


Fyorl:
Apologies that this PR has languished for so long, which means it's in need of another rebase. I tried to handle that for you, however, I noticed that many of the commits introduce linting issues and don't conform to this project's code style rules. It may well be that we introduced these rules after you had already submitted the original PR, so it's certainly not your fault, but it would make reviewing this much easier and less noisy if those could be cleaned up.

We have some information on code style in our [contributing guidelines](https://github.com/foundryvtt/dnd5e/blob/master/CONTRIBUTING.md#style). You should be able to run `npm run lint` to see any linting issues.

As part of preparing to review this, I tested the state of the polymorphing functionality in v10 and turned up a few bugs. You don't have to address them in this PR but I am noting them here in case we want to repurpose this PR for a general polymorphing tidy-up.

- #8120
- #8121
- #8122
- #8123
- #8124

p4535992:
For now i close this pr because i need to study fvtt10 and i wasn't happy on the code i made

PR #1608, opened 2022-06-23
[#1607] avoid setting NPC weapon proficiency/equipped if already set

### LINKED ISSUE #1607: NPC weapons cannot be created as "not proficient" via API
- Foundry v9, dnd5e 1.6.2

Explicitly setting `data.proficient` works for player characters, but not for NPCs.

```js
(async () => {
	const actor = await Actor.create({
		name: "Repro",
		type: "npc",
	});
	await actor.createEmbeddedDocuments(
		"Item",
		[
			{
				name: "Weapon",
				type: "weapon",
				data: {
					proficient: false,
				},
			},
		],
	);
	actor.items.contents[0].sheet.render(true);
	console.log(actor.items.contents[0].data.data.proficient);
})();
```

Expected: `false`

Actual: `true`

TheGiddyLimit:
If the user has specified these directly, then we should not clobber the existing values when creating the item. This matches the check already found in e.g. `_onCreateOwnedTool`

Closes #1607.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -2128,7 +2128,12 @@ export default class Item5e extends Item {
   _onCreateOwnedWeapon(data, isNPC) {
 
     // NPCs automatically equip items and are proficient with them
-    if ( isNPC ) return {"system.equipped": true, "system.proficient": true};
+    if ( isNPC ) {
+      const updates = {};
+      if ( !foundry.utils.hasProperty(data, "system.equipped") ) updates["system.equipped"] = true;
+      if ( !foundry.utils.hasProperty(data, "system.proficient") ) updates["system.proficient"] = true;
+      return updates;
+    }
     if ( data.system?.proficient !== undefined ) return {};
 
     // Some weapon types are always proficient


arbron:
@TheGiddyLimit Do you think you could rebase this and target the latest `v10-dev` branch?

module/item/entity.js
```diff
@@ -1907,8 +1907,8 @@ export default class Item5e extends Item {
 
     // NPCs automatically equip items and are proficient with them
     if ( isNPC ) {
-      updates["data.equipped"] = true;
-      updates["data.proficient"] = true;
+      if ( foundry.utils.getProperty(data, "data.equipped") === undefined ) updates["data.equipped"] = true;
+      if ( foundry.utils.getProperty(data, "data.proficient") === undefined ) updates["data.proficient"] = true;
```
arbron, line 1911: For v10 these will need to point to `system`, plus you can probably make it shorter using optional chaining rather than getProperty:

```suggestion
      if ( data.system?.equipped === undefined ) updates["system.equipped"] = true;
      if ( data.system?.proficient === undefined ) updates["system.proficient"] = true;
```

arbron (changes requested):

module/item/entity.js
TheGiddyLimit: The `getProperty` was to ensure that these two cases are equivalent:

```js
(async () => {
	const item1 = await Item.create({
		name: "Item 1",
		type: "weapon",
		// property path as string
		"system.proficient": true,
	});
	console.log(item1.system.proficient);

	const item2 = await Item.create({
		name: "Item 2",
		type: "weapon",
		// property path as nested objects
		system: {proficient: true},
	});
	console.log(item2.system.proficient);

	// ////////////////

	item1.sheet.render(true);
	item2.sheet.render(true);
})();
```

Is there some magic I'm missing here that renders that unnecessary?

arbron: You are correct, forgot that non-expanded data can appear in this updates :).

arbron (approved):

module/documents/item.mjs
```diff
@@ -1909,7 +1909,12 @@ export default class Item5e extends Item {
   _onCreateOwnedWeapon(data, isNPC) {
 
     // NPCs automatically equip items and are proficient with them
-    if ( isNPC ) return {"system.equipped": true, "system.proficient": true};
+    if ( isNPC ) {
+      const updates = {};
+      if ( foundry.utils.getProperty(data, "system.equipped") === undefined ) updates["system.equipped"] = true;
+      if ( foundry.utils.getProperty(data, "system.proficient") === undefined ) updates["system.proficient"] = true;
```
aaclayton, line 1915: Perhaps this might be a bit cleaner as:
```suggestion
      if ( foundry.utils.hasProperty(data, "system.equipped") ) updates["system.equipped"] = true;
      if ( foundry.utils.hasProperty(data, "system.proficient") ) updates["system.proficient"] = true;
```

aaclayton (changes requested):

module/documents/item.mjs
TheGiddyLimit: Sure would; cheers. Amended my commit

module/documents/item.mjs
```diff
@@ -2128,7 +2128,12 @@ export default class Item5e extends Item {
   _onCreateOwnedWeapon(data, isNPC) {
 
     // NPCs automatically equip items and are proficient with them
-    if ( isNPC ) return {"system.equipped": true, "system.proficient": true};
+    if ( isNPC ) {
+      const updates = {};
+      if ( !foundry.utils.getProperty(data, "system.equipped") ) updates["system.equipped"] = true;
```
aaclayton, line 2133: These checks need to be `hasProperty` instead of `getProperty` because the value may be `false`

aaclayton (changes requested):

module/documents/item.mjs
TheGiddyLimit: 🤦 fixed

aaclayton (approved):

PR #1614, opened 2022-06-25
[#1610] Apply temp HP from context menu on dice rolls

### LINKED ISSUE #1610: Apply Temp HP from right-clicking on a die roll
When you have a token selected and right-click on a die roll chat message, like a damage roll result, you get a context menu that allows you to apply damage and healing. There are a few damage options but there's no "Apply Temp HP" option. I think we should add that.

I have written this up as a world script in a test world already. If you're willing to accept it, I'll create a PR for it. Also we luck out that there's a user with a clock that makes for a great icon for it.

image

kaelad02:
Also adds an `applyTempHp` function to `Actor5e`. I could have just done that inside the chat function, but I figured it could be useful for macros and module developers to have the function.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -222,7 +222,7 @@
 "DND5E.Charges": "Charges",
 "DND5E.ChatContextDamage": "Apply Damage",
 "DND5E.ChatContextHealing": "Apply Healing",
-"DND5E.ChatContextTempHP": "Apply Temp HP",
+"DND5E.ChatContextTempHP": "Apply Temporary HP",
 "DND5E.ChatContextDoubleDamage": "Apply Double Damage",
 "DND5E.ChatContextHalfDamage": "Apply Half Damage",
 "DND5E.ChatFlavor": "Chat Message Flavor",
--- a/lang/en.json
+++ b/lang/en.json
@@ -222,6 +222,7 @@
 "DND5E.Charges": "Charges",
 "DND5E.ChatContextDamage": "Apply Damage",
 "DND5E.ChatContextHealing": "Apply Healing",
+"DND5E.ChatContextTempHP": "Apply Temp HP",
 "DND5E.ChatContextDoubleDamage": "Apply Double Damage",
 "DND5E.ChatContextHalfDamage": "Apply Half Damage",
 "DND5E.ChatFlavor": "Chat Message Flavor",
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -753,18 +753,20 @@ export default class Actor5e extends Actor {
     return allowed !== false ? this.update(updates, {dhp: -amount}) : this;
   }
 
+  /* -------------------------------------------- */
+ 
   /**
-   * Apply a certain amount of tempoary hit point, but only if it's more then the actor currently has.
+   * Apply a certain amount of temporary hit point, but only if it's more than the actor currently has.
    * @param {number} amount       An amount of temporary hit points to set
    * @returns {Promise<Actor5e>}  A Promise which resolves once the temp HP has been applied
    */
-  async applyTempHp(amount = 0) {
+  async applyTempHP(amount=0) {
     amount = parseInt(amount);
-    const hp = this.data.data.attributes.hp;
+    const hp = this.system.attributes.hp;
 
     // Update the actor if the new amount is greater than the current
     const tmp = parseInt(hp.temp) || 0;
-    return amount > tmp ? this.update({ "data.attributes.hp.temp": amount }) : this;
+    return amount > tmp ? this.update({"system.attributes.hp.temp": amount}) : this;
   }
 
   /* -------------------------------------------- */
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -753,6 +753,20 @@ export default class Actor5e extends Actor {
     return allowed !== false ? this.update(updates, {dhp: -amount}) : this;
   }
 
+  /**
+   * Apply a certain amount of tempoary hit point, but only if it's more then the actor currently has.
+   * @param {number} amount       An amount of temporary hit points to set
+   * @returns {Promise<Actor5e>}  A Promise which resolves once the temp HP has been applied
+   */
+  async applyTempHp(amount = 0) {
+    amount = parseInt(amount);
+    const hp = this.data.data.attributes.hp;
+
+    // Update the actor if the new amount is greater than the current
+    const tmp = parseInt(hp.temp) || 0;
+    return amount > tmp ? this.update({ "data.attributes.hp.temp": amount }) : this;
+  }
+
   /* -------------------------------------------- */
 
   /**
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -86,6 +86,12 @@ export function addChatMessageContextOptions(html, options) {
       callback: li => applyChatCardDamage(li, -1)
     },
     {
+      name: game.i18n.localize("DND5E.ChatContextTempHP"),
+      icon: `<i class="fas fa-user-clock"></i>`,
+      condition: canApply,
+      callback: li => applyChatCardTemp(li)
+    },
+    {
       name: game.i18n.localize("DND5E.ChatContextDoubleDamage"),
       icon: '<i class="fas fa-user-injured"></i>',
       condition: canApply,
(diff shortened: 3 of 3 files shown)


lang/en.json
```diff
@@ -218,6 +218,7 @@
 "DND5E.Charges": "Charges",
 "DND5E.ChatContextDamage": "Apply Damage",
 "DND5E.ChatContextHealing": "Apply Healing",
+"DND5E.ChatContextTempHP": "Apply Temp HP",
```
Varriount, line 221: This should probably be
```suggestion
"DND5E.ChatContextTempHP": "Apply Temporary HP",
```

module/actor/entity.js
```diff
@@ -915,6 +915,20 @@ export default class Actor5e extends Actor {
...
+   * @param {number} amount       An amount of temporary hit points to set
+   * @returns {Promise<Actor5e>}  A Promise which resolves once the temp HP has been applied
+   */
+  async applyTempHp(amount = 0) {
+    amount = parseInt(amount);
+    const hp = this.data.data.attributes.hp;
+
+    // Update the actor if the new amount is greater than the current
+    const tmp = parseInt(hp.temp) || 0;
+    return amount > tmp ? this.update({ "data.attributes.hp.temp": amount }) : this;
```
Varriount, line 929:
```suggestion
    const temp = parseInt(hp.temp) || 0;
    return amount > temp ? this.update({ "data.attributes.hp.temp": amount }) : this;
```

arbron:
@kaelad02 Think you could get this rebased on top of the `v10-dev` branch for potential inclusion in `2.0`?

kaelad02:
@arbron rebase done, thanks for considering this for v2.0

module/documents/actor/actor.mjs
```diff
@@ -753,6 +753,20 @@ export default class Actor5e extends Actor {
     return allowed !== false ? this.update(updates, {dhp: -amount}) : this;
   }
 
+  /**
+   * Apply a certain amount of tempoary hit point, but only if it's more then the actor currently has.
```
arbron, line 757:
```suggestion
   * Apply a certain amount of temporary hit point, but only if it's more than the actor currently has.
```

module/documents/actor/actor.mjs
```diff
@@ -753,6 +753,20 @@ export default class Actor5e extends Actor {
...
+   * @param {number} amount       An amount of temporary hit points to set
+   * @returns {Promise<Actor5e>}  A Promise which resolves once the temp HP has been applied
+   */
+  async applyTempHp(amount = 0) {
```
arbron, line 761: I think the HP in the method name should be capitalized to match methods like `_prepareNPCData`.

```suggestion
  async applyTempHP(amount=0) {
```

module/documents/actor/actor.mjs
```diff
@@ -753,6 +753,20 @@ export default class Actor5e extends Actor {
...
+    amount = parseInt(amount);
+    const hp = this.data.data.attributes.hp;
```
arbron, line 763:
```suggestion
    const hp = this.system.attributes.hp;
```

module/documents/actor/actor.mjs
```diff
@@ -753,6 +753,20 @@ export default class Actor5e extends Actor {
...
+
+    // Update the actor if the new amount is greater than the current
+    const tmp = parseInt(hp.temp) || 0;
+    return amount > tmp ? this.update({ "data.attributes.hp.temp": amount }) : this;
```
arbron, line 767:
```suggestion
    return amount > tmp ? this.update({"system.attributes.hp.temp": amount}) : this;
```

module/documents/chat-message.mjs
```diff
@@ -85,6 +85,12 @@ export function addChatMessageContextOptions(html, options) {
       condition: canApply,
       callback: li => applyChatCardDamage(li, -1)
     },
+    {
+      name: game.i18n.localize("DND5E.ChatContextTempHP"),
+      icon: `<i class="fas fa-user-clock"></i>`,
```
arbron, line 90:
```suggestion
      icon: '<i class="fas fa-user-clock"></i>',
```

module/documents/chat-message.mjs
```diff
@@ -120,6 +126,20 @@ function applyChatCardDamage(li, multiplier) {
   }));
 }
```
arbron, line 128:
```suggestion

/* -------------------------------------------- */
 
```

module/documents/chat-message.mjs
```diff
@@ -120,6 +126,20 @@ function applyChatCardDamage(li, multiplier) {
...
 
+/**
+ * Apply rolled dice as tempoary hit points to the controled token(s).
+ * @param {HTMLElement} li The chat entry which contains the roll data
```
arbron, line 131:
```suggestion
 * Apply rolled dice as temporary hit points to the controlled token(s).
 * @param {HTMLElement} li  The chat entry which contains the roll data
```

module/documents/chat-message.mjs
```diff
@@ -120,6 +126,20 @@ function applyChatCardDamage(li, multiplier) {
...
+ * @returns {Promise}
+ */
+function applyChatCardTemp(li) {
+  const message = game.messages.get(li.data("messageId"));
+  const roll = message.roll;
+  return Promise.all(canvas.tokens.controlled.map(t => {
+    const a = t.actor;
+    return a.applyTempHp(roll.total);
```
arbron, line 139:
```suggestion
    return a.applyTempHP(roll.total);
```

module/documents/chat-message.mjs
```diff
@@ -120,6 +126,20 @@ function applyChatCardDamage(li, multiplier) {
...
 }
 
+/**
+ * Apply rolled dice as tempoary hit points to the controled token(s).
+ * @param {HTMLElement} li The chat entry which contains the roll data
+ * @returns {Promise}
+ */
+function applyChatCardTemp(li) {
+  const message = game.messages.get(li.data("messageId"));
+  const roll = message.roll;
```
arbron, line 136: This throws a deprecation warning at the moment. Not sure what our overall strategy for these buttons when there are multiple rolls in a chat message will be, but for now you can just access the first roll directly:

```suggestion
  const roll = message.rolls[0];
```

arbron (changes requested):
Just a few changes :)

arbron (approved):

lang/en.json
aaclayton: Agree, should be "Temporary"

module/documents/actor/actor.mjs
```diff
@@ -753,6 +753,22 @@ export default class Actor5e extends Actor {
...
   }
 
+  /* -------------------------------------------- */
+ 
+  /**
+   * Apply a certain amount of temporary hit point, but only if it's more than the actor currently has.
+   * @param {number} amount       An amount of temporary hit points to set
+   * @returns {Promise<Actor5e>}  A Promise which resolves once the temp HP has been applied
+   */
+  async applyTempHP(amount=0) {
```
aaclayton, line 763: It seems this function is specifically intended for cases where the character may **gain** temporary HP. As such I think perhaps we should explicitly name it `gainTemporaryHP` and reject cases where `amount < 0`, or do I misunderstand?

module/documents/chat-message.mjs
```diff
@@ -85,6 +85,12 @@ export function addChatMessageContextOptions(html, options) {
       condition: canApply,
       callback: li => applyChatCardDamage(li, -1)
     },
+    {
+      name: game.i18n.localize("DND5E.ChatContextTempHP"),
+      icon: '<i class="fas fa-user-clock"></i>',
```
aaclayton, line 90: I'm not certain a clock is the most intuitive icon here.

aaclayton (changes requested):

module/documents/actor/actor.mjs
kaelad02: If you pass in a negative number, then `amount > tmp` will be false since it would be something like `-5 > 0` and it won't update anything.

module/documents/chat-message.mjs
kaelad02: The clock was meant to symbolize time. Temporary hit points expire at some point, so using the clock seemed appropriate. I also looked through the FA icons, the `user` ones anyways, and didn't see anything better IMO.

aaclayton: I think that's a much better choice!

kaelad02: Just that it's not consistent with the existing icons, that was my point. But it's not a hill I'm willing to die on

shemetz: That's a good point, the other four icons show a person silhouette while this does not.  Can you maybe take a screenshot of how each of them looks?  it's possible that using a clock is indeed better just for the consistency.

kaelad02: @shemetz Here's a screenshot with the `fa-user-clock` icon:

image

And one with the `fa-shield-heart` icon:

image

shemetz: so yeah, I think user-clock does fit the best for this.  the one for Double Damage is also slightly different but... oh well.

kaelad02: True, the double damage doesn't have an icon in the corner but it's clever IMO

Fyorl: I'd agree that a clock is reasonably intuitive to represent something that is 'temporary', and is consistent with our other `fa-user` icons. We can change it later very easily if people still feel strongly.

Fyorl (approved):

PR #1627, opened 2022-07-03
[#1401, #1620, #1626] Item Choice Advancement

### LINKED ISSUE #1401: Advancement Type: Item Choice
This advancement type is designed for any feature that gives the player a choice of options to add to their character at a certain level. This covers features with a limited set of options such as Fighting Style and more open ended features such as Magical Secrets.

## Level Up Interface

If a limited pool of options are provided, the interface will present the list with checkboxes so the player can select which options they would like. If any arbitrary items can be added, then the player is presented with a drop area where items can be dragged from the items sidebar or a compendium. Both a list of items and the drop area can be used at the same time.

It should be clear how many choices are available (Fighting Style only allows one, Magical Secrets lets you take two, etc.).

## Configuration

The `hint` string is a brief bit of text to assist players in what choices they can make. Typically this will be a portion of the feature's description and shouldn't be more than a single paragraph of one or two sentences.

The `choices` object indicates how many choices are available to the player and at which levels. A choice that can be taken only once such as a Fighter's Fighting Style would be simple as `{ 1: 1 }`, indicating that player should be presented with choosing a single style at first level. Something like Bard's Magical Secrets which gives the option of choosing two options at levels 10, 14, and 18 is shown below.

The boolean `allowDrops` indicates whether the drop area should be shown to the player to allow arbitrary items to be added.

Item types can be restricted using the `type` field, which is `null` if any item type is acceptable, or can be set to one of the top-level types (except things like class, subclass, and background).

Finally `pool` contains an array of item UUIDs indicating a fixed list of options presented to the player.

```javascript
{
  hint: "Choose two spells from any classes, including this one. A spell you choose must be of a level you can cast, as shown on the Bard table, or a cantrip.",
  choices: { 10: 2, 14: 2, 18: 2 },
  allowDrops: true,
  type: "spell",
  pool: null
}
```

## Value

The value object contains an entry for each level where a choice is offered. Each of these entries is another object containing the ID of the item on the actor as the key and the UUID of the chosen or dropped item as the value.

```javascript
{
  10: {
    "aonJ2YjkqkYB9WYB": "Compendium.dnd5e.spells.JLTQyqXEaJDrTXyW",
    "GBYN5rH4nh1ocRlY": "Compendium.dnd5e.spells.7UwUjJ6owIQkEPrs"
  }
}
```

## Future Directions

In the future, this can potentially be tied with a more advanced system for matching item types in compendiums. So the advancement would be able to present a list of all Metamagic options available to the player, regardless of which module introduced them.

### LINKED ISSUE #1620: Title text for delete item icon on Item Grant Advancement config isn't localized

### LINKED ISSUE #1626: Add spell configuration to ItemGrant advancement
The advancements for granting (#1400) and choosing items (#1401) should have the option to modify any spells added:

- Select preparation mode (regular, always prepared, pact magic, at will, innate)
- Change limited uses
- Specify spellcasting ability

These options should be part of these advancements, but the controls should only be visible if a spell is added to an Item Grant advancement or the "Spell" type is selected in an Item Choice advancement.

The interface would be similar to this mockup by [@MaxPat931](https://github.com/MaxPat931)

image

The configuration for these advancement types would gain something like this:

```javascript
{
  configuration: {
    spells: {
      ability: null,          // Don't change the spellcasting ability
      preparation: "innate",  // Set to innate spellcasting
      uses: {
        max: "1",
        per: "lr"             // One casting per long rest
      }
    }
  }
}
```

arbron:
Introduces the item choice advancement for choosing one or more options when you reach a level, usable for things like Fighting Style, Metamagic, or Signature Spells.

The player can select an option from a predefined list like Fighting Style:

[Selecting a Fighting Style](https://user-images.githubusercontent.com/19979839/177053144-13e843a7-ba35-4db6-9406-4d9d26b6146f.webm)

Or they can drop items onto the advancement such as with Magical Secrets:

[Dropping spells for Magical Secrets](https://user-images.githubusercontent.com/19979839/177053199-a75a73cf-3b03-428b-9ce1-1f4ee6e4653a.webm)

If a choice is available over multiple levels like Metamagic, the previously made choices will be shown:

[Selecting Metamagic with previous choices shown](https://user-images.githubusercontent.com/19979839/177053218-9bff4411-3274-498c-bbc8-2b0f55d71d75.webm)

Configuration is handled through an interface similar to Grant Items, but with the level list allowing to specify how many choices are available at which levels:

Item Choice configuration interface

This also adds the ability to configure spells that are added through Item Grant or Item Choice advancements, setting ability, preparation mode, and limited uses:

Item Grant - Spell Config

Resolves #1401
Resolves #1620 
Resolves #1626


MaxPat931:
Warlock Invocations provides the option to swap out previous choices for a new one (and Tasha's added a few Optional Class Features that allow for this as well), would it be feasible to build in a Modify Choices button for the Previous Choices view, or some other way to modify them after making the choice?

caprichozow:
Something else that could be added that'd be neat would be the option to better accommodate classes like the "Path of the Totem Warrior", giving us the option to set pre requirements on class advancements, using the Totem Warrior itself as an example on a setup is having so if you pick the Bear one at level 3, you'll automatically get the Bear one on 6th and 14th level, that'd decrease the number of prompts you'd have to go through while also locking away your options not enabling people to make the mistake of choosing Bear at the 3rd level and then say Eagle on the 6th and 14th.

Fyorl:
Rather than this being a new type of advancement, I believe we planned to add this functionality to the existing `ItemGrant` advancement. Conceptually they are the same: An item grant where you are given X options and must take all X of them, vs an item grant where you are given X options and must take Y of them.

arbron:
> Rather than this being a new type of advancement, I believe we planned to add this functionality to the existing `ItemGrant` advancement. Conceptually they are the same: An item grant where you are given X options and must take all X of them, vs an item grant where you are given X options and must take Y of them.

I believe there are enough differences, particularly with choices that span multiple levels. We'd either have to graft the level code onto `ItemGrant` where it doesn't have any use and have it exist for all item grants or force users to duplicate `ItemChoice` advancements for different levels, potentially leading to a large data duplication (see Warlock invocations which have dozens of entries across nearly as many levels).

Fyorl:
Is it not the case that `ItemChoice` is a strict superset of `ItemGrant` though? As in, any existing `ItemGrant` could be implemented as an `ItemChoice` instead? Maybe there would have to be some small tweaks to make that true, like a checkbox for 'optional' where, if unchecked, all the available items are granted instead of offering a choice.

It seems like for the Fighting Style example, which has only a single item in its level table, it could be equally implemented as an `ItemGrant` with `optional: true`, and an additional field like `choices: 1`.

If that's the case and they are interchangeable in this way, or where one is a strict superset of the other, then I don't think it makes sense to have both of them. It feels like it would be preferable to combine them in some way in order to keep code reuse to a maximum and reduce our API surface.

I think that's what our goal originally was with `ItemGrant`, which is why it has the `optional` field at all. But it seems we've deviated from that trajectory somewhat here.

Issue #1622, opened 2022-06-29
Rename backpack item to container


arbron:
Selecting an item named `"Backpack"` for things like bags of holding and component pouches is a bit odd. Changing it to a more generic `"Container"` would help clarify the item creation process. This would also help match to the character sheet where these items are listed under the `"Containers"` header.

This would not change the underling item type of `backpack`.

Issue #729, opened 2020-10-23
Native item container support


aaclayton:
Originally in GitLab by @saethone

Native support for item containers such as backpacks and belt pouches, with weight tracking and visibility of items from the item tab (including collapsing and expanding to view inside, instead of opening a separate item sheet as is the case currently through the support of another module)

(example of a backpack in pathfinder 2e system)
![image](https://gitlab.com/foundrynet/dnd5e/uploads/b0ff62d0681486e2b273276733f21ac7/image.png)

Ikaguia:
@arbron is this still in progress? Last update seems to be from july 2023. If you are busy with other stuff, I could try and see if I can do it

Fyorl:
> @arbron is this still in progress? Last update seems to be from july 2023. If you are busy with other stuff, I could try and see if I can do it

It is still in progress, and should be ready for 2.5, but thank you for the offer

PR #1633, opened 2022-07-10
[#1622] Rename "backpack" item type to "container"

djinnstar:
Potentially closes issue #1622 (rename backpack item to container)

- Refactors the old "backpack" item type with a new item type "container" - this reduces misnomers like a pouch being a "backpack"
- Updates all relevant compendia items to this new type

[This is my first dnd5e contribution - please review carefully!]
[Note: the actual "Backpack" item is preserved; only the "backpack" item type has been refactored to the new item type "container"]
[Packs were successfully recompiled; linting was unnecessary since syntax was preserved]

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3,7 +3,7 @@
 "ACTOR.TypeNpc": "Non-Player Character",
 "ACTOR.TypeVehicle": "Vehicle",
 "ITEM.TypeBackground": "Background",
-"ITEM.TypeBackpack": "Backpack",
+"ITEM.TypeBackpack": "Container",
 "ITEM.TypeClass": "Class",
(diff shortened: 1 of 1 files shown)


arbron:
Unfortunately there is no good way to alter an underlying type like this at the moment without causing issues with all worlds that have existing `backpack` items. There is theoretically a way to perform that migration, but Foundry core makes it a bit hacky.

I think our best bet is to just modify the localization keys for now to the new name, but keep the underlying type name as `backpack`.

djinnstar:
Sounds good! Rebased to v10 and kept just the localization name change.

arbron (approved):

djinnstar:
Closing - `container` should be a new type to answer #729.

PR #1636, opened 2022-07-13
Refactoring of Actor5e & ActorArmorConfig

arbron:
- Create new `_prepareAbilities` method to match the existing `_prepareSkills`
- Rename `_compute*` methods to `_prepare*` and take away their return values
- Reorder methods so that methods are logically grouped (base data preparation methods, derived data preparation methods) and in (mostly) alphabetical order
- Move `simplifyBonus` into utilities
- Move attunement calculation to `prepareBaseData` to allow it to be used in active effects (for 20th level Artificers)

### DIFF
--- a/module/apps/actor-armor.js
+++ b/module/apps/actor-armor.js
@@ -1,8 +1,18 @@
 /**
  * Interface for managing a character's armor calculation.
- * @extends {DocumentSheet}
  */
 export default class ActorArmorConfig extends DocumentSheet {
+  constructor(...args) {
+    super(...args);
+
+    /**
+     * Cloned copy of the actor for previewing change.s
+     * @type {Actor5e}
+     */
+    this.clone = this.object.clone();
+  }
+
+  /* -------------------------------------------- */
 
   /** @inheritdoc */
   static get defaultOptions() {
@@ -26,19 +36,16 @@ export default class ActorArmorConfig extends DocumentSheet {
 
   /** @inheritdoc */
   async getData() {
-    // Clone actor and get AC data
-    this.clone ??= this.object.clone();
     const ac = this.clone.system.attributes.ac;
 
-    // Get configuration data for the calculation mode
+    // Get configuration data for the calculation mode, reset to flat if configuration is unavailable
     let cfg = CONFIG.DND5E.armorClasses[ac.calc];
     if ( !cfg ) {
       ac.calc = "flat";
       cfg = CONFIG.DND5E.armorClasses.flat;
-      this.clone._prepareArmorClass();
+      this.clone.updateSource({ "system.attributes.ac.calc": "flat" });
     }
 
-    // Return context data
     return {
       ac,
       calculations: CONFIG.DND5E.armorClasses,
@@ -64,29 +71,8 @@ export default class ActorArmorConfig extends DocumentSheet {
   async _onChangeInput(event) {
     await super._onChangeInput(event);
 
-    const ac = this.clone.system.attributes.ac;
-
-    // Reference form data
-    ac.calc = this.form["ac.calc"].value;
-    const cfg = CONFIG.DND5E.armorClasses[ac.calc];
-    const enableFlat = ["flat", "natural"].includes(ac.calc);
-
-    // Handle changes to the calculation mode specifically
-    ac.formula = this.form["ac.formula"].value;
-    let flat = this.form["ac.flat"].value;
-    if ( event.currentTarget.name === "ac.calc" ) {
-      ac.formula = ac.calc === "custom" ? ac.formula : cfg.formula;
-      if ( enableFlat ) flat = ac.flat;
-    }
-
-    // Recompute effective AC
-    if ( enableFlat ) ac.flat = flat;
-    this.clone._prepareArmorClass();
-
-    // Update fields
-    this.form["ac.formula"].value = ac.formula;
-    this.form["ac.formula"].disabled = ac.calc !== "custom";
-    this.form["ac.flat"].value = enableFlat ? ac.flat : ac.value;
-    this.form["ac.flat"].disabled = !enableFlat;
+    // Update clone with new data & re-render
+    this.clone.updateSource({ [`system.attributes.${event.currentTarget.name}`]: event.currentTarget.value });
+    this.render();
   }
 }


module/apps/actor-armor.js
```diff
@@ -26,10 +26,12 @@ export default class ActorArmorConfig extends DocumentSheet {
 
   /** @inheritdoc */
   async getData() {
-
     // Get actor AC data
-    const actorData = this.object.toObject(false).system;
-    const ac = actorData.attributes.ac;
+    const ac = this.object.constructor.computeArmorClass(
```
aaclayton, line 30: I propose an alternative approach here. Instead of making `computeArmorClass` a method that can accept alternative inputs, I think the workflow for the `ActorArmorConfig` sheet should be to `clone()` the actor, apply changes to the clone via `updateSource()`, and then when the form is saved commit those changes back to the original actor.

This avoids us needing to have methods like `Actor#computeArmorClass` which operate on data other than the actor's own data which I think is preferable.

module/documents/actor.js
```diff
@@ -91,79 +92,51 @@ export default class Actor5e extends Actor {
...
-    this._prepareSkills(bonusData, checkBonus, originalSkills);
-
-    // Determine Initiative Modifier
-    this._computeInitiativeModifier(checkBonus, bonusData);
-
-    // Reset class scaling and compute scaling values
-    this._classes = undefined;
-    this._computeScaleValues();
+    // Prepare armor class data
+    this.system.attributes.ac = this.constructor.computeArmorClass(
```
aaclayton, line 124: I think this method should just be `_prepareArmorClass()` similar to other actor preparation helpers, enabled by the workflow change to clone the original actor as part of AC configuration.

module/documents/actor.js
```diff
@@ -91,79 +92,51 @@ export default class Actor5e extends Actor {
...
+      this.system.attributes.ac, this.itemTypes.equipment, this.getRollData({ deterministic: true })
+    );
+    this.armor = this.system.attributes.ac.equippedArmor ?? null;
+    this.shield = this.system.attributes.ac.equippedShield ?? null;
+    if ( this.system.attributes.ac.warnings ) this._preparationWarnings.push(...this.system.attributes.ac.warnings);
+
+    // Compute encumbrance, initiative, and scale values
+    this.system.attributes.encumbrance = this.constructor.computeEncumbrance(this.system, this.items, flags);
```
aaclayton, line 132: I don't see the need for `Actor.computeEncumbrance` to be a static method, can we go back to this being an instance method? Is there some other use case which necessitates this being called externally using arbitrary (non-instanced) data?

module/documents/actor.js
```diff
@@ -91,79 +92,51 @@ export default class Actor5e extends Actor {
...
+    this.system.attributes.init = this.constructor.computeInitiative(this.system, bonusData, flags);
+    this.system.scale = this.constructor.computeScaleValues(this.classes);
```
aaclayton, line 134: similar to `computeEncumbrance`, I would feel more at ease about this being an instance method since it operates on `this.classes`.

module/documents/actor.js
```diff
@@ -246,6 +219,21 @@ export default class Actor5e extends Actor {
 
   /* -------------------------------------------- */
 
+  /**
+   * Initialize derived AC fields for Active Effects to target.
+   * Mutates the system.attributes.ac object.
+   * @private
+   */
+  _prepareBaseArmorClass() {
```
aaclayton, line 227: A consideration: in the main foundry VTT code base we are moving towards using true private methods like `#prepareBaseArmorClass` in cases where we feel the method can/should be private, otherwise treating methods as `@protected` if they are underscored. Might that be a convention we want to employ here as well?

module/documents/actor.js
```diff
@@ -361,100 +385,32 @@ export default class Actor5e extends Actor {
...
+   * Mutates the value of the system.spells object.
    * @private
    */
-  _computeScaleValues() {
-    const scale = this.system.scale ??= {};
-    for ( const [identifier, cls] of Object.entries(this.classes) ) {
-      scale[identifier] = cls.scaleValues;
-      if ( cls.subclass ) scale[cls.subclass.identifier] = cls.subclass.scaleValues;
-    }
+  _computeSpellcastingProgression() {
```
aaclayton, line 401: A few comments here:
1. If something was a public method, we should add a deprecation warning for it and shift it to the bottom of section of the file, make sure to tag it as `@deprecated` for IDE hinting.
2. If something was not previously public like `_computeSpellcastingProgression`, I think we can (even should) remove it directly without any sort of backwards-compatible warning.

aaclayton (changes requested):
Refactor looks good @arbron, thanks! Take a look at my comments and let me know what you think.

module/documents/actor.js
arbron: While normally I'd agree, I think in the instance of these preparation methods we should still leave the deprecation warnings in. Because of the lack of hooks in 5e up to this point, the only way to enhance any of this logic has been to wrap these private methods, so I think it is preferable to avoid immediately bricking a whole bunch of modules.

module/apps/actor-armor.js
```diff
@@ -65,31 +64,28 @@ export default class ActorArmorConfig extends DocumentSheet {
...
+      ac.formula = ac.calc === "custom" ? ac.formula : cfg.formula;
       if ( enableFlat ) flat = ac.flat;
     }
 
     // Recompute effective AC
-    actorData = foundry.utils.mergeObject(actorData, {"system.attributes.ac": {calc, formula}});
-    if ( enableFlat ) actorData.system.attributes.ac.flat = flat;
-    ac = this.object._computeArmorClass(actorData.data);
+    if ( enableFlat ) ac.flat = flat;
+    this.clone._prepareArmorClass();
```
aaclayton, line 84: Rather than modifying the actor data directly and then calling `_prepareArmorClass` it should be better to instead compute an object of `changes` to apply to the clone and then call `this.clone.updateSource(changes)` - AC preparation for the clone will happen automatically.

module/documents/actor.js
```diff
@@ -86,84 +108,34 @@ export default class Actor5e extends Actor {
...
-    this.shield = ac.equippedShield || null;
-    if ( ac.warnings ) this._preparationWarnings.push(...ac.warnings);
+    const checkBonus = simplifyBonus(globalBonuses?.check, bonusData);
+    this._prepareAbilities(bonusData, globalBonuses, checkBonus, originalSaves);
+    this._prepareSkills(bonusData, globalBonuses, checkBonus, originalSkills);
+    this._prepareArmorClass();
+    this._prepareEncumbrance();
+    this._prepareInitiative(bonusData, checkBonus);
+    this._prepareScaleValues();
+    this._prepareSpellcasting();
```
aaclayton, line 138: This looks great, nice and clean!

aaclayton (changes requested):
This is nearly perfect in my view, I'd just like to see a slightly different approach for managing changes to the `clone` actor in AC config.

module/apps/actor-armor.js
```diff
@@ -26,23 +26,22 @@ export default class ActorArmorConfig extends DocumentSheet {
...
+    // Clone actor and get AC data
+    this.clone ??= this.object.clone();
+    const ac = this.clone.system.attributes.ac;
 
     // Get configuration data for the calculation mode
     let cfg = CONFIG.DND5E.armorClasses[ac.calc];
     if ( !cfg ) {
       ac.calc = "flat";
       cfg = CONFIG.DND5E.armorClasses.flat;
+      this.clone._prepareArmorClass();
```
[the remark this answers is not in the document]
arbron: But we are changing the calculation mode here, so we need to reprepare the AC.

Issue #1532, opened 2022-05-04
Add tool category to tool sheet header rather than generic "Tool"


arbron:
If a tool is selected within a certain category such as `"Musical Instrument"` or `"Artisans' Tools"`, that category should be displayed in the tool sheet header rather than the generic `"Tool"` that always displays now.

kaelad02:
Are you referring to the item type that's in the upper-right corner of the item sheet?

![image](https://user-images.githubusercontent.com/19502084/174918723-3af9c0f9-61ac-4bdf-a5b8-941e3629c8b8.png)

If so, then changing that sounds like a bad idea to me. Other item types have subtypes (e.g. equipment and weapons) and they show their subtypes to the left of the Rarity.

![image](https://user-images.githubusercontent.com/19502084/174918795-0110dbf5-c5d3-4c18-8b31-07f89f2ecb0f.png)

Updating the tool sheet to show the tool type like the second screenshot makes sense to me. It'd be consistent with what some of the other item types look like.

arbron:
This is the screenshot from the old MR created by @jeremyregnerus which I hope will eventually be recreated on GitHub:

![image](https://user-images.githubusercontent.com/19979839/175064302-4ec222b8-c01f-43aa-82dc-363cba2e82b9.png)

PR #1637, opened 2022-07-13
[#1532] Displaying toolType in tool sheet header + adding general toolType

djinnstar:
[possible solution to #1532]

This PR adds the tool's type [from among Artisan's, Musical Instrument, or Gaming Set] to the tool sheet's header. 

The first commit (3060096a825fa922b4cc80cea0914b0c9228284a) is just a few lines that accomplishes this goal, BUT it leaves a white space in the header for "general tools" not in those 3 categories. 

Removing the white space in the header would require either some sort of conditional logic in-place to remove the white space (kinda gross), or filling that space with a generic type. So:

The second commit (9adaf613adba901b8f5f15416b98f768c1b366be) adds a fourth toolType, called "General Tool", so that every tool has a toolType - so that space in the header is always filled. 

Warning: the second commit is opinionated™ - the SRD rules neither confirm nor deny that non-artisan, non-musical, non-gaming tools fall under an umbrella category, but this commit adds an umbrella category to make tool data more unified and asserts that the category is called "General Tool".

If people are opposed to this opinion, we could either keep just commit 1 (so no umbrella category) or we could rename the umbrella category to something more preferable, maybe like "Miscellaneous Tool".

### DIFF
--- a/templates/items/tool.html
+++ b/templates/items/tool.html
@@ -16,12 +16,15 @@
 
             <ul class="summary flexrow">
                 <li>
-                    <select name="system.rarity">
-                        {{selectOptions config.itemRarity selected=system.rarity blank=""}}
+                    {{lookup config.toolTypes data.toolType }}
+                </li>
+                <li>
+                    <select name="data.rarity">
+                        {{selectOptions config.itemRarity selected=data.rarity blank="&nbsp;"}}
                     </select>
                 </li>
                 <li>
-                    <input type="text" name="system.source" value="{{system.source}}" placeholder="{{ localize 'DND5E.Source' }}"/>
+                    <input type="text" name="data.source" value="{{data.source}}" placeholder="{{ localize 'DND5E.Source' }}"/>
                 </li>
             </ul>
         </div>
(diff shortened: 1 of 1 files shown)


arbron:
I don't think we should add the `"General Tool"` category. The tools in the tool list include top-level tools and tool groups, and there are specific situations where the groups provided like `"Artisans' Tools"` and `"Musical Instruments"` come into effect (aka, Bards get 3 musical instruments) that don't apply to a new `"General Tools"` category.

templates/items/tool.html
```diff
@@ -15,6 +15,9 @@ <h4 class="item-type">{{itemType}}</h4>
             </div>
 
             <ul class="summary flexrow">
+                <li>
+                    {{lookup config.toolTypes data.toolType }}
+                </li>
```
arbron, line 20: Rather than leaving a blank space for no tool, you can just revert to the default `"Tool"` label:

```suggestion
                <li>
                    {{#if data.toolType}}
                      {{lookup config.toolTypes data.toolType}}
                    {{else}}
                      {{localize "DND5E.ItemTypeTool"}}
                    {{/if}}
                </li>
```

arbron (changes requested):

djinnstar:
Yep, good point!

djinnstar:
Alright, removed the junk commit that updated toolTypes in compendia - should be good to go.

arbron (approved):

arbron (changes requested):
If you could rebase atop the `v10-dev` branch I'll try to get this into `2.0`.

djinnstar:
@arbron  -  not confident that I did this rebase correctly; could you take a look?

arbron:
> @arbron - not confident that I did this rebase correctly; could you take a look?

Something definitely didn't work right.

```
# Fetch latest changes from remote
git fetch upstream

# If you have a local v10-dev branch already, fast-forward it with upstream changes
git merge upstream/v10-dev

# Otherwise you'll need to create one
git branch v10-dev upstream/v10-dev

# Checkout the working branch
git checkout 1.7.x-toolType

# Rebase onto most recent v10-dev, fixing any merge issues
git rebase v10-dev

# Force push the changes back up to GitHub
git push --force origin 1.7.x-toolType
```

djinnstar:
The help was very kind! I believe it works now.

arbron (changes requested):
Okay, something is still being funky here with the rebase. You might just need to redo the commit entirely.

```
# Revert previous commit but keep changes
git reset HEAD~

# Now adjust the changes to `tool.hbs`

# And redo the commit
git add templates/items/tool.hbs
git commit -m "[#1532] Adds the tool type to the first bar for visibility"
```

djinnstar:
It seemed vastly easier to just remake a new MR: #1710

Issue #974, opened 2021-01-25
Add 'any' to DND5E.targetTypes


aaclayton:
Originally in GitLab by @jeremyregnerus

D&D 5e uses the language **_target_** often in their stat blocks, which encompasses a creature, object or location as a single defined target type. This is different than targeting a creature or an object exclusively. This target type is not currently included in DND5E.targetTypes.

PR #1638, opened 2022-07-15
[#974] Add Target to DND5E.targetTypes

djinnstar:
[Potentially solves #974]

Adds a generic "Target" to the targetTypes - "Target" here refers to the group of creatures and objects, i.e. "you throw a fire bolt at one creature or object" or "this spell affects one target". 

The actual fix is in the first commit - a single line that just adds the targetType.
The second commit updates all the SRD compendia packs with this option.
The third commit compiles the packs.


djinnstar:
> If you can rebase this to target the `v10-dev` branch and fix the merge issues then I'll push to get it into 2.0.

Done!

arbron (approved):

arbron (changes requested):
`config.js` should be `config.mjs` in the latest version.

Also, looks like you'll have to recompile `spells.db` because of some changes Atropos is making at the moment.

Varriount:
I feel like just "Target" is a bit *too* ambiguous - if this *only* refers to entities (creatures and objects), then it should be named something like "TargetEntity". If it also refers to areas, then "TargetAny" (or "TargetAnything") would be better.

djinnstar:
@arbron Thanks! Updated config.mjs and spells.db.

djinnstar:
> As a programmatic identifier, I feel like just `target` is a bit too ambiguous - if this _only_ refers to entities (creatures and objects), then it might be better to use something like `entity`. If it _also_ refers to areas, then `anything` would be better.
> 
> With regards to labels, "Creature or Object" or "Creature, Object, or Area" would also be my suggestions, respectively.

I agree that `entity` and `anything` can serve as better identifiers for an API to use, but they're also not real DnD terminology. Instead, DnD rules sometimes say "target" ambiguously to mean either "Creature or Object" or "Creature, Object, or Area", and we might want to preserve that. Ultimately we should get input from the higher-ups before deciding!

arbron (changes requested):
I think `target` is the right choice because of how it aligns with accepted 5e terminology.

Looks like `spells.db` will be needed to be rebuilt again atop the latest `v10-dev`, but other than that everything looks good.

aaclayton:
`target: target` conveys no useful information. I would really prefer we have a different term.

Varriount:
As a compromise, what about `anyTarget`?

As an aside, this kind of problem exposes some of the existing weaknesses in how targeting is handled - in the future, it might be better to allow multiple target categories to be selected, rather than a single one (I'm aware this would be a breaking change).

aaclayton:
I think some options I would be okay with are `single`, `creatureOrObject`, or `any`.

arbron (changes requested):
If you could rebase this atop `2.1.x` and switch to using a more descriptive key as discussed (I'm thinking `creatureOrObject`) then we could see about getting this moved along.

Fyorl:
Closed in favour of #2378

PR #1641, opened 2022-07-16
Slight reorganization of application files

arbron:
- Create `item` folder within `applications`, move `AbilityUseDialog` into it because it is only used by items, not actors
- Rename `actor-armor.mjs` to `armor-config.mjs, `actor-type.mjs` to `type-config.mjs`, `actor-flags.mjs` to `sheet-flags.mjs` to match with other files and their class names


aaclayton:
Looks like this hit a conflict with your other merge, but if you can sort that out we can get these changes merged in no problem.

arbron:
> Looks like this hit a conflict with your other merge, but if you can sort that out we can get these changes merged in no problem.

That's the trouble with including the compiled javascript in the repo.

PR #1644, opened 2022-07-17
Rewrite gulpfile in ESM format, update release workflow to use proper compiled output

arbron:
- No longer includes compiled `dnd5e.mjs` in repo
- Now does the compilation during the release build process
- Adds source maps to built output
- Reworks gulpfile and utils using ES modules


aaclayton (approved):
Works for me, thanks @arbron

Issue #1082, opened 2021-04-15
Additions to Actor proficiencies should automatically update current owned Items which should now be flagged as proficient.


aaclayton:
Originally in GitLab by @denisPoifol

What happened:
If my character is equiped with a leather armor and its proficiency to light armors is not yet set, when setting it the armor is still marked as "not proficient"

What I expected to happen:
Ideally setting light armor proficiency to the character would make him proficient with any light armor he owns

Additional notes:

This is also true when :
- dealing with a weapon
- removing proficiency and the object was marked as proeficient

System version:
Foundry v0.7.9
D&D5e v1.2.4

djinnstar:
I think I have a working commit that just needs refactoring, JSDoc annotations, etc. - should be ready to submit the PR in a day or so!

arbron:
Based on discussion in the original pass at this, we decided that rather than storing proficiency on the items, we would automatically calculate it during `prepareData` while adding a `proficiencyMode` property that allows an item to be set to `"Automatic"`, `"Always Proficient"`, or `"Never Proficient"`.

This solution is currently blocked by #774

djinnstar:
We might want a bit more flexibility depending on the final solution of #774. `“Automatic`, `“Always”`, and `“Never”` don’t allow the user to specify double/half proficiency, and it might be better to have `proficiencyMode` be a value from an enumeration, maybe something like `config.dnd5e.proficiencyValues.{expertise|half|full|none|auto}`. It would also give devs an easy way to modify the options and add something else (maybe like 1/3rd proficiency - nonsense in SRD RAW but maybe something a user might want to do).

Issue #774, opened 2020-11-08
Expand the Tool Proficiencies in the actor traits to encompass expertise


aaclayton:
Originally in GitLab by @Kandashi

The current Armour/Weapon/Tool proficiency lists only currently allow for boolean values, which works fine for armour or weapons but donst fully encompass the Tool proficiencies. 

Expanding this system to include procifiency levels (half/expertise) with tools (and potentially weapons) would allow for more accurate representation of the actors stas (and allow for custom weapons that allow for expertise)

PR #1646, opened 2022-07-18
[#1082] Add automatic proficiency updates for items when the owning actor's proficiency list changes

djinnstar:
[Potentially solves #1082]

Whenever _onUpdate is triggered on an Actor, update the proficiencies of any items related to the original update. 

Note: There might be a better place to attach this than "_onUpdate", but I couldn't find a less frequent method that gets called every time the actor's proficiencies change. The current slowdown is small (checks if weaponProf, armorProf, or toolProf is included in every update), but there might be a more efficient location for this.

Note: When we get working tool proficiencies in 1.8.x (see #774), this section should be updated to handle non-binary tool proficiencies like expertise and half-proficiency.

### DIFF
--- a/module/actor/entity.js
+++ b/module/actor/entity.js
@@ -2060,6 +2060,28 @@ export default class Actor5e extends Actor {
   _onUpdate(data, options, userId) {
     super._onUpdate(data, options, userId);
     this._displayScrollingDamage(options.dhp);
+    this._updateOwnedProficiencies(data);
+  }
+
+  /* -------------------------------------------- */
+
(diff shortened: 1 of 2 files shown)


arbron:
This was like my original attempt to resolve this issue, but after discussion we decided to move in a different direction of automatically determining the proficiency in `prepareData`. Unfortunately some of that conversation got lost in the transition over from GitLab.

I have a branch that implements the automatic proficiency, but it is held up because we don't have a method for representing tool expertise on a character sheet.

PR #1649, opened 2022-07-20
Fix several bugs in item grant advancement

arbron:
- Fix issue with item grant choices not appearing in advancement list or advancement config due to `TextEditor._createContentLink` returning node rather than raw HTML
- Fix issue with advancement values not being unset when leveling down due to `mergeObject` no long performing deletions by default
- Fix issue with granted item ID not being stored in advancement data

### DIFF
--- a/module/advancement/advancement.mjs
+++ b/module/advancement/advancement.mjs
@@ -259,8 +259,8 @@ export default class Advancement {
     const advancement = foundry.utils.deepClone(this.item.system.advancement);
     const idx = advancement.findIndex(a => a._id === this.id);
     if ( idx < 0 ) throw new Error(`Advancement of ID ${this.id} could not be found to update`);
-    foundry.utils.mergeObject(this.data, updates);
-    foundry.utils.mergeObject(advancement[idx], updates);
+    foundry.utils.mergeObject(this.data, updates, { performDeletions: true });
+    foundry.utils.mergeObject(advancement[idx], updates, { performDeletions: true });
     this.item.updateSource({"system.advancement": advancement});
     return this;
   }
--- a/module/advancement/types/item-grant.mjs
+++ b/module/advancement/types/item-grant.mjs
@@ -71,19 +71,17 @@ export class ItemGrantAdvancement extends Advancement {
     for ( const [uuid, selected] of Object.entries(data) ) {
       if ( !selected ) continue;
 
-      const ri = retainedData[uuid];
-      if ( ri && !ri._id ) ri._id = foundry.utils.randomID();
-      const item = ri ? new Item.implementation(ri) : (await fromUuid(uuid))?.clone();
-      if ( !item ) continue;
-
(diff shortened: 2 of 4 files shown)


module/advancement/advancement.mjs
```diff
@@ -259,8 +259,8 @@ export default class Advancement {
     const advancement = foundry.utils.deepClone(this.item.system.advancement);
     const idx = advancement.findIndex(a => a._id === this.id);
     if ( idx < 0 ) throw new Error(`Advancement of ID ${this.id} could not be found to update`);
-    foundry.utils.mergeObject(this.data, updates);
-    foundry.utils.mergeObject(advancement[idx], updates);
+    foundry.utils.mergeObject(this.data, updates, { performDeletions: true });
+    foundry.utils.mergeObject(advancement[idx], updates, { performDeletions: true });
     this.item.updateSource({"system.advancement": advancement});
```
aaclayton, line 264: It *should* now in V10 be possible to do this without merging first:
```js
this.item.updateSource({[`system.advancement.${idx}`]: updates});
```

module/advancement/types/item-grant.mjs
```diff
@@ -71,19 +71,17 @@ export class ItemGrantAdvancement extends Advancement {
     for ( const [uuid, selected] of Object.entries(data) ) {
       if ( !selected ) continue;
 
-      const ri = retainedData[uuid];
-      if ( ri && !ri._id ) ri._id = foundry.utils.randomID();
-      const item = ri ? new Item.implementation(ri) : (await fromUuid(uuid))?.clone();
-      if ( !item ) continue;
-
-      item.updateSource({
+      const itemData = retainedData[uuid] ?? (await fromUuid(uuid))?.clone().toObject();
```
aaclayton, line 74: You can clean this up a bit by passing in the extra data as part of the `clone` operation:
```js
let itemData = retainedData[uuid];
if ( !itemData ) {
  const source = await fromUuid(uuid);
  itemData = source.clone({
    _id: foundry.utils.randomID(),
    "flags.dnd5e.sourceId": uuid,
    "flags.dnd5e.advancementOrigin": `${this.item.id}.${this.id}`
  }, {keepId: true}).toObject();
}
```

aaclayton (changes requested):
Two comments for you to investigate.

module/advancement/advancement.mjs
arbron: That seems to be saving it as an array, not an object. Does this only work with `DataModel` enforced types?

aaclayton: > That seems to be saving it as an object, not an array. Does this only work with `DataModel` enforced types?

Ah, yes. My mistake. It only works when the data model specifically defines an `ArrayField` type. Sorry, maybe one for the future!

aaclayton (approved):

PR #1650, opened 2022-07-20
[#1530] Fix bug preventing remarkable athlete from being used when non-default ability is selected with skill checks

### LINKED ISSUE #1530: Remarkable Athlete flag doesn't contribute to skill roll if normally non-applicable skill is rolled with different ability
The new system for rolling skill checks with alternate skills doesn't take Remarkable Athlete into account. Because the proficiency bonus from that flag is calculated during data preparation, changes to the ability use with a skill in the roll dialog doesn't alter how the roll is performed. So if you roll a `"Intimidation (Strength)"` check, and you are not proficient in Intimidation, it doesn't add the half proficiency from Remarkable Athlete despite this being a strength check. Conversely, rolling a `"Stealth (Charisma)"` check doesn't remove said proficiency.

arbron:
Resolves #1530

### DIFF
--- a/module/dice/d20-roll.mjs
+++ b/module/dice/d20-roll.mjs
@@ -11,7 +11,9 @@
  *                                               compared
  * @param {boolean} [options.elvenAccuracy=false]      Allow Elven Accuracy to modify this roll?
  * @param {boolean} [options.halflingLucky=false]      Allow Halfling Luck to modify this roll?
+ * @param {boolean} [options.jackOfAllTrades=false]    Allow Jack of All Trades to modify this roll?
  * @param {boolean} [options.reliableTalent=false]     Allow Reliable Talent to modify this roll?
+ * @param {boolean} [options.remarkableAthlete=false]  Allow Remarkable Athlete to modify this roll?
  */
 export default class D20Roll extends Roll {
   constructor(formula, data, options) {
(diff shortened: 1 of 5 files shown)


module/dice/d20-roll.mjs
```diff
@@ -251,7 +253,21 @@ export default class D20Roll extends Roll {
     // Customize the modifier
     if ( form.ability?.value ) {
       const abl = this.data.abilities[form.ability.value];
+      let prof = this.data.baseProf;
+
+      // Make proficiency adjustments based on remarkable athlete
+      if ( this.options.remarkableAthlete ) {
```
aaclayton, line 259: I don't like that something as specific as remarkable athlete interjects itself into the fundamental rolling workflow for every `D20Roll` throughout the entire system. I know that for specific types of checks we handle remarkable athlete further upstream. Is there any other way to approach this? I am uncomfortable with this design.

aaclayton (changes requested):
I'm uncomfortable with this change from a design perspective. We have a funnel of logic which progressively coalesces gameplay mechanics and character data down into executed and evaluated dice rolls.

I think a very character-centric and mechanical concept like "remarkable athlete" needs to be higher up that funnel and by the time we get to making a `D20Roll` that sort of logic should have already been handled and baked into the inputs provided to the d20.

Can we solve the referenced bug in a different way by shifting this into the `rollSkill` method specifically?

module/dice/d20-roll.mjs
arbron: Hmm, we are already swapping out bonuses when the ability is changed.

I suppose we could add a hook here to allow skills to customize the roll.

arbron:
Closing because this is going to be resolved during the roll pipeline refactor #2138

PR #1655, opened 2022-07-21
Fix v10 macro & token migration issues

arbron:
- Adjust macro export to match filename (`dnd5e.documents.macro` rather than `dnd5e.documents.macros`)
- Add compatibility exports for `game.dnd5e.rollItemMacro` & `game.dnd5e.macros`
- Add migration to convert existing macros from using the old paths to the new paths
- Fix issue with migrations being unable to be created by dropping on macro bar
- Fix issue with token image migrations pointing to old `img` rather than `texture.src`
- Fix a few instances where `logComptabilityWarning` were provided with `from` rather than `since`


aaclayton (approved):
Great fixes, thanks @arbron!

PR #1656, opened 2022-07-22
Additional V10 bug fixes

arbron:
- Move `simplifyRollFormula` and its helper methods into their own file
- Fixed compatibility warning in `AbilityUseDialog`
- Fixed compatibility warning in `createScrollFromSpell`
- Fixed compatibility warnings in `transformInto`
- Fix bug with displaying scrolling damage numbers
- Fix bug preventing spells from being upcast
- Fix bug with `highlightCriticalSuccessFailure` for formulas without dice


module/dice/_module.mjs
```diff
@@ -1,3 +1,4 @@
 export {default as D20Roll} from "./d20-roll.mjs";
 export {default as DamageRoll} from "./damage-roll.mjs";
 export * from "./dice.mjs";
+export {default as simplifyRollFormula} from "./simplify-roll-formula.mjs";
```
aaclayton, line 4: I don't really understand the reason to move this, but it's not a problem necessarily. It felt to me like simplification of a roll formula belonged in the `dice` module with other dice-related helpers.

aaclayton (approved):
Thanks @arbron

Issue #1140, opened 2021-06-06
Automatic doubling of dice for critical hits does not work if the number of dice rolled (or the denomination of those dice) is a parenthetical expression.


aaclayton:
Originally in GitLab by @estenioss96

Hello,

I use this equation in the damage formula of the unarmed strike of a Monk Pc: 1d(4+floor((@classes.monk.levels+1)/6)*2).
When it is a critical hit is not doubling the die:

![image](https://gitlab.com/foundrynet/dnd5e/uploads/e9fe9acf8567d726453fca21578c61bb/1.PNG)
![image](https://gitlab.com/foundrynet/dnd5e/uploads/301e0f753323909b6d7f36d976ab634e/2.PNG)

Foundry 0.8.6
dnd5e 1.33
0 Modules

PR #1659, opened 2022-07-23
[#1140] Fix rolling criticals when number of dice is a formula

arbron:
Attempting to fix the issue with rolling critical damage when a parenthetical or math expression is used:

1. `(@details.level)d6` ==> `5d6`
2. `floor(@details.level / 2)d6` ==> `2d6`

It also evaluates properly if a deterministic formula is used for the dice type:

4. `1d(4+floor((@classes.monk.levels+1)/6)*2)` ==> `1d6`
5. `1d(@prof)rr=1` ==> `1d3rr=1`

It does not attempt to simplify things if dice are used on either side of the expression:

6. `(1d4)d6` ==> Nope
7. `1d(2 * 1d10)` ==> Nope

### DIFF
--- a/module/dice/damage-roll.mjs
+++ b/module/dice/damage-roll.mjs
@@ -12,6 +12,7 @@
 export default class DamageRoll extends Roll {
   constructor(formula, data, options) {
     super(formula, data, options);
+    if ( !this.options.preprocessed ) this.preprocessFormula();
     // For backwards compatibility, skip rolls which do not have the "critical" option defined
     if ( (this.options.critical !== undefined) && !this.options.configured ) this.configureDamage();
   }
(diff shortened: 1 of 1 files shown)


aaclayton (commented):
The code itself looks good. I *think* the logic is okay, but dice manipulation is a treacherous space with lots of unexpected edge cases and pitfalls. I don't see a reason not to merge this aside from my general unease that doing dice related stuff is hard. 

One change I would like to see before merging is to separate out this parenthetical pre-processing step to a helper function in the `DamageRoll` class instead of putting it all inline. 

For code organization we should separate pre-processing from the standard "this is how we do critical hits" logic.

aaclayton (approved):

Issue #1543, opened 2022-05-11
Add options object to `item.roll()` to disable different consumptions


aaclayton:
Originally in GitLab by @krbz999

The `item.roll()` function has the `configureDialog` key in its options, and it would be good if this could be paired with an additional object that explicitly "unticks" options such as placing a template, or consuming limited uses, spell slots, or resources.

This also helps with with magical weapons that commonly appear with charges, where the configuration dialog is usually more of a hassle than it's worth (magical weapons, despite having charges, are used for attacking more often than anything else).

Currently the recommendation is to place the charges elsewhere than on the weapons, which is a workaround but not ideal.

Issue #1002, opened 2021-02-16
Allow parts of the item use workflow to be hooked and/or overridden.


aaclayton:
Originally in GitLab by @clay.sweetser

Currently modules like Midi-QoL, Item Macro, Better Rolls, etc. have to either patch the `Item5e.roll` method or the actor sheet HTML in order to implement their functionality. The former technique doesn't work well when multiple modules use, while the latter technique requires sheet-specific HTML modification.

Ideally, the 5e system would support arbitrary overriding and hooking for parts of the item usage workflow, so that such measures wouldn't be required. I don't believe just adding hooks would be sufficient, as there are a number of modules that modify how a workflow is carried out.

A more extended idea of what I mean is described in [this gist](https://gist.github.com/Varriount/c9ab3ca5abc7df017f525836f5e66531). I didn't know whether it would appropriate to post it here, since it's rather long.

PR #1662, opened 2022-07-24
[#1002, #1543, #1677] Add hooks to `Item5e#roll` & `Item5e#displayCard` to allow modules to enhance that workflow

arbron:
Adds a number of hooks to the item usage process to allow modules to better customize that workflow without having to override or wrap system methods:

- `dnd5e.preUseItem`: Called before the item is configured, allows changing configuration data, controlling whether the dialog should be displayed, and upcasting spells even if the dialog isn't shown
- `dnd5e.preItemUsageConsumption`: Called before usage updates are calculated
- `dnd5e.itemUsageConsumption`: Called after resource consumption is calculated to allow any modifications before those changes are applied to the actor or other items
- `dnd5e.preDisplayCard`: Called before the chat card is created with the chat data that will be used to display the card to allow for any changes before the creation occurs
- `dnd5e.displayCard`: Called after the chat card is created to allow for any changes now that the card is saved to the database
- `dnd5e.useItem`: Called after consumption updates, the chat card is created, and any measured template is created to respond to the effects of the usage (e.g. auto applying damage to all creatures within the template)

This also makes the process for placing a measured template async so that it can be awaited and sets the creating item's UUID as a flag so their origin can be tracked.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -649,7 +649,7 @@ export default class Item5e extends Item {
    * @property {boolean} consumeQuantity         Should the item's quantity be consumed?
    * @property {boolean} consumeRecharge         Should a recharge be consumed?
    * @property {boolean} consumeResource         Should a linked (non-ammo) resource be consumed?
-   * @property {string|null} consumeSpellLevel   Specific spell slot to consume.
+   * @property {number|string|null} consumeSpellLevel  Specific spell level to consume, or "pact" for pact level.
    * @property {boolean} consumeSpellSlot        Should any spell slot be consumed?
    * @property {boolean} consumeUsage            Should limited uses be consumed?
    * @property {boolean} needsConfiguration      Is user-configuration needed?
@@ -693,7 +693,7 @@ export default class Item5e extends Item {
       consumeQuantity: is.uses?.autoDestroy ?? false,
       consumeRecharge: !!is.recharge?.value,
       consumeResource: !!resource.target && (!item.hasAttack || (resource.type !== "ammo")),
-      consumeSpellLevel: requireSpellSlot ? is.preparation.mode === "pact" ? "pact" : `spell${is.level}` : null,
+      consumeSpellLevel: requireSpellSlot ? is.preparation.mode === "pact" ? "pact" : is.level : null,
       consumeSpellSlot: requireSpellSlot,
       consumeUsage: !!is.uses?.per
     };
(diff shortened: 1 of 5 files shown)


module/documents/item.mjs
```diff
@@ -642,101 +642,169 @@ export default class Item5e extends Item {
...
+    /**
+     * A hook event that fires before an item roll is configured.
+     * @function dnd5e.preRollConfiguration
+     * @memberof hookEvents
+     * @param {Item5e} item                      Item being rolled.
+     * @param {ItemRollConfiguration} config     Configuration data for the item roll being prepared.
+     * @param {ItemRollOptions} options          Additional options used for configuring item rolls.
+     * @returns {boolean}                        Explicitly return false to prevent item from being rolled.
+     */
+    if ( Hooks.call("dnd5e.preRoll", item, config, options) === false ) return;
```
aaclayton, line 717: I think this hook name is possibly too general. The method here is effectively "play an item card to chat" - the method is called `roll` because in early versions of `dnd5e` that is typically what resulted from this workflow.... but a side comment is that perhaps `roll` is not the right name for this method either.

Regardless, I think creating hooks for `dnd5e.preRoll` or `dnd5e.roll` are too easily confused with functionality that might apply at the time that an actual dice roll occurs rather than as part of the item chat card.

aaclayton (changes requested):
Looks good, thank you @arbron - I would like to see us tighten up our naming conventions a bit with hooks and methods. 

Also, as a separate request: it would be helpful for some topics which are conceptually distinct to be separate MRs. The AbilityTemplate changes could have been reviewed and merged separately from the item chat card rolling stuff.

module/documents/item.mjs
arbron: How about `dnd5e.preUseItem` & `dnd5e.useItem` (and other related hook changes) to match the term of `_getUsageUpdates`?

arbron:
@aaclayton With the new hook names, should I also create a new preferred `Item5e#use` method and indicate it should be used over `roll`?

aaclayton:
> @aaclayton With the new hook names, should I also create a new preferred `Item5e#use` method and indicate it should be used over `roll`?

I'm open to your preferences here, but I do think there is a good argument in favor of migrating to `Item5e#use` with a backwards-compatible deprecation for `Item5e#roll`

less/actors.less
```diff
@@ -501,10 +501,10 @@
     .item {
       .item-name {
         cursor: pointer;
-        &.rollable:hover .item-image {
+        &:is(.rollable, .usable):hover .item-image {
```
aaclayton, line 504: Can we move to using `.usable` directly rather than needing to support both `.rollable` and `.usable`? OR can we leave CSS alone here and just keep using `.rollable` and make this just an JS API change for now?

arbron:
> I'm open to your preferences here, but I do think there is a good argument in favor of migrating to `Item5e#use` with a backwards-compatible deprecation for `Item5e#roll`

Renamed and deprecated. I also adjusted the css class from `rollable` to `usable`, but left backwards compatibility so we don't break 3rd party sheets.

aaclayton (changes requested):
When I suggested switching to `Item5e#use` I was envisioning a JS-only change. Seeing all the CSS changes makes me immediately regret that and want to switch back. Can we avoid having `:is(.rollable, .usable)` everywhere which is ugly.

aaclayton:
> Renamed and deprecated. I also adjusted the css class from `rollable` to `usable`, but left backwards compatibility so we don't break 3rd party sheets.

We are *very likely* already breaking 3rd party sheets so I would prefer to lean in rather than just volunteer to take on future tech debt.

less/actors.less
arbron: I'll just leave the CSS alone for now.

aaclayton (approved):
Thank you!

PR #1670, opened 2022-07-26
[#1002] Add hooks to `Actor5e` rolling methods

arbron:
Adds a number of hooks to the actor rolling process to allow modules to better customize rolling without having to override or wrap system methods:

- `dnd5e.preRollSkill` & `dnd5e.rollSkill`
- `dnd5e.preRollAbilityTest` & `dnd5e.rollAbilityTest`
- `dnd5e.preRollAbilitySave` & `dnd5e.rollAbilitySave`
- `dnd5e.preRollDeathSave`: Modify normal death save parameters and set the target & critical numbers.
- `dnd5e.rollDeathSave`: Modify or prevent actor changes based on the result of a death save.
- `dnd5e.preRollHitDie` & `dnd5e.rollHitDie`: Modify or prevent actor changes based on a hit die roll.

### DIFF
--- a/module/dice/dice.mjs
+++ b/module/dice/dice.mjs
@@ -3,44 +3,56 @@
 /* -------------------------------------------- */
 
 /**
+ * Configuration data for a D20 roll.
+ *
+ * @typedef {object} D20RollConfiguration
+ *
+ * @property {string[]} [parts=[]]  The dice roll component parts, excluding the initial d20.
+ * @property {object} [data={}]     Data that will be used when parsing this roll.
+ * @property {Event} [event]        The triggering event for this roll.
+ *
+ * ## D20 Properties
+ * @property {boolean} [advantage]     Apply advantage to this roll (unless overridden by modifier keys or dialog)?
+ * @property {boolean} [disadvantage]  Apply disadvantage to this roll (unless overridden by modifier keys or dialog)?
+ * @property {number} [critical=20]    The value of the d20 result which represents a critical success.
+ * @property {number} [fumble=1]       The value of the d20 result which represents a critical failure.
+ * @property {number} [targetValue]    The value of the d20 result which should represent a successful roll.
+ *
+ * ## Flags
+ * @property {boolean} [elvenAccuracy]   Allow Elven Accuracy to modify this roll?
+ * @property {boolean} [halflingLucky]   Allow Halfling Luck to modify this roll?
+ * @property {boolean} [reliableTalent]  Allow Reliable Talent to modify this roll?
+ *
+ * ## Roll Configuration Dialog
+ * @property {boolean} [fastForward=false]     Should the roll configuration dialog be skipped?
+ * @property {boolean} [chooseModifier=false]  If the configuration dialog is shown, should the ability modifier be
+ *                                             configurable within that interface?
+ * @property {string} [template]               The HTML template used to display the roll configuration dialog.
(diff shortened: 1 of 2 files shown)


module/dice/dice.mjs
```diff
@@ -3,44 +3,56 @@
...
-  parts=[], data={}, // Roll creation
-  advantage, disadvantage, fumble=1, critical=20, targetValue, elvenAccuracy, halflingLucky, reliableTalent, // Roll customization
-  chooseModifier=false, fastForward=false, event, template, title, dialogOptions, // Dialog configuration
-  chatMessage=true, messageData={}, rollMode, speaker, flavor // Chat Message customization
+  parts=[], data={}, event,
+  advantage, disadvantage, critical=20, fumble=1, targetValue,
+  elvenAccuracy, halflingLucky, reliableTalent,
+  fastForward=false, chooseModifier=false, template, title, dialogOptions,
+  chatMessage=true, messageData={}, rollMode, flavor,
+  speaker // Deprecated
```
akrigline, line 55: should we take this opportunity to nuke the deprecated speaker, or is that a job for another PR in 2.0.0?

module/documents/actor/actor.mjs
```diff
@@ -1117,29 +1213,58 @@ export default class Actor5e extends Actor {
...
+     * @param {Actor5e} actor        Actor for which the hit die has been rolled.
+     * @param {D20Roll} roll         The resulting roll.
+     * @param {object} updates
+     * @param {object} updates.actor  Updates that will be applied to the actor.
+     * @param {object} updates.class  Updates that will be applied to the class.
+     */
+    if ( Hooks.call("dnd5e.rollHitDie", this, roll, updates) === false ) return roll;
+
+    // Perform updates
+    if ( !foundry.utils.isEmpty(updates.actor) ) this.update(updates.actor);
```
akrigline, line 1265: Is it worth awaiting these operations, or are we trusting that the server handles concurrent requests on embedded items/parent documents?

akrigline (approved):
This gets a 👍 from me, My only further suggestion might be to try splitting out the Dice class changes from the simple addition of Actor hooks. That would result in some pretty gnarly conflicts without precise ordering of PR merges so I get why they're all in here, but I figured I'd bring it up.

module/dice/d20-roll.mjs
```diff
@@ -60,6 +75,28 @@ export default class D20Roll extends Roll {
...
 
+  /* -------------------------------------------- */
+
+  /**
+   * Is this roll a critical success? Returns undefined if roll isn't evaluated.
+   * @type {boolean|void}
+   */
+  get isCritical() {
+    if ( !this._evaluated ) return undefined;
+    return this.dice[0].total >= (this.options.critical ?? 20);
```
aaclayton, line 86: I think it should be possible to have a d20 roll where a critical success is not possible. If we wanted to be a bit more buttoned-up about the rules, for example, we might not highlight a nat 20 as a "critical success" for things like skill checks or saving throws for which that result has no particular meaning. 

I think for such rolls, it should be allowed that `this.options.critical` would be `null` to indicate "it is not possible to achieve a critical success" - which would require the logic to only consider cases in which `this.options.critical` has `Number.isNumeric()`

aaclayton (changes requested):

aaclayton (approved):

PR #1674, opened 2022-07-29
Switch to using .hbs extension for templates, add short name for Handlebars partials

arbron:
- Renames all templates from `.html` to `.hbs`
- Changes how partials are registered so they now use just the filename (no path or extension) with the `dnd5e.` prefix (`"systems/dnd5e/templates/actors/parts/actor-spellbook.hbs"` becomes `"dnd5e.actor-spellbook"`)
- Register full path partials using old extension for backwards compatibility


aaclayton (approved):
Looks tidy to me, thanks @arbron

Issue #1675, opened 2022-07-30
Add UUID to Advancement and allow them to be fetched using `fromUuid`


arbron:
To make it easier to store references to specific advancement "documents", they should have a UUID similar to normal documents and be accessible using the `fromUuid` & `fromUuidSync`.

PR #1676, opened 2022-07-30
[#1675] Add UUID to `Advancement` & make work with `fromUuid` & `fromUuidSync`

arbron:


aaclayton (commented):
I'm curious how essential this is to improving aspects of the advancement system. I am reluctant and wary of having support for a "uuid" which does not point to a `Document`. I realize this might be convenient to route some things through the same UUID-based links that items use, but is there some other way that doesn't conflate the concepts?

arbron:
> I'm curious how essential this is to improving aspects of the advancement system. I am reluctant and wary of having support for a "uuid" which does not point to a `Document`. I realize this might be convenient to route some things through the same UUID-based links that items use, but is there some other way that doesn't conflate the concepts?

Not essential, but it would be helpful to use in a few places to have a consistent way to reference advancements. Right now we are using a format `<itemId>.<advancementId>` as a placeholder for advancement UUID when advancements add items for example, but I felt a method more consistent with the core system would be helpful.

This might also be helpful for things like dragging & dropping advancements between different items and other cases where we need to reference a specific advancement from outside the item itself.

Fyorl:
The fact that it's relatively few lines of code to enable this functionality is quite compelling. There would be more work if we wanted to integrate it with drag-and-drop though, I think. Most of the core API infrastructure relies on calling `getDocumentClass(dragData.type)` and then `cls.fromDropData(dragData)`. Given that we'd potentially have to special-case a lot of code, maybe it's better if we don't try to extend the UUID format in this way.

`JournalEntryPage`s have a related feature where you can link to a heading within that `JournalEntryPage`, and these are expressed as 'JournalEntry.abc.JournalEntryPage.xyz#heading-slug` where the supplemental data is appended after a `#`. Similarly, ProseMirror instances are given unique identifiers with the same format. We've tried to make this quite general so that downstream consumers could use this format for their own purposes, so perhaps that's an option here?

aaclayton:
Okay, sounds like @Fyorl is comfortable here so I'm good to go also.

PR #1679, opened 2022-08-03
[#1624] incorporated forgotten adventures token pack 33 to 35

### LINKED ISSUE #1624: Incorporate Forgotten Adventures creature tokens packs 33, 34, and 35
- Pack 33: https://www.patreon.com/posts/creature-tokens-68483730
- Pack 34: https://www.patreon.com/posts/creature-tokens-68483902
- Pack 35: https://www.patreon.com/posts/creature-tokens-69830449

Kulkodar:
Added the new tokens from Forgotten Adventures.

One of the packs adds the Frost Giant which version should be used?

Closes #1624


aaclayton:
This merge looks almost perfect, but I spot an issue with the Griffon and with the Mammoth. We need to rotate the tokens to be south-facing without changing their size. What this entails is expanding the canvas size in order to accommodate the true south-facing token and increasing the token scale appropriately. @anathemamask has some experience with the changes that are needed to achieve that and may provide guidance.

aaclayton (changes requested):
See comment regarding our need for a south-facing Mammoth and Griffon tokens.

Kulkodar:
I should be able to calculate the new scale, if i need help i'll poke @anathemamask and ask for guidance.

aaclayton (changes requested):
Sorry, but the token upscaling did not go quite right. The resolution change needs to be completely exact to avoid any loss of quality or loss of true sizing relative to other objects. Your resized Mammoth is 1697px square when it should be 1800px. 

ACTUALLY - I notice an even more significant problem. Something strange is going on here. FA provides true south-facing tokens directly. I see in their library the file `Mammoth_Huge_Scale150_Beast_01.webp` which is not the source image that you have used for this MR. Where did the source tokens come from that they were not initially south-facing?

We need to make sure we are using the true source webp files directly from FA - they have done work to (I think) make sure that every token is south-facing in what they make available directly. I think this needs a more careful look.

Also, we don't need to include the .png images anymore. We are going to be removing those with the next 5e system release.

Kulkodar:
Hmm i used the diagonal of the image as the new side length and calculated the scaling factor. Is there something i did not think of?

Strange i redownloaded the free and premium pack from the website and there are not south facing. Where did you see it, in the live gallery of FA?

I have seen that some have png and others don't so i included both, will remove it.

aaclayton:
I get the FA webp tokens directly through their pCloud share drive. I guess maybe they don't make the webps available in the Patreon ZIP files, although I thought they did.

I can circle back here later today or this week and update the files with the official ones.

aaclayton (approved):
I'm going to merge this in as-is and then make some corrections on my end.

PR #1688, opened 2022-08-07
[#1505, #1684, #1686, #1687] Improve scale value advancement with better formatting & conversion

### LINKED ISSUE #1505: Allow customization of dice sizes in scale value advancement
Originally in GitLab by @sharkbruhaha

Please change the `Dice` Scalar Advancement from a dropdown to a numeric field.

Having this hard coded as a dropdown prevents the use of d100, but also larger non-physical dice numbers. Homebrew Wild Magic comes to mind immediately, but why limit it as well.

### LINKED ISSUE #1684: Allow access to count & dice size for single dice scale values
Rather than:

unknown

You could create this: 

unknown-1

And then access like this:

```
@scale.monster-hunter.superiority-dice -> 4d6
@scale.monster-hunter.superiority-dice.count -> 4
@scale.monster-hunter.superiority-dice.denomination -> d6
```

### LINKED ISSUE #1686: Add Challenge Rating to list of Scale Value types
Adding a dedicated CR type will be useful in presenting CRs properly for display:

Screen Shot 2022-08-06 at 21 59 42

### LINKED ISSUE #1687: Attempt to convert values when changing scale value type
Currently, when the type of a scale value is changed, all of the values previously entered are cleared. This can be annoying if you start entering values only to realize you need to change the type and thus have to reenter all the previous values. In many cases a conversion is possible between the different types, so it should be attempted.

arbron:
- Allows dice size & type to be access separately from dice scale values (resolves #1684)
  - `@scale.battle-master.superiority-dice` => `4d8`
  - `@scale.battle-master.superiority-dice.number` => `4`
  - `@scale.battle-master.superiority-dice.dice` => `d8`
- Adds new `"Challenge Rating"` type for things like Wildshape to properly display fractions (`¼` instead of `0.25`) (resolves #1686)
- Attempts to convert between types if they are changed to improve usability (resolves #1687)
- Allows die types to be customizable by script (e.g. `dnd5e.advancement.types.ScaleValueTypeDice.FACES.push(200)`) (resolves #1505)
- Add `d100` as a default die type


Varriount:
`@scale.battle-master.superiority-dice.number` might be better as `@scale.battle-master.superiority-dice.count`

module/advancement/types/scale-value.mjs
```diff
@@ -2,6 +2,195 @@ import Advancement from "../advancement.mjs";
...
+  /* -------------------------------------------- */
+
+  /**
+   * The die value to be rolled with the leading "d" (e.g. "d4").
+   * @type {string}
+   */
+  get dice() {
+    if ( !this.die ) return "";
+    return `d${this.die}`;
+  }
```
Varriount, line 172: Perhaps it might be better to return just the number?

arbron: You can get just the number using `@scale.battle-master.superiority-dice.die`, which is admittedly a bit close to `dice`.

Varriount: I guess, I don't quite feel that this really needs to be a function. I would at least give it a more specific name, like `prefixedDie`.

lang/en.json
```diff
@@ -144,12 +144,14 @@
 "DND5E.AdvancementScaleValueTitle": "Scale Value",
 "DND5E.AdvancementScaleValueHint": "A single value that changes as a class increases in level that is made available in roll formulas (such as a Monk's martial arts die).",
 "DND5E.AdvancementScaleValueIdentifierHint": "This scale value for the current level will be available using <strong>@scale.{class}.{identifier}</strong> in a formula.",
+"DND5E.AdvancementScaleValueTypeCR": "Challenge Rating",
 "DND5E.AdvancementScaleValueTypeDice": "Dice",
 "DND5E.AdvancementScaleValueTypeDistance": "Distance",
-"DND5E.AdvancementScaleValueTypeHintString": "Use this if you are unsure of the type of the scale values, or if the type does not matter.",
-"DND5E.AdvancementScaleValueTypeHintNumber": "Use this for purely numerical values like rage damage bonus or number of uses per day.",
+"DND5E.AdvancementScaleValueTypeHintCR": "Use this for representing creature challenge ratings for features like Wildshape.",
```
Fyorl, line 150: I don't know how consistent we've been with this in the past but the Druid feature is actually two words: 'Wild Shape'.

module/advancement/types/scale-value.mjs
```diff
@@ -2,6 +2,195 @@ import Advancement from "../advancement.mjs";
...
+    return { value: original.prepared };
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * This scale value prepared to be used in roll formulas.
+   * @type {string|null}
+   */
+  get prepared() { return this.value; }
```
Fyorl, line 56: Maybe we could call this `formula` instead? Assuming it's possible to use this value in such a way: `new Roll(scaleValue.formula)` then I think that name would align with the `Roll` parameter name.

module/advancement/types/scale-value.mjs
```diff
@@ -2,6 +2,195 @@ import Advancement from "../advancement.mjs";
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * This scale value formatted for display.
+   * @type {string|null}
+   */
+  get formatted() { return this.prepared; }
```
Fyorl, line 64: Personal preference for calling this `display` but I don't feel strongly, I'll leave it up to you.

module/advancement/types/scale-value.mjs
```diff
@@ -2,6 +2,195 @@ import Advancement from "../advancement.mjs";
...
+   * Shortcut to the prepared value when used in roll formulas.
+   * @returns {string}
+   */
+  toString() {
+    return this.prepared;
+  }
+}
+
+
+export class ScaleValueTypeNumber extends ScaleValueType {
```
Fyorl, line 78: Missing JSDoc and `@inheritdoc` on the methods.

module/advancement/types/scale-value.mjs
```diff
@@ -2,6 +2,195 @@ import Advancement from "../advancement.mjs";
...
+      hint: "DND5E.AdvancementScaleValueTypeHintNumber",
+      isNumeric: true
+    });
+  }
+
+  /* -------------------------------------------- */
+
+  static converted(original) {
+    const value = Number(original.prepared);
+    if ( value.isNaN ) return null;
```
Fyorl, line 92: Is this really a property on `Number` instances? I think this should be `isNaN(value)` or `Number.isNaN(value)` (I think they're functionally identical given that we've already cast to `Number` here).

module/advancement/types/scale-value.mjs
```diff
@@ -2,6 +2,195 @@ import Advancement from "../advancement.mjs";
...
+      case 0.125: return "&frac18;";
+      case 0.25: return "&frac14;";
+      case 0.5: return "&frac12;";
+      default: return super.formatted;
+    }
+  }
+}
+
+
+export class ScaleValueTypeDice extends ScaleValueType {
```
Fyorl, line 120: Missing JSDoc and `@inheritdoc` on the methods.

module/advancement/types/scale-value.mjs
```diff
@@ -69,7 +259,7 @@ export class ScaleValueAdvancement extends Advancement {
 
   /** @inheritdoc */
   titleForLevel(level, { configMode=false }={}) {
-    const value = this.formatValue(level);
+    const value = this.valueForLevel(level)?.formatted;
```
Fyorl, line 262: Since we have `toString` defined, I think this can just be:

```suggestion
    const value = this.valueForLevel(level);
```

Fyorl (changes requested):
Looks pretty reasonable to me, thanks for doing this work, Jeff. If you require your `MappingField` in order to integrate this with the new Advancement Data Model, then feel free to add it to this PR and I will review.

module/advancement/types/scale-value.mjs
arbron: `toString` returns the formula value, not the display value (this way when it is in a roll formula you can just do `@scale.monk.martial-arts` to get the proper formula version and not the display version, which might not work in formulas).

arbron:
@Fyorl This has been converted over to use data models and could use another review pass.

module/data/advancement/scale-value.mjs
```diff
@@ -1,17 +1,325 @@
...
+   * List of die faces that can be chosen.
+   * @type {number[]}
+   */
+  static FACES = [2, 3, 4, 6, 8, 10, 12, 20, 100];
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  static convertFrom(original) {
+    let [number, face] = (original.formula ?? "").split("d");
```
Fyorl, line 257:
```suggestion
    const [number, face] = (original.formula ?? "").split("d");
```

module/data/advancement/scale-value.mjs
```diff
@@ -1,17 +1,325 @@
...
+ * @property {number} n    Number of dice.
+ * @property {number} die  Die face.
+ */
+export class ScaleValueTypeDice extends ScaleValueType {
+
+  /** @inheritdoc */
+  static defineSchema() {
+    return {
+      number: new foundry.data.fields.NumberField({nullable: true, integer: true, positive: true}),
+      face: new foundry.data.fields.NumberField({required: true, integer: true, positive: true})
```
Fyorl, line 231: `DiceTerm` uses the plural `faces` (e.g., an 8-sided die has 8 faces on the polyhedron). I think the singular `face`, probably does not make sense in this context.

templates/advancement/scale-value-config.hbs
```diff
@@ -34,19 +34,19 @@
     </div>
   </div>
 
-  <div class="right-column scale-values">
+  <div class="right-column scale-values">{{log this}}
```
Fyorl, line 37: Errant debugging statement.

Fyorl (changes requested):

Fyorl (approved):

PR #1690, opened 2022-08-07
[#1620, #1626, #1689] Add spell config & type restriction to ItemGrant advancement

### LINKED ISSUE #1620: Title text for delete item icon on Item Grant Advancement config isn't localized

### LINKED ISSUE #1626: Add spell configuration to ItemGrant advancement

### LINKED ISSUE #1689: Prevent `ItemGrantAdvancement` from adding classes, subclasses, or backgrounds
Should help prevent this type from being used in a way the advancement system is not currently able to support.

arbron:
- Add ability to override spellcasting ability, preparation mode, and max uses for spells added using item grant advancement (resolves #1626)
- Prevent `class`, `subclass`, or `background` items from being added using item grant to avoid issues caused by the use of the advancement system in a way that isn't yet supported (existing advancements that do this won't break, but new ones can't be created through the interface) (resolves #1689)
- Localize delete title that was missed (resolves #1620)
- Move drag & drop code into `AdvancementConfig` to prepare for sharing with `ItemChoiceAdvancement`

### DIFF
--- a/module/advancement/advancement-config.mjs
+++ b/module/advancement/advancement-config.mjs
@@ -192,6 +192,7 @@ export default class AdvancementConfig extends FormApplication {
    * if the item is invalid.
    * @param {Event} event  Triggering drop event.
    * @param {Item5e} item  The materialized Item that was dropped.
+   * @throws An error if the item is invalid.
    * @protected
    */
   _validateDroppedItem(event, item) {}
--- a/module/advancement/advancement-config.mjs
+++ b/module/advancement/advancement-config.mjs
@@ -84,7 +84,7 @@ export default class AdvancementConfig extends FormApplication {
     super.activateListeners(html);
 
     // Remove an item from the list
-    if ( this.options.dropKeyPath ) html.on("click", ".item-delete", this._onItemDelete.bind(this));
+    if ( this.options.dropKeyPath ) html.on("click", "[data-action='delete']", this._onItemDelete.bind(this));
   }
 
   /* -------------------------------------------- */
@@ -132,7 +132,6 @@ export default class AdvancementConfig extends FormApplication {
     const updates = { configuration: await this.prepareConfigurationUpdate({
       [this.options.dropKeyPath]: items.filter(uuid => uuid !== uuidToDelete)
     }) };
-    console.log(updates);
     await this.advancement.update(updates);
     this.render();
   }
@@ -148,7 +147,7 @@ export default class AdvancementConfig extends FormApplication {
 
   /** @inheritdoc */
   async _onDrop(event) {
-    if ( !this.options.dropKeyPath ) return console.error(
+    if ( !this.options.dropKeyPath ) throw new Error(
       "AdvancementConfig#options.dropKeyPath must be configured or #_onDrop must be overridden to support"
       + " drag and drop on advancement config items."
     );
@@ -164,14 +163,17 @@ export default class AdvancementConfig extends FormApplication {
     if ( data.type !== "Item" ) return false;
     const item = await Item.implementation.fromDropData(data);
 
-    const verified = this._verifyDroppedItem(event, item);
-    if ( !verified ) return false;
+    try {
+      this._validateDroppedItem(event, item);
+    } catch(err) {
+      return ui.notifications.error(err.message);
+    }
 
     const existingItems = foundry.utils.getProperty(this.advancement.data.configuration, this.options.dropKeyPath);
 
     // Abort if this uuid is the parent item
     if ( item.uuid === this.item.uuid ) {
-      return ui.notifications.warn(game.i18n.localize("DND5E.AdvancementItemGrantRecursiveWarning"));
+      return ui.notifications.error(game.i18n.localize("DND5E.AdvancementItemGrantRecursiveWarning"));
     }
 
     // Abort if this uuid exists already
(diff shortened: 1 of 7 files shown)


less/items.less
```diff
@@ -251,7 +251,7 @@
         .configuration-mode-control {
           flex: 1 0;
           margin-inline-start: 0.5em;
-          a { text-align: inline-start; }
+          a { text-align: start; }
```
aaclayton, line 254: I think we generally use `text-align: left` elsewhere in the code, if consistency is desired.

module/advancement/advancement-config.mjs
```diff
@@ -76,6 +80,15 @@ export default class AdvancementConfig extends FormApplication {
 
   /* -------------------------------------------- */
 
+  activateListeners(html) {
+    super.activateListeners(html);
+
+    // Remove an item from the list
+    if ( this.options.dropKeyPath ) html.on("click", ".item-delete", this._onItemDelete.bind(this));
```
aaclayton, line 87: Not sure if we envision this app might in the future have other sort of control buttons, but if so I do prefer the paradigm of handling all such buttons with one aggregated handler, usually with the class selector `.action-button` or similar and using `data-action` in the HTML markup to differentiate button intent.

module/advancement/advancement-config.mjs
```diff
@@ -101,4 +114,86 @@ export default class AdvancementConfig extends FormApplication {
...
+   */
+  async _onItemDelete(event) {
+    event.preventDefault();
+    const uuidToDelete = event.currentTarget.closest("[data-item-uuid]")?.dataset.itemUuid;
+    if ( !uuidToDelete ) return;
+    const items = foundry.utils.getProperty(this.advancement.data.configuration, this.options.dropKeyPath);
+    const updates = { configuration: await this.prepareConfigurationUpdate({
+      [this.options.dropKeyPath]: items.filter(uuid => uuid !== uuidToDelete)
+    }) };
+    console.log(updates);
```
aaclayton, line 135: This was probably for debugging during development and should be removed

module/advancement/advancement-config.mjs
```diff
@@ -101,4 +114,86 @@ export default class AdvancementConfig extends FormApplication {
...
+    await this.advancement.update(updates);
+    this.render();
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  _canDragDrop() {
+    return this.isEditable;
```
aaclayton, line 144: Idle thought: this should probably be the default behavior of the base `FormApplication`

module/advancement/advancement-config.mjs
```diff
@@ -101,4 +114,86 @@ export default class AdvancementConfig extends FormApplication {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  async _onDrop(event) {
+    if ( !this.options.dropKeyPath ) return console.error(
```
aaclayton, line 151: This seems like it should be a `throw new Error` rather than a `return console.error`

module/advancement/advancement-config.mjs
```diff
@@ -101,4 +114,86 @@ export default class AdvancementConfig extends FormApplication {
...
+    try {
+      data = JSON.parse(event.dataTransfer.getData("text/plain"));
+    } catch(err) {
+      return false;
+    }
+
+    if ( data.type !== "Item" ) return false;
+    const item = await Item.implementation.fromDropData(data);
+
+    const verified = this._verifyDroppedItem(event, item);
```
aaclayton, line 167: I think this method name might be more clear with two changes:
1. `validate` instead of `verify`
2. The `_validateDroppedItem` method would `throw` and Error if invalid rather than returning `false`

module/advancement/advancement-config.mjs
```diff
@@ -101,4 +114,86 @@ export default class AdvancementConfig extends FormApplication {
...
+    if ( !verified ) return false;
+
+    const existingItems = foundry.utils.getProperty(this.advancement.data.configuration, this.options.dropKeyPath);
+
+    // Abort if this uuid is the parent item
+    if ( item.uuid === this.item.uuid ) {
+      return ui.notifications.warn(game.i18n.localize("DND5E.AdvancementItemGrantRecursiveWarning"));
```
aaclayton, line 174: This one also feels like it should be a `throw new Error`, and/or `ui.notifications.error` rather than just a UI warning.

module/advancement/advancement-config.mjs
```diff
@@ -101,4 +114,86 @@ export default class AdvancementConfig extends FormApplication {
...
+    }
+
+    // Abort if this uuid exists already
+    if ( existingItems.includes(item.uuid) ) {
+      return ui.notifications.warn(game.i18n.localize("DND5E.AdvancementItemGrantDuplicateWarning"));
```
aaclayton, line 179: This is a good use case for a UI warning, as a comparison to the others.

aaclayton (changes requested):

less/items.less
arbron: I've been generally trying to use logical properties and values wherever possible over physical to allow for potential future support of other reading directions.

Fyorl (changes requested):

module/advancement/types/item-grant.mjs
```diff
@@ -129,72 +143,28 @@ export class ItemGrantConfig extends AdvancementConfig {
...
-  }
-
-  /* -------------------------------------------- */
-
   /** @inheritdoc */
-  _canDragDrop() {
-    return this.isEditable;
+  async getData() {
+    const data = super.getData();
+    data.items = data.data.configuration.items.map(fromUuidSync);
```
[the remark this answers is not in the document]
arbron: I don't think there is a need for `getData` to be async in this situation, so I just removed that keyword.

Fyorl (approved):

PR #1710, opened 2022-08-18
[#1532] Displaying toolType in tool sheet header

djinnstar:
[possible solution to #1532]

This PR adds the tool's type [from among Artisan's, Musical Instrument, or Gaming Set] to the tool sheet's header.

### DIFF
--- a/templates/items/tool.hbs
+++ b/templates/items/tool.hbs
@@ -16,7 +16,11 @@
 
             <ul class="summary flexrow">
                 <li>
-                    {{lookup config.toolTypes system.toolType}}
+                    {{#if system.toolType}}
+                      {{lookup config.toolTypes system.toolType}}
+                    {{else}}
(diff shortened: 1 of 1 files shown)


templates/items/tool.hbs
```diff
@@ -15,6 +15,9 @@
             </div>
 
             <ul class="summary flexrow">
+                <li>
+                    {{lookup config.toolTypes system.toolType}}
```
arbron, line 19:
```suggestion
                    {{#if system.toolType}}
                      {{lookup config.toolTypes system.toolType}}
                    {{else}}
                      {{localize "DND5E.ItemTypeTool"}}
                    {{/if}}
```

arbron (changes requested):
I think the handling of non-typed tools got lost in the shuffle :).

djinnstar:
Good point! Added back in.

arbron (approved):

Issue #1692, opened 2022-08-08
Convert "Foundry notes" in descriptions to using `<secret>` blocks


arbron:
A number of items have Foundry implementation notes for users in addition to their normal descriptions. These notes should be moved inside `<secret>` blocks so they only show up for the owner and not in chat messages.

arbron:
Here are some class ones I found:
- [ ] Bardic Inspiration
- [ ] Font of Inspiration
- [ ] Channel Divinity (Cleric)
- [ ] Ki
- [ ] Deflect Missiles
- [ ] Favored Enemy
- [ ] Natural Explorer
- [ ] Font of Magic
- [ ] Spell Mastery

arbron:
- [ ] Sling Bullet
- [ ] Necklace of Prayer Beads
- [ ] Ring of Regeneration
- [ ] Ring of the Ram
- [ ] Robe of Useful Items
- [ ] Potion of Poison
- [ ] Potion of Heroism
- [ ] Restorative Ointment
- [ ] Rod of Absorption
- [ ] Feather Token Bird
- [ ] Necklace of Fireballs
- [ ] Sovereign Glue
- [ ] Defender Swords
- [ ] Dwarven Thrower
- [ ] Mace of Smiting
- [ ] Nine Lives Stealer Swords
- [ ] Staff of Fire
- [ ] Staff of Frost
- [ ] Staff of Charming
- [ ] Staff of Healing
- [ ] Staff of Striking
- [ ] Staff of Power
- [ ] Staff of Swarming Insects
- [ ] Staff of the Magi
- [ ] Staff of the Woodlands
- [ ] Staff of Withering
- [ ] Unarmed Strike Monk
- [ ] Breath Weapon
- [ ] Savage Attacks

PR #1712, opened 2022-08-20
[#1692] Convert Foundry Notes in journal entries to use secret blocks

arbron:
Converts the various Foundry notes in various items into secret blocks so they don't appear in chat. They are also all given the `foundry-note` class in case we want to do additional styling or handling later.


Fyorl:
This looks good, thanks again for the hard work, Jeff.

When testing this I uncovered a bug with our core ProseMirror editor, noted here: https://github.com/foundryvtt/foundryvtt/issues/8014

Also, the 'reveal' buttons that we now generate on secret blocks is inheriting some poor styling from the dnd5e sheet, which we should address before we merge:

image

Related to that, the 'reveal' button does not work on these secret blocks in item sheets, and I haven't had time to investigate it yet. We can potentially just hide them here but I don't think there's any technical limitations on why the reveal button shouldn't function here.

Fyorl (approved):
I found more bugs with core implementation of secret reveal but it only applies to those in locked compendia. I've hidden the reveal buttons if they appear in a locked compendium for now (they won't work anyway due to the compendium being locked), and also applied some style tweaks to make them match the core reveal buttons.

Thanks for doing this work, Jeff.

PR #1715, opened 2022-08-22
Update rules compendium with new journals v2 features

Fyorl:
* Condenses the separate journal entries into single-chapter entries with multiple pages.
* Use relative UUID links where appropriate.
* Adjust heading levels where appropriate.
* Implement a custom compendium sheet for displaying the whole of the content in an easily-accessible way.

The demo video is too big to attach here but I uploaded it to discord here: https://discord.com/channels/170995199584108546/795664425255174195/1010263684091740160


module/applications/srd-compendium.mjs
```diff
@@ -0,0 +1,91 @@
+Hooks.once("ready", () => {
+  const pack = game.packs.get("dnd5e.rules");
+  pack.apps = [new SRDCompendium(pack)];
+});
```
aaclayton, line 4: I would rather we put this code alongside our other init and ready behaviors in the root dnd5e module file.

templates/apps/srd-compendium.hbs
```diff
@@ -0,0 +1,33 @@
+<article>
+  <aside>
+    <header>
+      <h2>{{localize "DND5E.Disclaimer"}}</h2>
+    </header>
+    <section>{{{disclaimer}}}</section>
+  </aside>
+  <main>
```
aaclayton, line 8: I believe you are not supposed to have more than one `<main>` element in an HTML document: https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element

I think we should perhaps just use `<div>` instead since the reason for this element container is just for layout.

aaclayton (approved):
Approved with comments.

PR #1723, opened 2022-08-23
Switch actor and item sheets over to using the ProseMirror editor.

Fyorl:


aaclayton (approved):
I'm happy to see that the core changes you made allow this to be so simple. Please feel free to merge this in once the upstream core changes are also merged.

PR #1728, opened 2022-08-25
[#1722] Create custom journal page for displaying an automatically generated class summary

arbron:
Uses advancement data stored in a class to auto-generate a journal entry for that class styled in a manner similar to the SRD and other official books:

Class - Wizard w: Spell Table

This includes a full table with proficiency bonus, feature summary, scale values, and spell slots per level if available. There is a large descriptive area at the top, and several smaller descriptive areas throughout to allow for populating details not available on the class item:

Class - Editing Screen

Classes with optional features will have them displayed in a separate table & section:

Class - Optional Features

It also supports adding subclasses beneath the class details with their own tables and feature lists:

Class - Rogue Subclass


Fyorl:
I have some high-level questions before I get started with a full review:

1. Where does this leave our class item descriptions? These journal pages display roughly the same information. Do we anticipate keeping those two in-sync indefinitely? Do we retire the class item descriptions entirely? If we have the information available to automatically generate the class descriptions, would it be better to do that as part of the class item description rather than as a separate journal page?
2. I notice we're using scale values to encode class spell progression, but we do have plans to have a dedicated advancement type for these later. Should we wait for that first to avoid having to migrate these later?

arbron:
> 1. Where does this leave our class item descriptions? These journal pages display roughly the same information. Do we anticipate keeping those two in-sync indefinitely? Do we retire the class item descriptions entirely? If we have the information available to automatically generate the class descriptions, would it be better to do that as part of the class item description rather than as a separate journal page?

I was thinking we would reduce the description of class items to a short description and a link to the journal entry. This is a bit hard for the SRD classes because they don't really provide a description, they just go straight into the rules. We could also pull out the code for generating the table and add like a table enricher for class descriptions so we can use the auto-generated table.

> 2. I notice we're using scale values to encode class spell progression, but we do have plans to have a dedicated advancement type for these later. Should we wait for that first to avoid having to migrate these later?

I was thinking that instead of duplicating a bunch of the scale value code for a future spellcasting advancement, we could just have that spellcasting advancement point to the existing scale values to get their data on spells & cantrips available.

module/applications/journal/class-summary-sheet.mjs
```diff
@@ -0,0 +1,439 @@
...
+      }
+    }
+
+    const cols = [{ class: "level", span: 1 }];
+    if ( item.type === "class" ) cols.push({class: "prof", span: 1});
+    if ( hasFeatures ) cols.push({class: "features", span: 1});
+    if ( scaleValues.length ) cols.push({class: "scale", span: scaleValues.length});
+    if ( spellProgression ) cols.push(...spellProgression.cols);
+
+    const makeLink = uuid => TextEditor._createContentLink(["", "UUID", uuid]).outerHTML;
```
Fyorl, line 139: I think the synchronous version of this will be going away eventually too. But we might prefer to do this instead:

```suggestion
    const makeLink = uuid => (await fromUuid(uuid))?.toAnchor({classes: ["content-link"]}).outerHTML;
```

The `classes: ["content-link"]` should eventually be unnecessary when we've fixed this small oversight in core.

module/applications/journal/class-summary-sheet.mjs
```diff
@@ -0,0 +1,439 @@
...
+      }
+    }
+
+    // Leveled Progression
+    else {
+      // Get the max spell slot size based on progression
+      actor.items.get(classId).updateSource({"system.levels": CONFIG.DND5E.maxLevel});
+      actor.reset();
+      const largestSlot = Object.entries(actor.system.spells).reduce((slot, [key, data]) => {
+        if ( data.max === 0 ) return slot;
```
Fyorl, line 215:
```suggestion
        if ( !data.max ) return slot;
```

module/applications/journal/class-summary-sheet.mjs
```diff
@@ -0,0 +1,439 @@
...
+        const level = parseInt(key.replace("spell", ""));
```
Fyorl, line 216: Slight preference for `slice` rather than `replace`.

```suggestion
        const level = parseInt(key.slice(5)); // Remove 'spell' prefix.
```

module/applications/journal/class-summary-sheet.mjs
```diff
@@ -0,0 +1,439 @@
...
+        if ( !Number.isNaN(level) && level > slot ) return level;
```
Fyorl, line 217:
```suggestion
        if ( !Number.isNaN(level) && (level > slot) ) return level;
```

module/applications/journal/class-summary-sheet.mjs
```diff
@@ -0,0 +1,439 @@
...
+    const spellcasting = item.spellcasting;
+    if ( !spellcasting || (spellcasting.progression === "none") ) return null;
+
+    const items = [item.toObject()];
+    let classId = item.id;
+    if ( item.type === "subclass" ) {
+      classId = foundry.utils.randomID();
+      items.push({_id: classId, name: "tmp-class", type: "class", system: {identifier: item.system.classIdentifier}});
+    }
+    const actor = await Actor.implementation.create({name: "tmp", type: "character", items}, {temporary: true});
```
Fyorl, line 186: It's a clever way to reuse our spell preparation code, but it does a feel a little hacky to me. Can we instead factor out the spell preparation code in such a way that it will accept, say, a list of class items and return a structure that's suitable for use here as well as part of actor preparation?

module/applications/journal/class-summary-sheet.mjs
```diff
@@ -0,0 +1,439 @@
...
+      { class: "level", span: 1 },
+      { class: "features", span: 1 }
+    ];
+
+    const makeLink = uuid => TextEditor._createContentLink(["", "UUID", uuid]).outerHTML;
+
+    const rows = [];
+    for ( const level of Array.fromRange(CONFIG.DND5E.maxLevel, 1) ) {
+      const features = [];
+      for ( const advancement of item.advancement.byLevel[level] ) {
```
Fyorl, line 267: We could consider factoring this out to share code with `_getTable`, similar to what we've done with `_getFeatures`, or is that maybe premature if we plan to add additional feature types here?

module/applications/journal/class-summary-sheet.mjs
```diff
@@ -0,0 +1,439 @@
...
+    editor.render(true);
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  async _onDrop(event) {
+    let data;
+    try { data = JSON.parse(event.dataTransfer.getData("text/plain")); }
+    catch(err) { return false; }
```
Fyorl, line 423:
```suggestion
    const data = TextEditor.getDragEventData(event);
```

module/applications/journal/class-summary-sheet.mjs
```diff
@@ -0,0 +1,439 @@
...
+
+    if ( data.type !== "Item" ) return false;
+    const item = await Item.implementation.fromDropData(data);
+    switch (item.type) {
```
Fyorl, line 427:
```suggestion
    switch ( item.type ) {
```

Fyorl (changes requested):
Good stuff, Jeff, this will be really nice for content creators.

module/applications/journal/class-summary-sheet.mjs
arbron: I have another branch where I am reworking spell preparation, so that can be handled there rather than adding more to this PR.

module/applications/journal/class-summary-sheet.mjs
arbron: This will eventually also handle ability score improvement & subclass advancements in the main table but not the optional table, but there might still be a way to factor out some of this code.

module/applications/journal/class-summary-sheet.mjs
Fyorl: Is that a PR already? Or just a branch? What's the scope of the refactor there?

arbron: Just a branch for now: https://github.com/arbron/dnd5e/blob/spell/api/module/documents/actor/actor.mjs#L608

This splits `_prepareSpellcasting` into several separate methods and cleans up the logic a bunch, as well as adding hooks to allow for overriding default spellcasting preparation or adding custom types.

Fyorl: It doesn't look too big or controversial. Can you submit it as a PR maybe and I can prioritise it. I'd rather we do it that way than have this code merged as-is, in case it does get merged and then we never get around to the spellcasting PR in this release.

arbron: Okay, PR created: #1888

module/applications/journal/class-sheet.mjs
```diff
@@ -0,0 +1,459 @@
...
+      /**
+       * A hook event that fires to generate the table for custom spellcasting types.
+       * The actual hook names include the spellcasting type (e.g. `dnd5e.buildPsionicSpellcastingTable`).
+       * @param {object} table  Table definition being built. *Will be mutated.*
+       * @param {Item5e} cls    Synthetic class with the proper spellcasting set.
+       * @function dnd5e.buildSpellcastingTable
+       * @memberof hookEvents
+       */
+      Hooks.callAll(
+        `dnd5e.build${type.capitalize()}SpellcastingTable`, table, cls
```
Fyorl, line 255:
```suggestion
        `dnd5e.build${spellcasting.type.capitalize()}SpellcastingTable`, table, cls
```

Fyorl (changes requested):

module/applications/journal/class-sheet.mjs
```diff
@@ -0,0 +1,459 @@
...
+    const table = { rows: [] };
+
+    let cls;
+    if ( item.type === "class" ) {
+      cls = item.clone();
+    } else {
+      cls = await Item.implementation.create({
+        name: "tmp", type: "class", system: {spellcasting: { progression: spellcasting.progression }}
+      }, {temporary: true});
+    }
```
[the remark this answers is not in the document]
arbron: Take a look at what I've got now. No need to create synthetic items but I believe it still allows for custom progressions.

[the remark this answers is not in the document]
Fyorl: Yeah I like that solution. One suggestion for further refining the API: We could shift the `spellcasting` and `count` variables into a final `options` parameter for those methods and use `spellcasting = options.spellcasting ?? item.spellcasting` (you can use some destructuring of course, this is just for illustrative purposes). That might make the API a little more ergonomic and make it clear which arguments are optional. Let me know what you think.

module/documents/actor/actor.mjs
```diff
@@ -632,83 +632,111 @@ export default class Actor5e extends Actor {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Contribute to the actor's spellcasting progression.
+   * @param {object} progression                    Spellcasting progression data. *Will be mutated.*
+   * @param {Actor5e|null} actor                    Actor for whom the data is being prepared.
+   * @param {Item5e} cls                            Class for whom this progression is being computed.
+   * @param {SpellcastingDescription} spellcasting  Spellcasting descriptive object.
```
Fyorl, line 652: Does this `SpellcastingDescription` need a `typedef`? I can't see one but I might have missed it.

arbron: Good catch, meant to add that.

Fyorl (approved):
LGTM. Consider whether you want to apply the same `options` treatment to `computeLeveledProgression` and `computePactProgression`. I'll leave it up to you, and you can merge when ready.

arbron:
> LGTM. Consider whether you want to apply the same `options` treatment to `computeLeveledProgression` and `computePactProgression`. I'll leave it up to you, and you can merge when ready.

I fell like those two methods should be written with the same signature as the hook methods, and I want to pre-split out spellcasting and stuff before calling the hook so I think this is the best method.

Issue #1707, opened 2022-08-18
Handle warnings from `prepareMaxUses` as actor sheet warnings rather than global warnings


arbron:
Currently when preparing max uses on an item it can potentially throw multiple UI errors which have a number of issues:

1. They clutter up the screen, the old formula for `"Rage"` would throw like 6 warnings if it could find the Barbarian class
2. They disappear automatically, potentially preventing the user from being able to figure out the issue before they disappear
3. They didn't include any information on what item is having the actual issue, making it hard to fix the problem

If these errors were instead moved to the standard `_preparationWarnings` on an actor, then they would only clutter the actor sheet, they will stay until the issue is fixed, and they can include more system-specific information on how to resolve the problem.

PR #1747, opened 2022-08-28
[#1707] Display max uses formula warnings on sheet

arbron:
Uses custom `replaceFormulaData` when preparing max uses formula so that missing data can be displayed as an actor preparation warning rather than a global UI warning:

Formula Warning 2

This allows the warning to include the item name and what formula had the issue. It also consolidates multiple warnings into one so if a single formula references multiple properties, only one warning will be shown:

Multiple Properties

The warnings have the added ability to be made into links, so if you click on a warning about a formula issue it will open the sheet for the item having the problem. This functionality has also been extended to the error about bad AC formula which is now displayed in the error style and will open the AC config app when clicked:

AC Error


module/documents/item.mjs
```diff
@@ -641,6 +644,38 @@ export default class Item5e extends Item {
...
+    });
+    if ( (missingReferences.size > 0) && this.actor ) {
+      const listFormatter = new Intl.ListFormat(game.i18n.lang, { style: "long", type: "conjunction" });
+      const message = game.i18n.format("DND5E.FormulaMissingReferenceWarn", {
+        property, name: this.name, references: listFormatter.format(missingReferences)
+      });
+      this.actor._preparationWarnings.push({ message, link: this.uuid, type: "warning" });
+    }
+    return formula;
+  }
```
Varriount, line 675:
```suggestion
    let dataRegex = new RegExp(/@([a-z.0-9_-]+)/gi);
    const missingReferences = new Set();

    formula = formula.replace(dataRegex, (match, term) => {
      let value = foundry.utils.getProperty(data, term);
      if ( value == null ) {
        missingReferences.add(match);
        return "0";
      }
      return String(value).trim();
    });

    if ( missingReferences.size > 0 && this.actor ) {
      const listFormatter = new Intl.ListFormat(game.i18n.lang, { style: "long", type: "conjunction" });
      const message = game.i18n.format("DND5E.FormulaMissingReferenceWarn", {
        property, name: this.name, references: listFormatter.format(missingReferences)
      });
      this.actor._preparationWarnings.push({ message, link: this.uuid, type: "warning" });
    }
    return formula;
  }
```

module/applications/actor/base-sheet.mjs
```diff
@@ -251,6 +251,26 @@ export default class ActorSheet5e extends ActorSheet {
...
+  /**
+   * Prepare preparation warnings for display.
+   * @param {Array<string|object>} warnings  Array of warnings.
+   * @returns {object[]}                     Prepared warnings.
+   * @protected
+   */
+  _getWarnings(warnings) {
+    const prepared = [];
+    for ( const warning of warnings ) {
+      if ( typeof warning === "string" ) {
```
Fyorl, line 265: Do you foresee any issues with converting all of our preparation warnings into these richer objects so we don't need two separate branches when handling them?

module/documents/item.mjs
```diff
@@ -641,6 +644,38 @@ export default class Item5e extends Item {
...
+  /**
+   * Replace referenced data attributes in the roll formula with values from the provided data.
+   * If the attribute is not found in the provided data, display a warning on the actor.
+   * @param {string} formula           The original formula within which to replace.
+   * @param {object} data              The data object which provides replacements.
+   * @param {object} options
+   * @param {string} options.property  Name of the property to which this formula belongs.
+   * @returns {string}                 Formula with replaced data.
+   */
+  replaceFormulaData(formula, data, { property }) {
```
Fyorl, line 656: I wonder if, instead of duplicating some of the `Roll.replaceFormulaData` code here, we can instead use the `missing` value to accomplish the same thing. Since it's a string, we could supply a special value to `Roll.replaceFormulaData` that we then look for afterwards:

```js
const replaced = Roll.replaceFormulaData(formula, data, {missing: "$missing$"});
if ( replaced.includes("$missing$") && this.actor ) {
  // Format warning here.
}
return replaced.replaceAll("$missing$", "0");
```
It's a little bit less efficient because we're doing more passes/searches/replaces over the string, but formula strings are pretty short and I think the code ends up a little bit neater as a result.

Fyorl (changes requested):

module/documents/item.mjs
arbron: The problem with that method is it would leave no way to access the replaced `@` formula, so I wouldn't be able to include that in the warning.

Fyorl: Ah you're right, that would defeat a lot of the usefulness here. This should be on the core system to improve, at least adding a static property for the data formula regex.

module/documents/item.mjs
```diff
@@ -641,6 +644,38 @@ export default class Item5e extends Item {
...
+    let dataRgx = new RegExp(/@([a-z.0-9_-]+)/gi);
```
Fyorl, line 657:
```suggestion
    const dataRgx = new RegExp(/@([a-z.0-9_-]+)/gi);
```

Fyorl (approved):

PR #1750, opened 2022-08-29
[#1748, #1749] Adjust `dnd5e.useItem` hook & clean up hook documentation

### LINKED ISSUE #1748: v2.0.0 alpha3: If a measured template is cancelled, `dnd5e.useItem` hook is never called.
To reproduce:
- use item, tick "Place Measured Template"
- cancel the placement

### LINKED ISSUE #1749: Pass `AbilityTemplate` through to `dnd5e.useItem` hook if created

arbron:
- Ensure `dnd5e.useItem` hook is called even if the template creation is canceled (resolves #1748)
- Pass created measured template into hook (resolves #1749)
- Clean up various issues with hook documentation

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -787,23 +787,26 @@ export default class Item5e extends Item {
     if ( resourceUpdates.length ) await this.actor.updateEmbeddedDocuments("Item", resourceUpdates);
 
     // Prepare card data & display it if options.createMessage is true
-    const cardData = item.displayCard(options);
+    const cardData = await item.displayCard(options);
 
     // Initiate measured template creation
+    let templates;
     if ( config.createMeasuredTemplate ) {
-      const template = dnd5e.canvas.AbilityTemplate.fromItem(item);
-      if ( template ) await template.drawPreview();
+      try {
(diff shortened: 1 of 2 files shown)


module/documents/item.mjs
```diff
@@ -787,23 +787,26 @@ export default class Item5e extends Item {
...
-     * @param {Item5e} item                  Item being used.
-     * @param {ItemUseConfiguration} config  Configuration data for the roll.
-     * @param {ItemUseOptions} options       Additional options for configuring item usage.
+     * @param {Item5e} item                                Item being used.
+     * @param {ItemUseConfiguration} config                Configuration data for the roll.
+     * @param {ItemUseOptions} options                     Additional options for configuring item usage.
+     * @param {MeasuredTemplateDocument[]|null} templates  The measured templates if they were created.
      */
-    Hooks.callAll("dnd5e.useItem", item, config, options);
+    Hooks.callAll("dnd5e.useItem", item, config, options, templates ?? null);
```
akrigline, line 809: Do we/core explicitly return `null` in any other hooks or should we let this slip through as `undefined`?

aaclayton: I think null is fine here.

aaclayton (approved):
Looks good to me.

Issue #1711, opened 2022-08-19
Various SRD items were not properly converted to use core icons


arbron:
I suspect there is not direct match between the icons they use and the ones in core. We will need to find a new icon, and then also update `icon-migration.json` & `spell-icon-migration.json` with these new matches to ensure that players' icons are able to be properly migrated.

Class Features:
- Bonus Cantrip (Druid)
- Invocation: Eldritch Spear

Items:
- Elven Chain
- Robe of Useful Items
- Robes
- Vestments

Monster Features:
- Air Form
- Angelic Weapons
- Charm
- Create Whirlwind
- Darkness Aura
- Death Throes
- Enlarge
- Fire Aura
- Keen Senses
- Lightning Storm
- Misty Escape
- Petrifying Breath
- Possession
- Reflective Carapace
- Rejuvenation
- Shimmering Shield
- Teleport
- Web Sense
- Whirlwind

Racial Features:
- Dragonborn
- Dwarf
- Elf
- Gnome
- Half-Elf
- Halfling
- Hellish Resistance
- High Elf
- Hill Dwarf
- Human
- Lightfoot Halfling
- Rock Gnome
- Tiefling

Spells:
- Instant Summons
- Tiny Hut

arbron:
Also need to make sure we update the items in monsters & starter heroes.

alaustin0:
Icon locations:

**Class Features**
Bonus Cantrip (Druid) - icons/magic/nature/vines-thorned-glow-green.webp
Invocation: Eldritch Spear - messed up for me, not sure which icon it's supposed to be.

**Items**
Elven Chain - icons/equipment/back/mantle-collared-green.webp
Robe of Useful Items - icons/equipment/back/mantle-collared-black.webp
Robes - icons/equipment/back/mantle-collared-black.webp
Vestments - icons/equipment/back/mantle-collared-black.webp

**Monster Features**
Air Form - icons/magic/control/debuff-energy-hold-levitate-yellow.webp
Angelic Weapons - icons/magic/fire/dagger-rune-enchant-flame-strong-blue-yellow.webp
Charm - icons/magic/air/wind-tornado-spiral-pink-purple.webp
Create Whirlwind - icons/magic/air/wind-tornado-cyclone-white.webp
Darkness Aura - icons/magic/perception/eye-ringed-glow-angry-small-red.webp
Death Throes - icons/magic/fire/explosion-fireball-large-red-orange.webp
[...]

PR #1754, opened 2022-08-30
[#1711] Remove debugging code from icon migration, fix icons that didn't migrate

arbron:
- Removes leftover `debugger` from icon migration code
- Fix two icon migrations that pointed to `png` rather than `jpg` files
- Add migration for core icon that moved
- Ensure migration applied to all missing icons


Fyorl (approved):
LGTM thanks very much, Jeff.

PR #1760, opened 2022-09-01
[#1350] Minimum Roll for skills, abilities and attacks

### LINKED ISSUE #1350: Support minimum roll values on skill checks

szefo09:
Allows the global attack, ability or skills to have minimum roll modifier defined and applied.

https://user-images.githubusercontent.com/19979839/174321154-5a558619-a0b6-41bd-ac28-d1521ae580aa.mp4

Resolves #1350


module/dice/d20-roll.mjs
```diff
@@ -180,13 +185,14 @@ export default class D20Roll extends Roll {
...
    * @param {string} [data.defaultAbility]    For tool rolls, the default ability modifier applied to the roll
+   * @param {string} [data.skillId]           For skill rolls, the skillId of the skill that is being rolled
    * @param {string} [data.template]          A custom path to an HTML template to use instead of the default
    * @param {object} options                  Additional Dialog customization options
    * @returns {Promise<D20Roll|null>}         A resulting D20Roll object constructed with the dialog, or null if the
    *                                          dialog was closed
    */
   async configureDialog({title, defaultRollMode, defaultAction=D20Roll.ADV_MODE.NORMAL, chooseModifier=false,
-    defaultAbility, template}={}, options={}) {
+    defaultAbility, skillId, template}={}, options={}) {
```
Fyorl, line 195: I'm perhaps not following this correctly. We've added an `options.minimum` to the `D20Roll` class so that it can be applied as part of `D20Roll#configureModifiers`. Why do we also need to pass the `skillId` through `configureDialog` and `_onDialogSubmit`? Is it not enough to configure a `minimum` that we pass to `d20Roll` inside `Actor5e#rollSkill`, like we do with `Actor5e#rollToolCheck`?

Fyorl (changes requested):
Thank you for the hard work, this will be a nice feature to have.

In addition to the review feedback I've posted, I had a higher-level suggestion too. Currently, we have special-case handling for the `reliableTalent` flag. However, in this PR we have introduced the concept of a skill that can have a `minimum` value. Using this, we should be able to instead handle Reliable Talent as part of actor data preparation.

In `Actor5e#_prepareSkills` we can derive the `skl.minimum` based on the configured skill minimum plus the `reliableTalent` flag. That would allow us to remove any special-case handling for `reliableTalent` in our `d20Roll` workflow, and instead rely on the standard `minimum` handling.

module/dice/d20-roll.mjs
szefo09: This is specifially to handle the special case when we use a different ability for a skill check, that was recently added.

template.json
```diff
@@ -9,47 +9,59 @@
             "proficient": 0,
             "bonuses": {
               "check": "",
-              "save": ""
+              "save": "",
+              "checkMinimum": null,
+              "saveMinimum": null
```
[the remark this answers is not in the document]
szefo09: I agree!

template.json
```diff
@@ -138,143 +150,161 @@
             "ability": "dex",
             "bonuses": {
               "check": "",
-              "passive": ""
+              "passive": "",
+              "minimum": null
```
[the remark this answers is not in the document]
szefo09: done.

module/documents/item.mjs
```diff
@@ -619,6 +619,17 @@ export default class Item5e extends Item {
 
   /* -------------------------------------------- */
 
+  /**
+   * Retrieve an item's minimum attack roll.
+   * @returns {number|null}  The minimum roll value that should be added to the attack roll formula.
+   */
+  getMinimumRoll() {
```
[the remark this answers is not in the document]
szefo09: done.

szefo09:
> Thank you for the hard work, this will be a nice feature to have.
> 
> In addition to the review feedback I've posted, I had a higher-level suggestion too. Currently, we have special-case handling for the `reliableTalent` flag. However, in this PR we have introduced the concept of a skill that can have a `minimum` value. Using this, we should be able to instead handle Reliable Talent as part of actor data preparation.
> 
> In `Actor5e#_prepareSkills` we can derive the `skl.minimum` based on the configured skill minimum plus the `reliableTalent` flag. That would allow us to remove any special-case handling for `reliableTalent` in our `d20Roll` workflow, and instead rely on the standard `minimum` handling.

It could be done, but when I was working on this, I didn't feel comfortable removing parts of the already written code, so I built on top of it. The logic could be moved from `d20Roll#configureModifiers` to `Actor5e#_prepareSkills`. The only downside I see is a small regression - potentially losing the logic for reliableTalent in `d20Roll#toMessage`

module/dice/d20-roll.mjs
Fyorl: OK, I think I understand now: The `skillMinimum` that we derive in `Actor5e#rollSkill` is based on the skill minimum and the ability check minimum. Because the ability used for the skill might change after the user has confirmed the roll dialog, we need to re-derive the minimum in `_onDialogSubmit` based on the new ability, and we need to re-call the skill again to perform that calculation.

I was confused because the `skillId` never changes from `rollSkill` to `_onDialogSubmit` but we still need it to perform the calculation again.

This is actually a similar problem to #1650 where we have a complicated rolling pipeline due to the fact that the user may change the ability score during the rolling process.

Fyorl:
> It could be done, but when I was working on this, I didn't feel comfortable removing parts of the already written code, so I built on top of it. The logic could be moved from `d20Roll#configureModifiers` to `Actor5e#_prepareSkills`. The only downside I see is a small regression - potentially losing the logic for reliableTalent in `d20Roll#toMessage`

I think it's fine to deprecate the `reliableTalent` option and replace it with the `minimum` option that we've added here. It's probably best done as part of this MR, but I can add a commit myself if you prefer.

module/documents/actor/actor.mjs
```diff
@@ -815,6 +815,9 @@ export default class Actor5e extends Actor {
       data.skillBonus = Roll.replaceFormulaData(globalBonuses.skill, data);
     }
 
+    // Calculate skill minimum roll
+    const skillMinimum = Math.max(skl?.minimum ?? 0, abl?.minimum?.check ?? 0) || false;
```
Fyorl, line 819: Probably ok to leave these as `undefined` rather than setting them to `false`. `getMinimumAttackRoll` returns `null`, for example, and that value is passed to `d20Roll` without issue, so using any falsy value is probably fine.

template.json
```diff
@@ -10,6 +10,10 @@
             "bonuses": {
               "check": "",
               "save": ""
+            },
+            "minimum": {
+              "check":null,
```
Fyorl, line 15: Sorry to nitpick but these are missing a space after the colon:

```suggestion
              "check": null,
```

Fyorl (changes requested):

module/documents/actor/actor.mjs
```diff
@@ -931,6 +935,9 @@ export default class Actor5e extends Actor {
     // Add provided extra roll parts now because they will get clobbered by mergeObject below
     if ( options.parts?.length > 0 ) parts.push(...options.parts);
 
+    // Add Check Minimum
+    const abilityCheckMinimum = abl?.minimum?.check || false;
```
Fyorl, line 939: Probably ok to leave these as `undefined` rather than setting them to `false`. `getMinimumAttackRoll` returns `null`, for example, and that value is passed to `d20Roll` without issue, so using any falsy value is probably fine.

szefo09:
> I think it's fine to deprecate the `reliableTalent` option and replace it with the `minimum` option that we've added here. It's probably best done as part of this MR, but I can add a commit myself if you prefer.

I would really appreciate if You could add that commit!

thatlonelybugbear:
Is this something which could be added in the 2.5.0 milestone maybe?

Fyorl:
> Is this something which could be added in the 2.5.0 milestone maybe?

It's currently blocked on the roll refactor.

arbron:
So the roll refactor is complete in `4.1` and the system will support minimums on skill checks, concentration, and initiative. Currently there isn't support for minimums on generic ability checks or saves. If you wish to revamp this in light of the new changes then go ahead, but it might be a big job given the extent of recent changes.

PR #1766, opened 2022-09-03
Allow drag and drop of effects on items

akrigline:
This change introduces drag and drop handlers for the Item Sheet to allow Effects to be dragged and dropped between items.

This also allows effects to be drag and dropped from an item onto an actor to emulate applying the effect.

### Use Cases:
- Having configured a "Poisoned Dagger" with a "Poisioned" Effect, I would like to be able to quickly copy that effect to a "Poisioned Longsword".
  - Said items would not need to transfer their effect to their parent actor in order for the GM to use them on enemies.
- An item which replicates the effects of a spell that has active effects configured can quickly clone the effect of that spell.

### Methodology
Copies what core does for dropping active effects on Actor Sheets, with one change:
- Sets the `origin` of the created effect to be the `uuid` of the new parent item.

### Research
I've been doing this with with the module: Drop Effects on Items for several months and it's been very useful.

### DIFF
--- a/module/applications/item/item-sheet.mjs
+++ b/module/applications/item/item-sheet.mjs
@@ -30,7 +30,8 @@ export default class ItemSheet5e extends ItemSheet {
       classes: ["dnd5e", "sheet", "item"],
       resizable: true,
       scrollY: [".tab.details"],
-      tabs: [{navSelector: ".tabs", contentSelector: ".sheet-body", initial: "description"}]
+      tabs: [{navSelector: ".tabs", contentSelector: ".sheet-body", initial: "description"}],
+      dragDrop: [{dragSelector: "[data-effect-id]", dropSelector: '.effects-list'}],
     });
   }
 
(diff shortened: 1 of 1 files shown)


module/applications/item/item-sheet.mjs
```diff
@@ -500,6 +501,70 @@ export default class ItemSheet5e extends ItemSheet {
...
+    // Create drag data
+    let dragData;
+
+    // Active Effect
+    if ( li.dataset.effectId ) {
+      const effect = this.item.effects.get(li.dataset.effectId);
+      dragData = effect.toDragData();
+    }
+
+    if ( !dragData ) return;
```
Fyorl, line 520:
```suggestion
    const effect = this.item.effects.get(li.dataset.effectId);
    const dragData = effect?.toDragData();
    if ( !dragData ) return;
```

module/applications/item/item-sheet.mjs
```diff
@@ -30,7 +30,8 @@ export default class ItemSheet5e extends ItemSheet {
       classes: ["dnd5e", "sheet", "item"],
       resizable: true,
       scrollY: [".tab.details"],
-      tabs: [{navSelector: ".tabs", contentSelector: ".sheet-body", initial: "description"}]
+      tabs: [{navSelector: ".tabs", contentSelector: ".sheet-body", initial: "description"}],
+      dragDrop: [{dragSelector: "[data-effect-id]", dropSelector: '.effects-list'}],
```
Fyorl, line 34:
```suggestion
      dragDrop: [{dragSelector: "[data-effect-id]", dropSelector: ".effects-list"}],
```

module/applications/item/item-sheet.mjs
```diff
@@ -500,6 +501,70 @@ export default class ItemSheet5e extends ItemSheet {
...
+    /**
+     * A hook event that fires when some useful data is dropped onto an ItemSheet5e.
+     * @function dnd5e.dropItemSheetData
+     * @memberof hookEvents
+     * @param {Item5e} item                  The Item5e
+     * @param {ItemSheet5e} sheet            The ItemSheet5e application
+     * @param {object} data                  The data that has been dropped onto the sheet
+     * @returns {boolean}                    Explicitly return `false` to prevent normal drop handling.
+     */
+    const allowed = Hooks.call('dnd5e.dropItemSheetData', item, this, data);
```
Fyorl, line 542:
```suggestion
    const allowed = Hooks.call("dnd5e.dropItemSheetData", item, this, data);
```

module/applications/item/item-sheet.mjs
```diff
@@ -500,6 +501,70 @@ export default class ItemSheet5e extends ItemSheet {
...
+    }
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Handle the dropping of ActiveEffect data onto an Item Sheet
+   * @param {DragEvent} event                  The concluding DragEvent which contains drop data
+   * @param {object} data                      The data transfer extracted from the event
+   * @returns {Promise<ActiveEffect|boolean>}  The created ActiveEffect object or false if it couldn't be created.
```
Fyorl, line 557: Probably should be annotated with `@protected`.

module/applications/item/item-sheet.mjs
```diff
@@ -500,6 +501,70 @@ export default class ItemSheet5e extends ItemSheet {
...
+   */
+  async _onDropActiveEffect(event, data) {
+    const effect = await ActiveEffect.implementation.fromDropData(data);
+    if ( !this.item.isOwner || !effect ) return false;
+    if ( this.item.uuid === effect.parent.uuid || this.item.uuid === effect.origin ) return false;
```
Fyorl, line 562:
```suggestion
    if ( (this.item.uuid === effect.parent.uuid) || (this.item.uuid === effect.origin) ) return false;
```

Fyorl (changes requested):
Good stuff, thanks. Just some code style suggestions from me.

module/applications/item/item-sheet.mjs
arbron: I believe the `if` is there to handle future use cases where other items such as advancements are also draggable.

akrigline:
changes made @Fyorl thanks for the suggestions :)

Issue #1539, opened 2022-05-08
Add identifier to all items


arbron:
Add a similar identifier to spells to the one that was added to classes and subclasses. This will open up linking to spells via identifier like requested in #1509, as well as help build things like class spell lists and other features that need a consistent way to reference spells.

For example, an `"Eldritch Spear"` invocation could contain and advancement that locates the `"Eldritch Blast"` cantrip on an actor using its identifier `eldritch-blast`, and then modifies its distance appropriately.

PR #1767, opened 2022-09-03
[#1539] Add identifiers to spells

arbron:
This adds an `identifier` field to spells and sets them up on the SRD. This should be useful in identifying matching spells, representing spell lists, applying advancements to spells, and more.

This also modifies the default method for creating slugs to support spells with slashes like `"Blindness/Deafness"`. Formerly that would have created the identifier `blindnessdeafness` and now it adds a dash in the middle to create `blindness-deafness`.


Varriount:
How is this different from a spell's name?

arbron:
> How is this different from a spell's name? Is this something that would also make sense for other item types (features, regular items, etc.)?

The key difference between this and names is that this is stable regardless of localization. This should allow for building spell lists and certain other features without relying solely on full UUIDs.

I think the other item type that might most benefit from this is `feat`, which would allow some more powerful features in the advancements system. I figure that could be done as a separate PR though.

Varriount:
I think that this would be useful to implement for all item types, as it would allow, for example, a module to synchronize updates between items in a player's character sheet, the "unassociated" items available in the `items` sidebar, and items in compendiums. This is actually a real-world request I've made for the DDB Importer module: https://github.com/MrPrimate/ddb-importer/issues/251

p4535992:
the same mechanism should be implemented for effects with the addition of an "identifier" field.

Issue #1531, opened 2022-05-04
Move carrying capacity size multipliers to CONFIG & fix Powerful Build implementation


arbron:
Currently how much an actor's carrying capacity is multiplied based on their size is hardcoded into the `_computeEncumbrance` method. This data should be moved to the config so existing multipliers can be changed and new sizes can be added.

The implementation of `"Powerful Build"` at the moment also simply multiplies the encumbrance by two, rather than treating the character as one size-class larger. For the most part this is fine, but it means that a small creature with that flag would be treated like a large creature rather than a medium one. It also wouldn't support any custom multipliers that aren't powers of two.

Issue #1276, opened 2021-09-01
Implement the Variant Encumbrance rule from the SRD


aaclayton:
Originally in GitLab by @ben-barbour

From page 80 of the [SRD v5.1](http://media.wizards.com/2016/downloads/DND/SRD-OGL_V5.1.pdf).

It would also be great if we could alter the STR multipliers and maybe supply optional active effects for the thresholds, but that might be module territory.

benbarbour:
I've started taking a stab at this. Right now it's just cosmetic changes: the weight bar in the inventory changes color and tooltips are added over the bar for each encumbrance level. I've got a couple of questions: 

1.Would it be appropriate to add active effects (or optional active effects) for the encumbrance levels. Should they actually reduce speed?
2. I probably need help choosing good colors, if color changes for the weight bar are desirable at all... I don't do a lot of UX/UI work.

I think that if there are no status effects (or at least color changes on the weight bar) then maybe this shouldn't be done in the system. It would be a whole system setting to turn on some tooltips over the three weight bar sections, and people might not even notice.

aaclayton:
> Right now it's just cosmetic changes: the weight bar in the inventory changes color and tooltips are added over the bar for each encumbrance level. I've got a couple of questions:

I think that sounds like a good place to start.

> Would it be appropriate to add effects (or optional effects) for the encumbrance levels. Should they actually reduce speed? I don't think they can add disadvantage.

I think that's a good idea to explore, adding a status effect when the user is encumbered would be a good integration. Depending on the size of the change, it might be better to make that a follow-up MR to the cosmetic changes for simplicity of review and merge.

> Are color changes for the weight bar desirable, and if so, which colors?

I think color changes are a good idea but I would suggest two general tenets:

1. Make sure that information is not provided *only* using color, there needs to be some other indication also.
2. Make sure the color palette is not dramatically in contrast with the rest of the UI of the sheet. I'm not sure exactly which colors to suggest, but perhaps a light blue when unencumbered that translates towards a darker purplish-blue as you become more heavily encumbered? I'm not dead set on that and open to other choices. Regardless of which way we go it would be advisable to make those colors CSS variables so that modules can override the color palette with alternate choices.

benbarbour:
Got a couple of questions about this.

1. Should I add colour and/or tooltips to the sheet when the variant rule is not in use? There are only two states in that case: "under carrying capacity" and "exceeding carrying capacity".
[...]

PR #1775, opened 2022-09-06
[#1276, #1531] Implement variant encumbrance option rule

benbarbour:
This adds a toggle to the DnD5E system settings called "Use Variant Encumbrance Rules". If enabled, the encumbrance bar starts to turn from the default blue to a more purple/red as it passes the various thresholds. Please suggest better colours! A tooltip is also added to the entire encumbrance bar at each stage.

image

The rules don't have defined terms for "below 'Encumbered' (<5 times STR score) or "above 'heavily encumbered'" (>15 times STR score), so I used the words "Unencumbered" and "Over Encumbered" for the tooltip on those stages.

I have another commit that consolidates `tokenSizes` into `actorSizes` as recommended [here](https://github.com/foundryvtt/dnd5e/pull/1623#discussion_r912129520), and can add it to the PR if that's still a desirable change.

I'm also trying to sneak in a whitespace fix that was annoying me every time I ran the linter with `--fix`. I can open another PR for it if you'd rather.

----------------

I'd like to consider also adding status effects for the encumbrance levels above "Unencumbered" in this PR, but I'm unsure if they should be active effects or not. The speed modifications should be easy, the only question is if they should apply to all speeds or just walking/land speed. I'm pretty sure the former, but it's up for debate. However, if I understand the current state of active effects correctly, then we can't add the second effect of "Heavily Encumbered" (which is "disadvantage on d20 tests involving the 3 physical attributes"), right?

----------------

Closes [#1276]. May also fix [#1531].

### DIFF
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -1210,26 +1210,12 @@
   position: relative;
 }
 .dnd5e.sheet.actor .encumbrance .encumbrance-bar {
-  --encumbered-0-color: #6c8aa5;
-  --encumbered-33-color: #7e8197;
-  --encumbered-66-color: #907789;
-  --encumbered-max-color: #a56c79;
   position: absolute;
   top: 1px;
   left: 1px;
-  background: var(--encumbered-0-color);
   height: 8px;
   border: 1px solid #cde4ff;
   border-radius: 2px;
-}
-.dnd5e.sheet.actor .encumbrance.encumbered-33 .encumbrance-bar {
-  background: var(--encumbered-33-color);
-}
-.dnd5e.sheet.actor .encumbrance.encumbered-66 .encumbrance-bar {
-  background: var(--encumbered-66-color);
-}
-.dnd5e.sheet.actor .encumbrance.encumbered-max .encumbrance-bar {
-  background: var(--encumbered-max-color);
 }
 .dnd5e.sheet.actor .encumbrance .encumbrance-label {
   height: 10px;
@@ -1247,12 +1233,6 @@
   display: block;
   position: absolute;
 }
-.dnd5e.sheet.actor .encumbrance .encumbrance-breakpoint.encumbrance-33 {
-  left: 33%;
-}
-.dnd5e.sheet.actor .encumbrance .encumbrance-breakpoint.encumbrance-66 {
-  left: 66%;
-}
 .dnd5e.sheet.actor .encumbrance .arrow-up {
   bottom: 0;
   width: 0;
@@ -1260,6 +1240,9 @@
   border-left: 4px solid transparent;
   border-right: 4px solid transparent;
   border-bottom: 4px solid #666;
+}
+.dnd5e.sheet.actor .encumbrance .arrow-up.encumbered {
+  border-bottom: 4px solid #000;
 }
 .dnd5e.sheet.actor .encumbrance .arrow-down {
   top: 0;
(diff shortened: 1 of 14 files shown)


Varriount:
Is "Over Encumbered" (which should actually be "Over-encumbered") necessary? I don't see anything in the rules about it. Additionally, there isn't really much distinction between "over-encumbered" and "heavily encumbered" grammatically, so I'm not sure that "over-encumbered" is the best term. Perhaps "Very Heavily Encumbered"?

It's also worth putting some thought into how this might be customized by modules. If feasible, it would be nice to be able to set the points at which a encumbrance condition exists, the colors to use, and what active effects or functions (if any) should be applied/called (perhaps a hook should even be added).

benbarbour:
The non-variant rules say that (15xSTR)lbs is the maximum you can carry and the variant rules don't change that. I guess there's not a name for carrying more than that because that's an illegal state for a character to be in. I don't think the system should stop character sheets from holding over that limit (that would be annoying), but I do think the state should be highlighted and differentiated from "Heavily Encumbered". Maybe "Over Weight Limit" would be a better tooltip? If we decide to keep the current wording, I'll at least add the hyphen though, thanks!

> It's also worth putting some thought into how this might be customized by modules.

All the colours and thresholds are in variables that modules can override fairly easily, I think. Please correct me if I'm wrong there. Although, the code assumes three evenly spaced thresholds, which _might_ be worth changing to make it easier for modders? The three tickmarks on the bar would need to be placed by code instead of just placed by CSS if that's implemented.

Speaking of mods, I was thinking of one popular (_citation needed_) homebrew rule already. That is: "equipped items only add some fraction of their weight to encumbrance". This stops many classes from being encumbered by their starting equipment and punishes heavy armour wearers a bit less. Does anyone think it would be worth it to add `DND5E.encumbrance.equippedMultipler: 1` to the system and use it in `_prepareEncumbrance()` so that a simple world script can adjust it to 0.5 or whatever, even if there's no UI for it? If not, I may split out the tallying of item weights into it's own function so that modders can overwrite just that part of the algorithm easily.

Varriount:
> Does anyone think it would be worth it to add DND5E.encumbrance.equippedMultipler: 1 to the system and use it in _prepareEncumbrance() so that a simple world script can adjust it to 0.5 or whatever, even if there's no UI for it? If not, I may split out the tallying of item weights into it's own function so that modders can overwrite just that part of the algorithm easily.

I think this would be better implemented by an outside module. The real question is, do we expect modules to overwrite this via patching (a la https://github.com/ruipin/fvtt-lib-wrapper), via a CONFIG value (setting it to a function), or via an event hook?

benbarbour:
While I'm generally familiar with patching and (the idea behind) libWrapper, and event hooks, I don't have much experience with the Foundry and dnd5e codebases and so it makes comparing the pros and cons a bit difficult. I also don't see any examples of storing a function in the config (if you're talking about config.mjs / the DND5E object), but I suppose I could add one.

I don't see a problem with splitting the part of `_prepareEncumbrance()` that tallies inventory weight into it's own function, maybe `_tallyInventoryWeight()`, so that a module could overwrite it with a version that applies a multiplier to equipped items.

Should I bother making the tickmarks on the weight bar data-driven so that you can make arbitrary thresholds, and arbitrary numbers of thresholds, just my modifying `DND5E.encumbrance.variant`? That seems more like module territory to me, but I'm a greenhorn! :)

I'm thinking I'll change "Over Encumbered" to "Exceeding Carrying Capacity", unless there are objections. Those are words used in the actual rules, although not that exact phrasing.

Varriount:
>  I don't have much experience with the Foundry and dnd5e codebases and so it makes comparing the pros and cons a bit difficult.

Hm, I've reviewed how the CONFIG data is currently used, and it doesn't seem to store any functions, so ignore that previous suggestion.

Here's a quick rundown of the other two possibilities:
  - Hooks allow one or more functions to be called in response to an event. While useful, there's a certain amount of consideration that needs to be made by developers adding the hook (just like adding any public API).
  - `libWrapper` allows a module to patch just about any function or method. The benefit is that this enables modules to override mechanisms that were not otherwise overrideable. The downside is that this is very prone to breakage - if the function or method being patched changes (for example, its return type is changed), then the patched functions will break as well.

> Should I bother making the tickmarks on the weight bar data-driven so that you can make arbitrary thresholds, and arbitrary numbers of thresholds, just my modifying DND5E.encumbrance.variant?

Personally, that's what I would go for, but I think something like that is probably best decided by @arbron .

> How about changing "Over Encumbered" to "Exceeding Carrying Capacity"? Those are words used in the actual rules, although not that exact phrasing.

I think that's actually better than "Over Encumbered".

p4535992:
_How about changing "Over Encumbered" to "Exceeding Carrying Capacity"? Those are words used in the actual rules, although not that exact phrasing._
 I called **Overburdened** on my [Variant Encumbrance Module](https://github.com/VanirDev/VariantEncumbrance/).

itteerde:
I would like it to be with AE, and with the AE affecting all movement speeds the character has.

benbarbour:
I'm happy to add Active Effects if it will aid in getting this merged, but it was suggested in #1276 that I start with a cosmetic change and then add the effects in another PR.

arbron:
@benbarbour Most of this has been incorporated on the new 3.0 release of the system. If you find anything still missing for encumbrance, don't hesitate to file a new issue

PR #1780, opened 2022-09-07
[#1465, #1771, #1774] Add ability to move advancements from one item to another

### LINKED ISSUE #1465: Advancement migrations
As we are still building out the different types of advancements, there is a going to be a period of time where the SRD classes have some but not all of the advancement types. When future types are added, the system should be able to intelligently update classes on actors with new advancement types without requiring the manual replacement of those classes and the re-selection of advancement choices.

This could involve fetching data from the `"Classes (SRD)"` compendium, and then checking the advancement data on all world classes and adding any new advancements as necessary.

### LINKED ISSUE #1771: Add ability to drag individual advancements from one item to another
Dragging an advancement from one item's advancement tab to another should copy the advancement over. If the destination item is already on an actor, prompt for whether the new advancement should be applied.

### LINKED ISSUE #1774: Unapply advancement from actor when it is deleted
Currently, if an advancement is deleted from an item while on an actor then it's changes are left behind. Deleting an advancement should remove any changes it made.

arbron:
This adds a few tools to improve the usability of advancements, specifically to aid in the process of adding advancements to an item that is already on a character sheet.

This adds the ability to use drag-and-drop to copy a single advancement from one item to another. It also allows for dragging an entire item onto the Advancement tab of another to copy all advancements over. Regardless of which method is used, the system will check to ensure that each advancement doesn't already exist on the target (by comparing IDs) and that it is valid (only one Hit Points Advancement allowed). If the whole item is dragged over, it will display a prompt to give users an option of only copying over certain advancements.

If the item the advancements are being copied onto is embedded, the system will then show the Advancement Manager to  apply any new advancements.

In addition, deleting an advancement from an embedded item will now unapply that advancement before deletion.

Resolves #1465 
Resolves #1771 
Resolves #1774


akrigline:
likely to conflict with #1766 because we're both adding dragDrop to ItemSheet
I'd recommend splitting your drop handling of Advancement Data into a method dedicated to that `_onDropAdvancement`.

module/applications/item/item-sheet.mjs
```diff
@@ -610,6 +619,53 @@ export default class ItemSheet5e extends ItemSheet {
...
+    if ( showDialog ) {
+      try {
+        advancements = await AdvancementMigrationDialog.createDialog(this.item, advancements);
+      } catch(err) {
+        return false;
+      }
+    }
+
+    if ( !advancements.length ) return false;
+    if ( this.item.isEmbedded && !game.settings.get("dnd5e", "disableAdvancements") ) {
```
Fyorl, line 656: Shouldn't we have this check for disabled advancements much earlier? If advancements are disabled we can bail out right at the top of the method, no?

Fyorl (changes requested):
Looks good to me, only minor suggestions, and one point of clarification: Where we talk about 'migrating' advancements here, we mean manual, UI-based migrations with users dragging and dropping individual advancements or advancement-containing-items onto other items? We're not attempting to perform automated migrations on users worlds at all?

module/applications/item/item-sheet.mjs
arbron: We still want to copy over the advancements to the item even if advancements are disabled, we just don't want to launch the `AdvancementManager` to apply their changes to the actor.

arbron:
> Looks good to me, only minor suggestions, and one point of clarification: Where we talk about 'migrating' advancements here, we mean manual, UI-based migrations with users dragging and dropping individual advancements or advancement-containing-items onto other items? We're not attempting to perform automated migrations on users worlds at all?

This PR is just for manually dragging advancements to an item.

Fully-automated migrations on advancements aren't going to be possible because they require per-player action, but I was thinking we could implement a system in the UI to notify players that there are new advancements available and make it easy to apply them. That feature if we decide to implement it will be built on this foundation.

module/applications/advancement/advancement-manager.mjs
```diff
@@ -303,6 +381,17 @@ export default class AdvancementManager extends Application {
...
+  /* -------------------------------------------- */
+
+  /**
+   * Determine the proper working level either from the provided item or from the cloned actor.
+   * @param {Item5e} item  Item being advanced. If class or subclass, its level will be used.
+   * @returns {number}     Working level.
+   */
+  currentLevel(item) {
+    return item.system.levels ?? item.class?.system.levels ?? manager.clone.system.details.level;
+  }
```
Fyorl, line 393: I guess this also needs to take an actor parameter since `manager` is not defined in this context:

```suggestion
  currentLevel(item, actor) {
    return item.system.levels ?? item.class?.system.levels ?? actor.system.details.level;
  }
```

module/applications/advancement/advancement-manager.mjs
```diff
@@ -197,10 +242,43 @@ export default class AdvancementManager extends Application {
...
+   * @returns {AdvancementManager}  Prepared manager. Steps count can be used to determine if advancements are needed.
+   */
+  static forDeletedAdvancement(actor, itemId, advancementId, options) {
+    const manager = new this(actor, options);
+    const clonedItem = manager.clone.items.get(itemId);
+    const advancement = clonedItem?.advancement.byId[advancementId];
+    if ( !advancement ) return manager;
+
+    const minimumLevel = advancement.levels[0];
+    const currentLevel = this.currentLevel(clonedItem);
```
Fyorl, line 260:
```suggestion
    const currentLevel = this.currentLevel(clonedItem, manager.clone);
```

module/applications/advancement/advancement-manager.mjs
```diff
@@ -123,6 +123,54 @@ export default class AdvancementManager extends Application {
...
+   * @param {Advancement[]} advancements  Dropped advancements to add.
+   * @param {object} options              Rendering options passed to the application.
+   * @returns {AdvancementManager}  Prepared manager. Steps count can be used to determine if advancements are needed.
+   */
+  static forNewAdvancement(actor, itemId, advancements, options) {
+    const manager = new this(actor, options);
+    const clonedItem = manager.clone.items.get(itemId);
+    if ( !clonedItem || !advancements.length ) return manager;
+
+    const currentLevel = this.currentLevel(clonedItem);
```
Fyorl, line 139:
```suggestion
    const currentLevel = this.currentLevel(clonedItem, manager.clone);
```

Fyorl (changes requested):

module/applications/item/item-sheet.mjs
Fyorl: I suppose that does make sense; fair enough then

Fyorl (approved):

Issue #1779, opened 2022-09-07
Using max() in an attack roll bonus field on an item results in StringTerm error


IHaveThatPower:
Foundry 10.284, dnd5e 2.0.2, all modules disabled. Conferring on Discord in the dnd5e channel, this does _not_ happen on Foundry v9.

Repro steps:
1. Create a weapon item and give it to an actor
2. Set the Weapon Attack Action Type to Ranged Weapon Attack
3. Set Ability Modifier to Dexterity
4. Set the Attack Roll Bonus to `1 + max(@ablities.int.mod, @abilities.dex.mod) - @abilities.dex.mod`
5. Set the Damage Formula to `2d10 + max(@abilities.dex.mod, @abilities.int.mod) + 1`
6. Click the 'Roll' icon for the weapon
7. Click the 'Attack' option for the weapon

Expected:
Roll evaluates

Actual:
No evaluation occurs, and the following error appears in the console:
```
foundry.js:11551 Uncaught (in promise) Error: Unresolved StringTerm 5,6 requested for evaluation
[Detected 1 package: system:dnd5e]
    at StringTerm.evaluate (foundry.js:11551:11)
    at Roll._evaluate (foundry.js:9113:42)
    at async MathTerm._evaluate (foundry.js:10966:7)
    at async D20Roll._evaluate (foundry.js:9100:9)
    at async d20Roll (dice.mjs:96:3)
    at async Item5e.rollAttack (item.mjs:1323:18)
    at async Item5e._onChatCardAction (item.mjs:1803:9)
```
(`5,6` was the evaluation of the two terms in the `max()` function above in my test use case)

The damage term evaluates without issue.

arbron:
@Fyorl Can you check this out and see if this might be a core issue?

Fyorl:
It's being caused by a quite obscure interaction between the `D20Roll` class and the underlying `Roll` code. So technically a `dnd5e` bug. If you select the token and run this in the console:

```
Roll.parse("1d20 + 1 + max(@abilities.int.mod, @abilities.dex.mod) - @abilities.dex.mod + @mod", _token.actor.getRollData());
```

[...]

PR #1786, opened 2022-09-08
[#1779] Fix issue with d20 rolls failing when they include min() or max()

arbron:
The `D20Roll` class will no longer throw an error if it isn't provided a valid D20 formula. Instead it will now just check validity at certain points before performing D20-specific customization.

### DIFF
--- a/module/dice/d20-roll.mjs
+++ b/module/dice/d20-roll.mjs
@@ -16,10 +16,7 @@
 export default class D20Roll extends Roll {
   constructor(formula, data, options) {
     super(formula, data, options);
-    if ( !((this.terms[0] instanceof Die) && (this.terms[0].faces === 20)) ) {
-      throw new Error(`Invalid D20Roll formula provided ${this._formula}`);
-    }
-    if ( !this.options.configured ) this.configureModifiers();
+    if ( this.#validD20Roll && !this.options.configured ) this.configureModifiers();
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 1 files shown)


module/dice/d20-roll.mjs
```diff
@@ -16,10 +16,7 @@
 export default class D20Roll extends Roll {
   constructor(formula, data, options) {
     super(formula, data, options);
-    if ( !((this.terms[0] instanceof Die) && (this.terms[0].faces === 20)) ) {
-      throw new Error(`Invalid D20Roll formula provided ${this._formula}`);
-    }
-    if ( !this.options.configured ) this.configureModifiers();
+    if ( this.#validD20Roll && !this.options.configured ) this.configureModifiers();
```
Fyorl, line 19: Given that `configureModifiers` checks for `#validD20Roll`, I don't think we also need to guard it here.

module/dice/d20-roll.mjs
```diff
@@ -57,6 +54,16 @@ export default class D20Roll extends Roll {
 
   /* -------------------------------------------- */
 
+  /**
+   * Does this roll start with a d20?
+   * @type {boolean}
+   */
+  get #validD20Roll() {
```
Fyorl, line 61: I don't think this needs to be hard-private.

Fyorl (changes requested):

module/dice/d20-roll.mjs
```diff
@@ -57,6 +54,17 @@ export default class D20Roll extends Roll {
 
   /* -------------------------------------------- */
 
+  /**
+   * Does this roll start with a d20?
+   * @type {boolean}
+   * @protected
+   */
+  get _validD20Roll() {
```
Fyorl, line 62: This is my fault for not specifying originally, but it seemed like it could be a useful public method. For example, creating a `new D20Roll` and checking `roll.validD20Roll`, then performing some action. Might be more relevant now that the constructor no longer throws an error. I'll leave it up to you if you think it's worth changing before merge.

Fyorl (approved):

Issue #1769, opened 2022-09-04
Hit die rolls can incorrectly cause characters to lose health


Haxxer:
**Describe the bug**
When a character has a negative constitution modifier and roll a hit die, they can incorrectly lose health. A negative result should instead be treated as a zero:

"Short Rest (p. 186). After applying your Constitution modifier to a Hit Die roll, you regain a minimum of 0 hit points (6th printing)."

https://media.wizards.com/2017/dnd/downloads/PH-Errata.pdf 

**To Reproduce**
1. Set up a wizard with a 1 in constitution
2. Start a short rest
3. Roll a hit dice
4. More than likely, wizard has lost health

**Expected behavior**
The roll should be treated as a zero.

**Setup**
 - Foundry Version: 10.283
 - DND5e Version: 2.0.2

PR #1792, opened 2022-09-09
[#1769] Fix hit dice subtracting health with negative con mod, rework rollHitDie to use base Roll rather than DamageRoll

arbron:
This fixes the issue with rolling hit dice with a negative con modifier potentially subtracting hit points.

It also reworks `rollHitDie` to use the base `Roll` rather than `DamageRoll`, which avoids some formatting weirdness from the formula simplification present in `DamageRoll` that is not necessary for this purpose. This is a breaking change because it removes the ability to show a roll configuration dialog for hit die rolls (disabled by default, but possible before by passing `{dialog: true}` to the `rollHitDie` method). It also alters the config data passes to the `dnd5e.preRollHitDie` hook, potentially causing issues with people who have implemented that new hook already.

That said, I think it is better for hit die to be rolled as vanilla `Roll` rather than `DamageRoll`.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1224,14 +1224,13 @@ export default class Actor5e extends Actor {
   /* -------------------------------------------- */
 
   /**
-   * Roll a hit die of the appropriate type, gaining hit points equal to the die roll plus your CON modifier
-   * @param {string} [denomination]       The hit denomination of hit die to roll. Example "d8".
-   *                                      If no denomination is provided, the first available HD will be used
-   * @param {object} options              Additional options which modify the roll.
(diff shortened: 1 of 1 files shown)


arbron:
> Is it the case then that there was really no way to spawn the hit die configuration dialog via the UI? It was purely available via the API?

Nowhere in the core system are we displaying that dialog, but it is possible that 3rd party sheets or other modules might have been using it.

Fyorl (approved):

Fyorl:
Given that we expect the impact to be minimal, and that there is recourse for modules to take over the process entirely with `preRollHitDie` if they want to re-add the dialog, I think we can put this in the next patch. Would you mind raising an issue to add an (optional) intermediate hit die dialog for situational bonuses?

Issue #1788, opened 2022-09-09
Support for nested object structures in `module/config.mjs`


djinnstar:
It would be great if `config.mjs` could support nested object structures for localization. Currently, there is no way to localize the following structure, for example:

```
DND5E.itemSubtypes = {
  armor: {
    light: {
      label: "DND5E.ItemTypeLight",
      baseItems: {
        padded: "DND5E.BaseItemPadded",
        leather: "DND5E.BaseItemLeather",
        studded: "DND5E.BaseItemStudded"
      }
    }
  }
}
```

Issue #1016, opened 2021-03-02
Add Armor, Tool, and Weapon Proficiencies to Class Item


arbron:
Similar to issue #270 which was just implemented, add proficiency data for armor, weapons, and tools so that information can be programmatically available when adding classes.

```json
{
  "proficiencies": {
    "armor": {
      "grants": [],
      "multiclass": [],
      "value": []
    },
    "weapon": {
      "grants": [],
      "multiclass": [],
      "value": []
    },
    "tool": {
      "grants": [],
      "multiclass": [],
      "value": []
    }
  }
}
```

Here is an example for the Bard class:

```json
{
  "proficiencies": {
    "armor": {
      "grants": ["lgt"],
      "multiclass": ["lgt"],
      "value": ["lgt"]
    },
    "weapon": {
      "grants": ["sim", "handcrossbow", "longsword", "rapier", "shortsword"],
      "multiclass": [],
      "value": ["sim", "handcrossbow", "longsword", "rapier", "shortsword"]
    },
    "tool": {
      "grants": [{ "choices": ["music"], "count": 3 }],
      "multiclass": [{ "choices": ["music"], "count": 1 }],
      "value": ["drum", "flute", "viol"]
    }
  }
}
```

PR #1794, opened 2022-09-09
[#1788] Recursive `config` localization + item subtyping

djinnstar:
Solves [#1788]. 

Adds the ability to recursively localize nested object structures in `module/config.mjs` using a new method, `module/utils.mjs -> preLocalizeDeep`. 

Also adds an example where this could be useful - `DND5E.itemSubtypes`, which specifies the standard item subtypes with their supertypes. This sort of structure would be helpful to create more robust solutions for #1016 and #774 (but is not strictly necessary).


arbron:
This approach for representing item subtypes was specifically discussed when the current `baseItem` system was built and it was decided that storing base item data in compendiums was the better solution.

Without a specific use case that would depend on this localization technique, I don't see any reason to modify the current behavior.

PR #1800, opened 2022-09-10
[#774, #1082] Item Proficiency Rework

djinnstar:
[this is large but all the changes rely on each other, sorry!]

[relies on #1794]

Reworks the system's item proficiency with the goal of fixing a few problems:

- #1082 - item proficiencies are reclassified as derived data (heavily borrowed from a commit by @arbron), which causes item proficiencies to auto-update with an option for manual selection
- #774 - item proficiencies on a character sheet are now represented by structures like this (i.e. for a character with expertise in alchemist's supplies and regular proficiency in mason's tools): `thisActor.system.itemProficiencies.tool => {alchemist: 2, mason: 1}`
- Provide UI support for these changes
- Provide migrations to the new data structures

Known shortcomings: 
- The CSS on the new `itemProficiencyConfig` sheet could be prettier.


arbron:
These are two major changes (not even mentioning the change to base items) that should be handled as two different Pull Requests so they can be reasonable reviewed in isolation. It seems like the change to actor proficiency to support expertise should be handled first so we can discuss just those changes.

I don't see any reason to completely replace the current "base items" system with a completely new implementation. If you feel that such a major change is necessary, you'll have to explain why.

arbron:
Closing this because a different approach for handling tool expertise has been merged #2113

PR #1814, opened 2022-09-12
[#1776, #1810, #1811] Fix various bugs

### LINKED ISSUE #1776: Global Skill Check bonus does not display when re-editing
Foundry 10.284
dnd5e: 2.0.2
All modules disabled: Yes

If you set a Global Skill Bonus, then when you go back in to re-edit it, the value does not appear.  E.g...

1. Click the cog next to acrobatics to bring up the skill editor.  Enter a 5 in the "Global Skill Check" and then click Submit...
image

2. Roll some skill checks, see that the 5 is being added to the rolls...
image

3. Edit the Acrobatics (or any other) skill settings again.  See that the 5 is no longer being displayed...
image

4. Click submit and then roll some skill checks again.  The 5 is no longer added...
image

### LINKED ISSUE #1810: When a spell is upcast, the type and DC of the saving throw is not shown
Cast a spell at its normal level. Everything works fine.

Cast the spell at a higher level, and the button for the saving throw only displays 'Saving Throw'.

image

(Foundry 10.285, dnd5e 2.0.2)

### LINKED ISSUE #1811: Class/subclass sheets throw error when invalid identifier is used
The `ItemSheet5e#_getSubmitData` method still points to `data.identifier` rather than `system.identifier`.

arbron:
- Fix bug with Global Skill Check Bonus not appearing in Skill Config dialog (resolves #1776)
- Fix bug with `prepareFinalAttributes` not being called on upcast spells (resolves #1810)
- Fix bug when invalid identifier is set on class or subclass sheet (resolves #1811)

### DIFF
--- a/module/applications/actor/skill-config.mjs
+++ b/module/applications/actor/skill-config.mjs
@@ -40,7 +40,7 @@ export default class ActorSkillConfig extends DocumentSheet {
       skill: src.system.skills?.[this._skillId] || {},
       skillId: this._skillId,
       proficiencyLevels: CONFIG.DND5E.proficiencyLevels,
-      bonusGlobal: src.system.bonuses?.skill
+      bonusGlobal: src.system.bonuses?.abilities.skill
     };
   }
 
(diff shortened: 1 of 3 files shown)


Fyorl (approved):
Good fixes, thanks Jeff

Issue #1809, opened 2022-09-12
The new compendium SRD rules window does not take account of compendium translations


riccisi:
Hi, I'm the developer of the Babele module. The new custom window of the SRD Rules compendium expects English language content. This is a problem for those who use Babele with a translated compendium content. 
I recommend extracting the words 'Chapter' and 'Appendix' in keys that can be translated in order to make the code language independent.
![CleanShot 2022-09-12 at 10 27 45@2x](https://user-images.githubusercontent.com/6490765/189609769-30b8a0fd-82c8-439f-87a7-d5ac0cc5ae19.png)

Thanks.

arbron:
My recommendation is that we store the chapter and appendix data in the flags on the journal entries to avoid potentially messy string parsing. This will avoid difficult situations like languages where the number or letter might come before the other words.

PR #1815, opened 2022-09-12
[#1809] Convert chapter & appendix data in rules compendium into flags

arbron:
Added some flags to the journal entries in the SRD rules compendium to allow sorting of entries regardless of localization:

```json
"flags": {
  "dnd5e": {
    "type": "chapter",
    "position": 1
  }
}
```


module/applications/srd-compendium.mjs
```diff
@@ -31,23 +31,16 @@ export default class SRDCompendium extends Compendium {
...
-      const [, appB] = b.name.match(/^Appendix ([A-Z])/) || [];
-      if ( chA && !chB ) return -1;
-      else if ( chB && !chA ) return 1;
-      if ( chA && chB ) return Number(chA) - Number(chB);
-      if ( appA && appB ) return appA.charCodeAt(0) - appB.charCodeAt(0);
-      return 0;
+      const positions = { chapter: 0, appendix: 100 };
+      const getOrder = o => (positions[o.flags?.dnd5e?.type] ?? 200) + (o.flags?.dnd5e?.position ?? 0);
+      return getOrder(a) - getOrder(b);
     });
```
Fyorl, line 44: `getOrder` could probably be moved outside since it doesn't need to close over any of the variables inside the `sort` call.

```js
const getOrder = o => {
  return ({chapter: 0, appendix: 100}[o.flags?.dnd5e?.type] ?? 200) + (o.flags?.dnd5e?.position ?? 0);
};
```

Then `sort` can become:
```js
.sort((a, b) => getOrder(a) - getOrder(b));
```

Fyorl (changes requested):

Fyorl (approved):

PR #1816, opened 2022-09-13
[#1437] Changed price spell scroll lvl9

### LINKED ISSUE #1437: Spell scroll price data is homebrew?

The price of "spell scroll" items (e.g. [1st level](https://gitlab.com/foundrynet/dnd5e/-/blob/master/packs/src/items/scroll/spell-scroll-1st-level.json#L15)) seems to be taken from the "[Sane Magic Item Prices](https://forums.giantitp.com/showthread.php?424243-Sane-Magic-Item-Prices)" homebrew (CTRL-F "Spell Scroll").

While these prices are perhaps more sensible than those one could conjure up from the DMG/XGE rules/tables, this is an opinionated solution inappropriate for inclusion in the core dnd5e system. (Or if it's staying, then the 9th level price is wrong; it should be `10240` rather than `5120`!)

kjolern:
I have changed the price on the spell scroll 9th level from 5120 to 10240.  This fixes #1437


kjolern:
Do you need any changes on this @arbron ?

PR #1828, opened 2022-09-24
Centralize fastForward detection for roll configs, allow explicit false

akrigline:
## Goals
1. Centralize `fastForward` and `critical` detection within the `dice.mjs` helper functions, rather than also have some event key detection within `rollDamage` and similar methods.
    - `_determineCriticalMode` and `_determineAdvantageMode` should be the only methods concerned with determining these roll modifiers to make it easier to refactor those methods later (e.g. in #1591)
2. Respect when `damageRoll` or `d20Roll` is invoked with `fastForward` explicitly set to `false`, showing the roll config dialog as normal.
3. Maintain the current API call signatures and behaviors, except in the case where `fastForward` is explicitly set to false.

### Problem being Solved

> It is not possible to explicitly override a roll config to be `fastForward: false` if the config also has an `event` which would normally fast-forward a roll config.

<details>
<summary>Example scenario</summary>

```javascript
item.rollAttack({ event: { altKey: true }, fastForward: false, advantage: true });
```
As an API consumer I expect the attack roll configuration dialog to appear, but with `advantage` as the default button (This is the current behavior if I leave out the `event`):

Demonstration of the expectation working with this PR's code changes

When the `event` is populated, the `fastForward: false` is ignored because of the `event` handling in various places of the workflow, but most notably the two utility functions `_determineCriticalMode` and `_determineAdvantageMode`.

My assertion is that determining the `fastForward`-ness of a roll should solely be the responsibility of these methods, and that is what this PR makes true. It also uses nullish coalescing, so as not to overwrite an explicit `false`.
</details>

### Methodology

Remove the `event` checking that happened during item damage and attack roll workflows (from `_onChatCardAction` and `rollDamage`).

Remove the defaults for the `critical` and `fastForward` booleans from all methods/configs, allowing them to be `undefined` until `_determineAdvantageMode` and `_determineCriticalMode`'s evaluation (this did not affect `d20Roll`'s `critical` number argument).

### DIFF
--- a/module/dice/dice.mjs
+++ b/module/dice/dice.mjs
@@ -26,7 +26,7 @@
  * @property {boolean} [reliableTalent]  Allow Reliable Talent to modify this roll?
  *
  * ## Roll Configuration Dialog
- * @property {boolean} [fastForward]     Should the roll configuration dialog be skipped?
+ * @property {boolean} [fastForward]           Should the roll configuration dialog be skipped?
  * @property {boolean} [chooseModifier=false]  If the configuration dialog is shown, should the ability modifier be
  *                                             configurable within that interface?
  * @property {string} [template]               The HTML template used to display the roll configuration dialog.
@@ -136,7 +136,7 @@ function _determineAdvantageMode({event, advantage=false, disadvantage=false, fa
  *
  * ## Critical Handling
  * @property {boolean} [allowCritical=true]  Is this damage roll allowed to be rolled as critical?
- * @property {boolean} [critical]      Apply critical to this roll (unless overridden by modifier key or dialog)?
+ * @property {boolean} [critical]            Apply critical to this roll (unless overridden by modifier key or dialog)?
  * @property {number} [criticalBonusDice]    A number of bonus damage dice that are added for critical hits.
  * @property {number} [criticalMultiplier]   Multiplier to use when calculating critical damage.
  * @property {boolean} [multiplyNumeric]     Should numeric terms be multiplied when this roll criticals?
@@ -144,7 +144,7 @@ function _determineAdvantageMode({event, advantage=false, disadvantage=false, fa
  * @property {string} [criticalBonusDamage]  An extra damage term that is applied only on a critical hit.
  *
  * ## Roll Configuration Dialog
- * @property {boolean} [fastForward]  Should the roll configuration dialog be skipped?
+ * @property {boolean} [fastForward]        Should the roll configuration dialog be skipped?
  * @property {string} [template]            The HTML template used to render the roll configuration dialog.
  * @property {string} [title]               Title of the roll configuration dialog.
  * @property {object} [dialogOptions]       Additional options passed to the roll configuration dialog.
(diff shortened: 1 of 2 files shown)


module/dice/dice.mjs
```diff
@@ -26,7 +26,7 @@
  * @property {boolean} [reliableTalent]  Allow Reliable Talent to modify this roll?
  *
  * ## Roll Configuration Dialog
- * @property {boolean} [fastForward=false]     Should the roll configuration dialog be skipped?
+ * @property {boolean} [fastForward]     Should the roll configuration dialog be skipped?
```
Fyorl, line 29:
```suggestion
 * @property {boolean} [fastForward]           Should the roll configuration dialog be skipped?
```

module/dice/dice.mjs
```diff
@@ -136,15 +136,15 @@ function _determineAdvantageMode({event, advantage=false, disadvantage=false, fa
  *
  * ## Critical Handling
  * @property {boolean} [allowCritical=true]  Is this damage roll allowed to be rolled as critical?
- * @property {boolean} [critical=false]      Apply critical to this roll (unless overridden by modifier key or dialog)?
+ * @property {boolean} [critical]      Apply critical to this roll (unless overridden by modifier key or dialog)?
```
Fyorl, line 139:
```suggestion
 * @property {boolean} [critical]            Apply critical to this roll (unless overridden by modifier key or dialog)?
```

module/dice/dice.mjs
```diff
@@ -136,15 +136,15 @@ function _determineAdvantageMode({event, advantage=false, disadvantage=false, fa
...
  * @property {number} [criticalBonusDice]    A number of bonus damage dice that are added for critical hits.
  * @property {number} [criticalMultiplier]   Multiplier to use when calculating critical damage.
  * @property {boolean} [multiplyNumeric]     Should numeric terms be multiplied when this roll criticals?
  * @property {boolean} [powerfulCritical]    Should the critical dice be maximized rather than rolled?
  * @property {string} [criticalBonusDamage]  An extra damage term that is applied only on a critical hit.
  *
  * ## Roll Configuration Dialog
- * @property {boolean} [fastForward=false]  Should the roll configuration dialog be skipped?
+ * @property {boolean} [fastForward]  Should the roll configuration dialog be skipped?
```
Fyorl, line 147:
```suggestion
 * @property {boolean} [fastForward]        Should the roll configuration dialog be skipped?
```

module/documents/item.mjs
```diff
@@ -1352,7 +1352,8 @@ export default class Item5e extends Item {
    * @returns {Promise<DamageRoll>}        A Promise which resolves to the created Roll instance, or null if the action
    *                                       cannot be performed.
    */
-  async rollDamage({critical=false, event=null, spellLevel=null, versatile=false, options={}}={}) {
+  async rollDamage({critical, event=null, spellLevel=null, versatile=false, options={}}={}) {
+    debugger;
```
Fyorl, line 1356: Leftover `debugger` statement.

Fyorl (changes requested):

module/documents/item.mjs
akrigline:
```suggestion
```

Fyorl (approved):

PR #1829, opened 2022-09-24
Use RAW GP as standard for currency conversions

cstby:
I'm the maintainer of World Currency 5e, a module that allows for currency customization. This pull request changes the currency conversion schema so that all currencies are defined in terms of GP (according to rules as written).

There are a few problems with the current schema:
- Module creators often hard code their own conversion logic because it's convoluted to use the system conversion numbers to convert a currency to/from GP.
- Completely omitting a currency like Electrum can't be achieved while keeping copper and silver.
- We can't have more than one currency that's worth more than GP.
- Defining new currencies in terms of each other isn't intuitive for most GMs.

I believe that the current schema was designed to make the "Currency Converter" logic straightforward. I have updated the currency converter logic so that it'll work exactly the same as before.

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -14,7 +14,6 @@ ______      ______ _____ _____
 |___/ \\___/\\/___/ \\____/\\____/
 _______________________________`;
 
-
 /**
  * The set of Ability Scores used within the system.
  * @enum {string}
@@ -627,27 +626,28 @@ preLocalize("consumableTypes", { sort: true });
 DND5E.currencies = {
   pp: {
     label: "DND5E.CurrencyPP",
-    abbreviation: "DND5E.CurrencyAbbrPP"
+    abbreviation: "DND5E.CurrencyAbbrPP",
+    conversion: 0.1
   },
   gp: {
     label: "DND5E.CurrencyGP",
     abbreviation: "DND5E.CurrencyAbbrGP",
-    conversion: {into: "pp", each: 10}
+    conversion: 1
   },
   ep: {
     label: "DND5E.CurrencyEP",
     abbreviation: "DND5E.CurrencyAbbrEP",
-    conversion: {into: "gp", each: 2}
+    conversion: 2
   },
   sp: {
     label: "DND5E.CurrencySP",
     abbreviation: "DND5E.CurrencyAbbrSP",
-    conversion: {into: "ep", each: 5}
+    conversion: 10
   },
   cp: {
     label: "DND5E.CurrencyCP",
     abbreviation: "DND5E.CurrencyAbbrCP",
-    conversion: {into: "sp", each: 10}
+    conversion: 100
   }
 };
 preLocalize("currencies", { keys: ["label", "abbreviation"] });
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -617,10 +617,12 @@ preLocalize("consumableTypes", { sort: true });
 
 /**
  * The valid currency denominations with localized labels, abbreviations, and conversions.
+ * The conversion number defines how many of that currency are equal to one GP (according
+ * to the rules as written).
  * @enum {{
  *   label: string,
  *   abbreviation: string,
- *   [conversion]: {into: string, each: number}
+ *   conversion: number,
  * }}
  */
 DND5E.currencies = {
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1783,20 +1783,25 @@ export default class Actor5e extends Actor {
   convertCurrency() {
     const curr = foundry.utils.deepClone(this.system.currency);
     const conversion = Object.entries(CONFIG.DND5E.currencies);
-    conversion.sort((a, b) => a[1].conversion-b[1].conversion);
-    var change = 0
-    for ( let [c, data] of conversion) {
-      const rate = data.conversion
-      if ( !rate ) continue
-      change = change + curr[c] / rate
-      }
+    conversion.sort((a, b) => a[1].conversion - b[1].conversion);
+
+    var change = 0;
     for ( let [c, data] of conversion) {
       const rate = data.conversion;
       if ( !rate ) continue;
+
+      change = change + curr[c] / rate;
+    }
+
+    for ( let [c, data] of conversion) {
+      const rate = data.conversion;
+      if ( !rate ) continue;
+
       let amt = Math.floor(change * rate);
       curr[c] = amt;
       change = (change - amt / rate);
     }
+
     return this.update({"data.currency": curr});
   }
 
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1793,9 +1793,9 @@ export default class Actor5e extends Actor {
     for ( let [c, data] of conversion) {
       const rate = data.conversion;
       if ( !rate ) continue;
-      let amt = (change * rate).toFixed(8);
-      curr[c] = Math.floor(amt);
-      change = (change - (Math.floor(amt) / rate)) ;
+      let amt = Math.floor(change * rate);
+      curr[c] = amt;
+      change = (change - amt / rate);
     }
     return this.update({"data.currency": curr});
   }
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1779,18 +1779,25 @@ export default class Actor5e extends Actor {
    * carried by an Actor.
    * @returns {Promise<Actor5e>}
    */
+
   convertCurrency() {
     const curr = foundry.utils.deepClone(this.system.currency);
     const conversion = Object.entries(CONFIG.DND5E.currencies);
-    conversion.reverse();
-    for ( let [c, data] of conversion ) {
-      const t = data.conversion;
-      if ( !t ) continue;
-      let change = Math.floor(curr[c] / t.each);
-      curr[c] -= (change * t.each);
-      curr[t.into] += change;
+    conversion.sort((a, b) => a[1].conversion-b[1].conversion);
+    var change = 0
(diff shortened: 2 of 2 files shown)


module/documents/actor/actor.mjs
```diff
@@ -1779,18 +1779,25 @@ export default class Actor5e extends Actor {
...
+      const rate = data.conversion
+      if ( !rate ) continue
+      change = change + curr[c] / rate
+      }
+    for ( let [c, data] of conversion) {
+      const rate = data.conversion;
+      if ( !rate ) continue;
+      let amt = (change * rate).toFixed(8);
+      curr[c] = Math.floor(amt);
+      change = (change - (Math.floor(amt) / rate)) ;
```
Varriount, line 1798: Why the calls to `toFixed` and `Math.floor`? Wouldn't a single call to `Math.floor` at the end work?

module/documents/actor/actor.mjs
```diff
@@ -1779,18 +1779,25 @@ export default class Actor5e extends Actor {
    * carried by an Actor.
    * @returns {Promise<Actor5e>}
    */
+
   convertCurrency() {
     const curr = foundry.utils.deepClone(this.system.currency);
```
Varriount, line 1784: Hm, I wonder if a deep clone is really needed here, or if a shallow clone would suffice.

cstby: I only copied this from the previous convert currency function. I'm not too keen on the exact difference. Do you know why the original author may have used deep clone?

module/documents/actor/actor.mjs
cstby: Great point. I thought `toFixed` was needed to prevent small decimals from appearing. I've refactored based on your suggestions and I'm still getting the expected result.

module/documents/actor/actor.mjs
cstby: I can't find a shallow clone. What would you suggest using here instead?

Varriount: Use `array.slice()` (or rather, `this.system.currency.slice`) with no arguments.

cstby: Since `this.system.currency` isn't an array, the slice method won't work. If we want to avoid a deep clone, `const curr = this.system.currency;` does work as expected.

cstby:
I'd propose that the current iteration be merged. Refactoring working code is outside the scope of my original goal. Am I correct in understanding that `deepClone` here doesn't produce any measurable problem?

Varriount:
It doesn't. I don't think that's a problem which is preventing this PR from being merged though.

cstby:
Thanks. Any other changes you'd like me to consider at this point?

module/documents/actor/actor.mjs
```diff
@@ -1779,18 +1779,25 @@ export default class Actor5e extends Actor {
...
+      change = change + curr[c] / rate
+      }
+    for ( let [c, data] of conversion) {
+      const rate = data.conversion;
+      if ( !rate ) continue;
+      let amt = Math.floor(change * rate);
+      curr[c] = amt;
+      change = (change - amt / rate);
+    }
+    return this.update({"data.currency": curr});
```
Varriount, line 1800:
```suggestion
    conversion.sort((a, b) => a[1].conversion - b[1].conversion);

    var change = 0;
    for ( let [c, data] of conversion) {
      const rate = data.conversion;
      if ( !rate ) continue;

      change = change + curr[c] / rate;
    }

    for ( let [c, data] of conversion) {
      const rate = data.conversion;
      if ( !rate ) continue;

      let amt = Math.floor(change * rate);
      curr[c] = amt;
      change = (change - amt / rate);
    }

    return this.update({"data.currency": curr});
```

cstby:
Thanks for the above. I'm still not as familiar with Javascript as I'd like to be, so I appreciate the formatting suggestions. Anything else that can be tightened up?  Also, are you in agreement with the proposed currency conversion schema in `config.mjs`?

Varriount:
Is the conversion the same as it was previously?

cstby:
Yes, the values of each currency are the same relative to each other.

Varriount:
Then the only thing I can really think of is adding to the documentation on [config.msj, line 620](https://github.com/foundryvtt/dnd5e/pull/1829/files#diff-5ab7cd05ad5f0b1b20878ec2638ebb732cb4466abc03963a02fc1c1aba759dc9L620), mentioning that the conversions are relative to a single "base" unit (which in this case has a one-to-one equivalency to GP).

cstby:
Good call. I've updated the documentation, as you suggested.

module/documents/actor/actor.mjs
```diff
@@ -1779,18 +1779,30 @@ export default class Actor5e extends Actor {
    * carried by an Actor.
    * @returns {Promise<Actor5e>}
    */
+
```
arbron, line 1782:
```suggestion
```

module/config.mjs
```diff
@@ -618,36 +617,39 @@ preLocalize("consumableTypes", { sort: true });
 
 /**
  * The valid currency denominations with localized labels, abbreviations, and conversions.
+ * The conversion number defines how many of that currency are equal to one GP (according
+ * to the rules as written).
  * @enum {{
```
arbron, line 622: To make this more in line with other new configuration changes, this should probably be converted to a `typedef` which allows for a more detailed description of each property.

```suggestion
/**
 * Configuration data for a currency.
 *
 * @typedef {object} CurrencyConfiguration
 * @property {string} label         Localized label for the currency.
 * @property {string} abbreviation  Localized abbreviation for the currency.
 * @property {number} conversion    Number by which this currency should be multiplied to arrive at a standard value.
 */

/**
 * The valid currency denominations available in actors & containers.
 * @enum {CurrencyConfiguration}
```

module/documents/actor/actor.mjs
```diff
@@ -1779,18 +1779,30 @@ export default class Actor5e extends Actor {
...
   convertCurrency() {
     const curr = foundry.utils.deepClone(this.system.currency);
```
arbron, line 1784: It would probably be a good time to change this to the whole word for clarity.

```suggestion
    const currency = foundry.utils.deepClone(this.system.currency);
```

module/documents/actor/actor.mjs
```diff
@@ -1779,18 +1779,30 @@ export default class Actor5e extends Actor {
...
-    return this.update({"system.currency": curr});
+    conversion.sort((a, b) => a[1].conversion - b[1].conversion);
+
+    var change = 0;
+    for ( let [c, data] of conversion) {
+      const rate = data.conversion;
+      if ( !rate ) continue;
+
+      change = change + curr[c] / rate;
+    }
```
arbron, line 1794: - No need to use `var`
- The configuration definition states that the `conversion` is always required, so we shouldn't need to verify it exists here
- This would probably be clearer as a reduce

```suggestion
    let change = conversion.reduce((change, [c, data]) => {
      return change + (currency[c] / data.conversion);
    }, 0);
```

module/documents/actor/actor.mjs
```diff
@@ -1779,18 +1779,30 @@ export default class Actor5e extends Actor {
...
+
+    for ( let [c, data] of conversion) {
+      const rate = data.conversion;
+      if ( !rate ) continue;
+
+      let amt = Math.floor(change * rate);
+      curr[c] = amt;
+      change = (change - amt / rate);
+    }
```
arbron, line 1803:
```suggestion
    for ( const [c, data] of conversion) {
      const amt = Math.floor(change * data.conversion);
      currency[c] = amt;
      change -= amt / data.conversion;
    }
```

module/documents/actor/actor.mjs
```diff
@@ -1779,18 +1779,30 @@ export default class Actor5e extends Actor {
...
+
+    return this.update({"data.currency": curr});
```
arbron, line 1805:
```suggestion
    return this.update({"data.currency": currency});
```

arbron (changes requested):
Some documentation and code style changes, but overall it looks good.

module/documents/actor/actor.mjs
cstby: Thank you for these suggestions! They all look great, and I just have one comment on this one.

One use case I'm trying to support is the removal of currencies. Many players who use my module simply want to remove electrum and platinum. Until now, I had intended to support their need by allowing `conversion` to be undefined. That is why I explicitly check that `conversion` exists.

I'd like to get your advice on how best to support the need to remove currencies. 

One approach might be to advise 3rd-party modules to simply remove unneeded currency objects _entirely_ from `DND5E.currencies`. The currency converter code above would work as expected. However, any hardcoded references to specific currencies would break. Those hardcoded references would need to be changed to an approach that iterates through the currencies stored in `DND5E.currencies`.

Another approach might be to add a "removed" or "hidden" boolean value to the currency configuration schema. We would have to change the logic of the currency converter above to check for that value and skip any currencies that aren't supposed to exist.

arbron: I think in order to function in the same way as removing `skills` or `abilities`, users should just need to remove a currency from the `CONFIG.currencies` object (or replace it entirely with their own set) and everything else should work.

There are a few changes I have planned with the upcoming data model stuff that should also help improve support for this. For now I'd say just worry about conversion stuff and we can take a look at anything else needed later.

aaclayton (approved):
Merging this in and then handling remaining requested changes in a follow-up commit.

aaclayton:
See <https://github.com/foundryvtt/dnd5e/commit/7153fce3def063d7e7d53498e3b1be6cadc86fd2> for follow-up tidying.

PR #1838, opened 2022-10-01
Move Source Input field

EndlesNights:
Within the current NPC actor sheet I find that the upper list elements used directly underneath the actors name to use inefficient spacing, and the text is rather cramped. For this example below, we can see that the most NPCs who have both a type and a subtype will likely not have enough space to full display both words as one is cut off, along with any other additional text within the alignment text field that isn't part of the core nine. 
Furthermore the only reason there is enough space in the Source field is because acronyms are being used,  Which is fine for any core material, since the community has a shared core glossary, there is just not enough space to easily display the sources for 3rd party content.

default

The basic changes within this PR is to move the location of the source field from the upper list, to the form groups near the bottom of the page. This dramatically increases the amount of space to display information at the top of the page, making the type + subtype, field of an NPC readable from a glance, and so the users does not need to move their mouse cursor over and read the hover title text. Any additional text in the alignment field is also more accessibly read further from a simple change to the flex css on the size display which has been limited in its flex width by 50%.
The source text input field also has significantly more space for text, there is now enough room that acronyms are no longer a necessity, core acronyms  should still be used, but this creates more space for 3rd party creators to have their names properly displayed.

Another small change that I made to the sheet, was ever so slightly changing the height of the ability attributes boxes, so that the bottom of their list aligns with the bottom of the skills list.

image

### DIFF
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -365,6 +365,9 @@
 .dnd5e.sheet .sheet-header .summary li:last-child {
   border-right: none;
 }
+.dnd5e.sheet.npc .sheet-header .summary li:first-child {
+  flex:0.5;
+}
 .dnd5e.sheet .sheet-navigation {
   flex: 0 0 30px;
   margin-bottom: 5px;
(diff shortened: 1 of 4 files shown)


Varriount:
Perhaps it would be better in the header section, above the CR rating?

arbron:
Closing this because we are beginning the complete re-design of NPC sheets, so this isn't relevant anymore.

PR #1840, opened 2022-10-02
Polymorph Improvements

### LINKED ISSUE #1100: Transformation Tool - No Stats option

### LINKED ISSUE #1139: Add preserve passive/inactive/temporary effects toggle when using polymorph/wild-shape

### LINKED ISSUE #1391: [Feature Request] Add option to duplicate token display options when transforming an actor
Originally in GitLab by @Fyorl

Description copied from original report by @woutermeuwis

Currently when transforming/polymorphing an actor, the resulting token seems to use some default settings. Some of the token settings would make sense to be duplicated from the original actor, for example:

* Display Name
* Display Attribute bars
* Attribute bar settings

### LINKED ISSUE #1820: Deprecation warnings in Actor5e#transformInto and Actor5e#revertOriginalForm
image

image

### LINKED ISSUE #1823: Issues with reverting transformed unlinked token actors
Currently, when transforming an unlinked token actor, the 'revert transformation' header button is missing. However, if you close the sheet and re-open it, the button re-appears. Clicking it does not appropriately revert the transformation: Only the token's size seems to change but not its image or name. The actor remains in the transformed state.

### LINKED ISSUE #1825: Sight is not inherited from transform target when transforming an actor
## Steps to Reproduce
1. Create a creature and modify its prototype token: Enable 'has vision' and give it a 60 ft. vision range.
2. Drag that creature onto the sheet of another creature that has vision enabled by 0 ft. vision range.
3. Select the 'polymorph' transform option.
4. Drag the transformed creature to the scene.
5. Observe that its vision range is 0 rather than the 60 ft of the creature it transformed into.

The deprecation warning in #1820 mentions token sight so this is likely related.

MrPrimate:
A reimplementation of #1604.

This adds some new features:
- Keep Self: Transforms only the image, useful for disguise self, alter self etc. Resolves #1100
- Remove Effects: A selection of options to allow removal of various types of effects. Resolves #1139
- Retains a tokens display name and attribute bar display settings during transform. Resolves #1391
- Fixes Deprecation warnings Resolves #1820
- Allows unlinked tokens to transform back to original token data. Resolves #1823
- Fixes sight transfer issues. Resolves #1825

Some other improvements to follow. But this seemed like a good time for a first review.

Changed dialogue looks like:
new-apperance


Varriount:
Regarding the UI, perhaps it might be better to have the options be contained within a scrollable box? Similar to how the module settings window works.

Some other things that spring to mind (but that may be outside the scope of this PR):
  - Have the presets be in a `select` element. When a preset is selected, the relevant inputs would be updated to reflect the preset's configuration.
  - Have the polymorph presets be data driven. That is, there would be a value in the system's CONFIG variable like:
    ```javascript
      CONFIG.polymorphPresets = {
        "DND5E.PolymorphSelf": {
            keepSelf: true,
            transformTokens: true,
            removeAE: true,
            removeOriginAE: true,
            removeOtherOriginAE: true,
            removeFeatAE: true,
            removeSpellAE: true,
            removeEquipmentAE: true,
            removeClassAE: true,
            removeBackgroundAE: true
        },
        ...
      }
  ```

module/documents/actor/actor.mjs
```diff
@@ -1945,11 +2040,13 @@ export default class Actor5e extends Actor {
      * @param {Actor5e} target                 The target actor into which to transform.
      * @param {object} data                    The data that will be used to create the new transformed actor.
      * @param {TransformationOptions} options  Options that determine how the transformation is performed.
+     * @param {boolean} [renderSheet=true]   Render the sheet of the transformed actor after the polymorph
```
Fyorl, line 2043: Let's put this in an options object to send to the hook too, to match my earlier suggestion for the method arguments.

module/documents/actor/actor.mjs
```diff
@@ -1973,52 +2074,105 @@ export default class Actor5e extends Actor {
    * If this actor was transformed with transformTokens enabled, then its
    * active tokens need to be returned to their original state. If not, then
    * we can safely just delete this actor.
+   * @param {boolean} [renderSheet] Render Sheet after revert the transformation.
```
Fyorl, line 2077: See my suggestion above about moving this into an object.

module/documents/actor/actor.mjs
```diff
@@ -1962,6 +2059,10 @@ export default class Actor5e extends Actor {
       newTokenData._id = t.id;
       newTokenData.actorId = newActor.id;
       newTokenData.actorLink = true;
+
+      const dOriginalActorData = foundry.utils.getProperty(d, "flags.dnd5e.originalActor");
```
Fyorl, line 2063: `originalActor` should just be an `id`, rather than some object of data, no?

```suggestion
      const dOriginalActor = foundry.utils.getProperty(d, "flags.dnd5e.originalActor");
```

module/documents/actor/actor.mjs
```diff
@@ -1864,78 +1881,156 @@ export default class Actor5e extends Actor {
...
+    // Gather previous actor data
+    const previousActorIds = this.getFlag("dnd5e", "previousActorIds") || [];
+    previousActorIds.push(this._id);
+    foundry.utils.setProperty(d.flags, "dnd5e.previousActorIds", previousActorIds);
+
+    // Update unlinked Tokens, and grab a copy of any actorData adjustments to re-apply
     if ( this.isToken ) {
       const tokenData = d.prototypeToken;
       delete d.prototypeToken;
+      d.flags.dnd5e.previousActorData = duplicate(this.token.actorData);
```
Fyorl, line 2024:
```suggestion
      d.flags.dnd5e.previousActorData = this.token.toObject().actorData;
```

And should this flag be stored on the `TokenDocument` itself rather than placed in its `actorData` override?

module/documents/actor/actor.mjs
```diff
@@ -1973,52 +2074,105 @@ export default class Actor5e extends Actor {
...
+      const baseActor = original ? original : game.actors.get(this.token.actorId);
+      if ( !baseActor ) {
+        ui.notifications.warn(game.i18n.format("DND5E.PolymorphRevertNoOriginalActorWarn", {
+          reference: this.getFlag("dnd5e", "originalActor")
+        }));
+        return;
+      }
+      const prototypeTokenData = await baseActor.getTokenDocument();
+      const actorData = foundry.utils.getProperty(this, "flags.dnd5e.previousActorData");
+      const tokenUpdate = duplicate(this.token);
```
Fyorl, line 2110:
```suggestion
      const tokenUpdate = this.token.toObject();
```

module/documents/actor/actor.mjs
```diff
@@ -1973,52 +2074,105 @@ export default class Actor5e extends Actor {
...
+      }
+      tokenUpdate.sight = prototypeTokenData.sight;
+      tokenUpdate.detectionModes = prototypeTokenData.detectionModes;
+
       await this.sheet.close();
-      const actor = this.token.getActor();
-      actor.sheet.render(true);
+      await canvas.scene?.deleteEmbeddedDocuments("Token", [this.token._id]);
+      const results = await canvas.scene?.createEmbeddedDocuments("Token", [tokenUpdate], { keepId: true, render: true });
+      const actor = results.find(r => r._id === tokenUpdate._id).actor;
```
Fyorl, line 2125:
```suggestion
      const actor = await TokenDocument.implementation.create(tokenUpdate, {parent: canvas.scene, keepId: true, render: true}).actor;
```

module/documents/actor/actor.mjs
```diff
@@ -1973,52 +2074,105 @@ export default class Actor5e extends Actor {
...
+      await this.unsetFlag("dnd5e", "previousActorIds");
     }
 
-    // Delete the polymorphed version of the actor, if possible
-    const isRendered = this.sheet.rendered;
-    if ( game.user.isGM ) await this.delete();
-    else if ( isRendered ) this.sheet.close();
-    if ( isRendered ) original.sheet.render(isRendered);
+    // Delete the polymorphed version(s) of the actor, if possible
+    if (game.user?.isGM) {
```
Fyorl, line 2160: Shouldn't be possible for `game.user` to not exist.
```suggestion
    if ( game.user.isGM ) {
```

module/documents/actor/actor.mjs
```diff
@@ -1973,52 +2074,105 @@ export default class Actor5e extends Actor {
...
+        id !== original.id // Is not original Actor Id
+        && game.actors?.get(id) // Actor still exists
+      ).concat([this.id]); // Add this id
+
+      for (const id of idsToDelete) {
+        const actorToDelete = game.actors.get(id);
+        if (actorToDelete) {
+          await actorToDelete.delete();
+        }
+      }
```
Fyorl, line 2171:
```suggestion
      await Actor.implementation.deleteDocuments(idsToDelete);
```

templates/apps/polymorph-prompt.hbs
```diff
@@ -1,22 +1,36 @@
...
-		<label class="checkbox" for="transformTokens">
-			<input type="checkbox" id="transformTokens" name="transformTokens"
-				{{checked content.options.transformTokens}}>
-			{{localize 'DND5E.PolymorphTokens'}}
-		</label>
-	</div>
-	{{/unless}}
+<div class="polymorph two-column">
+    <div class="left-column">
+		<h3>{{localize 'DND5E.PolymorphCustomTransformation'}}</h3>
```
Fyorl, line 3: Looks like some tabs snuck in. I know the original was tabs too, but it's good to correct it now.

templates/apps/polymorph-prompt.hbs
```diff
@@ -1,22 +1,36 @@
...
+            <label class="checkbox" for="transformTokens">
+                <input type="checkbox" id="transformTokens" name="transformTokens"
+                    {{checked content.options.transformTokens}}>
+                {{localize 'DND5E.PolymorphTokens'}}
+            </label>
+        </div>
+        {{/unless}}
+    </div>
+    <div class="right-column">
+		<h3>{{localize 'DND5E.PolymorphEffectTransformation'}}</h3>
```
Fyorl, line 24: Tabs here

Fyorl (changes requested):
Before this is merged, I need to sit down with it and run through all the polymorphing tests, which I haven't had time to do yet. But this is looking good.

In addition to the code review, I had some additional comments:

1. We should make the polymorph dialog a little wider now so that the button text doesn't wrap.
2. `transformInto` and `revertOriginalForm` have become such gigantic methods that I wonder if we should be splitting this functionality out into its own separate helper class.

Let's not do (2) as part of this PR though as it will obfuscate the technical changes.

module/applications/actor/base-sheet.mjs
```diff
@@ -816,14 +817,46 @@ export default class ActorSheet5e extends ActorSheet {
             keepMental: true,
             mergeSaves: true,
             mergeSkills: true,
-            transformTokens: rememberOptions(html).transformTokens
+            transformTokens: rememberOptions(html).transformTokens,
```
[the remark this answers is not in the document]
MrPrimate: I don't know if this will work, as it will take all the options selected in the dialogue, rather than using the defaults plus the over-rides that we want for these "quick select" options.

module/documents/actor/actor.mjs
MrPrimate: I'll loop back to this, I had problems storing it on the token for reasons I need to remember.

module/applications/actor/base-sheet.mjs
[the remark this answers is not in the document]
Fyorl: You are right, I had forgotten that we don't take all the sheet options when clicking these buttons. In fact, it seems we're taking all the AE options from the sheet, but actually I think we should treat these the same as the rest of the options. For example, with Polymorph and Wildshape, you explicitly do not benefit from equipment, so AEs from equipment should always be disabled. For Polymorph, you shouldn't retain benefits from your class features, but in Wildshape you do, etc.

I also like @Varriount's suggestion to have these presets on the `CONFIG` object since they're getting a bit lengthy:

```js
CONFIG.transformationPresets = {
  wildshape: {
    icon: '<i class="fas fa-paw"></i>',
    label: "DND5E.PolymorphWildShape",
    options: {
      keepBio: true,
      removeEquipmentAE: true,
      // ...
    }
  }
};
```

Fyorl:
This is probably a bit of churn at this stage, so I apologise for not mentioning it sooner, but it only just occurred to me that we write most of the 'custom transformation' options in terms of 'keeping' something, but the 'effect transformations' are written in terms of 'removing' them. Is that because we used to always keep all effects before?

module/applications/actor/base-sheet.mjs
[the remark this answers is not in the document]
MrPrimate: So the thing with wildshape is that sometimes you _can_ retain the item and effect, if the DM allows it/fits with the transformed creature, thus I left these as always customisable, as I suspect you can't please everyone with a default for the active effect removal/retention.

MrPrimate:
> This is probably a bit of churn at this stage, so I apologise for not mentioning it sooner, but it only just occurred to me that we write most of the 'custom transformation' options in terms of 'keeping' something, but the 'effect transformations' are written in terms of 'removing' them. Is that because we used to always keep all effects before?

Yes, the default was always to keep all effects on an actor, and for things like status or spell effects you probably want to keep most of them It seemed like a more natural fit here, and was also slightly less convoluted when implementing a removal system rather than figuring out which effects to keep.

module/applications/actor/base-sheet.mjs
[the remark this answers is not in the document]
Fyorl: The whole point of the presets is to provide a quick way to apply those specific transformation effects according to the rules. For polymorph, the rules are clear: You do not retain equipment benefits, you do not retain class/subclass feature benefits, so all of those active effects should be disabled by the preset and not left customisable.

For wildshape, you do always retain your class/subclass features so those active effects should always be enabled by the preset, etc. We do not offer a way to piecemeal select which equipment you do or do not retain the benefit of for wild shape, so keeping or removing active effects is going to be wrong in some case or another. In this case I would vote to set the preset to keep the active effects for equipment since that has been the status quo for this feature in our system until now.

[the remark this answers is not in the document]
MrPrimate: Okay, I'll make those changes, I'm away for the weekend so it will be next week.

Fyorl:
> Yes, the default was always to keep all effects on an actor, and for things like status or spell effects you probably want to keep most of them It seemed like a more natural fit here, and was also slightly less convoluted when implementing a removal system rather than figuring out which effects to keep.

Could we try reframing these options in terms of 'keeping' rather than 'removing', unless you have a strong reason why we shouldn't. I worry we might have some regret here later if we do not opt for this piece of consistency now. I also feel it currently offers a slight UX hitch with users having to switch contexts from 'keep' to 'remove' when looking across the two lists.

module/applications/actor/base-sheet.mjs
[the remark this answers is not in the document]
MrPrimate: Okay, this is done.

MrPrimate:
> > Yes, the default was always to keep all effects on an actor, and for things like status or spell effects you probably want to keep most of them It seemed like a more natural fit here, and was also slightly less convoluted when implementing a removal system rather than figuring out which effects to keep.
> 
> Could we try reframing these options in terms of 'keeping' rather than 'removing', unless you have a strong reason why we shouldn't. I worry we might have some regret here later if we do not opt for this piece of consistency now. I also feel it currently offers a slight UX hitch with users having to switch contexts from 'keep' to 'remove' when looking across the two lists.

Changed

lang/en.json
```diff
@@ -651,21 +651,34 @@
...
 "DND5E.PolymorphMergeSkills": "Merge skill proficiencies (take the highest)",
 "DND5E.PolymorphPromptTitle": "Transforming Actor",
+"DND5E.PolymorphRemoveAE": "Remove All Active effects (overrides other effect retention options).",
+"DND5E.PolymorphKeepBackgroundAE": "Keep effects that originate from your background?",
+"DND5E.PolymorphKeepClassAE": "Keep effects that originate from your class or subclass?",
+"DND5E.PolymorphKeepEquipmentAE": "Keep effects that originate from your equipment?",
+"DND5E.PolymorphKeepFeatureAE": "Keep effects that originate from your features?",
+"DND5E.PolymorphKeepOriginAE": "Keep effects that originate on the actor i.e. effects not attached to features/items/spells",
+"DND5E.PolymorphKeepOtherOriginAE": "Keep effects that originate on a different actor i.e. spells/items/features from opponents",
+"DND5E.PolymorphKeepSpellAE": "Keep effects that originate from your spells?",
```
Varriount, line 677: Looking through the (user-visible) terms used by the rest of the system, I think "effect" (rather than "Active Effect" or "active effect") should be used.

```suggestion
"DND5E.PolymorphRemoveAE": "Remove all effects (overrides other effect retention options).",
"DND5E.PolymorphKeepBackgroundAE": "Keep effects that originate from your background?",
"DND5E.PolymorphKeepClassAE": "Keep effects that originate from your class or subclass?",
"DND5E.PolymorphKeepEquipmentAE": "Keep effects that originate from your equipment?",
"DND5E.PolymorphKeepFeatureAE": "Keep effects that originate from your features?",
"DND5E.PolymorphKeepOriginAE": "Keep effects that originate from this actor (i.e. effects not attached to features, items, or spells)?",
"DND5E.PolymorphKeepOtherOriginAE": "Keep effects that originate from a different actor (i.e. effects attached to another actor's feature, items, or spells)?",
"DND5E.PolymorphKeepSpellAE": "Keep effects that originate from your spells?",
```

lang/en.json
```diff
@@ -651,21 +651,34 @@
...
 "DND5E.PolymorphRestoreTransformation": "Restore Transformation",
+"DND5E.PolymorphRevertNoOriginalActorWarn": "No original actor is found with reference '{reference}'",
```
Varriount, line 679:
```suggestion
"DND5E.PolymorphRevertNoOriginalActorWarn": "No original actor was found with reference '{reference}'",
```

lang/en.json
MrPrimate: Ah, thanks! I thought I had changed all of these!

module/documents/actor/actor.mjs
```diff
@@ -1864,78 +1882,158 @@ export default class Actor5e extends Actor {
...
-      const oa = o.system.abilities[k];
-      const prof = abilities[k].proficient;
-      if ( keepPhysical && ["str", "dex", "con"].includes(k) ) abilities[k] = oa;
-      else if ( keepMental && ["int", "wis", "cha"].includes(k) ) abilities[k] = oa;
-      if ( keepSaves ) abilities[k].proficient = oa.proficient;
-      else if ( mergeSaves ) abilities[k].proficient = Math.max(prof, oa.proficient);
-    }
+    if ( !keepSelf ) {
+      const sightSource = keepVision ? o.prototypeToken : source.prototypeToken;
+      for ( const k of ["range", "angle", "visionMode", "color", "attenuation", "brightness", "saturation", "contrast"] ) {
```
Fyorl, line 1900: Need to add `enabled` to this list also.

module/documents/actor/actor.mjs
```diff
@@ -1973,52 +2077,102 @@ export default class Actor5e extends Actor {
...
-      return actor;
+      await canvas.scene?.deleteEmbeddedDocuments("Token", [this.token._id]);
+      const token = await TokenDocument.implementation.create(tokenUpdate, {
+        parent: canvas.scene, keepId: true, render: true
+      });
+      if ( isOriginalActor ) {
+        await this.unsetFlag("dnd5e", "isPolymorphed");
+        await this.unsetFlag("dnd5e", "previousActorIds");
+      }
+      if ( isRendered && renderSheet ) token.actor.sheet?.render(true);
```
Fyorl, line 2135: Rare, but this came up in testing where I would transform an actor, then drag its transformed version onto the canvas, then revert the original actor, and then attempt to revert the token. It worked well but threw an error here because (obviously) I had deleted the original actor.

```suggestion
      if ( isRendered && renderSheet ) token.actor?.sheet?.render(true);
```

Fyorl (changes requested):
Thanks for making those changes, this is coming along nicely. Also, good work with all the `renderSheet` logic by the way, that felt quite good while I was testing this.

module/documents/actor/actor.mjs
```diff
@@ -1864,78 +1882,158 @@ export default class Actor5e extends Actor {
...
-    // Keep biography
-    if ( keepBio ) d.system.details.biography = o.system.details.biography;
+      // Keep biography
+      if ( keepBio ) d.system.details.biography = o.system.details.biography;
+
+      // Keep senses
+      if ( keepVision ) d.system.traits.senses = o.system.traits.senses;
+
+      // Remove active effects
+      if ( removeAE ) {
```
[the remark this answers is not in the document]
MrPrimate: Mmmm, yes I like this.

less/apps.less
```diff
@@ -732,3 +732,20 @@
...
+/* ----------------------------------------- */
+
+
+.polymorph {
+  /* ----------------------------------------- */
+  /*  Two Column Configurations                */
+  /* ----------------------------------------- */
+  &.two-column {
+    display: grid;
+    grid-template-columns: 45% 65%;
```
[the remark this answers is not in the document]
MrPrimate: 👀

module/documents/actor/actor.mjs
```diff
@@ -1864,78 +1882,158 @@ export default class Actor5e extends Actor {
     d.system.attributes.ac.flat = target.system.attributes.ac.value; // Override AC
 
     // Token appearance updates
-    d.prototypeToken = {name: d.name, texture: {}};
+    if ( !hasProperty(d, "prototypeToken.texture.src") ) {
```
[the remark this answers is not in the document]
MrPrimate: Ah sorry, this can be cleaned up

MrPrimate:
@Fyorl ready for another round. 👯

MrPrimate:
> This looks pretty much good to go, thanks for the continued iteration. Did you manage to verify if we could store the `actorData` on the `TokenDocument` flags instead of as a flag within the `actorData` itself, per [this](https://github.com/foundryvtt/dnd5e/pull/1840#discussion_r990131219) comment?

Okay, changes made to this, my testing indicates it's fine on the token.

Fyorl (approved):
Looks good to me, thanks for doing this work. Polymorphing is quite complicated and fiddly so let's get this in 2.1 and people can start using and testing it before release.

PR #1844, opened 2022-10-05
Allow for `actor.applyTempHP(0)` to reset the `hp.temp` to 0.

thatlonelybugbear:
Change to the `applyTempHP` function to allow for resetting the `hp.temp` to 0 if needed with a `actor.applyTempHP(0)`

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -793,9 +793,9 @@ export default class Actor5e extends Actor {
     amount = parseInt(amount);
     const hp = this.system.attributes.hp;
 
-    // Update the actor if the new amount is greater than the current
+    // Update the actor if the new amount is greater than the current or if the new amount is 0.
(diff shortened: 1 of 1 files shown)


Fyorl:
I'm not sure we should be treating `0` as a special value in a function that works purely numerically. I can imagine code that performs some calculation and passes the result to this method, and it would be very surprising if it removed all temporary hit points simply because the calculation resulted in a `0`.

Passing a non-numeric value like `null` might be safer this behaviour. But I think if you want to reset temporary hit points, you should probably not be using `applyTempHP` at all, and instead should do it directly, i.e.:
```js
actor.update({"system.attributes.hp.temp": null});
```

thatlonelybugbear:
Agreed. 
`null` is understandably a better choice, but I think that I will close this PR as it doesn't seem that useful at a second glance.
More of a fleeting thought when I first used the `applyTempHP`.

PR #1845, opened 2022-10-05
[#1795] Add ability for items added with through advancement to trigger their own advancements

### LINKED ISSUE #1795: Adding a Subclass to an Actor Through Class Advancements does not add the Subclass's Advancement Features for that level
If you have an class with an advancement to add a subclass (for example, granting a Subclass item at level 3 for Artificer) does not give a prompt for accepting the subclass features for the level it was added (for example, the Alchemist level 3 features).

arbron:
Detects items added or removed during the advancement process and adds "synthetic" advancement steps if necessary for those items. This fixes the issue that prevents subclasses from automatically adding their initial features.

Adds a `AdvancementVisualizer` application that can be used to help debugging the advancement steps.

Changes subclass advancement to allow for selection of the subclass during the advancement process.

Resolves #1795

### DIFF
--- a/module/applications/advancement/advancement-manager.mjs
+++ b/module/applications/advancement/advancement-manager.mjs
@@ -11,6 +11,7 @@ import Advancement from "../../documents/advancement/advancement.mjs";
  * @property {Item5e} [class.item]        Class item that caused this advancement step.
  * @property {number} [class.level]       Level the class should be during this step.
  * @property {boolean} [automatic=false]  Should the manager attempt to apply this step without user interaction?
+ * @property {boolean} [synthetic=false]  Was this step created as a result of an item introduced or deleted?
  */
 
 /**
@@ -553,6 +554,7 @@ export default class AdvancementManager extends Application {
       do {
         const flow = this.step.flow;
         const type = this.step.type;
+        const preEmbeddedItems = this.clone.items.map(i => i);
 
         // Apply changes based on step type
         if ( (type === "delete") && this.step.item ) this.clone.items.delete(this.step.item.id);
@@ -564,6 +566,7 @@ export default class AdvancementManager extends Application {
         else if ( automaticData && flow ) await flow.advancement.apply(flow.level, automaticData);
         else if ( flow ) await flow._updateObject(event, flow._getSubmitData());
 
+        this._synthesizeSteps(preEmbeddedItems);
         this._stepIndex++;
 
         // Ensure the level on the class item matches the specified level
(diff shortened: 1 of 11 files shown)


module/applications/advancement/subclass-flow.mjs
```diff
@@ -1,13 +1,133 @@
...
+    } catch(err) {
+      return false;
+    }
+
+    if ( data.type !== "Item" ) return false;
+    const item = await Item.implementation.fromDropData(data);
+
+    // Ensure the dropped item is a subclass
+    if ( item.type !== "subclass" ) {
+      ui.notifications.warn(game.i18n.localize("DND5E.ADVANCEMENT.Subclass.Warning.InvalidType"));
```
krbz999, line 119:
```suggestion
      ui.notifications.warn("DND5E.ADVANCEMENT.Subclass.Warning.InvalidType", { localize: true });
```

module/applications/advancement/subclass-flow.mjs
```diff
@@ -1,13 +1,133 @@
...
+      data = JSON.parse(event.dataTransfer.getData("text/plain"));
+    } catch(err) {
+      return false;
+    }
+
+    if ( data.type !== "Item" ) return false;
+    const item = await Item.implementation.fromDropData(data);
+
+    // Ensure the dropped item is a subclass
+    if ( item.type !== "subclass" ) {
```
Fyorl, line 118: Should we also check `identifier` here? If you're dragging a subclass from outside the Compendium Browser (because you created it yourself maybe), then it will always be accepted as a valid subclass for the class because we override the existing identifier in `SubclassAdvancement#apply`.

module/data/fields/local-document-field.mjs
```diff
@@ -57,10 +57,23 @@ export default class LocalDocumentField extends foundry.data.fields.DocumentIdFi
...
 
+  /**
+   * Step up through model's parents to find the specified collection.
+   * @param {DataModel} model
+   * @param {string} collection
+   * @returns {EmbeddedCollection|void}
+   */
+  _findCollection(model, collection) {
+    if ( !model.parent ) return;
+    return model.parent[collection] ?? this._findCollection(model.parent, collection);
```
Fyorl, line 68:
```suggestion
    return model.parent.getEmbeddedCollection(collection) ?? this._findCollection(model.parent, collection);
```

templates/advancement/subclass-flow.hbs
```diff
@@ -0,0 +1,24 @@
...
+    <h3>{{{ title }}}</h3>
+    {{#if advancement.hint}}<p>{{ advancement.hint }}</p>{{/if}}
+
+    <div class="drop-target">
+        {{#if subclass}}
+        <div class="item-name flexrow">
+            <div class="item-image" style="background-image: url({{ subclass.img }});"></div>
+            <label class="flexrow">
+                <h4><a data-uuid="{{ subclass.uuid }}">{{ subclass.name }}</a></h4>
+                <a class="item-control item-delete" title="{{ localize 'DND5E.ItemDelete' }}">
```
Fyorl, line 11:
```suggestion
                <a class="item-control item-delete" title="{{ localize 'DND5E.ItemDelete' }}" data-action="deleteItem">
```

Fyorl (approved):
I tried quite a few things and couldn't really break it. Aside from the bugs reported, I don't think any of my comments need to hold up this PR, they are considerations for future improvements. Great to finally land this feature, nice work.

module/applications/advancement/subclass-flow.mjs
arbron: I didn't want to fully block support for "Universal Subclasses" (something WotC tried with the Strixhaven UA but didn't stick with).

Fyorl: Would such subclasses not have a blank identifier?

module/applications/advancement/advancement-manager.mjs
```diff
@@ -373,15 +388,25 @@ export default class AdvancementManager extends Application {
...
+   * @param {object} [options={}]
+   * @param {AdvancementStep[]} [options.findExisting]  Find if an existing matching flow exists.
+   * @returns {AdvancementFlow[]}                       Created or matched flow applications.
    * @protected
    */
-  static flowsForLevel(item, level) {
+  static flowsForLevel(item, level, { findExisting }={}) {
+    const match = (advancement, step) => (step.flow?.item.id === item.id)
+      && (step.flow?.advancement.id === advancement.id)
+      && (step.flow?.level === level);
```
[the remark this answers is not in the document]
arbron: I don't think caching on the advancement is feasible because on modify choices the item is completely deleted from the clone and then re-created, so all of its advancements are new instances.

This method of identification should be pretty robust because these three data points are the same we use to set the unique flow app ID, and no two flows should have the same IDs.

Perhaps the manager could have a cache map of flows if we need quicker lookup.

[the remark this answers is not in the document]
Fyorl: > This method of identification should be pretty robust because these three data points are the same we use to set the unique flow app ID, and no two flows should have the same IDs.

Ah ok, that gives some extra confidence then

PR #1847, opened 2022-10-06
[#1846] Allow ability type & defaults to be defined in config

### LINKED ISSUE #1846: Remove hardcoded default ability scores
Currently, ability scores all default to `10` except in a few cases where there are hardcoded exceptions like vehicle mental scores defaulting to 0 and Honor & Sanity on NPCs defaulting to equal Charisma and Wisdom by default. It would be useful to move these defaults into configuration so new abilities can specify their own defaults or the core ones can be overridden.

### LINKED ISSUE #1848: Allow specifying type on ability to control how it is handled in Polymorphing
Currently the list of ability scores that are retained during polymorphing when "Keep Mental Abilities" and "Keep Physical Abilities" are hardcoded, and thus don't support Honor, Sanity, or any custom ability scores. Adding a "type" value to configuration that specifies `physical` or `mental` would allow the polymorphing system to respond to these new abilities if desired.

arbron:
- Merge `DND5E.abilities` & `DND5E.abilityAbbreviations` into single object
- Add `type` to abilities to specify whether they are `"mental"` or `"physical"` and adjust polymorphing to make use of those types rather than hardcoded values (resolves #1848)
- Add a `defaults` object to those where different default values can be specified for different actor types and adjust `Actor5e#_prepareBaseAbilities` to make use of these defaults rather than hardcoding values (resolves #1846)

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -229,22 +229,16 @@ export default class Actor5e extends Actor {
   _prepareBaseAbilities() {
     if ( !("abilities" in this.system) ) return;
     const abilities = {};
-    for ( const key of Object.keys(CONFIG.DND5E.abilities) ) {
+    for ( const [key, config] of Object.entries(CONFIG.DND5E.abilities) ) {
       abilities[key] = this.system.abilities[key];
       if ( !abilities[key] ) {
         abilities[key] = foundry.utils.deepClone(game.system.template.Actor.templates.common.abilities.cha);
 
-        // Honor: Charisma for NPC, 0 for vehicles
-        if ( key === "hon" ) {
-          if ( this.type === "vehicle" ) abilities[key].value = 0;
-          else if ( this.type === "npc" ) abilities[key].value = this.system.abilities.cha?.value ?? 10;
+        let defaultValue = config.defaults?.[this.type] ?? 10;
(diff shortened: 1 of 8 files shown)


module/documents/actor/actor.mjs
```diff
@@ -2308,8 +2302,8 @@ export default class Actor5e extends Actor {
       for ( let k of Object.keys(abilities) ) {
         const oa = o.system.abilities[k];
         const prof = abilities[k].proficient;
-        if ( keepPhysical && ["str", "dex", "con"].includes(k) ) abilities[k] = oa;
-        else if ( keepMental && ["int", "wis", "cha"].includes(k) ) abilities[k] = oa;
+        if ( keepPhysical && (CONFIG.DND5E.abilities[k]?.type === "physical") ) abilities[k] = oa;
+        else if ( keepMental && (CONFIG.DND5E.abilities[k]?.type === "mental") ) abilities[k] = oa;
```
Fyorl, line 2306:
```suggestion
        const type = CONFIG.DND5E.abilities[k]?.type;
        if ( keepPhysical && (type === "physical") ) abilities[k] = oa;
        else if ( keepMental && (type === "mental") ) abilities[k] = oa;
```

Fyorl (approved):
LGTM. Minor code suggestion. Feel free to merge when ready.

PR #1854, opened 2022-10-07
[#1852] Allow for choosing a skill's ability on the character sheet

### LINKED ISSUE #1852: Allow for selection of skill ability on the character sheet
Being able to change the ability used for the skill check partway through the process causes some issues with our roll pipeline whereby we must revisit any ability check bonuses and modifiers after the intermediary dialog has been confirmed, rather than resolving these up-front.

If we move the ability selection to the sheet, we keep the flexibility of choosing an ability to use for the skill check, while also simplifying our roll pipeline as we can guarantee the ability used for the roll will not change partway though.

image

Fyorl:
- Closes #1852 

Unblocks #1760 & #1650

image

### DIFF
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -981,7 +981,7 @@
   cursor: default;
 }
 .dnd5e.sheet.actor ul.skills-list {
-  flex: 0 0 180px;
+  flex: 0 0 212px;
   list-style: none;
   margin: 0 5px 0;
   padding: 0;
@@ -991,6 +991,7 @@
 .dnd5e.sheet.actor ul.skills-list li.skill {
   height: 24px;
   padding: 3px 2px;
+  align-items: center;
 }
 .dnd5e.sheet.actor ul.skills-list li.skill:nth-child(even) {
   background: rgba(0, 0, 0, 0.05);
@@ -1019,6 +1020,7 @@
 }
 .dnd5e.sheet.actor ul.skills-list li.skill .skill-ability {
   flex: 0 0 26px;
+  border: none;
   text-transform: capitalize;
 }
 .dnd5e.sheet.actor ul.skills-list li.skill .skill-mod {
@@ -1934,16 +1936,12 @@
   margin: 0;
   top: -6px;
 }
-.dnd5e.sheet.actor.character ul.skills-list {
-  flex: 0 0 212px;
-}
 .dnd5e.sheet.actor.character ul.skills-list li.skill {
   padding: 3px;
 }
 .dnd5e.sheet.actor.character ul.skills-list li.skill h4 {
   flex: 1px;
   margin: 0;
-  font-size: 13px;
 }
 .dnd5e.sheet.actor.character .item-detail.player-class {
   flex: 0 0 180px;
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -1019,9 +1019,10 @@
   display: block;
 }
 .dnd5e.sheet.actor ul.skills-list li.skill .skill-ability {
-  flex: 0 0 26px;
+  flex: 0;
   border: none;
   text-transform: capitalize;
+  margin-right: 8px;
 }
 .dnd5e.sheet.actor ul.skills-list li.skill .skill-mod {
   flex: 0 0 20px;
--- a/less/actors.less
+++ b/less/actors.less
@@ -337,9 +337,10 @@
       }
 
       .skill-ability {
-        flex: 0 0 26px;
+        flex: 0;
         border: none;
         text-transform: capitalize;
+        margin-right: 8px;
       }
 
       .skill-mod {
--- a/less/actors.less
+++ b/less/actors.less
@@ -293,7 +293,7 @@
   /* ----------------------------------------- */
 
   ul.skills-list {
-    flex: 0 0 180px;
+    flex: 0 0 212px;
     list-style: none;
     margin: 0 5px 0;
     padding: 0;
@@ -303,6 +303,7 @@
     li.skill {
       height: 24px;
       padding: 3px 2px;
+      align-items: center;
 
       &:nth-child(even) {
         background: rgba(0, 0, 0, 0.05);
@@ -337,6 +338,7 @@
 
       .skill-ability {
         flex: 0 0 26px;
+        border: none;
         text-transform: capitalize;
       }
 
--- a/less/character.less
+++ b/less/character.less
@@ -111,14 +111,11 @@
   }
 
   ul.skills-list {
-    flex: 0 0 212px;
-
     li.skill {
       padding: 3px;
       h4 {
         flex: 1px;
         margin: 0;
-        font-size: 13px;
       }
     }
   }
--- a/module/applications/actor/npc-sheet.mjs
+++ b/module/applications/actor/npc-sheet.mjs
@@ -9,7 +9,7 @@ export default class ActorSheet5eNPC extends ActorSheet5e {
   static get defaultOptions() {
     return foundry.utils.mergeObject(super.defaultOptions, {
       classes: ["dnd5e", "sheet", "actor", "npc"],
-      width: 600
+      width: 632
     });
   }
 
--- a/module/applications/actor/skill-config.mjs
+++ b/module/applications/actor/skill-config.mjs
@@ -40,6 +40,7 @@ export default class ActorSkillConfig extends DocumentSheet {
       skill: src.system.skills?.[this._skillId] || {},
       skillId: this._skillId,
       proficiencyLevels: CONFIG.DND5E.proficiencyLevels,
+      abilities: CONFIG.DND5E.abilities,
       bonusGlobal: src.system.bonuses?.abilities.skill
     };
   }
--- a/module/dice/d20-roll.mjs
+++ b/module/dice/d20-roll.mjs
@@ -187,23 +187,18 @@ export default class D20Roll extends Roll {
    * @param {string} [data.title]             The title of the shown dialog window
    * @param {number} [data.defaultRollMode]   The roll mode that the roll mode select element should default to
    * @param {number} [data.defaultAction]     The button marked as default
-   * @param {boolean} [data.chooseModifier]   Choose which ability modifier should be applied to the roll?
-   * @param {string} [data.defaultAbility]    For tool rolls, the default ability modifier applied to the roll
    * @param {string} [data.template]          A custom path to an HTML template to use instead of the default
    * @param {object} options                  Additional Dialog customization options
    * @returns {Promise<D20Roll|null>}         A resulting D20Roll object constructed with the dialog, or null if the
    *                                          dialog was closed
    */
-  async configureDialog({title, defaultRollMode, defaultAction=D20Roll.ADV_MODE.NORMAL, chooseModifier=false,
-    defaultAbility, template}={}, options={}) {
+  async configureDialog({title, defaultRollMode, defaultAction=D20Roll.ADV_MODE.NORMAL, template}={}, options={}) {
 
     // Render the Dialog inner HTML
     const content = await renderTemplate(template ?? this.constructor.EVALUATION_TEMPLATE, {
       formula: `${this.formula} + @bonus`,
       defaultRollMode,
       rollModes: CONFIG.Dice.rollModes,
-      chooseModifier,
-      defaultAbility,
       abilities: CONFIG.DND5E.abilities
     });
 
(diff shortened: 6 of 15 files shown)


arbron (commented):
Code looks good, but at least on Firefox the dropdown chevron ends up closer to the modifier than the ability, which is a bit visually confusing:

Screen Shot 2022-10-07 at 10 42 30

Perhaps a bit of an extra margin after the `<select>` would help:

Screen Shot 2022-10-07 at 10 46 56

Varriount:
I feel like this makes the character sheet a tad too noisy - is performing the selection in the roll dialogue really not possible?

Fyorl:
> I feel like this makes the character sheet a tad too noisy - is performing the selection in the roll dialogue really not possible?

It's of course possible since that's how it works currently. But it complicates our rolling pipeline, causing us to have to essentially re-prepare the roll again once the dialog is confirmed in case the ability used in the initial preparation changed.

Varriount:
> In case the ability used in the initial preparation changed

What preparation logic is carried out before the dialog is shown?

arbron:
> What preparation logic is carried out before the dialog is shown?

The whole roll formula and all of its data are prepared before hand because it is needed in case the dialog is skipped and to display the proper initial values in the dialog itself.

arbron (approved):

arbron (commented):
It might be good to add the ability selection to the skill config dialog also.

Varriount:
So, just to make sure I understand this change correctly, what are the workflows for rolling a single skill check with a non-default ability and setting the default ability for a skill check?

Fyorl:
> So, just to make sure I understand this change correctly, what are the new workflows for:
> 
> * Rolling a single skill check with a non-default ability?
> * Setting the default ability for a skill check?

You select the ability to use for the skill on the sheet, then roll it. There's no distinction between default or non-default, it will use whichever ability is currently selected when you click to roll it.

Varriount:
Hm, then I feel like this is a case where code simplification is being done at the expense of user-friendliness:
  - A part of the roll workflow is being split off in a way that violates typical expectations of consistency. Given that there are dialogues for per-roll skill check configuration and default skill check configuration, users are typically going to expect that *all* per-roll configuration options will be in the former dialogue, and *all* default configuration options will be in the latter dialogue.  
    I've seen this kind of inconsistency enough with other pieces of software (_\*cough\* Windows \*cough\*_) to be confident in stating that this _**will**_ annoy users.
  - A single UI component (the new selector) is being used for two different tasks. It is acting both as a default ability selector and a per-roll ability selector. While it performs the former function well-enough, for the latter function the user must always remember to set the skill back to its previous value.
  - A UI component with explicitly labelled behavior (in the per-roll skill check configuration) is being swapped for a component with implicit, unlabeled behavior. I have players with varying degrees of technical intuition, and can definitely see them either being confused by this new UI component works, or simply not noticing it.

Fyorl:
> A part of the roll workflow is being split off in a way that violates typical expectations of consistency.
Given that there are dialogues for per-roll skill check configuration and default skill check configuration, users are typically going to expect that all per-roll configuration options will be in the former dialogue, and all default configuration options will be in the latter dialogue.
Can you expand on this? Which part of the roll workflow is being split off, and which two dialogs are you talking about?

Varriount:
The two dialogues I'm talking about are the "Configure [Skill Name]" and "[Skill Name] Skill Check" dialogues, respectively:
image
image

This PR removes the the "Ability Modifier" field from the "[Skill Name] Skill Check" dialogue and places it on the character sheet. On the sheet, it is meant to function as a way to set the ability modifier to use for a particular skill, for all rolls of that skill.

Before this PR, a user had to perform these steps to override the ability modifier for a single skill check roll:
  - Click the skill name.
  - In the dialogue that appears, change the "Ability Modifier" selector to the overriding ability, along with any other customizations for that single roll.
  - Click "Submit".

After this PR, the user must (for a single skill check roll):
  - Select the ability modifier from the drop-down list next to the skill,
  - Click the skill name.
  - Input any customizations for that single roll.
  - Click "Submit".
  - Select the previous ability modifier from the drop-down list next to the skill.

This PR removes the any way to change the ability modifier for _only_ a _single_ skill check roll, in favor of a way to change the ability modifier for _all_ skill check rolls (of a particular skill). Furthermore, this functionality's corresponding UI component is not located where similar components are located (the "Configure [Skill Name]" and "[Skill Name] Skill Check" dialogues).

Fyorl:
We can (and probably should) add the ability configuration to the skill configuration dialog, Jeff made a similar comment.

I recognise that this is a change to how variable ability skill rolls are performed, though I don't really agree with the implication that it's strictly less user friendly. It's just different, and it has different trade-offs. For example, having the ability selection in the intermediary roll dialog means that it's impossible to perform a fast-forward skill roll with a non-default ability. With this change, fast-forward and non-fast-forward rolls can both have a different ability selected in a consistent way.

One other big point of contention I think I have is with the framing of the intermediary roll dialog as 'the skill check roll dialog'. It's titled as such, but this dialog is actually shared across all ability checks, saving throws and attack rolls, and it's intended for generic modifications to any roll. You'll notice that skill rolls are an aberration here with the option to change something specific like the ability modifier used.

Finally, this feature to adjust the ability modifier used as part of skill checks is actually relatively new. There was no way to do this in the system before (via the UI). With the benefit of hindsight I believe this feature implemented in its current state would not have been approved in the first place (I was the one that approved it), given some of the tech debt it has incurred. So rather than removing the feature entirely, this PR is an attempt to retain it while also smoothing back out some of the wrinkles it has caused in our rolling pipeline. I'm open to any alternative proposals here that might solve this issue in a different way.

Varriount:
I'd still propose specializing the intermediary skill-check dialogue with the ability modifier to use (for that specific roll). Even with the functionality to change the default ability modifier added in the most recent commit, users are still unable to change the ability modifier in only one-off scenarios - they must always remember to change the ability modifier back to what it was previously set to.

I have real-world experience to back this assertion too. Roll20's 5e character sheet has a similar component for selecting whether rolls should be made normally, with advantage, or with disadvantage:
image
When a user needs to make a roll with advantage through their character sheet, they must click the "Advantage" toggle, then make the relevant roll. If only that single roll should have been made with advantage, they then must remember to click the "Normal" toggle after the roll.

I can personally attest that, in my weekly game with 4 players, most sessions involve at least once instance of someone forgetting to set their rolls back to normal.

In this case the potential for annoyance is even worse; it's not just a simple matter of selecting "Normal", but rather selecting the correct ability the skill was originally set to use (which may not even be the "default" ability that the 5e rules dictate).

Fyorl:
> I'd still propose specializing the intermediary skill-check dialogue with the ability modifier to use (for that specific roll).

This is off the table due to the associated technical cost and the fact that it does not address fast-forward rolls. Do you have any additional suggestions?

The only things I've been able to come up with currently is:
1. Indicate, in some way, when a non-default ability is currently selected.
2. Reset the drop-down to the default ability once a roll has been executed.

I think I'd only consider (2) if it could be implemented cleanly, and there are some additional UX considerations that come with that option.

krbz999:
> Indicate, in some way, when a non-default ability is currently selected.

DDB does so by changing the color of the skill to blue.

What if there were keybinding shortcuts for resetting the skill's ability to default? However just having option (2) seems like a decently clean solution that _mostly_ satisfies everyone.

Varriount:
So, I initially wrote up a long-winded reply explaining what my observations, areas for code improvement, etc. If you still want me to give that, I can, but after writing it all out, I figured it might be better to just get straight to the point.

The root of the problem here is that the roll customization dialogue is created rather late in the skill check process, because it's tied to the [`D20Roll`](https://github.com/foundryvtt/dnd5e/blob/f2c7c99193f5c00d327e9fb515b3759919b86e91/module/dice/d20-roll.mjs#L197) class. This then ends up complicating things because the `D20Roll` class's dialogue needs to know what an ability modifier actually *is*. 

So why not extract this logic out? From what I can see, the primary purpose of the `D20Roll` class is to contain data about a d20 dice formula and evaluate it. It shouldn't be showing the user prompts, nor (arguably) should it be manipulating or reading any other parts of the UI. The only data it should contain is data common to all d20 rolls\*.

Splitting this dialogue logic out would simplify both the D20Roll class and the [`d20roll`](https://github.com/foundryvtt/dnd5e/blob/f2c7c99193f5c00d327e9fb515b3759919b86e91/module/dice/dice.mjs#L51) function, while allowing customization of the prompt.

Just to be clear, I'm not against a default ability selector in a skill's *configuration dialogue* (which *would* impact fast-forwarded rolls); I'm against placing them directly on the character sheet, then indicating to users that the selectors are suitable for one-off rolls.

\*with some exceptions

Fyorl:
I'm trying to understand how factoring out the dialog code would solve the core issue of having to re-prepare the roll data if the ability used changes halfway through the roll process. Are you suggesting that we move the dialog to the start of the process in some way? So we immediately check the click event for whether the roll is fast-forwarded or not. If not, we display the dialog *first*, then do roll preparation after, or if the roll was fast-forwarded we skip the dialog and go straight to roll preparation?

I think it would work but it would mean that we cannot show a preview of the actual bonus values in the dialog. I think it's also quite a disruptive change to the roll workflow that I suspect will break a lot of modules. I won't discount it for that but I think it would mean that we don't do this in 2.1.

Fyorl:
> Just to make sure I'm understanding things correctly, where is the re-preparation is being done? Do you mean the logic in `Actor5e.rollSkill`, `d20roll`, and `D20Roll._onDialogSubmit` for dealing with the `@abilityCheckBonus` term?

Yes, `@abilityCheckBonus` is one such example. You can see that we resolve all bonuses in `Actor5e#rollSkill`, allowing for the roll to be made immediately in the case of a fast-forward roll. But then we have to un-set some of them (like `@mod` and `@abilityCheckBonus`) because the user might change the ability modifier being used. This is why, for all other rolls, the intermediary dialog is able to show the exact resolved bonus, but for skill rolls, you see the raw `@abilityCheckBonus` string in the preview.

We then have to splice the bonus back into the partially-resolved `Roll` in `_onDialogSubmit`, and we need to make sure we have enough information to do so. The ability check bonus is relatively simple, but consider Remarkable Athlete which has its own little package of logic that can potentially change the proficiency bonus. We don't really want to carry that logic around with us through the entire roll process, see this comment [here](https://github.com/foundryvtt/dnd5e/pull/1650#pullrequestreview-1075848262). Ideally we want to resolve all the bonuses once up-front, and then the intermediary roll dialog can be concerned only with modifications to the d20 itself.

Varriount:
Yeah, I definitely feel like this is due to misplaced responsibilities. The `Roll` class and its descendants should really only be concerned with evaluating dice rolls. If dialogue functionality really needs to be attached to them, it should probably be done through dependency injection.

A broader idea here would be to invert the control flow here - instead of having the `D20Roll` object create the dialogue, have the dialogue create the `D20Roll` object. This would allow the dialogue to update the formula it displays as the user modifies the UI.

In pseudo-code:
```
function rollSkill(...) {
    if (fastforward) {
        let roll = new SkillRoll(data, options)
    } else {
        // Could also be a regular function, `skillRollfromDialog(...)`
        let roll = await SkillRoll.fromDialog(data, options)
    }
    return roll.evaluate()
}


class SkillRoll extends D20Roll {
    dialogueFunction = (data, options) => {
        // Create the dialog, then when it closes, return the roll it constructed.
        const dialog = new SkillRollDialog(data, options)
        const roll = await dialog.render(true)
        return roll
    }

    constructor(data, options) {
        // Assemble skill roll formula and terms.
        ...
    }

    static async fromDialog(data, options) {
        await this.dialogueFunction(data, options)
    }
}


class SkillRollDialogue extends FormApplication {
    async _onChangeInput(event) {
        // Or whatever method is called when a from field is changed.
        // Update the roll, then re-render the dialogue to display the new roll.
        this.roll = new SkillRoll(this.data, this.options)
        await this.render(true)
    }
}
```

Fyorl:
Yeah, bringing the dialog forward in the pipeline would work. Even if we do that though, the experience for fast-forward rolls isn't as good: You have to open the skill dialog, change the ability score, close the dialog, then make the roll. It erodes any of the time-saving you might have wanted from fast-forwarding the roll in the first place.

So while I understand the concern that the proposal in this PR might make things worse for non-fast-forward rolls, I'm not sure that overall it's a strictly worse design. Ideally we'd have a solution that has the best of both worlds.

Either way, I think this has to be parked for now as it has been taking too long and we have other priorities to get to in 2.1.

Varriount:
I guess, I don't quite see why this needs to be implemented for fast-forward rolls. To me, the whole point of the fast-forward process is to say, "I'm ok with the defaults, and have no need to customize them for this roll".

p4535992:
> 

100% agree and I think it should be the logic in general when it comes to fast forward roll to use the default preset settings.

Possibly external modules such as https://github.com/misthero/dnd5e-custom-skills can allow you to prepare multiple versions of the same skill and "customize" multiple fast forward rolls.

Issue #1806, opened 2022-09-11
Positive deltas are not accepted in many character sheet fields


benbarbour:
**Foundry Version:** 10.285
**DnD5e Version:** 2.0.2
**Modules:** None

If I enter "+X" in a currency box on the inventory panel of the Actor sheet then that currency is set to X. I expected X to be added to the current value. If I enter "-X" instead then X is subtracted from the current value. Using no prefix sets the value to whatever I entered, as expected.

*UPDATE*: This affects more than just the currency fields (see below).

unsoluble:
Note this is also the case for other places where deltas should be accepted, like XP and HP.

kaelad02:
I remember something about some `<input>` fields changing from input type string to number. Could that be it? Could the fields now only be accepting numbers, so when they're submitted they're changed from `+10` as a string to just `10` the number?

LiamFallon213:
So I did some testing. What I found, all testing done on the currency fields.

- This bug only happens when you add a positive delta to a non zero field.

- If you manually go into the HTML of foundry and change the input of the file to text instead of number. The bug is not present.

- If you manually set the currency as text before run, neither deltas work.

- The code for handling delta's simply is not running when the bug is present.

- The code for activating listeners, (around line 600 in base-sheet.mjs) does not run when bug is present.

This is where my investigation into the bug currently is stuck at, I can't follow the data up past the listener activation failure. I will provide more info when I discover it.

arbron:
Looks like this is only an issue in Chrome, it works fine in Firefox. It seems like Chrome is removing the leading plus from the value before we get a chance to handle it (e.g. `"+15"` turns into `"15"`) while Firefox leaves it alone.

LiamFallon213:
> Looks like this is only an issue in Chrome, it works fine in Firefox. It seems like Chrome is removing the leading plus from the value before we get a chance to handle it (e.g. `"+15"` turns into `"15"`) while Firefox leaves it alone.

I can confirm this, I was able to get a proof of concept working where I changed the input type to text. Then changed the delta function to handle this by parsing to float in the addition line, ~line 700 base-sheet.mjs.
`input.value = parseFloat(foundry.utils.getProperty(this.actor, input.name)) + delta;`

Notably this breaks a bunch of other things, so I can't just do this then push a fix. Specifically it's breaking the CSS right now, which I'm working on.

LiamFallon213:
I am about 90% done fixing this. Will get the code up soon.

PR #1860, opened 2022-10-11
[#1806] Fix number deltas not being accepted on actor sheet

LiamFallon213:
This is a bug fix to bug #1806.

Changed most of the numericInput fields on the character sheet to text input.
Added regex based input validation to accommodate this into the _onChangeInputDelta function on base-sheet.mjs.
Added parsing to the property return in _onChangeInputDelta to make JS preform addition instead of concatenation.
Added a new HTML class, "delta-input", which is now what jquery looks for to determine if some input field is one that can have deltas entered.

What I discovered with some help from the discord was that certain browsers would cut off the + from an input like "+100". I made the inputs text inputs to fix this. To accommodate this I added input validation and parsing.


templates/actors/character-sheet.hbs
```diff
@@ -54,23 +54,23 @@
...
                         <input name="system.attributes.hp.temp" type="number" step="any" class="temphp"
                             placeholder="+{{ localize 'DND5E.Temp' }}"
                             value="{{system.attributes.hp.temp}}"
-                            title="{{localize 'DND5E.HitPointsTemp'}}">
+                            title="{{localize 'DND5E.HitPointsTemp'}}" class="delta-input">
                         <input name="system.attributes.hp.tempmax" type="number" step="any" class="temphp"
                             placeholder="+{{ localize 'DND5E.Max' }}"
                             value="{{system.attributes.hp.tempmax}}"
-                            title="{{localize 'DND5E.HitPointsTempMax'}}">
+                            title="{{localize 'DND5E.HitPointsTempMax'}}" class="delta-input">
```
arbron, line 73:
```suggestion
                            title="{{localize 'DND5E.HitPointsTempMax'}}" data-dtype="Number">
```

templates/actors/character-sheet.hbs
```diff
@@ -54,23 +54,23 @@
...
-                        <input name="system.attributes.hp.value" type="number" step="any"
+                        <input name="system.attributes.hp.value" type="text" step="any"
                             value="{{system.attributes.hp.value}}" placeholder="10"
-                            title="{{localize 'DND5E.HitPointsCurrent'}}">
+                            title="{{localize 'DND5E.HitPointsCurrent'}}" class="delta-input">
                         <span class="sep"> / </span>
                         <input name="system.attributes.hp.max" type="number" step="any"
                             value="{{system.attributes.hp.max}}" placeholder="10"
-                            title="{{localize 'DND5E.HitPointsMax'}}">
+                            title="{{localize 'DND5E.HitPointsMax'}}" class="delta-input">
```
arbron, line 63:
```suggestion
                            title="{{localize 'DND5E.HitPointsMax'}}" data-dtype="Number">
```

templates/actors/character-sheet.hbs
```diff
@@ -54,23 +54,23 @@
                 <li class="attribute health">
                     <h4 class="attribute-name box-title">{{ localize "DND5E.HitPoints" }}</h4>
                     <div class="attribute-value multiple">
-                        <input name="system.attributes.hp.value" type="number" step="any"
+                        <input name="system.attributes.hp.value" type="text" step="any"
                             value="{{system.attributes.hp.value}}" placeholder="10"
-                            title="{{localize 'DND5E.HitPointsCurrent'}}">
+                            title="{{localize 'DND5E.HitPointsCurrent'}}" class="delta-input">
                         <span class="sep"> / </span>
                         <input name="system.attributes.hp.max" type="number" step="any"
```
arbron, line 61:
```suggestion
                        <input name="system.attributes.hp.max" type="text"
```

templates/actors/character-sheet.hbs
```diff
@@ -54,23 +54,23 @@
                 <li class="attribute health">
                     <h4 class="attribute-name box-title">{{ localize "DND5E.HitPoints" }}</h4>
                     <div class="attribute-value multiple">
-                        <input name="system.attributes.hp.value" type="number" step="any"
+                        <input name="system.attributes.hp.value" type="text" step="any"
```
arbron, line 57:
```suggestion
                        <input name="system.attributes.hp.value" type="text"
```

templates/actors/character-sheet.hbs
```diff
@@ -54,23 +54,23 @@
...
                             value="{{system.attributes.hp.value}}" placeholder="10"
-                            title="{{localize 'DND5E.HitPointsCurrent'}}">
+                            title="{{localize 'DND5E.HitPointsCurrent'}}" class="delta-input">
```
arbron, line 59:
```suggestion
                            title="{{localize 'DND5E.HitPointsCurrent'}}" data-dtype="Number">
```

templates/actors/character-sheet.hbs
```diff
@@ -153,8 +153,8 @@
             {{#each system.abilities as |ability id|}}
                 <li class="ability {{#if ability.proficient}}proficient{{/if}}" data-ability="{{id}}">
                     <h4 class="ability-name box-title rollable">{{ability.label}}</h4>
-                    {{numberInput ability.value name=(concat "system.abilities." id ".value")
-                        class="ability-score" placeholder="10"}}
+                    <input type="text" value="{{ability.value}}" name="{{concat "system.abilities." id ".value"}}"
+                        class="ability-score delta-input" placeholder="10">
```
arbron, line 157:
```suggestion
                    <input type="text" name="system.abilities.{{id}}.value" value="{{ability.value}}"
                        class="ability-score" placeholder="10" data-dtype="Number">
```

templates/actors/character-sheet.hbs
```diff
@@ -203,12 +203,9 @@
                             <label class="recharge checkbox flexcol">
                                 <span>{{ localize "DND5E.AbbreviationSR" }}</span><input name="system.resources.{{res.name}}.sr" type="checkbox" {{checked res.sr}}/>
                             </label>
-
-                            {{numberInput res.value name=(concat "system.resources." res.name ".value")
-                                placeholder="0"}}
+                            <input type="text" value="{{res.value}}" name="{{concat "system.resources." res.name ".value"}}" placeholder="0" class="delta-input">
```
arbron, line 206:
```suggestion
                            <input type="text" name="system.resources.{{res.name}}.value" value="{{res.value}}" placeholder="0" data-dtype="Number">
```

templates/actors/character-sheet.hbs
```diff
@@ -203,12 +203,9 @@
...
                             <span class="sep"> / </span>
-                            {{numberInput res.max name=(concat "system.resources." res.name ".max") placeholder="0"}}
-
+                            <input type="text" value="{{res.max}}" name="{{concat "system.resources." res.name ".max"}}" placeholder="0" class="delta-input">
```
arbron, line 208:
```suggestion
                            <input type="text" name="system.resources.{{res.name}}.max" value="{{res.max}}" placeholder="0" data-dtype="Number">
```

templates/actors/character-sheet.hbs
```diff
@@ -223,18 +220,15 @@
                         <h4 class="rollable" data-action="rollDeathSave">{{ localize "DND5E.DeathSave" }}</h4>
                         <div class="counter-value">
                             <i class="fas fa-check"></i>
-                            {{numberInput system.attributes.death.success name="system.attributes.death.success"
-                                placeholder="0"}}
+                            <input type="text" value="{{system.attributes.death.success}}" name="system.attributes.death.success" class="delta-input" placeholder="0">
```
arbron, line 223: I'm not sure delta inputs are really necessary for death saves, because generally that is handled by the core roller. They weren't enabled here before the recent changes anyway.

templates/actors/character-sheet.hbs
```diff
@@ -223,18 +220,15 @@
...
-                                placeholder="0"}}
+                            <input type="text" value="{{system.attributes.death.failure}}" name="system.attributes.death.failure" class="delta-input" placeholder="0">
                         </div>
                     </div>
                     <div class="counter flexrow exhaustion">
                         <h4>{{ localize "DND5E.Exhaustion" }}</h4>
                         <div class="counter-value">
-                            {{numberInput system.attributes.exhaustion name="system.attributes.exhaustion"
-                                placeholder="0"}}
+                            <input type="text" value="{{system.attributes.exhaustion}}" name="system.attributes.exhaustion" class="delta-input" placeholder="0">
```
arbron, line 231:
```suggestion
                            <input type="text" name="system.attributes.exhaustion" value="{{system.attributes.exhaustion}}" placeholder="0" data-dtype="Number">
```

templates/actors/parts/actor-inventory.hbs
```diff
@@ -9,7 +9,7 @@
         </h3>
         {{#each system.currency as |v k|}}
         <label class="denomination {{k}}">{{ lookup ../labels.currencies k }}</label>
-        {{numberInput v name=(concat "system.currency." k)}}
+       <input type="text" value="{{v}}" step="any" name="{{concat "system.currency." k}}" class="delta-input" >
```
arbron, line 12:
```suggestion
       <input type="text" name="system.currency.{{k}}" value="{{v}}" data-dtype="Number"> 
```

templates/actors/parts/actor-spellbook.hbs
```diff
@@ -4,8 +4,7 @@
         <label>{{localize "DND5E.Spellcasting"}}</label>
         {{else}}
         <label>{{localize "DND5E.Level"}}</label>
-        {{numberInput system.details.spellLevel name="system.details.spellLevel" class="spellcasting-level"
-            placeholder="0"}}
+        <input type="text" value = "{{system.details.spellLevel name="system.details.spellLevel"}}" class="spellcasting-level delta-input" placeholder="0">
```
arbron, line 7:
```suggestion
        <input type="text" name="system.details.spellLevel" value="{{system.details.spellLevel}}" class="spellcasting-level" placeholder="0" data-dtype="Number">
```

templates/actors/parts/actor-spellbook.hbs
```diff
@@ -35,7 +34,7 @@
             <h3>{{section.label}}</h3>
             <div class="spell-slots">
                 {{#if section.usesSlots}}
-                {{numberInput section.uses name=(concat "system.spells." section.prop ".value") placeholder="0"}}
+                <input type="text" value="{{section.uses}}"  name="{{concat "system.spells." section.prop ".value"}}" placeholder="0" class = "delta-input">
```
arbron, line 37:
```suggestion
                <input type="text" name="system.spells.{{section.prop}}.value" value="{{section.uses}}" placeholder="0" data-dtype="Number"> 
```

templates/actors/character-sheet.hbs
LiamFallon213: I will do this after my uni homework tonight.

dnd5e.css
```diff
@@ -1127,7 +1127,7 @@
   line-height: 20px;
   color: #7a7971;
 }
-.dnd5e.sheet.actor .inventory-filters .currency input[type="number"] {
+.dnd5e.sheet.actor .inventory-filters .currency input[type="text"] {
```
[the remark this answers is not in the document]
LiamFallon213: I will get this working after I finish my uni homework tonight.

templates/actors/character-sheet.hbs
LiamFallon213: I mean, that's true but I guess I don't see why we wouldn't do this unless it breaks something else. I tried to more or less make all inputs on the sheet that use only numbers compatible.

arbron: Its fine by me :)

templates/actors/character-sheet.hbs
```diff
@@ -223,18 +220,15 @@
                         <h4 class="rollable" data-action="rollDeathSave">{{ localize "DND5E.DeathSave" }}</h4>
                         <div class="counter-value">
                             <i class="fas fa-check"></i>
-                            {{numberInput system.attributes.death.success name="system.attributes.death.success"
-                                placeholder="0"}}
+                            <input type="text" value="{{system.attributes.death.success}}" name="system.attributes.death.success" class="delta-input" placeholder="0">
                             <i class="fas fa-times"></i>
-                            {{numberInput system.attributes.death.failure name="system.attributes.death.failure"
-                                placeholder="0"}}
+                            <input type="text" value="{{system.attributes.death.failure}}" name="system.attributes.death.failure" class="delta-input" placeholder="0">
```
arbron, line 225:
```suggestion
                            <input type="text" name="system.attributes.death.failure" value="{{system.attributes.death.failure}}" placeholder="0" data-dtype="Number">
```

templates/actors/parts/actor-inventory.hbs
```diff
@@ -9,7 +9,7 @@
         </h3>
         {{#each system.currency as |v k|}}
         <label class="denomination {{k}}">{{ lookup ../labels.currencies k }}</label>
-        {{numberInput v name=(concat "system.currency." k)}}
+       <input type="text" name="system.currency.{{k}}" value="{{v}}" data-dtype="Number">
```
arbron, line 12:
```suggestion
        <input type="text" name="system.currency.{{k}}" value="{{v}}" data-dtype="Number"> 
```

module/applications/actor/base-sheet.mjs
```diff
@@ -695,7 +693,7 @@ export default class ActorSheet5e extends ActorSheet {
     const value = input.value;
     if ( ["+", "-"].includes(value[0]) ) {
       let delta = parseFloat(value);
-      input.value = foundry.utils.getProperty(this.actor, input.name) + delta;
+      input.value = parseFloat(foundry.utils.getProperty(this.actor, input.name)) + delta;
```
Fyorl, line 696: I think we want to use `Number` here as `Number(null)` is `0`. If we use `parseFloat` for the same we get `NaN`.
```suggestion
      input.value = Number(foundry.utils.getProperty(this.actor, input.name)) + delta;
```

templates/actors/parts/actor-spellbook.hbs
```diff
@@ -4,8 +4,7 @@
         <label>{{localize "DND5E.Spellcasting"}}</label>
         {{else}}
         <label>{{localize "DND5E.Level"}}</label>
-        {{numberInput system.details.spellLevel name="system.details.spellLevel" class="spellcasting-level"
-            placeholder="0"}}
+        <input type="text" name="system.details.spellLevel" value="{{system.details.spellLevel}}" class="spellcasting-level" placeholder="0" data-dtype="Number">
```
Fyorl, line 7: Not sure I agree with this one. Is it really the case that we need to apply deltas to an NPC spellcaster's spellcasting level? I would expect you only ever set it once to a particular value and then leave it alone.

Fyorl (changes requested):
Thanks for doing this work, it will be a good fix to get in this release. As Jeff mentioned, the npc and vehicle sheets will probably need the same treatment to identify areas where we want to allow for deltas.

dnd5e.css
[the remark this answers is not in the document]
LiamFallon213: ok have some time, gonna try this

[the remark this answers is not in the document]
LiamFallon213: I have fixed this with the commit on 10/25/22

LiamFallon213:
> 



> 

Perhaps, I tried to do some of those, but I will look at their files and run some updates

templates/actors/parts/actor-spellbook.hbs
Fyorl: We should use the most appropriate input field for the value it is holding. In many cases a `number` input (and our `numberInput` helper) has the advantage that it can catch invalid input early before the form is even submitted. So we should be using `numberInput` with appropriate `min`, `max`, and `step`, for all strictly numeric values except in cases where we want to relax these constraints to allow for deltas. My current intuition is that we only need deltas for HP, XP, and currency. I don't think there are really many other places where the numbers are large enough that deltas are particularly useful.

module/applications/actor/base-sheet.mjs
```diff
@@ -686,18 +684,27 @@ export default class ActorSheet5e extends ActorSheet {
...
+    if(value.match(/^([\d+-]\d*$)/)){
+      if ( ["+", "-"].includes(value[0]) ) {
+        let delta = parseFloat(value);
+        input.value = Number(foundry.utils.getProperty(this.actor, input.name)) + delta;
+      } else if ( value[0] === "=" ) input.value = value.slice(1);
+    }else if(value.match(/^([\d+-]\d*[+-/*]?[-]?\d*$)/)){
+      if(value.length <= 10){ 
+        input.value = Math.round(eval(value));
+      }
+    }else{
```
Fyorl, line 705: It looks like this is checking for something like a formula and evaluating it? What is the use-case for this? Also, your PR is targeted at fixing a bug, and this seems to be additional behaviour that was not previously in the system. If you want to add functionality, please raise an issue separately with the intended use-case and submit a separate PR for it.

LiamFallon213: Good idea

LiamFallon213: Part of why this is part of the same thing, is that this requires the logic I built in the original PR.

LiamFallon213: I have removed that section and placed it in new PR, #1897 .

templates/actors/parts/actor-spellbook.hbs
LiamFallon213: You make a good point here.

LiamFallon213: Updated to resolve this.

LiamFallon213:
@Fyorl I updated this to fix actor-spellbook.hbs as per your suggestion.

Fyorl:
> @Fyorl I updated this to fix actor-spellbook.hbs as per your suggestion.

I can see there are still many places that have been converted to `text` input fields where it's very debateable how useful having deltas would be. For example, vehicle mishap threshold, damage threshold, and action thresholds I am almost certain would never need delta support.

Character resources are probably fine, but I don't think ability scores should have deltas even if they might be occasionally useful. Really our metric for whether a field should have delta support should be 'is this field almost exclusively updated via deltas?'. For HP, XP, and currency, I believe that's the case, but for most other fields I am doubtful.

.gitignore
```diff
@@ -12,3 +12,5 @@ node_modules/
 
 # OS Specific
 .DS_Store
+dnd5e.code-workspace
+.vscode/launch.json
```
Fyorl, line 16: These look like they should be under the `# IDE` heading.

module/applications/actor/base-sheet.mjs
```diff
@@ -686,18 +684,23 @@ export default class ActorSheet5e extends ActorSheet {
   /* -------------------------------------------- */
 
   /**
-   * Handle input changes to numeric form fields, allowing them to accept delta-typed inputs
+   * Handle input changes to numeric form fields, allowing them to accept delta-typed inputs, uses REGEX to determine if input is valid or not.
    * @param {Event} event  Triggering event.
    * @private
    */
+  // Uses strict regex validation to prevent malicious injection
+  // Has a limit on size to further prevent malicious injection, it's going to be hard to do malicious code in  10 characters.
```
Fyorl, line 692: Malicious injection should not be an issue as we are not `eval`ing anything.

module/applications/actor/base-sheet.mjs
```diff
@@ -686,18 +684,23 @@ export default class ActorSheet5e extends ActorSheet {
   /* -------------------------------------------- */
 
   /**
-   * Handle input changes to numeric form fields, allowing them to accept delta-typed inputs
+   * Handle input changes to numeric form fields, allowing them to accept delta-typed inputs, uses REGEX to determine if input is valid or not.
```
Fyorl, line 687: Probably not necessary to describe the regex implementation detail here.

module/applications/actor/base-sheet.mjs
```diff
@@ -686,18 +684,23 @@ export default class ActorSheet5e extends ActorSheet {
...
-    if ( ["+", "-"].includes(value[0]) ) {
-      let delta = parseFloat(value);
-      input.value = foundry.utils.getProperty(this.actor, input.name) + delta;
+    if(value.match(/^([\d+-]\d*$)/)){
+      if ( ["+", "-"].includes(value[0]) ) {
+        let delta = parseFloat(value);
+        input.value = Number(foundry.utils.getProperty(this.actor, input.name)) + delta;
+      } else if ( value[0] === "=" ) input.value = value.slice(1);
+    }else{
+      input.value = Number(foundry.utils.getProperty(this.actor, input.name));
```
Fyorl, line 702:
```suggestion
    if ( /^([\d+-]\d*$)/.test(value) ){
      if ( ["+", "-"].includes(value[0]) ) {
        const delta = parseFloat(value);
        input.value = Number(foundry.utils.getProperty(this.actor, input.name)) + delta;
      } else if ( value[0] === "=" ) input.value = value.slice(1);
    } else {
      input.value = Number(foundry.utils.getProperty(this.actor, input.name));
```

Fyorl (changes requested):

module/applications/actor/base-sheet.mjs
LiamFallon213: Fair

LiamFallon213:
> 
This is fair, since you are more experienced with this codebase feel free to change some. I really missed this on currency and HP, so I can see why we might want to not have it for many things. 

However I would say that ability scores are often effected by delta's. Examples being monsters that change scores, e.g. intelligence devourer, and items like books. While the math for these are not difficult, I do see this as a good source of utility.

Fyorl:
> However I would say that ability scores are often effected by delta's. Examples being monsters that change scores, e.g. intelligence devourer, and items like books. While the math for these are not difficult, I do see this as a good source of utility.

I would probably disagree with 'often'. You have to adjust HP, XP, and currency pretty much every encounter, but you don't fight Intellect Devourers or Shadows, or find Very Rare magical items every encounter by a wide margin. I can definitely understand the argument for delta-enabling ability scores though, so if you feel strongly, we can keep them in. Ultimately, the cost of having a delta-enabled field is not particularly high.

LiamFallon213:
As far as I'm concerned, this is done.

Issue #1761, opened 2022-09-01
Please return the ability to add text to a currency field


Elmaro:
# Suggested Changes
In the version of 5e that worked with v9 of foundry, on an item, you could write "5 gp" or "27 ep" etc as a price for the item. You just needed to add a currency suffix next to the numbers, and viola.

But upon the release of the latest version, that feature has been removed, and you can no longer type letters into the "price" field, thus only way to distinguish between currencies is via decimal places.

# Reason for Changes
The 5th edition of D&D uses CP, SP, EP, GP and PP - and being able to go to an item and write how much it costs in whole numbers was such an unbelievably useful thing to do. 

In 5e Game Modules such as Waterdeep Dragon Heist for example, the book gives you new specific currencies to use, the Toal, the Harbor Moon and the Sambar. Due to the changes in the foundry 5e version it now isn't possible to signify these currencies at all.

In Curse of Strahd and the other Domains of Death from, Van Richten's Guide Silver is supposed to be a rarer commodity, and prior to the changes in the newest version, it was very straightforward for a DM to change items so they use a different currency denomination, but are still of the same value. Especially if currency needs to be exchanged, and does not instantly transmogrify into another denomination when it is spent or gained.

# Other Uses
In conjunction with foundry modules that provide the ability to build a shop in game that the players can interact with, the removal of the ability to define what kind of currency an item us using causes all of the items to have a wide array of different decimal places. It is no so difficult when looking at an item to determine the actual value in real coins - "0.23" doesn't look like a price to me

Things get especially confusing if a DM were to use the silver standard, which helps make electrum a more viable currency on its own

# Afterward
On a personal note, I have dyscalculia which really impacts my ability to do or read anything involving maths, decimals included - and the ability to really quickly change an item's price by writing "gp, ep, cp" etc - next to it was such an unbelievably useful feature

I am concerned that this might be seen as a niche issue, but I really do not think it is, I think having the ability to signify different coins that an item uses by typing the suffix next to a whole number is exceedingly useful

(An example of items with the currency denomination displayed)
![currencything](https://user-images.githubusercontent.com/29361562/188008722-65a34851-5772-4c31-8d32-f2f05903d1a5.PNG)

PR #1862, opened 2022-10-11
[#1761] Allow text in currency field

LiamFallon213:
This introduces the functionality requested in #1761.  Specifically it allows text in the price field.

### DIFF
--- a/templates/items/parts/item-description.hbs
+++ b/templates/items/parts/item-description.hbs
@@ -14,7 +14,7 @@
 
         <div class="form-group">
             <label>{{ localize "DND5E.Price" }}</label>
-            {{numberInput system.price name="system.price"}}
+            <input type="text" value="{{system.price}}" name="system.price">
         </div>
         {{/if}}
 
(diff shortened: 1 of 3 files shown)


templates/items/parts/item-description.hbs
```diff
@@ -14,7 +14,7 @@
 
         <div class="form-group">
             <label>{{ localize "DND5E.Price" }}</label>
-            {{numberInput system.price name="system.price"}}
+            <input type="text" value="{{system.price}}" name="system.price">
```
Varriount, line 17: Just curious - what happens when an invalid value is given?

arbron:
I'm not sold that allowing text in this input is the right option because it would require any modules that make use of this data to do text parsing and hope users entered their information in something of a standard format. I believe a better option would be adding a section property for `currency` with a dropdown to allow for changing between the defined currency types while still keeping the value as a number.

```javascript
price: {
  value: 15,
  currency: "gp"
}
```

This will require a more complicated migration, but it should be easy to handle with the new data models.

arbron:
I'm closing this in favor of the more structured alternative that is getting added to 2.1: #1908

PR #1870, opened 2022-10-12
Add Devil's Sight as a sense in creature data and on sheets

### LINKED ISSUE #1869: Devil's Sight should be among the structured creature senses
Currently the Devil's Sight invocation has an active effect that increases darkvision to 120 feet. Were it to be present as a distinct sense, the AE and any future integration with vision modes could represent the mechanic more precisely.

stwlam:
Closes #1869 

Notes:
- The current senses are all single words, so there was no casing precedent. I went with camel case but am happy to change it to anything else.
- I notice the compendium build results are committed. Should a rebuild be done with this PR or in a followup?

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -719,6 +719,7 @@
 "DND5E.SensesConfig": "Configure Senses",
 "DND5E.SensesConfigHint": "Configure any special sensory perception abilities that this actor possesses.",
 "DND5E.SenseDarkvision": "Darkvision",
+"DND5E.SenseDevilsSight": "Devil's Sight",
 "DND5E.SenseBlindsight": "Blindsight",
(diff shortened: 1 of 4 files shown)


Fyorl:
This way is probably more flexible and homebrew-friendly but arguably Devil's Sight is more of a special trait than its own type of sense. It functions the same way across all official material, and is not part of the list of 'senses' in the MM or PHB. We can come back to this when we get around to implementing dnd5e-specific vision modes (hopefully in this release).

> I notice the compendium build results are committed. Should a rebuild be done with this PR or in a followup?

Usually we include the `packs/src/` changes and the rebuilt compendium `.db` files in the PR itself.

Issue #1835, opened 2022-09-30
Create new system data models for Actors and Items


arbron:
- [x] Create Actor system data models
- [x] Create Item system data models
- [x] Move migrations into data models where possible
- [x] Refresh all SRD compendiums using new data models

PR #1872, opened 2022-10-14
[#1835] Add `loot` data model

arbron:
Adds the data model for Loot items, which includes the Item Description template and the Physical Item template.

### DIFF
--- a/module/data/abstract.mjs
+++ b/module/data/abstract.mjs
@@ -3,6 +3,17 @@
  *
  * **Note**: This uses some advanced Javascript techniques that are not necessary for most data models.
  * Please refer to the [advancement data models]{@link BaseAdvancement} for an example of a more typical usage.
+ *
+ * In template.json, each Actor or Item type can incorporate several templates which are chunks of data that are
+ * common across all the types that use them. One way to represent them in the schema for a given Document type is to
+ * duplicate schema definitions for the templates and write them directly into the Data Model for the Document type.
+ * This works fine for small templates or systems that do not need many Document types but for more complex systems
+ * this boilerplate can become prohibitive.
+ *
+ * Here we have opted to instead create a separate Data Model for each template available. These define their own
+ * schemas which are then mixed-in to the final schema for the Document type's Data Model. A Document type Data Model
+ * can define its own schema unique to it, and then add templates in direct correspondence to those in template.json via
+ * SystemDataModel.mixin.
  */
 export default class SystemDataModel extends foundry.abstract.DataModel {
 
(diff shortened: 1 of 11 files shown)


module/data/abstract.mjs
```diff
@@ -0,0 +1,79 @@
...
+ * **Note**: This uses some advanced Javascript techniques that are not necessary for most data models.
+ * Please refer to the [advancement data models]{@link BaseAdvancement} for an example of a more typical usage.
+ *
+ */
+export default class SystemDataModel extends foundry.abstract.DataModel {
+
+  /**
+   * Base templates used for construction.
+   * @type {*[]}
+   * @protected
```
Fyorl, line 13: We should consider `@internal` here. I don't think subclasses should really be messing with this property, it's exclusively part of the mixin mechanism.

module/data/item/templates/physical-item.mjs
```diff
@@ -0,0 +1,66 @@
...
+ * @property {boolean} equipped    Is this item equipped on its owning actor.
+ * @property {string} rarity       Item rarity as defined in `DND5E.itemRarity`.
+ * @property {boolean} identified  Has this item been identified?
+ * @mixin
+ */
+export default class PhysicalItemTemplate extends foundry.abstract.DataModel {
+  static defineSchema() {
+    return {
+      quantity: new foundry.data.fields.NumberField({
+        required: true, nullable: false, integer: true, positive: true, initial: 1, label: "DND5E.Quantity"
```
Fyorl, line 17: `positive: true` here is perhaps too strict. I know there are some users who like to keep things at `0` quantity so they don't forget to restock them and can do so more easily.

module/data/item/templates/physical-item.mjs
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Migrate the item's rarity from freeform string to enum value.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static migrateRarity(source) {
+    if ( !source.rarity ) return;
```
Fyorl, line 60:
```suggestion
    if ( !source.rarity || CONFIG.DND5E.itemRarity[key] ) return;
```
And then we can remove the or-branch further down.

Fyorl (changes requested):
Nice work, Jeff. Some very minor change requests.

module/data/abstract.mjs
arbron: `@internal` isn't a JSDoc property. We currently don't have TSDoc definitions, so that throws a linting error.

Fyorl (approved):
This is obviously a very well thought-out design, so thank you for putting it together, Jeff. I'm just going to append a few comments to cover all our bases, but then I think we can merge it and move onto the other data models.

PR #1876, opened 2022-10-14
[#1258, #1858, #1867] Conditionally show/hide item activation fields

### LINKED ISSUE #1258: Spell Duration options are missing "Until Dispelled" and "Until Dispelled or Triggered"
Originally in GitLab by @DrYoshiyahu

There are a dozen spells in the game that have a duration of "Until Dispelled" or "Until Dispelled or Triggered." These spells in the SRD compendium use "Permanent" instead, which is perhaps some form of compromise, even if strictly inaccurate.

If that is the case, and "Permanent" is supposed to replace the two of them, then might I suggest at least giving the option to input plain text as a duration, rather than users relying on the dropdown menu, for the duration purists out there.

### LINKED ISSUE #1858: Conditionally show fields on item sheet
There are number of fields on the item sheet that are not necessary depending on what is selected, for example the units associated with target is only needed when an area target is selected, and should be hidden if an individual target type is selected (`"2 Feet Enemies"` doesn't make sense, `"30 Feet Cone"` does).

### LINKED ISSUE #1867: Expand the Limited Uses Maximum field for better visibility
The Maximum Uses field is too small to view attribute references, if used
image

Messing around in the Elements (I know nothing of css) changing `.dnd5e.sheet.item .details .form-group.uses-per input` to `{flex: 1;}` is a bit better 
image

but it may be more desirable to differentiate the `text` Max Uses field from the `number` Available Uses and Resource Consumption fields also controlled by `.dnd5e.sheet.item .details .form-group.uses-per input` so that just this field can be expanded, leaving those a bit smaller
image

arbron:
Hides any fields that are not valid based on the currently select units for item's `"Target"`, `"Range"`, `"Duration"`, and `"Resource Consumption"`. So if your target is `"Enemy"`, it will only display the numeric field and not the distance units (cause 5 Feet Enemy as a target makes no sense).

Conditional Fields - Empty

Conditional Fields - Single

Conditional Fields - Multi

This also expands the size of the limited uses input fields to allow more of the max uses formula to be visible.

Based off the work of @zeel01 from over on GitLab.

Resolves #1258
Resolves #1858 
Resolves #1867


Varriount:
Yes! This is a great improvement - it's definitely something that has confused both me and my players.

Fyorl (changes requested):
Felt very good to use in my testing, nice work. Let's take this opportunity to replace any of the `<input type="number">` fields that were edited here with the `{{numberInput}}` helper. This might also be a good opportunity to add the missing 'Until Dispelled' duration.

arbron:
> Felt very good to use in my testing, nice work. Let's take this opportunity to replace any of the `<input type="number">` fields that were edited here with the `{{numberInput}}` helper.

I think we've been staying away from that in a few places because it doesn't support adding tooltips.

> This might also be a good opportunity to add the missing 'Until Dispelled' duration.

Good idea, I'll stick that in.

Fyorl:
> I think we've been staying away from that in a few places because it doesn't support adding tooltips.

Oh, ugh, ok. I'll do another pass then and make sure our `step` values are all correct then at least.

Fyorl (approved):

PR #1879, opened 2022-10-17
Add support for per-ability-score global DC bonuses

TheGiddyLimit:
This supports the following use-case:

A GM has a creature with spellcasting using one ability (say int), and a feature which has a saving throw DC based on another ability (say cha). They want to scale the creature's saving throw DCs on-the-fly using active effects.

Currently, modifying DCs with active effects is only possible for spells. This change allows other DCs linked to an ability score to be easily modified.


module/documents/actor/actor.mjs
```diff
@@ -365,7 +365,8 @@ export default class Actor5e extends Actor {
 
       abl.save = abl.mod + abl.saveBonus;
       if ( Number.isNumeric(abl.saveProf.term) ) abl.save += abl.saveProf.flat;
-      abl.dc = 8 + abl.mod + this.system.attributes.prof + dcBonus;
+      abl.spellDc = 8 + abl.mod + this.system.attributes.prof + spellDcBonus;
```
TheGiddyLimit, line 368: Probably breaking change: renamed this to better represent what it does
Thoughts?

Varriount: While I'll leave it to the project maintainers to approve/reject this, keep in mind that a change like this requires:
  - [Migrating}(https://github.com/foundryvtt/dnd5e/blob/f2c7c99193f5c00d327e9fb515b3759919b86e91/module/migration.mjs) existing actor data.
  - Changing the data model in [template.json](https://github.com/foundryvtt/dnd5e/blob/master/template.json).
  - Renaming any other references to this attribute.

Fyorl:
I just want to clarify so I understand correctly. Presently, the value in 'global spell DC bonus' is added to the `dc` value for *all* abilities, even if they're not the character's 'spellcasting ability', is that right? And what's desired here is a way to modify the `dc` value of a specific ability, without also necessarily modifying the `dc` of all other abilities?

TheGiddyLimit:
@Fyorl ah, then this PR is probably redundant for the use-case I posted (which was not my own -- @CrackJackFlood , your thoughts?)
A way to modify each DC independently may still have its uses, but equally may be more annoying than useful 🤔

Fyorl:
Well, I do wonder if we should be making some changes here regardless. It seems a little unintuitive to me that if you want to modify your Battle Master Manoeuvre DC that you need to add a value to the 'global spell DC bonus' field. Maybe that should be renamed to just 'global DC bonus', or should be modified to actually only affect spell save DCs. Probably a bit of a tangential issue to this PR though.

TheGiddyLimit:
True. Equipping a Robe of the Archmagi shouldn't change my Trip Attack DC, so a rename doesn't quite cut it. I inadvertently made the change to "have spell DC bonus only affect spells" here; I'm happy to split that into another PR if so desired. I'll give my "client" here a chance to weigh in first, though!

CrackJackFlood:
@TheGiddyLimit Thanks again for doing this.

@Fyorl Hello, I am the guy who mentioned this functionality to Giddy and he put together this PR.

To start off I think this is niche but this functionality is very useful especially when it comes to modifying monster save DC's with active effects. 

I have currently just started a homebrew high level DnD5e campaign where the players will be starting at level 20 and with multiclassing be able to get to level 30.  I am relying on monsters from the MM for most of my random encounters issue is monsters in DnD are not really balanced for this level of play so I have created drag and drop features for monster which boost their stats. I am relying on DnDbeyond and other bestiary compendiums for quick importing of creatures though most have their save DC set to flat or spellcasting.

My understanding is the majority of monsters in DnD though I am sure exceptions exist, calculate their save DC off of one of their attributes. (Attribute + Prof + 8). Global DC increase is fine and has its own uses but when it comes down to having a creature like a half-dragon who has its breath weapon calculate off of one of its physical attributes if you were to give it a spellcasting ability you may only want to increase the save DC for its abilities not its spellcasting.

Now I know the easy answer here is just manually configure each monster. I would though like to automate most of this task so to speak which is why I have endeavored to make drag and drop boosts for my monsters to give even basic weak monsters a bump in power. The area I have struggled in though is boosting save DC. There is a way round this by just boosting attributes but then this has other gameplay and balance implications. 

For example a monster ability may use its Con modifier for working out its save DC, if you boost Con here to increase the save DC you also increase its health and its own saving throw. 

Again its niche but I can see many other uses for it as well.

aaclayton:
I don't feel that this PR addresses a clear need in the D&D5e rules or handles potential edge cases which arise from configuring additional DC bonuses in this way. Sorry.

PR #1887, opened 2022-10-21
[#1835] Add `feat` data definition with action & activated effect templates

arbron:
Add data definitions for the `feat` type and the required `ActionTemplate` & `ActivtedEffectTemplate`.


module/data/item/templates/activated-effect.mjs
```diff
@@ -0,0 +1,106 @@
...
+        units: new foundry.data.fields.StringField({required: true, blank: true, label: "DND5E.TargetUnits"}),
+        type: new foundry.data.fields.StringField({required: true, blank: true, label: "DND5E.TargetType"})
+      }, {label: "DND5E.Target"}),
+      range: new foundry.data.fields.SchemaField({
+        value: new foundry.data.fields.NumberField({required: true, min: 0, label: "DND5E.RangeNormal"}),
+        long: new foundry.data.fields.NumberField({required: true, min: 0, label: "DND5E.RangeLong"}),
+        units: new foundry.data.fields.StringField({required: true, blank: true, label: "DND5E.RangeUnits"})
+      }, {label: "DND5E.Range"}),
+      uses: new foundry.data.fields.SchemaField({
+        value: new foundry.data.fields.NumberField({required: true, min: 0, label: "DND5E.LimitedUsesAvailable"}),
```
Fyorl, line 58: `integer: true`

module/data/item/templates/activated-effect.mjs
```diff
@@ -0,0 +1,106 @@
...
+          required: true, blank: false, nullable: true, initial: null, label: "DND5E.LimitedUsesPer"
+        }),
+        recovery: new FormulaField({required: true, label: "DND5E.RecoveryFormula"})
+      }, {label: "DND5E.LimitedUses"}),
+      consume: new foundry.data.fields.SchemaField({
+        type: new foundry.data.fields.StringField({required: true, blank: true, label: "DND5E.ConsumeType"}),
+        target: new foundry.data.fields.StringField({
+          required: true, blank: false, nullable: true, initial: null, label: "DND5E.ConsumeTarget"
+        }),
+        amount: new foundry.data.fields.NumberField({required: true, label: "DND5E.ConsumeAmount"})
```
Fyorl, line 70: `integer: true`

module/data/item/templates/activated-effect.mjs
```diff
@@ -0,0 +1,106 @@
...
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static migrateMaxUses(source) {
+    if ( (source.uses?.max === 0) || (source.uses?.max === "0") ) source.uses.max = "";
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Fix issue with some imported range data that uses the format "100/400" in the range field,
```
Fyorl, line 96: Imported from where? Is this actually a migration from a format we used in the system previously?

module/data/item/templates/activated-effect.mjs
```diff
@@ -0,0 +1,106 @@
...
+   * rather than splitting it between "range.value" & "range.long".
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static migrateRanges(source) {
+    if ( typeof source.range?.value !== "string" ) return;
+    const split = source.range.value.split("/");
+    if ( !Number.isNaN(split[0]) ) source.range.value = Number(split[0]);
+    if ( (split.length > 1) && !Number.isNaN(split[1]) ) source.range.long = Number(split[1]);
```
Fyorl, line 104: We should use `Number.isNumeric`. `Number.isNaN` only checks for the `NaN` value explicitly, which these values can't be because they're strings.

```suggestion
    const [value, long] = source.range.value.split("/");
    if ( Number.isNumeric(value) ) source.range.value = Number(value);
    if ( Number.isNumeric(long) ) source.range.long = Number(long);
```

templates/items/parts/item-activation.hbs
```diff
@@ -75,7 +75,7 @@
         <span class="sep">/</span>
         <input type="number" step="any" name="system.range.long"
             value="{{system.range.long}}" placeholder="{{localize 'DND5E.Max'}}"
```
Fyorl, line 77: Should this placeholder also change if we're calling it the long range now?

Fyorl (changes requested):
Nice work, thanks Jeff, just a couple of queries and some areas where I think we might want to be strict about integer numbers, let me know what you thing.

module/data/item/feat.mjs
```diff
@@ -0,0 +1,31 @@
...
+ */
+export default class FeatData extends SystemDataModel.mixin(
+  ItemDescriptionTemplate, ActivatedEffectTemplate, ActionTemplate
+) {
+  static defineSchema() {
+    return this.mergeSchema(super.defineSchema(), {
+      requirements: new foundry.data.fields.StringField({required: true, label: "DND5E.Requirements"}),
+      recharge: new foundry.data.fields.SchemaField({
+        value: new foundry.data.fields.NumberField({
+          required: true, integer: true, minimum: 1, label: "DND5E.FeatureRechargeOn"
```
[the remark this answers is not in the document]
arbron: My concern here is that with `positive: true`, any existing values of `0` will throw an error and not load the item. With `min: 1`, the `0` is just changed to `1` and the item is loaded properly.

module/data/abstract.mjs
```diff
@@ -42,13 +42,26 @@ export default class SystemDataModel extends foundry.abstract.DataModel {
...
   /* -------------------------------------------- */
 
+  /**
+   * Merge two schema definitions together as well as possible.
+   * @param {object} a  First schema that forms the basis for the merge. *Will be mutated.*
+   * @param {object} b  Second schema that will be merged in, overwriting any non-mergeable properties.
+   * @returns {object}  Fully merged schema.
+   */
+  static mergeSchema(a, b) {
+    Object.assign(a, b);
```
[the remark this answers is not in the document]
arbron: Yeah, I left is simple now until `ConsumableData` where we will actually need to add more logic to this. I figured it would be good to get the merge method in now though so the types can call it as I add them.

module/data/item/templates/activated-effect.mjs
arbron: This wasn't used in the system, but I encountered this format in one of my older worlds with content imported from D&D Beyond. I figured it would be good to handle cases like this for the moment to ensure people don't lose items when we add these models.

module/data/item/feat.mjs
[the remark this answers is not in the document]
Fyorl: Yeah that's fair, I don't think we need to be strict here.

module/data/abstract.mjs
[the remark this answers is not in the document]
Fyorl: Yeah that makes sense, we can do that.

module/data/item/templates/activated-effect.mjs
Fyorl: What currently happens to that data now that we changed those inputs on the sheet to `number` inputs? It's just lost?

arbron: If you open the sheet it is set to `null`. We could just change the migrate to discard this information if you think this log is too specific, but we need something to prevent these items from erroring out.

Fyorl: It's fine, I don't think there's any problem with leaving this in. I'm a little wary of setting a precedent for migrating module-specific data, but if this was a string field at some point then it's plausible a user might enter such data themselves.

Fyorl (approved):

Issue #1575, opened 2022-06-02
[Feature Request] Custom Spell Progression


aaclayton:
Originally in GitLab by @etoy22

I would appreciate it if there was a way to make custom spell progression. As there exists an easier way automating class progression it would be helpful if this could exist so that i can pair the two together to design the custom class that i am looking to make.

Issue #1550, opened 2022-05-18
Refactor `_computeSpellcastingProgression` to allow easier customization of spell level logic


arbron:
Currently all of an actor's spell slots are prepared in one monolithic method which doesn't allow much room for modules to modify the preparation logic. For example, because the logic for determining pact slots is buried within this method, it isn't currently possible to rework that to support pact casters higher then level 20 or the partial pact caster logic in the "Order of the Profane Soul" Blood Hunter subclass.

- [x] Split logic that determines how classes contribute to spellcaster level into separate methods (one for leveled, one for pact)
- [x] Split logic that generates and populates spell slots in actor data into separate methods
- [x] Add hooks to both of those methods to allow modules to complement or override their functionality
- [x] Move pact slot data into `CONFIG` rather than hardcoded formulas

Issue #1540, opened 2022-05-08
Advancement Type: Spellcasting


arbron:
This advancement type allows a class & subclass's spellcasting to be configured. This will replace the current spellcasting configuration in those items' system data.

## Level Up Interface

The interface will display what spellcasting type is configured and the associated ability.

## Configuration

The `type` property indicates whether `leveled` or `pact` casting is used. It can also support custom spellcasting types if introduced by modules.

The second `progression` property controls how much spellcasting the character gets per class level (`full`, `half`, `third`, or something else for custom spellcasting types).

The primary spellcasting ability is set with `ability`.

If a class requires spells to be prepared, then `preparation` is set to true. Number of prepared spells is equal to the spellcasting ability + spellcasting levels for the class, though custom spellcasting types can define their own formulas.

If the class allows ritual casting, `ritual` can be set to `true`.

```javascript
{
  "type": "leveled",
  "progression": "half",
  "ability": "wis",
  "preparation": true,
  "ritual": false
}
```

## Value

No player data is necessary.

## Future Directions

Potentially the ability to customize the primary spellcasting ability could be defined. The advancement could include a list of abilities that that player could choose from when gaining spellcasting.

The ability to customize the preparation formula might be helpful, but that could probably be better handled through custom spellcasting types.

Keeping track of type of magic granted (e.g. arcane, divine, or primal) might be useful, though that information isn't used for much mechanically in 5e.

Information on how the character gains spells and how many they can know will at any given level will be handled by a separate `SpellsAdvancement`.

Zanderaf:
"Potentially the ability to customize the primary spellcasting ability could be defined. The advancement could include a list of abilities that that player could choose from when gaining spellcasting."

Does this mean the ability to choose your spellcasting ability score if a class has that option? It would be nice if that was the case.

DaveParkinglot:
I was linked to this issue because I was asking about how to add advancements to Classes and Features that handle Spell selection.

One of the biggest challenges I have getting new players into Foundry is character creation. It has come a very long way in a fairly short time, but selecting Spells is still a huge headache for my players (and therefore for me, and other DMs on my server). Other systems like D&D Beyond have this implemented, and it would be great to have streamlined spell selection in Foundry 5e.

From discussing in the Foundry Discord, I realize that there are edge cases like the Bard's level 10 feature Magical Secrets that make this harder than just setting up Advancements for each Class/Feature to handle spellcasting.

[...]

PR #1888, opened 2022-10-24
[#1550, #1575] Refactor spellcasting preparation & add support for custom progressions

arbron:
- Refactors `_prepareSpellcasting` method to make the spellcasting preparation process clearer and remove special cases
- Adds `_computeLeveledProgression` & `_computePactProgression` methods for build progression object from classes
- Adds `dnd5e.computeLeveledProgression` & `dnd5e.computePactProgression` hooks to allow for overriding the default behavior
- Adds `_prepareLeveledSlots` & `_preparePactSlots` and related hooks to allow for customizing slot creation behavior

This also introduces a `CONFIG.DND5E.spellcastingTypes` object that will eventually replace `spellProgression`. It defines two top-level types (`leveled` and `pact`) with the option for new additions by modules. Individual progressions (`full`, `half`, `third`, `artificer`) are defined underneath that top-level type. The type determines which hooks are called, so if you add a third type for psionics then the `dnd5e.computePsionicProgression` & `dnd5e.preparePsionicSlots` hooks will be called to allow for configuring that new spellcasting type.

Eventually when #1540 is implemented it would store `type` & `progression` as two separate configuration values rather than a single one like now. For the moment this simply derives the necessary information from the single `progression` value.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -621,16 +621,20 @@ export default class Actor5e extends Actor {
 
     // NPCs don't get spell levels from classes
     if ( this.type === "npc" ) {
-      // TODO: Add better support for half, third, or pact caster NPCs
       progression.slot = this.system.details.spellLevel ?? 0;
     }
 
     else {
+      const typeCounts = {};
       // Grab all classes with spellcasting
       const classes = this.items.filter(c => {
-        return (c.type === "class") && (c.system.spellcasting.progression !== "none");
+        if ( c.type !== "class" ) return false;
+        const prog = CONFIG.DND5E.spellProgression[c.system.spellcasting.progression];
+        if ( !prog.type ) return false;
+        typeCounts[prog.type] ??= 0;
+        typeCounts[prog.type] += 1;
+        return true;
       });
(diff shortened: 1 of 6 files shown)


module/documents/actor/actor.mjs
```diff
@@ -605,72 +607,161 @@ export default class Actor5e extends Actor {
...
+      pactLevel = actor.system.details.spellLevel;
+    }
+
+    // TODO: Allow pact level and slot count to be configured
+    if ( pactLevel > 0 ) {
+      spells.pact.level = Math.ceil(Math.min(10, pactLevel) / 2); // TODO: Allow custom max pact level
+      if ( override !== null ) spells.pact.max = Math.max(override, 1);
+      else {
+        spells.pact.max = Math.max(1, Math.min(pactLevel, 2), Math.min(pactLevel - 8, 3), Math.min(pactLevel - 13, 4));
+      }
```
Fyorl, line 760: Prefer to swap these branches so that the `else` doesn't read like a double-negative.

Fyorl (changes requested):
Nice refactor, Jeff. Only some minor comments.

Fyorl (approved):

PR #1890, opened 2022-10-25
[#1835] Create system data models for spells

### LINKED ISSUE #1572: Remove components.value from spell Item template
This isn't used anywhere in the system, it seems to be left over from a former method of describing materials.

arbron:
Spells!

Resolves #1572


module/data/fields.mjs
```diff
@@ -91,3 +91,81 @@ export class IdentifierField extends foundry.data.fields.StringField {
...
+  _cleanType(value, options) {
+    Object.values(value).forEach(v => this.model.clean(v, options));
+    return value;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  getInitialValue(data) {
+    let keys = this.options.initialKeys;
```
Fyorl, line 129: We typically try to document these extra options in `_defaults`, even if it's just declared as `initialKeys: undefined`.

module/data/fields.mjs
```diff
@@ -91,3 +91,81 @@ export class IdentifierField extends foundry.data.fields.StringField {
...
+    if ( !keys || !foundry.utils.isEmpty(this.initial()) ) return super.getInitialValue(data);
```
Fyorl, line 130: `this.initial` is not guaranteed to be a function, though it would probably be incorrect to supply a simple value in this `ObjectField` case. Still, `super.getInitialValue` should take of this for us so we could be a bit defensive here and do this instead:

```suggestion
    const initial = super.getInitialValue(data);
    if ( !keys || !foundry.utils.isEmpty(initial) ) return initial;
```

module/data/fields.mjs
```diff
@@ -91,3 +91,81 @@ export class IdentifierField extends foundry.data.fields.StringField {
...
+    const initial = {};
+    for ( const key of keys ) initial[key] = this.model.getInitialValue();
+    return initial;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _validateType(value, options={}) {
+    if ( typeof value !== "object" ) throw new Error("must by an Object");
```
Fyorl, line 141:
```suggestion
    if ( foundry.utils.getType(value) !== "Object" ) throw new Error("must be an Object");
```

module/data/fields.mjs
```diff
@@ -91,3 +91,81 @@ export class IdentifierField extends foundry.data.fields.StringField {
...
+    if ( !foundry.utils.isEmpty(errors) ) throw new foundry.data.fields.ModelValidationError(errors);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Validate each value of the object.
+   * @param {object} value    The object to validate.
+   * @param {object} options  Validation options.
+   * @returns {object}        An object of value-specific errors by key.
```
Fyorl, line 152:
```suggestion
   * @returns {Object<Error>}        An object of value-specific errors by key.
```

module/data/item/spell.mjs
```diff
@@ -0,0 +1,89 @@
...
+        initialKeys: [...Object.keys(CONFIG.DND5E.spellComponents), ...Object.keys(CONFIG.DND5E.spellTags)]
+      }),
+      materials: new foundry.data.fields.SchemaField({
+        value: new foundry.data.fields.StringField({required: true, label: "DND5E.SpellMaterialsDescription"}),
+        consumed: new foundry.data.fields.BooleanField({required: true, label: "DND5E.SpellMaterialsConsumed"}),
+        cost: new foundry.data.fields.NumberField({
+          required: true, initial: 0, min: 0, label: "DND5E.SpellMaterialsCost"
+        }),
+        supply: new foundry.data.fields.NumberField({
+          required: true, initial: 0, min: 0, label: "DND5E.SpellMaterialsSupply"
```
Fyorl, line 53: `integer: true` maybe?

Fyorl (changes requested):

module/data/item/spell.mjs
arbron: I'm not 100% sure how people are using these fields so I'm reluctant to rule out having decimal values here.

module/data/fields.mjs
```diff
@@ -91,3 +91,96 @@ export class IdentifierField extends foundry.data.fields.StringField {
...
+    return value;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  getInitialValue(data) {
+    let keys = this.initialKeys;
+    const initial = super.getInitialValue(data);
+    if ( !keys || !foundry.utils.isEmpty(initial) ) return super.getInitialValue(data);
```
Fyorl, line 146:
```suggestion
    if ( !keys || !foundry.utils.isEmpty(initial) ) return initial;
```

module/data/item/spell.mjs
Fyorl: Yeah that's fair enough, I don't feel particularly strongly.

Fyorl (approved):

Issue #1857, opened 2022-10-10
Remove `none` from various item fields


arbron:
Several item fields ("Activation Cost", "Target", & "Range") include a blank value as well as a `"None"` value, both of which mean the same thing. This `none` value should be removed and any items that currently use it should be migrated to a blank value.

Varriount:
Are you sure it shouldn't be the other way around? Remove `""` and display "None" in the UI?

Fyorl:
For the vast majority of drop-downs in the UI where we want to represent 'no active choice made' or 'not applicable', we use a blank value. Jeff has identified a few places where have a blank option as well as a 'none' option, with no functional or semantic difference between the two. Removing the blank value and leaving 'none' would make these places inconsistent with the rest of the UI.

Varriount:
I guess, my general argument is that the entire UI should use "none", rather than an empty selection, since it's more explicit to the user. I realize that would take an outsized amount of effort though.

aaclayton:
Addressed <https://github.com/foundryvtt/dnd5e/issues/1857> and closed <https://github.com/foundryvtt/dnd5e/pull/1893/files> via <https://github.com/foundryvtt/dnd5e/commit/92c4b59fc1195ec438357b55cfbc05f4e4903c58>

PR #1893, opened 2022-10-26
[#1857] Remove the `none` option from various item fields

LiamFallon213:
Fixes some of #1857


module/applications/actor/character-sheet.mjs
```diff
@@ -81,7 +81,7 @@ export default class ActorSheet5eCharacter extends ActorSheet5e {
       item.hasUses = uses && (uses.max > 0);
       item.isOnCooldown = recharge && !!recharge.value && (recharge.charged === false);
       item.isDepleted = item.isOnCooldown && (uses.per && (uses.value > 0));
-      item.hasTarget = !!target && !(["none", ""].includes(target.type));
+      item.hasTarget = !!target && !([""].includes(target.type));
```
Varriount, line 84: I think this is equivalent?
  - `includes` doesn't need to be used anymore, since we're only checking for one value.
  - `target` should always be an object.
  - `!!target.type` will be false if `target.type` is a blank string, null, undefined, etc.
```suggestion
      item.hasTarget = !!target.type;
```

.gitignore
```diff
@@ -12,3 +12,6 @@ node_modules/
 
 # OS Specific
 .DS_Store
+.vscode/launch.json
+*.code-workspace
+.vscode/launch.json
```
arbron, line 17: These are IDE entries, so they should go under the IDE comment. Also, `.vscode/launch.json` is here twice. It should probably just be `.vscode/` to the entire directory is ignored.

arbron (changes requested):
This will also require removing `"none"` from the dropdowns on items sheets and adding a migration to convert any existing `"none"` values to a blank string.

aaclayton:
Addressed <https://github.com/foundryvtt/dnd5e/issues/1857> and closed <https://github.com/foundryvtt/dnd5e/pull/1893/files> via <https://github.com/foundryvtt/dnd5e/commit/92c4b59fc1195ec438357b55cfbc05f4e4903c58>

PR #1895, opened 2022-10-26
[Not ready] Integration of the Sidekick sheet

p4535992:
**THIS IS NOT A READY PR AND IS A VERY LOW PRIORITY**

A tentative to add the Sidekick implementation from TCE (https://www.dndbeyond.com/sources/tcoe/dungeon-masters-tools#Sidekicks)

I am not an very expert in the rules or in the system code, I open this PR to keep track of comments and advice regarding the code i'm trying to made for adding a character sheet dedicated to the sidekicks mechanism.

The most popular solution is to simply use the "Character" type, and indeed I started with extending that class by limiting the choice of classes, but I would like to keep the type and the sheet separate, as was done for the "Vehicle" type .

So here my questions:
- Is the creation of a new type "Sidekick" the right solution ? or is it better to create an alternative sheet for the type "character" ?
 - What might be the best way to handle the three classes "Expert", "Spellcaster" , and "Warrior" , based on the feedback from the first question, (I would also like to give the possibility to add via modules new classes for the sidekick, just like the classic class for characters).
- What attributes and/or features should possibly be disabled on the "Sidekick" sheet ? 
for now I have started to use the character sheet and i disabled/hidden:
  
  - Level
  - Subclass


krbz999:
It seems more logical to me that 'Sidekick' is the actor's class, while Warrior / Spellcaster / Expert is the actor's subclass (chosen at level 1). They definitely do have levels. Going strictly by the wording, however, Warrior/Spellcaster/Expert is the class, so only 'Subclass' should be disabled.

Additionally, they (might) have more hit dice than they have class levels.

p4535992:
> It seems more logical to me that 'Sidekick' is the actor's class, while Warrior / Spellcaster / Expert is the actor's subclass (chosen at level 1). They definitely do have levels. Going strictly by the wording, however, Warrior/Spellcaster/Expert is the class, so only 'Subclass' should be disabled.
> 
> Additionally, they (might) have more hit dice than they have class levels.

This is actually... a very good idea ? I don't think there will be people commenting on the fact that sidekick classes are treated as subclasses.... i'll wait for some other commnet, but i really like this idea.

Fyorl:
I don't have a good feeling for what the correct solution is here either. These are some pros and cons of either option that I can think of which might help us decide:

### Sidekicks as `character` type
**Pros**
* Limited impact on the codebase. We make a few assumptions across the codebase about there being only really 3 actor types. So keeping it to 3 would mean changes in this PR can be quite localised.
* Character actors already have existing support for classes, subclasses, and advancements.
* Sidekicks behave most like characters from a rules perspective: They have class levels and ASIs, PB and HD are tied to class levels, they explicitly get death saves, etc.
* Don't necessarily need to create an entirely new sheet.

**Cons**
* Slight weirdness with HD which will require special-casing. DoIP sidekicks have level + 1 HD, and Tasha's sidekicks have level + whatever their original creature's HD was (A CR 1/2 Thug starts with 5HD and 32 HP!)
* If we do want to create a new sheet, it won't be tied to the creature type.

### Sidekicks as their own type
**Pros**
* They can have their own simplified sheet that is tied to their actor type.
* Other modules can add their own sheets that target only sidekicks.
* Easier to split out logic that applies solely to sidekicks.

**Cons**
* Messy implementation. We have to go through a lot of the codebase and change places that are restricted to `character` to `character` and `sidekick`.

Well, writing these out helped a bit. I think I'm leaning more towards the `character` implementation. It feels like sidekicks are really just a collection of 3 classes/subclasses that are a lot simpler than 'full' classes, and they are identical to PCs in almost every other way. Happy to hear other opinions though, and other pros/cons I might have missed.

p4535992:
since I can't keep up with it at the moment and the large amount of updates going on this project, I decided to pick it up again after the release of 2.1.0.
I thank everyone for the excellent advice and views.

PR #1897, opened 2022-10-26
Formulas in delta fields

LiamFallon213:
This adds functionality to execute formulas in Delta Fields. This pull also implements the logic within #1860, this is because the new logic is an extension, and dependent on the logic in #1860 .

The use case for this is primarily complex situations where multiple effects may be added nearly simultaneously. 

For example, if a DM mishears their player, they may want to both add and subtract at the same time. 

This also works very well for currency, where a player may in the same transaction receive and spend money. This shortens the amount of time needed for them to process the change.

This is also very useful for damage multipliers. If you need to apply damage in a way like, they rolled for damage IRL and you don't want to do that math, the sheet can now do it for you. I would at some point like to expand this to work as a delta but for now I believe this is added functionality at a very low complexity cost.


LiamFallon213:
@Fyorl This is the PR I opened that contains the new functionality of evaluating formulas in delta input fields.

Varriount:
Hm, how will this impact existing modules?

PR #1898, opened 2022-10-26
[#1835] Create system data model for weapons

arbron:
Adds `WeaponData` & `MountableTemplate`


Fyorl (approved):
LGTM

Fyorl (changes requested):
I was only checking the GoS rules when reviewing, and forgot to check BG:DiS rules for Land vehicles. Their equipment doesn't seem to have AC and HP so we might want to change those to `nullable: true`, and modify the `template.json` so those values start as `null`.

Fyorl (approved):

PR #1903, opened 2022-10-27
[#1901] Take ammo's critical threshold into account when calculating attack's threshold

### LINKED ISSUE #1901: Have attack rolls be affected by the critical threshold of consumed ammunition
It would make sense that a weapon that normally crits on a roll of 20 but has been supplied ammunition that has been set to crit on a 19 or 20, for example, would also crit on a 19. The ammunition attack bonus and damage formulas are already added to the weapon, so this follows that logic.

krbz999:
Added ammunition's critical threshold to an item's `getCriticalThreshold`. Resolves #1901.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -609,7 +609,7 @@ export default class Item5e extends Item {
    * - the constant '20'
    * @returns {number|null}  The minimum value that must be rolled to be considered a critical hit.
    */
-   getCriticalThreshold() {
+  getCriticalThreshold() {
     const actorFlags = this.actor.flags.dnd5e || {};
     if ( !this.hasAttack ) return null;
     let actorThreshold = null;
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -613,14 +613,14 @@ export default class Item5e extends Item {
     const actorFlags = this.actor.flags.dnd5e || {};
     if ( !this.hasAttack ) return null;
     let actorThreshold = null;
-    let it = this.system.critical?.threshold ?? 20;
-    let at = 20;
+    let itemThreshold = this.system.critical?.threshold ?? Infinity;
+    let ammoThreshold = Infinity;
     if ( this.type === "weapon" ) actorThreshold = actorFlags.weaponCriticalThreshold;
     else if ( this.type === "spell" ) actorThreshold = actorFlags.spellCriticalThreshold;
     if ( this.system.consume?.type === "ammo" ) {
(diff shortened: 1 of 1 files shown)


module/documents/item.mjs
```diff
@@ -605,16 +605,22 @@ export default class Item5e extends Item {
...
+   * - item document's ammunition (if it has any)
    * - the constant '20'
    * @returns {number|null}  The minimum value that must be rolled to be considered a critical hit.
    */
   getCriticalThreshold() {
     const actorFlags = this.actor.flags.dnd5e || {};
     if ( !this.hasAttack ) return null;
     let actorThreshold = null;
+    let it = this.system.critical?.threshold ?? 20;
+    let at = 20;
```
Varriount, line 617: Perhaps it would be better to use more descriptive variables names, `itemThreshold` and `ammoThreshold`?

arbron, line 617:
```suggestion
    let itemThreshold = this.system.critical?.threshold ?? Infinity;
    let ammoThreshold = Infinity;
```

module/documents/item.mjs
```diff
@@ -605,16 +605,22 @@ export default class Item5e extends Item {
...
     if ( this.type === "weapon" ) actorThreshold = actorFlags.weaponCriticalThreshold;
     else if ( this.type === "spell" ) actorThreshold = actorFlags.spellCriticalThreshold;
-    return Math.min(this.system.critical?.threshold ?? 20, actorThreshold ?? 20);
+    if ( this.system.consume?.type === "ammo" ) {
+      at = this.actor.items.get(this.system.consume.target)?.system.critical?.threshold ?? 20;
```
arbron, line 621:
```suggestion
      ammoThreshold = this.actor.items.get(this.system.consume.target)?.system.critical.threshold ?? Infinity;
```

module/documents/item.mjs
```diff
@@ -605,16 +605,22 @@ export default class Item5e extends Item {
...
+    }
+    return Math.min(it, at, actorThreshold ?? 20);
```
arbron, line 623: Falling back to `Infinity` for `itemThreshold` & `ammoThreshold` will allow `actorThreshold` to be set to higher than 20 to prevent crits.

```suggestion
    return Math.min(itemThreshold, ammoThreshold, actorThreshold ?? 20);
```

arbron (changes requested):

module/documents/item.mjs
Varriount: Clever!

Fyorl (approved):

PR #1904, opened 2022-10-27
[#1787] Create Spell Lists custom journal page

### LINKED ISSUE #1787: Add custom journal entry page for spell lists
Add a journal entry page type that represents a spell list. This will contain a type (class, subclass, background, race, other), an identifier (matching the associated item, so the Bard spell lists would be `bard`), and a list of spells. This would automatically produce a properly formatted spell lists such as that in the SRD.

### LINKED ISSUE #1855: Paladin 5th Level Spell List Error
image

This should include Dispel Evil and Good, not Detect Evil and Good

arbron:
Adds a new journal entry page type to represent spell lists:

Spell - By Level

Spell - Editing Screen

The `"Spell List Type"` and `"Identifier"` values are there to support future features to allow limited spell selections to a specific class list, automatically adding spells for a specific class or subclass, or other options that require a formatted list of spells.

Four grouping modes are supported for spells, `"By Level"` which can be seen above as well as `"By School"`:

Spell - By School

`"By First Letter"`:

Spell - By First Letter

and `"No Grouping"`:

Spell - No Grouping

A default grouping mode can be set while editing the spell list, and the displayed grouping mode can be selected while viewing the list (this box doesn't save to the DB, so it is changeable even in a locked compendium).

Resolves #1787 
Resolves #1855


Fyorl:
I expect we'll have some solution to spell lists in the core system at some point. Would it make more sense for these pages to read from that source and display the information, rather than having the list configured and stored directly inside the journal page? Or is this the actual proposed solution for spell lists, and we expect to be reading from the data stored in these pages to manage spell lists in the rest of the system?

arbron:
> I expect we'll have some solution to spell lists in the core system at some point. Would it make more sense for these pages to read from that source and display the information, rather than having the list configured and stored directly inside the journal page? Or is this the actual proposed solution for spell lists, and we expect to be reading from the data stored in these pages to manage spell lists in the rest of the system?

My current though is that these journal pages would be the canonical source of spell lists that are then compiled by the system on load into a central data store. So we and various modules could each define a journal page for their spell lists, and then either automatically via journal scanning or via an explicit call in a hook those could be merged together into an in-memory spell list.

Initially I was thinking this would be a good place to use generic documents, but then those were nixed and custom journal pages added so this seemed like a solution. At least I think spell lists in journal pages makes more sense than spell lists as a new item type.

Fyorl:
> Initially I was thinking this would be a good place to use generic documents, but then those were nixed and custom journal pages added so this seemed like a solution. At least I think spell lists in journal pages makes more sense than spell lists as a new item type.

Yeah, ok, I did wonder if that might be the case. We should align on what our strategy is going to be for spell lists. It may well turn out that the best solution is indeed a custom journal page, but we probably need to do some design work first.

DreamingHero1997:
I think one important feature that is missing from spells at the moment is the ability to assign it to classes, if this were added then that would add the ability to automate the spell list generation, this could also make it possible to be filtered for use with other modules such as [Compendium Browser](https://github.com/League-of-Foundry-Developers/compendium-browser) which allow filtering through compendiums in a neat manner.

Zanderaf:
Is there room to add an option to denote a spell as having the Ritual Casting option?

kaelad02:
@Zanderaf This already exists. It's the last checkbox in the Spell Components section.

PR #1905, opened 2022-10-27
[#1835] Create system data model for consumables

arbron:
This one makes use of the more complex schema merging behavior to add the `autoDestroy` field to the `uses` object from `ActivatedEffectTemplate`.

### DIFF
--- a/module/data/abstract.mjs
+++ b/module/data/abstract.mjs
@@ -12,8 +12,8 @@
  *
  * Here we have opted to instead create a separate Data Model for each template available. These define their own
  * schemas which are then mixed-in to the final schema for the Document type's Data Model. A Document type Data Model
- * can define its own schema unique to it, and then add templates in direct correspondence to those in template.json via
- * SystemDataModel.mixin.
+ * can define its own schema unique to it, and then add templates in direct correspondence to those in template.json
+ * via SystemDataModel.mixin.
  */
 export default class SystemDataModel extends foundry.abstract.DataModel {
 
@@ -56,23 +56,7 @@ export default class SystemDataModel extends foundry.abstract.DataModel {
    * @returns {DataSchema}  Fully merged schema.
    */
   static mergeSchema(a, b) {
-    for ( const key of Object.keys(b) ) {
-      if ( !(key in a) || (a[key].constructor !== b[key].constructor) ) {
-        a[key] = b[key];
-        continue;
-      }
-      switch ( b[key].constructor ) {
-        case foundry.data.fields.SchemaField:
-          const fields = this.mergeSchema(a[key].fields, b[key].fields);
-          Object.values(fields).forEach(f => f.parent = undefined);
-          a[key] = new foundry.data.fields.SchemaField(
-            fields, foundry.utils.mergeObject(a[key].options, b[key].options)
-          );
-          break;
-        default:
-          a[key] = b[key];
-      }
-    }
+    Object.assign(a, b);
     return a;
   }
 
--- a/module/data/abstract.mjs
+++ b/module/data/abstract.mjs
@@ -51,12 +51,28 @@ export default class SystemDataModel extends foundry.abstract.DataModel {
 
   /**
    * Merge two schema definitions together as well as possible.
-   * @param {object} a  First schema that forms the basis for the merge. *Will be mutated.*
-   * @param {object} b  Second schema that will be merged in, overwriting any non-mergeable properties.
-   * @returns {object}  Fully merged schema.
+   * @param {DataSchema} a  First schema that forms the basis for the merge. *Will be mutated.*
+   * @param {DataSchema} b  Second schema that will be merged in, overwriting any non-mergeable properties.
+   * @returns {DataSchema}  Fully merged schema.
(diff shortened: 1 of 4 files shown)


module/data/abstract.mjs
```diff
@@ -56,7 +56,23 @@ export default class SystemDataModel extends foundry.abstract.DataModel {
    * @returns {object}  Fully merged schema.
```
Fyorl, line 56: I guess this should actually be `@returns {DataSchema}`, as well as the two parameters above it that GH won't let me highlight.

module/data/abstract.mjs
```diff
@@ -56,7 +56,23 @@ export default class SystemDataModel extends foundry.abstract.DataModel {
...
    */
   static mergeSchema(a, b) {
-    Object.assign(a, b);
+    for ( const key of Object.keys(b) ) {
+      if ( !(key in a) || (a[key].constructor !== b[key].constructor) ) {
+        a[key] = b[key];
+        continue;
+      }
+      switch (b[key].constructor) {
```
Fyorl, line 64:
```suggestion
      switch ( b[key].constructor ) {
```

module/data/abstract.mjs
```diff
@@ -56,7 +56,23 @@ export default class SystemDataModel extends foundry.abstract.DataModel {
...
+        case foundry.data.fields.SchemaField:
+          const fields = this.mergeSchema(a[key].fields, b[key].fields);
+          Object.values(fields).forEach(f => f.parent = undefined);
+          a[key] = new foundry.data.fields.SchemaField(
+            fields, foundry.utils.mergeObject(a[key].options, b[key].options)
+          );
+          break;
+        default:
+          a[key] = b[key];
```
Fyorl, line 73: Perhaps some solution for merging the `options` of the two fields at least here.

arbron: I have a solution for that over on my actor data branch where we actually would need it, but I figured I would only include the code here we are actually using.

arbron:
> Yeah this is a tricky one, particularly around merging the `options` of fields, which are very important to the functioning of the field. I will probably have to talk to Atro and see how he envisioned such a thing might be solved, or whether we're treading entirely new ground here.

I figure since all of the inputs into this merging function are well known there shouldn't be a big issue. There is no user- or dynamically-generated inputs here, it is all hard-coded into the system so if any problems do arise we can find them out now.

Fyorl:
Sorry this has been left in limbo a bit, we've been busy ironing out some issues around system data models that this work (and other devs) have uncovered.

We are treading new ground a bit where merging `SchemaField`s is concerned, as they weren't really designed for such a thing. I'd like your opinion on whether the maintenance burden of *not* doing automatic schema merging would be too high, or whether it is reasonable to write reusable classes that are tailored directly to our `template.json` schema. For example, in this very simple case, we might declare:

```js
class ItemUses extends fields.SchemaField {
  constructor({autoDestroy=false, ...options}={}) {
    const schema = {
      value: new fields.NumberField(),
      max: new fields.NumberField(),
      per: new fields.StringField(),
      recovery: new fields.StringField()
    };
    if ( autoDestroy ) schema.autoDestroy = new fields.BooleanField();
    super(schema, options);
  }
}
```

It wouldn't be necessary to have a single option per additional field. One option could 'enable' several extra fields for example, depending on our needs.

I understand the actor data is the biggest challenge here where merging schemas is involved, but it should be possible for something like the 'creature' template to dynamically change the schema it returns depending on whether the parent is an actor or an npc.

arbron:
I worry about the decreased legibility of this approach with not having the `CharacterData` own its own fields. We would end up with a like a large `DetailsSchema` class with a bunch of different definitions for different actor types. Doable, but I think harder to read through than the other way around.

As far as the difficulties of merging schemas, I would be more concerned if we were working on a Foundry core `foundry.utils.mergeSchema` method which would have to take into account how all systems and modules might make use if it. I think the problem space is a lot more constrained because this is an internal method and we can test every place it is being used thoroughly to make sure it works. We could even make it clearer that this is a private method by underscoring it and changing the documentation.

Fyorl:
I take your point about our implementation here being much more constrained, and I think an underscore-prefixed method is a good idea if we do decide to go that route.

I think we definitely want to avoid illegibility, but we have a few techniques available for composing schemas in a more deliberate fashion instead of automatic merging. For example, we can still have `CharacterData` owns its own fields if that is the concern:

```js
const Biography = options => new fields.SchemaField({
  value: new fields.HTMLField(),
  public: new fields.HTMLField()
}, options);

const creatureDetails = () => ({
  alignment: new fields.StringField(),
  race: new fields.StringField()
});

class CharacterData extends DataModel {
  static defineSchema() {
    return {
      details: new fields.SchemaField({
        background: /* ... */,
        originalClass: /* ... */,
        xp: /* ... */,
        biography: Biography(),
        ...creatureDetails()
      });
    };
  }
}
```

There's a certain amount of elegance to the automatic merge design, but there's also value in having our schema be more explicit and avoiding extra layers of indirection, so we should consider the trade-offs carefully.

PR #1907, opened 2022-10-27
[#1316] Further configuration on initiative

### LINKED ISSUE #1316: Add more extensive configuration for initiative checks
Originally in GitLab by @Pjb518

1.5.x introduces detailed configuration options for individual skills and ability check bonuses, but this release doesn't add the same options for initiative bonuses.

In a future release, I think we should add a similar configuration window for initiative, where players can set arbitrary initiative bonuses, including dice terms. Currently, only static bonuses can be added, and those bonuses cannot have flavour text. Attribute references also cannot be used in the modifier field.

Recent official releases have also indicated that WotC plans to use initiative proficiency as a mechanic going forward, so it's worth including proficiency level as part of this new configuration. See the Oath of the Watchers Paladin and Harengon race as examples.

LiamFallon213:
Setting this up so it can tracked as draft. This seeks to resolve #1316


LiamFallon213:
I doubt this is 100% ready, but I did add some requested functionality. This now uses a hidden skill to determine initiative, allowing for the dialog and potential proficiency.  This also greatly streamlines abilities like Gift of Alacrity. 

To do this I added new functionality to the overall rolling classes which will allow for use of the dialogue in the future without explicitly making a roll. Currently this is triggered by a simple if statement, but could be configured to be more useful in the future.

LiamFallon213:
No longer in draft, ready for review.

LiamFallon213:
Requesting review @arbron or @Fyorl

LiamFallon213:
@arbron I would like this reviewed so I can continue work if necessary.

module/applications/actor/character-sheet.mjs
```diff
@@ -214,7 +214,7 @@ export default class ActorSheet5eCharacter extends ActorSheet5e {
    * @returns {Promise}         Dialog or roll result.
    * @private
    */
-  _onSheetAction(event) {
+  async _onSheetAction(event) {
```
LiamFallon213, line 217: This must be asynchronous to allow the promise used in in [lines 230 - 234](https://github.com/foundryvtt/dnd5e/blob/6998d9bbbb5f722f401851bfb2b75a563227e057/module/applications/actor/character-sheet.mjs#L230-L234) to function.

module/applications/actor/character-sheet.mjs
```diff
@@ -227,8 +227,14 @@ export default class ActorSheet5eCharacter extends ActorSheet5e {
       case "rollDeathSave":
         return this.actor.rollDeathSave({event: event});
       case "rollInitiative":
-        return this.actor.rollInitiative({createCombatants: true});
+        const initiative = new Promise((resolve) => {
+          resolve(this.actor.rollSkill("ini"));
+        });
+        let z = (await initiative);
+        let x = this.actor.rollInitiative({ createCombatants: true, initiativeOptions: { formula: (""+ z +"+"+ this.actor.system.attributes.init.bonus) } });
+        return (x);
```
LiamFallon213, line 235: This section first calls a hidden initiative skill roll that builds the formula. Then passes that as a formula to the rollInitiative function to handle the rest.

module/dice/dice.mjs
```diff
@@ -51,7 +52,7 @@
 export async function d20Roll({
   parts=[], data={}, event,
   advantage, disadvantage, critical=20, fumble=1, targetValue,
-  elvenAccuracy, halflingLucky, reliableTalent,
+  elvenAccuracy, halflingLucky, reliableTalent, formulaOnly,
```
LiamFallon213, line 55: Adding a new function to the d20Roll function. Instead of always rolling a dice, you can ask for something akin to a skill check etc, and if passed a true value, instead of rolling a dice, it will make the formula then return that formula. This is used in my updated initiative function where it uses this to generate the formula then pass that to the roll initiative function.

module/dice/dice.mjs
```diff
@@ -92,6 +93,10 @@ export async function d20Roll({
     if ( configured === null ) return null;
   } else roll.options.rollMode ??= defaultRollMode;
 
+  if(formulaOnly == true){
+    return (roll.formula);
+  }
```
LiamFallon213, line 98: This is more of a temporary thing that is used only for initiative skill rolls.

module/documents/actor/actor.mjs
```diff
@@ -989,6 +989,12 @@ export default class Actor5e extends Actor {
     // Reliable Talent applies to any skill check we have full or better proficiency in
     const reliableTalent = (skl.value >= 1 && this.getFlag("dnd5e", "reliableTalent"));
 
+    let formulaOnly = false;
+
+    if(skillId == "ini"){
+      formulaOnly = true;
+    }
```
LiamFallon213, line 996: Checks if its an init check and if so sets the formulaOnly flag to true.

templates/actors/character-sheet.hbs
```diff
@@ -170,6 +170,7 @@
             <ul class="skills-list">
             {{#each config.skills as |obj s|}}
             {{#with (lookup ../system.skills s) as |skill|}}
+                {{#unless (eq skill.label "Initiative")}}
```
LiamFallon213, line 173: This allows for the "initiative" box to not be rendered in the character sheet.

.gitignore
```diff
@@ -12,3 +12,5 @@ node_modules/
 
 # OS Specific
 .DS_Store
+.vscode/launch.json
+dnd5e.code-workspace
```
Varriount, line 16: These can be removed, as a PR to add the `.vscode` folder to .gitignore is [currently open](https://github.com/foundryvtt/dnd5e/pull/1923).

module/applications/actor/character-sheet.mjs
Varriount: Hm, I'm not sure this is the correct approach to reusing the skill-rolling logic for initiative calculation. There are effects (like Bard's Jack of All Trades) that affect all skills, but don't affect initiative.

krbz999: If you don't mind me butting in, it would be more correct to be simply a Dexterity check, since that is how it is determined as per the Basic Rules.

> At the beginning of every combat, you roll initiative by making a Dexterity check. Initiative determines the order of creatures' turns in combat.

However I think it deserves to rather get the Death Saving Throw treatment. Like a Death Saving Throw is a special kind of Saving Throw, Initiative is a special kind of Ability Check, and constructing it fully without use of helpers from skills would likely end up being a lot cleaner in the end.

It would allow for roll hooks as well, e.g., `dnd5e.preRollInitiative` and `dnd5e.rollInitiative` just like there's `dnd5e.preRollDeathSave`.

LiamFallon213: > If you don't mind me butting in, it would be more correct to be simply a Dexterity check, since that is how it is determined as per the Basic Rules.
> 
> > At the beginning of every combat, you roll initiative by making a Dexterity check. Initiative determines the order of creatures' turns in combat.
> 
> However I think it deserves to rather get the Death Saving Throw treatment. Like a Death Saving Throw is a special kind of Saving Throw, Initiative is a special kind of Ability Check, and constructing it fully without use of helpers from skills would likely end up being a lot cleaner in the end.
> 
> It would allow for roll hooks as well, e.g., `dnd5e.preRollInitiative` and `dnd5e.rollInitiative` just like there's `dnd5e.preRollDeathSave`.

The reason this is like this is that WOTC has indicated things like proficiency in Initiative are going to be a thing in later books. This allows for that.

LiamFallon213: > If you don't mind me butting in, it would be more correct to be simply a Dexterity check, since that is how it is determined as per the Basic Rules.
> 
> > At the beginning of every combat, you roll initiative by making a Dexterity check. Initiative determines the order of creatures' turns in combat.
> 
> However I think it deserves to rather get the Death Saving Throw treatment. Like a Death Saving Throw is a special kind of Saving Throw, Initiative is a special kind of Ability Check, and constructing it fully without use of helpers from skills would likely end up being a lot cleaner in the end.
> 
> It would allow for roll hooks as well, e.g., `dnd5e.preRollInitiative` and `dnd5e.rollInitiative` just like there's `dnd5e.preRollDeathSave`.

Furthermore, this allows a few things. Example: this allows any skill to be used when rolling initiative which is a large upgrade, it also allows advantage and disadvantage rolls and roll bonuses, which is important for things like gift of alacrity.

krbz999: (Assuming you meant ability, not skill) Isn't that going to be set on the sheet, not in the roll dialog? Plus death saves allow for proficiency, bonuses, and advantage/disadvantage in their own method, I think that should be doable for Initiative as well in the same way.

LiamFallon213: > (Assuming you meant ability, not skill) Isn't that going to be set on the sheet, not in the roll dialog? Plus death saves allow for proficiency, bonuses, and advantage/disadvantage in their own method, I think that should be doable for Initiative as well in the same way.

This also makes it very easy to interact with it for future changes.

aaclayton:
Addressed in https://github.com/foundryvtt/dnd5e/pull/1994

PR #1908, opened 2022-10-28
[#1761] Convert item price to object to support different currencies

arbron:
Converts `price` on items into an object that contains the numeric value as well as currency units, which default to GP:

```javascript
price: {
  value: 10,
  currency: "gp"
}
```

Price field with currency dropdown

This should allow selecting alternative currencies to avoid things like setting the price of sling bullet to `0.002` or other similar issues.

- [x] Change `DataModel` & `template.json
- [x] Add migration from old data
- [x] Add into UI
- [x] Update compendiums to use new format


Varriount:
I wonder if it might be worth making this into a proper class/object, so that something like `item.system.price.toString()` would print "3 gp" (or similar)?

arbron:
> I wonder if it might be worth making this into a proper class/object, so that something like `item.system.price.toString()` would print "3 gp" (or similar)?

I don't think it would be worth the overhead. If we find that is really needed, we could add it to the label object generated during `prepareData`.

module/data/item/templates/physical-item.mjs
```diff
@@ -19,9 +19,14 @@ export default class PhysicalItemTemplate extends foundry.abstract.DataModel {
...
         required: true, nullable: false, initial: 0, min: 0, label: "DND5E.Weight"
       }),
-      price: new foundry.data.fields.NumberField({
-        required: true, nullable: false, initial: 0, min: 0, label: "DND5E.Price"
-      }),
+      price: new foundry.data.fields.SchemaField({
+        value: new foundry.data.fields.NumberField({
+          required: true, nullable: false, initial: 0, min: 0, label: "DND5E.Price"
+        }),
+        currency: new foundry.data.fields.StringField({
```
Fyorl, line 26: Slight preference for 'denomination', or 'denom' instead. What do you think?

Fyorl (changes requested):

module/data/item/templates/physical-item.mjs
arbron: Probably a more accurate term.

arbron:
@Fyorl If there are no other issues with the structure, I'll work on getting the compendiums migrated

Fyorl:
> @Fyorl If there are no other issues with the data model, I'll work on getting the compendiums migrated

Please, go ahead, no objections here.

Fyorl (approved):

PR #1913, opened 2022-10-31
[#1293] Suppress unavailable effects on synthetic actors

### LINKED ISSUE #1293: activeEffects.js determineSuppression() does not disable/enable effects for items added to unlinked Tokens

If an item, which has a transfer active effect, is added to the synthetic actor of an unlinked token, the origin field of the active effect is of the form Scene.Id.Token.Id.Item.id, not Actor.Id.Item.Id.
  
The code
```
   const [parentType, parentId, documentType, documentId] = this.data.origin?.split(".") ?? [];
```
will not correctly split out the parentId/documentId in this case and hence the item will never be suppressed. I'm unsure if this is the intended behaviour, but I would have expected that since it works for equipment added to the base Actor for the unlinked token it should work for equipment added to the token's synthetic actor as well.

krbz999:
This is an attempt to resolve #1293, changing the logic of determineSuppression given that uuids for linked and unlinked actors are different.

Updated: Using `fromUuidSync` and getting the item (the origin) immediately, and checking the equality of the effect's and item's (if any) parents, it should be a lot simpler.

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -56,20 +56,10 @@ export default class ActiveEffect5e extends ActiveEffect {
    */
   determineSuppression() {
     this.isSuppressed = false;
-    if ( this.disabled || (this.parent.documentName !== "Actor") ) return;
-    const [parentType, parentId, documentType, documentId, syntheticItem, syntheticItemId] = this.origin?.split(".") ?? [];
-    // case 1: linked or sidebar actor
-    if ( parentType === "Actor" ) {
-      if ( (parentId !== this.parent.id) || (documentType !== "Item") ) return;
-    }
-    // case 2: synthetic actor on scene
-    else if ( parentType === "Scene" ) {
-      if ( (documentId !== this.parent.token?.id) || (syntheticItem !== "Item") ) return;
-    }
-    // case 3: not an actor
-    else return;
-    const item = this.parent.items.get(documentId) ?? this.parent.items.get(syntheticItemId);
-    if ( !item ) return;
+    if ( this.disabled || (this.parent.documentName !== "Actor") || !this.origin ) return;
+    // Determine if this is an effect from an item
+    const item = fromUuidSync(this.origin);
+    if ( item?.parent !== this.parent ) return;
     this.isSuppressed = item.areEffectsSuppressed;
   }
 
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -57,9 +57,18 @@ export default class ActiveEffect5e extends ActiveEffect {
   determineSuppression() {
     this.isSuppressed = false;
     if ( this.disabled || (this.parent.documentName !== "Actor") ) return;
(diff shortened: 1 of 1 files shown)


Fyorl:
@krbz999 
> I considered using `fromUuidSync` and getting the item (the origin) immediately, but this would blow past the `parentId` (or `documentId`) checks.

I think this would still allow you to validate the parent.
```js
if ( !this.origin ) return;
const item = fromUuidSync(this.origin);
if ( !item?.parent !== this.parent ) return;
```

`item.parent` in this case should be the synthetic actor instance if the `origin` was of the `Scene.abc.Token.def.Item.ghi` variety.

module/documents/active-effect.mjs
```diff
@@ -57,9 +57,18 @@ export default class ActiveEffect5e extends ActiveEffect {
   determineSuppression() {
     this.isSuppressed = false;
     if ( this.disabled || (this.parent.documentName !== "Actor") ) return;
-    const [parentType, parentId, documentType, documentId] = this.origin?.split(".") ?? [];
-    if ( (parentType !== "Actor") || (parentId !== this.parent.id) || (documentType !== "Item") ) return;
-    const item = this.parent.items.get(documentId);
+    const [parentType, parentId, documentType, documentId, syntheticItem, syntheticItemId] = this.origin?.split(".") ?? [];
+    // case 1: linked or sidebar actor
```
Fyorl, line 61: I think our linter will complain if comments don't start with a capital letter. You can check it yourself with `npm run lint`.

module/documents/active-effect.mjs
```diff
@@ -56,11 +56,10 @@ export default class ActiveEffect5e extends ActiveEffect {
...
     this.isSuppressed = false;
-    if ( this.disabled || (this.parent.documentName !== "Actor") ) return;
-    const [parentType, parentId, documentType, documentId] = this.origin?.split(".") ?? [];
-    if ( (parentType !== "Actor") || (parentId !== this.parent.id) || (documentType !== "Item") ) return;
-    const item = this.parent.items.get(documentId);
-    if ( !item ) return;
+    if ( this.disabled || (this.parent.documentName !== "Actor") || !this.origin ) return;
+    // Determine if this is an effect from an item
+    const item = fromUuidSync(this.origin);
+    if ( item?.parent !== this.parent ) return;
```
Fyorl, line 62: Partly my fault for not giving you fully-baked code suggestions, but `fromUuidSync` can throw an error in some cases (e.g., if the UUID points to an embedded item within a document stored in a compendium), so we probably need to wrap this in a `try/catch`.

krbz999: In those cases, is there a preference for logging the error, or should it in this case simply `return`?

Fyorl: It should just `return` and treat the item as not suppressed. It's usually a mistake if the `origin` ended up in that state.

Fyorl (approved):

PR #1914, opened 2022-11-01
[#1853] Relieve automatic resizing of Item Sheets when dragged

### LINKED ISSUE #1853: Resize not respected when dragging the item sheet.
If you resize these sheets and then drag the window around, the height of the sheet gets reset back to its default size. Normally this wouldn't be too big of an issue because the height of a sheet gets reset when you switch between tabs.

However, this issue because far more annoying once you include the fact that these sheets don't account for the height of the left-side items, which causes a double scrollbar.

One or the other issue isn't too bad, but both quickly become an annoyance. lol

Here is a video showing the issue: https://streamable.com/krixez

krbz999:
This is an attempt to resolve #1853. By taking the largest of the *current* height and the default height, the rather annoying resizing of item sheets is somewhat relieved.

The automatic resizing was mainly an issue for the Description tab. There was no noticeable issue with the Effects tab unless an item had 10 or more effects. Also, this PR does not affect the Details tab at all.

### DIFF
--- a/module/applications/item/item-sheet.mjs
+++ b/module/applications/item/item-sheet.mjs
@@ -110,7 +110,7 @@ export default class ItemSheet5e extends ItemSheet {
     /** @deprecated */
     Object.defineProperty(context, "data", {
       get() {
-        const msg = `You are accessing the "data" attribute within the rendering context provided by the ItemSheet5e 
+        const msg = `You are accessing the "data" attribute within the rendering context provided by the ItemSheet5e
         class. This attribute has been deprecated in favor of "system" and will be removed in a future release`;
         foundry.utils.logCompatibilityWarning(msg, { since: "DnD5e 2.0", until: "DnD5e 2.2" });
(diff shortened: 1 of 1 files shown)


arbron (approved):

module/applications/item/item-sheet.mjs
```diff
@@ -110,7 +110,7 @@ export default class ItemSheet5e extends ItemSheet {
     /** @deprecated */
     Object.defineProperty(context, "data", {
       get() {
-        const msg = `You are accessing the "data" attribute within the rendering context provided by the ItemSheet5e 
+        const msg = `You are accessing the "data" attribute within the rendering context provided by the ItemSheet5e
```
Fyorl, line 113: I actually wonder why this is a template string at all. I thought our linter would flag these sorts of usages. Not relevant to this review though.

Fyorl (approved):

PR #1916, opened 2022-11-01
[#881] Add spell level and item id to message flags when a spell is cast

### LINKED ISSUE #881: Query: dnd1.2.0 Rolling a spell makes working out what level the spell was cast at tricky.

Prior to 1.2.0 and via actor.useSpell(), one could inspect the item that did the item.roll() to determine the spell level that was used to cast the spell. 

It seems with the new scheme the only way to determine the spell level at which the spell was cast is to inspect the result.content of the item.roll() return.

It would be convenient for cast spells if the spell level was explicitly called out in the return value of the item.roll() call, since I want to use the dnd5e standard use/cast/consume code, rather than reinventing it.

krbz999:
This is a one-liner to add a spell's level, as well as its item id, as `flags.dnd5e.use.spellLevel: <Number>`, and hopefully resolve #881.

I have put this in a separate PR to get some feedback on what the best flag would be; attack rolls use `flags.dnd5e.roll.type: "attack"` and damage rolls use `flags.dnd5e.roll.type: "damage"`, and this is not entirely in line with that, but using `flags.dnd5e.roll.type: "use"` didn't seem correct either.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -732,7 +732,7 @@ export default class Item5e extends Item {
     options = foundry.utils.mergeObject({
       configureDialog: true,
       createMessage: true,
-      "flags.dnd5e.use": {type: this.type, itemId: this.id}
+      "flags.dnd5e.use": {type: this.type, itemId: this.id, itemUuid: this.uuid}
     }, options);
 
     // Reference aspects of the item data necessary for usage
@@ -1342,7 +1342,7 @@ export default class Item5e extends Item {
         left: window.innerWidth - 710
       },
       messageData: {
-        "flags.dnd5e.roll": {type: "attack", itemId: this.id},
+        "flags.dnd5e.roll": {type: "attack", itemId: this.id, itemUuid: this.uuid},
         speaker: ChatMessage.getSpeaker({actor: this.actor})
       }
     }, options);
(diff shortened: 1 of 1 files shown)


module/documents/item.mjs
```diff
@@ -783,6 +783,7 @@ export default class Item5e extends Item {
         item.prepareFinalAttributes();
       }
     }
+    if ( isSpell ) foundry.utils.setProperty(options.flags, "dnd5e.use", {spellLevel: item.system.level, itemId: this.id});
```
Varriount, line 786: Huh, is the original item ID not already attached to the chat card?

krbz999: On `item.use()` there are no system flags at all. The item id and the spell level have always been present in the dataset, however. But as the issue says, this makes it easier.

arbron:
I'm wondering if it would just be simpler to put these at the top level, so `flags.dnd5e.spellLevel` and `flags.dnd5e.itemId`. Or perhaps something like `flags.dnd5e.item.level` & `flags.dnd5e.item.id`.

Also, it would probably be better to store the item's UUID rather than the ID otherwise it would be difficult to retrieve the original item. And the ID should be stored for all items rather than just spells.

krbz999:
I had the same thought about putting it at the top level. Something like `flags.dnd5e.use: {type: item.type, itemId: item.id}` and `spellLevel` when it is a spell.

I chose the item id because that is what is stored on attack rolls and damage rolls. Should those be changed?

Fyorl:
@krbz999
> I chose the item id because that is what is stored on attack rolls and damage rolls. Should those be changed?

Perhaps, in order to avoid breakage, we can include both for now, and retire `itemId` later? Could you update attack and damage rolls to include `itemUuid` as well?

krbz999:
> @krbz999
> 
> > I chose the item id because that is what is stored on attack rolls and damage rolls. Should those be changed?
> 
> Perhaps, in order to avoid breakage, we can include both for now, and retire `itemId` later? Could you update attack and damage rolls to include `itemUuid` as well?

I realise it would veer out of scope of the PR slightly, but I did so and added itemUuid to all other item roll types since those used itemId too. So in total, tool checks, other formula, attack, damage, and use.

Fyorl (approved):

PR #1920, opened 2022-11-05
[#1919] Disable grid snapping when placing ability templates onto gridless scenes

### LINKED ISSUE #1919: Disable grid snapping when placing ability templates onto gridless scenes
Currently, when attempting to place an ability template onto a gridless scene, the template will snap to certain points as if a square grid is enabled.

- **Do the Rules as Written (RAW) support this feature? If so, provide some examples.**
    - Yes. The [v5.1 SRD](https://media.wizards.com/2016/downloads/DND/SRD-OGL_V5.1.pdf) makes no mention of playing on a grid. The only mention of playing on a grid appears in the PHB as a variant rule.
- **Is the missing feature in the System Reference Document? If not, it might still be supportable, but it is worth mentioning in the request.**
    - Playing without a grid is not explicitly mentioned in the [v5.1 SRD](https://media.wizards.com/2016/downloads/DND/SRD-OGL_V5.1.pdf). However, neither is playing on a grid.
- **Does this feature help a GM run a fifth edition game in Foundry VTT?**
    - Yes. GMs (and the players of such games) who do not use a grid, but still use battlemaps (like myself) are plagued by inaccurate ability template placement due to snapping being the default behavior, which is not modifiable via the dnd5e settings configuration menu.

SethGallucci:
This pull request modifies the `_onMovePlacement` and `_onConfirmPlacement` methods in the `AbilityTemplate` class to cause ability templates to only undergo pixel-level snapping (effectively no snapping) when placing ability templates onto a gridless scene. Snapping behavior on gridded scenes is unaffected.

Closes #1919

### DIFF
--- a/module/canvas/ability-template.mjs
+++ b/module/canvas/ability-template.mjs
@@ -150,7 +150,8 @@ export default class AbilityTemplate extends MeasuredTemplate {
     let now = Date.now(); // Apply a 20ms throttle
     if ( now - this.#moveTime <= 20 ) return;
     const center = event.data.getLocalPosition(this.layer);
-    const snapped = canvas.grid.getSnappedPosition(center.x, center.y, 2);
+    let interval = canvas.grid.type === CONST.GRID_TYPES.GRIDLESS ? 0 : 2;
+    const snapped = canvas.grid.getSnappedPosition(center.x, center.y, interval);
     this.document.updateSource({x: snapped.x, y: snapped.y});
     this.refresh();
     this.#moveTime = now;
@@ -180,7 +181,8 @@ export default class AbilityTemplate extends MeasuredTemplate {
    */
   async _onConfirmPlacement(event) {
     await this._finishPlacement(event);
-    const destination = canvas.grid.getSnappedPosition(this.document.x, this.document.y, 2);
+    let interval = canvas.grid.type === CONST.GRID_TYPES.GRIDLESS ? 0 : 2;
+    const destination = canvas.grid.getSnappedPosition(this.document.x, this.document.y, interval);
     this.document.updateSource(destination);
     this.#events.resolve(canvas.scene.createEmbeddedDocuments("MeasuredTemplate", [this.document.toObject()]));
   }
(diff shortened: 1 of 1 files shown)


Varriount:
Out of curiosity, what is the current behavior when placing templates on gridless scenes?

SethGallucci:
https://github.com/foundryvtt/dnd5e/issues/1919#issue-1436908923
> Currently, when attempting to place an ability template onto a gridless scene, the template will snap to certain points as if a square grid is enabled.

module/canvas/ability-template.mjs
```diff
@@ -150,7 +150,8 @@ export default class AbilityTemplate extends MeasuredTemplate {
     let now = Date.now(); // Apply a 20ms throttle
     if ( now - this.#moveTime <= 20 ) return;
     const center = event.data.getLocalPosition(this.layer);
-    const snapped = canvas.grid.getSnappedPosition(center.x, center.y, 2);
+    let interval = canvas.grid.type === CONST.GRID_TYPES.GRIDLESS ? 0 : 2;
```
arbron, line 153:
```suggestion
    const interval = canvas.grid.type === CONST.GRID_TYPES.GRIDLESS ? 0 : 2;
```

module/canvas/ability-template.mjs
```diff
@@ -180,7 +181,8 @@ export default class AbilityTemplate extends MeasuredTemplate {
    */
   async _onConfirmPlacement(event) {
     await this._finishPlacement(event);
-    const destination = canvas.grid.getSnappedPosition(this.document.x, this.document.y, 2);
+    let interval = canvas.grid.type === CONST.GRID_TYPES.GRIDLESS ? 0 : 2;
```
arbron, line 184:
```suggestion
    const interval = canvas.grid.type === CONST.GRID_TYPES.GRIDLESS ? 0 : 2;
```

arbron (changes requested):

module/canvas/ability-template.mjs
SethGallucci: In the `_onRotatePlacement` method, should the `delta` and `snap` variables be instead declared with `const`, as well? I figured there was some reason they are declared with `let` and went with that to keep consistency.

arbron: Yeah, anything that isn't redefined later should be a `const`, so you can change those and also `now` in `_onMovePlacement`.

arbron (approved):

Fyorl (approved):

PR #1922, opened 2022-11-05
[#729] Containers!

arbron:
Containers that can actually contain items! Wowza! Actors get containers! Sidebar gets containers! Compendiums get containers! Containers can get other containers!

Backpack

Containers are tracked by a new `system.container` value defined on `PhysicalItemTemplate` which is the ID of the container item. The `backpack` item type then has methods for building a collection from the current `DocumentCollection`, as well as calculating properties like the number of contained items and their weight.

In addition to modifications to `ActorSheet5e` to ensure only the non-contained items are displayed, this also includes a custom version of `ItemDirectory` and `Compendium` that handle hiding items in contains and dragging and dropping full contains. This means that contains can be moved anywhere in the system you want and they will continue to behave properly.


Varriount:
Would this eventually allow for items with multiple "actions" associated with them? For example, with a [Bag of Beans](https://www.dndbeyond.com/magic-items/4579-bag-of-beans) a Player can either plant a single bean (causing a random effect) or dump the bag's contents to the ground (causing a small explosion).

arbron:
> Would this eventually allow for items with multiple "actions" associated with them? For example, a player with [Bag of Beans](https://www.dndbeyond.com/magic-items/4579-bag-of-beans) can either plant a single bean (causing a random effect) or dump the bag's contents to the ground (causing a small explosion).

No, this only supports adding physical items into containers. I have been do some design work on something like that though, but it would be a completely different design more like advancements.

arbron:
> What does the container look like on the character sheet? Can you collapse/expand the contents? And if so, can you use a contained item from that view?

Right now none of the contents are displayed on the character sheet, you have to open up the item itself. Displaying on character sheet requires a bunch of additional changes and I didn't want to bloat the scope of this PR beyond the basic requirements of containers.

p4535992:
Can you add a system setting for disable/enable this ? For anyone want to use external module for accomplish this.

RossGosling:
> Can you add a system setting for disable/enable this ? For anyone want to use external module for accomplish this.

To me it feels like if you wanted to have containers exclusively setup with external modules the onus is on those modules to prevent vanilla containers
If there's appropriate hooks for inventory changes and stuff it shouldn't be a problem?

PR #1925, opened 2022-11-10
[#1884] Add activation cost to Monk's unarmed strike

### LINKED ISSUE #1884: Unarmed Strike (Monk) Has No Activation Cost
This is a small inconsistency that actually does have some impact with other modules.

Your Unarmed Strike (Monk) weapon that is listed under Class Features doesn't have an activation cost. This is inconsistent with the other weapons. Because of this difference, it doesn't show up for the Token Action HUD module, for instance.

kjolern:
I have changed the monk class feature unarmed strike from 0 activation cost to 1 action. Fixes #1884


arbron:
Looks like this was already fixed with #1712 for the upcoming 2.1.0 release.

PR #1927, opened 2022-11-11
[#1926] Add try-catch handling to placeTemplate case of Item5e._onChatCardAction

### LINKED ISSUE #1926: Uncaught rejection of AbilityTemplate.drawPreview promise from placeTemplate case in Item5e._onChatCardAction
### Description
This bug is trivial and seems not to produce any negative side effects.
The promise created as part of creating an ability template from a chat card button is not caught when rejected.

### Expected Result
It is expected that this workflow would not produce an uncaught error.

### Actual Result
This workflow produces an uncaught error.

### Steps to Reproduce

1. Boot up a new, vanilla, dnd5e-2.0.2 world in Foundry 10.288.
2. Create a new actor of any type.
3. Add to the embedded items of the actor an item that has the placement of an ability template in its use workflow.
4. Use the item so that its chat card is created and contains a _Place Measured Template_ button. Place or discard the created template; it makes no difference, as the rejection of the promise in this case is caught.
5. Click the _Place Measured Template_ button on the newly created chat card to being the workflow for placing the ability template.
6. Right-click to cancel the placement of the ability template.
7. View the new error in the console.

SethGallucci:
This pull request adds try-catch handling to the placeTemplate case of the `_onChatCardAction` method of the `Item5e` class. The absence of handling the rejection case of the promise from the `drawPreview` method of the created `AbilityTemplate` resulted in an uncaught error. The code within the placeTemplate case has also been refactored so as to resemble the code responsible for the creation of the initial `AbilityTemplate`.

Closes #1926

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1907,8 +1907,9 @@ export default class Item5e extends Item {
       case "toolCheck":
         await item.rollToolCheck({event}); break;
       case "placeTemplate":
-        const template = dnd5e.canvas.AbilityTemplate.fromItem(item);
-        if ( template ) template.drawPreview();
+        try {
+          await (dnd5e.canvas.AbilityTemplate.fromItem(item))?.drawPreview();
+        } catch(err) {}
         break;
       case "abilityCheck":
         targets = this._getChatCardTargets(card);
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1908,7 +1908,7 @@ export default class Item5e extends Item {
         await item.rollToolCheck({event}); break;
       case "placeTemplate":
         try {
-          await (dnd5e.canvas.AbilityTemplate.fromItem(item))?.drawPreview();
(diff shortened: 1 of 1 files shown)


module/documents/item.mjs
```diff
@@ -1907,8 +1907,9 @@ export default class Item5e extends Item {
       case "toolCheck":
         await item.rollToolCheck({event}); break;
       case "placeTemplate":
-        const template = dnd5e.canvas.AbilityTemplate.fromItem(item);
-        if ( template ) template.drawPreview();
+        try {
+          await (dnd5e.canvas.AbilityTemplate.fromItem(item))?.drawPreview();
```
arbron, line 1911: The parentheses aren't necessary, so its a bit cleaner without them.

```suggestion
          await dnd5e.canvas.AbilityTemplate.fromItem(item)?.drawPreview();
```

arbron (changes requested):

arbron (approved):

PR #1935, opened 2022-11-18
Add support for a group actor type to represent parties of characters traveling together in a region map.

aaclayton:
I believe this MR is at a state where it could be merged as a MVP which satisfies the "must haves" and provides a framework that can be further developed and expanded to incorporate additional future work. Comments welcome.

## Overview
A new Actor type, `group` which has its own data schema defined as the `GroupActor` data model

A new sheet for `ActorGroupSheet` which displays group membership, inventory, description, and other attributes.
* Drop Actors on the sheet to add them to the group
* Click the [x] button to remove an Actor from the group
* View the HP bars of all actors in the group, sheet updates automatically in real-time when actors HP changes
* Configure movement speeds for the group defined in units of `land`, `air`, and `water`. **Note: perhaps this needs to be more strongly anchored in some 5e rules somewhere**
* Basic management of group inventory with categories for `weapon`, `equipment`, `backpack`, and `loot` (catchall) items
* Group actors can be added to combat encounters, but they do not have a specialized initiative formula and instead roll only a basic `1d20`. This could perhaps be improved in the future.

## Future Work
There is a lot that can still be done to improve the concept of the group actor. I propose deferring the following items as "future work" unless it is viewed that something on this list is a must have for the initial implementation.
* Support group skill or ability checks
* Support application of damage or healing to be spread among all group members
* Support differentiation of public `summary` vs private `description` as different HTML fields.

## Screenshots
image

image

image


Varriount:
What would this new kind of actor type do, that can't already be done with the existing actor types?

module/applications/actor/group-sheet.mjs
```diff
@@ -0,0 +1,198 @@
...
+        name: member.name,
+        img: member.img,
+        hp: {}
+      };
+
+      // HP bar
+      const hp = member.system.attributes.hp;
+      m.hp.current = hp.value + (hp.temp || 0);
+      m.hp.max = hp.max + (hp.tempMax || 0);
+      m.hp.pct = ((m.hp.current / m.hp.max) * 100).toFixed(2);
```
arbron, line 85: Temp HP and temp max HP lead to the bars expanding past the end of the container:

Screenshot 2022-11-19 at 10 49 41

```suggestion
      m.hp.max = hp.max + (hp.tempmax || 0);
      m.hp.pct = Math.clamped((m.hp.current / m.hp.max) * 100, 0, 100).toFixed(2);
```

That will fix the immediate problem, but we'll probably want to do a temp HP overlay like on the token bars.

module/applications/actor/group-sheet.mjs
```diff
@@ -0,0 +1,198 @@
...
+    return {sections, stats};
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Prepare movement speed data for rendering on the sheet.
+   * @returns {{secondary: string, primary: string}}
+   */
+  #prepareMovementSpeed() {
```
arbron, line 106: I'd say we should probably use the DMG's calculation rules to auto calculate travel pace based on the speed set here:

> - In 1 minute, you can move a number of feet equal to your speed times 10.
> - In 1 hour, you can move a number of miles equal to your speed divided by 10.
> - For daily travel, multiply your hourly rate of travel by the number of hours traveled (typically 8 hours).
> - For a fast pace, increase the rate of travel by one-third.
> - For a slow pace, multiply the rate by two-thirds.

Then we can display the pace values, rather than the base speed, perhaps even with a toggle between Fast/Normal/Slow pace.

dnd5e.mjs
```diff
@@ -84,6 +84,8 @@ Hooks.once("init", function() {
   CONFIG.DND5E = DND5E;
   CONFIG.ActiveEffect.documentClass = documents.ActiveEffect5e;
   CONFIG.Actor.documentClass = documents.Actor5e;
+  CONFIG.Actor.systemDataModels.group = documents.GroupActor;
```
arbron, line 87: The item and advancement data models we've implemented so far in `2.1.x` have taken a slightly different organization approach, separating the system data models from the documents in their own folder.

module/documents/actor/group.mjs
```diff
@@ -0,0 +1,93 @@
...
+ */
+export default class GroupActor extends foundry.abstract.DataModel {
+
+  /** @override */
+  static defineSchema() {
+    const fields = foundry.data.fields;
+    return {
+      description: new fields.SchemaField({
+        full: new fields.StringField(),
+        summary: new fields.StringField()
```
arbron, line 22: Should these be `HTMLField`. I know the behavior isn't any different, but it would communicate a bit better the type of string they contain.

mnoreke:
Not sure if this is the right spot for it, but please add a setting to hide the HP values from players for non-owned items.

There are some DMs (like me) that only provide range data to their players regarding how injured a party member is, so having a spot to look for the actual numbers would mitigate this.

aaclayton:
> There are some DMs (like me) that only provide range data to their players regarding how injured a party member is, so having a spot to look for the actual numbers would mitigate this.

My current plan is that you will only see the health bars of group members in this UI if you have Observer permission or greater to their actor. (which would allow you to see health on their character sheet).

LukasPrism:
Conversely, our table likes seeing each others' hit points, but we don't want full Observer permissions to view each other's sheets and (especially) see with each other's vision/detection modes. But perhaps we're in the minority – just putting it out there.

ghost91-:
I don’t really have any stakes in this, but I find the way the health bars + numbers are laid out in the original screenshot to be quite confusing. It looks like the number in the left would be the minimum value, which of course doesn’t make sense.

If you really want to display the numerical values, something that could work better is to put them in the middle inside health bar itself. Just need to make sure the contrast between font and bar is always good.

mnoreke:
> Conversely, our table likes seeing each others' hit points, but we don't want full Observer permissions to view each other's sheets and (especially) see with each other's vision/detection modes. But perhaps we're in the minority – just putting it out there.
 
 This is why a few settings to control the display of bars/numbers would make sense. :)

module/applications/actor/group-sheet.mjs
aaclayton: What I take away from this suggestion is that I can remove the "units" field from the movement object, since the same speed value works for different distance units according to the DMG rules.

module/applications/actor/group-sheet.mjs
```diff
@@ -0,0 +1,299 @@
...
+        return movementConfig.render(true);
+    }
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Handle clicks to item control buttons on the group sheet.
+   * @param {PointerEvent} event      The initiating click event
+   */
```
arbron, line 232:
```suggestion
  /**
   * Handle clicks to item control buttons on the group sheet.
   * @param {PointerEvent} event      The initiating click event
   * @returns {Promise<Item5e>|ItemSheet5e}
   */
```

module/applications/actor/group-sheet.mjs
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Handle clicks on an item name to expand its description
+   * @param {PointerEvent} event      The initiating click event
+   */
+  #onClickItemName(event) {
+    return game.system.applications.actor.ActorSheet5e.prototype._onItemSummary.call(this, event);
```
arbron, line 286: Maybe too much for this PR, but perhaps this and the drop handling would be improved by descending from a new base actor sheet.

module/applications/actor/group-sheet.mjs
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  async _onDropActor(event, data) {
```
arbron, line 292: Should probably add support for dropping folders of actors, but probably not necessary for initial implementation.

module/applications/actor/movement-config.mjs
```diff
@@ -23,15 +23,36 @@ export default class ActorMovementConfig extends DocumentSheet {
...
+      land: "DND5E.MovementLand",
+      water: "DND5E.MovementWater",
+      air: "DND5E.MovementAir"
+    }: {
+      walk: "DND5E.MovementWalk",
+      burrow: "DND5E.MovementBurrow",
+      climb: "DND5E.MovementClimb",
+      fly: "DND5E.MovementFly",
+      swim: "DND5E.MovementSwim"
     }
```
arbron, line 47:
```suggestion
    } : {
      walk: "DND5E.MovementWalk",
      burrow: "DND5E.MovementBurrow",
      climb: "DND5E.MovementClimb",
      fly: "DND5E.MovementFly",
      swim: "DND5E.MovementSwim"
    };
```

module/data/actor/group.mjs
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+  /* -------------------------------------------- */
+  /*  Data Preparation                            */
+  /* -------------------------------------------- */
+
+  /**
+   * Prepare base data for group actors.
+   * @internal
```
arbron, line 48: `@internal` is a TSDoc code, not JSDoc so our linter isn't a fan.

```suggestion
   * @private
```

```suggestion
   * @protected
```

module/data/actor/group.mjs
```diff
@@ -0,0 +1,108 @@
...
+        if ( a.type === "group" ) console.warn(`Group "${this._id}" may not contain another Group "${a.id}" as a member.`)
+        else this.members.add(a);
+      }
+      else console.warn(`Actor "${id}" in group "${this._id}" does not exist within the World.`);
+    }
+  }
+
+  /**
+   * Prepare derived data for group actors.
+   * @internal
```
arbron, line 64: Same as above

module/data/actor/group.mjs
```diff
@@ -0,0 +1,108 @@
...
+  /*  Data Preparation                            */
+  /* -------------------------------------------- */
+
+  /**
+   * Prepare base data for group actors.
+   * @internal
+   */
+  _prepareBaseData() {
+    this.members.clear();
+    for ( const id of this._source.members) {
```
arbron, line 52:
```suggestion
    for ( const id of this._source.members ) {
```

module/data/actor/group.mjs
```diff
@@ -0,0 +1,108 @@
...
+      const a = game.actors.get(id);
+      if ( a ) {
+        if ( a.type === "group" ) console.warn(`Group "${this._id}" may not contain another Group "${a.id}" as a member.`)
```
arbron, line 55:
```suggestion
        if ( a.type === "group" ) console.warn(`Group "${this._id}" may not contain another Group "${a.id}" as a member.`);
```

module/data/actor/group.mjs
```diff
@@ -0,0 +1,108 @@
...
+  /* -------------------------------------------- */
+
+  /**
+   * Add a new member to the group.
+   * @param {Actor5e} actor           A non-group Actor to add to the group
+   * @returns {Promise<Actor5e>}      The updated group Actor
+   */
+  async addMember(actor) {
+    if ( actor.type === "group" ) throw new Error(`You may not add a group within a group.`);
+    if ( actor.pack ) throw new Error(`You may only add Actors to the group which exist within the World.`);
```
arbron, line 81:
```suggestion
    if ( actor.type === "group" ) throw new Error("You may not add a group within a group.");
    if ( actor.pack ) throw new Error("You may only add Actors to the group which exist within the World.");
```

module/data/actor/group.mjs
```diff
@@ -0,0 +1,108 @@
...
+      }
+    });
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Remove a member from the group.
+   * @returns {Promise<Actor5e>}      The updated group Actor
+   */
```
arbron, line 96:
```suggestion
  /**
   * Remove a member from the group.
   * @param {Actor5e} actor           Actor to remove.
   * @returns {Promise<Actor5e>}      The updated group Actor
   */
```

module/documents/item.mjs
```diff
@@ -1828,7 +1830,7 @@ export default class Item5e extends Item {
 
     // Include an ability score modifier if one exists
     const abl = this.abilityMod;
```
arbron, line 1832: `Item5e#abilityMod` is also throwing an error.

Line 35:
```javascript
      const abilities = this.actor.system.abilities ?? {};
```

Line 50:
```javascript
            return abilities.dex?.mod >= abilities.str?.mod ? "dex" : "str";
```

arbron (changes requested):

arbron:
Not necessary for this initial version, but something to add to the to-do list later is the ability to display the aggregate group HP as a token bar.

krbz999:
Two questions:
- Was `consumable` left out intentionally?
- What happens if you attempt to add an unlinked actor as a member?

module/applications/actor/group-sheet.mjs
```diff
@@ -0,0 +1,299 @@
...
+  static get defaultOptions() {
+    return foundry.utils.mergeObject(super.defaultOptions, {
+      classes: ["dnd5e", "sheet", "actor", "group"],
+      template: "systems/dnd5e/templates/actors/group-sheet.hbs",
+      tabs: [{navSelector: ".tabs", contentSelector: ".sheet-body", initial: "members"}],
+      width: 620,
+      height: 620
+    });
+  }
+
```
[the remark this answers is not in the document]
aaclayton: It's awkward. Inheriting from ActorSheet5e would require overriding and removing almost all methods, but it does miss some things that otherwise we do want. It's an example that we probably need to re-rationalize what logic is in the base actor sheet and what logic is in type-specific sheets.

module/applications/actor/group-sheet.mjs
```diff
@@ -0,0 +1,299 @@
...
+    // Item summaries
+    html.find(".item .rollable h4").click(event => this.#onClickItemName(event));
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Handle clicks to action buttons on the group sheet.
+   * @param {PointerEvent} event      The initiating click event
+   */
```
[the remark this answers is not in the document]
aaclayton: This method isn't intended to have a meaningful return value, doing:
```js
return this.object.system.removeMember(removeMemberId);
```
Is just a more code-frugal approach than:
```js
this.object.system.removeMember(removeMemberId);
break;
```
If we want to be strict about our return types I will remove the return statements from the method so that it returns void.

module/applications/actor/group-sheet.mjs
aaclayton: Same as above.

module/applications/actor/group-sheet.mjs
aaclayton: That feels well out of scope for this PR and should be handled as a separate set of changes. I do, however, agree.

module/applications/actor/group-sheet.mjs
aaclayton: I think we need to limit scope. It's part of the reason we have so many un-merged PRs in this repo is because we struggle to make changes small enough that they can be reviewed and merged iteratively. We have to keep everything from becoming a thousand LOC change otherwise we never get anywhere.

module/data/actor/group.mjs
aaclayton: We should change our linter then, because as of V10 our docstring syntax that we use is typedoc. We should use `@internal` tags for methods which are not intended for the public API but do need to be called outside of the scope in which they are defined.

aaclayton:
> Was `consumable` left out intentionally?

Eh, it was intentional, but I see the justification for having a separate header section for consumables so I have added it in.

> What happens if you attempt to add an unlinked actor as a member?

There's not really a UX which allows this to be done out of the box, since there's no way for you to be dragging an unlinked actor in order to drop it on the sheet. Group members refer to base world actors, you cannot have an unlinked actor in the group.

krbz999:
> > Was `consumable` left out intentionally?
>
> Eh, it was intentional, but I see the justification for having a separate header section for consumables so I have added it in.

My question was more, why the difference between groupings of items between character sheets and group sheets (I actually forgot `tool` existed when I asked).

> > What happens if you attempt to add an unlinked actor as a member?
>
> There's not really a UX which allows this to be done out of the box, since there's no way for you to be dragging an unlinked actor in order to drop it on the sheet. Group members refer to base world actors, you cannot have an unlinked actor in the group.

Makes sense. Could this actor type down the line be used to group groups of enemies during combat? I could see an immediate use case for this, which also aligns with how DND5E initiative is technically _proposed_ to work (each NPC type being grouped with the same initiative).

Lastly (and leastly), "__A group__ of 2 members and 1 vehicles" I personally feel would look better as simply "Group of 2 members and 1 vehicles".

templates/actors/group-sheet.hbs
```diff
@@ -0,0 +1,93 @@
...
+                        </div>
+                    </li>
+                    {{/each}}
+                </ol>
+            </section>
+            {{/each}}
+        </div>
+
+        <div class="tab inventory" data-group="primary" data-tab="inventory">
+            {{> "dnd5e.actor-inventory" sections=inventory}}
```
arbron, line 84: For some reason the inventory isn't scrollable if the number of items extends beyond the bottom of the sheet.

arbron (changes requested):

module/data/actor/group.mjs
```diff
@@ -0,0 +1,111 @@
...
+        })
+      }),
+      currency: new fields.SchemaField({
+        pp: new fields.NumberField({nullable: false, initial: 0, integer: true, min: 0}),
+        gp: new fields.NumberField({nullable: false, initial: 0, integer: true, min: 0}),
+        ep: new fields.NumberField({nullable: false, initial: 0, integer: true, min: 0}),
+        sp: new fields.NumberField({nullable: false, initial: 0, integer: true, min: 0}),
+        cp: new fields.NumberField({nullable: false, initial: 0, integer: true, min: 0})
+      })
+    }
```
Fyorl, line 39:
```suggestion
    };
```

module/data/actor/group.mjs
```diff
@@ -0,0 +1,111 @@
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Remove a member from the group.
+   * @param {Actor5e} actor           An Actor to remove from this group
+   * @returns {Promise<Actor5e>}      The updated group Actor
+   */
+  async removeMember(actor) {
+    const memberIds = Array.from(this._source.members);
```
Fyorl, line 101: `this._source.members` should already be an array I believe.

module/data/actor/group.mjs
```diff
@@ -0,0 +1,111 @@
...
+    const actorId = actor instanceof Actor ? actor.id : actor;
```
Fyorl, line 102: Seems to suggest this method will also take a string ID as well as an `Actor5e` instance, do we want to add that to the parameter annotation?

templates/apps/movement-config.hbs
```diff
@@ -1,34 +1,24 @@
...
-    </div>
-    <div class="form-group">
-        <label>{{localize "DND5E.MovementFly"}}</label>
-        <input name="system.attributes.movement.fly" type="number" step="0.1" value="{{movement.fly}}"/>
-    </div>
-    <div class="form-group">
-        <label>{{localize "DND5E.MovementSwim"}}</label>
-        <input name="system.attributes.movement.swim" type="number" step="0.1" value="{{movement.swim}}"/>
+        <label>{{localize label}}</label>
+        {{numberInput (lookup @root.movement id) name=(concat "system.attributes.movement." id) placeholder=0 step=0.1}}
```
Fyorl, line 6:
```suggestion
        {{numberInput (lookup @root.movement id) name=(concat "system.attributes.movement." id) placeholder=0 step=0.1 min=0}}
```

templates/actors/group-sheet.hbs
```diff
@@ -0,0 +1,93 @@
...
+    </nav>
+
+    <section class="sheet-body">
+        <div class="tab members" data-group="primary" data-tab="members">
+            {{#each sections as |section id|}}
+            <section class="group-members directory" data-section="{{id}}">
+                <header class="directory-header flexrow">
+                    <h3 class="name">{{section.label}}</h3>
+                    <span class="hp">Hit Points</span>
+                    <span class="controls">Controls</span>
```
Fyorl, line 54: Missing localisation.

Fyorl (changes requested):
Looking really nice. Only some minor suggestions from me.

Fyorl:
@krbz999 
> Edit: With the addition of 'Space vehicle' as a proficiency, does 'Space' make sense as a movement type in addition to Land, Water, Air?

Spelljammer ships have a fly speed and a 100-million-miles-every-24-hours speed. They don't really have a concept of a 'space' speed, unless you count the latter, but that's a fixed value that doesn't change.

module/applications/actor/group-sheet.mjs
```diff
@@ -0,0 +1,308 @@
...
+    let speeds = [
+      [movement.land, `${game.i18n.localize("DND5E.MovementLand")} ${movement.land}`],
+      [movement.water, `${game.i18n.localize("DND5E.MovementWater")} ${movement.water}`],
+      [movement.air, `${game.i18n.localize("DND5E.MovementAir")} ${movement.air}`]
+    ];
+    speeds = speeds.filter(s => !!s[0]).sort((a, b) => b[0] - a[0]);
+    const primary = speeds.shift();
+    return {
+      primary: `${primary ? primary[1] : "0"}`,
+      secondary: speeds.map(s => s[1]).join(", ")
```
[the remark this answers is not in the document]
aaclayton: I think that one is overkill, it doesn't actually preserve the commas which I think is ideal.

module/applications/actor/group-sheet.mjs
```diff
@@ -0,0 +1,308 @@
...
+  activateListeners(html) {
+    super.activateListeners(html);
+    html.find(".group-member .name").click(this.#onClickMemberName.bind(this));
+
+    // Action buttons
+    html.find(".action-button").click(this.#onClickActionButton.bind(this));
+    html.find(".item-control").click(this.#onClickItemControl.bind(this));
+
+    // Item summaries
+    html.find(".item .rollable h4").click(event => this.#onClickItemName(event));
```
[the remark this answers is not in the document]
aaclayton: Reluctant agreement.

module/data/actor/group.mjs
aaclayton: The purpose here was to clone the source field so I can mutate it with `findSplice` but that change is not applied until I call the `.update` function. I will use `foundry.utils.deepClone()` here instead to make it more readable that is what I'm doing.

aaclayton:
Thanks for the feedback @Fyorl - updated the PR with suggested changes.

aaclayton:
Something I noticed while handling some localization changes is that we had duplicative definition of item types and their pluralizations. The standard Item document types were defined as `ITEM.Type{type}` while we had duplicative localizations of the form `DND5E.ItemType{type}`. I have gotten rid of the second format and moved pluralized item type labels to be adjacent to the former style as `ITEM.Type{type}Pl`.

I can split this commit over to a separate PR if it's controversial, but I don't think it should be.

arbron (approved):

Fyorl (approved):

Fyorl:
@aaclayton 

> I can split this commit over to a separate PR if it's controversial, but I don't think it should be.

No objections here.

airbreather:
> Something I noticed while handling some localization changes is that we had duplicative definition of item types and their pluralizations. The standard Item document types were defined as `ITEM.Type{type}` while we had duplicative localizations of the form `DND5E.ItemType{type}`. I have gotten rid of the second format and moved pluralized item type labels to be adjacent to the former style as `ITEM.Type{type}Pl`.
> 
> See: [1aa7a7a](https://github.com/foundryvtt/dnd5e/commit/1aa7a7a1612557a699274797cb16cb952ad8efe8)
> 
> I can split this commit over to a separate PR if it's controversial, but I don't think it should be.

FYI: p4535992/foundryvtt-inventory-plus#28

*I'm just cross-referencing the issue to add visibility to this downstream impact, not claiming that it was a mistake.  I appreciate the need to cut back unnecessary work for people doing translations :+1:.*

PR #1953, opened 2022-12-02
[#1835] Create system data model for containers (aka backpacks)

arbron:
I named it `ContainerData` in expectation of an eventual renaming of the `backpack` item type to `container` (#1622). If that seems too confusing, I can rename it to `BackpackData` and we can change it later.

Also, I moved the currency data to its own template so it can eventually contain currency conversion methods so they are available on containers in addition to actors. More work can be done on that once #1829 is integrated.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -557,6 +557,7 @@
 "DND5E.ItemConsumableType": "Consumable Type",
 "DND5E.ItemContainerCapacity": "Capacity",
 "DND5E.ItemContainerCapacityItems": "Items",
+"DND5E.ItemContainerCapacityMax": "Max Capacity",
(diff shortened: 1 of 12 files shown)


Fyorl (approved):
It's not strictly required but let's add the `currency` fields to the `template.json` under backpack to keep them in sync with the model for now. Otherwise this looks good to me, feel free to merge when ready.

arbron:
> It's not strictly required but let's add the `currency` fields to the `template.json` under backpack to keep them in sync with the model for now. Otherwise this looks good to me, feel free to merge when ready.

They are already there, that is why I added them to the data model :).

PR #1954, opened 2022-12-02
[#1835] Create system data model for tools

arbron:


module/data/item/tool.mjs
```diff
@@ -0,0 +1,34 @@
...
+  static defineSchema() {
+    return this.mergeSchema(super.defineSchema(), {
+      toolType: new foundry.data.fields.StringField({required: true, label: "DND5E.ItemToolType"}),
+      baseItem: new foundry.data.fields.StringField({required: true, label: "DND5E.ItemToolBase"}),
+      ability: new foundry.data.fields.StringField({
+        required: true, initial: "int", blank: false, label: "DND5E.DefaultAbilityCheck"
+      }),
+      chatFlavor: new foundry.data.fields.StringField({required: true, label: "DND5E.ChatFlavor"}),
+      proficient: new foundry.data.fields.NumberField({
+        required: true, nullable: false, integer: true, initial: 0, min: 0, label: "DND5E.ItemToolProficiency"
```
Fyorl, line 29: Half proficiency is represented as `0.5` so this can't be `integer: true`.

Fyorl (changes requested):

Fyorl (approved):

Issue #1956, opened 2022-12-04
Roll damage with flavour


TPNils:
I have been looking into making an automation module where I want to take damage types into consideration. I wanted to reuse Item5e.rollDamage() but this seems to [omit the flavour texts](https://github.com/foundryvtt/dnd5e/blob/f2c7c99193f5c00d327e9fb515b3759919b86e91/module/documents/item.mjs#L1400), would it be possible to add those to the roll?

RossGosling:
This seems like it may duplicate #1818 ?

TPNils:
It's somewhat similar yes, though I would like it to also work when passing the option "without generating chat message"

Fyorl:
Leaving this open to make sure that we implement #1818 such that damage parts are split into their own separate `Roll` instances (with flavour) at the `damageRoll` level, and not just at the level at which they are converted to the chat card.

arbron:
Not sure when this got fixed, but it isn't a problem anymore

PR #1957, opened 2022-12-04
[#1956] fix: include roll flavour in damage rolls

TPNils:
Suggested for for issue #1956

> I have been looking into making an automation module where I want to take damage types into consideration. I wanted to reuse Item5e.rollDamage() but this seems to [omit the flavour texts](https://github.com/foundryvtt/dnd5e/blob/f2c7c99193f5c00d327e9fb515b3759919b86e91/module/documents/item.mjs#L1400), would it be possible to add those to the roll?

I wasn't 100% sure on what the best practice is on merging flavors, but currently flavors seem to be appended so I have copied that behaviour. Example, if I have the damage formula `1d8[if uninjured]` with damage type `fire` it would become `1d8[if uninjured fire]`. An alternative result I could think of (but did not implement) would be `1d8[fire: if uninjured]`

This would be a rather unique case, in 99% of the cases the flavor is simply appended. Example: damage formula `8d6` with damage type `fire` it would become `8d6[fire]`.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1397,9 +1397,28 @@ export default class Item5e extends Item {
 
     // Get roll data
     const dmg = this.system.damage;
-    const parts = dmg.parts.map(d => d[0]);
+    const dmgParts = deepClone(dmg.parts)
     const rollData = this.getRollData();
     if ( spellLevel ) rollData.item.level = spellLevel;
+
+    // Adjust damage from versatile usage
+    if ( versatile && dmg.versatile ) {
+      parts[0][0] = dmg.versatile;
+      messageData["flags.dnd5e.roll"].versatile = true;
+    }
+
+    // Convert damage parts to formula parts
+    const parts = dmgParts.map(d => {
+      if (!d[1]) {
+        return d[0];
+      } else {
+        const endsWithFlavor = /\[([^[\]]*)]$/.exec(d[0]);
+        if (endsWithFlavor) {
+          return `${d[0].substring(0, d[0].length - endsWithFlavor[0].length)}[${endsWithFlavor[1]} ${d[1]}]`;
+        }
+        return `${d[0]}[${d[1]}]`;
+      }
+    });
 
     // Configure the damage roll
     const actionFlavor = game.i18n.localize(this.system.actionType === "heal" ? "DND5E.Healing" : "DND5E.DamageRoll");
@@ -1420,12 +1439,6 @@ export default class Item5e extends Item {
       },
       messageData
     };
-
-    // Adjust damage from versatile usage
-    if ( versatile && dmg.versatile ) {
-      parts[0] = dmg.versatile;
-      messageData["flags.dnd5e.roll"].versatile = true;
-    }
 
     // Scale damage from up-casting spells
     const scaling = this.system.scaling;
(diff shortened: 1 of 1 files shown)


krbz999:
Is it possible to instead put the flavor somewhere in the internal rolls' data or in the damage roll messages' flags, rather than changing the visuals of every roll for all users?

TPNils:
Would this not be an improvement _as is_ regarding Dice So Nice?

If I use the roll command `/roll 1d6[bludgeoning] + 1d8[fire] + 1d10[force]` I get the associated dice for their damage types, but if I make this configuration for an item damage roll it does not.

If this is too much of a breaking change I have something in mind to accommodate but I want to check if the current suggestion might be fine.

trioderegion:
What about complex damage formulas like `1d8 + 1d(@item.level)` ?

As @krbz999 mentioned, hooking on damage rolls in your module and adding extra information to the message/roll's flags would be the less invasive option.

TPNils:
Hm.. I see what you mean, there seems to be a bug in DamageRoll that deletes flavor texts for ParentheticalTerm. That would need to be fixed in order for this to work.
image

In a situation where DamageRoll evaluates without removing flavor, I would expect the result to be `1d8 + 1d(@item.level)[force]` (or whatever damage type it should be).

However, how would you suggest I hook into it externally? Let us create a fictional item with 2 damage parts
- 8d6 (fire)
- 1d8 + 1d3 (force)

Today this evaluates to `8d6 + 1d8 + 1d3` which is not accounting for actor bonuses, ammo, spell scaling and other modules editing the roll like the new Build-a-bonus. I would then somehow need to reverse link the resulting roll to the original damage types which I don't really see how I could do that.

For me I see 2 clear solutions to solve this problem.
1. It would be added to the dnd5e damage roll, but I am feeling that this will not be the case.
2. Before I call the damage roll, I can temporary edit the data of the item and add the damage type as a flavor and revert after the roll. 

I would like to avoid option 2 as it does feel quite hacky, though it would get the job done but can break with any new version release. I could also just copy paste the roll damage method. While this would make it "stable", that would mean my module would never be able to enjoy the improvements added from the dnd5e system until I notice it and also update so I do not consider this an option.

Allow me to present a more complex scenario. I can fairly easily detect the damage types, while I wouldn't have a clue on how do detect this without the flavor.
- I have a magical mace which deals 1d6 + @mod bludgeoning damage & flat 3 (no damage type) & 1d8 fire damage
- I have an actor bonus of 1d2
- I have used smite at lvl1 which is added through a hook in dnd5e.preRollDamage => 2d8 radiant damage
- I have added a situational bonus of 1d12

The result would be the following (assuming the smite bonus uses flavor)
image

krbz999:
> However, how would you suggest I hook into it externally?

`Hooks.on("dnd5e.rollDamage", (item, roll) => { /* do things */ });`

stwlam:
I think there is an incorrect assumption about operator precedence at work in the above cases. With `1d8 + 1d(@item.level)[force]`, the flavor is part of the `ParentheticalTerm` rather than the ultimately resolved `Die`. The dnd5e system simplifies the above by breaking out the `faces` number from the intermediate parenthetical, and the flavor is tossed along with it. If you want to associate the "force" flavor with the second `Die` expression, it should be `1d@item.level[force]`, `(1d@item.level)[force]`, or even `(1d(@item.level))[force]`.

Issue #1237, opened 2021-08-12
Add weight units to vehicles


arbron:
Vehicle cargo capacity is measured in tons but that isn't indicated anywhere on the sheet making it confusing when adding items and having the weights not seem to add up. An indicator of the units on the encumbrance bar and next to the `"Cargo Capacity"` field would help resolve this confusion.

erikhuck:
@arbron agreed. The units should be clearly labeled and we should be able to customize the units of measurement. Tons as the default unit for dnd5e doesn't make much sense since almost every mount carrying capacity is some number of pounds.

Ackerfe:
I know this isn't a fix to this request. But from what I can see the weight is defined in short tons (2000 pounds = 1 ton) and while it's true that the capacity for land vehicles is calculated by multiplying the carrying capacity of the pulling animal by 5 and you can't put non-integers in the capacity field on the sheet, what you CAN do is set up the capacity using effects on the sheet, so just go in the effects tab of the sheet and create a new on targeting system.attributes.capacity.cargo and set that to override and give it whatever value you need, just remember to take the calculated pulling capacity of your mule/horse/elephant/etc.. and divide it by 2000 so that you get proper weight calculation in the sheet

PR #1959, opened 2022-12-04
[#1909] Add units to item weight & properly convert for calculating encumbrance

### LINKED ISSUE #1909: Add units to item weight
Currently all items are assumed to be weighed in pounds. This makes it difficult to represent large vehicle equipment that might weigh tons or to support a metric mode in the system. I propose converting weight into an object with a `units` parameter:

```javascript
weight: {
  value: 5,
  units: "lbs"
}
```

This would also require defining different weight units on config:

```javascript
DND5E.weightUnits = {
  lb: {
    label: "Pounds",
    abbreviation: "lb",
    conversion: 1
  },
  tn: {
    label: "Tons",
    abbreviation: "tn",
    conversion: 2000
  },
  kg: {
    label: "Kilograms",
    abbreviation: "kg",
    conversion: 2.2046
  },
  t: {
    label: "Tonnes",
    abbreviation: "t",
    conversion: 2204.6
  }
};
```

Calculating weight for character sheets would be a simple matter of multiplying the weight `value` by the conversion multiplier specified its unit's config.

arbron:
Item weights now define a specific unit which will make it possible to have a mixture of imperial and metric object weights, as well as represent large items in vehicle inventories. Defines these weights along with a conversion method in `CONFIG.weightUnits`.

Closes #1909

### To Do
- [ ] Convert SRD items

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1817,6 +1817,25 @@
 "DND5E.WeaponSimpleProficiency": "Simple",
 "DND5E.WeaponSimpleR": "Simple Ranged",
 "DND5E.Weight": "Weight",
+"DND5E.WeightUnit": {
+  "Label": "Weight Units",
+  "Kilograms": {
+    "Label": "Kilograms",
+    "Abbreviation": "kg"
+  },
+  "Megagrams": {
+    "Label": "Tonnes",
+    "Abbreviation": "t"
+  },
+  "Pounds": {
+    "Label": "Pounds",
+    "Abbreviation": "lb"
+  },
+  "Tons": {
+    "Label": "Tons",
+    "Abbreviation": "tn"
+  }
+},
 "DND5E.WhisperedTo": "Whispered to",
 "DND5E.Wiki": "Wiki",
 "DND5E.available": "available",
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -1773,7 +1773,7 @@ DND5E.weightUnits = {
     conversion: 2.2046,
     type: "metric"
   },
-  mg: {
+  Mg: {
     label: "DND5E.WeightUnit.Megagrams.Label",
     abbreviation: "DND5E.WeightUnit.Megagrams.Abbreviation",
     conversion: 2204.6,
@@ -1857,7 +1857,7 @@ DND5E.encumbrance = {
     },
     vehicle: {
       imperial: "tn",
-      metric: "mg"
+      metric: "Mg"
     }
   }
 };
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -1770,13 +1770,13 @@ DND5E.weightUnits = {
   kg: {
     label: "DND5E.WeightUnit.Kilograms.Label",
     abbreviation: "DND5E.WeightUnit.Kilograms.Abbreviation",
-    conversion: 2.2046,
+    conversion: 2.5,
     type: "metric"
   },
   Mg: {
     label: "DND5E.WeightUnit.Megagrams.Label",
     abbreviation: "DND5E.WeightUnit.Megagrams.Abbreviation",
-    conversion: 2204.6,
+    conversion: 2500,
     type: "metric"
   }
 };
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -1741,17 +1741,62 @@ preLocalize("distanceUnits");
 /* -------------------------------------------- */
 
 /**
+ * Configuration data for a weight unit.
+ *
+ * @typedef {object} WeightUnitConfiguration
+ * @property {string} label         Localized label for the unit.
+ * @property {string} abbreviation  Localized abbreviation for the unit.
+ * @property {number} conversion    Number that by which this unit should be multiplied to arrive at a standard value.
+ * @property {string} type          Whether this is an "imperial" or "metric" unit.
+ */
+
+/**
+ * The valid units for measurement of weight.
+ * @enum {WeightUnitConfiguration}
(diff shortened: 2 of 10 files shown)


p4535992:
@arbron do you think the same concept can make sense for meters and feet as well ?

erikhuck:
fixes #1237

erikhuck:
@arbron Does this PR enable changing the units of measurement of vehicles? I don't think having tons as the only unit of measurement really makes sense for dnd5e vehicles.

Fyorl:
Going to try to get this and #3415 in 3.2 as I understand the encumbrance levels are causing some disruption for metric users. If I understand the gist of this PR, it's similar to how we added units to currency, but for weight now too?

The new sheets at the moment simply use an icon to represent weight, rather than a unit, and I'd prefer to keep it that way as I think it's a bit easier to parse visually. Is it reasonable to assume users won't be mixing and matching items with different weights on their sheets? Or is this likely to become a point of confusion?

krbz999:
Mixing lb and kg would be strange to do (in my opinion). But if it ends up being allowed, a simple tooltip could be added on the weight of the item on the actor sheet.

Fyorl:
Yeah, a tooltip would be no problem

lang/en.json
```diff
@@ -1817,6 +1817,25 @@
...
 "DND5E.WeaponSimpleR": "Simple Ranged",
 "DND5E.Weight": "Weight",
+"DND5E.WeightUnit": {
+  "Label": "Weight Units",
+  "Kilograms": {
+    "Label": "Kilograms",
+    "Abbreviation": "kg"
+  },
+  "Megagrams": {
+    "Label": "Tonnes",
```
krbz999, line 1827: What's wrong with "Megagram" and "mg"?

lang/en.json
```diff
@@ -1817,6 +1817,25 @@
 "DND5E.WeaponSimpleProficiency": "Simple",
 "DND5E.WeaponSimpleR": "Simple Ranged",
 "DND5E.Weight": "Weight",
+"DND5E.WeightUnit": {
+  "Label": "Weight Units",
+  "Kilograms": {
```
krbz999, line 1822: "Kilograms" and "Megagrams" are not actual words. The plural forms are "Kilogram" and "Megagram".

module/documents/actor/actor.mjs
```diff
@@ -501,18 +501,20 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
   _prepareEncumbrance() {
     const config = CONFIG.DND5E.encumbrance;
     const encumbrance = this.system.attributes.encumbrance ??= {};
-    const units = game.settings.get("dnd5e", "metricWeightUnits") ? "metric" : "imperial";
+    const baseUnits = CONFIG.DND5E.encumbrance.baseUnits[this.type] ?? CONFIG.DND5E.encumbrance.baseUnits.default;
+    const unitSystem = game.settings.get("dnd5e", "metricWeightUnits") ? "metic" : "imperial";
```
krbz999, line 505:
```diff
- const unitSystem = game.settings.get("dnd5e", "metricWeightUnits") ? "metic" : "imperial";
+ const unitSystem = game.settings.get("dnd5e", "metricWeightUnits") ? "metric" : "imperial";
```

module/config.mjs
```diff
@@ -1805,9 +1850,15 @@ DND5E.encumbrance = {
...
-    imperial: 2000, // 2000 lbs in an imperial ton
-    metric: 1000 // 1000 kg in a metric ton
+  baseUnits: {
+    default: {
+      imperial: "lb",
+      metric: "kg"
+    },
+    vehicle: {
+      imperial: "tn",
+      metric: "mg"
```
krbz999, line 1860: Using "mg" here so presumably just a mistake up above.

lang/en.json
thatlonelybugbear: > "Kilograms" and "Megagrams" are not actual words. The plural forms are "Kilogram" and "Megagram".

I don't believe that this is correct. Kilograms and Megagrams are totally valid plural forms 👍

lang/en.json
arbron: I thought `Tonnes` was the more generally used term, but I'd welcome feedback from someone in a metric country :)

module/config.mjs
arbron: What mistake are you referring to?

krbz999: Using "t" as the abbreviation for Megagrams in the localization.

arbron: This is the static `key`, not the localization value, so it references the key in `CONFIG.weightUnits`.

lang/en.json
krbz999: I am, lol, but here's another to make a fool of me.

GamerFlix:
Just to comment on the megagrams point and suggested alternative:

According to convention (compare: https://en.wikipedia.org/wiki/Metric_prefix) `mg` would stand for miligrams while megagrams (which I have never heard uses before) would be `Mg` (note the case). I would also like to note that tons vs tonnes for metric vs imperial seems rather confusing to me when it could simply be Imperial tons vs Metric tons for perfect clarity. The latter point is especially noteworthy for people for whom English is not their first language and therefore aren't familiar with the spelling difference's meaning.

Fyorl (approved):
Looks good, thanks. Please see my comments but let's get this merged and get some testing in.

Issue #461, opened 2020-04-24
Allow for an item to have multiple action profiles for damage rolls, saving throws, etc...


aaclayton:
Originally in GitLab by @DarKDinDoN

## Original Suggestion

Hi, It would be nice if we can add more than one save roll to a feat or any item.
Example: the monk "open hand technique":

```
Open Hand Technique
Starting when you choose this tradition at 3rd level, you can manipulate your enemy’s ki when you harness your own. Whenever you hit a creature with one of the attacks granted by your Flurry of Blows, you can impose one of the following effects on that target: 

* It must succeed on a Dexterity saving throw or be knocked prone.
* It must make a Strength saving throw. If it fails, you can push it up to 15 feet away from you.
* It can’t take reactions until the end of your next turn.
```

## [Feature Request] Break Items separate actions into parts
From @jeremyregnerus

I understand that this is both probably complicated, and hard to describe, so I'll do my best here.
It would be really useful for more complex items to have *parts* to their actions section that can be added individually to provide all of the actions an item provides in one place as opposed to having having just one action type that can be added to an item. This is sort of like the ability to add multiple damage can add multiple lines, just more broken up.

Let's use an example of a versatile weapon that does 1d6 poison damage on a failed dc save, has a once per day ability as an action to cure poison, and for complexities sake, let's also say on a critical strike this weapon slips the dc save and automatically does the poison damage. And for icing on top, I use maximum weapon damage, roll critical dice in my game. This is all just theoretical but scenarios like these come into play with a lot of wondrous items and complex weapons.

Today, there is one action, by default the attack roll. The first issue is this attack roll will try consume the daily use of the weapon. The user can deselect the checkbox to use this, but they would have to do this on every attack. They would also be presented with warnings when trying to make an attack after this daily use has been consumed.

[...]

PR #1963, opened 2022-12-14
[#461] Support multiple item usage profiles

RossGosling:
## #461 Allow for an item to have multiple action profiles for damage rolls, saving throws, etc...

The goal here is to implement a new system of item 'Usage-Profiles', these are an array of activations an item can have and use independently of one another

Use-cases:

- Shortbow with different usage-profiles for ammo, so each can be configured to consume different ammunition
- Any weapon can be configured with Great Weapon Master to simply apply the attack roll and damage alterations
- Hexblade's Pact Weapon with a different usage-profile for versus normal or cursed targets (When crit threshold is lowered)
- Spells like Arcane Hand with a different usage-profile for the 4 distinct attack types: Clenched Fist, Forceful Hand, Grasping Hand, and Interposing Hand
- Features like Open Hand Technique with two usage-profiles for knocking an enemy prone, or pushing them 15ft

----

### Support Item Types

- [X] Weapon
- [X] Spell
- [X] Feature
- [X] Equipment
- [X] Consumable

### User Stories derived from @schultzcole - https://github.com/foundryvtt/dnd5e/issues/461#issuecomment-1154123067

- [x] I would like to be able to navigate between usage-profiles when editing an item
- [x] I would like to be able to add usage-profiles when editing an item
- [x] I would like to be able to delete usage-profiles when editing an item
- [x] I would like to be able to duplicate usage-profiles when editing an item
- [x] I would like to be able to select usage-profiles when rolling an item
- [x] I would like to be able to see which usage-profile was used in the chat item-card
- [x] I would like to be able to see which usage-profile was used in the chat attack-card
- [x] I would like to be able to see which usage-profile was used in the chat damage-card
- [x] I would like to be able to select usage-profiles when rolling via a macro

### Programmatic User Stories

- [ ] I would like to be able to select a usage-profile when calling item.use()
- [ ] I would like to be able to select a default usage-profile
- [ ] I would like to be able to add usage-profiles
- [ ] I would like to be able to delete usage-profiles

### Technical Tasks

- [ ] Fix item update issues after rebasing to 2.1.x
- [ ] Work out the most appropriate place to store profile-selection data within the ItemSheet5e
- [ ] Likely migrate to an object-based storage structure rather than arrays
- [ ] The system should migration data for pack items and user data to the new format

### Additional Feature Goals

- [ ] I would like to be able to alter the order of usage-profiles when editing an item

----

### Visual Design In-Progress

[Video Demo](https://youtu.be/GMV49vro8xE)

| Profiles |
| -------- |
| image |

| Selecting Profiles |
| -------- |
| image |

| Chat With Profile |
| -------- |
| image |


Varriount:
Ooo, I like this. Would it be possible to use this with spells as well? Lots of spells have different "profiles" to change how they are used (e.g., [Toll the Dead](https://www.dndbeyond.com/spells/toll-the-dead).

RossGosling:
> Ooo, I like this. It's especially useful for spells (e.g., [Arcane Hand](https://www.dndbeyond.com/spells/arcane-hand)).

Ah thank you, that's an amazing example - Been working on Spells too yep, just hadn't got the best example to work from

Varriount:
From a UI perspective, how do you propose users activate/use a specific profile? Will they be able to set a "default" profile?

RossGosling:
> From a UI perspective, how do you propose users activate/use a specific profile? Will users be able to set a "default" profile?

The approach I'm working to right now is that when users roll a Longsword they will have an additional dialog to pick first, there's not defaulting or anything

RossGosling:
I've uploaded a video of my current implementation to demonstrate
[https://youtu.be/GMV49vro8xE](https://youtu.be/GMV49vro8xE)

RossGosling:
I think there's too many issues I've encountered to keep up with this, I think I found so many UX implications I'd never considered
- The Item Properties, and display details within the Inventory become a problem when an item has many profiles with many charges, etc
I've got so many issues around not being able to update item data and I've just been going in circles

Varriount:
@RossGosling What about displaying each profile as a separate entry in the user's inventory?

In a similar vein, an alternative I've considered to this is a module that transparently groups items together:
  - Grouped items would be represented as a virtual "composite" item in compendiums and the "items" sidebar.
  - When a composite item is added to an actor's inventory, the module would automatically extract the embedded items from the composite item, and then delete the composite item.
  - If one of the composite items is deleted, then the other items in the group would be deleted.

RossGosling:
@Varriount that sounds like a great approach, Staff seem to have another thought around their Advancement system, not sure the specifics but it's laid out in #1964

PR #1965, opened 2022-12-16
[#1835] Apply refresh to all items in compendiums

arbron:
This uses the `dnd5e.migrations.refreshAllCompendiums()` method to force all items in our compendiums to reflect the current data models. This should result in the removal of all long-deprecated fields and the addition of any missing fields with their default values.


Fyorl (approved):
Nice work, this seems to have fixed the validation errors I was encountering whenever turning on `_enableV10Validation` in our models. It's good to know that was just caused by old data, rather than an issue with our models. We probably still need a solution for that where world data is concerned, but let's get the compendium data in shape first.

Issue #1970, opened 2022-12-19
Item Context Menu with option to duplicate/clone items, features, and spells


RossGosling:
Pull Request #1971

---

Provides a new Context Menu on all items, features, spells, and effects across Characters, NPCs, Groups, and Vehicles.

This includes a tailored set of options with a new Duplicate, which simply clones your item, feature, spell, or effect.

Duplicate is not available for Backgrounds, Class, or Subclass after to some discussion in Discord.

---

<details><summary><b>Inventory Screenshots</b></summary>

Lacking the equip icon due to #1972 

  | Inventory (Character Sheet) |
  | --- |
  | ![image](https://user-images.githubusercontent.com/879127/209038209-116d4c3f-f4ea-4ae4-8be8-981830b6e0ec.png) |
  | ![image](https://user-images.githubusercontent.com/879127/209038319-58a30d76-ad19-4545-9d29-4b987f9ef669.png) |

</details>
<details><summary><b>Spellbook Screenshots</b></summary>

  | Spellbook (Character Sheet) |
  | --- |
  | ![image](https://user-images.githubusercontent.com/879127/209021425-9eaaa616-0345-4fb4-9032-6c0e4f044151.png) |
  | ![image](https://user-images.githubusercontent.com/879127/209021462-39ef180d-f44a-4470-b5ac-bfa225eb8d57.png) |
  
</details>
<details><summary><b>Feature Screenshots</b></summary>

  | Features (Character Sheet) |
  | --- |
  | ![image](https://user-images.githubusercontent.com/879127/209021544-a48e5025-8a03-4fcc-8e7c-5780f70d2811.png) |
  | ![image](https://user-images.githubusercontent.com/879127/209038397-410ba5a1-77cf-4808-af3c-555d195a580d.png) |

</details>
<details><summary><b>Effect Screenshots</b></summary>

  | Effects (Character Sheet) |
  | --- |
  | ![image](https://user-images.githubusercontent.com/879127/209038447-e4dea825-a694-4280-9761-33d260195543.png) |

</details>

arbron:
While duplication functionality would certainly be helpful in certain situations, I don't think it warrants a whole new button to further clutter up the UI. This might be a good place to begin adding context menus to the inventory list. That is how I added the ability to duplicate advancements into the system: https://github.com/foundryvtt/dnd5e/pull/1588

RossGosling:
Makes a lot of sense - I'll see about setting that up when I get a chance

[...]

PR #1971, opened 2022-12-19
[#1970] implement item context menus within Actor/NPC/Group/Vehicle

RossGosling:
Details in Issue #1970


arbron:
I’ll give a more extensive review in a few weeks, but on quick review it would probably be a good idea to use variables in the localizations. So instead of “Edit Effect”, “Edit Feature”, and “Edit Item” it would be “Edit {type}”. Then you’d just have to pre-localize the context menus using `game.i18n.format(“DND5E.ContextMenuActionEdit”, {type: game.i18n.localize(“DND5E.Effect”)});`

Varriount:
I think a better place to put such an item's context menu would be to the right of the item's entry, where all the other item manipulation buttons are. For example, using the ubiquitous "more options" icon:
image

RossGosling:
> I’ll give a more extensive review in a few weeks, but on quick review it would probably be a good idea to use variables in the localizations. So instead of “Edit Effect”, “Edit Feature”, and “Edit Item” it would be “Edit {type}”. Then you’d just have to pre-localize the context menus using `game.i18n.format(“DND5E.ContextMenuActionEdit”, {type: game.i18n.localize(“DND5E.Effect”)});`

Thanks for the heads up

I had initially run with "Edit Weapon", "Duplicate Equipment", etc, but had 2 areas that felt off later:
- With Hooks, are more the merrier? I had hooks for each type, but I didn't know whether that made sense, so reduced it to one, and later shifted to four types: Item, Feature, Spell, and Effect
- When I implemented the contextual options, like equipping items, "Equip Equipment" felt weird, like the specificity was maybe overkill

> I think a better place to put such an item's context menu would be to the right of the item's entry, where all the other item manipulation buttons are. For example, using the ubiquitous "more options" icon: \*image\*

The impression I've had when talking to others is a strong disdain for any new 'clutter' in that area, the current approaching being to right-click any part of the Item

I'm not sure about styling, I've just adopted what was present in another piece of work I was pointed at here: https://github.com/foundryvtt/dnd5e/pull/1588

Varriount:
> The impression I've had when talking to others is a strong disdain for any new 'clutter' in that area, the current approaching being to right-click any part of the Item

Which is why it's better to put a context menu there (and possibly move some existing actions to it). Many modules add actions there, which makes sense from a UX perspective, but isn't scalable. Adding a context menu button there fixes that problem.

RossGosling:
> Which is why it's better to put a context menu there (and possibly move some existing actions to it). Many modules add actions there, which makes sense from a UX perspective, but isn't scalable. Adding a context menu button there fixes that problem.

Oh sorry, so to try and summarise all of these bits

- "More Options" could be left-clicked to bring up the context menu
- A "More Options" icon could be a bit more intuitive? / obvious that it exists
- I think right-clicking an item will always bring up the context menu

Is that kinda what you're getting at?

Varriount:
> > Which is why it's better to put a context menu there (and possibly move some existing actions to it). Many modules add actions there, which makes sense from a UX perspective, but isn't scalable. Adding a context menu button there fixes that problem.
> 
> Oh sorry, so to try and summarise all of these bits
> 
>     * "More Options" could be left-clicked to bring up the context menu
> 
>     * A "More Options" icon could be a bit more intuitive? / obvious that it exists
> 
>     * I think right-clicking an item will always bring up the context menu
> 
> 
> Is that kinda what you're getting at?

Yes, pretty much. I feel that, especially these days, users don't really expect application-specific behavior on right-click when using the browser, which makes that kind of UX pattern less effective.

arbron:
> Yes, pretty much. I feel that, especially these days, users don't really expect application-specific behavior on right-click when using the browser, which makes that kind of UX pattern less effective.

My vote would be to leave this as a context menu for now and propose this as an option in a second PR. I think even if we go this route it would be wise to keep the context menu as an alternative options, with the "…" as just a more obvious shortcut to that.

> * Ideally, other modules should be able to add entries to the context menu in a sheet-agnostic way, rather than having to resort to directly modifying the context menu's HTML. This allows such modules to easily integrate with third-party sheets.

If we generate the context menu and the "…" menu using the same context menu system and hooks, then it will be extensible in the same way all context menus are since v10.

aaclayton (approved):
Conceptually this is in good shape. There are some changes needed to the actual implementation but I'll get this merged and then tidy those up as a follow-up commit.

aaclayton:
Code tidied up in https://github.com/foundryvtt/dnd5e/commit/755a0e297913ddf257405bb274aea644906e2192

RossGosling:
@aaclayton - Thank you!

Just one note would be the tidy-up not covering the [Actor Group Sheet](https://github.com/foundryvtt/dnd5e/blob/2.1.x/module/applications/actor/group-sheet.mjs#L211-L284)

aaclayton:
> @aaclayton - Thank you!
> 
> Just one note would be the tidy-up not covering the [Actor Group Sheet](https://github.com/foundryvtt/dnd5e/blob/2.1.x/module/applications/actor/group-sheet.mjs#L211-L284)

Oops, thank you! Got that taken care of.

PR #1976, opened 2022-12-23
[#1969]Compendiums "Items (SRD)" and "Spells (SRD) are no longer in alphabetical order 

### LINKED ISSUE #1969: Compendiums "Items (SRD)" and "Spells (SRD) are no longer in alphabetical order
I mentiond this on Discord and @krbz999 pointed out they have a `sort` property set that is affecting their order in the compendium.

This would appear to be a bug/unintended as it prevents the compendiums being sorted alphabetically.

kjolern:
I have removed the sort parameter from all items and spells to get them to sort alphabetical. Fixes #1969


arbron:
Looking at the definition for the document data model, it looks like `sort` is a required field. While this shouldn't cause any errors, it will just end up resulting in the sorts for all of these documents being set to `0` (the default value) on load. A better solution would be to set everything as `0` ourselves.

The easiest way to accomplish this would be to modify `utils/packs.mjs` and add `data.sort = 0;` as the final line in the `cleanPackEntry` function:

```
  if ( data.label ) data.label = cleanString(data.label);
  if ( data.name ) data.name = cleanString(data.name);
  data.sort = 0;
}
```

Then just run `gulp cleanPacks` and all of the packs will get the updated sort value data.

kjolern:
I have made the change to the utils/packs.mjs file @arbron.

aaclayton:
Implemented in https://github.com/foundryvtt/dnd5e/commit/16c334a752ffc62a24d348ddb25ca04c63b6551b

Issue #1977, opened 2022-12-25
`ammo` type consumables are not localized in the AbilityUseDialog.


krbz999:
There is localization for `DND5E.ConsumableAmmunition` but not for `DND5E.ConsumableAmmo`.

![image](https://user-images.githubusercontent.com/50169243/209472139-e3d9eb67-9361-4306-8891-3858987f82b7.png)

A quick glance leads me to believe that simply changing all instances of `DND5E.ConsumableAmmunition` to `DND5E.ConsumableAmmo` fixes it, but there may be places I have missed.

Issue #1955, opened 2022-12-03
Chatlog saving throw buttons trigger deprecation warnings.


krbz999:
![image](https://user-images.githubusercontent.com/50169243/205443060-a6771f8b-16b8-433a-b6e4-8b7d4fd7e096.png)

Issue #1930, opened 2022-11-13
World compendium packs are not migrated


dev7355608:
https://github.com/foundryvtt/dnd5e/blob/f2c7c99193f5c00d327e9fb515b3759919b86e91/module/migration.mjs#L72

```diff
- if ( p.metadata.package !== "world" ) continue; 
+ if ( p.metadata.packageType !== "world" ) continue; 
```

PR #1978, opened 2022-12-25
[#1977] [#1955] [#1930] [#1979] [#1984] Ammo localization; chatlog deprecation warning; world pack migration; actor spellcasting mod; null limited uses in ability dialog

krbz999:
* Changed localization of Ammunition to use 'ammo', matching the key used for the consumable type. To the best of my ability, I've not found any other references to `DND5E.ConsumableAmmunition`.
* Passing the token document to the speaker method instead of the placeable, avoiding the only lingering deprecation warning I've encountered post v10 release.
* For 1930: Weirdly `metadata.package` does exist (as `"world"`) immediately when a compendium is created, but this property disappears when refreshed.
* Added `data.attributes.spellmod = data.abilities[data.attributes.spellcasting || "int"]?.mod ?? 0;` to actor's roll data. This one-liner may be a duplicate of a different issue; if so, I can easily remove it, though supporting a 'default' spellcasting modifier for actors that have none is still worthwhile.
* When determining whether Limited Uses should force the AbilityUseDialog, added a check for `uses.max > 0` rather than just `uses.per`, which is identical to whether the checkbox appears in the dialog.

Related issues:
- #1977
- #1955
- #1930 
- #1979 
- #1984

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -302,7 +302,7 @@
 "DND5E.ConsumeWarningZeroAttribute": "{name} has run out of its designated attribute resource pool!",
 "DND5E.ConsumeResource": "Consume Resource?",
 "DND5E.ConsumeRecharge": "Consume Recharge?",
-"DND5E.ConsumableAmmunition": "Ammunition",
+"DND5E.ConsumableAmmo": "Ammunition",
 "DND5E.ConsumableFood": "Food",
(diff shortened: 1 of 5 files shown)


aaclayton (approved):
Looks good to me, thank you!

PR #1988, opened 2022-12-29
Add grouping back to item target type

arbron:
The grouping of item target type got removed when the "None" options was modified, this adds it back.


aaclayton (approved):
Aah, I missed that was an optgroup. My mistake. Thanks Jeff.

PR #1990, opened 2022-12-29
[#1835] Create system data models for actors

arbron:
- Add system data models for `character`, `npc`, and `vehicle`
- Refactors data model for `group` to match styling of other data models
- Add ability to `SystemDataModel` to perform multiple levels of `mixin` merging


module/data/actor/templates/common.mjs
```diff
@@ -0,0 +1,67 @@
...
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static migrateACData(source) {
+    if ( !source.attributes?.ac ) return;
+    const ac = source.attributes.ac;
+
+    // If the actor has a numeric ac.value, then their AC has not been migrated to the auto-calculation schema yet.
+    if ( Number.isNumeric(ac.value) ) {
+      ac.flat = parseInt(ac.value);
+      ac.calc = this._systemType === "npc" ? "natural" : "flat";
```
aaclayton, line 44: Can we avoid the complexity of needing a `static _systemType` by just making an assumption here like picking `"flat"` if we have to do a migration from a numeric value?

module/data/abstract.mjs
```diff
@@ -80,7 +88,7 @@ export default class SystemDataModel extends foundry.abstract.DataModel {
   static mixin(...templates) {
```
aaclayton, line 88: We should annotate the return value here as `{typeof SystemDataModel}` to indicate that we are returning the constructor, not an instance.

module/data/actor/ability.mjs
```diff
@@ -0,0 +1,26 @@
...
+ * An embedded data structure for actor ability scores.
+ * @see ActorData5e
+ *
+ * @property {number} value          Ability score.
+ * @property {number} proficient     Proficiency value for saves.
+ * @property {object} bonuses        Bonuses that modify ability checks and saves.
+ * @property {string} bonuses.check  Numeric or dice bonus to ability checks.
+ * @property {string} bonuses.save   Numeric or dice bonus to ability saving throws.
+ */
+export default class AbilityData extends foundry.abstract.DataModel {
```
aaclayton, line 13: I would prefer for something relatively small like each ability to instead be a `SchemaField` instead of a full embedded `DataModel`. Using an embedded data model is more heavyweight and unless we need that embedded data model to contain a lot of specialized logic we would be better served using the more lightweight tool. 

You've used this approach elsewhere like with `makeResourceField` which returns a `SchemaField` for each resource pool. Can we do the same here?

module/data/actor/skill.mjs
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * An embedded data structure for individual skill data.
+ *
+ * @property {number} value            Proficiency level creature has in this skill.
+ * @property {string} ability          Default ability used for this skill.
+ * @property {object} bonuses          Bonuses for this skill.
+ * @property {string} bonuses.check    Numeric or dice bonus to skill's check.
+ * @property {string} bonuses.passive  Numeric bonus to skill's passive check.
+ */
+export default class SkillData extends foundry.abstract.DataModel {
```
aaclayton, line 12: Same comment here as with `AbilityData`. I would really prefer these to be more lightweight `SchemaField`.

module/data/actor/templates/common.mjs
```diff
@@ -0,0 +1,67 @@
...
+      }, {label: "DND5E.Abilities"})
+    });
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  static migrateData(source) {
+    this.migrateACData(source);
+    this.migrateMovementData(source);
```
aaclayton, line 28: Can we make the internal migration functions private? Or do they need to be part of the public API here?

module/data/actor/templates/details.mjs
```diff
@@ -0,0 +1,37 @@
+/**
+ * Shared contents of the details schema between various actor types.
+ */
+export default class DetailsField {
```
aaclayton, line 4: It feels a little awkward using a class in this way. It feels like we're not really saving that much in terms of lines of code compared to the case where we just declare these explicitly inline. Could we move to do that, even if we repeat some of this for character/npc/etc?

module/migration.mjs
```diff
@@ -194,6 +194,7 @@ export const migrateArmorClass = async function(pack) {
 
       // Perform the normal migration.
       _migrateActorAC(src, update);
+      // TODO: See if AC migration within DataModel is enough to handle this
```
aaclayton, line 197: At what point do we know the answer to this question?

aaclayton (changes requested):
Looking great @arbron - particularly good for a first pass. My comments attached.

module/data/actor/ability.mjs
Fyorl: It could be useful, particularly for ability scores and skills, to be able to move preparation logic for preparing a single ability score/skill inside these data model classes. Those two have reasonably involved preparation logic.

module/data/actor/templates/details.mjs
Fyorl: Not only LoC though, it's about code reuse as well, and not having to make the same change in 3/4 separate places.

module/data/actor/ability.mjs
arbron: Yeah, moving the preparation logic for abilities and skills here is what I was thinking.

module/data/actor/templates/common.mjs
Fyorl: I remember we took some pains to make sure NPC actors were given a `natural` AC as part of this migration. We could move it back to the general `migration.mjs` code. The `_systemType` concept might come in useful elsewhere though.

module/data/actor/templates/common.mjs
arbron: Hard private isn't going to work for these methods because of how the mixin operates:

`TypeError: Failed data migration for NPCData: can't access private field or method: object is not the right class`

module/data/actor/templates/common.mjs
```diff
@@ -0,0 +1,67 @@
...
+      }, {label: "DND5E.Abilities"})
+    });
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  static migrateData(source) {
+    this.#migrateACData(source);
+    this.#migrateMovementData(source);
```
aaclayton, line 28: To avoid issues with subclasses not being able to access this private migration you should do:
```suggestion
    CommonTemplate.#migrateACData(source);
    CommonTemplate.#migrateMovementData(source);
```

module/data/actor/templates/common.mjs
Fyorl: It indeed does not work for instance methods, but it should work for static methods so long as we explicitly use the class name per this comment: https://github.com/foundryvtt/dnd5e/pull/1990#discussion_r1059464136

PR #1991, opened 2022-12-29
Module art mapping support.

Fyorl:


aaclayton (changes requested):
This looks really good. As mentioned in Discord the one additional thing we need is a mechanism to add credit to actors who have their artwork replaced. We specifically need this for Forgotten Adventures because it's part of the agreement I have with them that allows us to include their token artwork, although if we can add a general solution that allows anyone who provides an artwork mapping to denote a credit line for where that artwork comes from that would be ideal IMO.

aaclayton (approved):
Would be good to test this with some FA tokens to ensure that it's working as intended.

PR #1894, opened 2022-10-26
[#1632] Adding ability to roll initiative with advantage from character sheet

LiamFallon213:
partially resolves #1632

Issue #1632, opened 2022-07-10
Add option to roll initiative with advantage and disadvantage


IngeRasch:
All in the title really, I would greatly appreciate it if the option to roll initiative with advantage and disadvantage could be added to this system.

oestred:
following

LiamFallon213:
I'll take a look into this

Dumbhuman:
Just some examples of why this is needed:

- Any creature with at least one level of Exhaustion has disadvantage on all ability checks (which would include initiative rolls).
- Any creature with the Poisoned condition has disadvantage on all ability checks (which would include initiative rolls).
- 7th level Barbarians get the Feral Instinct feature which gives advantage on all initiative rolls.
- 7th level Warrior Sidekicks get the Battle Readiness feature which gives advantage on all initiative rolls.
- 10th level Artificers can use an infusion to create the Helm of Awareness magic item which gives advantage on all initiative rolls.
- 1st level Wild Magic Sorcerers get the Tides of Chaos feature which can give advantage on one ability check (which would include initiative rolls).
- 1st level Twilight Clerics get the Vigilant Blessing feature which gives one creature advantage on their next initiative roll.
- The 2nd level spell Enhance Ability can be used for the Cat's Grace effect which grants one creature advantage on all dexterity checks (which would include initiative rolls) for up to 1 hour duration.
- The 8th level spell Foresight grants advantage on all ability checks (which would include initiative rolls) for up to 8 hours.

aaclayton:
Addressed in https://github.com/foundryvtt/dnd5e/pull/1994

Issue #1069, opened 2021-04-08
Initiative and disadvantage


aaclayton:
Originally in GitLab by @nulmas

We have ways to give a character advantage on Initiative rolls (via an effect or the toggle option in the character sheet), but not disadvantage. Being that Initiative rolls are ability checks, effects like Exhaustion and Hex should cause characters to roll with disadvantage, something that as far as I can tell is impossible to do in Foundry at the moment unless if rolled as a Dex check (which may ignore bonuses specifically to Initiative) and changed in the combat tracker.

Issue #811, opened 2020-11-12
Make initiative internally a skill.


aaclayton:
Originally in GitLab by @cmiller.coder

I am writing this as a proposed change to the 5e game system for foundry. With the advent of the active effects system, many traps and gaps exist in how the initiative is currently handled. The initiative is inherently a dexterity ability check. By treating it internally as a skill, there are many optimization and simplifications which can happen.

Pros:

- Bonuses that apply to all ability checks (for example, the Stone of Good Luck) wouldn't need separate active effects for initiative and skills.
- Jack of all Trades wouldn't need to handle initiative in its own special way.
- Better handling of homebrews which might allow for [alternative ability scores](https://www.enworld.org/threads/dms-option-intelligence-for-initiative.405631/) to be used.
- Allow for better support of the 7th level feature Quickdraw from Matt Mercer's Gunslinger Fighter subclass. This allows you to add your proficiency bonus to the initiative roll.
- If there is ever an automated way of granting advantage to ability checks (Something like Enhance Ability: Cat's Grace), it would automatically apply to initiative correctly.

Cons:

Initiative is not a skill, so care might have to be taken in some places to not include it. For example, being careful to not include it a list of skills for skill gaining feats/features such as Expertise, Skilled, or Prodigy.

Overall, I think it would result in better handling with less code.

aaclayton:
We are taking a different route with initiative that does not treating it as a skill.

PR #1994, opened 2022-12-30
Support an initiative roll configuration Dialog for characters

aaclayton:
Closes Issues:
- https://github.com/foundryvtt/dnd5e/issues/1632
- https://github.com/foundryvtt/dnd5e/issues/1069
- https://github.com/foundryvtt/dnd5e/issues/811 (wontfix)

Closes PRs:
- https://github.com/foundryvtt/dnd5e/pull/1894 (close without merging)
- https://github.com/foundryvtt/dnd5e/issues/1632 (close without merging)

## Objectives
- Ensure that initiative formula includes bonuses that are currently absent but should be included
- Enable a d20 roll dialog when rolling initiative off the character sheet to support circumstances where initiative can be conditionally rolled with a bonus or advantage/disadvantage

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1403,6 +1403,94 @@ export default class Actor5e extends Actor {
   /* -------------------------------------------- */
 
   /**
+   * Get an un-evaluated D20Roll instance used to roll initiative for this Actor.
+   * @param {object} [options]                        Options which modify the roll
+   * @param {D20Roll.ADV_MODE} [options.advantageMode]    A specific advantage mode to apply
+   * @param {string} [options.flavor]                     Special flavor text to apply
+   * @returns {D20Roll}                               The constructed but unevaluated D20Roll
+   */
+  getInitiativeRoll({advantageMode, flavor}={}) {
+
+    // Use a temporarily cached initiative roll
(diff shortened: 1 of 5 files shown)


module/documents/actor/actor.mjs
```diff
@@ -1402,6 +1402,87 @@ export default class Actor5e extends Actor {
...
+
+    // Standard initiative formula
+    const parts = ["1d20"];
+
+    // Special initiative bonuses
+    if ( "init" in this.system.attributes ) {
+      const init = this.system.attributes.init;
+      parts.push(init.mod);
+      if ( init.prof.term !== "0" ) parts.push(init.prof.term);
+      if ( init.bonus !== 0 ) parts.push(init.bonus);
```
Fyorl, line 1430: I think the way these are typically done in other roll methods is by using data references:

```suggestion
      parts.push("@prof");
      data.prof = init.prof.term;
      parts.push("@bonus");
      data.bonus = init.bonus;
```

Similarly for the `dexCheckBonus` and `globalCheckBonus` below. It's not strictly necessary, but I think it might have some use for 3rd-party consumers. Might also be useful to enforce rules about only adding proficiency bonus once to a given roll.

Fyorl (approved):
Looks good to me. Only one comment, but it's optional.

PR #1998, opened 2022-12-30
Add support for data models in TokenDocument.getTrackedAttributes.

Fyorl:
I have a branch for support for this as part of the core software also, but here we have overridden the (pending) core methods in order to support the custom `MappingField` that we have in the `dnd5e` system.

### DIFF
--- a/module/documents/token.mjs
+++ b/module/documents/token.mjs
@@ -1,3 +1,5 @@
+import { MappingField } from "../data/fields.mjs";
+
 /**
  * Extend the base TokenDocument class to implement system-specific HP bar logic.
  */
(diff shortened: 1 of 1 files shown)


module/documents/token.mjs
```diff
@@ -18,6 +20,7 @@ export default class TokenDocument5e extends TokenDocument {
 
   /** @inheritdoc */
   static getTrackedAttributes(data, _path=[]) {
+    if ( data instanceof foundry.abstract.DataModel ) return this._getTrackedAttributesFromSchema(data.schema, _path);
```
aaclayton, line 23: Looks good, but I wonder if it would be better for this logic to live as a static method on our data models themselves, so you would instead do:
```js
if ( data instanceof SystemDataModel ) data.constructor.getTrackedAttributes()
```

aaclayton (approved):

module/documents/token.mjs
Fyorl: I think that would work fine. Let me submit my PR for the core half of this work, as it may impact what we decide to do here. But it would still work if we take over the process in `TokenDocument5e.getTrackedAttributes` and redirect it to `SystemDataModel`.

Issue #2001, opened 2022-12-31
HP throws validation error when max is 0


arbron:
The HP validation is checking that `hp.min >= `hp.max` and throwing an error if true, but this causes an error to be thrown when characters are created because their default max HP is 0.

Issue #2000, opened 2022-12-31
Remove `xp.min` and `xp.max` from stored data


arbron:
Both of these fields are included in `template.json` and the new data models, but their values are completely derived from the character's level.

Issue #1996, opened 2022-12-30
2.1.0: Issues with Resource Consumption


krbz999:
The target is populated with Objects when the type is set to 'Attribute'.

![image](https://user-images.githubusercontent.com/50169243/210105253-d875744d-c0cc-492e-9163-b0ecec88496e.png)

Relatedly, items that do not have `system.quantity` can be set to consume their own quantity.

![image](https://user-images.githubusercontent.com/50169243/210105393-6d771cf0-c1d2-49c9-93b8-83d41389eefe.png)

Issue #1975, opened 2022-12-21
Failing to determine base item prevents entire item sheet from rendering


Fyorl:
In `ItemSheet5e#_getItemBaseTypes`

PR #2002, opened 2022-12-31
[#1578, #1900, #1934, #1972, #1975, #1996, #2000, #2001] Grab bag of bug fixes

arbron:
- Fixed optional chaining #1578
- Removed updates during `prepareBaseAbilities` and `prepareBaseSkills` because that can be handled by data models #1900 
- Add width/height to SVGs #1934
- Add `item.canToggle` to items in character inventory #1972
- Fixed optional chaining #1975
- Moved call to `_getItemConsumptionTargets` out of `mergeObject` to prevent it from being expanded & removed item from its own ammunition consumption list #1996
- Remove `xp.min` & `xp.max` from template and data models #2000
- Fix issue with HP validation #2001


Fyorl (approved):
Thanks very much, Jeff

PR #2003, opened 2022-12-31
Dynamically attach FA token artwork and credits rather than hard-coding it into the SRD data

Fyorl:
Also adds support for enabling or disabling portrait and token art provided by any given module.


aaclayton (approved):
Thanks for doing this @Fyorl - I have two thoughts for you to consider as part of a final merge.

1. We had originally worked with a volunteer (Penelope/Vyrnali) who offered to write little biographies for every SRD creature, however this effort got abandoned fairly early on. While I think the biographies that were written are *okay* I think the inconsistency of having a very incomplete set of biographies is worse than having the value of some of them - it also increases the burden for other languages who want to localize the compendium content. I think we should remove all the biographies and set them all to an empty string.

2. We may want to also set `img` and `token.texture.src` to an empty string and instead allow the core software to populate whatever default value it is configured to do. If modules override or update the default token to be something other than `mystery-man.svg` we could allow that configuration to apply by making these fields blank in the source data.

PR #2004, opened 2022-12-31
Refresh monsters & heroes compendiums with new actor data models

arbron:


Fyorl:
Good stuff, Jeff. I didn't realise we had so much derived data saved in our SRD content. The only issue I've noticed after review was that some creatures seem to have lost their `traits.dr.bypasses`. You can take a look at the Earth Elemental or the Shadow to see this. It might be an issue with our schema but the definition at least looked ok to me.

Fyorl:
I located the issue with bypasses and pushed a fix. I'll update the JSONs in just a moment.

Fyorl (approved):

PR #2009, opened 2023-01-01
[#1997] Add support for preRollInitiative and rollInitiative hooks.

### LINKED ISSUE #1997: Request: Add `dnd5e.preRollInitiative` and `dnd5e.rollInitiative` hooks
The ability to modify the roll behavior through these hooks, akin to all other d20Roll hooks (abilityTest and abilitySave, etc) would be very helpful.

Some useful things a user (or module) might want to modify prior to the roll being resolved:
- What ability is used.
- Adding more terms to the `parts` or modifying the present ones.
- Modifying the `data` used to evaluate the terms.
- Possibly even setting a minimum or maximum value of the d20.

(Relatedly, after playing around with the code recently merged into 2.1.0, I found that shift/ctrl/alt did not skip the dialog nor apply advantage/disadvantage.)

Fyorl:
Some adjustments for things I missed in my initial review that came up in later testing such as missing fast-forward rolls and the advantage button not being highlighted in the roll dialog when 'advantage on initiative' was checked. Also adds rudimentary support for `preRollInitiative` and `rollInitiative` hooks.

- Closes #1997

### DIFF
--- a/module/dice/dice.mjs
+++ b/module/dice/dice.mjs
@@ -110,8 +110,9 @@ export async function d20Roll({
  * @param {boolean} [config.disadvantage]  Is something granting this roll disadvantage?
  * @param {boolean} [config.fastForward]   Should the roll dialog be skipped?
  * @returns {{isFF: boolean, advantageMode: number}}  Whether the roll is fast-forward, and its advantage mode.
+ * @internal
  */
-function _determineAdvantageMode({event, advantage=false, disadvantage=false, fastForward}={}) {
+export function _determineAdvantageMode({event, advantage=false, disadvantage=false, fastForward}={}) {
   const isFF = fastForward ?? (event && (event.shiftKey || event.altKey || event.ctrlKey || event.metaKey));
   let advantageMode = CONFIG.Dice.D20Roll.ADV_MODE.NORMAL;
   if ( advantage || event?.altKey ) advantageMode = CONFIG.Dice.D20Roll.ADV_MODE.ADVANTAGE;
--- a/module/dice/dice.mjs
+++ b/module/dice/dice.mjs
@@ -58,7 +58,9 @@ export async function d20Roll({
 
   // Handle input arguments
   const formula = ["1d20"].concat(parts).join(" + ");
-  const {advantageMode, isFF} = _determineAdvantageMode({advantage, disadvantage, fastForward, event});
+  const {advantageMode, isFF} = CONFIG.Dice.D20Roll.determineAdvantageMode({
+    advantage, disadvantage, fastForward, event
+  });
   const defaultRollMode = rollMode || game.settings.get("core", "rollMode");
   if ( chooseModifier && !isFF ) {
     data.mod = "@mod";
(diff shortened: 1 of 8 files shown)


module/documents/actor/actor.mjs
```diff
@@ -1488,15 +1488,27 @@ export default class Actor5e extends Actor {
...
+    /**
+     * A hook event that fires before initiative is rolled for an Actor.
+     * @function dnd5e.preRollInitiative
+     * @memberof hookEvents
+     * @param {Actor5e} actor                The Actor that is rolling initiative.
+     * @param {object} data                  The roll data.
+     * @param {string[]} parts               The roll parts.
+     * @param {D20RollConfiguration} config  Configuration data for the pending roll.
+     */
+    Hooks.callAll("dnd5e.preRollInitiative", this, data, parts, options);
```
Fyorl, line 1507: I don't think we can support cancelling the roll by returning `false` from a hook here because this workflow is integrated with core `Combat#rollInitiative`, which does not have any provision for cancelling a roll based on the return value of `Combatant#getInitiativeRoll`.

arbron: That seems fine here. Other places where hooks can cancel rolls it is expected that modules would divert to an alternative, but for initiative everything eventually needs to end up back in the core initiative system.

module/documents/actor/actor.mjs
```diff
@@ -1488,15 +1488,27 @@ export default class Actor5e extends Actor {
       if ( Number.isNumeric(abilityValue) ) parts.push(String(abilityValue / 100));
     }
 
-    // Create the d20 roll
-    const formula = parts.join(" + ");
-    return new CONFIG.Dice.D20Roll(formula, data, {
-      flavor: flavor ?? game.i18n.localize("DND5E.Initiative"),
-      advantageMode: advantageMode,
-      halflingLucky: flags.halflingLucky ?? false,
+    foundry.utils.mergeObject(options, {
```
arbron, line 1491: I believe this merge should be reversed, so options provided into the method take priority over the ones set here. That is how other rolling methods are set up:

```javascript
const rollData = foundry.utils.mergeObject({
  flavor: options.flavor ?? game.i18n.localize("DND5E.Initiative"),
  halflingLuck: flags.halflingLucky ?? false,
  critical: null,
  fumble: null
}, options);
```

krbz999: Had a look at the other rolls, and this should be `halflingLucky`, not `halflingLuck`.

Fyorl: Good catches, thanks

arbron (approved):

module/dice/dice.mjs
```diff
@@ -110,8 +110,9 @@ export async function d20Roll({
  * @param {boolean} [config.disadvantage]  Is something granting this roll disadvantage?
  * @param {boolean} [config.fastForward]   Should the roll dialog be skipped?
  * @returns {{isFF: boolean, advantageMode: number}}  Whether the roll is fast-forward, and its advantage mode.
+ * @internal
  */
-function _determineAdvantageMode({event, advantage=false, disadvantage=false, fastForward}={}) {
+export function _determineAdvantageMode({event, advantage=false, disadvantage=false, fastForward}={}) {
```
aaclayton, line 115: Instead of exporting a function, can we please reframe this as a static method on the `D20Roll` class?

module/documents/actor/actor.mjs
```diff
@@ -1488,15 +1488,27 @@ export default class Actor5e extends Actor {
...
+    /**
+     * A hook event that fires before initiative is rolled for an Actor.
+     * @function dnd5e.preRollInitiative
+     * @memberof hookEvents
+     * @param {Actor5e} actor                The Actor that is rolling initiative.
+     * @param {object} data                  The roll data.
+     * @param {string[]} parts               The roll parts.
+     * @param {D20RollConfiguration} config  Configuration data for the pending roll.
+     */
+    Hooks.callAll("dnd5e.preRollInitiative", this, data, parts, options);
```
aaclayton, line 1507: A main reason I *did not* add these hooks here is I really don't like the mess of having a pre-hook here in `getInitiativeRoll` and a post-hook in an overridden `rollInitiative()` method. This is - to me - really ugly and I would vote to remove it and either not offer hooks here or offer a different framing of hooks.

module/documents/actor/actor.mjs
```diff
@@ -1507,16 +1519,24 @@ export default class Actor5e extends Actor {
    * @returns {Promise<void>}           A promise which resolves once initiative has been rolled for the Actor
    */
   async rollInitiativeDialog(rollOptions={}) {
+    let {advantageMode, fastForward, event} = rollOptions;
+    const ff = _determineAdvantageMode({fastForward, event});
+    if ( ff.isFF ) advantageMode ??= ff.advantageMode;
```
aaclayton, line 1524: I don't personally agree with fast-forwarding the initiative roll. Unlike other D20 rolls where Adv/Disadv are common, having those modifiers for initiative is very rare.  It's more likely to result in a mis-click that incorrectly applies adv/disadv to initiative than it is to save a user time. Furthermore, people aren't used to checking the chat log for initiative to ensure that their initiative was rolled properly, so an initiative roll that incorrectly has advantage or disadvantage is unlikely to be noticed.

module/documents/actor/actor.mjs
```diff
@@ -1526,6 +1546,24 @@ export default class Actor5e extends Actor {
 
   /* -------------------------------------------- */
 
+  /** @inheritdoc */
+  async rollInitiative(options={}) {
+    const combat = await super.rollInitiative(options);
+    const combatant = this.isToken ? combat.getCombatantByToken(this.token.id) : combat.getCombatantByActor(this.id);
```
aaclayton, line 1552: The `Actor#rollInitiative` method may create initiative rolls for multiple combatants, not necessarily only one.

aaclayton (changes requested):

Fyorl:
Fast-forwarding initiative rolls has been removed for now due to their relative permanence compared with other sheet rolls. I've moved the hooks into `Actor5e#rollInitiative`.

PR #2011, opened 2023-01-02
[#1605] Prepare scale values earlier to allow them to be modified by active effects

### LINKED ISSUE #1605: Make scale values available for active effects
Move `_computeScaleValues` to `prepareBaseData`, thus allowing scale values to be used on active effects with the help of DAE

arbron:
Moves the data preparation for scale value between when item data is prepared but before active effects are applied so they can be modified by active effects. This also allows them to be used as in input for an active effect if a player is using a module like DAE.

Resolves #1605

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -124,17 +124,9 @@ export default class Actor5e extends Actor {
 
   /* --------------------------------------------- */
 
-  /** @override */
-  prepareEmbeddedDocuments() {
-    super.prepareEmbeddedDocuments();
-    this._prepareScaleValues();
-    this.applyActiveEffects();
-  }
-
-  /* --------------------------------------------- */
-
   /** @inheritDoc */
   applyActiveEffects() {
+    this._prepareScaleValues();
     // The Active Effects do not have access to their parent at preparation time, so we wait until this stage to
     // determine whether they are suppressed or not.
     this.effects.forEach(e => e.determineSuppression());
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -93,6 +93,7 @@ export default class Actor5e extends Actor {
   prepareData() {
     this._classes = undefined;
     this._preparationWarnings = [];
+    this.items.forEach(item => item.prepareInitialAttributes());
     super.prepareData();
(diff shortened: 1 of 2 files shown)


Fyorl:
A couple of high-level design issues that bear discussing before I review:

1. We have a longstanding issue in the system whereby some active effects are intended to modify values *before* calculations are done. For example, an AE that modifies an ability score would want that modification reflected in that ability's derived `mod` value. Some active effects are intended to modify values *after* calculations are done. For example, an AE that modifies `ac.value` would want that modification done after the `ac.value` has already been calculated so that it is not subsequently overridden as part of normal data preparation.

It seems to me that that this line lies along the base/derived line. An AE targeting a base field should probably be applied before derived data is calculated, and an AE targeting a derived field should probably be applied afterwards, but I don't know if that's true in all cases. If it is true though, then we have a clear path to providing a smart AE system that could be quite powerful.

The question, then, is does this proposal fit into that? If we do the above work, would this particular issue be solved as part of that?

2. If this issue would not be solved by some wider work on AEs in general, is there a solution that does not involve looping over all embedded items twice on every data preparation? The answer might be no but data preparation is not free and can become hot code in certain contexts (migrations, world initialisation) so we should try to be frugal where we can.

arbron:
@Fyorl Upon further examination of how `Actor` is implemented in core, I've realized that there is a simpler way to implement this. The only thing that needs to happen is we need to prepare these scale values after `Item#prepareData` is called but before active effects are applied, so no need to loop over items again.

module/documents/actor/actor.mjs
```diff
@@ -124,6 +124,15 @@ export default class Actor5e extends Actor {
 
   /* --------------------------------------------- */
 
+  /** @override */
+  prepareEmbeddedDocuments() {
+    super.prepareEmbeddedDocuments();
```
Fyorl, line 129: This call to `super.prepareEmbeddedDocuments` (in this case `Actor#prepareEmbeddedDocuments`) will also call `this.applyActiveEffects`, no? I suppose we could move `this._prepareScaleValues` to the top of the `applyActiveEffects` override instead?

Fyorl (changes requested):

Fyorl (approved):

Issue #1944, opened 2022-11-26
[2.1.x] Editing scale values doesn't retain data properly


arbron:

krbz999:
If this is the bug that the UI for a scale value 'resets' with each change (despite the data being added most of the time), then it's also happening for Item Grants and Hit Points advancements.

PR #2014, opened 2023-01-02
[#1944] Fix bug with displaying updates in AdvancementConfig

arbron:
Fixes a bug introduced by the interaction between item data models and advancements. Replicates core's app updating behavior using `apps` to automatically re-render config applications when an advancement is updated.

### DIFF
--- a/module/applications/advancement/advancement-config.mjs
+++ b/module/applications/advancement/advancement-config.mjs
@@ -12,10 +12,10 @@ export default class AdvancementConfig extends FormApplication {
     super(advancement, options);
 
     /**
-     * The advancement being created or edited.
-     * @type {Advancement}
+     * The ID of the advancement being created or edited.
+     * @type {string}
      */
-    this.advancement = advancement;
+    this._advancementId = advancement.id;
 
     /**
      * Parent item to which this advancement belongs.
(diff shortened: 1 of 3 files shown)


module/documents/item.mjs
```diff
@@ -2039,10 +2039,17 @@ export default class Item5e extends Item {
...
-    if ( source ) return this;
+    if ( source ) {
+      advancement.render();
+      return this;
+    }
+
     const advancementCollection = this.toObject().system.advancement;
     advancementCollection[idx] = advancement.toObject();
-    return this.update({"system.advancement": advancementCollection});
+    this.update({"system.advancement": advancementCollection}).then(r => {
```
Fyorl, line 2049: This needs to be returned.

module/applications/advancement/advancement-config.mjs
```diff
@@ -12,10 +12,10 @@ export default class AdvancementConfig extends FormApplication {
     super(advancement, options);
 
     /**
-     * The advancement being created or edited.
-     * @type {Advancement}
+     * The ID of the advancement being created or edited.
+     * @type {string}
      */
-    this.advancement = advancement;
+    this._advancementId = advancement.id;
```
Fyorl, line 18: Should probably be hard-private unless we expect external code to inspect and modify this?

Fyorl (changes requested):

Fyorl (approved):

PR #2016, opened 2023-01-02
[#1405] Trait Advancement

### LINKED ISSUE #1405: Advancement Type: Trait
The trait advancement type would be responsible for assigning an actor any proficiencies. It is designed only for values that are `true` or `false` such as saving throw & skill proficiencies, armor, weapon, & tool proficiencies, languages, damage resistances, etc. It is not designed to handle expertise or properties with numeric values such as movement.

This would be able to handle cases where traits are simply granted (becoming a Fighter always gives you proficiency in simple and martial weapons) or when a choice is needed (Bards can choose between three different musical instruments).

This would resolve #1016. The closed Background MR !361 contains a complete implementation of this design outside the advancement system.

## Level Up Interface

For advancements that simply grant traits this will simple show what the player is given. If a choice of traits is required then the player is given a dropdown that contains all of their options. When an option is selected it is added to the list of selected options and the dropdown is updated with a modified list showing any remaining options. An example of this UI can be seen on the old Background item type MR !358.

## Configuration

The `type` property in the configuration object would indicate which type of trait is being represented. `"armor"` would represent armor proficiencies, `"dv"` would be damage vulnerabilities, `"saves"` would be saving throw proficiencies, etc.

The boolean `allowReplacements` indicates whether an alternative can be chosen if the character already has proficiencies in the types granted. While not useful for classes, this will be used on backgrounds where, for example, the player can chose another skill to gain proficiency if they ones provided by the background are already proficient.

The `grants` property is an array of strings containing a list of proficiencies always granted. If a category like Simple Weapons (`"sim"`) is included it grants the player proficiency in that entire category.

The `choices` property contains an array of objects with a `count` value and an optional `choices` array. The `count` number indicates how many choices are given to the player. If no `choices` array exists then they are able to choose from anything of that type, otherwise the `choices` array lists any combination of individual entries or categories to choose from. If a category like Simple Weapons is included then the player can choose any number of items from within that category, but not the category itself.

This example indicates that the player is granted proficiency in Light Armor if they don't already have it. If they are already proficient in Light Armor, nothing happens.

```javascript
{
  type: "armor",
  allowReplacements: false,
  grants: ["lgt"],
  choices: []
}
```

This will grant the player proficiency in the Insight and Religion skills. If they are already proficient in either of those skills, they will be able to choose a replacement from any other skill.

```javascript
{
  type: "skills",
  allowReplacements: true,
  grants: ["ins", "rel"],
  choices: []
}
```

This will give the player a choice from any two languages in which they do not already have proficiency.

```javascript
{
  type: "languages",
  allowReplacements: false,
  grants: [],
  choices: [{ count: 2 }]
}
```

This will give the player a choice of any three musical instruments.

```javascript
{
  type: "tool",
  allowReplacements: false,
  grants: [],
  choices: [{ count: 3, choices: ["mus"] }]
}
```

## Value

The value object contains `chosen` which is an array of the choices the player made. This list will also contain any traits that were automatically granted (because it makes some behind the scenes stuff easier).

```javascript
{
  chosen: ["bagpipes", "flute", "viol"]
}
```

## Progress
- [x] #2498
- [x] #2499
- [x] #2522 
- [x] #2543
- [x] #2546 
- [x] #2544
- [x] #2547

arbron:
This advancement gives the ability to grant traits to a character upon level up.

Bard - Advancements

### Configuration

The advancement configuration window tries to make it as easy as possible to set up the most common use cases. The first step is always selecting the trait type. For something like saving throw proficiencies on classes, which always just grant two choices, the next step is to simple check off those saves granted:

Bard - Configure Saves

For something like weapon proficiencies, you can also just click a category to instead grant proficiency in that entire category:

Bard - Configure Weapons

For choices, rather than immediately selection entries the user first clicks the "+" button to add a new choice. By default the choices is from all available options of the type. So for a Bard's skills, all that is required is setting the number of skills to be chosen:

Bard - Configure Skills

For a Rogue on the other hand, the individual skills will need to be checked after the choice is created:

Rogue - Configure Skills

Categories can also be selected for choices in which case the player will be given the option of selecting any item within that category, such as will Bard's giving the choice of any musical instrument:

Bard - Configure Tools

More advanced configurations are also available, for example the "Cloistered Scholar" background which grants proficiency in History, plus the player's choice from Arcana, Nature, and Religion:

Cloistered Scholar - Advancements

Cloistered Scholar - Configure Skills

Or the "Haunted One", which offers a choice of two languages, but one of which must be from a specific list:

Haunted One - Configure Languages

### Application

During the Advancement process, each trait advancement will display what it offers and what of those things have actually been applied:

Bard - Apply Armor

If there are choices, then a dropdown appears to let the player choose:

Bard - Apply Skills

This dropdown automatically updates after each choice to reduce the number of options to those available.

Resolves #1405

### To Do
- [ ] Add "Expertise" option
- [ ] Add trait advancements to SRD classes & backgrounds
- [ ] Display traits on Class Summary journal pages


arbron:
Closing in favor of several smaller PRs

PR #2019, opened 2023-01-03
Rework sheet and roll data preparation to avoid mutating actor and item state

Fyorl:
Apologies for how clumsy some of this is. The guiding principles here were:
 - It is ok to pass the real actor or item instance into the template data as it's only for reading.
 - Avoid mutating a property of the real actor or item instance as part of template data or roll data preparation.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -203,13 +203,14 @@ export default class Actor5e extends Actor {
    *                                            either a die term or a flat term.
    */
   getRollData({ deterministic=false }={}) {
-    const data = foundry.utils.deepClone(super.getRollData());
+    const data = {...super.getRollData()};
     data.prof = new Proficiency(this.system.attributes.prof, 1);
     if ( deterministic ) data.prof = data.prof.flat;
+    data.attributes = foundry.utils.deepClone(data.attributes);
     data.attributes.spellmod = data.abilities[data.attributes.spellcasting || "int"]?.mod ?? 0;
     data.classes = {};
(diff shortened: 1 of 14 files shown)


module/documents/actor/actor.mjs
```diff
@@ -203,13 +203,14 @@ export default class Actor5e extends Actor {
    *                                            either a die term or a flat term.
    */
   getRollData({ deterministic=false }={}) {
-    const data = foundry.utils.deepClone(super.getRollData());
+    const data = {...super.getRollData()};
```
aaclayton, line 206: I'm confused about why we need this to be a different object instead of just:
```suggestion
    const data = super.getRollData();
```

aaclayton (approved):
You claimed this was clumsy, but I don't think it is. This feels like a clean solution to me that probably results in some nice side-effect performance benefits. I suspect sheet rendering time is measurably faster.

module/documents/actor/actor.mjs
Fyorl: It's as Jeff says: `Actor#getRollData` returns `this.system` and, since `deepClone` returns the original instance for complex objects, this was now returning the actor's actual `this.system` instance since it's a complex `DataModel` object now. We now copy over the properties of `this.system` since we need their derived values, but it's housed in a separate object instance so that mutations at the top-level do not mutate `this.system`. Since the object spread copies over properties and does not clone them, there's still some risk that those inner properties could be mutated, but it's difficult to find a better solution right now without support from the core software.

PR #2027, opened 2023-01-05
Assorted bug fixes

### LINKED ISSUE #1238: Immediately evaluated inline rolls which occur within the context of an Actor Sheet (or owned Document sheet) should obtain that Actor's roll data object for the purposes of evaluation
Originally in GitLab by @unsoluble1

(New ticket for this rather than resurrecting foundryvtt#4631)

Obtain and render rollData for *non-deferred* inline rolls, so that their full results may be displayed inline with text in places like Description fields, etc).

"The target makes a `[[1d20 + @abilities.dex.mod]]{Dexterity Check}`", for example.

### LINKED ISSUE #1239: When rolling an item with charges and ammunition, the checkbox for resource consumption is incorrectly displayed
Originally in GitLab by @Kandashi

If an item it setup with both self consuming charges and ammo consumption, the ammo is consumed twice. Once upon confirming "use charge" and once upon rolling the attack.

### LINKED ISSUE #1639: Spell Scroll to Hit and Save DC corrections
The spell scrolls do not calculate the To Hit or Save DCs correctly, instead the To Hit is getting a bonus of what should be the flat value, and Save DCs are set to the owner's spell save DCs.
image

Fixing the To Hit is as easy as adding `- @mod` in the Attack Roll Bonus field (where the flat attack bonus already exists).
image

Fixing the Save DCs is probably a little more difficult, currently the saving throw info is generated from the spell's information, but sets the DC based on the spellcasting attribute, this should change to a Flat DC defined in the Spell Scroll, and only the Save Ability populated from the Spell item. Currently, setting the DC on the Spell Scroll item gets overwritten
image

### LINKED ISSUE #1817: Dice without an amount do not double on critical hits
A damage roll "1d10" will double to "2d10" on a critical hit, but if the damage part of the roll is just "d10," it gets replaced with "1d10" on hits and crits and is never doubled.

(Foundry 10.285, dnd5e 2.0.2)

### LINKED ISSUE #1891: Shield is affecting Natural Armor twice
When using a shield item with natural armor the bonus effects it twice over.

For my example I am using the SRD shield which has a +2 to AC.

This only happens when I have the shield equipped first.
 
When I input x into the config window for natural armor it forces it to x+2. Then if I hit update it has x+2+2.

Input 12
unknown
Forces it to 14
unknown
AC shows it is 16
unknown

### LINKED ISSUE #1958: Spells start with empty school, but it automatically changes to "Abjuration" as soon as any change is made
`template.json` and `SpellData` both define spell school as blank by default, but the "Spell School" dropdown doesn't allow for a blank school. This results in the spell's school being changed to the first value in the school list (in this case, Abjuration) as soon as any other change is made to the spell.

Two possible solutions:
1. Set a default spell school in `template.json` & `SpellData` and update the school field to disallow blank values
2. Update the "Spell School" select to allow blank values

Fyorl:
- Closes #1817 
- Closes #1639 
- Closes #1891 
- Closes #1958 
- Closes #1238 
- Closes #1239

### DIFF
--- a/module/applications/item/item-sheet.mjs
+++ b/module/applications/item/item-sheet.mjs
@@ -71,6 +71,7 @@ export default class ItemSheet5e extends ItemSheet {
       labels: item.labels,
       isEmbedded: item.isEmbedded,
       advancementEditable: (this.advancementConfigurationMode || !item.isEmbedded) && context.editable,
+      rollData: this.item.getRollData.bind(this.item),
 
       // Item Type, Status, and Details
       itemType: game.i18n.localize(`ITEM.Type${this.item.type.titleCase()}`),
@@ -78,13 +79,6 @@ export default class ItemSheet5e extends ItemSheet {
       itemProperties: this._getItemProperties(),
       baseItems: await this._getItemBaseTypes(),
       isPhysical: item.system.hasOwnProperty("quantity"),
-
-      // Enrich HTML description
-      descriptionHTML: await TextEditor.enrichHTML(item.system.description.value, {
-        secrets: item.isOwner,
-        async: true,
-        relativeTo: this.item
-      }),
 
       // Action Details
       hasAttackRoll: item.hasAttack,
@@ -132,6 +126,14 @@ export default class ItemSheet5e extends ItemSheet {
         context.spellComponents = {...CONFIG.DND5E.spellComponents, ...CONFIG.DND5E.spellTags};
         break;
     }
+
+    // Enrich HTML description
+    context.descriptionHTML = await TextEditor.enrichHTML(item.system.description.value, {
+      secrets: item.isOwner,
+      async: true,
+      relativeTo: this.item,
+      rollData: context.rollData
+    })
     return context;
   }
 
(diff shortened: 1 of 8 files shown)


arbron (approved):

module/applications/item/item-sheet.mjs
```diff
@@ -71,6 +71,7 @@ export default class ItemSheet5e extends ItemSheet {
       labels: item.labels,
       isEmbedded: item.isEmbedded,
       advancementEditable: (this.advancementConfigurationMode || !item.isEmbedded) && context.editable,
+      rollData: this.item.getRollData.bind(this.item),
```
aaclayton, line 74: I think I'm missing an insight on why this helps us instead of just `rollData: this.item.getRollData()`. Is the idea that by passing a bound function it avoids potentially costly roll data preparation in cases where there are no inline rolls that require enrichment? I think I've probably just answered my own question, but I'll ask it anyways.

aaclayton (approved):
Looks good, thanks @Fyorl!

PR #2035, opened 2023-01-06
Change backgrounds on start heroes to proper background items, fix wildshape CR on Aoth

arbron:


packs/src/heroes/akra-dragonborn-cleric.json
```diff
@@ -5295,6 +5211,40 @@
...
+      "img": "icons/sundries/documents/document-torn-diagram-tan.webp",
+      "effects": [],
+      "folder": null,
+      "sort": 0,
+      "ownership": {
+        "default": 0
+      },
+      "flags": {
+        "core": {
+          "sourceId": "Item.VY4Dc9uJPNVlvu1D"
```
Fyorl, line 5236: This is a little suspicious, it looks like a world item maybe? Since these backgrounds are custom, I probably wouldn't expect them to have a `sourceId` at all.

Issue #2064, opened 2023-01-16
System data model migrations disrupt core data migrations


Fyorl:
When the core API adds a migration from `data` -> `system`, it will refuse if `system` already exists. In some cases, our `SystemDataModel` migrations will add properties indiscriminately, causing a `system` object to be created alongside the original, legacy `data` object, and causing the core migration to not add the migrated `system` property. For example, `PhysicalItemTemplate##migratePrice` will do this.

We should probably modify `SystemDataModel.migrateData` to call `super.migrateData` first, so that these core migrations can apply, and then apply the individual template migrations.

PR #2070, opened 2023-01-17
[#2064] Fix system DataModel migrations causing issues with core migrations, up minimum core version

arbron:
**Note**: I've removed the return statement in out overridden `migrateData` methods. The core JSDoc states that `migrateData` should return `source`, but `migrateDataSafe` just ignores that value. Do we think is will be necessary to add `return source;` to each of these methods or are we good as is?

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -125,8 +125,8 @@ export default class CharacterData extends CreatureTemplate {
 
   /** @inheritdoc */
   static migrateData(source) {
+    super.migrateData(source);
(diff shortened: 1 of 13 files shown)


Fyorl:
> **Note**: I've removed the return statement in out overridden `migrateData` methods. The core JSDoc states that `migrateData` should return `source`, but `migrateDataSafe` just ignores that value. Do we think is will be necessary to add `return source;` to each of these methods or are we good as is?

This is fine for now, but we should be ready to make changes here if it becomes an issue.

Fyorl (approved):

Issue #2066, opened 2023-01-17
Clicking "+ Add" under the action header on actor sheets creates passive item


arbron:
Before clicking the "+ Add" button on the "Actions" header for NPCs or the "Active Abilities" on PCs would create an item with "Action" already selected so it appears in the correct section, now it appears in the "Features" section on NPCs and the "Passive Abilities" section on PCs.

Issue #2058, opened 2023-01-15
Can't use decimal for movement


dmizrahi:
Hello,
Since the 2.1.x update, we can't use decimal for movement. The character sheet round up the value. It's annoying because Halfling for instance have a movement of 7.5m.
Thanks !

dev7355608:
The movement speed fields should be noninteger with step size 0.1.
https://github.com/foundryvtt/dnd5e/blob/b28a8680290ca47675fa86c28f1ec28843a2ac6d/module/data/actor/templates/attributes.mjs#L30-L44
https://github.com/foundryvtt/dnd5e/blob/b28a8680290ca47675fa86c28f1ec28843a2ac6d/module/data/actor/group.mjs#L41-L49

Issue #2057, opened 2023-01-14
Unable to change units in the movement config


dev7355608:
https://github.com/foundryvtt/dnd5e/blob/b28a8680290ca47675fa86c28f1ec28843a2ac6d/module/applications/actor/movement-config.mjs#L53
https://github.com/foundryvtt/dnd5e/blob/b28a8680290ca47675fa86c28f1ec28843a2ac6d/templates/apps/movement-config.hbs#L9

Issue #2055, opened 2023-01-13
Spells added to vehicle are not converted to scrolls


krbz999:
For both types, one can programmatically add a `spell` type item to each actor type using a script.

When using the drag-and-drop method, only `group` creates a scroll instead, where `vehicle` simply adds a Spell under 'Cargo'.

arbron:
We currently don't do any restriction on item types that are added programmatically, but adding spells to vehicles should definitely convert them to scrolls.

PR #2072, opened 2023-01-17
[#2055, #2057, #2058, #2066] Fix various issues with sheets

arbron:
- Create spell scrolls from spells dropped onto vehicle sheets #2055 
- Fix units not being available for actor movement #2057 
- Allow decimals in movement values (to support metric values) #2058
- Fix items created directly on sheets not getting proper properties #2066

### DIFF
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -1003,7 +1003,8 @@ export default class ActorSheet5e extends ActorSheet {
     }
 
     // Create a Consumable spell scroll on the Inventory tab
-    if ( (itemData.type === "spell") && (this._tabs[0].active === "inventory") ) {
+    if ( (itemData.type === "spell")
+      && (this._tabs[0].active === "inventory" || this.actor.type === "vehicle") ) {
       const scroll = await Item5e.createScrollFromSpell(itemData);
       return scroll.toObject();
     }
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -1181,7 +1181,7 @@ export default class ActorSheet5e extends ActorSheet {
     const itemData = {
       name: game.i18n.format("DND5E.ItemNew", {type: game.i18n.localize(`ITEM.Type${type.capitalize()}`)}),
       type: type,
-      system: { ...header.dataset }
+      system: foundry.utils.expandObject({ ...header.dataset })
     };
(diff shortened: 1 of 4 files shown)


module/utils.mjs
```diff
@@ -144,13 +144,35 @@ export async function preloadHandlebarsTemplates() {
...
+  if ( arguments.length !== 2 ) throw new Error("#dnd5e-with requires exactly one argument");
+  if ( foundry.utils.getType(context) === "function" ) context = context.call(this);
+
+  if ( foundry.utils.isEmpty(context) ) {
+    const inverse = options.inverse(this);
+    if ( inverse ) return options.inverse(this);
+  }
+
+  return options.fn(context, { data: options.data, blockParams: [context] });
+}
```
Fyorl, line 164: I'm not sure this is the right solution. It feels like it's probably a bug if we're looping over the Actor's items and we cannot find an `itemContext` for it. Can we instead ensure that every item has an associated `itemContext` or that the item is not included in the item array that is being traversed.

Fyorl (changes requested):

module/utils.mjs
arbron: Well, the problem is that vehicle crew & passenger entries don't have actual IDs to look up because they are not actually items, so we can't add them to the context object. The other solution I was testing is adding an `{{else}}` section to the existing `{{#with}}` block and adding the code to display those data types there, but that resulted in a bunch of duplicated template code.

Fyorl: OK, I understand now. It's because we're reusing the inventory partial and the vehicle context does not quite fit with the rest. So we could, alternatively, give the entries IDs, but that feels a bit hacky.

I think I would feel a bit better about this if the functionality of the helper were a bit more bespoke and not such an obvious 'patch' to the core handlebars helper. Could we instead call it something like `dnd5e-item-context` and have it take only the item ID as a parameter?

```hbs
{{#dnd5e-item-context item.id as |ctx|}}
```

I think we have access to the handlebars context root in `options`, so we could perform that lookup automatically. Something like:
```js
function itemContext(id, options) {
  // ...
  const context = options.data.root.itemContext?.[id];
  // ...
}
```

arbron:
@Fyorl I popped the vehicle sheet changes off to another branch so we can merge the other fixes while I work in that

Fyorl (approved):
Thanks very much, Jeff

PR #2073, opened 2023-01-17
[#2071] Default to metric length units for movement and senses

### LINKED ISSUE #2071: Adjust default units to match metric setting
Currently, movement & senses units always default to `ft`. If metric units are used, this should be default to `m`.

arbron:
Sets the default value for movement, senses, and grid units to meters if metric length units are enabled in settings.

Updates newly created scenes setting them to the correct units and adjusting the grid distance (so using default settings, scenes in metric mode will be created with a grid of 1.5 m, instead of 5 ft).

Resolves #2071

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -587,6 +587,21 @@ Hooks.on("renderJournalPageSheet", applications.journal.JournalSheet5e.onRenderJ
 
 Hooks.on("targetToken", canvas.Token5e.onTargetToken);
 
+Hooks.on("preCreateScene", (doc, createData, options, userId) => {
+  // Set default grid units based on metric length setting
+  const units = utils.defaultUnits("length");
+  if ( (units !== dnd5e.grid.units) && !foundry.utils.getProperty(createData, "grid.distance")
+    && !foundry.utils.getProperty(createData, "grid.units") ) {
+    const C = CONFIG.DND5E.movementUnits;
+    doc.updateSource({
+      grid: {
+        // TODO: Replace with `convertLength` method once added
+        distance: dnd5e.grid.distance * (C[dnd5e.grid.units]?.conversion ?? 1) / (C[units]?.conversion ?? 1), units
+      }
+    });
+  }
+});
+
 // TODO: Generalize this logic and make it available in the re-designed transform application.
 Hooks.on("dnd5e.transformActor", (subject, target, d, options) => {
   const isLegacy = game.settings.get("dnd5e", "rulesVersion") === "legacy";
--- a/lang/en.json
+++ b/lang/en.json
@@ -4008,8 +4008,6 @@
 "SETTINGS.5eHonorN": "Honor Ability Score",
 "SETTINGS.5eInitTBL": "Append the raw Dexterity ability score to break ties in Initiative.",
 "SETTINGS.5eInitTBN": "Initiative Dexterity Tiebreaker",
-"SETTINGS.5eMetricN": "Use Metric Weight Units",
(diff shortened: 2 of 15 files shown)


Fyorl (approved):

Fyorl (changes requested):
A couple of things we need to address first, I think:
1. Our setting is currently `metricWeightUnits`, and is labelled as such in the settings screen. If we want to handle distance units, we should add a separate setting for that. I do think this should be two separate settings because, for whatever reason, some localised versions of the rules mix metric distances with imperial weights and vice versa (even if the country in question does not use imperial measurements for anything).
2. There's not currently any way (that I can see) where a user can change the movement units for an actor. It seems to only be possible via the API.

arbron:
> 1. Our setting is currently `metricWeightUnits`, and is labelled as such in the settings screen. If we want to handle distance units, we should add a separate setting for that. I do think this should be two separate settings because, for whatever reason, some localised versions of the rules mix metric distances with imperial weights and vice versa (even if the country in question does not use imperial measurements for anything).

I can see that, I'll push this off to `2.2` since it is growing in scope.

> 2. There's not currently any way (that I can see) where a user can change the movement units for an actor. It seems to only be possible via the API.

That's another bug that is fixed in #2072

Fyorl (changes requested):
Looks good, I just have one thing that I'm not sure about. For senses and speeds we have a drop-down of units, and one of the options is 'automatic'. If a content module is packaging monsters with all their speeds and senses set to automatic, will that mean that they have 60ft Darkvision when imported into an imperial world, and 60m Darkvision when imported into a metric World? This can't be a new problem though, it must already be the case?

Should we remove this 'automatic' option now and instead have the drop-down default to the `defaultUnit("length")`? This would be the same as weapon and spell ranges, and item weights, which require a unit, so they do not have this problem.

arbron:
> Looks good, I just have one thing that I'm not sure about. For senses and speeds we have a drop-down of units, and one of the options is 'automatic'. If a content module is packaging monsters with all their speeds and senses set to automatic, will that mean that they have 60ft Darkvision when imported into an imperial world, and 60m Darkvision when imported into a metric World? This can't be a new problem though, it must already be the case?
> 
> Should we remove this 'automatic' option now and instead have the drop-down default to the `defaultUnit("length")`? This would be the same as weapon and spell ranges, and item weights, which require a unit, so they do not have this problem.

I'll take a look at it, but I think that might have been an unintended consequence of sharing the dialog between actors and species. It feels like the correct behavior is to have automatic only for when the dialog is shown from actor sheets, but require units for species sheets.

Fyorl:
> I'll take a look at it, but I think that might have been an unintended consequence of sharing the dialog between actors and species. It feels like the correct behavior is to have automatic only for when the dialog is shown from actor sheets, but require units for species sheets.

Doesn't the issue still exist for NPCs? They don't get their senses and speed from a species item (unless manually added). If the data I am inputting says the creature has a speed of 30 feet, I should of course set the unit explicitly to feet, but the drop-down kind of implies that I don't have to do that since it says 'Automatic (feet)'. But if this creature is included in a module and then imported into a World set to metric, then that creature suddenly has a speed of 30m no?

arbron:
> Doesn't the issue still exist for NPCs? They don't get their senses and speed from a species item (unless manually added). If the data I am inputting says the creature has a speed of 30 feet, I should of course set the unit explicitly to feet, but the drop-down kind of implies that I don't have to do that since it says 'Automatic (feet)'. But if this creature is included in a module and then imported into a World set to metric, then that creature suddenly has a speed of 30m no?

Adjusted

Fyorl (approved):

PR #2074, opened 2023-01-17
[#2018] Replace titles with tooltips on sheets

### LINKED ISSUE #2018: Switch from `title` to `data-tooltip` on actor and item sheets

arbron:
Resolves #2018

### DIFF
--- a/templates/actors/character-sheet.hbs
+++ b/templates/actors/character-sheet.hbs
@@ -36,7 +36,7 @@
                 </li>
                 <li>
                     {{#if labels.background}}
-                        <span data-tooltip="{{localize 'DND5E.Background'}}">{{labels.background}}</span>
+                        <span data-tooltip="DND5E.Background">{{labels.background}}</span>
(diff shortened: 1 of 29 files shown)


templates/actors/character-sheet.hbs
```diff
@@ -36,7 +36,7 @@
                 </li>
                 <li>
                     {{#if labels.background}}
-                        <span title="{{localize 'DND5E.Background'}}">{{labels.background}}</span>
+                        <span data-tooltip="{{localize 'DND5E.Background'}}">{{labels.background}}</span>
```
krbz999, line 39: I don't think `localize` is necessary here, the tooltip should translate the strings automatically with just `data-tooltip="DND5E.Background"`.

arbron: Good point, though I'll still have to leave `localize` in places where we pass in parameters.

Fyorl (approved):

PR #2081, opened 2023-01-19
[#1564] Convert property attribution to use tooltips

### LINKED ISSUE #1564: Convert PropertyAttribution to use new tooltips system
Imported from Gitlab

arbron:
Makes the property attributions more consistent with other parts of the sheet and fixes the rendering issue that was preventing the expansion of attributions to other parts of the sheet.

Resolves #1564

### DIFF
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -562,56 +562,56 @@ h5 {
 /* ----------------------------------------- */
 /*  Property Attribution                     */
 /* ----------------------------------------- */
-.property-attribution {
-  display: none;
-  position: absolute;
+#tooltip.property-attribution {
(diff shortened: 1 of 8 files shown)


Fyorl (approved):
Code looks good. Since this is a deprecation, should we wait for 2.2? Seems like a pretty small potential blast radius though.

arbron:
> Code looks good. Since this is a deprecation, should we wait for 2.2? Seems like a pretty small potential blast radius though.

There shouldn't be any actual change in behavior for 3rd party sheets using this, other than the deprecation warning, so I'd probably say it is good now. Up to you.

PR #2088, opened 2023-01-21
[#244] Preserve expanded state of items on actor sheets across re-render

### LINKED ISSUE #244: Preserve the expanded/collapsed state of owned item descriptions on either side of an actor sheet re-render workflow.

As per title; changing something in one of these sections refreshes the view, which collapses sheet items one had expanded for easier viewing. This makes for a frustrating experience.

arbron:
Tracks which sections are expanded using the new `ActorSheet5e#_expanded` property.

Moves the actual HTML for the expanded sections into its own template so it can be easily rendered with the other templates.

Resolves #244

### DIFF
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -39,6 +39,15 @@ export default class ActorSheet5e extends ActorSheet {
 
   /* -------------------------------------------- */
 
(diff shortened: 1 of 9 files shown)


Fyorl (approved):
Very nice and neat, thank you, Jeff

PR #2093, opened 2023-01-22
feat: add control to skip initiative dialog using modifier keys

matjam:
# What?

This PR adds a control that is defaulted to false that can be configured per client to enable the alt/shift processing of initiative rolls via the combat tracker.

## Feedback

I'm a BE dev by trade, so JS is outside my comfort zone, so I am ok with (gentle) feedback on the code.

## TODO

The one missing piece to this PR is that I think the modifiers should be configurable. Right now it's hard-coded, but if someone can point me to documentation or code that explains how to create a new keybinding entry and use it within the system, that would be great.

## Discussion

I realize this may be a point of contention, given [this comment by @arbron ](https://discord.com/channels/170995199584108546/670336046164213761/1066629882639044658)

The rationale here is that having a dialog box pop up every time, at least for me and for many others, is disruptive to the workflow. Players can easily be trained on using alt/shift if they find it annoying, and can toggle the feature on. Additionally, modules like midi-qol automate a lot of things precisely to be able to avoid having lots of pop-ups for people to click for things to happen. The current implementation removed the ability for the `advantageMode` to be calculated and set by a module. I'm not sure this PR fixes that; maybe we should have an event that can be hooked for this purpose.

I can probably make this a module if we're absolutely set on not having this feature, but I still would argue strongly at the very least for a hook so `_onCombatantControl()` doesn't need to be wrapped by a module.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1512,13 +1512,16 @@ export default class Actor5e extends Actor {
   async rollInitiativeDialog(rollOptions={}) {
     // Create and configure the Initiative roll
     const roll = this.getInitiativeRoll(rollOptions);
-    const choice = await roll.configureDialog({
-      defaultRollMode: game.settings.get("core", "rollMode"),
-      title: `${game.i18n.localize("DND5E.InitiativeRoll")}: ${this.name}`,
-      chooseModifier: false,
-      defaultAction: rollOptions.advantageMode ?? dnd5e.dice.D20Roll.ADV_MODE.NORMAL
-    });
-    if ( choice === null ) return; // Closed dialog
+
+    if (rollOptions.advantageMode === null) {
+      const choice = await roll.configureDialog({
+        defaultRollMode: game.settings.get("core", "rollMode"),
+        title: `${game.i18n.localize("DND5E.InitiativeRoll")}: ${this.name}`,
+        chooseModifier: false,
+        defaultAction: rollOptions.advantageMode ?? dnd5e.dice.D20Roll.ADV_MODE.NORMAL
+      });
+      if ( choice === null ) return; // Closed dialog
+    }
 
     // Temporarily cache the configured roll and use it to roll initiative for the Actor
     this._cachedInitiativeRoll = roll;
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1513,7 +1513,7 @@ export default class Actor5e extends Actor {
     // Create and configure the Initiative roll
     const roll = this.getInitiativeRoll(rollOptions);
 
-    if (rollOptions.advantageMode === null) {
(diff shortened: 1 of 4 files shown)


module/settings.mjs
```diff
@@ -90,6 +90,16 @@ export default function registerSystemSettings() {
     type: Boolean
   });
 
+  // Use keyboard controls to set initiative advantage
+  game.settings.register("dnd5e", "initiativeAdvantageKeyboardControl", {
```
kaelad02, line 94: I don't see why we need a setting to control this considering we don't have it for any other rolls (e.g. ability checks, attack rolls).

module/documents/actor/actor.mjs
```diff
@@ -1512,13 +1512,16 @@ export default class Actor5e extends Actor {
...
     const roll = this.getInitiativeRoll(rollOptions);
-    const choice = await roll.configureDialog({
-      defaultRollMode: game.settings.get("core", "rollMode"),
-      title: `${game.i18n.localize("DND5E.InitiativeRoll")}: ${this.name}`,
-      chooseModifier: false,
-      defaultAction: rollOptions.advantageMode ?? dnd5e.dice.D20Roll.ADV_MODE.NORMAL
-    });
-    if ( choice === null ) return; // Closed dialog
+
+    if (rollOptions.advantageMode === null) {
```
kaelad02, line 1516: This would introduce a difference between how initiative checks work compared to all other d20 rolls. Looking at how the `d20Roll` function works, and the `determineAdvantageMode` function it uses, it requires either a `fastForward` property or holding down one of the special keys (e.g. alt for advantage). This skips the dialog if the `advantageMode` is set rather than using an explicit `fastForward` or one of the key presses.

For the Advantage Reminder module, I'm looking at setting the `advantageMode` property and I think it would be a bad user experience to fast forward the roll in this case. And while not related to your PR, while I understand we can't use `d20Roll` in this case since we need to return an unrolled `Roll` instance, I think we should keep it as close to that for consistency sake for module/macro developers.

matjam: Yup, just pushed a change; it's using the `determineAdvantageMode()` function and rather sets the `isFF` property instead.

Given the new changes, what do you suggest? I am not super comfortable in the code-base - I've just started - so I have no idea how the existing module/macro developers use things.

module/settings.mjs
matjam: Huh, I guess modules are patching in that behavior themselves. Maybe this would be better as a standalone module.

matjam:
Gonna close this PR for now. I will maybe just make a module for this, as it really is a narrow thing. And I couldn't really come up with a good answer for

> I don't see why we need a setting to control this considering we don't have it for any other rolls (e.g. ability checks, attack rolls).

from @kaelad02.

PR #2103, opened 2023-01-24
[#2041] Disable fields that are overridden by an Active Effect and display a tooltip explaining why they cannot be edited

### LINKED ISSUE #2041: Show a warning notification when attempting to change an overridden armour calculation, or have the field be disabled with a tooltip
If 'Unarmored Defense' (from both Monk and Barbarian) from the 'Class Features (SRD)' is added to the character sheet, it is no longer possible to change the AC calculation to anything else.
'Unarmored Defense' should only be active if no armor is being worn, however, equipping armor does not change the AC, and it's not possible to change the AC calculation to 'Equipped Armor'.
I believe that the issue is that the 'Change Mode' of the item's effect is set to 'Override'. I believe that it should either not be set to 'Override', or it should automatically switch to 'Equipped Armor' if the user has equipped any.

Tested on Foundry v10.291 and 5e v2.1.1 and v2.0.2

Fyorl:
- Closes #2041

### DIFF
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -459,6 +459,7 @@ export default class ActorSheet5e extends ActorSheet {
 
     // Format a spellbook entry for a certain indexed level
     const registerSection = (sl, i, label, {prepMode="prepared", value, max, override}={}) => {
+      const aeOverride = foundry.utils.hasProperty(this.actor.overrides, `system.spells.spell${i}.override`);
       spellbook[i] = {
         order: i,
         label: label,
@@ -470,7 +471,8 @@ export default class ActorSheet5e extends ActorSheet {
         slots: useLabels[i] || max || 0,
         override: override || 0,
         dataset: {type: "spell", level: prepMode in sections ? 1 : i, "preparation.mode": prepMode},
-        prop: sl
+        prop: sl,
+        editable: context.editable && !aeOverride
       };
     };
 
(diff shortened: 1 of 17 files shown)


aaclayton (approved):
Looks okay at surface level. If it's working in practice go ahead!

module/applications/actor/base-sheet.mjs
```diff
@@ -643,6 +643,13 @@ export default class ActorSheet5e extends ActorSheet {
 
       // Active Effect management
       html.find(".effect-control").click(ev => ActiveEffect5e.onManageActiveEffect(ev, this.actor));
+
+      for ( const override of Object.keys(foundry.utils.flattenObject(this.actor.overrides)) ) {
+        html.find(`input[name="${override}"],select[name="${override}"]`).each((i, el) => {
+          el.disabled = true;
+          el.dataset.tooltip = "DND5E.ActiveEffectOverrideWarning";
```
arbron, line 650: For fields that have an existing tooltip, do we want to take that into account? Maybe prepending the tooltip with it?

arbron (commented):
Seems good for most things, but doesn't work on the proficiency buttons on the character sheet because they aren't normal inputs.

module/applications/actor/base-sheet.mjs
Fyorl: I tried a version where we didn't replace existing tooltips, and a version where we did (this version). I felt that replacing the tooltip worked best from a UX perspective. I didn't try a version where we concatenate the values at all, but it feels like that would be too long and also read strangely.

Fyorl:
> Seems good for most things, but doesn't work on the proficiency buttons on the character sheet because they aren't normal inputs.

I will add some special-case handling for proficiency buttons.

module/applications/actor/base-sheet.mjs
krbz999: Curious if this method covers the case where the input field isn't present, such as spell slot overrides, where the input field is created by clicking the 'edit' icon.

Fyorl: > Curious if this method covers the case where the input field isn't present, such as spell slot overrides, where the input field is created by clicking the 'edit' icon.

I will add some handling for this too 😩. Any others you can think of?

krbz999: Anything in Special Traits and other sub-applications?

Fyorl: Those are already handled with the current version.

arbron (approved):

PR #2104, opened 2023-01-24
Fix various linting issues

arbron:
Just a few code style cleanup issues.

### DIFF
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -1127,7 +1127,7 @@ export default class ActorSheet5e extends ActorSheet {
     event.preventDefault();
     const itemId = event.currentTarget.closest(".item").dataset.itemId;
     const item = this.actor.items.get(itemId);
-    if ( item ) return item.use({}, {event});
+    return item.use({}, {event});
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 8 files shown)


module/applications/actor/base-sheet.mjs
```diff
@@ -402,7 +402,7 @@ export default class ActorSheet5e extends ActorSheet {
       const physical = [];
       if ( data.bypasses?.size ) {
         values = values.filter(t => {
-          if ( !CONFIG.DND5E.physicalDamageTypes[t] ) return true;
+          if ( !globalThis.CONFIG.DND5E.physicalDamageTypes[t] ) return true;
```
Fyorl, line 405: Not sure the `globalThis` prefix is necessary to access `CONFIG` here, or below.

Fyorl (changes requested):

module/applications/actor/base-sheet.mjs
arbron: It’s the only way I could find that would fix the `no-loop-func` ESLint warning that was being incorrectly thrown here. The other option would be to manually annotate it to skip that warning.

Fyorl: Honestly, I might vote to just turn off that warning entirely. I don't think I've ever had it flag something that wasn't a false positive. What do you think?

arbron: I have no objections to that :)

Fyorl (approved):

Issue #2094, opened 2023-01-22
Shift type-specific logic into data models where possible


arbron:
- [x] Move Item getters
- [x] Move chat properties
- [x] Move Item preparation methods
- [x] Move Actor preparation methods
  - [x] Actor base data
  - [x] Actor derived data
  - [ ] ~~Actor spellcasting~~ (most of this logic has already been moved out to the spellcasting method data models, so this should be okay as is)
- [x] Move preCreate/Update/Delete handlers
  - [x] Actor
  - [x] Item

JPMeehan:
`SpellData#getCardData` returns `context.isSpell = true`, but the places `isSpell` gets used just do a plain type check for `type === "spell"`. This makes it difficult for module subtypes that are spell-like (psionic powers, martial maneuvers from 3pp) to display info as richly as the system-native spells.

JPMeehan:
Similar issue with property sorting - right now it's a hard check of `if ( item.type !== "spell" ) context.properties = sortObjectEntries(context.properties, "label");`, but having this be a (static?) property of a module subtype would allow for more customization

PR #2110, opened 2023-01-25
[#2094] Move Item getters into system data models

arbron:
Moves most of the getter logic from `Item5e` into the various system data models and splits up `abilityMod` and `criticalThreshold` so that each individual type can handle their own type-specific logic.

Adjust the logic of `hasDamage`, `hasSave`, and `isVersatile` to require that `actionType` be set.

Adds `isMountable`, `hasScalarDuration`, `hasScalarRange`, and `hasScalarTarget` getters.

This also refactors `ItemSheet5e` & templates to make more use of getters directly and reorders the getters in `Item5e` to be in alphabetical order.

### DIFF
--- a/module/data/item/weapon.mjs
+++ b/module/data/item/weapon.mjs
@@ -108,4 +108,15 @@ export default class WeaponData extends SystemDataModel.mixin(
   get _typeCriticalThreshold() {
     return this.parent?.actor?.flags.dnd5e?.weaponCriticalThreshold ?? Infinity;
   }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Is this item a separate large object like a siege engine or vehicle component that is
+   * usually mounted on fixtures rather than equipped, and has its own AC and HP?
+   * @type {boolean}
+   */
+  get isMountable() {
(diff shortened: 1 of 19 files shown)


module/data/item/weapon.mjs
```diff
@@ -81,4 +83,40 @@ export default class WeaponData extends SystemDataModel.mixin(
...
     if ( source.weaponType === null ) source.weaponType = "simpleM";
   }
+
+  /* -------------------------------------------- */
+  /*  Getters                                     */
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  get _typeAbilityMod() {
+    // Ranged weapons - Dex (PHB pg. 194)
```
Fyorl, line 93: Let's remove these PHB references

Fyorl (approved):
LGTM. Only one comment, and you can merge once that's taken care of.

krbz999:
This doesn't seem to touch `activation.type`, meaning items can still have an area of effect or limited uses despite having Activation Type: none.

This doesn't seem quite right here.

image

PR #2113, opened 2023-01-26
[#774] Refactor tools implementation to be more like skills

Fyorl:
The changes here address two main issues in the system currently:

* If a character has expertise in a tool, that can only be recorded on the tool item itself. Removing the tool item therefore removes the character's record of expertise with that tool (#774).
* It's not currently possible for a character to make a check with one of their vehicle proficiencies, which is required when operating various vehicles.

Instead of only storing proficiency and ability information on tool items, we now store that information on the actor itself, and move the character's record of tool proficiencies from `traits.toolProf` to the `tools` field. A lot of the code to do with rolling and configuring skills has been reused and merged with tools, which they closely resemble in game terms.

image


aaclayton (commented):
I'll try and do a more complete code review later, but I have a philosophical comment/question to share first. It seems to me that there is a parity here with the way that armor and weapons work. 

With Armor/Weapons we store the character's proficiencies on the Actor, but we *also* record the proficiency flag on the item itself. This change moves tools to a place where we are no longer recording proficiency on the Item.

I think whichever way we go, we should have consistency in this regard.

aaclayton:
Also - shouldn't the default ability score which is most typically used with a certain tool remain something that can be configured on the tool item itself?

Fyorl:
> This change moves tools to a place where we are no longer recording proficiency on the Item.

Oops, that was some earlier experimentation that I forgot to rollback before committing. The intention here is that we achieve consistency across weapons, armour, and tools as you point out. The proficiency checkbox on weapons and armour will change to a drop-down with three values, something like 'Determined Automatically', 'Proficient', 'Not Proficient', with the first value controlled by the actor's proficiencies. The same will then happen for tools, where we add an additional 'Determined Automatically' option to the proficiency drop-down.

> Also - shouldn't the default ability score which is most typically used with a certain tool remain something that can be configured on the tool item itself?

I feel that the tool ability should be configurable on the actor, just like with skills, since tool proficiencies much more closely resemble skill proficiencies in game terms, than they do weapon or armour proficiencies. There's no reason we can't have it configurable on the item additionally though, but with an added 'default' option, like with weapons, that will use the actor's configured ability for that tool.

templates/items/tool.hbs
```diff
@@ -84,7 +84,7 @@
             <div class="form-group">
                 <label>{{ localize "DND5E.DefaultAbilityCheck" }}</label>
                 <select name="system.ability">
-                    {{selectOptions config.abilities selected=system.ability}}
+                    {{selectOptions config.abilities selected=system.ability blank=(localize "DND5E.Default")}}
```
arbron, line 87: Seems like there should also be a default or automatic option for proficiency.

module/documents/item.mjs
```diff
@@ -1739,79 +1739,7 @@ export default class Item5e extends Item {
...
-     * A hook event that fires after a tool check has been rolled for an Item.
-     * @function dnd5e.rollToolCheck
-     * @memberof hookEvents
-     * @param {Item5e} item   Item for which the roll was performed.
-     * @param {D20Roll} roll  The resulting roll.
-     */
-    if ( roll ) Hooks.callAll("dnd5e.rollToolCheck", this, roll);
-
-    return roll;
+    return this.actor?.rollToolCheck(this.system.baseItem, {ability: this.system.ability, ...options});
```
arbron, line 1742: It seems like the `"Tool Bonus"` field on tool items themselves is ignored.

module/applications/actor/tool-selector.mjs
```diff
@@ -0,0 +1,39 @@
...
+ * A specialized version of the TraitSelector used for selecting tool and vehicle proficiencies.
+ * @extends {TraitSelector}
+ */
+export default class ToolSelector extends TraitSelector {
+  /** @inheritdoc */
+  async getData() {
+    return {
+      ...super.getData(),
+      choices: await Trait.choices(this.trait, Object.entries(this.document.system.tools).reduce((set, [key, tool]) => {
+        if ( tool.value > .5 ) set.add(key);
```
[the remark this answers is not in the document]
Fyorl: You are right. The checkboxes should control what appears on the sheet, with proficiency kept separate.

templates/items/tool.hbs
Fyorl: That will be coming in a followup PR. I intend to do an 'automatic proficiencies' PR across weapons, armour, and tools, but it's too big for one PR. For now, the proficiency on the tool is just ignored.

arbron (approved):

aaclayton (approved):
Conceptually I'm good with this direction of change. I haven't reviewed the code in detail, would that level of review be useful from me, or were you most interested in verifying that I'm aligned with the basic design?

Fyorl:
> Conceptually I'm good with this direction of change. I haven't reviewed the code in detail, would that level of review be useful from me, or were you most interested in verifying that I'm aligned with the basic design?

So long as you're ok with the high-level design, you can skip the in-depth review, I think, since Jeff has done that.

PR #2122, opened 2023-01-27
[#2119, #2121] Support spells above level 9, allow pact progression customization

### LINKED ISSUE #2119: Spell levels above 9 not supported
There are a few places in the spell preparation code that are hardcoded to assume the maximum spell level is 9. It would be great to make this more flexible.

### LINKED ISSUE #2121: Allow pact casting progression to be configured
Currently modifying pact casting progression involves creating a totally custom `preparePactSlots` method and overriding the default using the `DND5E.preparePactSlots` hook. While this works, it would be nice if this calculation logic was available in the config to more easily add support for things like pact casters above 20th level.

arbron:
Removes the default spell levels defined in template and adjusts `prepareLeveledSlots` so it is based on the spell levels defined in config, not in the actor data. Also refactors a few other places that were hard-coded to handle nine levels of spells.

Resolves #2119 

Moves pact progression configuration out of the `preparePactSlots` method into config for easier customization.

Resolves #2121


module/config.mjs
```diff
@@ -1078,6 +1078,31 @@ DND5E.SPELL_SLOT_TABLE = [
...
+ * @typedef {object} PactProgressionConfig
+ * @property {number} slots  Number of spell slots granted.
+ * @property {number} level  Level of spells that can be cast.
+ */
+
+/**
+ * Define the pact slot & level progression by pact caster level.
+ * @enum {PactProgressionConfig}
+ */
+DND5E.pactCastingProgression = {
```
krbz999, line 1093: Part of me is wondering if this is a perfect opportunity to make a nested object to allow for multiple pact progressions.
```js
DND5E.pactCastingProgression = {
  regular: {...},
  someOther: {...}
};
```

arbron: It seems to me that would be better handled by custom spellcasting types using the hooks.

krbz999:
What happens if you attempt to make a spell scroll from a 10th level spell? `DND5E.spellScrollIds` only goes up to 9.

Fyorl (approved):

Issue #532, opened 2020-05-28
Edit physical item quantity in the sheet directly instead of having to open the item sheet.


aaclayton:
Originally in GitLab by @Tercept

Make the inventory item quantities quick-editable inline without needing to open the item editor to edit.

Issue #2132, opened 2023-01-30
Group inventory items can be expanded repeatedly. Expanded items do not stay expanded when the sheet is re-rendered.


krbz999:
To reproduce: Click item. Click item. Click item. Click item--

![image](https://user-images.githubusercontent.com/50169243/215528237-2f2dcaac-d357-4b0f-b9c1-e6e4d9786774.png)

Second part: Expand an item. Then cause the sheet to rerender.

PR #2133, opened 2023-01-30
[#532] Allow for editing the quantity of an item directly on the character sheet

krbz999:
Note that this touches the `character` sheet only, but by extension also the `group` sheet, where it is currently non-functional.

image

### DIFF
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -123,7 +123,7 @@ export default class ActorSheet5e extends ActorSheet {
     /** @deprecated */
     Object.defineProperty(context, "data", {
       get() {
-        const msg = `You are accessing the "data" attribute within the rendering context provided by the ActorSheet5e
+        const msg = `You are accessing the "data" attribute within the rendering context provided by the ActorSheet5e 
         class. This attribute has been deprecated in favor of "system" and will be removed in a future release`;
         foundry.utils.logCompatibilityWarning(msg, { since: "DnD5e 2.0", until: "DnD5e 2.2" });
         return context.system;
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -123,7 +123,7 @@ export default class ActorSheet5e extends ActorSheet {
     /** @deprecated */
     Object.defineProperty(context, "data", {
       get() {
-        const msg = `You are accessing the "data" attribute within the rendering context provided by the ActorSheet5e 
+        const msg = `You are accessing the "data" attribute within the rendering context provided by the ActorSheet5e
         class. This attribute has been deprecated in favor of "system" and will be removed in a future release`;
         foundry.utils.logCompatibilityWarning(msg, { since: "DnD5e 2.0", until: "DnD5e 2.2" });
         return context.system;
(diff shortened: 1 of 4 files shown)


krbz999:
> Quantity column looks good. I'm not clear on why we're removing the Charges column from the group actor sheet though.

My idea was that the group actor had a focus on inventory management, not item management, if that makes sense. I also couldn't come up with a use case for managing limited uses on an item on a group actor.

I can add it back easily, though adding the functionality of adjusting the limited uses results in even more copying and pasting from the base actor.

(532 should also be linked)

Fyorl:
> My idea was that the group actor had a focus on inventory management, not item management, if that makes sense. I also couldn't come up with a use case for managing limited uses on an item on a group actor.

I'm sure there could be items with charges that a group manages in a shared way. Like a Wand of Tiny Hut or something.

> I can add it back easily, though adding the functionality of adjusting the limited uses results in even more copying and pasting from the base actor.

There was some mention in the original group actor PR about creating an additional abstract parent sheet implementation that we could use to share code more easily. How much code duplication would there be to just display the charges on the sheet but not have them editable? I'm ok with removing the charges column entirely if it's nonfunctional without code duplication, and then we add it back in once we've created the common sheet class to share code across.

krbz999:
> How much code duplication would there be to just display the charges on the sheet but not have them editable?

The code duplication would be in making them save their changed values, not in displaying them. I feel like just adding them back later is a better long-term solution rather than repeated changes.

arbron:
@krbz999 I brought your fix for #2132 into the 2.1.5 release, so when you rebase you should be able to remove references to that particular fix.

module/applications/actor/base-sheet.mjs
```diff
@@ -860,7 +861,9 @@ export default class ActorSheet5e extends ActorSheet {
     const value = input.value;
     if ( ["+", "-"].includes(value[0]) ) {
       const delta = parseFloat(value);
-      input.value = Number(foundry.utils.getProperty(this.actor, input.name)) + delta;
+      const item = this.actor.items.get(input.closest("[data-item-id]")?.dataset.itemId);
+      if ( !item ) input.value = Number(foundry.utils.getProperty(this.actor, input.name)) + delta;
+      else input.value = Number(foundry.utils.getProperty(item, input.dataset.name)) + delta;
```
Fyorl, line 866: Prefer to switch these so that the `else` statement does not read like a double-negative:
```suggestion
      if ( item ) input.value = Number(foundry.utils.getProperty(item, input.dataset.name)) + delta;
      else input.value = Number(foundry.utils.getProperty(this.actor, input.name)) + delta;
```

module/applications/actor/base-sheet.mjs
```diff
@@ -1165,6 +1168,23 @@ export default class ActorSheet5e extends ActorSheet {
 
   /* -------------------------------------------- */
 
+  /**
+   * Change the quantity of an Owned Item within the actor.
+   * @param {Event} event        The triggering click event.
+   * @returns {Promise<Item5e>}  Updated item.
+   * @private
```
Fyorl, line 1175: This should be `@protected`. So should the two methods above it too, now that I look at it.

module/applications/actor/group-sheet.mjs
```diff
@@ -353,6 +360,21 @@ export default class GroupActorSheet extends ActorSheet {
...
   /* -------------------------------------------- */
 
+  /**
+   * Change the quantity or limited uses of an Owned Item within the actor.
+   * @param {Event} event        The triggering click event.
+   * @returns {Promise<Item5e>}  Updated item.
+   * @private
+   */
+  async _onQuantityChange(event) {
+    const proto = game.system.applications.actor.ActorSheet5e.prototype;
```
Fyorl, line 370: It's fine to call into the `ActorSheet5e` `prototype` for now, but we should call this method something different since it doesn't pertain only to quantity anymore. Something like `_onInlineItemChange` maybe. I'm open to suggestions.

Fyorl (changes requested):

PR #1696, opened 2022-08-11
[#79] Hide name/desc for items not identified

ccjmk:
Resolves #79 

Used `system.identified` to toggle between `description.value` and `description.unidentified` and between showing true name or a masked 'Unidentified XYZ' faked name in item sheet (weapon/equipment/consumable) and actor inventory.

Possible further steps:

* Leveraging baseItem on weapons and equipments to improve masked name, for example turning "Unidentified Martial Meele" into "Unidentified Rapier" or "Unidentified Medium Armor" to "Unidentified Breastplate".
* Showing a fake icon if not identified; either `mystery-man` or saving a different icon akin of the description treatment.
* Hiding possible spoiler data, ranging from values on the Description tab like Price and Properties, to hiding the Effects tab and everythin on the details tab save for the basic equipment details. Maybe even just showing the Identified flag on the Details tab at all.
* Following [a suggestion by Arcanist](https://discord.com/channels/732325252788387980/787815954725732393/1006991456998793237) in the League's discord, a context menu could be added for the GM to identify/unidentify items on the sidebar (similar to Forien's mystify mechanics)
* Some way of showing on the sidebar if an item is identified or not, like a little FA icon next to the name or showing the masked name with some style, and revealing the true name on hover for the GM, or the likes.

PR #2137, opened 2023-01-30
[#2135, #2136] Add multi editor to make additional descriptions editable

### LINKED ISSUE #2135: Add ability to edit "hidden" descriptions
Actors and items all have various "hidden" descriptions that are available in the data model but cannot be editor or viewed ("biography" vs. "public biography" on normal actors, "full description" vs. "summary" on group actors, "description" vs. "unidentified description" vs. "chat message" on items). These various descriptions should be exposed so players can use all of them.

### LINKED ISSUE #2136: Use item's "chat description" if specified
Items have `system.description.chat` in their data but that information isn't used in any way. The system should automatically use this description in place of the full description in chat messages if present, rather than requiring people to use some combination of hidden fields like they do currently.

arbron:
Creates a new `{{multiEditor}}` handlebar helper that takes configuration objects (prepared by the `{{source}}` helper) to display multiple descriptions within a single editor:

https://user-images.githubusercontent.com/19979839/215570200-dc93ab75-9a06-4e0d-a6dc-917fe840379a.mov

Also displays the contents of `system.description.chat` for item chat messages if present to allow for custom chat description.

Resolves #2135 
Resolves #2136


ccjmk:
I'm interested in using this for #1696 once this PR goes through, it would make for a much neater UX/UI imo.

Fyorl:
I think my only hangup at the moment is that I'm not sold on the UI of selecting a description to edit via a drop-down. In some discussions around this feature it was suggested that a slide-toggle might be nice for most cases where there's only two descriptions to choose from. Something like `fa-toggle-on` and `fa-toggle-off`. The problem is that what I think will be perhaps the most-used alternative description is the item unidentified description, which unfortunately has 3 descriptions (normal, unidentified, and chat), so a slide-toggle cannot work, and we would need to fall back to a drop-down.

So first maybe we could consider whether we need the `chat` description at all? Its use case is fairly obvious but it seems like more of a nice-to-have, rather than a necessary feature. I personally quite like the idea of being able to quickly toggle between identified and unidentified description as a DM, that feels like a nice UX to me.

Alternatively, because I also quite liked what @ccjmk proposed in #1696 with the ability to see all descriptions at once, we could potentially do something like an accordion style UI (maybe using the `<details>` element or something?)

image

Having multiple editors in the same tab can be quite awkward to style, so I would propose that we side-step the issue by having the edit button instead pop out a separate editor dialog that is edit-only mode (similar to how journal page editing works.)

krbz999:
This seems like a use case for tabs.

However, if a pop-out text editor would allow collaboration that is to me 100% the ideal.

arbron:
@Fyorl Here is a quick prototype of showing all descriptions inline and replacing the whole tab with an editor:

https://github.com/foundryvtt/dnd5e/assets/19979839/72508c62-4a4e-42e0-9ec9-5ba36091cfe9

On locked sheets it hides the first header to avoid clutter:

Locked Headers

Fyorl:
> @Fyorl Here is a quick prototype of showing all descriptions inline and replacing the whole tab with an editor:

Looks pretty good to me. One minor point of feedback would be to have the edit button always visible so it's more obvious what you need to do in order to edit the description rather than having to hover over it to find out. Otherwise I think we should be good to proceed in this direction.

arbron:
Closed in favor of a new PR: https://github.com/foundryvtt/dnd5e/pull/2137

PR #2143, opened 2023-01-31
[#1082] Add proficiency automation for weapons, armour, and tools.

### LINKED ISSUE #1082: Additions to Actor proficiencies should automatically update current owned Items which should now be flagged as proficient.

### LINKED ISSUE #2305: Tool Item - Tool Proficiency field
The Tool Proficiency field in a Tool Item has no effect on the roll, the proficiency bonus is governed by the Actor's Tool Proficiency traits
I think that when a Tool item is placed on an actor the field should be set by the Actor's traits, but anything set in that Item should take precedence over the Actor's traits. 

image

### LINKED ISSUE #2327: Outdated `system.traits.toolProf` in `item.mjs` causes dropped tools to never set proficiency.
```js
const actorToolProfs = this.parent.system.traits?.toolProf?.value || new Set();
  ```
Should use `system.tools` now.

### PR DESCRIPTION
**Pull request by Fyorl**
- Closes #1082 
- Closes #2327 
- Closes #2305

### DIFF
--- a/module/data/item/equipment.mjs
+++ b/module/data/item/equipment.mjs
@@ -150,9 +150,8 @@ export default class EquipmentData extends SystemDataModel.mixin(
   /**
    * The proficiency multiplier for this item.
    * @returns {number}
-   * @protected
    */
-  get _proficiencyMultiplier() {
+  get proficiencyMultiplier() {
     if ( Number.isFinite(this.proficient) ) return this.proficient;
     const actor = this.parent.actor;
     if ( !actor ) return 0;
--- a/module/data/item/equipment.mjs
+++ b/module/data/item/equipment.mjs
@@ -26,7 +26,7 @@ import MountableTemplate from "./templates/mountable.mjs";
  * @property {string} speed.conditions  Conditions that may affect item's speed.
  * @property {number} strength          Minimum strength required to use a piece of armor.
  * @property {boolean} stealth          Does this equipment grant disadvantage on stealth checks when used?
- * @property {boolean} proficient       Does the owner have proficiency in this piece of equipment?
+ * @property {number} proficient        Does the owner have proficiency in this piece of equipment?
  */
 export default class EquipmentData extends SystemDataModel.mixin(
   ItemDescriptionTemplate, PhysicalItemTemplate, EquippableItemTemplate,
@@ -51,7 +51,9 @@ export default class EquipmentData extends SystemDataModel.mixin(
         required: true, integer: true, min: 0, label: "DND5E.ItemRequiredStr"
       }),
       stealth: new foundry.data.fields.BooleanField({required: true, label: "DND5E.ItemEquipmentStealthDisav"}),
-      proficient: new foundry.data.fields.BooleanField({required: true, initial: true, label: "DND5E.Proficient"})
+      proficient: new foundry.data.fields.NumberField({
+        required: true, min: 0, max: 1, integer: true, initial: null, label: "DND5E.ProficiencyLevel"
+      })
     });
   }
 
@@ -64,6 +66,7 @@ export default class EquipmentData extends SystemDataModel.mixin(
     super.migrateData(source);
     EquipmentData.#migrateArmor(source);
     EquipmentData.#migrateStrength(source);
+    EquipmentData.#migrateProficient(source);
   }
 
   /* -------------------------------------------- */
@@ -93,6 +96,16 @@ export default class EquipmentData extends SystemDataModel.mixin(
     if ( typeof source.strength !== "string" ) return;
     if ( source.strength === "" ) source.strength = null;
     if ( Number.isNumeric(source.strength) ) source.strength = Number(source.strength);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Migrate the proficient field to convert boolean values.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static #migrateProficient(source) {
+    if ( typeof source.proficient === "boolean" ) source.proficient = Number(source.proficient);
   }
 
   /* -------------------------------------------- */
@@ -132,4 +145,22 @@ export default class EquipmentData extends SystemDataModel.mixin(
     return this.armor.type === "vehicle";
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * The proficiency multiplier for this item.
+   * @returns {number}
+   * @protected
+   */
+  get _proficiencyMultiplier() {
+    if ( Number.isFinite(this.proficient) ) return this.proficient;
+    const actor = this.parent.actor;
+    if ( !actor ) return 0;
+    if ( actor.type === "npc" ) return 1; // NPCs are always considered proficient with any armor in their stat block.
+    const config = CONFIG.DND5E.armorProficienciesMap;
+    const itemProf = config[this.armor?.type];
+    const actorProfs = actor.system.traits?.armorProf?.value ?? new Set();
+    const isProficient = (itemProf === true) || actorProfs.has(itemProf) || actorProfs.has(this.baseItem);
+    return Number(isProficient);
+  }
 }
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: module/documents/item.mjs [hunk 1]
```diff
@@ -470,6 +470,48 @@ export default class Item5e extends Item {
...
+        multiplier = Math.max(baseItemProf?.value ?? 0, categoryProf?.value ?? 0);
+      } else {
+        const type = this.type === "equipment" ? this.system.armor?.type : this.system.weaponType;
+        const config = CONFIG.DND5E[`${this.type === "equipment" ? "armor" : "weapon"}ProficienciesMap`];
+        const itemProf = config[type];
+        const actorProfs = traits?.[this.type === "equipment" ? "armorProf" : "weaponProf"]?.value ?? new Set();
+        const isProficient = (itemProf === true) || actorProfs.has(itemProf) || actorProfs.has(this.system.baseItem);
+        multiplier = Number(isProficient);
+      }
+    }
```
arbron, line 508: All this can be brought up a level.

```suggestion
    // Tools based on actor's proficiencies.
    else if ( this.type === "tool" ) {
      const baseItemProf = tools?.[this.system.baseItem];
      const categoryProf = tools?.[this.system.toolType];
      multiplier = Math.max(baseItemProf?.value ?? 0, categoryProf?.value ?? 0);
    }

    // Otherwise, determine proficiency based on the actor's proficiencies.
    else {
      const type = this.type === "equipment" ? this.system.armor?.type : this.system.weaponType;
      const config = CONFIG.DND5E[`${this.type === "equipment" ? "armor" : "weapon"}ProficienciesMap`];
      const itemProf = config[type];
      const actorProfs = traits?.[this.type === "equipment" ? "armorProf" : "weaponProf"]?.value ?? new Set();
      const isProficient = (itemProf === true) || actorProfs.has(itemProf) || actorProfs.has(this.system.baseItem);
      multiplier = Number(isProficient);
    }
```

module/documents/item.mjs
```diff
@@ -470,6 +470,48 @@ export default class Item5e extends Item {
 
   /* -------------------------------------------- */
 
+  /**
+   * Determine an item's proficiency level based on its parent actor's proficiencies.
+   * @protected
+   */
+  _prepareProficiency() {
```
arbron, line 477: Since this is a new method, it might be a good time to start moving preparation methods into the system data models.

We could add a `prepareProficiency` method, possibly in a new `ProficiencyTemplate` or within the existing `EquippableTemplate` (though that is also inherited by some types without proficiency), that is called by `prepareFinalAttributes` as `this.system.prepareProficiency?.()`.

Then, similar to how we are handling the `abilityMod` getter, there could be a method for getting type-specific multipliers that certain types can override. So `SpellData#_proficiencyMultiplier` would always return `1`, `ToolData#_proficiencyMultiplier` would have the tool-specific code, etc.

arbron (changes requested):
Everything seems to work great!

I have some thoughts about moving the `_prepareProficiency` method into the data models.

module/documents/item.mjs
Fyorl: I was able to utilise a `_proficiencyMultiplier` getter, which definitely felt a lot better as it was able to remove some ugly ternaries and clean up the logic in `prepareProficiency`. I wasn't able to successfully move `prepareProficiency` out of `Item5e` and into the Data Models though. I couldn't find a good place for it: The main outlier was spells. I think a `ProficiencyTemplate` that defined the `proficient` field, and had a `prepareProficiency` method would be a good way to do it. But that would mean adding the `proficient` field to spells, which I was reluctant to do.

Let me know if you have any alternate suggestions, or if you think we should reconsider putting the `proficient` field on spells.

module/data/item/equipment.mjs
```diff
@@ -51,7 +51,9 @@ export default class EquipmentData extends SystemDataModel.mixin(
         required: true, integer: true, min: 0, label: "DND5E.ItemRequiredStr"
       }),
       stealth: new foundry.data.fields.BooleanField({required: true, label: "DND5E.ItemEquipmentStealthDisav"}),
-      proficient: new foundry.data.fields.BooleanField({required: true, initial: true, label: "DND5E.Proficient"})
+      proficient: new foundry.data.fields.NumberField({
+        required: true, min: 0, max: 1, integer: true, initial: null, label: "DND5E.ProficiencyLevel"
```
aaclayton, line 55: What would it mean for this to be `nullable` and for the initial to be `null`. Should the initial be `0` and the field be `nullable: false`?

module/data/item/equipment.mjs
```diff
@@ -132,4 +145,22 @@ export default class EquipmentData extends SystemDataModel.mixin(
...
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * The proficiency multiplier for this item.
+   * @returns {number}
+   * @protected
+   */
+  get _proficiencyMultiplier() {
```
aaclayton, line 155: Perhpas this is worth considering as "public API" instead of `@protected`? Could be useful for various things.

module/data/item/tool.mjs
```diff
@@ -30,7 +30,7 @@ export default class ToolData extends SystemDataModel.mixin(
       }),
       chatFlavor: new foundry.data.fields.StringField({required: true, label: "DND5E.ChatFlavor"}),
       proficient: new foundry.data.fields.NumberField({
-        required: true, nullable: false, initial: 0, min: 0, label: "DND5E.ItemToolProficiency"
+        required: true, initial: null, min: 0, max: 2, step: 0.5, label: "DND5E.ItemToolProficiency"
```
aaclayton, line 33: Similar question here, feels like this should be `nullable: false, initial: 0` rather than null?

module/data/item/weapon.mjs
```diff
@@ -35,7 +35,9 @@ export default class WeaponData extends SystemDataModel.mixin(
       properties: new MappingField(new foundry.data.fields.BooleanField(), {
         required: true, initialKeys: CONFIG.DND5E.weaponProperties, label: "DND5E.ItemWeaponProperties"
       }),
-      proficient: new foundry.data.fields.BooleanField({required: true, initial: true, label: "DND5E.Proficient"})
+      proficient: new foundry.data.fields.NumberField({
```
aaclayton, line 38: Same question about `nullable`

module/documents/item.mjs
aaclayton: Could this be part of data preparation on the sub-type itself rather than preparation that is managed by the Item? We are assigning the `Proficiency` instance to `this.system`. Could we just shift this to sub-type prearation and instead assign like:
```js
this.prof = new Proficiency(this.parent.system.attributes.prof, this.proficiencyMultiplier);
```
Or we have to do it here because of order-of-operations?

template.json
```diff
@@ -685,7 +685,7 @@
       },
       "strength": null,
       "stealth": false,
-      "proficient": true
+      "proficient": null
```
aaclayton, line 688: This is fine for the purposes of this PR, but I would encourage you to consider ditching the template.json file (or rather just minimizing it) in `2.3.0` since we have data models now.

Example: https://github.com/foundryvtt/crucible/blob/master/template.json

aaclayton (approved):
@Fyorl looks good to me, giving an approve. My comments are hopefully useful, but I trust your perspective on this. Overall looks good!

module/documents/item.mjs
Fyorl: Ideally, yes, but there's not a good place to put such preparation at the moment. `EquipmentData`, `WeaponData`, `ToolData`, and `SpellData` do not share a common ancestor (aside from the base `SystemDataModel`), so it would have to be duplicated for each of them.

template.json
Fyorl: Yes that should be fine to do for 2.3. I was holding off before due to some lingering issues with system data models in v10, but those were resolved with the 303 closing release.

Issue #1554, opened 2022-05-19
Add flags to advancement data


arbron:
To allow modules that want to expand the functionality of built-in advancement types, the advancement pseudo-documents should have a `flags` object that can be used to store arbitrary data like normal documents.

PR #2148, opened 2023-02-03
[#1554] Add flags to advancements

arbron:
Adds flags and the `getFlag`, `setFlag`, `unsetFlag` methods to `Advancement`.


Fyorl:
In principle this makes sense to do, but once we open the floodgates, there's no closing them. My concern here is that we're leaning too hard into 'Advancements are Documents' without them actually being Documents. It's convenient to use familiar APIs, but ultimately Advancements cannot be updated atomically, they propagate their updates up to the parent Item, and I don't want to be left with another mini-unlinked-token situation on our hands again.

I think I would have more confidence here if we had some concrete use-cases of what Advancement flags would be needed for. It would allow us to weigh the costs of attempting to accommodate those use-cases in some other way.

Fyorl (approved):

PR #2156, opened 2023-02-05
[#2024] More advanced drop-logic for spells on the spellbook tab

### LINKED ISSUE #2024: Set spells dropped onto NPCs with no spellcaster level to innate spellcasting
Right now all spells dropped onto NPCs start as prepared spells even if that NPC doesn't have any spellcaster levels. A lot of monsters, including most new official monsters, handle their spellcasting through "Innate Spellcasting" or some variant rather than giving the monster spellcaster levels. It would save a lot of time manually setting the casting mode on those spells if the system changed them to innate itself.

krbz999:
This changes the spellbook template, wrapping each section in a `DIV` tag for easier traversal. The look itself is unchanged.

The base actor sheet also saves the most recent `_dragEvent` when an item is dropped to be able to reference it in the needed functions.

Then finally, the base actor sheet has logic for PCs and NPCs to determine what the preparation mode of the dropped spell should be. The logic applies to both cantrips and leveled spells.

If a cantrip is dropped on...
- the pact section or any leveled section (1st to Xth), it goes under "Cantrips" (i.e. "prepared").
- if it's dropped on no section, it goes under "Cantrips" if the actor has any spell slot progression, otherwise "innate".

The cantrip otherwise falls under the section it is dropped onto.

If a leveled spell is dropped on the cantrip section or on no section...
- if the actor is an NPC, it falls under "innate" if the actor does not have caster levels, otherwise "prepared".
- if the actor is a PC, it falls under "innate" if the actor does not have spell slot progressions, otherwise "pact" or "prepared".

The leveled spell otherwise falls under the section it is dropped onto.

Attempts to resolve #2024.


module/applications/actor/base-sheet.mjs
```diff
@@ -1076,6 +1091,35 @@ export default class ActorSheet5e extends ActorSheet {
         return false;
       }
     }
+
+    // Adjust the preparation mode of a leveled spell depending on the section on which it is dropped.
+    if ( itemData.type === "spell" && itemData.system.level != 0 ) {
```
Varriount, line 1096: It might be worth extracting this into a dedicated method (named something like `onDropSpell`) and then use the same calling pattern as [line 1083](https://github.com/foundryvtt/dnd5e/blob/f3fe4c72285e62a3a9eb8cca2b54541a43d5aa52/module/applications/actor/base-sheet.mjs#L1083..L1084).

krbz999: I agree, it grew a bit big.

module/applications/actor/base-sheet.mjs
```diff
@@ -1159,6 +1134,38 @@ export default class ActorSheet5e extends ActorSheet {
...
 
+  /**
+   * Adjust the preparation mode of a dropped spell depending on the drop location on the sheet.
+   * @param {object} itemData    The item data requested for creation. **Will be mutated.**
+   * @param {DragEvent} event    The original drag event.
+   */
+  _onDropSpell(itemData, event) {
+    if ( itemData.type !== "spell" || itemData.system.level == 0 ) return;
+
+    // Determine the section it is dropped on, if any.
```
Varriount, line 1145: I like your philosophy regarding comments. 😄

module/applications/actor/base-sheet.mjs
```diff
@@ -1159,6 +1134,38 @@ export default class ActorSheet5e extends ActorSheet {
...
+    const mode = event.target.closest("div[data-type='spell']")?.dataset;
+
+    // If the spell is dropped on a section, set the mode to be that type.
+    if ( mode && mode.level > 0 ) itemData.system.preparation.mode = mode["preparation.mode"];
+
```
Varriount, line 1150:
```suggestion
    if ( mode && mode.level > 0 ) {
      itemData.system.preparation.mode = mode["preparation.mode"];
    }
```
Personally, I would add braces here, since the `if` statement has an `else if` statement, but ultimately that's up to @Fyorl and/or @arbron.

module/applications/actor/base-sheet.mjs
```diff
@@ -1159,6 +1134,38 @@ export default class ActorSheet5e extends ActorSheet {
...
+    }
+
+    // If the spell is dropped on an actor with only pact magic (and other classes having 'none'), set the spell to be pact.
+    else if ( this.document.type === "character" ) {
+      const modes = this.document.itemTypes.class.reduce((acc, cls) => {
+        if ( cls.spellcasting?.type === "pact" ) acc.pact = true;
+        else if ( cls.spellcasting?.type === "leveled" ) acc.leveled = true;
+        return acc;
+      }, {pact: false, leveled: false});
+      if ( modes.pact && !modes.leveled ) itemData.system.preparation.mode = "pact";
```
Varriount, line 1163: I might be misreading this, but wouldn't it theoretically be more efficient to rewrite this using a `for` loop?
```suggestion
      let pact = false;
      let leveled = false;
      for (const itemClass of this.document.itemTypes.class) {
        pact |= (itemClass.spellcasting?.type === "pact");
        leveled |= (itemClass.spellcasting?.type === "leveled");
        if (pact && leveled) break;
      }
    if (pact && !leveled) itemData.system.preparation.mode = "pact";
```
This could technically be made even more efficient with an if/else if/else statement, but I feel the above is a good balance between efficiency and readability for what is going to likely be a very small array.

Also, what is `itemData.system.preparation.mode` set to if `pact` is false or `leveled` is true (i.e., that last `if` condition is false)?

krbz999: Then no change is made, which is the idea.

I feel like the current `reduce` is much more readable. The only 'gain' from something more efficient would be for abserd actors with 20+ different classes.

module/applications/actor/base-sheet.mjs
```diff
@@ -39,6 +39,15 @@ export default class ActorSheet5e extends ActorSheet {
 
   /* -------------------------------------------- */
 
+  /**
+   * Track the most recent drag event.
+   * @type {DragEvent}
+   * @protected
+   */
+  _event = null;
```
arbron, line 47: I'm don't think storing the event like this is the best idea. I would suggest instead overriding the default `Actor#_onDropItem` method so that it passes the event through:

```js
  /** @override */
  async _onDropItem(event, data) {
    if ( !this.actor.isOwner ) return false;
    const item = await Item.implementation.fromDropData(data);
    const itemData = item.toObject();

    // Handle item sorting within the same Actor
    if ( this.actor.uuid === item.parent?.uuid ) return this._onSortItem(event, itemData);

    // Create the owned item
    return this._onDropItemCreate(event, itemData);
  }
```

I created an issue on core to hopefully get this change integrated there also: https://github.com/foundryvtt/foundryvtt/issues/8872

templates/actors/parts/actor-spellbook.hbs
```diff
@@ -30,91 +30,93 @@
...
-                <input type="text" name="system.spells.{{section.prop}}.value" value="{{section.uses}}" placeholder="0"
-                       data-dtype="Number">
-                <span class="sep"> / </span>
-                <span class="spell-max" data-level="{{section.prop}}" data-slots="{{section.slots}}">
-                    {{{section.slots}}}
-                    {{#if section.editable}}
-                    <a class="slot-max-override" data-tooltip="DND5E.SpellProgOverride">
-                        <i class="fas fa-edit"></i>
-                    </a>
+    <div {{#each section.dataset as |v k|}}data-{{k}}="{{v}}"{{/each}}>
```
arbron, line 33: This ends up with us repeating these dataset options here and with the `"+ Add"` button. It might be worth it to remove the repeated data below and add a bit of logic to `_onItemCreate` so it can get the properties from either location if necessary.

```js
  _onItemCreate(event) {
    event.preventDefault();
    const dataset = (event.currentTarget.closest(".spellbook-section") ?? event.currentTarget).dataset;
    const type = dataset.type;

    // skipped lines

    const itemData = {
      name: game.i18n.format("DND5E.ItemNew", {type: game.i18n.localize(`ITEM.Type${type.capitalize()}`)}),
      type: type,
      system: foundry.utils.expandObject({ ...dataset })
    };

    // more skipped lines
  }
```

Probably be worth it to add a class here to make it easier to identify the purpose of this div.
```suggestion
    <div class="spellbook-section" {{#each section.dataset as |v k|}}data-{{k}}="{{v}}"{{/each}}>
```

module/applications/actor/base-sheet.mjs
```diff
@@ -1115,6 +1135,39 @@ export default class ActorSheet5e extends ActorSheet {
...
+  _onDropSpell(itemData) {
+    if ( itemData.system.level == 0 || !["npc", "character"].includes(this.document.type) ) return;
+
+    // Determine the section it is dropped on, if any.
+    const mode = this._event.target.closest("div[data-type='spell']")?.dataset;
+
+    // If the spell is dropped on a section, set the mode to be that type.
+    if ( mode && mode.level > 0 ) {
+      itemData.system.preparation.mode = mode["preparation.mode"];
+    }
```
arbron, line 1151: One thought I have here is that if a NPC has no spellcasting levels, even cantrips dropped should be changed to innate. Generally sheets that include an `"Innate Spellcasting"` feature will list cantrips in that feature within the `at will` list, so it would make sense that they would be listed with the rest of the innate spells.

module/applications/actor/base-sheet.mjs
```diff
@@ -1115,6 +1135,39 @@ export default class ActorSheet5e extends ActorSheet {
...
+
+    // If the spell is not dropped on a section, and on an NPC with no caster levels, set the spell to be innate.
+    else if ( this.document.type === "npc" && !this.document.system.details.spellLevel ) {
+      itemData.system.preparation.mode = "innate";
+    }
+
+    // If the spell is dropped on an actor with only pact magic (and other classes having 'none'), set the spell to be pact.
```
arbron, line 1158: Check the line length there, a bit longer than 120 characters. The parenthetical seems redundant here.

```suggestion
    // If the spell is dropped on an actor with only pact magic set the spell to be pact.
```

module/applications/actor/base-sheet.mjs
arbron: Rather than iterating over `this.document.itemTypes.class`, which has to first filter over all items on the actor, you can instead use the `this.actor.classes` object which is pre-filtered.

If no spellcasting classes are found on the player, it perhaps might make sense to set to preparation mode to something like `atwill` rather than leave it as `leveled`. This would accommodate non-spellcasters who get certain spells from their race or from a feat.

arbron (changes requested):

module/applications/actor/base-sheet.mjs
krbz999: Having it integrated in core would be ideal. I did override these in an earlier commit, but then realised it broke several functions on several actor types, notably tossing an error when dropping folders of documents. Is there a downside to leaving it as is as a temporary solution? I can add a 'TODO' comment to have `event` removed later on.

krbz999:
I did a second run of it, will update the top message.

module/applications/actor/base-sheet.mjs
arbron: I'll defer this to Kim's review.

module/applications/actor/base-sheet.mjs
```diff
@@ -1115,6 +1135,52 @@ export default class ActorSheet5e extends ActorSheet {
...
+
+    // Determine the actor's spell slot progressions, if any.
+    const progs = Object.values(this.document.classes).reduce((acc, cls) => {
+      if ( cls.spellcasting?.type === "pact" ) acc.pact = true;
+      else if ( cls.spellcasting?.type === "leveled" ) acc.leveled = true;
+      return acc;
+    }, {pact: false, leveled: false});
+
+    // Case 1: Drop a cantrip.
+    if ( itemData.system.level == 0 ) {
```
arbron, line 1156:
```suggestion
    if ( itemData.system.level === 0 ) {
```

module/applications/actor/base-sheet.mjs
```diff
@@ -1115,6 +1135,52 @@ export default class ActorSheet5e extends ActorSheet {
...
+        if ( this.document.type === "npc" ) {
+          const npcLevel = this.document.system.details.spellLevel;
+          itemData.system.preparation.mode = npcLevel ? "prepared" : "innate";
+        } else {
+          itemData.system.preparation.mode = progs.pact ? "pact" : progs.leveled ? "prepared" : "innate";
+        }
+      } else {
+        itemData.system.preparation.mode = mode["preparation.mode"];
+      }
+    }
```
arbron, line 1179: Linter isn't a big fan of `else` blocks that only contain another `if` statement, so this can be refactored up a level:

```suggestion
    // Case 2: Drop a leveled spell in a section without a mode.
    else if ( (mode.level === 0) || !mode["preparation.mode"] ) {
      if ( this.document.type === "npc" ) {
        itemData.system.preparation.mode = this.document.system.details.spellLevel ? "prepared" : "innate";
      } else {
        itemData.system.preparation.mode = progs.pact ? "pact" : progs.leveled ? "prepared" : "innate";
      }
    }

    // Case 3: Drop a leveled spell in a specific section.
    else {
      itemData.system.preparation.mode = mode["preparation.mode"];
    }
```

arbron (changes requested):
Functionality seems good, just a few style changes from me. I'll throw it over to Kim :).

module/applications/actor/base-sheet.mjs
Fyorl: Agreed that having the store the event is not the ideal solution. I think foundryvtt/foundryvtt#8872 would be a breaking change (assuming we want to have the `event` be the first argument to match the rest of the API), so the earliest it can go in is v12. The implementation here with overriding `_onDrop` is only very minor tech debt though so it's probably ok as an interim solution with a TODO to change this later.

module/applications/actor/base-sheet.mjs
```diff
@@ -1233,7 +1299,7 @@ export default class ActorSheet5e extends ActorSheet {
     const itemData = {
       name: game.i18n.format("DND5E.ItemNew", {type: game.i18n.localize(`ITEM.Type${type.capitalize()}`)}),
       type: type,
-      system: foundry.utils.expandObject({ ...header.dataset })
+      system: foundry.utils.expandObject(dataset)
```
Fyorl, line 1302: Fairly sure this doesn't work because `dataset` is actually a `DOMStringMap` and not a plain object, so this needs to be:

```suggestion
      system: foundry.utils.expandObject({ ...dataset })
```

templates/actors/parts/actor-spellbook.hbs
```diff
@@ -30,91 +30,92 @@
...
-                <input type="text" name="system.spells.{{section.prop}}.value" value="{{section.uses}}" placeholder="0"
-                       data-dtype="Number">
-                <span class="sep"> / </span>
-                <span class="spell-max" data-level="{{section.prop}}" data-slots="{{section.slots}}">
-                    {{{section.slots}}}
-                    {{#if section.editable}}
-                    <a class="slot-max-override" data-tooltip="DND5E.SpellProgOverride">
-                        <i class="fas fa-edit"></i>
-                    </a>
+    <div class="spellbook-section" {{#each section.dataset as |v k|}}data-{{k}}="{{v}}"{{/each}}>
```
Fyorl, line 33: `<div>` is not actually a valid child of an `<ol>` element, but neither is another `<ol>`, which is something we also already do here. I think the 'strictly correct' markup we should use here (and in `actor-inventory.hbs`) is something like:
```html
<ol class="items-list inventory-list">
  <li class="spellbook-section">
    <header class="items-header spellbooks-header flexrow"><!-- ... --></header>
    <ol class="item-list"><!-- ... --></ol>
  </li>
  <!-- ... -->
</ol>
```

That's probably too disruptive of a change to make as part of this PR, but we could keep the markup mostly the same as before at the expense of some slightly more complicated logic in `_onDropSpell`:

```hbs
<!-- actor-spellbook.hbs -->
<li class="items-header spellbook-header flexrow" {{#each section.dataset as |v k|}}data-{{k}}="{{v}}"{{/each}}>
```

```js
// _onDropSpell
let header = this._event.currentTarget.closest(".items-header"); // Dropped directly on the header.
if ( !header ) {
  const list = this._event.currentTarget.closest(".items-list"); // Dropped inside an existing list.
  header = list?.previousElementSibling;
}
const mode = header?.dataset ?? {};
```

Let me know what you think.

module/applications/actor/base-sheet.mjs
```diff
@@ -1115,6 +1135,52 @@ export default class ActorSheet5e extends ActorSheet {
...
+        itemData.system.preparation.mode = mode["preparation.mode"];
+      }
+    }
+
+    // Case 2: Drop a leveled spell in a section without a mode.
+    else if ( (mode.level === 0) || !mode["preparation.mode"] ) {
+      if ( this.document.type === "npc" ) {
+        itemData.system.preparation.mode = this.document.system.details.spellLevel ? "prepared" : "innate";
+      } else {
+        itemData.system.preparation.mode = progs.pact ? "pact" : progs.leveled ? "prepared" : "innate";
```
Fyorl, line 1172: If the character has both Pact Magic and Spellcasting, I understand we need to choose one to take precedence, but I might suggest the other way around here with `prepared` taking precedence over `pact` as I think that matches more closely the original behaviour.

Fyorl (changes requested):
The feature makes sense to me, and the code looks good, I just have an alternative proposal for how we do the markup.

templates/actors/parts/actor-spellbook.hbs
krbz999: > Let me know what you think.

My (very blunt) opinion is that don't see this as *too* important, right now at least, and that the actor sheets are due for a revisit overall. I think it could be good to see to this part in particular at such a time.

Fyorl (approved):
Nice UX enhancement, thank you Zhell

PR #2158, opened 2023-02-06
[#1403] Ability Score Improvement Advancement

### LINKED ISSUE #1403: Advancement Type: Ability Score Improvement
This advancement can be used to represent an ability score change performed by `race` or `background` (for support of D&D One content) and can be used to indicate at which levels a `class` grants an ability score improvement with optional feat support.

## Level Up Interface

For classes, first the player would be asked whether they wish to improve their ability scores or add a feat. Since feats are an optional rule, the feat option can potentially be disabled using a system setting. For races and backgrounds, they will be taken directly to the ASI interface.

If they choose to take the ASI, they will be shown a list of their current ability scores and given the option to distribute up to two points among them.

If they choose to take a feat, they will be presented with a drop area where they can place the new feat.

## Configuration

The `points` property is a number indicating how many points a player can allocate among their ability scores.

The `fixed` property is an object containing how many points are allocated to specific ability scores.

One or both of these properties must be set. The system will also support granting fixed to one score and still allocating points to another (aka you get a +2 to dexterity and one other point to allocate as you wish).

``` json
{
  "points": 2
  "fixed": null
}
```

```json
{
  "points": null,
  "fixed": {
    "dex": 2,
    "con": 1
  }
}
```

## Value

The value object contains a `selected` property indicated whether the player selected the ability score improvement (`"asi"`) or the feat (`"feat"`).

If they chose the ASI then there is an `improvements` object containing the keys of the ability scores they improved and a value with the amount they improved that score. This will include any changes made by fixed values because they might have been capped by the maximum for that score.

```javascript
{
  selected: "asi",
  improvements: {
    cha: 2
  }
```

If a feat is chosen, the object contains a `feat` property containing another object with the origin UUID of the feat and the ID on the actor. This structure intentionally mirrors the structure proposed for item grants (#1400) and choices (#1401) to hopefully be able to reuse some code.

```javascript
{
  selected: "feat",
  feat: {
    "hpLNiGq7y67d2EHA": "Compendium.dnd5e.classfeatures.Dop1RcU5UdhSVLYS"
  }
}
```

## Future Directions

A more advanced items type and compendium scanning system would allow for a list of feats to be presented to the player built automatically from the various compendiums available rather than require the user to drag an item in manually.

arbron:
Implements the ability score improvement advancement to allow for selecting ability scores and feats.

Rogue - Advancements

### Configuration

The ability score improvement can be configured to allow for a number of variable points that can be assigned as the user chooses (like with classes):

Configure Points Improvement

Or with a fixed improvement to certain traits:

Configure Fixed Improvement

Or a combination of the two.

### Application

During the level up process the player is presented with a list of their ability scores with the ability to modify them:

Apply Points Improvement

When added to classes, the advancement also presents the option of selecting a feat:

https://user-images.githubusercontent.com/19979839/217079416-6d802657-995a-4bb1-8733-717464b6fb4d.mov

If a fixed improvement is configured, then that is indicated on the screen and those scores cannot be modified by the player:

Apply Fixed Improvement

Resolves #1403 

### To Do
- [ ] Find a way to set default points to `2` for classes, and `0` for other items
- [x] Consider whether adding feats should be a system option, since it is technically an optional rule
- [ ] Add advancement to SRD classes
- [x] Display ASIs on Class Summary journal pages


Zanderaf:
Any plans to add function with feats increasing ability scores when dragged in?  There arent any I know of off the top of my head in the SRD but stuff like Keen Mind has Int or Wis

module/applications/advancement/ability-score-improvement-config.mjs
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  getData() {
+    const abilities = Object.entries(CONFIG.DND5E.abilities).reduce((obj, [key, data]) => {
+      const fixed = this.advancement.configuration.fixed[key] ?? 0;
+      obj[key] = {
+        key: key,
```
Fyorl, line 22:
```suggestion
        key,
```

module/data/advancement/ability-score-improvement.mjs
```diff
@@ -0,0 +1,50 @@
...
+  /** @inheritdoc */
+  static defineSchema() {
+    return {
+      type: new foundry.data.fields.StringField({
+        required: true, initial: "asi", choices: ["asi", "feat"]
+      }),
+      assignments: new MappingField(new foundry.data.fields.NumberField({
+        nullable: false, integer: true
+      }), {required: false, initial: undefined}),
+      feat: new MappingField(new foundry.data.fields.StringField(), {
```
Fyorl, line 45: The fact that this is a `MappingField` is awkward, but I understand that it fits better into the rest of our framework around adding items via advancements.

module/documents/advancement/ability-score-improvement.mjs
```diff
@@ -0,0 +1,168 @@
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Information on the ASI points available.
+   * @type {{ assigned: number, total: number }}
+   */
+  get points() {
+    return {
+      assigned: Object.values(this.value.improvements ?? {}).reduce((n, c) => n + c, 0),
```
Fyorl, line 53: Should this be `assignments`?
```suggestion
      assigned: Object.values(this.value.assignments ?? {}).reduce((n, c) => n + c, 0),
```

Fyorl (changes requested):
Really great stuff, Jeff. The code looks good, only some minor feedback, and in particular the UI looks very crisp and polished.

Fyorl (approved):

PR #2159, opened 2023-02-06
[#1013] Add max to ability scores

### LINKED ISSUE #1013: Add "max" field to Character abilities

Character abilities (Str, Dex, etc) are capped at 20 by default. Some abilities (Barbarian class features and items) can increase this beyond this cap. It would be good to have this max tracked in the character data somewhere.

My example use case for this is to be able to restrict the bonus applied by an Active Effect by using a formula such as in a DAE custom effect. e.g. `{@data.abilities.dex.max, @data.abilities.dex.value + 4} kl`.

I wouldn't see the system actively preventing the ability been set higher, just providing a method of determining what the max can be for modules like DAE and ddb-importer.

If you're happy with this I would be happy to try to implement the feature.

arbron:
Split the maximum ability score from #2158 for separate review.

Adds `max` to abilities that defaults to `20` to allow `AbilityScoreImprovementAdvancement` and other uses to cap an ability score.

- Currently doesn't prevent entering a value over the max on the character sheet
- This ends up also adding `max` to NPCs and vehicles where it isn't useful, should the data models be modified so it is only included on characters?

Resolves #1013

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -94,6 +94,7 @@
 "DND5E.AbilitySan": "Sanity",
 "DND5E.AbilitySanAbbr": "san",
 "DND5E.AbilityScore": "Ability Score",
+"DND5E.AbilityScoreMax": "Maximum Ability Score",
 "DND5E.AbilityModifier": "Ability Modifier",
 "DND5E.AbilityPromptText": "What type of {ability} check?",
 "DND5E.AbilityPromptTitle": "{ability} Ability Check",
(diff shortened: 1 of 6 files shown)


Fyorl:
> This ends up also adding max to NPCs and vehicles where it isn't useful, should the data models be modified so it is only included on characters?

I think it's fine if every ability score object has a `max`, other actor types are free to ignore it.

Do we want to make the maximum ability score a `CONFIG` value? So the default `max` would be `null`, which would draw from the globally configured maximum. And then you can override it with a specific value.

Fyorl (approved):

Issue #2170, opened 2023-02-08
Add ammo, spellcasting focus, and container types to config


arbron:
Adding these in addition to the existing weapon, tool, and armor types will be helpful when implementing the new `EquipmentAdvancement`.

Blocks #1871

PR #2172, opened 2023-02-08
[#2170] Add ammo, focus, and container types

arbron:


ccjmk:
It exceeds the scope of this ticket, but adding the focus type to a class' `spellcasting` will be handy to link these two together (for example, for being able to search and select an appropriate focus for a spellcating class`

Fyorl (approved):

Issue #885, opened 2020-12-14
Feature request: Add a way to apply a flat bonus to attack rolls


aaclayton:
Originally in GitLab by @cs96and

Items such as the circlet of blasting have a flat +5 bonus to attack rolls, regardless of the wearer's ability scores.  It would be good if there was a separate "Flat" or "None" option in the ability score drop down so that `@mod` is not automatically added to the roll.

PR #2173, opened 2023-02-09
[#885] [#2026] Add 'none' as an option when choosing the ability modifier

### LINKED ISSUE #2026: Add 'none' as an option in the ability dropdown for weapon and spell attacks to indicate that it has a flat bonus and should not add any ability modifier

krbz999:
This introduces an additional option when selecting an item's ability modifier; ~~"flat"~~ "none", which will ignore ability modifiers, ~~proficiency bonus, and any bonuses to melee/ranged weapon/spell attacks~~.

circ_select

This also adjusts the logic such that an 'equipment' type item is required to have proficient toggled on for proficiency bonus to be added, and removes `@mod` from created spell scrolls, which is not needed any longer (these are set to ~~`flat`~~ `none` if they have an attack).

Example using Circlet of Blasting:

Before (using an actor with a +2 PB and +2 INT):

circ_before

After (using that same actor, with Proficient toggled off):

circ_after

Resolves #2026.
Does something to #885.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -599,7 +599,7 @@ export default class Item5e extends Item {
     // Take no further action for un-owned items
     if ( !this.isOwned ) return {rollData, parts};
 
-    if( this.system.ability !== "flat" ) {
+    if ( this.system.ability !== "flat" ) {
       // Ability score modifier
       parts.push("@mod");
 
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -2038,7 +2038,8 @@ export default class Item5e extends Item {
     // Get spell data
     const itemData = (spell instanceof Item5e) ? spell.toObject() : spell;
     let {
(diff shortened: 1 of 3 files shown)


module/documents/item.mjs
```diff
@@ -587,18 +587,18 @@ export default class Item5e extends Item {
...
+    if ( !["weapon", "consumable", "equipment"].includes(this.type) || this.system.proficient ) {
       parts.push("@prof");
       if ( this.system.prof?.hasProficiency ) rollData.prof = this.system.prof.term;
     }
 
     // Actor-level global bonus to attack rolls
     const actorBonus = this.actor.system.bonuses?.[this.system.actionType] || {};
     if ( actorBonus.attack ) parts.push(actorBonus.attack);
-
+
```
arbron, line 601:
```suggestion

```

arbron (commented):
Just one minor style issue, but otherwise looks good to me.

arbron (approved):

Fyorl (approved):

PR #2182, opened 2023-02-11
Tweaks and fixes to the scroll creation process

### LINKED ISSUE #1241: Add a note declaring a spell as "requiring concentration" when creating a spell scroll item from a spell item

Whether or not a spell has a concentration requirement is crucial information which can't be expressed in the spell scroll item itself. A one-line auto-generated text note in the description would suffice instead.

Repro: Drag the "Alter Self" spell from the "Spells (SRD)" compendium to an actor's "Inventory" tab.

### LINKED ISSUE #2181: Spell to scroll conversion is missing original text
Normally, when a spell is dragged in the ineventory tab, it will turn into a spell scroll.
However, as of the latest 2.1.4 version (or even earlier) only the description of srcolls in general is added instead of having original text + scroll description text.

image

krbz999:
Added missing localization in the scroll description.

Added a hook (`dnd5e.createScrollFromSpell`) for those who want to tweak the scroll creation process further (the process cannot be cancelled).

Add concentration tag as a piece of descriptive text under the spell name (closes #1241), only for concentration spells.

image

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -2039,7 +2039,7 @@ export default class Item5e extends Item {
     const itemData = (spell instanceof Item5e) ? spell.toObject() : spell;
     let {
       actionType, description, source, activation, duration, target,
-      range, damage, formula, save, level, attackBonus, ability
+      range, damage, formula, save, level, attackBonus, ability, components
     } = itemData.system;
 
     // Get scroll data
@@ -2056,7 +2056,11 @@ export default class Item5e extends Item {
     const scrollDetails = scrollDescription.slice(scrollIntroEnd + pdel.length);
 
     // Create a composite description from the scroll description and the spell details
-    const desc = `${scrollIntro}<hr/><h3>${itemData.name} (Level ${level})</h3><hr/>${description.value}<hr/><h3>Scroll Details</h3><hr/>${scrollDetails}`;
+    const desc = scrollIntro
+    + `<hr/><h3>${itemData.name} (${game.i18n.format("DND5E.LevelNumber", {level})})</h3>`
+    + (components.concentration ? `<em>${game.i18n.localize("DND5E.ScrollRequiresConcentration")}</em>` : "")
+    + `<hr/>${description.value}<hr/>`
+    + `<h3>${game.i18n.localize("DND5E.ScrollDetails")}</h3><hr/>${scrollDetails}`;
 
     // Used a fixed attack modifier and saving throw according to the level of spell scroll.
     if ( ["mwak", "rwak", "msak", "rsak"].includes(actionType) ) {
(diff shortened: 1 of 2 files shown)


module/documents/item.mjs
```diff
@@ -2323,10 +2328,20 @@ export default class Item5e extends Item {
...
         save, level, attackBonus
       }
     });
+
+    /**
+     * A hook event that fires after the item data for a scroll is created but before the item is returned.
+     * @function dnd5e.createScrollFromSpell
+     * @memberof hookEvents
+     * @param {Item5e|object} spell       The spell or item data triggering this function.
+     * @param {object} itemData           The item data of the spell triggering this function.
```
arbron, line 2341: I'm not sure there is much reason to pass both the spell as an document and an object. We aren't doing any actual modifications to `itemData` so consumers of this hook could just get it themselves from the first property.

module/documents/item.mjs
```diff
@@ -2323,10 +2328,20 @@ export default class Item5e extends Item {
...
+        description: {value: desc.trim()}, source, actionType, activation, duration, target, range, damage, formula,
         save, level, attackBonus
       }
     });
+
+    /**
+     * A hook event that fires after the item data for a scroll is created but before the item is returned.
+     * @function dnd5e.createScrollFromSpell
+     * @memberof hookEvents
+     * @param {Item5e|object} spell       The spell or item data triggering this function.
```
arbron, line 2340: Not related to your changes, but it would be a good idea to update the JSDoc at the start of this method to match this. We've always accepted either a Document or object, but the JSDoc param doesn't reflect that.

module/documents/item.mjs
```diff
@@ -2282,15 +2282,17 @@ export default class Item5e extends Item {
...
   /**
    * Create a consumable spell scroll Item from a spell Item.
-   * @param {Item5e} spell      The spell to be made into a scroll
-   * @returns {Item5e}          The created scroll consumable item
+   * @param {Item5e|object} spell     The spell or item data to be made into a scroll
+   * @param {object} options          Additional options that modify the created scroll
+   * @returns {Item5e}                The created scroll consumable item
    */
-  static async createScrollFromSpell(spell) {
+  static async createScrollFromSpell(spell, options={}) {
```
krbz999, line 2289: The `options` object is new and added for purely selfish reasons, and for people who call this method manually and prefer not to rely on a hook. The system itself does not make use of it.

arbron:
@krbz999 Could you split the bug fix off into a separate PR from the other changes? The rest of the changes are nice, but I don't want them to hold up the bug fix from getting into `2.2`.

krbz999:
> @krbz999 Could you split the bug fix off into a separate PR from the other changes? The rest of the changes are nice, but I don't want them to hold up the bug fix from getting into `2.2`.

Wasn't that added to 2.1.5?

arbron:
> Wasn't that added to 2.1.5?

So it was, ancient history :)

Fyorl (approved):

Issue #2185, opened 2023-02-11
Long class names with multi-classing causes sheet layout issues


arbron:
<img width="554" alt="Screenshot 2023-02-11 at 14 56 52" src="https://user-images.githubusercontent.com/19979839/218284601-d2b17279-de86-4119-b2f6-f37a181fd8b5.png">

<img width="733" alt="Screenshot 2023-02-11 at 14 57 20" src="https://user-images.githubusercontent.com/19979839/218284606-c9a39db6-a6a9-4b75-b68c-668203d20be8.png">

Issue #2162, opened 2023-02-06
Longer Character Background Titles Cause Display Issues


Dialectikal:
Foundry: 10.291
DnD5e: 2.1.4
No modules active

Slightly longer names for character backgrounds bleed down into the character sheet and cause odd display issues instead of being cut off.

![image](https://user-images.githubusercontent.com/31572084/217110547-ff98135b-dd22-4451-818d-159c4be03951.png)

For clarification, the Background should be "City Watch/Investigator"

PR #2186, opened 2023-02-11
[#2162, #2185] Fix sheet layout issues with long background or class names

arbron:
Screenshot 2023-02-11 at 14 59 30


krbz999:
I'm normally against adding new css classes where they aren't needed, but could there be a `.race`, `.background`, and `.alignment` class added, to align with their `.proficiency` sibling?

Fyorl (approved):
I'll note that backgrounds with pathologically-long names still break the layout, but I couldn't find a quick fix aside from `overflow: hidden` which was not ideal.

PR #2207, opened 2023-02-21
[#2132, #2181, #2205] Fix few nagging bugs

### LINKED ISSUE #2132: Group inventory items can be expanded repeatedly. Expanded items do not stay expanded when the sheet is re-rendered.

### LINKED ISSUE #2181: Spell to scroll conversion is missing original text

### LINKED ISSUE #2205: transformInto() not respecting renderSheet:false
When using transformInto there doesn't seem to be any way of preventing the sheet from rendering – {renderSheet:false} doesn't have any effect, it still pops right on up.

arbron:
A few bugs that are annoying but easy to fix that should probably be solved sooner rather than waiting for `2.2.0`.

- Fix issue with items in group inventory repeatedly expanding
- Fix issue with spell scrolls not getting spell description
- Fix issue with `transformInto` ignoring `renderSheet` option for linked actors

Resolves #2132 
Resolves #2181 
Resolves #2205

### DIFF
--- a/module/applications/actor/group-sheet.mjs
+++ b/module/applications/actor/group-sheet.mjs
@@ -9,6 +9,15 @@ import Item5e from "../../documents/item.mjs";
  */
 export default class GroupActorSheet extends ActorSheet {
 
+  /**
+   * IDs for items on the sheet that have been expanded.
+   * @type {Set<string>}
+   * @protected
+   */
+  _expanded = new Set();
+
+  /* -------------------------------------------- */
+
   /** @inheritDoc */
   static get defaultOptions() {
     return foundry.utils.mergeObject(super.defaultOptions, {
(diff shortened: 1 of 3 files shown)


module/applications/actor/group-sheet.mjs
```diff
@@ -50,6 +59,11 @@ export default class GroupActorSheet extends ActorSheet {
     // Inventory
     context.itemContext = {};
     context.inventory = this.#prepareInventory(context);
+    context.expandedData = {};
+    for ( const id of this._expanded ) {
+      const item = this.actor.items.get(id);
+      if ( item ) context.expandedData[id] = await item.getChatData({secrets: this.actor.isOwner});
```
Fyorl, line 65: Looks like this might conflict with changes in #2206

Fyorl (approved):
LGTM. I'll let you handle the merging so you can navigate the potential conflicts

module/applications/actor/group-sheet.mjs
arbron: That was on the base sheet, this is on the group sheet.

Issue #2223, opened 2023-02-25
BUG  Negative temp max that exceed total hp


MotoMoto1234:
If you set an active effect to system.attributes.hp.tempmax and a negative value past the total hp, the bar does not clamp to the left side of the token and goes on endlessly depending on how crazy you go:

UPDATE:  This is also possible to replicate without an active effect at all, just put a crazy negative number in the tempmax field of an actor

![image](https://user-images.githubusercontent.com/116308065/221382130-b83266be-f659-4830-bba9-e83e04e901e4.png)

![image](https://user-images.githubusercontent.com/116308065/221382136-f9a25735-14d3-4f22-8eb6-d2b74f74d346.png)

PR #2238, opened 2023-03-11
HP bar underflow fix

Ikaguia:
Fixes https://github.com/foundryvtt/dnd5e/issues/2223


Fyorl:
Apologies, it appears I missed this PR. We have an alternative fix implemented now though.

Issue #2215, opened 2023-02-22
Save a reference to the level of an item when it creates a MeasuredTemplate


krbz999:
In 2.0.0, `flags.dnd5e.origin` (the item's uuid) was added to templates created from an item meaning it became possible to know what item created it long after creation. However the chat message is the only source if one wants to refer to the level of the spell.

Similar to other cases, such as when making an attack or damage roll, it would be nice if the spell level was also saved for future reference on the template in `flags.dnd5e.spellLevel`.

PR #2239, opened 2023-03-11
Add spellLevel to measured template flags

Ikaguia:
Closes https://github.com/foundryvtt/dnd5e/issues/2215

### DIFF
--- a/module/canvas/ability-template.mjs
+++ b/module/canvas/ability-template.mjs
@@ -48,6 +48,7 @@ export default class AbilityTemplate extends MeasuredTemplate {
       fillColor: game.user.color,
       flags: { dnd5e: { origin: item.uuid } }
     };
+    if ( item.type === "spell" ) foundry.utils.mergeObject(templateData.flags, {"dnd5e.spellLevel": item.system.level});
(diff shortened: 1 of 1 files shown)


module/canvas/ability-template.mjs
```diff
@@ -48,6 +48,7 @@ export default class AbilityTemplate extends MeasuredTemplate {
       fillColor: game.user.color,
       flags: { dnd5e: { origin: item.uuid } }
     };
+    if ( item.type === "spell" ) foundry.utils.mergeObject(templateData.flags, {"dnd5e.spellLevel": item.system.level});
```
krbz999, line 51: `mergeObject` might be a bit overkill here, since `flags.dnd5e` always exists. 🤔

Ikaguia: I just followed the same pattern already established here:
https://github.com/foundryvtt/dnd5e/blob/bdc76e271d1ebae7092824c4f0e7c5299d7172e4/module/documents/item.mjs#L842

Fyorl (approved):

Issue #2204, opened 2023-02-19
Request: Add back the ability to equip "Loot" inventory items.


GambetTV:
You can equip literally everything else. There's no reason to take away features that already existed like this. It just kills modules that utilize it, and basically limits what we can do with our games. Maybe I want to know if a player is holding his tinderbox. There are so many items that don't have a mechanical reason to be classified as Equipment or Weapons that we still might have good reasons for wanting to equip in game. The module Variant Encumbrance, among others, utilized equipping items to reduce carry weight. Removing features like this hinders my ability to homebrew.

Aize01:
Yep, very essential for my campaign as well, since we play heavily on inventory management (its a scavenging apocalyptic campaign), please bring this back.

Fyorl:
Seemingly a deliberate decision per #1992 but we've since moved it back to consumables and containers. I do not recall the original rationale for this so I will tag it as a feature request to consider.

PR #2240, opened 2023-03-11
Make 'loot' equippable

Ikaguia:
Closes https://github.com/foundryvtt/dnd5e/issues/2204


krbz999:
I think it was actually a conscious decision to remove the ability to equip/attune to loot-type items.
In either case, there's also the issue of equipping/attuning through the item sheet, not just the actor sheet. That's not available for loot-type items.

Issue #2168, opened 2023-02-08
Standardize item system type format


arbron:
Currently an item type is tracked in several ways depending on the type of item:
- **consumable**: `system.consumableType`
- **equipment**: `system.armor.type` (even if it isn't armor), `system.baseItem`
- **feat**: `system.type.value`, `system.type.subtype`
- **tool**: `system.toolType`, `system.baseItem`
- **weapon**: `system.weaponType`, `system.baseItem`

I propose migrating these values to a more standard format to match features:
```js
{
  system: {
    type: {
      value: "",
      subtype: "", // For features
      baseItem: "" // For equipment, tools, and weapons
    }
  }
}
```

This would make it easier to create indexes on these fields rather than needing to track a different indexed field for each item type.

Ikaguia:
I believe I caught all the existing uses you mentioned, but was unsure on how to do migration; on my PR it is done just via dataModel migration

Ikaguia:
@arbron I would suggest altering the name of the property from `system.type` and `system.type.subtype` to `system.category` and `system.type.subcategory`, since that would make it less likely to be confused with the item's `type` property.

Ikaguia:
@arbron sorry to ping you a second time, but what is your opinion on the suggestion above?

Ikaguia:
About the compendium migration, should that be done as a separate issue or... ?

Fyorl:
> About the compendium migration, should that be done as a separate issue or... ?

Best to wait for the changes planned in #2665 first.

PR #2242, opened 2023-03-11
Standardize item system type format

Ikaguia:
https://github.com/foundryvtt/dnd5e/issues/2168


Fyorl:
Thanks for doing this work. Do you think it would be possible to omit the compendium data changes at the moment as they add a lot of noise to the PR? It looks like it could do with a rebase as well. Please let me know if you're able to do that, otherwise I can probably handle it.

Issue #2247, opened 2023-03-12
Display currencies for NPCs


ArekExora:
Currently NPCs have the currencies data in their model, and several modules patch it to the sheet. Why not adding it in the system itself?
It may make the GM life easier by keeping an actor by actor track of the money the player will be able to loot.

It could be switched on or off with a config flag too, in case some people dislike it

![image](https://user-images.githubusercontent.com/20903647/224560765-74e7f29f-a4ed-4841-a543-79d65b0662bb.png)

PR #2248, opened 2023-03-12
Display currencies in NPC sheets too

ArekExora:
Closes https://github.com/foundryvtt/dnd5e/issues/2247

Display the currencies of the NPCs in the sheet instead of having to manually check the model.

image

The minimum width of the sheet had to be increased a bit to avoid tinkering with the currencies styles.


arbron:
Closing this due to the upcoming NPC sheet re-design.

PR #2263, opened 2023-04-15
RAW See Invisibility detection mode

dev7355608:
Core's See Invisibility detection mode is not RAW (or RAI). See Invisibility does not allow you to see an invisible creature or object that you wouldn't be able to see if the creature or object wasn't invisible. But core's See Invisibility allows that: for example, a token with 0 vision range and no other detection modes other than See Invisibility can see an invisible token in complete darkness provided vision is not obstructed by walls. See Invisibility cannot detect anything on its own. It just suppresses (the first bullet point of) the invisible condition and requires another sense to see the creature or object.

The proposed detection mode test whether this vision source sees the target without the invisible status effect.
1. All vision modes except this vision source are temporarily removed. If there's more than one active vision source and one of them has See Invisibility and the other does not, the vision source with See Invisibility must not make the other vision source see invisible tokens. That's why we need to remove all but this vision source.
2. Non-sight-based detection modes are temporarily removed. Non-sight-based detection modes ignore the invisible condition anyway. If we didn't remove the non-sight-based detection modes, then the invisible target might end up with the wrong detection filter: the See Invisibility detection filter, when in reality the invisible token was detected by Feel Tremor, which of course doesn't need See Invisibility to detect it.
3. The invisible status is removed temporarily, such that `hasStatusEffect(CONFIG.specialStatusEffects.INVISIBLE)` returns false.
4. Visibility is tested.
5. The original state is restored.

### DIFF
--- a/module/canvas/_module.mjs
+++ b/module/canvas/_module.mjs
@@ -1,3 +1,4 @@
 export {default as AbilityTemplate} from "./ability-template.mjs";
+export * from "./detection-mode.mjs";
 export {default as Token5e} from "./token.mjs";
 export {measureDistances} from "./grid.mjs";
--- a/module/canvas/detection-mode.mjs
+++ b/module/canvas/detection-mode.mjs
@@ -1,0 +1,158 @@
+/**
+ * The detection mode for See Invisibility.
+ */
+export class DetectionModeSeeInvisibility extends DetectionMode {
+  constructor() {
+    super({
+      id: "seeInvisibility",
+      label: "DETECTION.SeeInvisibility",
+      type: DetectionMode.DETECTION_TYPES.SIGHT,
+      walls: false
+    });
+  }
+
+  /** @override */
+  static getDetectionFilter() {
+    return this._detectionFilter ??= GlowOverlayFilter.create({
+      glowColor: [0, 0.60, 0.33, 1]
+    });
+  }
+
+  /** @override */
+  _canDetect(visionSource, target) {
+    // Only invisible tokens can be detected.
+    return target instanceof Token && target.document.hasStatusEffect(CONFIG.specialStatusEffects.INVISIBLE);
+  }
+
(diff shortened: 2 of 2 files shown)


module/canvas/detection-modes/see-invisibility.mjs
```diff
@@ -0,0 +1,158 @@
...
+    const visionSources = this.#removeOtherVisionSources(visionSource);
+    const detectionsModes = this.#removeNonSightDetectionModes(visionSource);
+    const activeEffects = this.#removeInvisibleStatusEffects(target);
+
+    // Test whether this vision source sees the target without the invisible status effect.
+    const isVisible = canvas.effects.visibility.testVisibility(test.point, { tolerance: 0, object: target });
+
+    this.#restoreOtherVisionSources(visionSources);
+    this.#restoreNonSightDetectionModes(visionSource, detectionsModes);
+    this.#restoreInvisibleStatusEffects(target, activeEffects);
```
Varriount, line 42: Rather than removing and restoring sources/detection modes/effects, would it be possible to instead extend or modify `testVisibility` so that these things can be ignored in its logic?

dev7355608: Vision source and detection mode filtering could be done in `testVisibility` without removing and restoring, but suppressing the status effect cannot. So it wouldn't really be a big improvement. Also there's the issue that having the system override `testVisibility` could cause module compatibility issues with those that override it too.

Varriount: Why does the status effect need to be suppressed?

dev7355608: We need the detection modes to be able to detect invisible tokens when they normally cannot. If we don't suppress the invisible status effect, See Invisibility doesn't do anything.

PR #2268, opened 2023-04-16
Add Detect Poison and Disease detection mode

dev7355608:


module/canvas/detection-modes/detect-poison-and-disease.mjs
```diff
@@ -0,0 +1,53 @@
...
+      label: "DND5E.DetectPoisonAndDisease",
+      type: DetectionMode.DETECTION_TYPES.OTHER,
+      walls: false
+    });
+  }
+
+  /** @override */
+  static getDetectionFilter() {
+    return this._detectionFilter ??= GlowOverlayFilter.create({
+      glowColor: [0, 1, 0, 1]
```
Varriount, line 17: It might be a good idea to allow the glow color (and other parameters?) to be modified by modules and/or settings (e.g. for those with color blindness).

PR #2269, opened 2023-04-16
Add Detect Evil and Good detection mode

dev7355608:


module/canvas/detection-modes/detect-evil-and-good.mjs
```diff
@@ -0,0 +1,36 @@
...
+    if (!(target instanceof Token)) return false;
+    const actor = target.actor;
+    if (!(actor && actor.type === "npc")) return false;
+    const type = actor.system.details.type.value;
+    return type === "aberration"
+      || type === "celestial"
+      || type === "elemental"
+      || type === "fey"
+      || type === "fiend"
+      || type === "undead";
```
krbz999, line 32: Maybe a `CONFIG.DND5E` object with the 'good' types and 'evil' types would be nice to have!

Fyorl: Yes, we should use a `CONFIG` to store these values rather than hard-coding it in the detection mode.

Fyorl (changes requested):

PR #2265, opened 2023-04-16
Fix mundane item rarity

dev7355608:
Summary of changes:
- Changed the rarity of Dust of Dryness from `""` to `"uncommon"`.
- Changed the rarity of all mundane items in Items to `""`.
- Changed the rarity of all items in Starter Heroes to `""`: the starter heroes carry only mundane items.
- Changed the rarity of all item in Trade Goods to `""`: trade goods are mundane items.
- Changed the rarity of all items in Monsters to `""`: items carried by monsters are not from the Magic Items Tables and therefore don't have a rarity.

PR #2271, opened 2023-04-17
Add Detect Magic detection mode

dev7355608:
This makes the assumption that mundane items have no rarity and only magic items have a rarity. Unfortunately the rarities of mundane items in the items pack are set to common (#2265): so all mundane items that originated from the pack are detected as magical.

### DIFF
--- a/module/canvas/detection-modes/detect-magic.mjs
+++ b/module/canvas/detection-modes/detect-magic.mjs
@@ -1,0 +1,106 @@
+/**
+ * The detection mode for Detect Magic.
+ */
+export class DetectionModeDetectMagic extends DetectionMode {
+  constructor() {
+    super({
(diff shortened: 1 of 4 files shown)


module/canvas/detection-modes/detect-magic.mjs
```diff
@@ -0,0 +1,106 @@
+/**
+ * The detection mode for Detect Magic.
+ */
+export class DetectionModeDetectMagic extends DetectionMode {
```
krbz999, line 4: 'Detect Magic' leaves a faint aura on a magical object/creature if one uses their action, which requires that the target is visible, but the Detection Mode is the other way around, only creating an effect for invisible targets. Could this be inverted or rather include both if possible?

PR #2278, opened 2023-05-08
Fixes for v11.

### LINKED ISSUE #2275: [v11]: SRD Rules Compendium triggers deprecation warning and ain't lookin' so hot
```diff
- rules.apps = [new applications.journal.SRDCompendium(rules)];
+ rules.apps = [new applications.journal.SRDCompendium({collection: rules})];
```
image

### LINKED ISSUE #2276: [v11] Remove restriction on editing owned items' effects.
`item-sheet.mjs`:
```diff
- if ( this.item.isOwned ) return ui.notifications.warn("Managing Active Effects within an Owned Item is not currently supported and will be added in a subsequent update.");
```

Related issue: #1921

krbz999:
* ~~Changed deprecated i18n strings~~ Added new strings that are compliant with v11.
* Changed deprecated method of instantiating the compendium app, applicable only if on v11.
* ~~Converted some hbs syntax to use `ifThen`.~~ (reverted this change for v10 compatibility.)
* ~~Changed all `ActiveEffect#label` instances into `ActiveEffect#name`.~~ Muted the `ActiveEffect#label` deprecation warning.
* Removed blocker on modifying owned items' effects (closes #2276), however only on v11.
* Tweaked the css on the SRD rules compendium (closes #2275).
* Added missing `new` in `combat.mjs`. (This one is not relevant to v11, it was just a bug.)
* Removed any calls to deprecated `ActiveEffect#_getSourceName` on v11.
* Removed a deprecation warning in `roll-dialog.hbs` due to the change to `CONFIG.DND5E.abilities`.


dnd5e.css
```diff
@@ -1794,6 +1794,10 @@ h5 {
 /* ----------------------------------------- */
 /*  SRD Compendium                           */
 /* ----------------------------------------- */
+.srd-compendium.app .window-content {
+  background: url(../ui/parchment.jpg) repeat;
+  color: var(--color-text-dark-primary);
+}
```
Fyorl, line 1800: This needs to be changed in the LESS files (specifically `journal.less` for the SRD compendium files)

module/applications/actor/base-sheet.mjs
```diff
@@ -1240,7 +1240,7 @@ export default class ActorSheet5e extends ActorSheet {
     }
 
     const itemData = {
-      name: game.i18n.format("DND5E.ItemNew", {type: game.i18n.localize(`ITEM.Type${type.capitalize()}`)}),
+      name: game.i18n.format("DND5E.ItemNew", {type: game.i18n.localize(`TYPES.Item.${type}`)}),
```
Fyorl, line 1243: Let's use `CONFIG.Item.typeLabels[type]` here to get a v10/v11 agnostic string

module/applications/actor/character-sheet.mjs
```diff
@@ -49,14 +49,10 @@ export default class ActorSheet5eCharacter extends ActorSheet5e {
...
-    const inventory = {
-      weapon: { label: "ITEM.TypeWeaponPl", items: [], dataset: {type: "weapon"} },
-      equipment: { label: "ITEM.TypeEquipmentPl", items: [], dataset: {type: "equipment"} },
-      consumable: { label: "ITEM.TypeConsumablePl", items: [], dataset: {type: "consumable"} },
-      tool: { label: "ITEM.TypeToolPl", items: [], dataset: {type: "tool"} },
-      backpack: { label: "ITEM.TypeContainerPl", items: [], dataset: {type: "backpack"} },
-      loot: { label: "ITEM.TypeLootPl", items: [], dataset: {type: "loot"} }
-    };
+    const inventory = {};
+    for(const type of ["weapon", "equipment", "consumable", "tool", "backpack", "loot"]){
```
Fyorl, line 53:
```suggestion
    for ( const type of ["weapon", "equipment", "consumable", "tool", "backpack", "loot"] ) {
```

module/applications/item/item-sheet.mjs
```diff
@@ -422,7 +422,6 @@ export default class ItemSheet5e extends ItemSheet {
       html.find(".damage-control").click(this._onDamageControl.bind(this));
       html.find(".trait-selector").click(this._onConfigureTraits.bind(this));
       html.find(".effect-control").click(ev => {
-        if ( this.item.isOwned ) return ui.notifications.warn("Managing Active Effects within an Owned Item is not currently supported and will be added in a subsequent update.");
```
Fyorl, line 425: We need to version-gate this

Fyorl (changes requested):
Thanks for doing this work, Zhell. I've flagged some areas where we want to explicitly version-gate to retain v10 compatibility. I didn't mark all the places where we should convert to using `typeLabels` but I hope the remaining areas should be obvious enough. Even without the back-compat motivation, converting to use `typeLabels` where we can is a good change regardless.

Feel free to add `ifThen` as a handlebars helper in the dnd5e system to allow it to work in v10 too. Just mark it as `@deprecated since v11` so we can find it again to remove it later.

For the ActiveEffect `label` -> `name` migration, it feels too awkward and risky to attempt to version-gate every instance of it. I would recommend that we stick to `label` in the codebase for now until we can safely remove it whenever we decide to make v11 the minimum for the dnd5e system. We can suppress the deprecation warning with an entry in `CONFIG.compatibility`:
```js
CONFIG.compatibility.excludePatterns.push(/\bActiveEffect5e#label\b/);
```
But if you feel otherwise, I'm happy to hear your thoughts.

module/documents/active-effect.mjs
```diff
@@ -165,7 +165,6 @@ export default class ActiveEffect5e extends ActiveEffect {
 
     // Iterate over active effects, classifying them into categories
     for ( let e of effects ) {
-      e._getSourceName(); // Trigger a lookup for the source name
```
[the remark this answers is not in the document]
krbz999: From what I could tell, it didn't seem needed in v11. It was deprecated with no mention of a replacement. Initial testing didn't reveal any issues.

dnd5e.mjs
```diff
@@ -190,8 +192,9 @@ Hooks.once("ready", function() {
   _configureValidationStrictness();
 
   // Apply custom compendium styles to the SRD rules compendium.
-  const rules = game.packs.get("dnd5e.rules");
-  rules.apps = [new applications.journal.SRDCompendium(rules)];
+  const rules = {collection: game.packs.get("dnd5e.rules")};
+  const isV10 = !foundry.utils.isNewerVersion(game.version, "10.291")
```
Fyorl, line 196: Previously we've attached this to the dnd5e object to make it easier to reference elsewhere. We also should probably use `game.release.generation` rather than checking an explicit version like `291` to make sure this check stays accurate even in the face of a closing v10 release.
```
game.dnd5e.isV10 = game.release.generation < 11;
```

Fyorl (approved):
Nice work, thanks. I had one comment but I'll mark this as approved, feel free to make the suggested changes yourself or I will when I get around to it a bit later.

dnd5e.mjs
krbz999: I appended it to the `game.dnd5e` object, an also version gated the `_getSourceName` call.

PR #2282, opened 2023-05-15
v11 compatibility updates.

Fyorl:
- Added LevelDB files to .gitignore.
 - Ensure actor preparation does not throw an error for actors with module-provided sub-types.
 - Use new explicit trackableAttributes configuration to include derived attributes.

### DIFF
--- a/templates/apps/initiative-config.hbs
+++ b/templates/apps/initiative-config.hbs
@@ -5,7 +5,7 @@
         <label>{{localize "DND5E.AbilityModifier"}}</label>
         <div class="form-fields">
             <select name="system.attributes.init.ability">
-                {{selectOptions abilities selected=ability blank="" label="label"}}
+                {{selectOptions abilities selected=ability blank="" labelAttr="label"}}
             </select>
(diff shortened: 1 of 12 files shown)


templates/apps/initiative-config.hbs
```diff
@@ -5,7 +5,7 @@
         <label>{{localize "DND5E.AbilityModifier"}}</label>
         <div class="form-fields">
             <select name="system.attributes.init.ability">
-                {{selectOptions abilities selected=ability blank=""}}
+                {{selectOptions abilities selected=ability blank="" label="label"}}
```
arbron, line 8:
```suggestion
                {{selectOptions abilities selected=ability blank="" labelAttr="label"}}
```

(`skill-config.hbs` also throws a warning)

arbron (changes requested):

templates/apps/initiative-config.hbs
Fyorl: Whoops, thanks

Issue #2288, opened 2023-05-20
Despite `Item#hasLimitedUses` being false, the charges still appear on the character sheet


krbz999:
Same thing with 'activation type' being 'None'.

![image](https://github.com/foundryvtt/dnd5e/assets/50169243/037c8fa7-a512-4666-a05b-aa4803a427cc)

Issue #2286, opened 2023-05-20
If resource/limited uses are filled in, but the item has no activation type, prompts are still forced


krbz999:
![image](https://github.com/foundryvtt/dnd5e/assets/50169243/72af116f-3c38-4cc0-bc64-628d12f54e91)

Issue #2285, opened 2023-05-20
If resource inputs on an item are filled in but set to 'None', a prompt is still forced.


krbz999:
![image](https://github.com/foundryvtt/dnd5e/assets/50169243/e5b91e8c-b1a1-4361-8231-822bb343095c)

PR #2317, opened 2023-05-26
Reject prompts of several properties when an item has no activation type

krbz999:
Related issues: #2288, #2286, #2285.

The `isActive` getter returns whether an activation type is set in an item.

With this, if an item's Activation Type is set to 'None' (which also hides limited uses, recharge, resource consumption in the item sheet), their respective prompts will no longer be forced. I.e., an item can be "used" with activation type "none" and uses/target/recharge/consume filled in, but no prompts will be made.

Taken the liberty of cleaning up some labels on the actor sheet as well, since those were related.


Fyorl:
Marking this as blocked on #2387 for now

Fyorl (approved):
Good changes, thanks Zhell

PR #2319, opened 2023-05-26
Set feature items on NPCs as 'Monster Feature' by default

### LINKED ISSUE #2067: Set features created on NPCs to "Monster Feature" type by default
Currently features created on NPCs default to not having a specific type. It may be useful to assume they will be Monster Features and set that as default.

krbz999:
This adds a `preCreate` method for feature type items on npc sheets such that the items dropped or created are set to "Monster Feature" unless a feature type already exists on the item data.

Closes #2067.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1901,6 +1901,9 @@ export default class Item5e extends Item {
       case "weapon":
         updates = this._onCreateOwnedWeapon(data, isNPC);
         break;
+      case "feat":
+        updates = this._onCreateOwnedFeature(data, isNPC);
+        break;
     }
     if ( updates ) return this.updateSource(updates);
   }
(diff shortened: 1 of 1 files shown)


module/applications/actor/character-sheet.mjs
```diff
@@ -148,7 +148,7 @@ export default class ActorSheet5eCharacter extends ActorSheet5e {
         hasActions: false, dataset: {type: "class"}, isClass: true },
       active: {
         label: "DND5E.FeatureActive", items: [],
-        hasActions: true, dataset: {type: "feat", "activation.type": "action"} },
+        hasActions: true, dataset: {type: "feat", "system.activation.type": "action"} },
```
Fyorl, line 151: If we're changing these datasets, the PR/issue should probably be marked as breaking.

module/applications/actor/npc-sheet.mjs
```diff
@@ -125,4 +126,15 @@ export default class ActorSheet5eNPC extends ActorSheet5e {
...
     return super._updateObject(event, formData);
   }
+
+  /* -------------------------------------------- */
+  /*  Event Listeners and Handlers                */
+  /* -------------------------------------------- */
+
+  /** @override */
+  async _onDropSingleItem(itemData) {
+    const monsterFeature = (itemData.type === "feat") && !itemData.system.type.value && ("monster" in CONFIG.DND5E.featureTypes);
```
Fyorl, line 136: This sort of automatic data population is typically handled in `Item5e#_preCreate`, that way it will apply to programmatic creation as well. Unless you specifically wanted to avoid that?

Fyorl (changes requested):

module/applications/actor/npc-sheet.mjs
krbz999: I did not know of it at the time. It simplifies the PR greatly.

module/documents/item.mjs
```diff
@@ -2061,6 +2064,21 @@ export default class Item5e extends Item {
     return updates;
   }
 
+  /**
+   * Pre-creation logic for the automatic configuration of owned feature type Items.
+   * @param {object} data       Data for the newly created item.
+   * @param {boolean} isNPC     Is this actor an NPC?
+   * @returns {object}          Updates to apply to the item data.
+   * @private
```
Fyorl, line 2072: These `_onCreate` methods should probably all be `@protected` but it's fine to leave this for now to be consistent with the others.

Fyorl (approved):

PR #2320, opened 2023-05-26
[#2279] Update context on resting

### LINKED ISSUE #2203: Move Gritty Long Rest Duration into the CONFIG
<https://github.com/foundryvtt/dnd5e/blob/master/module/documents/actor/actor.mjs#L2137>

It would be nice to be able to modify this value instead of having it hardcoded as 7.

### LINKED ISSUE #2279: Supply an update option when an actor takes a short/long rest
It would be nice for general `updateActor` hooks to be able to tell if an actor update was caused by a rest.

krbz999:
Adds `isRest: true` to actor and item updates context when resting.

Closes #2279.


module/documents/actor/actor.mjs
```diff
@@ -1981,7 +1981,7 @@ export default class Actor5e extends Actor {
     if ( Hooks.call("dnd5e.preRestCompleted", this, result) === false ) return result;
 
     // Perform updates
-    await this.update(result.updateData);
+    await this.update(result.updateData, { isRest: true });
     await this.updateEmbeddedDocuments("Item", result.updateItems);
```
Fyorl, line 1985: We should supply an `isRest: true` to the Item update as well.

module/config.mjs
```diff
@@ -376,6 +376,16 @@ preLocalize("timePeriods");
 
 /* -------------------------------------------- */
 
+/**
+ * The duration of resting features.
+ * @enum {number}
+ */
+DND5E.restDurations = {
+  gritty: { long: { value: 7 } }
```
Fyorl, line 384: The rest variant durations are pretty well-defined in the rules. I feel like if you change the duration of the 'gritty realism' rest duration, then it's no longer 'gritty realism', it's a custom rest variant. In that vein, I'd propose we just have something like:

```js
DND5E.restDurations = {
  short: null,
  long: null
};
```

A value of `null` can indicate 'use the value of the rest variant in the system settings', and they can otherwise be configured with custom values like this:

```js
DND5E.restDurations = {
  short: { value: 30, unit: "minutes" },
  long: { value: 2, unit: "hours" }
}
```

I think we could also enable modules to more easily add their own rest variants to the settings menu by adding a `default` case to our `restVariant` `switch`es to use the configured values.

What do you think?

Fyorl (changes requested):
It's a little awkward to have *only* the gritty rest duration configurable. I have an alternative proposal below.

module/config.mjs
krbz999: Seems slightly overkill given that rest does not affect time at all. The original intent was simply to let the scaling value used for the dice rolls to be configurable, and going hog wild with the durations of all the resting types is just as easily done with a module or world script by overriding the localization since it's for display only.

Fyorl: > going hog wild with the durations of all the resting types is just as easily done with a module or world script by overriding the localization since it's for display only.

It's not for display only though, because the gritty realism long rest does have mechanical consequences, i.e. rolling for recharges 7 times. If you changed standard long rests to 2 days instead of 8 hours, it would then also have mechanical consequences, no?

I'm not proposing that we make resting fully configurable in this PR, but it seems like a solution that's over-fitted to our existing implementation. Maybe it would make more sense to just make the `isRest` changes in this PR and raise a separate task for a more comprehensive rest configuration.

Fyorl (approved):

PR #2324, opened 2023-05-27
Attunement tracker on character sheet

### LINKED ISSUE #1309: Track total attuned items on the actor sheet
Originally in GitLab by @Kandashi

Branching from #1021 specific issue for tracking Attuned Items (and possibly max slots) on the actor sheet.
No hard cap to prevent further attunement, just a visual guide. 

If I were to work towards this, would it be best to calculate these values in `ActorSheet5eCharacter#getData` or within the template?

krbz999:
This adds a smol attunement tracker on the character sheet under Inventory.

It is displayed in the same fashion as spell slot overrides, and should work with active effects overriding it (removing the edit icon).

[att.webm](https://github.com/foundryvtt/dnd5e/assets/50169243/3d8127f1-210d-4401-8bce-e57adb329324)

Closes #1309.

### DIFF
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -1126,6 +1126,15 @@ h5 {
   flex: 0 0 20px;
   justify-content: flex-end;
 }
+.dnd5e.sheet.actor .inventory-filters .attunement {
+  color: #7a7971;
+  font-size: 12px;
+}
+.dnd5e.sheet.actor .inventory-filters .attunement [name="system.attributes.attunement.max"] {
+  display: inline;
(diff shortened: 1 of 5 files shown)


Varriount:
Wouldn't it be better to add this as a button next to the item?

krbz999:
"The item"?

Zanderaf:
can this be changed to reflect increased attunement slots similar to what artificers get in tashas?

Fyorl (approved):

PR #2340, opened 2023-06-08
Update spell-icon-migration.json for 'explosion-star' icons

### LINKED ISSUE #2360: Some items have their icon path broken by system updates
Migration entries for `icons/magic/light/explosion...` are incorrect.

They are trying to convert paths that do exist such as `explosion-star-large-orange-purple` into paths that do not exist: `explosion-star-orange-purple-large`. This results in broken icon paths whenever the migration runs.

This notably breaks several spell icons, including the `Light` cantrip every time the system updates.

M0nk3yy:
Migration entries for `icons/magic/light/explosion...` are incorrect.

They are trying to convert paths that do exist such as `explosion-star-large-orange-purple` into paths that do not exist: `explosion-star-orange-purple-large`. This results in broken icon paths whenever the migration runs.

The core icons should probably be renamed so alike colours are next to each other and the schema is consistent. This is the case with other icons such as `beam-rays-orange-small.webp` where the size is at the end. Until that time, this change amends the migration to work with the current icon naming scheme.

Fixes #2360


LukasPrism:
After updating to 2.2.1 I’m experiencing the joy of fixing all the items again that broke in 2.1. Hopefully this isn’t still a problem in 2.3

M0nk3yy:
Fixes #2360

M0nk3yy:
@Fyorl Please consider this for 2.2.2 to prevent broken icon paths

json/spell-icon-migration.json
```diff
@@ -411,20 +411,20 @@
...
-  "icons/magic/light/explosion-star-small-teal.webp": "icons/magic/light/explosion-star-teal-small.webp",
-  "icons/magic/light/explosion-star-large-teal-purple.webp": "icons/magic/light/explosion-star-teal-purple-large.webp",
-  "icons/magic/light/explosion-star-small-teal-purple.webp": "icons/magic/light/explosion-star-teal-purple-small.webp",
-  "icons/magic/light/explosion-star-large-orange.webp": "icons/magic/light/explosion-star-orange-large.webp",
-  "icons/magic/light/explosion-star-small-orange.webp": "icons/magic/light/explosion-star-orange-small.webp",
-  "icons/magic/light/explosion-star-large-orange-purple.webp": "icons/magic/light/explosion-star-orange-purple-large.webp",
-  "icons/magic/light/explosion-star-small-orange-purple.webp": "icons/magic/light/explosion-star-orange-purple-small.webp",
-  "icons/magic/light/explosion-star-large-pink.webp": "icons/magic/light/explosion-star-pink-large.webp",
-  "icons/magic/light/explosion-star-small-pink.webp": "icons/magic/light/explosion-star-pink-small.webp",
+  "icons/magic/light/explosion-star-blue-large.webp": "icons/magic/light/explosion-star-large-blue.webp",
```
Fyorl, line 414: I can see in both v10 and v11 that the icon is called `explosion-star-blue-large`, so this change would map a correct icon filename to an incorrect filename, no? It looks like the original line was correct. What am I missing here?

Fyorl: I think I picked the first line which just happened to be a mistake. The two `explosion-star-blue` lines need to be reverted, but I checked the other lines and they all seem to be correct.

Fyorl (changes requested):
See comments about `explosion-star-blue`

M0nk3yy (commented):
Changes to lines 414 and 415 reverted

Fyorl (approved):

M0nk3yy:
Thanks Kim, good catch. I hadn't spotted that those two use a different naming format. The changes have been made.

PR #2341, opened 2023-06-08
Icon updates for various misc items in items.db

M0nk3yy:
I submitted a bunch of icon changes for tools/kits/supplies etc. back when the repo was on gitlab, and have been meaning to submit more for some time. Future passes will take a more systematic approach but this selection includes some that I already had prepared and some particularly egregious icons (chain, manacles).

Preview below with new icons on the _left_.

image
image


Fyorl (approved):
I noticed the `items.db` had 1.6k changes which seemed suspicious. Turns out all the line-endings got converted to CRLF. I re-packed the DB and pushed a commit to fix it and it looks good now. Just letting you know for the future. I thought our pack scripts or NeDB would always write LF endings rather than CRLF, but apparently not. It might be related to your git's `core.autocrlf` settings, but I don't know for sure.

Thanks for doing this work though, the icon updates look good.

PR #2344, opened 2023-06-13
Max pat931 wiki update

MaxPat931:
Update links in Home.md


MaxPat931:
Made several updates:
New pages for all advancement types, including Item Choice advancements (with description from https://github.com/foundryvtt/dnd5e/issues/1401 and videos from the 2.0.0 release)
Updates to Advancement User Guide - removed details about each advancement type for links to the Type's page, added section for Item Choice
Updated all links to point to `.md` files
Updated publish-wiki.yml to include preprocessing step to strip `.md` from the link when the wiki is generated and link to the appropriate wiki page

Pushed all changes to my master branch so you can preview the wiki there https://github.com/MaxPat931/dnd5e/wiki

Fyorl (approved):
Looks great, thank you for doing this. One thing I hope we can expand on later is that the Item Grant Advancement page doesn't show how the dialog changes when a spell item is granted. Some discussion/screenshots about that would be a good future improvement.

MaxPat931:
Absolutely, I think I was more focused on getting an MVP of the existing docs over to this new process and didnt want to change too much, but certainly there are some updates that should be made.

What are your thoughts on a separate `Wiki` milestone and/or branch? Because these files are pretty separated from the code, I don't see why the commits would have to be tied to system releases, but I can understand it would add another layer of complexity.

Fyorl:
> What are your thoughts on a separate `Wiki` milestone and/or branch? Because these files are pretty separated from the code, I don't see why the commits would have to be tied to system releases, but I can understand it would add another layer of complexity.

Yes, something like a `wiki` branch that we could use to trigger the wiki build instead of `master` would be ideal I think. I'm just not sure of the best way to keep it in-sync with the current development branch yet though.

MaxPat931:
Would it need to be in-sync if we are only making changes to the wiki folder's files in that branch?
By creating a `wiki-master` branch and updating the yml to:
```yml
on:
  push:
    branches: [wiki-master]
    paths:
      - wiki/**
      - .github/workflows/publish-wiki.yml
```
it would only be looking for commits in that branch to the wiki folder (and the yml file) to trigger the action and build the Wiki, and I guess we wouldnt even need the wiki folder in the master branch at all.

Fyorl:
> Would it need to be in-sync if we are only making changes to the wiki folder's files in that branch?

No, we technically wouldn't need to. It just seems a bit weird to have a separate branch that has files which the main branch, and the rest of the repository never see. But maybe it's not that weird.

I think I'm ok with having a `wiki` branch that is the only branch that has the `wiki` folder, and then periodically merging `master` into the `wiki` branch every time we release. It's ok for `master` to never see the `wiki` folder, but I'd prefer that our `wiki` branch's code wasn't just the state of the dnd5e repo at some random point in time. If we only keep in-sync one way then it's pretty easy to do I think.

MaxPat931:
Sounds good to me! 
Is there anything that you'd like me to do for that?

Fyorl:
> Sounds good to me! Is there anything that you'd like me to do for that?

No, I don't think so. I can make the necessary changes here. I was just going to call the branch `wiki`. Do you have any strong preference for `wiki-master` instead?

MaxPat931:
Not specifically `wiki-master` but I do think a little something to differentiate it would be helpful, since we reference "wiki" a lot; the `wiki` branch, `wiki` folder, and the actional Wiki

Fyorl:
Alright, I created a dedicated branch for building the wiki called `publish-wiki`, and amended the action file. It seems to have worked. I did a quick look around the wiki and it seems in order. Very smooth, thank you for doing this.

I'm going to update the `2.3.x` branch to remove any wiki specific stuff now.

PR #2345, opened 2023-06-13
pack folders

### LINKED ISSUE #2310: Add `packFolders` to system manifest

MaxPat931:
Resolves #2310 

Implements Compendium Folders for new worlds with the following structure
```
**DnD5e SRD Content** #cd2c1e
    rules
    tables
    heroes
    **Character Features** #b72c1f
        classes
        subclasses
        classfeatures
        backgrounds
        races
    **Items & Spells** #941919 
        items
        spells
        tradegoods
    **Monsters** #7d160c
        monsters
        monsterfeatures
```

image

### DIFF
--- a/system.json
+++ b/system.json
@@ -2,14 +2,14 @@
   "id": "dnd5e",
   "title": "DnD5e - Fifth Edition System",
   "description": "A system for playing the fifth edition of the worlds most popular role-playing game in the Foundry Virtual Tabletop environment.",
-  "version": "2.2.0",
+  "version": "2.2.1",
   "compatibility": {
     "minimum": "10.291",
     "verified": "11"
   },
   "url": "https://github.com/foundryvtt/dnd5e/",
   "manifest": "https://raw.githubusercontent.com/foundryvtt/dnd5e/master/system.json",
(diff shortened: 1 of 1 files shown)


Fyorl:
This looks fine to me, but I don't want to be too opinionated here, as it will affect everyone, and we can't change it after the fact—any changes will only be reflected in new worlds. My initial thought was that we put them in the same categories as they were in v10, e.g. have an 'Actors' folder and an 'Items' folder, though I appreciate that we're not taking advantage of the opportunity to provide some more bespoke categorisation.

If you could update the description with a text representation of what the proposed folder structure is, so people aren't required to parse the JSON themselves, that would be appreciated. Then I'd prefer to have a few more eyes from the community on the proposed structure.

MaxPat931:
Yea sounds good to me, having some chatter about it in discord here https://discord.com/channels/170995199584108546/670336046164213761/1131635144235679785

Perhaps the least opinionated option would just be one "master" folder and users could add folders as they desire, but I like the idea of having an example of nested folders. (and looking at the code it looks like I forgot the Parent folder I intended so depending on the feedback I'll make that update)

MaxPat931:
Updated per community feedback in Discord, added folder structure in the initial comment

Fyorl (approved):
This looks like a pretty good arrangement to me, thanks for doing this work soliciting feedback from the community here.

Issue #2289, opened 2023-05-20
Item5e.use does not respect config.createMeasuredTemplate=false for area effect spells


pgervais70:
When calling Item5e.use with config.createMeasuredTemplate=false on a spell with an area effect this value is not taken into account when calling AbilityUseDialog because it only takes the item as parameter and it shows the create measured template checkbox based on the item having an area effect. 
The createMeasuredTemplate flag passed from the config object should have priority over the default behavior.

PR #2346, opened 2023-06-14
Fix item use create measure template option not working with aoe spells

elwin6446:


module/applications/item/ability-use-dialog.mjs
```diff
@@ -48,7 +49,7 @@ export default class AbilityUseDialog extends Dialog {
       consumeResource: resource.target && (!item.hasAttack || (resource.type !== "ammo")),
       consumeUses: uses.per && (uses.max > 0),
       canUse: recharges ? recharge.charged : sufficientUses,
-      createTemplate: game.user.can("TEMPLATE_CREATE") && item.hasAreaTarget,
+      createTemplate: game.user.can("TEMPLATE_CREATE") && (config?.createMeasuredTemplate ?? item.hasAreaTarget),
```
krbz999, line 52: Would this not cause an issue if passing `createMeasuredTemplate: true` while `item.hasAreaTarget` is `false`?
It seems like the purpose of the `config` object in this context is to explicitly only disable some properties in the `data` object, but here it is taking priority. Maybe adding `config.createMeasuredTemplate !== false` instead would work (or using `mergeObject`), like,
```js
createTemplate: game.user.can("TEMPLATE_CREATE") && item.hasAreaTarget && (config.createMeasuredTemplate !== false),
```

elwin6446: A change was pushed to make sure that it cannot be true if is not an AOE

Fyorl:
What issue is this fixing? Is there an associated bug report in the tracker?

elwin6446:
> What issue is this fixing? Is there an associated bug report in the tracker?

It fixes https://github.com/foundryvtt/dnd5e/issues/2289

Fyorl:
Marking as blocked on #2387

Fyorl:
This should now be resolved on the `2.4.x` branch and `AbilityUseDialog` will now be better-behaved about respecting options passed to it. Please feel free to re-open the original issue if not.

Issue #1448, opened 2022-03-01
[Suggestion] Add tags to features


aaclayton:
Originally in GitLab by @danz5000

Currently weapons are really cool; we can tag things like if the damage is magical, if it's versatile, is heavy, or has special properties.

Could we get the same for features?

Less relevant, but could we consider adding "magic resistance grants advantage on saves against this" and "requires concentration" to the list of tags as well? Thanks!

PR #2356, opened 2023-07-03
Add `magical` field to `PhysicalItemTemplate`

dev7355608:
By RAW only magic items have a rarity. So technically magic items are those items that have a rarity and mundane items are those that do not. But that is unreliable, because some people use rarity for mundane items or rule that common magic items are nonmagical.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -780,6 +780,7 @@
 "DND5E.LongRestResultHitDice": "{name} takes a long rest and recovers {dice} Hit Dice.",
 "DND5E.LongRestResultHitPoints": "{name} takes a long rest and recovers {health} Hit Points.",
 "DND5E.LongRestResultShort": "{name} takes a long rest.",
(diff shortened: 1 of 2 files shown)


Fyorl:
Weapons (and now ammunition) already have a 'Magical' option as part of their `properties`, so we would probably want to migrate that if we go with this strategy, and we'll need a way to configure this field via the UI.

The following is obviously more ambitious than this PR, but I personally feel that our strategy here should be:

1. Generalise the concept of `properties` for *all* items by adding the field to the model for all item types. We can use `MappingField` and `initialKeys` to customise the properties on a per-type basis.
2. Migrate spell `tags` to be `properties` instead.
3. Add 'Magical' and 'Concentration' properties to `feature` items, per #1448.

Does that sound reasonable?

arbron:
Being handled as part of https://github.com/foundryvtt/dnd5e/pull/2681

Issue #2382, opened 2023-07-25
Scrolling Status Text animations should be suppressed when the triggering token is toggled hidden


unsoluble:
(Moved from https://github.com/foundryvtt/foundryvtt/issues/9833)

Scrolling Status Text animations are visible to all players with line of sight to the token, even if the token has been toggled hidden with the Toggle Visibility State control. The text animations ought to be suppressed in that case.

This behaviour is present in both 10.303 and 11.306

### Reproduction Steps

1. Be in a game system that supports Scrolling Status Text
2. Enable Scrolling Status Text
3. Put a token within a Player's vision polygon
4. Toggle Visibility State on the token, making it hidden from the Player
5. Change a value on the token that triggers a Scrolling Status Text animation
6. View from the Player's login that the text animation is visible

### Bug Checklist

- [X] The issue occurs while all Modules are disabled

krbz999:
Are you seeing scrolling text on invisible tokens for conditions being toggled on or off or only damage numbers? I am not seeing the former, but slapped in a fix in #2383 for damage numbers specifically.

unsoluble:
Only tested with HP changes, so perhaps that's all that needs to be hit.

PR #2383, opened 2023-07-25
[#2382] Do not display scrolling damage numbers if a token cannot be seen by a user

krbz999:


module/documents/actor/actor.mjs
```diff
@@ -2694,6 +2694,7 @@ export default class Actor5e extends Actor {
     dhp = Number(dhp);
     const tokens = this.isToken ? [this.token?.object] : this.getActiveTokens(true);
     for ( const t of tokens ) {
+      if ( !t.isVisible ) continue;
```
Feu-Secret, line 2697: I would use the same logic from core. `isVisible` is too much overhead here I think. 
```suggestion
      if ( !t.visible || !t.renderable ) continue;
```

PR #2385, opened 2023-07-26
Transformation fixes

### LINKED ISSUE #1821: Selecting 'keep mental/physical ability scores' when transforming an actor also keeps the host's saving throw proficiencies for those abilities
## Steps to Reproduce
1. Import one of the Starter Heroes and note their saving throw proficiencies.
2. Import another creature that lacks those saving throw proficiencies.
3. Initiate a transform from the hero into the creature, and select 'keep mental/physical ability scores', but leave 'keep saving throw proficiencies' unchecked.
4. Observe that the transformed actor retains their saving throw proficiencies regardless.

The creature's saving throw proficiencies are also lost, they are essentially overridden by the hero's (host's).

The expected behaviour when 'keep saving throw proficiencies' is unchecked is that the resulting actor has the saving throw proficiencies of the transform target, not the host. This should be independent of whether 'keep mental/physical ability scores' is checked or not.

### LINKED ISSUE #1822: Selecting 'keep spells' when transforming an actor should probably keep the host's spellcasting ability
Currently the transformed actor inherits the spellcasting ability of the transform target, rather than that of the host. Not technically a bug, but I think it's likely the intention of the user when they have 'keep spells' checked.

krbz999:
Closes #1821, closes #1822

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -2281,7 +2281,7 @@ export default class Actor5e extends Actor {
    * @property {boolean} [mergeSkills=false]        Take the maximum of the skill proficiencies
    * @property {boolean} [keepClass=false]          Keep proficiency bonus
    * @property {boolean} [keepFeats=false]          Keep features
-   * @property {boolean} [keepSpells=false]         Keep spells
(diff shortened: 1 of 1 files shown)


Fyorl (approved):
Looks good, thanks for these fixes

PR #2386, opened 2023-07-26
Improve `SystemDataModel`

dev7355608:
These changes make it possible to use `cleanData`, `validateJoint`, and `shimData` in a template.


Fyorl (approved):
The motivation for the changes here make sense to me, though I would suggest we hold off on merging this in 2.3 and instead merge it early in 2.4 if we decide to go ahead with it.

PR #2393, opened 2023-07-29
Fix Devil's Sight spelling and remove the effect from the Devil's Sight invocation that grants 120 ft Darkvision

dev7355608:
Devil's Sight isn't Darkvision.


arbron (approved):

arbron:
@dev7355608 Could you rebase this upon `3.0.x`, the paths to the source files have changed preventing this from being merged.

dev7355608:
@arbron Rebased.

Fyorl:
I assume this was added because it's the closest approximation to darkvision that we have in the system currently. I am ok with removing it, but maybe it would be better to wait until we have an adequate replacement.

Issue #2257, opened 2023-04-07
Euclidean measurement using `canvas.grid.measureDistance` with `{gridSpaces: true}` should probably return Euclidean measurement


caewok:
Foundry: v10.291
dnd5e: v2.1.5
Browser: Version 112.0.5615.49 (arm64)
browserclient OS: macOS 13.3
Running Foundry on NodeJS locally
No modules running.

The values returned `canvas.grid.measureDistance` for `{gridSpaces: true}` and `{gridSpaces: false}` is inconsistent between the different distance settings and therefore problematic when using the ruler to measure values. Currently, the three distance settings return different values depending on the value of `gridSpaces` (assuming a grid; non-gridded appears to always return Euclidean):

|  Setting     |   gridSpace         | value  |
| ------------- |:-------------:| -----:|
| 5/5/5| true | 5/5/5 distance |
| 5/10/5 | true | 5/10/5 distance |
| Euclidean | true | ?? some sort of grid rounding |
| 5/5/5      | false      |   Euclidean distance |
| 5/10/5 | false      |   Euclidean distance  |
| Euclidean | false | Euclidean distance |

The Foundry ruler uses `{gridSpaces: true}` to determine distance. Thus, in dnd5e, the ruler returns the expected 5/5/5 and 5/10/5 distance, but returns some non-Euclidean value (a rounded value based on the grid?) when the distance setting is Euclidean. In addition, any module that wants to obtain the gameplay distance would likely also use `{gridSpaces: true}` and end up with the same incorrect result for the Euclidean distance setting. I would argue that the user, having set the preferred distance measurement to "Euclidean," would expect to see the Euclidean measurement at all times. 

Note that this really only affects long diagonal distance measurements; shorter distances tend to get rounded and so are mostly correct.

My recommendation is to change the measurement to return the same value when `gridSpaces: true` or `gridSpaces:false` if the Euclidean distance setting is selected. (I think it is reasonable to return Euclidean measurements whenever `gridSpaces:false` because there are instances when the true distance is required regardless of the user's chosen distance setting.)

PR #2403, opened 2023-08-08
[#2257] Fix: Remove rounding from Euclidean measurement

dev7355608:


module/canvas/grid.mjs
```diff
@@ -12,8 +12,8 @@ export function measureDistances(segments, options={}) {
     let r = s.ray;
 
     // Determine the total distance traveled
-    let nx = Math.abs(Math.ceil(r.dx / d.size));
-    let ny = Math.abs(Math.ceil(r.dy / d.size));
+    let nx = Math.ceil(Math.abs(r.dx / d.size));
+    let ny = Math.ceil(Math.abs(r.dy / d.size));
```
dev7355608, line 16: We have the same issue in core. But I guess it doesn't matter, because the segments passed by the `Ruler` always start and end in grid space centers. Either way it wouldn't give the right results if the segments didn't start and end in grid space centers. We should get the grid coordinates of the grid spaces that contain the origin and target of the rays and calculate the distance this way.

Fyorl (approved):

PR #2415, opened 2023-08-11
[#2404] Ignore honor and sanity for ability score improvements

### LINKED ISSUE #2404: Allow for an ability score to be entirely ignored by ASI advancements
For example, the DMG says honor cannot be raised by ASIs. Other optional abilities, like piety, are raised in different ways only as well.

krbz999:
This adds `improvement` to entries in `CONFIG.DND5E.abilities` which, when explicitly `false`, should entirely ignore the given ability for the purpose of Ability Score Improvement advancements.

Closes #2404


Fyorl (approved):
Thanks for doing this work. I made some adjustments to add a `canImprove` method which avoids having to repeat the `CONFIG.DND5E.abilities[ability].improvement` check, and allows subclasses to control this behaviour if they ever wanted to.

I also added some additional code to handle cases where there are existing advancements that have Honour/Sanity bonuses. Now they won't contribute to the unallocated points total and won't be increased or decreased when the advancement is applied or reversed.

PR #2424, opened 2023-08-12
[#2245] Stack consumables on the Group actor sheet

### LINKED ISSUE #2245: Group actor sheet does not stack consumables
They always create a different stack when drag-and-dropped unlike all the other actor sheets.

krbz999:
... This is getting silly.

Eventually the group sheet should simply extend the Actor5e sheet, or base system sheet implementation should be separated into an abstract utility sheet that can get mixed into specific actor sheets, possibly splitting them in two groups; 'living sheets' (character, npc, vehicle) and 'other sheets' (group).

Closes #2245


Fyorl:
> Eventually the group sheet should simply extend the Actor5e sheet, or base system sheet implementation should be separated into an abstract utility sheet that can get mixed into specific actor sheets, possibly splitting them in two groups; 'living sheets' (character, npc, vehicle) and 'other sheets' (group).

We keep talking about this but never raised an issue so I have done so now #2545. It would also work as a mixin if we want.

Fyorl (approved):

PR #2425, opened 2023-08-12
[#986] Allow for rotating a square/cube template

### LINKED ISSUE #986: Casting a square/cube spell creates a measured template of the wrong type
Originally in GitLab by @foresto

When casting a 5e spell with a square or cube area of effect, such as Faerie Fire or Entangle, Foundry (optionally) creates a measured template of *rectangle* type. This is the wrong template type, since it cannot be rotated without deforming it. The system should instead create a template of *ray* type with its width equal to its distance.

See also: [aaclayton's explanation](https://gitlab.com/foundrynet/foundryvtt/-/issues/4300#note_467221232)

krbz999:
_It is now hip to be square._

Ideally the ability to rotate a rectangle without changing its height or width would be a core improvement. However given that the system can neither rotate properly while placing it, nor afterwards, setting squares and cubes to be equal-side ray templates is a good solution in the meantime.

Closes #986


Fyorl:
A good change but let's add a system setting to allow DMs to have the old behaviour and prevent players from being able to rotate square templates since that is apparently a preference some DMs have.

Fyorl (approved):
Code looks good. I think we should lend some weight to maintaining the status quo here so that there are no surprises when people update, but I don't feel that strongly.

Fyorl:
@krbz999 I've made a change that I hope is a reasonable compromise. The setting is now renamed to frame it as opt-in to *prevent* square templates from being rotated. The setting is on by default however as that matches the system's previous behaviour, so this way it's a smoother migration path.

If you feel very strongly, we can discuss and perhaps revert the commit on your branch, but otherwise hopefully we can proceed.

krbz999:
Fine by me. Yeet.

PR #2428, opened 2023-08-13
[RFC] Rename Bardic Inspiration identifier to align with community modules

TheGiddyLimit:
The D&D Beyond importer module uses a more precise name for the bardic inspiration scale than that which is currently used by the SRD classes available with the dnd5e system. Automation modules are therefore likely to follow suit, and some have already done so (e.g. Chris's Premades).

Flip the SRD identifier to better match developing community expectations.

See e.g.:

- https://github.com/MrPrimate/ddb-importer/blob/e7fcbfec18000b671519928e0e2c51dd614df296/src/parser/classes/index.js#L52
- https://github.com/chrisk123999/chris-premades/blob/master/packs/cpr-class-features.db (CTRL+F `"@scale.bard.bardic-inspiration"`)

---

Disclaimer: haven't looked at what migration/how migrations are handled. Happy to do the work if this change is palatable.


krbz999:
This seems to touch the class item only, but not the features that use the scale value. Those would have to be adjusted as well, in which case there is a consideration for what would happen for existing actors that have the class, but not all the features, as well as those that use the SRD bard class in conjunction with their own bard subclass item.

When imported from the SRD compendium or from user's personal compendiums, each user would have to manually change the scale value to match one or the other. The alternative is to continuously migrate the name of the scale value (`inspiration` to `bardic-inspiration`) for all classes and feature items, in which case users will be fully unable to use their own (and shorter) name for the scale value.

There are probably nuances to a possible migration method I overlook here, but might it actually be more straightforward for importers and premade items modules like the ones mentioned to adjust and align with the SRD, given a user base that more often continuously updates their actors' items either manually or automatically?

I don't see what's best here, but those are my personal thoughts on the matter.

Fyorl:
> Disclaimer: haven't looked at what migration/how migrations are handled. Happy to do the work if this change is palatable.

I'm not even sure what a migration would look like here. Already-existing Actors would have Advancements configured with the old identifier and would continue to work fine. Newly-created Actors would get the new identifier when they dragged the Bard class from the SRD compendium. It would probably only be a problem if there are external modules or macros that are looking for a specific identifier, but those would run into trouble regardless since identifiers are user-editable and can be whatever the user wants, ultimately.

That said, unless there are severe technical challenges preventing them from doing so, we would generally encourage modules to align with the reference implementation where it exists. I don't see why this particular situation would be any different, so I'm tentatively marking this PR as won't change.

Fyorl:
>I'm not even sure what a migration would look like here. Already-existing Actors would have Advancements configured with the old identifier and would continue to work fine.

Actually, I just realised this would not be the case. An Actor that levels up will pull in an Item configured with the new identifier from the SRD compendium as part of their Advancement, while all their existing Items use the old identifier, creating an issue.

This would be a pretty complex migration to target all class and subclass Items across all Actors. So it makes even less sense to make this change and place a migration burden on *all* users of the dnd5e system for the sake of modules that they might not even use.

TheGiddyLimit:
Reasonable

PR #2434, opened 2023-08-23
Add more detail to formula simplification failure warning message

Gawdl3y:
This adds the item ID and the parent name and ID to the warning message output in the console when the formula simplification fails for an item, and also provides the full error object rather than just the message. The motivation behind this is to make it much easier to identify the source of this warning.


Fyorl (approved):

Fyorl (approved):
An easy win, thank you

PR #2447, opened 2023-08-28
[#2446] Toggle inspiration on the header element

### LINKED ISSUE #2446: Click events on the `H4` element for inspiration on the `character` sheet should toggle inspiration
It's larger and easier to hit than a small checkbox so would be pretty neat.

krbz999:
Closes #2446

### DIFF
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -1045,6 +1045,9 @@ h5 {
   height: 20px;
   line-height: 20px;
 }
+.dnd5e.sheet.actor .counters .counter label {
+  cursor: pointer;
+}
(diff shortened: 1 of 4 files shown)


Fyorl (approved):
Sorry for reaching in and rewriting your PR, but it was so small that I felt it wasn't worth the back-and-forth. I didn't want us to use `rollable` here as it's a UI cue that we have established that implies something will be rolled, but inspiration isn't rolled. I've used `<label>` here as the typical way to connect some label text to checkbox. Technically `<h4>` is not valid inside a `<label>` but I didn't want to mess with the styles too much for this small PR.

PR #2455, opened 2023-09-03
Update loot sheet with tabs

tavernaDelBeerholder:
Several modules tend to work with the concept of adding a tab to the item sheet unfortunately the loot sheet is the only one that "thinks" differently, to make it uniform and help developers can we put the tabs back to the loot sheet ?
I also think it is visually better because then people find the same coporting.


p4535992:
Just a clarification is the pr okay to point it to 2.3.x or should I point it directly to 2.4.X ?

Fyorl:
> Just a clarification is the pr okay to point it to 2.3.x or should I point it directly to 2.4.X ?

I will typically re-target branches just before I merge them. But if you'd like to do it yourself, that's fine too, it will save me from accidentally making a mistake and merging it into `2.3.x`.

p4535992:
I beg you to consider this PR

Fyorl:
> I beg you to consider this PR

The PR is small and uncontroversial so I have placed it in the milestone. The only delay is that I need to do a little bit of testing before merging, which I haven't had time for yet.

Fyorl (approved):

PR #2458, opened 2023-09-06
Rework build scripts to avoid the use of gulp

arbron:
Remove `gulp` and related dependencies to avoid outstanding security vulnerabilities and clean up the CLI a bit.


Varriount:
Just curious, why?

arbron:
> Just curious, why?

Vulnerabilities with Gulp dependencies. Since it hasn't had any new releases since 2019 the chance of those being fixed doesn't seem high.

Fyorl:
I just want to note that, since `gulp` is a dev dependency, users are not exposed to any of these vulnerabilities, so it is not a super high priority or anything. It does mean there's a fair bit of noise when running `npm audit` though, meaning we could miss more serious vulnerabilities. Since we can replace `gulp` fairly easily in this case then we might as well.

Fyorl:
One additional comment is that there seems to have been a critical vulnerability raised against `@babele/traverse` so feel free to bump some dependency versions additionally, or fix it with `npm audit` before we merge this.

PR #2467, opened 2023-09-09
Add dawn and dusk uses type, add 'none' as truthy activation type

### LINKED ISSUE #2462: Implement 'Dawn' and 'Dusk' as options for Limited Uses
They would function identically to 'Charges' and also allow for a recovery formula to be input on the item sheet, but could be handed differently if time management of any sort is implemented.

### LINKED ISSUE #2465: Reimplement 'None' as a valid activation type
Previously "None" was a truthy value, which easily served this purpose, allowing for creation of items with ranges, limited uses, etc., but nowadays 'Special' is the best fit.

However several actions are very clearly labeled as not requiring any action where 'Special' shouldn't need to serve as a catch-all.

### LINKED ISSUE #2466: Support 'Piety' as an ability score
This should have a default value of 1 on characters and npcs, and 0 for vehicles.

krbz999:
Closes #2462.
Closes #2465.


module/config.mjs
```diff
@@ -538,6 +547,18 @@ preLocalize("itemRarity");
...
 
+/**
+ * The limited use periods that support a recovery formula.
+ * @enum {string}
+ */
+DND5E.limitedUseFormulaPeriods = {
+  charges: "DND5E.Charges",
+  dawn: "DND5E.Dawn",
+  dusk: "DND5E.Dusk"
+};
```
Fyorl, line 558: Might be nice to display these in an `<optgroup>` in the drop-down on the sheet, but not a requirement. It can be follow-up work.

Fyorl (approved):

PR #2471, opened 2023-09-10
[#968] Update base-sheet.mjs

### LINKED ISSUE #968: In character sheet, split special senses with semi-colon
Originally in GitLab by @CDeenen

On the attributes page of the character sheet you can add custom attributes with the 'special' field such as for languages, damage immunities, etc. 
For pretty much all of these, you can split multiple attributes with a semi-colon, however, this does not work for 'Senses'. Could this be added?

Hoppyhob:
Closes #968 
Modified _getSenses(systemData)
Removes senses.special
Adds a senses.customN for each special senses separated by a semi-colon (N being the index starting from 1)


aaclayton (approved):
Looks good to me and matches the way that other custom traits are handled. @Fyorl if this looks good to you feel free to merge in.

PR #2475, opened 2023-09-15
Replace the core conditions with system-specific effects

### LINKED ISSUE #1279: Create (Active) Status Effects for all Conditions
Originally in GitLab by @akrigline

At the moment the default set of effects that come with Foundry Core does not completely cover the Conditions list of 5e. We should add the missing ones in `CONFIG.statusEffects`.

- [x] Blinded
- [ ] Charmed
- [x] Deafened
- [ ] Exhaustion
    - [ ] Level 1
    - [ ] Level 2
    - [ ] Level 3
    - [ ] Level 4
    - [ ] Level 5
    - [ ] Level 6 (Death)
- [x] Frightened
- [ ] Grappled
- [ ] Incapacitated
- [ ] Invisible
- [x] Paralyzed
- [ ] Petrified
- [x] Poisoned
- [x] Prone
- [x] Restrained
- [x] Stunned
- [x] Unconscious

Some additional status effects which would be extremely useful that are not strictly conditions:
- [ ] Concentrating

---

Additionally, it would be ideal if the effects conditions which can be automated through active effects are.

For example: Grappled, Paralyzed, Petrified, Prone, Restrained, Stunned, and Unconscious all set the creature's speed to 0, we can affect that like so:

| Attribute Key | Change Mode | Effect Value |
| -------- | -------- | -------- |
| `data.attributes.movement.walk`     | Override     | `0`     |
| `data.attributes.movement.fly`     | Override     | `0`     |
| `data.attributes.movement.swim`     | Override     | `0`     |
| `data.attributes.movement.burrow`     | Override     | `0`     |
| `data.attributes.movement.climb`     | Override     | `0`     |

### LINKED ISSUE #2390: Harmonize condition types and statuses
As of this writing (Foundry 11, dnd5e 2.2), condition types come from the system and statuses come directly from the Foundry default set (as far as I can tell).

This leads to the fact that an immunity can't be matched to the condition and so on in modules that care about such things, due to different wordings in the ids, e.g. `stun`/`stunned`, `restrain`/`restrained`.

Probably related to #1279.

krbz999:
Closes #1279
Closes #2390

The conditions that exist now are (in order represented on the token HUD):
- dead,
- bleeding,
- charmed,
- cursed,
- diseased,
- exhaustion,
- frightened,
- poisoned,
- blinded,
- deafened,
- silenced,
- grappled*,
- incapacitated,
- paralyzed*,
- petrified*,
- prone,
- restrained*,
- sleeping*,
- stunned*,
- unconscious*,
- invisible,
- marked,
- flying,
- burrowing,
- concentrating,
- transformed,
- hidden

These should be aligning with the default `CONFIG.DND5E.conditionTypes`.

_*These effects apply actual changes to the actor, mostly movement reductions._

Some of these effects also apply several statuses (e.g. 'Sleeping' is treated as 'Unconscious', which also causes 'Incapacitated', ergo 'Sleeping' applies three statuses, and 'Petrified' applies petrified, incapacitated, and silenced). This all shows on the token HUD.


module/config.mjs
```diff
@@ -1897,6 +1897,170 @@ DND5E.advancementTypes = {
...
+    name: "DND5E.ConDiseased",
+    description: "DND5E.ConDiseasedText",
+    icon: "icons/svg/biohazard.svg"
+  },
+  frightened: {
+    name: "DND5E.ConFrightened",
+    description: "DND5E.ConFrightenedText",
+    icon: "icons/svg/terror.svg"
+  },
+  grappled: {
```
aaclayton, line 2055: I have one key concern here that I'd hope to discuss before moving forward regarding directionality and control.

Does the `ActiveEffect` include a condition? Or does the condition generate the Active Effect? I prefer the first path, but the data structures in this PR imply the second path.

The design that core Foundry now has is that ActiveEffect embedded documents each contain a set of `statuses`, and those respective status sets can be unioned to define a set of `statuses` which currently apply to the Actor.

This allows us to encapsulate more nuanced logic inside methods like `#prepareMovementSpeed`. For example consider a creature that has an ActiveEffect that imparts the `grappled` status, but also has condition immunity to `grappled`. This approach where the condition generates a certain active effect breaks down, whereas a scenario where we prepare actor data conditional on the actor's current statuses allows us to automate those sorts of immunities or interactions.

I think this is very important to shift, and I may not have explained myself fully or clearly so I hope we can discuss.

lang/en.json
```diff
@@ -371,22 +371,49 @@
 "DND5E.ContextMenuActionEnable": "Enable",
 "DND5E.ContextMenuActionDisable": "Disable",
 "DND5E.ConBlinded": "Blinded",
```
aaclayton, line 404: As part of this rework I'd suggest maybe moving towards a nested namespace like:
```json
{
  "DND5E": {
    "CONDITIONS": {
      "Blinded": "Blinded",
      "BlindedTooltip": "A blinded creature can't see ..."
     }
   }
}
```
That being said, as related to a separate discussion regarding Rules Pages we may  or may not need BlindedTooltip in the localization.

aaclayton (changes requested):
Thank you @krbz999, I would like a chance to discuss with @Fyorl and @arbron but hopefully we can all move this forward soon.

module/config.mjs
krbz999: > I may not have explained myself fully or clearly so I hope we can discuss.

What I am understanding is that the conditions here should not have an array of `changes`.

That's all easily amended, though I'm interested in seeing how we will handle the 'changes' inside data preparation given that `CONFIG.DND5E.conditionTypes` is configurable.

PR #2482, opened 2023-09-22
Calego's AE guide

### LINKED ISSUE #2481: Wiki - Add Active Effect Guide
Calego's Active Effect Guide https://hackmd.io/@foundryvtt-dnd5e/active-effects is one of the most commonly referenced links in the Discord's pins, adding it to the wiki will be a major benefit to making the wiki a more centralized source of information, and also allows us to make updates to the information as required.

Thanks to Calego for sending me the md file so that I could migrate it over. 🙏

MaxPat931:
Adds a guide for Active Effect to the wiki

Closes #2481 

tried to include some of the additional md styles outlined here https://github.com/orgs/community/discussions/16925 like this:  
> [!WARNING]  
but it doesnt seem to work in the wiki for some reason https://github.com/orgs/community/discussions/16925#discussioncomment-6883415 hopefully soon, seems potentially useful

I did double check some info from the original guide and made updates/removed changed info, but a second pass would be helpful

Preview on my wiki here: https://github.com/MaxPat931/dnd5e/wiki/Active-Effect-Guide

### DIFF
--- a/wiki/Active-Effect-Guide.md
+++ b/wiki/Active-Effect-Guide.md
@@ -232,7 +232,7 @@ E.g. An Item or Spell which sets the Actor's AC to `12 + Int` for the duration.
 | Attribute Key | Change Mode | Effect Value |
 | -------- | -------- | -------- |
 | `system.attributes.ac.calc`     | Override     | `custom`     |
-| `system.attributes.ac.formula`     | Custom     | `12 + @abilities.int.mod`     |
+| `system.attributes.ac.formula`     | Override     | `12 + @abilities.int.mod`     |
 
 ---
 
--- a/wiki/Active-Effect-Guide.md
+++ b/wiki/Active-Effect-Guide.md
@@ -1,0 +1,410 @@
+# FoundryVTT dnd5e Active Effects Examples
+###### tags: `active-effects`
+
+![Up to date as of 2.1.x](https://img.shields.io/badge/dnd5e-v2.1.x-informational)
+
+See [Kandashi's Active Effects Guide](https://docs.google.com/document/d/1DuZaIFVq0YulDOvpahrfhZ6dK7LuclIRlGOtT0BIYEo) for a trove of useful information
+
+This document only covers Active Effects available to the Core dnd5e System.
+
+### Legend
+
+`[number]`
+
+These square brakets mean "replace this with your value of the type within the brackets".
+
+So this example: `+[number]` would mean you input `+3`.
+
+`[formula]`
+When `formula` is mentioned in this document it means this value can be populated with any dice formula. For example, Bless adds several effects with the Effect Value of `1d4`.
+
+
+As part of this, an Actor's Rolldata is available as ["@attributes."](https://github.com/foundryvtt/dnd5e/wiki/Roll-Formulas) Useful examples:
+
+| @attribute                 | Description                   |
+| -------------------------- | ----------------------------- |
+| `@abilities.dex.mod`       | Actor's Dexterity Modifier    |
+| `@prof`                    | Actor's Proficiency Bonus     |
+| `@details.level`           | Actor's overall Level         |
+| `@classes.barbarian.levels` | Actor's Barbarian Class Level |
+
+> [!NOTE]
+> Note that when using formulas in an Active Effect Value, the actor sheet display that corresponds to the changed value will not display the evaluated formula, but it will be applied when rolled.
+> E.g. When adding `@abilities.cha.mod` to `system.bonuses.abilities.save` to simulate a Paladin's Aura of Protection, the actor sheet will not display that bonus applied to saving throws. The bonus will be present when the saving throw is rolled.
+
+To find out more about these, this post from Unsoluble in the [FoundryVTT Discord server](https://discord.gg/foundryvtt)'s #core-how-to channel can help:
+
+> To explore the data model in order to refer to inline properties like character levels, attributes, and other values, here are a few approaches:
+>
+> • Select a token, then open up the dev tools (F12 on Win; ⌥⌘I on Mac), and paste this into the Console:
+`console.log(canvas.tokens.controlled[0].actor.getRollData());`
+>
+> • Or: Install the "Autocomplete Inline Properties" module, to be able to just start typing in a supported field and have the available properties pop up (not all systems supported yet).
+>
+> • Or: Install this module (by pasting this manifest URL into the bottom of the Install Module window), enable it in your world, find and import its macro from the bundled Compendium, and run it on the selected token. `https://raw.githubusercontent.com/relick/FoundryVTT-Actor-Attribute-Lists/master/module.json`
+>
+> • Or: Right-click an actor in the sidebar and choose Export Data, which will get you a JSON file you can browse through. 
+
+
+## Commonly Desired Effect Examples
+
+### Abilities
+
+```
+system.abilities.[abbreviation].value
+                                bonuses.check
+                                        save
+```
+
+> [!NOTE]
+> <details>
+> <summary>Ability Abbreviations</summary>
+> 
+> | Ability      | Abbreviation |
+> | ------------ | ------------ |
+> | Strength     | `str`        |
+> | Dexterity    | `dex`        |
+> | Constitution | `con`        |
+> | Wisdom       | `wis`        |
+> | Intelligence | `int`        |
+> | Charisma     | `cha`        |
+> 
+> Source: `CONFIG.DND5E.abilities`
+>   
+> </details>
+
+
+#### Overriding an Ability Value
+E.g. an Item or potion that sets an ability score to a set value while in use
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.abilities.[abbreviation].value`     | Override     | `[number]`     |
+
+#### Bonus to a Specific Ability Save
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.abilities.[abbreviation].bonuses.save`     | Add     | `+[number/formula]`     |
+
+
+#### Bonus to a Specific Ability Check
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.abilities.[abbreviation].bonuses.check`     | Add     | `+[number/formula]`     |
+
+
+#### Bonus to All Ability Checks
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.bonuses.abilities.skill`     | Add     | `+[number/formula]`     |
+
+
+#### Bonus to All Ability Saves
+
+E.g. Paladin Aura of Protection
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.bonuses.abilities.save`     | Add     | `+[number/formula]`     |
+
+---
+
+### Skills
+```
+system.skills.[abbreviation].bonuses.check
+                                     passive
+```
+
+> [!NOTE]
+> <details>
+> <summary>Skill Abbreviations</summary>
+> 
+> | Skill           | Abbreviation |
+> | --------------- | ------------ |
+> | Acrobatics      | `acr`        |
+> | Animal Handling | `ani`        |
+> | Arcana          | `arc`        |
+> | Athletics       | `ath`        |
+> | Deception       | `dec`        |
+> | History         | `his`        |
+> | Insight         | `ins`        |
+> | Investigation   | `inv`        |
+> | Intimidation    | `itm`        |
+> | Medicine        | `med`        |
+> | Nature          | `nat`        |
+> | Persuasion      | `per`        |
+> | Perception      | `prc`        |
+> | Performance     | `prf`        |
+> | Religion        | `rel`        |
+> | Sleight of Hand | `slt`        |
+> | Stealth         | `ste`        |
+> | Survival        | `sur`        |
+> 
+> Source: `CONFIG.DND5E.skills`
+> </details>
+
+
+#### Bonus to a Specific Skill Check
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.skills.[abbreviation].bonuses.check`     | Add     | `+[number/formula]`     |
+
+#### Bonus to a Specific Skill Passive
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.skills.[abbreviation].bonuses.passive`     | Add     | `+[number/formula]`     |
+
+#### Bonus to All Skill Checks
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.bonuses.abilities.skill`     | Add     | `+[number/formula]`     |
+
+#### Bonus to Initiative
+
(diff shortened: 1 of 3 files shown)


krbz999:
This would be a good time to update the guide for scale values (especially Dice scale values).

MaxPat931:
Absolutely. 
Quick draft here:

### Scale Value

```
system.scale.[classId].[scaleId].value
                                 .number
                                 .die
                                 .faces
                                 .formula
```

> <details>
> <summary>Dice Scale Values</summary>
> 
> The Dice Scale Values have a few unique keys, here is an example of the result for these keys based on a scale value that is 3d8
>
> | Key | Value    |
> | ------------- | -------- |
> | number | 3 |
> | die | d8  |
> | faces | 8    |
> |formula | 3d8|
> </details>

#### Increase the value of a Scale Values
E.g. An Item or Spell which allows additional use(s) of a Class Feature.

| Attribute Key | Change Mode | Effect Value |
| -------- | -------- | -------- |
| system.scale.[classId].[scaleId].value`     | Add     | `[number]`     |

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,440 @@
+# FoundryVTT dnd5e Active Effects Examples
+
+![Up to date as of 2.1.x](https://img.shields.io/badge/dnd5e-v2.1.x-informational)
```
Fyorl, line 3: This seems old

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,440 @@
...
+
+See [Kandashi's Active Effects Guide](https://docs.google.com/document/d/1DuZaIFVq0YulDOvpahrfhZ6dK7LuclIRlGOtT0BIYEo) for a trove of useful information
+
+This document only covers Active Effects available to the Core dnd5e System.
+
+## Legend
+
+`[number]` - These square brakets mean "replace this with your value of the type within the brackets".
```
Fyorl, line 11:
```suggestion
`[number]` - These square brackets mean "replace this with your value of the type within the brackets".
```

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,440 @@
...
+So this example: `+[number]` would mean you input `+3`.
+
+`[formula]` - When `formula` is mentioned in this document it means this value can be populated with any dice formula. For example, Bless adds several effects with the Effect Value of `1d4`.
+
+
+As part of this, an Actor's Rolldata is available as ["@attributes."](https://github.com/foundryvtt/dnd5e/wiki/Roll-Formulas) Useful examples:
```
Fyorl, line 17: Pretty sure we do not support roll data as Active Effect values.

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,440 @@
...
+### Abilities
+
+```
+system.abilities.[abbreviation].value
+                                bonuses.check
+                                        save
+```
+
+
+> <details>
```
Fyorl, line 55: Does this need to be inside a `<blockquote>`? Seems like it would look fine outside it too. Don't feel strongly though.

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,440 @@
...
+> | Intelligence | `int`        |
+> | Charisma     | `cha`        |
+> 
+> Source: `CONFIG.DND5E.abilities`
+>   
+> </details>
+
+
+#### Overriding an Ability Value
+E.g. an Item or potion that sets an ability score to a set value while in use
```
Fyorl, line 73: An 'Upgrade' example would be nice here too. Like how Gauntlets of Ogre Power 'upgrade' Strength to 19, but will not lower it if the score is already higher.

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,440 @@
...
+
+
+#### Override the AC Calculation to a custom formula
+
+E.g. An Item or Spell which sets the Actor's AC to something like `12 + Int` for the duration.
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.attributes.ac.calc`     | Override     | `custom`     |
+| `system.attributes.ac.formula`     | Custom     | `10 + @abilities.str.mod`     |
```
Fyorl, line 237:
```suggestion
| `system.attributes.ac.formula`     | Custom     | `12 + @abilities.int.mod`     |
```

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,440 @@
...
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.scale.[classIdentifier].[scaleIdentifier].value`     | Add     | `+[number]`     |
+
+#### Increase the number of die of a Dice Scale Value
+E.g. An Item or Spell which increases the number of die in a Dice Scale Value (e.g. adds a die to a Rogue's Sneak Attack).
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.scale.[classIdentifier].[scaleIdentifier].number`     | Add     | `+[number]`     |
```
Fyorl, line 440: Maybe these examples should use the concrete barbarian rage and sneak attack values in their keys. I think such an example will help a lot with understanding what is meant by 'classIdentifier' and 'scaleIdentifier'.

Fyorl (changes requested):

wiki/Active-Effect-Guide.md
krbz999: In fields that are strings, we absolutely do. It varies from attribute to attribute. For example, `system.bonuses.<type>.attack`, (a string) but *not* `system.attributes.ac.bonus` (an integer).

wiki/Active-Effect-Guide.md
MaxPat931: Yea, this just came from Calego's original document, but 2.1 was when the access to the different parts of a Dice Scale Value was added, so I think that 2.1 represents the last time a change was made to the relevant data here, but I can bump it to 2.3 or 2.4 to represent the release of when the document was updated

wiki/Active-Effect-Guide.md
Fyorl: Hmm, I think we have to invoke a separate roll data pass manually on a case-by-case basis. It's not a general thing supported by the Active Effects system.

wiki/Active-Effect-Guide.md
Fyorl: Yeah I think since this information is up-to-date as of 2.3 it's fine to bump this up.

wiki/Active-Effect-Guide.md
MaxPat931: This also comes from Calego's original doc, where it was in a `:::info` block to have some unique styling, which renders as such on hack.md
image

I agree that it isn't really doing much for us here, but for some reason I do like it...

Fyorl: If you prefer it, then by all means keep it

wiki/Active-Effect-Guide.md
MaxPat931: Unless we want to show that specific example, I think adding a table of the Change modes and their function at the top of the document might be more useful.

MaxPat931:
Thoughts on adding an example and warning about HP attributes?

### Hit Points
```
system.attributes.hp.value
                     max
                     temp
                     tempmax
                     bonuses.level
                             overall
```

> [!warning]
> Never alter the `value`, `max`, or `temp` attributes with an active effect, as this **will** cause issues.

#### Temporary Bonus to the Maximum HP
E.g. An Item or Spell which temporarily increases a character's Max HP (e.g. Aid).

| Attribute Key | Change Mode | Effect Value |
| -------- | -------- | -------- |
| `system.attributes.hp.tempmax`     | Add     | `+[number]`     |

#### Bonus to the Maximum HP
E.g. An Item or Feature which increases a character's Max HP.

| Attribute Key | Change Mode | Effect Value |
| -------- | -------- | -------- |
| `system.attributes.hp.bonuses.overall`     | Add     | `+[number/formula]`     |

#### Bonus HP for each Character Level
E.g. An effect that provides a bonus to the HP a character gains for each Level they acquire (e.g. Tough).

| Attribute Key | Change Mode | Effect Value |
| -------- | -------- | -------- |
| `system.attributes.hp.bonuses.level`     | Add     | `+[number/formula]`     |

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,442 @@
...
+| `@details.level`           | Actor's overall Level         |
+| `@classes.barbarian.levels` | Actor's Barbarian Class Level |
+
+> [!Note]
+> When using formulas in an Active Effect Value, the actor sheet display that corresponds to the changed value will not display the evaluated formula, but it will be applied when rolled.
+> E.g. When adding `+@abilities.cha.mod` to `system.bonuses.abilities.save` to simulate a Paladin's Aura of Protection, the actor sheet will not display that bonus applied to saving throws. The bonus will be present when the saving throw is rolled.
+
+| Change Mode | Description |
+|------------ | ------------|
+| Add         | Adds the provided value to a number. This can be used to both add and subtract from a particular value by specifying `+1` or `-1` as the value to add. |
```
Fyorl, line 29: This works on 'array' or 'set' values. We even have some examples in this wiki like adding a condition type. I'm not sure how best to explain that though as we shouldn't presume the average user understands what an 'array' or 'set' is.

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,442 @@
...
+This document only covers Active Effects available to the Core dnd5e System.
+
+## Legend
+
+`[number]` - These square brackets mean "replace this with your value of the type within the brackets".
+So this example: `+[number]` would mean you input `+3`.
+
+`[formula]` - When `formula` is mentioned in this document it means this value can be populated with any dice formula. For example, Bless adds several effects with the Effect Value of `1d4`.
+
+As part of this, an [Actor's Rolldata](https://github.com/foundryvtt/dnd5e/wiki/Roll-Formulas) is available as "@attributes." Useful examples:
```
Fyorl, line 14: We probably need another note here then, according to [this](https://github.com/foundryvtt/dnd5e/pull/2482#discussion_r1365998926) discussion, warning that supplying rolldata is not supported for all active effects, and it's kind of arbitrary which ones it is supported for. Not sure how best to explain that either.

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,442 @@
...
+> | Wisdom       | `wis`        |
+> | Intelligence | `int`        |
+> | Charisma     | `cha`        |
+> 
+> Source: `CONFIG.DND5E.abilities`
+>   
+> </details>
+
+
+#### Overriding an Ability Value
```
Fyorl, line 67: It occurs to me that we might want to try to stick to game terms where possible as that is most likely what people are thinking of when trying to translate game rules to active effects. I'm just going to go through and flag these to hopefully make it easier if you agree.

```suggestion
#### Overriding an Ability Score
```

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,442 @@
...
+E.g. an Item or potion that sets an ability score to a set value while in use
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.abilities.[abbreviation].value`     | Override     | `[number]`     |
+
+#### Upgrading an Ability Value
```
Fyorl, line 74:
```suggestion
#### Upgrading an Ability Score
```

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,442 @@
...
+E.g. an Item or potion that sets an ability score to a set value, if the value does not already exceed that value, such as the Gauntlets of Ogre Power
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.abilities.[abbreviation].value`     | Upgrade     | `[number]`     |
+
+#### Bonus to a Specific Ability Save
```
Fyorl, line 81:
```suggestion
#### Bonus to a Specific Saving Throw
```

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,442 @@
...
+
+
+#### Bonus to All Ability Checks
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.bonuses.abilities.skill`     | Add     | `+[number/formula]`     |
+
+
+#### Bonus to All Ability Saves
```
Fyorl, line 102:
```suggestion
#### Bonus to All Saving Throws
```

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,442 @@
...
+| -------- | -------- | -------- |
+| `system.bonuses.abilities.skill`     | Add     | `+[number/formula]`     |
+
+#### Bonus to Initiative
+
+Initiative is not quite a skill but behaves similarly.
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.attributes.init.bonus`     | Add     | `+[number/formula]`     |
```
Fyorl, line 170: We should consider moving this right under 'bonus to a specific ability check' since Initiative is just a special kind of ability check and not really related to skills (aside from skills also being ability checks). We can then avoid the wording about 'not quite a skill'.

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,442 @@
...
+system.attributes.ac.bonus
+                     formula
+                     calc
+                     cover
+                     flat
+```
+
+#### Add a Bonus to AC
+
+E.g. An Item or Spell which adds to the Actor's current AC to something for the duration.
```
Fyorl, line 224:
```suggestion
E.g. An Item or Spell which adds something to the Actor's current AC for the duration.
```

Fyorl (changes requested):

wiki/Active-Effect-Guide.md
MaxPat931: I think that anywhere a die formula is supported, a rolldata path will be too?

krbz999: > I think that anywhere a die formula is supported, a rolldata path will be too?

If it supports dice, then it's a string that is evaluated when needed (at least fully, some parts are sometimes evaluated immediately for sheets), and it supports roll data.

The other direction is not always true, some strings support roll data (for example `attributes.hp.bonuses`), but not dice.

Fyorl (approved):
Happy for this to be merged. It seems no-one has a good idea for how to explain the cases where adding roll data will work and when it won't so we can leave it as-is and will just have to field questions from confused users in the discord.

wiki/Active-Effect-Guide.md
```diff
@@ -0,0 +1,475 @@
...
+
+
+### Override the AC Calculation to a custom formula
+
+E.g. An Item or Spell which sets the Actor's AC to `12 + Int` for the duration.
+
+| Attribute Key | Change Mode | Effect Value |
+| -------- | -------- | -------- |
+| `system.attributes.ac.calc`     | Override     | `custom`     |
+| `system.attributes.ac.formula`     | Custom     | `12 + @abilities.int.mod`     |
```
krbz999, line 235: The change mode here should be 'Override'.

krbz999 (changes requested):

PR #2492, opened 2023-10-14
[#2463] Add configuration for whether to default to true/false for limited uses and templates in AbilityUseDialog

### LINKED ISSUE #2463: Add a toggle in the item sheet to 'ignore' limited uses and measured templates when using the item
Some items have limited uses that are meant only to serve as a resource for other items on the actor, such as magic weapons that have charges for their spells. It would be good if these did not prompt for usage every time the weapon was used.

Likewise, a few spells have the ability to pick a target area but do not do so when cast (such as Dragon's Breath or the Investiture spells), so it would also be good to not always force a prompt under these circumstances.

The item sheet could be improved to add two toggles (when applicable) in the same section as the 'Destroy on Empty' for consumable items.

krbz999:
Closes #2463


templates/items/parts/item-activation.hbs
```diff
@@ -160,4 +160,23 @@
         </select>
     </div>
 </div>
+
+{{!-- Prompt Configuration --}}
+<div class="form-group">
+    {{#if (eq item.type "consumable")}}
+    <label class="checkbox">
+        <input type="checkbox" name="system.uses.autoDestroy" {{checked system.uses.autoDestroy}}> {{ localize "DND5E.ItemDestroyEmpty" }}
```
Fyorl, line 168: Suggestion for a tooltip: "If this item has limited uses, it will be destroyed when those uses reaches 0. Otherwise it will be destroyed when used."

templates/items/parts/item-activation.hbs
```diff
@@ -160,4 +160,23 @@
...
+    </label>
+    {{/if}}
+    {{#if item.hasAreaTarget}}
+    <label class="checkbox">
+        <input type="checkbox" name="system.target.prompt" {{checked system.target.prompt}}> {{ localize "DND5E.TemplatePrompt" }}
```
Fyorl, line 173: Suggestion for tooltip: "If unchecked, the prompt for placing a Measured Template will be suppressed."

templates/items/parts/item-activation.hbs
```diff
@@ -160,4 +160,23 @@
...
+    </label>
+    {{/if}}
+    {{#if system.uses.per}}
+    <label class="checkbox">
+        <input type="checkbox" name="system.uses.prompt" {{checked system.uses.prompt}}> {{ localize "DND5E.LimitedUsesPrompt" }}
```
Fyorl, line 178: Suggestion for tooltip: "If unchecked, the prompt to consume an available use will be suppressed."

Fyorl (changes requested):

Fyorl (approved):

PR #2493, opened 2023-10-15
Add Coin support in _simplifyDiceTerms

### LINKED ISSUE #2281: '1dc' roll terms erroneously mapped to '1d2' in item derivedDamage labels
This is purely a cosmetic issue and doesn't affect rolling.  If you add one or more '1dc' terms in the damage formula for an item, the 'derivedDamage' labels shown in the description tab of that item will instead contain '1d2' (or 'nd2' for n terms).  This is incorrect as '1dc' is equivalent to '1d2 - 1' and not '1d2' despite having two faces.

Anyway, this happens with zero enabled modules, v. 2.1.5 of the dnd5e system, and v. 10.291 of Foundry.

1dc
1d2
labels

Hoppyhob:
Closes #2281
There was an issue in _simplifyDiceTerms where a Coin was reduced to its operator and number of faces, then assumed to be a die when passed back out. This ended up turning a coin into a die 2 and combining it with any other die 2s in the Dice Terms.
This was fixed by using a c for the faces value if the DiceTerm is Coin and returning a DiceTerm of Coin, instead of Die, if the face value is c.
(Any DiceTerm other than Coin should default to using the number of faces and return as a Die DiceTerm)

### DIFF
--- a/module/dice/simplify-roll-formula.mjs
+++ b/module/dice/simplify-roll-formula.mjs
@@ -107,7 +107,7 @@ function _simplifyDiceTerms(terms) {
   // Split the unannotated terms into different die sizes and signs
   const diceQuantities = unannotated.reduce((obj, curr, i) => {
     if ( curr instanceof OperatorTerm ) return obj;
-    let face = curr.constructor?.name === "Coin" ? "c":curr.faces;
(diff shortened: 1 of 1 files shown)


Fyorl (approved):
Nice fix, thanks

PR #2494, opened 2023-10-15
Add NaN logic to ASI flow

### LINKED ISSUE #2441: Ability scores that are at their maximum can be edited in an ASI
Dunno if we want to prevent editing them, but given that the "-" and "+" buttons are disabled I assume so.

Additionally, if the value is deleted, the entire advancement flow needs to be redone as the number becomes `NaN`.

One fix is to just give it the `readonly` property, same as if it had a `fixed` addition.

Hoppyhob:
Closes #2441
- Added NaN Check to _onClickButton to reset assignments value to 0 if it is NaN or undefined.
- Added NaN check to getData() to return the assignment to 0 if NaN.
- Added check to set ability as fixed if ability's initial value matched the ability's max value.

### DIFF
--- a/module/applications/advancement/ability-score-improvement-flow.mjs
+++ b/module/applications/advancement/ability-score-improvement-flow.mjs
@@ -69,7 +69,7 @@ export default class AbilityScoreImprovementFlow extends AdvancementFlow {
         delta: (value - ability.value) ? formatter.format(value - ability.value) : null,
         showDelta: true,
         isDisabled: !!this.feat,
-        isFixed: !!fixed || ability.value >= ability.max,
+        isFixed: !!fixed || (ability.value >= ability.max),
         canIncrease: (value < max) && !fixed && !this.feat,
         canDecrease: (value > ability.value) && !fixed && !this.feat
       };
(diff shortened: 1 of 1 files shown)


module/applications/advancement/ability-score-improvement-flow.mjs
```diff
@@ -57,7 +60,7 @@ export default class AbilityScoreImprovementFlow extends AdvancementFlow {
     const abilities = Object.entries(CONFIG.DND5E.abilities).reduce((obj, [key, data]) => {
       if ( !this.advancement.canImprove(key) ) return obj;
       const ability = this.advancement.actor.system.abilities[key];
-      const fixed = this.advancement.configuration.fixed[key] ?? 0;
+      const fixed = (this.advancement.configuration.fixed[key] || ability.max === ability.value) ?? 0;
```
Fyorl, line 63: The result of this expression will sometimes be a boolean, but `fixed` must always be a number because it's used in an arithmetic expression on the following line.

I think it would be better if we change `isFixed` on line 72 (75 in this PR) to:
```js
isFixed: !!fixed || (ability.value >= ability.max)
```

Fyorl (changes requested):
Thanks for doing this work. In addition to the comment I left, I feel it would be preferable to prevent the value from ever becoming `NaN` at all rather than spreading `NaN` checks around our code. I believe the `NaN` value is originating in `_onChangeInput` where we call `input.valueAsNumber`. We should instead do something like this:

```js
const value = isNaN(input.valueAsNumber) ? 0 : input.valueAsNumber;
const clampedValue = Math.clamped(value, Number(input.min), Number(input.max));
```

Hoppyhob:
Added the requested changes and removed the NaN checks that are now redundant.
For the _onChangeInput I added a NaN check that directly sets the assignments[key] to zero. (Instead of setting it to zero then clamping it to the minimum then finding the difference) Otherwise if not NaN it will go through the original logic.

Fyorl (approved):
Nice, thanks, looks good to me

PR #2498, opened 2023-10-19
[#1405] Data Structure for TraitAdvancement

arbron:
Implements configuration changes and the data structure necessary for the TraitAdvancement.


Fyorl (approved):
Looks good to me, thanks Jeff

PR #2499, opened 2023-10-19
[#1405] Modify trait utilities to support prefixed trait keys & add SelectChoices

arbron:
- Modifies a number of utility methods in trait utilities to support prefixed trait keys (`tool:art:potter` instead of just `potter`)
- Introduces the `SelectChoices` type which offers a number of methods for filtering, sorting, cloning, etc. nested sets of options like those represented by tool categories


module/documents/actor/trait.mjs
```diff
@@ -232,24 +369,73 @@ export function keyLabel(trait, key) {
...
-  if ( !choice.pool ) {
-    return game.i18n.format("DND5E.TraitConfigChooseAny", {
-      count: choice.count,
-      type: traitLabel(trait, choice.count).toLowerCase()
+export function choiceLabel(choice, { only=false, final=false }={}) {
+  if ( !choice.pool.size ) return "";
+
+  // Single entry in pool
+  // { count: 3, pool: ["skills:*"] } -> any three skills
+  // { count: 3, pool: ["skills:*"] } (final) -> three other skills
```
Fyorl, line 384: I might need a bit of elaboration of how the `final` option is supposed to work. Is it only for the purposes of formatting the list for display?

module/documents/actor/trait.mjs
```diff
@@ -232,24 +369,73 @@ export function keyLabel(trait, key) {
...
 }
+
+/* -------------------------------------------- */
+
+/**
+ * Create a human readable description of trait grants & choices.
+ * @param {Set<string>} grants                       Guaranteed trait grants.
+ * @param {TraitChoice[]} [choices=[]]               Trait choices.
+ * @param {object} [options={}]
+ * @param {string} [options.choiceMode="inclusive"]  Choice mode.
```
Fyorl, line 417: If there are only a handful of options we can annotate it like this:

```suggestion
 * @param {"inclusive"|"exclusive"} [options.choiceMode="inclusive"]  Choice mode.
```

module/documents/actor/trait.mjs
```diff
@@ -232,24 +369,73 @@ export function keyLabel(trait, key) {
...
+
+  let sections = Array.from(grants).map(g => keyLabel(g));
+  if ( choiceMode === "inclusive" ) {
+    sections = sections.concat(choiceSections);
+  } else {
+    const choiceListFormatter = new Intl.ListFormat(game.i18n.lang, { style: "long", type: "disjunction" });
+    sections.push(choiceListFormatter.format(choiceSections));
+  }
+
+  const listFormatter = new Intl.ListFormat(game.i18n.lang, { style: "long", type: "conjunction" });
```
Fyorl, line 436: We can use `game.i18n.getListFormatter` which will lazily cache these formatters though it's not really a big deal.

Fyorl (changes requested):
I can't help but feel like we perhaps wouldn't need to do quite so much work in `Trait` if we had more consistent and predictable structuring in our `CONFIG` as well as for where traits are stored on the actor. I'm not sure there's enough of a tangible benefit to changing it though at this stage given how much churn and breakage it would cause.

It seems that we only really ever have one level of `children` in our current trait configuration. I wonder, then, if having a fully-recursive tree structure and associated operations is necessary, and if there are maybe some simplifications that can be won from recognising this property. It might be that the general tree solution is the simplest though.

In general though I think it's fine that this complexity is tucked away in these utilities. It's an acceptable place to be currently.

If I could ask though for some small examples as to some of the inputs and outputs for the `SelectChoices` class, particularly in regard to how the wildcards work. That would help me a lot with contextualising some of this work. Since it's all sets of strings I'm hoping they won't be too arduous to produce. We could also include them as `@example` annotations.

Some examples for some of the trait methods would also help me, like you have in `choiceLabel`, they were very useful.

module/documents/actor/trait.mjs
arbron: Yes, it is to format lists properly when there is a mixture to grants & choices like `"Athletics and three other skills"`, which reads better than `"Athletics and any three skills"`.

module/documents/actor/select-choices.mjs
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Merge another SelectOptions object into this one.
+   * @param {SelectOptions} other
+   * @returns {SelectOptions}
+   */
+  merge(other) {
```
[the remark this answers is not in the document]
arbron: I can make that change.

This pattern was replicating the syntax of Swift which has a very clear naming to distinguish mutating methods and non-mutating methods unlike the mess of Javascript :).

module/documents/actor/trait.mjs
```diff
@@ -17,56 +19,92 @@ const _cachedIndices = {};
...
 }
 
 /* -------------------------------------------- */
 
 /**
- * Fetch the categories object for the specified trait.
- * @param {string} trait  Trait as defined in `CONFIG.DND5E.traits`.
- * @returns {object}      Trait categories defined within `CONFIG.DND5E`.
+ * Get the current trait values for the provided actor.
+ * @param {BlackFlagActor} actor  Actor from which to retrieve the values.
```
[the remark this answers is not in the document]
arbron: Whoops

module/documents/actor/trait.mjs
```diff
@@ -190,39 +293,73 @@ export function traitIndexFields() {
...
+ * object is provided in config.
+ * @param {string} key              Key for which to generate the label.
+ * @param {object} [config={}]
+ * @param {number} [config.count]   Number to display, only if a wildcard is used as final part of key.
+ * @param {string} [config.trait]   Trait as defined in `CONFIG.DND5E.traits` if not using a prefixed key.
+ * @param {boolean} [config.final]  Is this the final in a list?
+ * @returns {string}                Retrieved label.
  */
-export function keyLabel(trait, key) {
+export function keyLabel(key, { count, trait, final }={}) {
```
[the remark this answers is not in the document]
arbron: I think the old way can be handled pretty easily with a deprecation warning in case anyone was using it.

module/documents/actor/trait.mjs
arbron: Hey, new method I didn't know existed, great!

arbron:
> I can't help but feel like we perhaps wouldn't need to do quite so much work in `Trait` if we had more consistent and predictable structuring in our `CONFIG` as well as for where traits are stored on the actor. I'm not sure there's enough of a tangible benefit to changing it though at this stage given how much churn and breakage it would cause.

While I think it would be okay to switch things on the actor side, any major change of the config would be a breaking change for any 3rd party content that adds new types.

> It seems that we only really ever have one level of `children` in our current trait configuration. I wonder, then, if having a fully-recursive tree structure and associated operations is necessary, and if there are maybe some simplifications that can be won from recognising this property. It might be that the general tree solution is the simplest though.

We do have the chance of multiple levels when more than one trait type is represented, which mainly comes into play with expertise (e.g. you get a choice between gaining expertise in a any skill or thieves tools).

> If I could ask though for some small examples as to some of the inputs and outputs for the `SelectChoices` class, particularly in regard to how the wildcards work. That would help me a lot with contextualising some of this work. Since it's all sets of strings I'm hoping they won't be too arduous to produce. We could also include them as `@example` annotations.

Added some examples for `filter` and `exclude`.

> Some examples for some of the trait methods would also help me, like you have in `choiceLabel`, they were very useful.

Add a bunch and moved the examples into the JSDocs.

module/documents/actor/trait.mjs
```diff
@@ -84,29 +133,93 @@ export async function choices(trait, chosen=new Set()) {
...
+      if ( prefixed && any ) {
+        const anyKey = `${key}:*`;
+        children[anyKey] = {
+          label: keyLabel(anyKey).titleCase(),
+          chosen: chosen.has(anyKey), sorting: false, wildcard: true
+        };
+      }
+      Object.entries(data.children).forEach(([k, v]) => prepareCategory(k, v, children, key));
+    }
+  };
```
Fyorl, line 196: This is perhaps a silly suggestion, but could this be simplified somewhat by delegating the prefixing to the `SelectChoices` constructor (maybe with an option)? Things started falling into place a lot better once I realised that the children are already prefixed when supplied to the `SelectChoices` constructor. I thought we should maybe document that then, but then I thought we might be able to actually do the prefixing in the constructor, and then it would be obvious.

Maybe that's not feasible, I'm not sure, but maybe you could try it out. Otherwise, ultimately, this is just utility code so it's fine as-is as well with just a small mention in the constructor JSDoc that children are supplied already-prefixed.

module/documents/actor/trait.mjs
```diff
@@ -190,39 +303,113 @@ export function traitIndexFields() {
...
+ * // Returns "any Artisan's Tools"
+ * keyLabel("tool:art:*");
+ *
+ * @example
+ * // Returns "any 2 Artisan's Tools"
+ * keyLabel("tool:art:*", { count: 2 });
+ *
+ * @example
+ * // Returns "2 other Artisan's Tools"
+ * keyLabel("tool:art:*", { count: 2, final: true });
```
Fyorl, line 342: These are super helpful, thank you.

module/documents/actor/trait.mjs
```diff
@@ -232,24 +419,109 @@ export function keyLabel(trait, key) {
...
+ * @example
+ * // Returns "Thieves Tools or any skill"
+ * choiceLabel({ count: 1, pool: new Set(["tool:thief", "skills:*"]) }, { only: true });
+ *
+ * @example
+ * // Returns "Thieves' Tools or any artisan tool"
+ * choiceLabel({ count: 1, pool: new Set(["tool:thief", "tool:art:*"]) }, { only: true });
+ *
+ * @example
+ * // Returns
```
Fyorl, line 450: Don't keep me in suspense like this.

module/documents/actor/trait.mjs
```diff
@@ -232,24 +419,109 @@ export function keyLabel(trait, key) {
...
+ * // Returns "Choose any skill proficiency"
+ * localizedList(new Set(), [{ count: 1, pool: new Set(["skills:*"])}]);
+ *
+ * @example
+ * // Returns "Choose any 2 languages or any 1 skill proficiency"
+ * localizedList(new Set(), [
+ *   {count: 2, pool: new Set(["languages:*"])}, { count: 1, pool: new Set(["skills:*"])}
+ * ], {choiceMode: "exclusive"});
+ */
+export function localizedList(grants, choices=[], { choiceMode="inclusive" }={}) {
```
Fyorl, line 507: Since these seem all kind of optional from the examples, maybe it should just be one big options object:

```suggestion
export function localizedList({ grants=new Set(), choices=[], choiceMode="inclusive" }={}) {
```

module/documents/actor/select-choices.mjs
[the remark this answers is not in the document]
Fyorl: I guess I can kind of see how the tense of the verb might imply mutation if I think about it for a bit, and if the convention is in place already. That's quite interesting to know about Swift. An `inplace` option is definitely more in-line with the core API conventions though, so thanks for making the change.

Fyorl (approved):
Looks pretty good to me, thank you. I've left some minor linting comments and some suggestions for consideration, that you can take or leave.

module/documents/actor/trait.mjs
arbron: I think the issue is that generating the `chosen` value for each choice requires the fully prefixed key, so if we left prefixing to the `SelectChoices` object we'd probably have to iterate through the entire structure again to set `chosen` or generate the keys twice.

Fyorl: Ah ok, I see. Can we just document the `SelectChoicesEntry` typedef then to mention that the `children` are required to be already-prefixed? I suppose we could add an option to perform the prefixing in `SelectChoices` if we ever needed it. I guess this method is the only place we instantiate a `SelectChoices` object though currently.

Fyorl (approved):
Just one comment above about adding documentation then feel free to merge.

module/documents/actor/trait.mjs
arbron: I'll add a note about providing prefixed keys if desired, though we do also use this method with un-prefixed keys in `TraitSelector` so it isn't a requirement.

Issue #1865, opened 2022-10-11
Add custom enrichers for rolling checks & saves


arbron:
Currently people are attempting to implement skill check buttons in descriptions using the format `1d20 + @skills.stl.total`, but this doesn't take into account all of the potential bonuses that apply nor does it display the roll dialog to allow selection of Advantage. Adding custom enrichers to the various system rolls would allow us to handle these inline rolls properly.

- ~`@Check{dex}`~ `@Check[dex]`: Roll a flat Dexterity check, displays as `Dexterity Check`
- ~`@Check{dex.acr}`~ `@Check[dex.acr]`: Roll an Acrobatics check using dexterity, displays as `Dexterity (Acrobatics) Check`
- ~`@Check{dex.acr|Leap}`~ `@Check[dex.acr]{Leap}`: Roll an Acrobatics check using dexterity, displays as `Leap`
- `@Check[dex.acr|15]`: Roll an Acrobatics DC 15 check using dexterity, displays as `DC 15 Dexterity (Acrobatics) Check`
- `@Check[dex.acr|dc:15]`: Explicitly adding the DC prefix would also be supported
- ~`@Save{dex}`~ `@Save[dex]`: Roll a Dexterity save, displays as `Dexterity Save`

Potentially also include a ~`@Skill{acr}`~ `@Skill[acr]` version that just uses the ability associated with a particular skill on the selected actor sheet.

### Alternative Spellings

- `@RollCheck`, `@RollSave`
- `@AbilityCheck`, `@AbilitySave`, `@SkillCheck`

MaxPat931:
a `@RollItem{Bite}` which essentially calls `game.dnd5e.documents.macro.rollItem("Bite")` would also be useful for the Multiattack example I mentioned in #1546 which I've tinkered around and implemented with @kaelad02 's Inline Roll Commands module

arbron:
It would also be useful to be able to reference certain rules without pointing to a specific journal:

- `@Reference{skill:acr}`: Opens the journal entry for actobatics
- `@Reference{condition:prone}` Opens the journal entry for prone

kaelad02:
Like MaxPat mentioned, I made a module that does this that includes skills, ability checks, saves, and even items.

https://github.com/kaelad02/inline-roll-cmd

[...]

PR #2500, opened 2023-10-19
[#1865] Add dnd5e-specific rolling enrichers

arbron:
Screenshot 2023-10-19 at 14 21 26


module/enrichers.mjs
```diff
@@ -0,0 +1,195 @@
...
+ * @returns {Promise|void}
+ */
+export function rollAction(event) {
+  let target = event.target;
+  if ( !target.classList.contains("roll-link") ) target = target.closest(".roll-link");
+  if ( !target ) return;
+  event.stopPropagation();
+
+  const { type, ability, skill, dc } = target.dataset;
+  const options = {};
```
krbz999, line 174:
```suggestion
  const options = { event: event };
```

Fyorl:
> Seems a bit counter intuitive to me in the case of the 2nd group of examples to use `check` for what is actually skills.

Skills are ability checks.

module/enrichers.mjs
```diff
@@ -0,0 +1,195 @@
...
+/* -------------------------------------------- */
+
+/**
+ * Perform the provided roll action.
+ * @param {Event} event  The click event triggering the action.
+ * @returns {Promise|void}
+ */
+export function rollAction(event) {
+  let target = event.target;
+  if ( !target.classList.contains("roll-link") ) target = target.closest(".roll-link");
```
Fyorl, line 169: I think this is unnecessary because `Element#closest` can return itself if it matches the provided selector.
```suggestion
  const target = event.target.closest(".roll-link");
```

module/enrichers.mjs
Fyorl: Maybe just:
```js
const options = { event };
```

module/enrichers.mjs
```diff
@@ -0,0 +1,195 @@
...
+  if ( !target ) return;
+  event.stopPropagation();
+
+  const { type, ability, skill, dc } = target.dataset;
+  const options = {};
+  if ( dc ) options.targetValue = dc;
+
+  // Fetch the actor that should perform the roll
+  let actor;
+  const speaker = ChatMessage.getSpeaker();
```
Fyorl, line 179:
```suggestion
  const speaker = ChatMessage.implementation.getSpeaker();
```

module/enrichers.mjs
```diff
@@ -0,0 +1,195 @@
...
+  if ( speaker.token ) actor = game.actors.tokens[speaker.token];
+  actor ??= game.actors.get(speaker.actor);
+  if ( !actor ) return ui.notifications.warn(game.i18n.localize("EDITOR.DND5E.Inline.NoActorWarning"));
```
Fyorl, line 182: `Notifications#warn` returns a `number` (in v11 at least).

```suggestion
  if ( !actor ) {
    ui.notifications.warn(game.i18n.localize("EDITOR.DND5E.Inline.NoActorWarning"));
    return;
  }
```

module/enrichers.mjs
```diff
@@ -0,0 +1,195 @@
...
+
+  switch (type) {
```
Fyorl, line 184:
```suggestion
  switch ( type ) {
```

Fyorl (changes requested):
This looks good to me, thanks Jeff, only minor suggestions on the code side, though I had some more high-level thoughts:

1. Spellcasting ability checks are fairly frequently called for. Something like `[[/check spell 20]]` -> `DC 20 Spellcasting ability check` would be nice to support. Easy enough to add parsing support for but I'm unsure if it's a lot of work on the execution side, so we can save it for follow-up work if necessary.
2. I can imagine maybe wanting some dynamic calculations for the DC. I think that rapidly becomes very complicated in terms of parsing and evaluating though, so I'm very much not suggesting we do it as part of this work.
3. On Zhell's suggestion of supporting `[[/skill per str 20]]`, it's pretty much the equivalent of `[[/check]]` but with the arguments the other way around. In which case I might suggest we just have `[[/skill]]` be an alias of `[[/check]]` and allow the arguments to be in any order. It would cause some issues if there were ever an overlap of ability keys with skill keys, but it makes the API more ergonomic as you don't need to remember the order of arguments. We remove `enrichSkill`, and `enrichCheck` becomes:

```js
let ability;
let skill;
let dc;
while ( config.length ) {
  const arg = config.shift();
  if ( arg in CONFIG.DND5E.abilities ) ability = arg;
  else if ( arg in CONFIG.DND5E.skills ) skill = arg;
  else if ( Number.isNumeric(arg) ) dc = Number(arg);
} 
```

4. With the above changes it also becomes quite easy to support tool checks since the parsing could be handled in the same way but checking against `CONFIG.DND5E.toolIds` instead. The execution path would need to be a little different though.

arbron:
Okay, I've merged check and skill so you can use those terms interchangeably and the terms can be in any order. I've also added the ability to use formulas for DCs, as long as they don't include spaces:

Screenshot 2023-10-20 at 09 33 25

Fyorl (approved):
Thanks Jeff, this looks good to go, I'm just going to kick it over to Atro for final sign off here. In the meantime, did you have any thoughts on the tool checks suggestion? You might have missed it since I edited it into this comment https://github.com/foundryvtt/dnd5e/pull/2500#pullrequestreview-1690055725 later. It feels like it would be reasonably easy to add support for it here in `enrichCheck`, maybe with another `/tool` alias. It can be follow-up work though.

arbron:
> Thanks Jeff, this looks good to go, I'm just going to kick it over to Atro for final sign off here. In the meantime, did you have any thoughts on the tool checks suggestion? You might have missed it since I edited it into this comment [#2500 (review)](https://github.com/foundryvtt/dnd5e/pull/2500#pullrequestreview-1690055725) later. It feels like it would be reasonably easy to add support for it here in `enrichCheck`, maybe with another `/tool` alias. It can be follow-up work though.

Didn't catch that before, I'll work on that for a follow up PR.

arbron:
Updated based on feedback:

Screenshot 2023-10-23 at 08 33 02

Fyorl (changes requested):
This is shaping up nicely, thanks Jeff. I've left some additional suggestions around argument parsing that you can consider. Also, two other pieces from discord discussion:

(1) Support for a 'format' option, i.e. `format=long` or `format=short` which determines whether the full 'check' or 'saving throw' is appended and part of the clickable link. It seems Atro's preference would be for the default to be `format=short` (which is what this PR currently defaults to), so we should continue with that so long as there's no objection.

(2) There was some suggestion to allow for the full skill/ability name to be used since the abbreviations can be hard to remember (tools probably don't need this). So we would support, e.g. `[[/check dexterity acrobatics]]`. There are some open questions as to where we put these keys (on `CONFIG`? just in `enrichers.mjs`? exported?) and localisation issues. If the user changes their language, the check should not suddenly stop being enriched, for example.

My suggestion, then, since the relevant `CONFIG` objects are already full objects, is that we just append something like `fullKey` as a hard-coded English key (much like tools already have) that we can reference here:

```js
DND5E.skills = {
  acr: { label: "DND5E.SkillAcr", ability: "dex", fullKey: "acrobatics" }
};
```

And then we perform some static initialisation in `enrichers.mjs` at the top of the file:

```js
import DND5E from "./config.mjs";

const fullSkills = Object.fromEntries(Object.values(DND5E.skills).map(skl => [skl.fullKey, skl]));
```

Fyorl:
An addendum to the above that I forgot that was also discussed on discord: If we add boolean parameters that will allow us to add the `passive` option that converts a check to a passive description (this would not need to roll anything, it would just be for display/localisation purposes).

module/config.mjs
```diff
@@ -112,31 +121,32 @@ DND5E.hitPointsAbility = "con";
...
+  inv: { label: "DND5E.SkillInv", ability: "int", fullKey: "investigation" },
+  med: { label: "DND5E.SkillMed", ability: "wis", fullKey: "medicine" },
+  nat: { label: "DND5E.SkillNat", ability: "int", fullKey: "nature" },
+  prc: { label: "DND5E.SkillPrc", ability: "wis", fullKey: "perception" },
+  prf: { label: "DND5E.SkillPrf", ability: "cha", fullKey: "performance" },
+  per: { label: "DND5E.SkillPer", ability: "cha", fullKey: "persuasion" },
+  rel: { label: "DND5E.SkillRel", ability: "int", fullKey: "religion" },
+  slt: { label: "DND5E.SkillSlt", ability: "dex", fullKey: "slightOfHand" },
+  ste: { label: "DND5E.SkillSte", ability: "dex", fullKey: "stealth" },
+  sur: { label: "DND5E.SkillSur", ability: "wis", fullKey: "survival" }
```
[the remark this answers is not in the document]
arbron: These keys will not be replaced by localization, otherwise existing enrichers that point to them would break as soon as you enable a new localization.

arbron:
> An addendum to the above that I forgot that was also discussed on discord: If we add boolean parameters that will allow us to add the `passive` option that converts a check to a passive description (this would not need to roll anything, it would just be for display/localisation purposes).

Screenshot 2023-10-23 at 13 52 08

Ikaguia:
This looks real nice, would it be possible to add an additional parameter to force a specific roll mode when clicking that tag?
I.E: `[[/check ability=wis skill=per rollmode=blind]]` or `[[/check ability=int skill=arc rollmode=blind]]` would allow the dm to easily ask for checks that the players don't know the result. (if no `rollmode` was set, it would default to using the user's default roll mode.

Fyorl:
> This looks real nice, would it be possible to add an additional parameter to force a specific roll mode when clicking that tag? I.E: `[[/check ability=wis skill=per rollmode=blind]]` or `[[/check ability=int skill=arc rollmode=blind]]` would allow the dm to easily ask for checks that the players don't know the result. (if no `rollmode` was set, it would default to using the user's default roll mode.

I'm not sure about allowing authors to force a particular roll mode. Different DMs have different preferences around 'rolling in the open' or not, and I imagine they would find it quite annoying if they are forced to do it one way or the other because whoever wrote the content forced an explicit roll mode.

module/config.mjs
```diff
@@ -112,31 +121,32 @@ DND5E.hitPointsAbility = "con";
...
+  ins: { label: "DND5E.SkillIns", ability: "wis", fullKey: "insight" },
+  itm: { label: "DND5E.SkillItm", ability: "cha", fullKey: "intimidation" },
+  inv: { label: "DND5E.SkillInv", ability: "int", fullKey: "investigation" },
+  med: { label: "DND5E.SkillMed", ability: "wis", fullKey: "medicine" },
+  nat: { label: "DND5E.SkillNat", ability: "int", fullKey: "nature" },
+  prc: { label: "DND5E.SkillPrc", ability: "wis", fullKey: "perception" },
+  prf: { label: "DND5E.SkillPrf", ability: "cha", fullKey: "performance" },
+  per: { label: "DND5E.SkillPer", ability: "cha", fullKey: "persuasion" },
+  rel: { label: "DND5E.SkillRel", ability: "int", fullKey: "religion" },
+  slt: { label: "DND5E.SkillSlt", ability: "dex", fullKey: "slightOfHand" },
```
Fyorl, line 147:
```suggestion
  slt: { label: "DND5E.SkillSlt", ability: "dex", fullKey: "sleightOfHand" },
```

module/config.mjs
```diff
@@ -1972,6 +1982,21 @@ DND5E.advancementTypes = {
 
 /* -------------------------------------------- */
 
+let enrichmentLookup;
```
Fyorl, line 1985: Just a minor suggestion to prefix this with an underscore to make it clear that it's being used internally (even though it can't escape the module context). It's a little convention we've used sometimes in config files to do similar things.

module/enrichers.mjs
```diff
@@ -0,0 +1,303 @@
...
+      if ( config.dc ) label = game.i18n.format("EDITOR.DND5E.Inline.DC", { dc: config.dc, check: label });
+      label = game.i18n.format(`EDITOR.DND5E.Inline.Check${longSuffix}`, { check: label });
+    }
+  }
+
+  if ( config.passive ) {
+    const span = document.createElement("span");
+    span.classList.add("passive-check");
+    span.innerText = label;
+    return span;
```
Fyorl, line 180: Might be nice to factor this out into its own function too like `createPassiveElement` or something. I like how compartmentalised this module is currently.

Fyorl (approved):
This looks great, thank you Jeff. Just one or two suggestions, which you can take or leave, feel free to merge when ready.

Issue #2388, opened 2023-07-27
Truncate `template.json` and defer to the system data models where possible.


Fyorl:
Now that we have made a closing v10 release that fixes the issue whereby system data models still relied on `template.json` for their `getInitialValue`, it should be possible to largely truncate the `template.json`.

Related #1933

Fyorl:
Marking this as blocked since the core API still somewhat relies on template.json in order to infer data types for Active Effect application. We can either wait for https://github.com/foundryvtt/foundryvtt/issues/6631 to be resolved, or alternatively we can implement our own code in our existing Active Effect application override to always look at the data model (we might end up having to do this anyway if we want to retain v11 compatibility when v12 is out).

PR #2503, opened 2023-10-20
[#342] Add Race type, update Movement & Type config to work with items

arbron:
Add Race item type with data structure to handle advancement, movement, darkvision, and creature type. This required some changes to the `ActorTypeConfig` and `ActorMovementConfig` to handle editing items.

Screenshot 2023-10-20 at 11 29 14


MaxPat931:
The different display of Darkvision being its own line instead of a title and box, like Type and Movement looks just a little off to me. Even though all official races only provide Darkvision as an additional sense, perhaps it should be a title of Senses, then the specified sense in a box, for more consistency/flexibility for homebrew?

template.json
```diff
@@ -676,6 +676,24 @@
...
+        "fly": null,
+        "swim": null,
+        "walk": 30,
+        "units": "ft"
+      },
+      "type": {
+        "value": "humanoid",
+        "subtype": ""
+      }
+    },
```
aaclayton, line 696: Since we are defining system data models, the contents of `template.json` are actually not required beyond declaring the type as valid. We could just do:
```suggestion
    "race": {},
```
Same goes for all other types, if we wanted to make that change.

module/data/item/race.mjs
```diff
@@ -0,0 +1,73 @@
...
+ * @property {string} type.value       Creature type (e.g. "humanoid", "construct").
+ * @property {string} type.subtype     Additional type data (e.g. "elf" for Eladrin)
+ */
+export default class RaceData extends SystemDataModel.mixin(ItemDescriptionTemplate) {
+  /** @inheritdoc */
+  static defineSchema() {
+    return this.mergeSchema(super.defineSchema(), {
+      identifier: new IdentifierField({label: "DND5E.Identifier"}),
+      advancement: new foundry.data.fields.ArrayField(new AdvancementField(), {label: "DND5E.AdvancementTitle"}),
+      darkvision: new foundry.data.fields.NumberField({min: 0, integer: true, label: "DND5E.SenseDarkvision"}),
```
aaclayton, line 29: Would it be worthwhile to generalize that the `race` contains a `senses` data structure that then translates onto `actor.system.attributes.senses`? Might certain races have something other than darkvision that we would need to record? Perhaps not but it could accommodate for homebrew while also reusing an existing data structure that we have.

module/data/item/race.mjs
```diff
@@ -0,0 +1,73 @@
...
+      movement: new foundry.data.fields.SchemaField({
```
aaclayton, line 30: It would be nice to have this share the same movement schema that an actor has, similar to my comment about senses above. If we have a `MovementData` model we can embed it here like:
```js
movement: new fields.SchemaField(MomvementData.defineSchema())
```
Or as:
```js
movement: new fields.EmbeddedDataField(MomvementData)
```
Which is a bit more powerful, but also slightly heavier for performance.

module/data/item/race.mjs
```diff
@@ -0,0 +1,73 @@
...
+ * @property {number} movement.walk
+ * @property {string} movement.units   Units used to measure movement.
+ * @property {object} type
+ * @property {string} type.value       Creature type (e.g. "humanoid", "construct").
+ * @property {string} type.subtype     Additional type data (e.g. "elf" for Eladrin)
+ */
+export default class RaceData extends SystemDataModel.mixin(ItemDescriptionTemplate) {
+  /** @inheritdoc */
+  static defineSchema() {
+    return this.mergeSchema(super.defineSchema(), {
```
aaclayton, line 26: Recommend aliasing:
```js
const fields = foundry.data.fields;
```
at the top of the function so that the rest of the code reads a bit more concisely.

module/data/item/race.mjs
```diff
@@ -0,0 +1,73 @@
...
+        swim: new foundry.data.fields.NumberField({
+          initial: null, min: 0, integer: true, label: "DND5E.MovementSwim"
+        }),
+        walk: new foundry.data.fields.NumberField({
+          initial: 30, min: 0, integer: true, label: "DND5E.MovementWalk"
+        }),
+        units: new foundry.data.fields.StringField({initial: "ft", label: "DND5E.MovementUnits"})
+      }),
+      type: new foundry.data.fields.SchemaField({
+        value: new foundry.data.fields.StringField({initial: "humanoid", label: "DND5E.CreatureType"}),
```
aaclayton, line 49: `type.value` field should specify a valid set of `choices` I think. While we're here, do we like `value` as this attribute path? I know we've used it before, but maybe something like `type.primary` or `type.category` would be more accurate?

module/data/item/race.mjs
```diff
@@ -0,0 +1,73 @@
...
+ * @property {object} type
+ * @property {string} type.value       Creature type (e.g. "humanoid", "construct").
+ * @property {string} type.subtype     Additional type data (e.g. "elf" for Eladrin)
+ */
+export default class RaceData extends SystemDataModel.mixin(ItemDescriptionTemplate) {
+  /** @inheritdoc */
+  static defineSchema() {
+    return this.mergeSchema(super.defineSchema(), {
+      identifier: new IdentifierField({label: "DND5E.Identifier"}),
+      advancement: new foundry.data.fields.ArrayField(new AdvancementField(), {label: "DND5E.AdvancementTitle"}),
```
aaclayton, line 28: Can we have structured data for the ability score bonuses or penalties that a race provides? Or does that need to go in as a type of advancement? It seems like we should have a `SchemaField` for `abilities` here.

module/data/item/race.mjs
```diff
@@ -0,0 +1,73 @@
...
+      type: new foundry.data.fields.SchemaField({
+        value: new foundry.data.fields.StringField({initial: "humanoid", label: "DND5E.CreatureType"}),
+        subtype: new foundry.data.fields.StringField({label: "DND5E.CreatureTypeSelectorSubtype"})
+      })
+    });
+  }
+
+  /* -------------------------------------------- */
+
+  prepareDerivedData() {
```
aaclayton, line 57: I know we've done this elsewhere for other models, but it's probably worth considering whether we really *need* this system of prepared `labels` here. Feels like we could prepare labels elsewhere (like in the Race config sheet) without needing them to be cached on the `Item` instance.

module/applications/item/item-sheet.mjs
```diff
@@ -486,6 +490,28 @@ export default class ItemSheet5e extends ItemSheet {
     ];
   }
 
+  /* -------------------------------------------- */
+  /**
+   * Handle spawning the configuration applications.
+   * @param {Event} event   The click event which originated the selection.
+   * @private
```
aaclayton, line 497: Now might not be the right time to start making this change, but I would vote for moving to formal private methods `#onConfigMenu` in any situation where we would be declaring the method as `@private` and reserve underscores for either `@internal` or `@protected` methods. The convention we now use in the core software.

module/applications/actor/type-config.mjs
```diff
@@ -15,31 +15,41 @@ export default class ActorTypeConfig extends FormApplication {
...
   /* -------------------------------------------- */
 
-  /** @override */
-  getData(options={}) {
+  /**
+   * Key path to where the type is stored on the document.
+   * @type {string}
+   */
+  get keyPath() {
+    return this.document instanceof Actor ? "system.details.type" : "system.type";
```
aaclayton, line 44: Instead of having to deduce this key path from within the context of an `ActorTypeConfig` instance, would be better to pass in the config as an `option` to the constructor when we create a `new ActorTypeConfig`. i.e.
```js
new ActorTypeConfig(race, {keyPath: "system.details.type"});
new ActorTypeConfig(actor, {keyPath: "system.type"});
```

module/applications/actor/movement-config.mjs
```diff
@@ -5,34 +5,37 @@ import BaseConfigSheet from "./base-config.mjs";
...
   }
 
   /* -------------------------------------------- */
 
-  /** @override */
+  /** @inheritdoc */
   getData(options={}) {
     const source = this.document.toObject();
+    const isActor = this.document instanceof Actor;
+    const keyPath = isActor ? "system.attributes.movement" : "system.movement";
```
aaclayton, line 32: See similar comment for `ActorTypeConfig` about the merits of passing in `keyPath` as an option to the constructor.

templates/items/race.hbs
```diff
@@ -0,0 +1,67 @@
+<form class="{{cssClass}} flexcol" autocomplete="off">
```
aaclayton, line 1: Should perhaps consider naming the file `race-sheet.hbs` or `race-config.hbs` instead of just "race".

aaclayton (changes requested):
Looking good, some suggested changes from my end. Will defer final review and approval to @Fyorl

template.json
Fyorl: We do want to truncate the template.json entirely (#2388) at some point, so I agree that we should do so here as a start since it's a new item type without any associated baggage yet.

module/data/item/race.mjs
Fyorl: Yeah another +1 here that we should reuse senses and movement schema and allow races to provide more than just darkvision as a sense.

module/data/item/race.mjs
Fyorl: There's unfortunately a bit of tension with regards to putting explicit `choices` in the DataModel while still allowing the system to be configurable. If we set `choices` to something like `CONFIG.creatureTypes`, and a module adds something to that, any Actors that are subsequently configured with that creature type will become invalid if the module is ever disabled.

There are a few things we've thought about doing here: Deferring to preparation to strip out 'unconfigured' values, or writing custom validators to do something similar. At the moment we've opted to just not use `choices` for now, until we either implement some of the aforementioned options, or there is some core API solution in place.

module/data/item/race.mjs
Fyorl: I think we want to use the `AbilityScoreImprovementAdvancement` for this, it was built with some accommodations for race items already.

module/data/item/race.mjs
arbron: The problem with setting a `choices` is it easily leads to entire actors/items being invalid. If a module or script adds a choice and is then disabled, any actors or items that use that choice will just disappear entirely from the world and the only way a user can know what went wrong is through the inspector. I'd prefer to avoid `choices` and just degrade more gracefully when creating labels and whatnot.

Renaming would work, but would we also want to migrate the NPC data structure to match?

module/applications/item/item-sheet.mjs
arbron: In this case I think this should probably be a `@protected` in case subclasses of item sheet want to extend it.

Issue #2495, opened 2023-10-17
Allow module-provided actor subtypes to use core data preparation


JPMeehan:
As a developer making a module that leverages actor subtypes, I would like to be able to use the `system.prepareBaseData()` and `system.prepareDerivedData()` methods that are naturally part of Foundry's data preparation cycle. Unfortunately, the Actor5e class fully suppresses those methods. I would like to be able to use natural foundry data preparation processes to support my module subtypes.

Related to #2094

PR #2506, opened 2023-10-20
[#2495] Allow prepareData methods to be called on module actors

arbron:
- `SystemDataModel` now inherits from `TypeDataModel`
- Renamed `_prepareBaseData` on `GroupData` to match naming in core
- Adjusted data preparation to ensure `TypeDataModel` preparation methods are still called

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -132,8 +132,7 @@ export default class Actor5e extends Actor {
 
   /** @inheritDoc */
   prepareDerivedData() {
-    if ( !game.template.Actor.types.includes(this.type) ) return;
-    if ( this.type === "group" ) return;
+    if ( !game.template.Actor.types.includes(this.type) || (this.type === "group") ) return;
 
     const flags = this.flags.dnd5e || {};
     this.labels = {};
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -91,8 +91,7 @@ export default class Actor5e extends Actor {
 
   /** @inheritDoc */
   prepareData() {
-    // Do not attempt to prepare non-system types.
-    if ( !game.template.Actor.types.includes(this.type) ) return;
+    if ( !game.template.Actor.types.includes(this.type) ) return super.prepareData();
     this._classes = undefined;
     this._preparationWarnings = [];
     super.prepareData();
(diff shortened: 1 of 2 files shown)


Fyorl (changes requested):
I have a suspicion that `TypeDataModel` is v11-only, which is why we do not use it currently. That's also why I opted to check `game.template` for the types rather than checking `system.modelProvider`.

The rest of the changes look fine to me except that we might have to continue manually calling `GroupActor#prepareBaseData` for now while we have v10 compatibility.

It's unclear whether #2495 can be resolved without inheriting from `TypeDataModel`. I don't know if we want to punt this issue to a later version where we drop v10 support, or whether we can dynamically inherit from `DataModel` or `TypeDataModel` based on `game.release.generation`.

module/documents/actor/actor.mjs
```diff
@@ -103,13 +102,9 @@ export default class Actor5e extends Actor {
...
-
-    // Delegate preparation to type-subclass
-    if ( this.type === "group" ) {  // Eventually other types will also support this
-      return this.system._prepareBaseData();
-    }
-
-    this._prepareBaseArmorClass();
+    if ( !game.template.Actor.types.includes(this.type) ) return;
+    if ( this.type !== "group" ) this._prepareBaseArmorClass();
+    else if ( game.release.generation < 11 ) this.system.prepareBaseData();
```
Fyorl, line 107: I checked the core API and `system.prepareBaseData()` will only be called automatically if `system` is an instance of `TypeDataModel`, which `GroupActor` is not anymore. So I think we must always call it manually until we drop v10 support and `SystemDataModel` can inherit from `TypeDataModel`.

```suggestion
    if ( this.type === "group" ) this.system.prepareBaseData();
    else this._prepareBaseArmorClass();
```

module/documents/actor/actor.mjs
```diff
@@ -137,11 +132,8 @@ export default class Actor5e extends Actor {
 
   /** @inheritDoc */
   prepareDerivedData() {
-
-    // Delegate preparation to type-subclass
-    if ( this.type === "group" ) {  // Eventually other types will also support this
-      return this.system._prepareDerivedData();
-    }
+    if ( !game.template.Actor.types.includes(this.type) ) return;
+    if ( this.type === "group" ) return;
```
Fyorl, line 136:
```suggestion
    if ( !game.template.Actor.types.includes(this.type) || (this.type === "group") ) return;
```

Fyorl (changes requested):

module/documents/actor/actor.mjs
arbron: Core checks if it is `TypeDataModel` or if it has a Function called `prepareBaseData`, so this is needed to avoid calling that method twice.

```javascript
// Core
if ( isTypeData || (this.system?.prepareBaseData instanceof Function) ) this.system.prepareBaseData();
```

Fyorl: Ah, right, I totally glossed over that, that makes sense, thanks.

Issue #2479, opened 2023-09-19
Using a consumable item with a Negative Consumption value fails if the consumed resource is zero.


FrankIsAtank:
FoundryVTT V 11 Stable 306
DnD5e 2.3.1

To reproduce : 
1. Create a player character actor.
2. Add a Consumable item to the actor.
3. Give the new item these properties :
- Activation Cost : Action
- Resource Consumption -1 'resources.primary.value' Attribute
4. Return to the character sheet and try to consume the new item.

Observed behaviour :
The item fails to be consumed, presumably because it checks that the consumed resource is available.
A warning notification appears with "New Consumable has run out of its designated Attribute!".

Expected behaviour :
With a negative cost, the intent is for consuming the item to increase an attribute, such as adding to a resource counter. Doing so when that attribute is zero should not fail.

Hoppyhob:
I wasn't able to recreate this with:
FoundryVTT V 11 Stable 308
DnD5e 2.3.1 or the 2.4.x branch

It appears from the code we warn with this message(ConsumeWarningNoQuantity) here (in item.mjs):
```
let remaining = quantity - amount;
    if ( remaining < 0 ) {
      ui.notifications.warn(game.i18n.format("DND5E.ConsumeWarningNoQuantity", {name: this.name, type: typeLabel}));
      return false;
    }
```
Which with a quantity of zero (gotten from resources.primary.value or defaulted to 0 if null) would be working as expected 0 - -1 = 1 which is greater than zero. (This should only warn if the resource value is -2 or lower)

krbz999:
> Which with a quantity of zero (gotten from resources.primary.value or defaulted to 0 if null) would be working as expected 0 - -1 = 1 which is greater than zero. (This should only warn if the resource value is -2 or lower)

I suspect it is related to #2165.

`resource = foundry.utils.getProperty(this.actor.system, consume.target)` becomes `undefined`, and then fails here:
```js
    // Verify that a consumed resource is available
    if ( resource === undefined ) {
      ui.notifications.warn(game.i18n.format("DND5E.ConsumeWarningNoSource", {name: this.name, type: typeLabel}));
      return false;
    }
```

krbz999:
OK, maybe it's not so weird that these (value and max) get deleted given that we delete them.

https://github.com/foundryvtt/dnd5e/blob/master/module/applications/actor/character-sheet.mjs#L30

Issue #2232, opened 2023-03-09
Resource value is deleted when set to 0


sgsabbage:
Hi folks,

I've noticed a weird issue where if a feature pulls from one of the primary/secondary/tertiary resources on the sheet, if the value of that resource is 0, rather than being told the resource is out of charges, it says that the attribute no longer exists.

This has a separate knock-on effect if you edit a feature on a character while the resource is 0, which is that the attribute disappears from the list, e.g: 
character with resource at 0:
![image](https://user-images.githubusercontent.com/1389994/224042319-7459a305-aa9f-4380-af49-d3278c95bd98.png)
attribute options:
![image](https://user-images.githubusercontent.com/1389994/224042669-8e83cfa9-7afb-46d7-98f8-5998fcbed351.png)

Which means if you happen to edit a feature while its linked resource is 0, it immediately delinks the resource even if you make no other changes.

I believe this is down to this line of code: https://github.com/foundryvtt/dnd5e/blob/master/module/applications/actor/character-sheet.mjs#L30 where the value is deleted if the resource exists and the value is 0 - though if the resource is unconfigured, it returns a `null` instead, e.g:
```json
{
    "primary": {
        "max": 10,
        "sr": false,
        "lr": false,
        "label": "Psychic",
        "name": "primary",
        "placeholder": "Resource 1"
    },
    "secondary": {
        "value": null,
        "max": null,
        "sr": false,
        "lr": false,
        "label": "",
        "name": "secondary",
        "placeholder": "Resource 2"
    },
    "tertiary": {
        "value": null,
        "max": null,
        "sr": false,
        "lr": false,
        "label": "",
        "name": "tertiary",
        "placeholder": "Resource 3"
    }
}
```

I'm happy to raise a PR to either remove that line of code or replace it with a null, but as I have no other experience with the module, I don't know what this has the potential to break.

LukasPrism:
FWIW this issue screws up Monk’s Token Bar if that is set to display resources and they drop to 0. Perhaps Monk will circumvent it at his end.

krbz999:
This seems to be the same issue as #2165.

Issue #2165, opened 2023-02-07
`Actor5e#system.resources[x].value` is (sometimes) `undefined`


krbz999:
¯\\\_(ツ)\_/¯

![image](https://user-images.githubusercontent.com/50169243/217340935-8a5f5143-b40c-49a5-8805-44fd007d18eb.png)

Completely fresh actor that has never been interacted with in any way:

![image](https://user-images.githubusercontent.com/50169243/217341423-86a7c81e-3ce1-43a9-a1e7-d573ce915abf.png)

Completely fresh actor that has never been interacted with in any way except for being dragged to the scene and selected for this purpose:

![image](https://user-images.githubusercontent.com/50169243/217341587-2fc22cf5-8a41-4e63-8140-7e59df75dbc4.png)

PR #2510, opened 2023-10-22
[#2479, #2232] Prevent resource objects from being mutated

krbz999:
In the character sheet, not only was `value` (and sometimes `max`) getting deleted, all resource objects were getting mutated with `name` and `placeholder`.

While unsure if actually related, I am also unable to reproduce #2165 with this fix (with a character sheet open or not), but I am if I swap to a different branch without this fix.

image

### DIFF
--- a/module/applications/actor/character-sheet.mjs
+++ b/module/applications/actor/character-sheet.mjs
@@ -24,11 +24,10 @@ export default class ActorSheet5eCharacter extends ActorSheet5e {
 
     // Resources
     context.resources = ["primary", "secondary", "tertiary"].reduce((arr, r) => {
-      const res = context.actor.system.resources[r] || {};
-      res.name = r;
-      res.placeholder = game.i18n.localize(`DND5E.Resource${r.titleCase()}`);
-      if (res && res.value === 0) delete res.value;
-      if (res && res.max === 0) delete res.max;
+      const res = foundry.utils.mergeObject(context.actor.system.resources[r], {
+        name: r,
+        placeholder: game.i18n.localize(`DND5E.Resource${r.titleCase()}`)
+      }, {inplace: false});
       return arr.concat([res]);
     }, []);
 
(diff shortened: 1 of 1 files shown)


module/applications/actor/character-sheet.mjs
```diff
@@ -24,11 +24,10 @@ export default class ActorSheet5eCharacter extends ActorSheet5e {
 
     // Resources
     context.resources = ["primary", "secondary", "tertiary"].reduce((arr, r) => {
-      const res = context.actor.system.resources[r] || {};
-      res.name = r;
-      res.placeholder = game.i18n.localize(`DND5E.Resource${r.titleCase()}`);
-      if (res && res.value === 0) delete res.value;
-      if (res && res.max === 0) delete res.max;
```
Fyorl, line 31: Now that we've broken the reference and are (presumably) no longer mutating real actor data, it should be fine to still have these, no? Is it no longer relevant to the sheet logic?

module/applications/actor/character-sheet.mjs
```diff
@@ -24,11 +24,10 @@ export default class ActorSheet5eCharacter extends ActorSheet5e {
...
+      const res = foundry.utils.mergeObject(context.actor.system.resources[r], {
```
Fyorl, line 27: No harm in keeping this guard here unless we're confident it's never needed.
```suggestion
      const res = foundry.utils.mergeObject(context.actor.system.resources[r] || {}, {
```

Fyorl (changes requested):

module/applications/actor/character-sheet.mjs
krbz999: From what I gather either seems fine. Either the value is 0 and gets deleted and a placeholder (which is '0') gets shown, or we show the value (which may we '0').

Fyorl (approved):

Fyorl:
Good fixes, thank you, Zhell

Issue #2501, opened 2023-10-20
Add price in the `item-properties`


InnilToo:
**Use case:** Having all useful information in one place makes running a game always smoother. No having to open another window (item edit in this case) is always welcomed and can prevent accidental edits.

**Problem:** Items already display a lot of useful information if you "roll them out" so why not add more if there is room?

**Feature request:** Add price in the `item-properties`.

![image](https://github.com/foundryvtt/dnd5e/assets/33260912/e6e5247e-c5ee-4b67-ba52-377bae98a1cf)

krbz999:
This would be especially useful on loot.

Fyorl:
Prices are available as a column on actor sheets now, and in the item tooltip, and we also include price in loot item tags.

PR #2517, opened 2023-10-24
[#2501] Show price on loot and consumable items

krbz999:
This shows up as a 'tag' in chat messages and in the actor sheet when expanded.

### DIFF
--- a/module/data/item/templates/physical-item.mjs
+++ b/module/data/item/templates/physical-item.mjs
@@ -36,6 +36,20 @@ export default class PhysicalItemTemplate extends SystemDataModel {
   }
 
   /* -------------------------------------------- */
+  /*  Getters                                     */
+  /* -------------------------------------------- */
+
+  /**
+   * Get a human-readable label for the price and denomination.
+   * @type {string}
+   */
+  get priceLabel() {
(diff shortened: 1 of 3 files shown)


module/data/item/templates/physical-item.mjs
```diff
@@ -35,6 +35,20 @@ export default class PhysicalItemTemplate extends SystemDataModel {
...
+  /* -------------------------------------------- */
+
+  /**
+   * Get a human-readable label for the price and denomination.
+   * @type {string}
+   */
+  get priceLabel() {
+    const { value, denomination } = this.price;
+    const hasPrice = value && (denomination in CONFIG.DND5E.currencies);
+    return hasPrice ? `${value} ${game.i18n.localize(CONFIG.DND5E.currencies[denomination].label)}` : null;
```
Fyorl, line 49: `CONFIG.DND5E.currencies.x.label` should already be localised by this point I believe.
```suggestion
    return hasPrice ? `${value} ${CONFIG.DND5E.currencies[denomination].label}` : null;
```

Fyorl (changes requested):

Fyorl (approved):

Issue #2100, opened 2023-01-23
Allow largest and smallest available hit dice to be used in damage formulas


maxobremer:
At the moment if you want to add hit dice to a damage formula, you have to use the class's data of the hit dice or manually fill in the number. It would be nice if a damage formula can look at the smallest or largest hit dice that the player still has available and use that to calculate damage.
Would be cool to have a weapon that consumes hit dice but turns them into extra damage.

revilowaldow:
Have just had a fun chat in #macro-polo with Zhell that resulted in this delight:

```
1d(max(
	@classes.Artificer.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Barbarian.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Bard.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Cleric.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Druid.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Fighter.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Monk.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Paladin.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Ranger.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Rogue.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Sorcerer.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Warlock.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue,
	@classes.Wizard.advancement.0.parent.parent.advancement.byType.HitPoints.0.hitDieValue
))
```

+1 request for support for this please

PR #2519, opened 2023-10-24
[#2100] Turn `attributes.hd` into an object, save reference to smallest and largest

krbz999:
Instead of `attributes.hd` being an integer (number of remaining hit dice), it's now an object with `amount` (the remaining quantity), `smallest`, and `largest`.


revilowaldow:
Any reason this didn't get implemented? Still very keen to see support for #2100

Issue #2472, opened 2023-09-12
Weapons with ammunition not taking in account ammunition damage when only damage is rolled.


Zanderaf:
I'm not sure if its intended or not, but when using something such as a longbow with with +1 arrows:
- Clicking the item and sending it into chat with the attack / damage options and selecting attack to make a roll, and then damage consumes the ammo and applies the damage bonuses from the ammunition accordingly.

However, if you send the item to chat and only hit damage and skip the ammo consumption, it only rolls the base longbow's damage. Is the damage linked to the consumed ammo / this whole process intentional?

Fyorl:
Probably not intentional, no. Maybe a technical limitation, not sure. I'll mark this as a bug since I think if a weapon is using explicitly using ammunition then the ammunition properties should always apply when rolling damage.

krbz999:
The code adds a reference to the `_ammo` in the attack roll, then removes that reference when creating the damage roll.

It is also the attack roll that consumes ammo, not the damage roll.

PR #2525, opened 2023-10-26
[#2472] Change ammo logic for attack and damage rolls

krbz999:
Instead of a damage roll requiring a preceding attack roll to gain the benefit of 'ammo damage', the damage roll simply checks if the item has ammo and can attack. The attack roll simply serves to consume ammo, nothing more.

There are some additional benefits to this.
- If one user made the attack roll, another user can still gain the benefit of the ammo on the damage roll, in that weird edge case of two owners of one actor, where before the addition of `_ammo` to the item was local only.
- Users can make all their attack rolls first, and then roll all the damage rolls for easier calculation.
- Users can "redo" a damage roll that was meant to be a critical hit, without having to make another attack roll first.

### DIFF
--- a/module/data/item/templates/action.mjs
+++ b/module/data/item/templates/action.mjs
@@ -178,8 +178,8 @@ export default class ActionTemplate extends SystemDataModel {
   get criticalThreshold() {
     if ( !this.hasAttack ) return null;
     let ammoThreshold = Infinity;
-    if ( this.consume?.type === "ammo" ) {
-      ammoThreshold = this.parent?.actor?.items.get(this.consume.target).system.critical.threshold ?? Infinity;
+    if ( this.hasAmmo ) {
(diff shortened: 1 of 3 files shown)


arbron:
So the issue with this is that ammo's quantity is consumed on attack, which also means that ammo might have 0 uses remaining or the item might be fully deleted by the time damage is rolled (if `delete on empty` is checked), resulting in no ammo being available to calculate damage bonuses.

krbz999:
> So the issue with this is that ammo's quantity is consumed on attack, which also means that ammo might have 0 uses remaining or the item might be fully deleted by the time damage is rolled (if `delete on empty` is checked), resulting in no ammo being available to calculate damage bonuses.

Ammo does not get destroyed.

Fyorl (approved):

Issue #2521, opened 2023-10-25
Create a config for actor resources


krbz999:
This would improve the UI, and could be used as a helper to let users use roll data to determine the maximums.

The label and two toggles would also fit better here; only the current value is really needed for quick edits directly on the sheet.

Fyorl:
Obsoleted by the v2 sheet design.

PR #2526, opened 2023-10-26
[#2521] Move actor resources into a config

krbz999:
Also turns `resources.max` into a `FormulaField`, such that it may support roll data.

image


thatlonelybugbear:
Wouldn't it be better to also turn the `resources.value` into a `FormulaField`, as you already did half the work?

That would offer some more functionality even though usecases might not be immediatelly apparent thinking about core dnd5e ruleset.

PR #2527, opened 2023-10-26
[#2135, #2136] Add multiple inline editors to items & display chat description

### LINKED ISSUE #2135: Add ability to edit "hidden" descriptions

### LINKED ISSUE #2136: Use item's "chat description" if specified

arbron:
Adds multiple inline editors into items to allow editing normal description, unidentified description, and chat text. When the edit button is clicked the editor replaces the whole tab giving plenty of space to write.

https://github.com/foundryvtt/dnd5e/assets/19979839/09eb8de2-5b2b-409c-830c-5b9d3e559e50

This also will display the chat text in created chat messages rather than the default description if it is provided.
 
Closes #2135
Closes #2136

### DIFF
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -708,6 +708,25 @@ h5 {
   width: 16px;
   height: 16px;
 }
+/* ----------------------------------------- */
+/*  Accordion                                */
+/* ----------------------------------------- */
+.accordion {
+  transition: margin-bottom 500ms ease;
+}
+.accordion.collapsed {
+  margin-bottom: 0.5rem;
+}
+.accordion.collapsed > .accordion-content {
+  height: 0;
+}
+.accordion-heading {
+  cursor: pointer;
+}
+.accordion-content {
+  overflow: hidden;
+  transition: height 500ms ease;
+}
 .dnd5e.sheet.actor {
   /* ----------------------------------------- */
   /*  Shared Styles                            */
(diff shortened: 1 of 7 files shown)


krbz999:
My artistic skills are offensive, but hear me out. Rather than elongating the description field possibly thrice over vertically, can we split it horizontally?

image

Fyorl (changes requested):
Code looks good to me, just one comment. We do seem to be missing the actual accordion part of the 'accordion' design though. I've attached a diff that implements a (relatively) lightweight `Accordion` class to handle this, including some CSS transition animations. It will wrap any markup but we could maybe simplify things by mandating use of `<details>` elements.

https://github.com/foundryvtt/dnd5e/assets/207433/f8991122-5036-4916-bb68-170e516cffe6

Zhell raises a reasonable point that, with very long descriptions, we will have too much crammed into the pane. The implementation comes with a `collapseOthers` option that we could employ here to make better use of the space (this is the typical 'accordion' effect.)

https://github.com/foundryvtt/dnd5e/assets/207433/73b25b4a-c0cb-478d-8772-21d70ee4df68

At this point though, we should consider if going to all this effort is necessary when we could just leverage the existing tabs infrastructure.

Personally, I find it useful to be able to compare the descriptions and see them all at once. I also find it a bit more obvious which description is currently being edited when they are arranged vertically with headings in this way, rather than as tabs. I appreciate that these are fairly minor advantages, though I would hope that if we do opt to use the accordion widget, that it would open up some design space elsewhere also.

If we do decide to pursue this avenue, let me know if you'd prefer to incorporate the changes yourself referencing the diff, or if I should push a commit to this branch or something.

[accordion.patch](https://github.com/foundryvtt/dnd5e/files/13207767/accordion.patch)

arbron:
@Fyorl I merged your accordion changes in, with the `collapseOthers` set to `false` based on popular opinion. I also added a transition to `margin-bottom` to avoid that little jump that was happening :).

Fyorl (approved):

Issue #2478, opened 2023-09-18
Add Details tab and Loot subtypes


krbz999:
Building onto #2455, an additional tab ('Details') with very basic information could be added. While loot-type items have always been intended to be very basic, extending the system model with subtypes like most (all?) other item types have would be useful.

Some suggestions:
- gem
- treasure
- resource
- junk

PR #2532, opened 2023-10-29
[#2478] Add details tab and loot subtypes

krbz999:
image


module/config.mjs
```diff
@@ -841,6 +841,37 @@ preLocalize("featureTypes.class.subtypes", { sort: true });
...
+DND5E.lootTypes = {
+  junk: {
+    label: "DND5E.Loot.Junk"
+  },
+  resource: {
+    label: "DND5E.Loot.Resource"
+  },
+  treasure: {
+    label: "DND5E.Loot.Treasure",
+    subtypes: {
```
Fyorl, line 864: We have `loot` which is a sub-type of Item, and `treasure` which is a sub-type of `loot`, and now `gem` which is a sub-type of `treasure`. I think we could simplify things in a few places if we flattened this out and had `gem` and `art` just be direct sub-types of `loot`.

At most we could group them in an `<optgroup>` like we do for armour.

image

Fyorl:
I'm ok with a separate tab but since there's so little there I wonder if we'd be better off putting these drop-downs in the top section (alongside rarity and source), or on the left side of the description tab (underneath price and quantity.) What do you think?

krbz999:
> I'm ok with a separate tab but since there's so little there I wonder if we'd be better off putting these drop-downs in the top section (alongside rarity and source), or on the left side of the description tab (underneath price and quantity.) What do you think?

A second tab would both be useful for module developers (who have been injecting their own new tabs, which spurred this on), as well as make the Loot item sheet look less silly when it only has 1 tab.

PR #2533, opened 2023-10-29
[#2529] Implement damage enricher

### LINKED ISSUE #2529: Add damage enricher
This would be an enricher that would be able to display formatted damage including the average generated automatically from the roll, and a damage type that could be passed along with the roll.

```
Any creature standing on a pressure plate when its trap triggers must succeed on
a [[/save ability=dex dc=14]] saving throw or take [[/damage 2d6 type=bludgeoning]]
damage from the flailing armor.
```

```html
Any creature standing on a pressure plate when its trap triggers must succeed on a
<a class="roll-action" data-type="save" data-key="dex" data-dc="20">
<i class="fa-solid fa-dice-d20"></i> DC 14 Dexterity</a> saving throw or take 7
(<a class="roll-action" data-type="damage" data-damage-type="bludgeoning">
<i class="fa-solid fa-dice-d20"></i> 2d6</a>) bludgeoning damage from the flailing armor.
```

arbron:
Implements a damage enricher in the format `[[/damage 1d6 bludgeoning]]` or `[[/damage formula=1d6 type=bludgeoning]]` which will present the normal damage rolling dialog, allowing criticals to be rolled and damage types to be passed through.

Screenshot 2023-10-29 at 10 58 32

Closes #2529

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -1,4 +1,5 @@
 import { simplifyBonus } from "./utils.mjs";
+import { damageRoll } from "./dice/_module.mjs";
 import * as Trait from "./documents/actor/trait.mjs";
 
 /**
@@ -6,7 +7,7 @@ import * as Trait from "./documents/actor/trait.mjs";
  */
 export function registerCustomEnrichers() {
   CONFIG.TextEditor.enrichers.push({
-    pattern: /\[\[\/(?<type>check|save|skill|tool) (?<config>[^\]]+)]](?:{(?<label>[^}]+)})?/gi,
+    pattern: /\[\[\/(?<type>check|damage|save|skill|tool) (?<config>[^\]]+)]](?:{(?<label>[^}]+)})?/gi,
     enricher: enrichString
   });
 
@@ -25,8 +26,9 @@ export function registerCustomEnrichers() {
 export async function enrichString(match, options) {
   let { type, config, label } = match.groups;
   config = parseConfig(config, match.input);
-  config.input = match.input;
+  config.input = match[0];
   switch ( type.toLowerCase() ) {
+    case "damage": return enrichDamage(config, label, options);
     case "check":
     case "skill":
     case "tool": return enrichCheck(config, label, options);
@@ -43,14 +45,15 @@ export async function enrichString(match, options) {
  * @returns {object}
  */
 function parseConfig(match) {
-  const config = {};
+  const config = { values: [] };
   for ( const part of match.split(" ") ) {
     if ( !part ) continue;
     const [key, value] = part.split("=");
     const valueLower = value?.toLowerCase();
-    if ( ["true", "false"].includes(valueLower) ) config[key] = valueLower === "true";
+    if ( value === undefined ) config.values.push(key);
+    else if ( ["true", "false"].includes(valueLower) ) config[key] = valueLower === "true";
     else if ( Number.isNumeric(value) ) config[key] = Number(value);
-    else config[key] = value ?? true;
+    else config[key] = value;
   }
   return config;
 }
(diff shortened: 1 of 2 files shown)


Fyorl:
There are some (rare) cases where multiple damage types are called for. Meteor Swarm and Flame Strike come to mind, but I'm sure there are more.

>  A creature takes 20d6 fire damage and 20d6 bludgeoning damage on a failed save

That could of course be expressed with `[[/damage 20d6 fire]] and [[/damage 20d6 bludgeoning]]`, but for the purposes of applying damage and managing saving throws, etc., it becomes a lot easier if they could be included in a single chat card. Perhaps we could support a syntax like `[[/damage 20d6 fire, 20d6 bludgeoning]]`, separating on a comma or a pipe `|`, or something. What do you think?

arbron:
> There are some (rare) cases where multiple damage types are called for. Meteor Swarm and Flame Strike come to mind, but I'm sure there are more.
> 
> > A creature takes 20d6 fire damage and 20d6 bludgeoning damage on a failed save
> 
> That could of course be expressed with `[[/damage 20d6 fire]] and [[/damage 20d6 bludgeoning]]`, but for the purposes of applying damage and managing saving throws, etc., it becomes a lot easier if they could be included in a single chat card. Perhaps we could support a syntax like `[[/damage 20d6 fire, 20d6 bludgeoning]]`, separating on a comma or a pipe `|`, or something. What do you think?

I think it makes sense to be able to handle multiple formulas in one enricher, though maybe we should hold off implementing that portion until we can handle multiple rolls in a damage roll properly.

That format looks good, and I don't think it would be a problem to add support for multiple rolls later. We could even modify the formatter to still support keyed values (e.g. `[[/damage 20d6 type=fire, 20d6 type=bludgeoning]]` or even `[[/damage formula=20d6 type=fire, formula=20d6 type=bludgeoning]]` if you want to be really specific).

One question about the multiple damages format is what do we make the roll link. Right now I only have the dice itself as the link, but that becomes a bit more of an issue with multiple types. We could put a link on each of the dice values, but that might be confusing if they both produce the same damage roll with both types.

module/enrichers.mjs
```diff
@@ -180,6 +180,77 @@ export async function enrichCheck(config, label, options) {
 
 /* -------------------------------------------- */
 
+/**
+ * Enrich a damage link.
+ * @param {string[]} config            Configuration data.
+ * @param {string} label               Optional label to replace default text.
+ * @param {EnrichmentOptions} options  Options provided to customize text enrichment.
```
Fyorl, line 187: I might have missed this in the earlier review but these are documented as being optional but not marked in the JSDoc as optional, e.g.

```suggestion
 * @param {string} [label]               Optional label to replace default text.
 * @param {EnrichmentOptions} [options]  Options provided to customize text enrichment.
```

And that goes for `enrichCheck` too. I think `options` is guaranteed to be passed-in by the calling `TextEditor` methods, but since these functions are exported, presumably they might be called separately from that workflow. On that note, do we know why these methods are exported? It seems that the only entry-point we actually need is `registerCustomEnrichers`, and that the rest of the functions can be private to the module, no?

module/enrichers.mjs
```diff
@@ -180,6 +180,77 @@ export async function enrichCheck(config, label, options) {
...
+ * 666 (<a class="roll-action" data-type="damage" data-formula="8d4dl" data-damage-type="force">
+ *   <i class="fa-solid fa-dice-d20"></i> 8d4dl
+ * </a> force
+ * ````
+ */
+export async function enrichDamage(config, label, options) {
+  const formulaParts = [];
+  if ( config.formula ) formulaParts.push(config.formula);
+  for ( const value of config.values ) {
+    if ( value in CONFIG.DND5E.damageTypes ) config.type = value;
```
Fyorl, line 221: We could add our well-known boolean values here to allow them to be passed in as 'flag-style' still:

```suggestion
    if ( value in CONFIG.DND5E.damageTypes ) config.type = value;
    else if ( value === "average" ) config.average = true;
```

Fyorl (changes requested):
Only some minor comments/questions from me, looks good otherwise.

> One question about the multiple damages format is what do we make the roll link. Right now I only have the dice itself as the link, but that becomes a bit more of an issue with multiple types. We could put a link on each of the dice values, but that might be confusing if they both produce the same damage roll with both types.

I think that is a good question, and one that I do not have a good suggestion for either. I guess we should come back to this when we are doing the wider system work on multi-damage rolls.

module/enrichers.mjs
arbron: I was thinking just parsing any unrecognized values as flags:

```suggestion
    if ( value in CONFIG.DND5E.damageTypes ) config.type = value;
    else config[value] = true;
```

Fyorl (approved):
Approved with one suggestion.

Issue #2504, opened 2023-10-20
Expand the Actor and Item `source` field to contain more structured data


aaclayton:
Suggest that the Source field which is currently a string field be expanded to be displayed read-only on the sheet with a configuration cog to open a mini config app for the source. Properties of the source configuration would be:

## Source Data
- Book (optional, a select input populated with a configured list of known sources)
- Page (optional, a string in case some page numbers are non-numeric like "Appendix I")
- Custom (optional, a custom string which overrides Book and Page)
- UUID (optional, a UUID which represents an upstream source of the data `flags.core.sourceId`)
- Creative Commons (boolean, a checkmark to keep track of content which is has a creative commons license. Alternatively a license dropdown field if we wanted to support multiple options like CC, OGL, ORC, custom etc...)

## Module Extensibility
3rd party modules like "Tome of Beasts" or "Flee Mortals" could register themselves to `DND5E.CONFIG.SOURCES` in order to appear in the Book dropdown.

## Benefits
- Present source references in a standardized format
- Useful for content creators building adventures with imported data to customize content sources for custom import behaviors
- Useful for filtering content in a future compendium browser UI to identify content from a particular book
- Licensing rights are important and the future of D&D may involve pretty mixed content with CC/SRD/ORC/Unlicensed all in play

## Data Migration
The data migration to support this would be relatively easy. The current source string would be migrated to `source.custom`.

PR #2534, opened 2023-10-29
[#2504] Convert source on items & actors into object

arbron:
First part of #2504 that migrates the data structure to an object. This updates the existing templates to display and reference `source.custom` so a future PR can add a source configuration app.

```javascript
{
  book: "",
  page: "",
  custom: "",
  uuid: "",
  license: ""
}
```

### DIFF
--- a/module/data/actor/npc.mjs
+++ b/module/data/actor/npc.mjs
@@ -136,8 +136,9 @@ export default class NPCData extends CreatureTemplate {
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
   static #migrateSource(source) {
-    if ( !source.details?.source || foundry.utils.getType(source.details.source) === "Object" ) return;
-    source.details.source = { custom: source.details.source };
+    if ( source.details?.source && (foundry.utils.getType(source.details.source) !== "Object") ) {
+      source.details.source = { custom: source.details.source };
+    }
   }
 
   /* -------------------------------------------- */
--- a/module/data/actor/npc.mjs
+++ b/module/data/actor/npc.mjs
@@ -1,4 +1,5 @@
 import { FormulaField } from "../fields.mjs";
+import SourceField from "../shared/source-field.mjs";
 import AttributesFields from "./templates/attributes.mjs";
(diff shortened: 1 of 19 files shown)


module/data/actor/npc.mjs
```diff
@@ -123,12 +124,24 @@ export default class NPCData extends CreatureTemplate {
...
 
   /* -------------------------------------------- */
 
+  /**
+   * Convert source string into custom object.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static #migrateSource(source) {
+    if ( !source.details?.source || foundry.utils.getType(source.details.source) === "Object" ) return;
+    source.details.source = { custom: source.details.source };
```
aaclayton, line 140: I think we need to be careful and ensure that `source.details.source` is present:
```suggestion
    if ( ("source" in source.details) && (foundry.utils.getType(source.details.source) !== "Object") ) {
      source.details.source = {custom: source.details.source};
    } 
```

module/data/shared/source-field.mjs
```diff
@@ -0,0 +1,47 @@
...
+
+  /* -------------------------------------------- */
+
+  initialize(value, model, options={}) {
+    const obj = super.initialize(value, model, options);
+
+    Object.defineProperty(obj, "label", {
+      get() {
+        // TODO: Implement logic for generating source
+        return this.custom ?? "";
```
aaclayton, line 32: Perhaps for now (can be improved later):
```suggestion
        return this.custom ?? `${this.book} ${this.page}`;
```

module/data/shared/source-field.mjs
```diff
@@ -0,0 +1,47 @@
...
+      },
+      enumerable: false
+    });
+    Object.defineProperty(obj, "toString", {
```
aaclayton, line 36: Good call with assigning a `toString` redirect.

aaclayton (changes requested):
Looking good to me, thanks @arbron. Get any feedback from @Fyorl but this looks able to move forward.

Fyorl (approved):

Issue #2538, opened 2023-10-30
Allow data models & template to define custom socket event handlers


arbron:
Currently there is a lot of type-specific socket event handlers defined in `Actor5e` and `Item5e` and module-provided types can only define their own handlers using hooks. Adding a calling system similar to what is used for `_migrateData` to call DataModel-specific handlers would be useful and would allow us to shift a bunch of logic out of the base document classes.

PR #2541, opened 2023-10-31
[#342, #2538] Display race on sheet, add singleton logic to race & background items

arbron:
Displays race item above background on features tab of character sheet.

Adds a `LocalDocumentField` to reference a single item on an actor, and replace `background` & `race` properties with this new field. This field has a `fallback` option so that if the item isn't found, then string is displayed (allowing existing hand-typed Race and Background values to continue displaying properly).

Used a mixin for `Item5e` and `Actor5e` that forwards socket event handler requests to their data models, allowing each type to add custom socket event handler behavior. This behavior is used in `RaceData` and `BackgroundData` to set `system.details.race` and `system.details.background` to reference any newly added items, and unset it when they are deleted. This is also used for any item marked with the `singleton` metadata property to only allow one to be created at a time.

I've also tweaked all of our current socket event handlers to ensure they return `false` if the parent handler returns `false` so we can properly prevent items from being created.

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -1,4 +1,4 @@
-import { FormulaField } from "../fields.mjs";
+import { FormulaField, LocalDocumentField } from "../fields.mjs";
 import AttributesFields from "./templates/attributes.mjs";
 import CreatureTemplate from "./templates/creature.mjs";
 import DetailsFields from "./templates/details.mjs";
@@ -26,7 +26,7 @@ import TraitsFields from "./templates/traits.mjs";
  * @property {number} attributes.exhaustion               Number of levels of exhaustion.
  * @property {number} attributes.inspiration              Does this character have inspiration?
  * @property {object} details
- * @property {string} details.background                  Name of character's background.
+ * @property {Item5e|string} details.background           Character's background item or name.
  * @property {string} details.originalClass               ID of first class taken by character.
  * @property {XPData} details.xp                          Experience points gained.
  * @property {number} details.xp.value                    Total experience points earned.
@@ -91,7 +91,9 @@ export default class CharacterData extends CreatureTemplate {
       details: new foundry.data.fields.SchemaField({
         ...DetailsFields.common,
         ...DetailsFields.creature,
-        background: new foundry.data.fields.StringField({required: true, label: "DND5E.Background"}),
+        background: new LocalDocumentField(foundry.documents.BaseItem, {
+          required: true, fallback: true, label: "DND5E.Background"
+        }),
         originalClass: new foundry.data.fields.StringField({required: true, label: "DND5E.ClassOriginal"}),
         xp: new foundry.data.fields.SchemaField({
           value: new foundry.data.fields.NumberField({
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -127,19 +127,6 @@ export default class CharacterData extends CreatureTemplate {
     super._migrateData(source);
     AttributesFields._migrateInitiative(source.attributes);
   }
-
-  /* -------------------------------------------- */
-
-  prepareDerivedData() {
-    // Populate the background field with background if background item exists but is not set
-    if ( !(this.details.background instanceof dnd5e.documents.Item5e) ) {
-      const background = this.parent.itemTypes.background?.[0];
-      if ( background ) Object.defineProperty(this.details, "background", {
-        get() { return background; },
-        enumerable: false
-      });
-    }
-  }
 }
 
 /* -------------------------------------------- */
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -127,6 +127,19 @@ export default class CharacterData extends CreatureTemplate {
     super._migrateData(source);
     AttributesFields._migrateInitiative(source.attributes);
   }
+
+  /* -------------------------------------------- */
+
+  prepareDerivedData() {
+    // Populate the background field with background if background item exists but is not set
+    if ( !(this.details.background instanceof dnd5e.documents.Item5e) ) {
+      const background = this.parent.itemTypes.background?.[0];
+      if ( background ) Object.defineProperty(this.details, "background", {
+        get() { return background; },
+        enumerable: false
+      });
+    }
+  }
 }
 
 /* -------------------------------------------- */
--- a/module/data/fields.mjs
+++ b/module/data/fields.mjs
@@ -151,6 +151,86 @@ export class IdentifierField extends foundry.data.fields.StringField {
 /* -------------------------------------------- */
 
 /**
+ * @typedef {StringFieldOptions} LocalDocumentFieldOptions
+ * @property {boolean} [fallback=false]  Display the string value if no matching item is found.
+ */
+
+/**
+ * A mirror of ForeignDocumentField that references a Document embedded within this Document.
+ *
+ * @param {typeof Document} model              The local DataModel class definition which this field should link to.
+ * @param {LocalDocumentFieldOptions} options  Options which configure the behavior of the field.
+ */
+export class LocalDocumentField extends foundry.data.fields.DocumentIdField {
+  constructor(model, options={}) {
+    if ( !foundry.utils.isSubclass(model, foundry.abstract.DataModel) ) {
+      throw new Error("A ForeignDocumentField must specify a DataModel subclass as its type");
+    }
+
+    super(options);
+    this.model = model;
+  }
+
+  /* -------------------------------------------- */
+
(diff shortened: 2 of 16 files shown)


module/data/fields.mjs
```diff
@@ -150,6 +150,86 @@ export class IdentifierField extends foundry.data.fields.StringField {
...
+  /* -------------------------------------------- */
+
+  _validateType(value) {
+    if ( !this.options.fallback ) super._validateType(value);
+  }
+
+  /* -------------------------------------------- */
+
+  initialize(value, model, options={}) {
+    if ( this.idOnly ) return value;
```
Fyorl, line 210: I think this somewhat violates the semantics of the `DocumentIdField`. If this is a field with `fallback: false`, then the value returned in the `idOnly` case might be some arbitrary string and not a document ID. We can't check to see if the document exists first, since `idOnly` can be expected to return an ID even if the document doesn't exist. I think we might have to check `isValidId` here.
```suggestion
    if ( this.idOnly ) return this.options.fallback || foundry.data.validators.isValidId(value) ? value : null;
```

module/data/fields.mjs
```diff
@@ -150,6 +150,86 @@ export class IdentifierField extends foundry.data.fields.StringField {
...
+ * @property {boolean} [fallback=false]  Display the string value if no matching item is found.
+ */
+
+/**
+ * A mirror of ForeignDocumentField that references a Document embedded within this Document.
+ *
+ * @param {typeof Document} model              The local DataModel class definition which this field should link to.
+ * @param {LocalDocumentFieldOptions} options  Options which configure the behavior of the field.
+ */
+export class LocalDocumentField extends foundry.data.fields.DocumentIdField {
```
Fyorl, line 164: We're missing all our `@inheritDoc` and `@override` annotations in this class.

module/documents/mixin.mjs
```diff
@@ -0,0 +1,54 @@
+/**
+ * Mixin used to share some logic between Actor & Item documents.
+ * @type {function(Class)}
+ * @mixin
+ */
+export const DocumentMixin = Base => class extends Base {
```
Fyorl, line 6: Can we call this something like `SystemDocumentMixin`?

module/documents/mixin.mjs
```diff
@@ -0,0 +1,54 @@
...
+    super._onUpdate(changed, options, userId);
+    this.system._onUpdate?.(changed, options, userId);
+  }
+
+  /* -------------------------------------------- */
+
+  _onDelete(options, userId) {
+    super._onDelete(options, userId);
+    this.system._onDelete?.(options, userId);
+  }
```
Fyorl, line 53: Missing JSDoc for these. Probably best if we re-annotate the parameter types and also maybe put in a `@see {ClientDocumentMixin#_preCreate}`, etc. They should all also be `@protected`.

module/documents/item.mjs
```diff
@@ -1980,7 +1981,8 @@ export default class Item5e extends Item {
 
   /** @inheritdoc */
   async _preCreate(data, options, user) {
-    await super._preCreate(data, options, user);
+    let allowed = await super._preCreate(data, options, user);
+    if ( allowed === false ) return allowed;
```
Fyorl, line 1985: We're not re-assigning this anywhere, nor returning it. That might change later but for now we should inline it.

```suggestion
    if ( (await super._preCreate(data, options, user)) === false ) return false;
```

module/documents/item.mjs
```diff
@@ -2025,7 +2027,8 @@ export default class Item5e extends Item {
 
   /** @inheritdoc */
   async _preUpdate(changed, options, user) {
-    await super._preUpdate(changed, options, user);
+    let allowed = await super._preUpdate(changed, options, user);
+    if ( allowed === false ) return allowed;
```
Fyorl, line 2031:
```suggestion
    if ( (await super._preUpdate(changed, options, user)) === false ) return false;
```

module/documents/actor/actor.mjs
```diff
@@ -838,7 +839,9 @@ export default class Actor5e extends Actor {
 
   /** @inheritdoc */
   async _preCreate(data, options, user) {
-    await super._preCreate(data, options, user);
+    let allowed = await super._preCreate(data, options, user);
+    if ( allowed === false ) return allowed;
```
Fyorl, line 843:
```suggestion
    let allowed = await super._preCreate(data, options, user);
    if ( (await super._preCreate(data, options, user)) === false ) return false;
```

module/documents/actor/actor.mjs
```diff
@@ -859,7 +862,8 @@ export default class Actor5e extends Actor {
 
   /** @inheritdoc */
   async _preUpdate(changed, options, user) {
-    await super._preUpdate(changed, options, user);
+    let allowed = await super._preUpdate(changed, options, user);
+    if ( allowed === false ) return allowed;
```
Fyorl, line 866:
```suggestion
    if ( (await super._preUpdate(changed, options, user)) === false ) return false;
```

module/data/item/background.mjs
```diff
@@ -15,4 +15,27 @@ export default class BackgroundData extends SystemDataModel.mixin(ItemDescriptio
...
+    if ( (game.user.id !== userId) || this.parent.actor?.type !== "character" ) return;
+    this.parent.actor.update({"system.details.background": this.parent.id});
+  }
+
+  /* -------------------------------------------- */
+
+  async _preDelete(options, user) {
+    if ( this.parent.actor?.type !== "character" ) return;
+    await this.parent.actor.update({"system.details.background": null});
+  }
```
Fyorl, line 40: Missing JSDoc.

module/data/item/race.mjs
```diff
@@ -67,4 +74,20 @@ export default class RaceData extends SystemDataModel.mixin(ItemDescriptionTempl
...
+    if ( (game.user.id !== userId) || this.parent.actor?.type !== "character" ) return;
+    this.parent.actor.update({"system.details.race": this.parent.id});
+  }
+
+  /* -------------------------------------------- */
+
+  async _preDelete(options, user) {
+    if ( this.parent.actor?.type !== "character" ) return;
+    await this.parent.actor.update({"system.details.race": null});
+  }
```
Fyorl, line 92: Missing JSDoc.

Fyorl (changes requested):

arbron:
@Fyorl I added a bit of data preparation code to ensure that background labels from before the reference was added continue working displaying properly.

module/data/item/race.mjs
Fyorl: Seems this is still missing JSDoc.

module/data/actor/character.mjs
```diff
@@ -125,6 +127,19 @@ export default class CharacterData extends CreatureTemplate {
...
+  /* -------------------------------------------- */
+
+  prepareDerivedData() {
+    // Populate the background field with background if background item exists but is not set
+    if ( !(this.details.background instanceof dnd5e.documents.Item5e) ) {
+      const background = this.parent.itemTypes.background?.[0];
+      if ( background ) Object.defineProperty(this.details, "background", {
+        get() { return background; },
+        enumerable: false
+      });
```
Fyorl, line 140:
```suggestion
      if ( background ) Object.defineProperty(this.details, "background", {
        value: background,
        enumerable: false
      });
```

This would still miss the `toString` handler that we define on real `LocalDocumentField`s. I'm leaning towards just doing a system migration to convert this field rather than using `prepareDerivedData`. It should be quite simple and means we don't have to worry about these 'unmigrated' values lurking in any given Actor.

Fyorl (changes requested):

module/data/actor/character.mjs
arbron: Okay, I've swapped over to full migration.

module/migration.mjs
```diff
@@ -267,6 +267,8 @@ export const migrateActorData = function(actor, migrationData) {
     const itemData = i instanceof CONFIG.Item.documentClass ? i.toObject() : i;
     let itemUpdate = migrateItemData(itemData, migrationData);
 
+    if ( itemData.type === "background" ) updateData["system.details.background"] = itemData._id;
```
Fyorl, line 270: Let's make sure this is idempotent and we are not unnecessarily performing an update.

```suggestion
    if ( (itemData.type === "background") && (actor.system?.details?.background !== itemData._id) ) {
      updateData["system.details.background"] = itemData._id;
    }
```

Fyorl (changes requested):

Fyorl (approved):
LGTM, thank you Jeff

Issue #2220, opened 2023-02-24
Advancement Type: Size


arbron:
This advancement type allows a race to set a character's ~~type, movement, senses, and~~ size.

## Level Up Interface

~~This interface will display the values for the race being added.~~ If the race offers a choice of size, that option will be presented.

## Configuration

~~The `type`, `movement`, and `senses` object are the same as are presented in actor data, and are `null` if they are not set (e.g. Humans don't get darkvision, so `senses` will be `null` for them).~~

The `size` property is an set of size choices.

```javascript
{
  "size": ["med"]
}
```

## Value

Stores the size choice selected as well as the former value of any other properties that were replaced so they can be accurately reversed.

ccjmk:
I find the name Race Properties slightly deceiving as some classes also set movement or senses, but I am at a loss with suggesting a better name. Character Properties sounds a little too generic to me, swinging too far on the opposite direction..

kaelad02:
If you're still taking name suggestions, maybe Racial Traits since the feature sounds similar to #1405

emwhalen:
Perhaps what you're discussing is Body Properties or Physical Properties. I have no strong feelings about it.

PR #2542, opened 2023-11-01
[#2220] Add SizeAdvancement

arbron:


less/advancement.less
```diff
@@ -205,6 +205,23 @@
...
+    .trait-list {
+      display: flex;
+      flex-wrap: wrap;
+      gap: 0.25rem;
+      padding: 0;
+      list-style: none;
+
+      li {
+        flex: 1 1 40%;
+      }
```
Fyorl, line 221: CSS grid might make more sense here rather than using a flex-basis of 40% with flex-wrap.

```suggestion
      display: grid;
      grid-template-columns: 1fr 1fr;
      gap: 0.25rem;
      padding: 0;
      list-style: none;
```

module/applications/advancement/size-config.mjs
```diff
@@ -0,0 +1,39 @@
...
+ * Configuration application for size advancement.
+ */
+export default class SizeConfig extends AdvancementConfig {
+
+  /** @inheritdoc */
+  static get defaultOptions() {
+    return foundry.utils.mergeObject(super.defaultOptions, {
+      classes: ["dnd5e", "advancement", "size"],
+      dragDrop: [{ dropSelector: ".drop-target" }],
+      dropKeyPath: "items",
```
Fyorl, line 14: Does this config accept drops?

module/documents/advancement/size.mjs
```diff
@@ -0,0 +1,82 @@
...
+    });
+  }
+
+  /* -------------------------------------------- */
+  /*  Instance Properties                         */
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  get levels() {
+    return [0];
```
Fyorl, line 36: I thought this wouldn't apply to advancements granted by non-class and non-subclass items?

module/documents/advancement/size.mjs
```diff
@@ -0,0 +1,82 @@
...
+    return sizes.map(s => `<span class="tag">${CONFIG.DND5E.actorSizes[s]}</span>`).join("");
+  }
+
+  /* -------------------------------------------- */
+  /*  Editing Methods                             */
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  static availableForItem(item) {
+    return !item.advancement.byType.Race?.length;
```
Fyorl, line 55: Should this be:
```suggestion
    return !item.advancement.byType.Size?.length;
```

templates/advancement/size-config.hbs
```diff
@@ -0,0 +1,10 @@
+<form autocomplete="off">
+  {{> "dnd5e.advancement-controls"}}
+
+  <div class="form-group">
+    <label>{{localize "DND5E.AdvancementHint"}}</label>
+    <textarea name="configuration.hint">{{configuration.hint}}</textarea>
+  </div>
```
Fyorl, line 7: I saw in your screenshot that this looked quite cramped. It would be nicer to have the textarea able to fill the full width I think.

templates/advancement/size-flow.hbs
```diff
@@ -0,0 +1,9 @@
+<form id="{{appId}}" data-level="{{level}}" data-id="{{advancement.id}}" data-type="{{type}}">
+  <h3>{{{this.title}}}</h3>
+
+  <p>{{advancement.configuration.hint}}</p>
+
+  <select name="size">
+    {{selectOptions sizes selected=selectedSize}}
+  </select>
```
Fyorl, line 8: I think it's a little confusing to be presented with a drop-down that has only one option. A user might reasonably think there was some bug and that they were supposed to have a choice. One option that might work is to only show the drop-down if `sizes.length > 1`, and just display the hint otherwise. In your Elf example, the hint very clearly tells the user that their size is Medium. If the hint text is omitted, I think we can quite easily auto-generate some hint text like we do with traits that says something like 'Your size is {size}' in the case of `sizes.length < 2` and 'Choose one of the available options for your size' if `sizes.length > 1`.

Fyorl (changes requested):

module/documents/advancement/size.mjs
arbron: This is still relevant, it just refers to character level rather than class level. `0` here indicates that this should be applied regardless of current character level.

templates/advancement/size-config.hbs
arbron: Screenshot 2023-11-06 at 10 19 12

Fyorl (approved):
Good stuff, thanks Jeff

PR #2546, opened 2023-11-01
[#1405] Migrate existing class saves/skills data to TraitAdvancement

arbron:
Migrates existing entries from the `saves` and `skills` sections on class items into `TraitAdvancement`.


module/data/item/class.mjs
```diff
@@ -94,4 +78,48 @@ export default class ClassData extends SystemDataModel.mixin(ItemDescriptionTemp
...
+   * Migrate the class's saves & skills into TraitAdvancements.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static #migrateTraitAdvancement(source) {
+    source.advancement ??= [];
+    if ( source.advancement.find(a => a.type === "Trait") ) return;
+
+    if ( source.saves?.length ) {
+      const savesData = {
+        type: "Trait",
```
Fyorl, line 94: Do these traits need to be given `_id`s?

Fyorl (changes requested):
Some additional considerations:
1. We should consider doing the trait migration as a full migration to ensure the data gets persisted. In-memory migrations are fine for changing names or types, but since we're writing full advancement objects I'd feel more comfortable doing it as a full migration.
2. I think we can also strip out some code related to configuring skills and saves from `ItemSheet5e`, and potentially also trim some functionality from `TraitSelector`, though I'm less sure about that.

module/data/item/class.mjs
```diff
@@ -94,4 +78,55 @@ export default class ClassData extends SystemDataModel.mixin(ItemDescriptionTemp
       ability: ""
     };
   }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Migrate the class's saves & skills into TraitAdvancements.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   * @private
```
Fyorl, line 87: `@protected` or `@internal`

module/data/item/class.mjs
```diff
@@ -94,4 +78,55 @@ export default class ClassData extends SystemDataModel.mixin(ItemDescriptionTemp
...
+        skillsData.value = {
+          chosen: system.skills.value.map(t => `skills:${t}`)
+        };
+      }
+      system.advancement.push(new TraitAdvancement(skillsData).toObject());
+      delete system.skills;
+      needsMigration = true;
+    }
+
+    if ( needsMigration ) foundry.utils.setProperty(source, "flags.dnd5e.needsMigration", true);
```
Fyorl, line 130: `needsMigration` seems a bit too general to me. Can we call it something like `persistSourceMigration` instead? That goes for the naming in `migration.mjs` as well. I'm hoping to make it clear what we're doing in those cases.

module/migration.mjs
```diff
@@ -13,11 +13,13 @@ export const migrateWorld = async function() {
...
     try {
+      const flags = { needsMigration: false };
       const source = valid ? actor.toObject() : game.data.actors.find(a => a._id === actor.id);
-      const updateData = migrateActorData(source, migrationData);
-      if ( !foundry.utils.isEmpty(updateData) ) {
+      const updateData = migrateActorData(source, migrationData, flags);
+      if ( !foundry.utils.isEmpty(updateData) || flags.needsMigration ) {
         console.log(`Migrating Actor document ${actor.name}`);
-        await actor.update(updateData, {enforceTypes: false, diff: valid});
+        const update = foundry.utils.isEmpty(updateData) ? source : updateData;
```
Fyorl, line 21: The full source data needs to take priority in the 'needsMigration' case, otherwise we might not persist the source migration in cases where there's both a source migration and a non-source migration.

```suggestion
        if ( flags.needsMigration ) foundry.utils.mergeObject(source, updateData);
        const update = flags.needsMigration ? source : updateData;
```

module/migration.mjs
```diff
@@ -30,11 +32,14 @@ export const migrateWorld = async function() {
...
+      const flags = { needsMigration: false };
       const source = valid ? item.toObject() : game.data.items.find(i => i._id === item.id);
-      const updateData = migrateItemData(source, migrationData);
-      if ( !foundry.utils.isEmpty(updateData) ) {
+      const updateData = migrateItemData(source, migrationData, flags);
+      if ( !foundry.utils.isEmpty(updateData) || flags.needsMigration ) {
         console.log(`Migrating Item document ${item.name}`);
-        await item.update(updateData, {enforceTypes: false, diff: valid});
+        const update = foundry.utils.isEmpty(updateData) ? source : updateData;
+        update["flags.dnd5e.-=needsMigration"] = null;
```
Fyorl, line 41: Maybe it would be a little neater if this was unset at the point `flags.needsMigration` is set? So if we amend `migrateItemData` to:

```js
if ( foundry.utils.getProperty(item, "flags.dnd5e.needsMigration") ) {
  flags.needsMigration = true;
  updateData["flags.dnd5e.-=needsMigration"] = null;
}
```

Then perhaps we can avoid having to unset it elsewhere? Additionally it means that `updateData` will never be empty if `flags.needsMigration` is true, so we can simplify the `if` guards somewhat.

Fyorl (changes requested):
Thanks for making those changes, Jeff. I think this is a good opportunity to build up some framework for persisting source migrations that we can make more use of in the future.

Two things I want to make absolutely sure of when testing this:
* If the migration is run a second time, items do not repeat the 'persist source migration' workflow with `diff: false`.
* After the migration is performed, no items are left with `dnd5e.needsMigration` flags.

Fyorl (changes requested):
Looking good, Jeff, just a couple of things flagged. I have one issue raised in my testing that I hope you can reproduce:

1. (on 2.4.x branch) Create a new world
2. Create a new actor
3. Create a new class on that actor
4. Edit the class and give it any advancement (I used HitPointAdvancement)
5. Under its details tab, configure 2 saving throws and some skills (I used 6 skills)
6. Select 2 skills (also under the details tab)
7. Close down the world and switch to the trait-migration branch
8. Launch the world and run `migrateWorld`
9. Open the class item on the actor and select 'modify choices'
10. Advance to the skill trait flow
11. Observe no saved choices or choices to make

image

Also noticed a small display issue with the alignment of the icons in the advancement table. If you happen to know what the issue is there, feel free to make a quick fix directly. Otherwise I will take a look in due course.

image

Fyorl:
> Not sure what you're referring to. Is it because the icons are aligned at the top with the title?

Just that they seem completely flush as to be touching the border of the element above them. I think it might just be a property of these particular icons though. I think others seem to have a bit of space baked in at the top so that I didn't notice it before.

Fyorl (approved):
Was able to do some further testing after the rebase, and it looks pretty good to me, thanks

PR #2554, opened 2023-11-05
Add preHook on the "createScrollFromSpell" function

tavernaDelBeerholder:
There are at least for me a lot of cases where I want to keep the creation of spell scrolls under control.

For example, only mages can create them, or to prevent players from messing up when a spell is dragged on the sheet.

A simple preHook in the method is more than enough for any existing use case.


Fyorl (approved):

p4535992:
Sure

Issue #2548, opened 2023-11-01
Update ScaleValueAdvancement to work with backgrounds & races


arbron:
Currently `ScaleValueAdvancement` makes some assumptions about it existing on a class or subclass, but with some minor changes it could be made to work with character level and work on backgrounds & races. This could be useful for scaling things like a Dragonborn's breath weapon.

PR #2555, opened 2023-11-06
[#2548] Update ScaleValueAdvancement to work with backgrounds & races

arbron:
Minor tweaks to scale value to allow it to work with backgrounds & races based off character level, rather than class level.

### DIFF
--- a/module/applications/advancement/scale-value-config.mjs
+++ b/module/applications/advancement/scale-value-config.mjs
@@ -44,7 +44,9 @@ export default class ScaleValueConfig extends AdvancementConfig {
    */
   _prepareLevelData() {
     let lastValue = null;
-    return Array.fromRange(CONFIG.DND5E.maxLevel + 1).slice(1).reduce((obj, level) => {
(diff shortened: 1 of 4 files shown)


module/applications/advancement/scale-value-config.mjs
```diff
@@ -44,7 +44,9 @@ export default class ScaleValueConfig extends AdvancementConfig {
    */
   _prepareLevelData() {
     let lastValue = null;
-    return Array.fromRange(CONFIG.DND5E.maxLevel + 1).slice(1).reduce((obj, level) => {
+    let levels = Array.fromRange(CONFIG.DND5E.maxLevel + 1);
+    if ( ["class", "subclass"].includes(this.advancement.parent.type ) ) levels = levels.slice(1);
```
Fyorl, line 48:
```suggestion
    if ( ["class", "subclass"].includes(this.advancement.parent.type) ) levels = levels.slice(1);
```

Fyorl (approved):
LGTM, just a minor linting issue

PR #2556, opened 2023-11-06
[#342] Apply senses/movement to actor when race is added

arbron:
- Movement types and senses are applied automatically, assuming the current value on the actor isn't already higher
- ~~Creature type isn't applied yet because we don't have it displayed on the actor, I'll handle that in another PR~~
- When deleting the race, movement and senses are unapplied if the value is equal to that on the race (so we don't get rid of values if the player has changed them)
- We now prompt when removing backgrounds & races with advancement just like with classes & subclasses
- If the player unchecks "Remove advancement changes" when deleting a race, it will not unapply any of the direct changes

### DIFF
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -1403,15 +1403,12 @@ export default class ActorSheet5e extends ActorSheet {
     if ( !game.settings.get("dnd5e", "disableAdvancements") ) {
       const manager = AdvancementManager.forDeletedItem(this.actor, item.id);
       if ( manager.steps.length ) {
-        if ( ["class", "subclass"].includes(item.type) ) {
-          try {
-            const shouldRemoveAdvancements = await AdvancementConfirmationDialog.forDelete(item);
-            if ( shouldRemoveAdvancements ) return manager.render(true);
-          } catch(err) {
-            return;
-          }
-        } else {
-          return manager.render(true);
+        try {
+          const shouldRemoveAdvancements = await AdvancementConfirmationDialog.forDelete(item);
+          if ( shouldRemoveAdvancements ) return manager.render(true);
+          else return item.delete({ shouldRemoveAdvancements });
+        } catch(err) {
+          return;
         }
       }
     }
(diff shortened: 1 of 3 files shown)


module/applications/actor/base-sheet.mjs
```diff
@@ -1403,15 +1403,12 @@ export default class ActorSheet5e extends ActorSheet {
...
-            if ( shouldRemoveAdvancements ) return manager.render(true);
-          } catch(err) {
-            return;
-          }
-        } else {
-          return manager.render(true);
+        try {
+          const shouldRemoveAdvancements = await AdvancementConfirmationDialog.forDelete(item);
+          if ( shouldRemoveAdvancements ) return manager.render(true);
+          else return item.delete({ shouldRemoveAdvancements });
```
Fyorl, line 1409: Unnecessary `else`.

```suggestion
          return item.delete({ shouldRemoveAdvancements });
```

module/data/item/race.mjs
```diff
@@ -89,7 +89,31 @@ export default class RaceData extends SystemDataModel.mixin(ItemDescriptionTempl
...
+    updates["system.attributes.movement.units"] = this.movement.units;
+
+    for ( const key of Object.keys(CONFIG.DND5E.senses) ) {
+      if ( this.senses[key] > attributes.senses[key] ) {
+        updates[`system.attributes.senses.${key}`] = this.senses[key];
+      }
+    }
+    if ( this.senses.special ) {
+      updates[system.attributes.senses.special] = attributes.senses.special
+        ? `${attributes.senses.special}:${this.senses.special}` : this.senses.special;
```
Fyorl, line 110:
```suggestion
      updates[system.attributes.senses.special] = [attributes.senses.special, this.senses.special].filterJoin(":");
```

Fyorl (changes requested):

arbron:
@Fyorl Since creature type on PCs has been merged, I've added support for that into this PR.

module/data/actor/character.mjs
```diff
@@ -97,7 +97,7 @@ export default class CharacterData extends CreatureTemplate {
           required: true, fallback: true, label: "DND5E.Background"
         }),
         originalClass: new foundry.data.fields.StringField({required: true, label: "DND5E.ClassOriginal"}),
-        type: new CreatureTypeField({ swarm: false }),
+        type: new CreatureTypeField({ swarm: false }, { initial: { value: "humanoid" }}),
```
Fyorl, line 100:
```suggestion
        type: new CreatureTypeField({ swarm: false }, { initial: { value: "humanoid" } }),
```

Fyorl (approved):

PR #2562, opened 2023-11-07
add documentation for hooks on the create spell scroll function

tavernaDelBeerholder:


Fyorl (approved):
Thanks for following-up on this.

Issue #2571, opened 2023-11-08
(2.4) Creating a measured template via the chat card does not store spell upcast information


krbz999:
- Create 1st-level spell that creates a template.
- Cast at 2nd level and create template as part of the casting.
- The created template correctly stores '2' as the spell level.
- Create another template via the chatlog.
- The new template stores '1' as the spell level despite being 2nd level.

PR #2572, opened 2023-11-08
[#2571]

krbz999:
Considered cloning the created item and scaling up its level, however this would cause damage rolls to not be upscaled, so an options object modifying the template seemed the best solution currently.


Fyorl:
Can you clarify the wording in #2571? It says that 'subsequent' creations of measured templates do not include the spell upcast information, but looking at the proposed solution here, it seems that *any* creation of a measured template via the chat card would not contain the upcast information. Is that understanding correct?

krbz999:
The level of the item is used. When you upcast, the cloned item (with higher level) is used for the creation of the measured template. If you create a template from the chat log, the base item, not the clone, is referenced, and *its* level is used.

Fyorl (approved):

Issue #2569, opened 2023-11-08
(2.4) Add a visual indication that the description is expanded or collapsed in an item's sheet


krbz999:
Currently that's near impossible to see unless any text is actually visible.

Issue #2567, opened 2023-11-08
(2.4) The accordion for descriptions has a small jumpy motion when the sheet is re-rendered


krbz999:
Expand an empty description, then change something on the sheet like its rarity. Observe tiny jump due to the css transition.

PR #2573, opened 2023-11-08
[#2567, #2569] Add status indicator to accordions, fix jump on sheet re-render

arbron:
- Adds an indicator that helps indicate whether the accordion is open or closed (#2569)
- Fixes a jump that occurs when sheet is re-rendered (#2567)
- Fixes spacing for empty accordions that are expanded

https://github.com/foundryvtt/dnd5e/assets/19979839/55c19eda-7fd7-4d11-9177-9d787e048361

### DIFF
--- a/module/applications/accordion.mjs
+++ b/module/applications/accordion.mjs
@@ -54,11 +54,13 @@ export default class Accordion {
     this.#sections = new Map();
     this.#ongoing = new Map();
     const { headingSelector, contentSelector } = this.#config;
+    let idx = 0;
     for ( const heading of root.querySelectorAll(headingSelector) ) {
       const content = heading.querySelector(contentSelector) ?? heading.parentElement.querySelector(contentSelector);
       if ( !content ) continue;
       const wrapper = document.createElement("div");
       wrapper.classList.add("accordion");
+      if ( this.#collapsed[idx] ) wrapper.classList.add("collapsed");
       heading.before(wrapper);
       wrapper.append(heading, content);
       this.#sections.set(heading, content);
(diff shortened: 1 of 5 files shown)


krbz999:
This is incredibly minor, but can we have the accordion transition time match the time for item details on the actor sheets (which are noticeably faster).

module/applications/accordion.mjs
```diff
@@ -54,18 +54,21 @@ export default class Accordion {
     this.#sections = new Map();
     this.#ongoing = new Map();
     const { headingSelector, contentSelector } = this.#config;
+    let idx = 0;
```
Fyorl, line 57: Rather than assigning an index externally, let's just convert this to a normal `for` loop instead of a `for..of`.

templates/items/parts/item-description.hbs
```diff
@@ -63,6 +63,7 @@
             {{#if editable}}
                 <h3 class="description-header">
                     {{localize "DND5E.Description"}}
+                    <i class="accordion-indicator fa-solid fa-chevron-down"></i>
```
Fyorl, line 66: It could be down to personal experience but I feel like these indicators are typically prefix the title rather than postfix it. For example, the `<details>` tag renders with the indicator first.

Fyorl (changes requested):

module/applications/accordion.mjs
arbron: The issue is that we only want to increment the index if the content is found, because otherwise an entry in `#collapsed` isn't created. If we used a `for` loop there are situations where the indexes wouldn't match.

arbron:
Screenshot 2023-11-13 at 07 31 38

module/applications/accordion.mjs
Fyorl: Ah, you're right, I missed that detail

Fyorl (approved):

Issue #2528, opened 2023-10-27
Feature Request: Support for Common / Exotic language differentiation and language subtypes.


Zanderaf:
A rework to the way the language proficiency tab is set would be nice to maybe have in the future, with something like a way to tell apart common and exotic languages such as dwarvish and abyssal like the SRD states.

Also, Primordial has different dialects or "sublanguage" like terran, ignan, etc. that would maybe be valid to put as sub-things of primordial.

Not sure if I got the point across, hoping other people can chime in since it was talked about in the D&D5e channel on the discord.

arbron:
The system supports languages subtypes now if you wanted to add them using a world script:

<img width="252" alt="Screenshot 2023-10-27 at 09 58 44" src="https://github.com/foundryvtt/dnd5e/assets/19979839/a2bb38ee-42dc-46e9-a5f2-fc6810ad29e9">
<img width="420" alt="Screenshot 2023-10-27 at 09 58 51" src="https://github.com/foundryvtt/dnd5e/assets/19979839/b754e56e-b29e-4775-8990-b29a0cfee7e9">
<img width="335" alt="Screenshot 2023-10-27 at 09 59 05" src="https://github.com/foundryvtt/dnd5e/assets/19979839/dd8d2ecf-727e-4eba-8a92-2c1efc1cda2e">

So all this would take would be a bit of reorganization of the configuration and can be done by a world script at the moment (though that doesn't mean we shouldn't add them to the system).

krbz999:
Might make sense to do something similar for damage types. It would provide an easy way to temporarily grant resistance to 'all physical damage' for example, via an effect.

PR #2575, opened 2023-11-09
[#2528] Add categories to languages

arbron:
Groups languages into categories based on how they are defined in the SRD.

Language Selector
Language Advancement

This change does modify the `languages` object in config, but it shouldn't be a breaking change for most cases because adding languages in the old fashion (`CONFIG.DND5E.languages.test = "test";`) will still work correctly (those languages will just appear at the top level.

Retrieving languages using `Trait.keyLabel` will also continue working. The only case that will lead to unexpected results is if code was retrieving languages directly (`console.log(CONFIG.DND5E.languages.common`), but that will just return empty.


module/config.mjs
```diff
@@ -1683,38 +1683,62 @@ DND5E.conditionTypes = {
...
+  },
+  exotic: {
+    label: "DND5E.LanguagesExotic",
+    children: {
+      aarakocra: "DND5E.LanguagesAarakocra",
+      abyssal: "DND5E.LanguagesAbyssal",
+      celestial: "DND5E.LanguagesCelestial",
+      deep: "DND5E.LanguagesDeepSpeech",
+      draconic: "DND5E.LanguagesDraconic",
+      druidic: "DND5E.LanguagesDruidic",
```
krbz999, line 1716: I feel like Druidic falls under the same category (that is, no category) as Thieves' Cant.

Fyorl: I think the PHB does refer to them as 'secret' languages if we wanted to give them a category. But I am ok with leaving them at the top level too. It seems pretty unlikely that you would ever want to gain proficiency with 'all secret languages' given that Thieves' Cant and Druidic are totally unrelated.

Fyorl: Oh, also, this is marked as resolved but I can still see Druidic under the exotic category in the PR.

krbz999: > It seems pretty unlikely that you would ever want to gain proficiency with 'all secret languages'

Handy shorthand if one was ever to gain temporary knowledge of 'All languages', they'd only need common, exotic, and secret.

module/documents/actor/trait.mjs
```diff
@@ -383,8 +384,11 @@ export function keyLabel(key, config={}) {
   if ( lastKey === "*" ) {
     let type;
     if ( parts.length ) {
-      const category = CONFIG.DND5E[traitConfig.configKey ?? trait]?.[parts.pop()];
-      if ( !category ) return key;
+      let category = traitData;
+      do {
+        category = (category.children ?? category)[parts.shift()];
+        if ( !category ) return key;
+      } while (parts.length);
```
Fyorl, line 391:
```suggestion
      } while ( parts.length );
```

module/documents/actor/trait.mjs
```diff
@@ -410,9 +414,20 @@ export function keyLabel(key, config={}) {
       if ( index ) return index.name;
       break;
     }
-  }
 
-  return key;
+    // Explicit categories (e.g. languages)
```
Fyorl, line 418: Not clear to me how this differs from the above tool options where we search for an exact category match or a child match? Do we still need those, or are they covered by this new `searchCategory`?

Fyorl (changes requested):

module/documents/actor/trait.mjs
arbron: For vehicle categories in tools, those are stored in separate configuration objects, while these are nested categories in a single object. The joys of years of legacy configuration data :).

Fyorl: Got it, thanks, that makes sense.

Fyorl (approved):

PR #2576, opened 2023-11-09
[#2504] Implement editor for source data

arbron:
Add a source editor for configuring the new source data on items and NPCs:

Source Editor

The editor uses a data list to display recommendations for Book and License, but it doesn't require they match something in config:

Source List

I've only added on book and license at the moment for the SRD, but it might be worth it to add other options even if we aren't going to use them for built-in system content to aid players creating stuff themselves.

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2055,6 +2055,32 @@ DND5E.advancementTypes = {
 };
 
 /* -------------------------------------------- */
+/*  Sources                                     */
+/* -------------------------------------------- */
+
+/**
+ * List of books available as sources.
+ * @enum {string}
+ */
+DND5E.sourceBooks = {
+  "SRD 5.1": "SOURCE.BOOK.SRD"
+};
+preLocalize("sourceBooks", { sort: true });
+
+/* -------------------------------------------- */
+
+/**
+ * List of licenses available with sources.
+ * @enum {string}
+ */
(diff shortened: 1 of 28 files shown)


Fyorl (changes requested):

module/config.mjs
```diff
@@ -2054,6 +2054,32 @@ DND5E.advancementTypes = {
...
+};
+preLocalize("sourceBooks", { sort: true });
+
+/* -------------------------------------------- */
+
+/**
+ * List of licenses available with sources.
+ * @enum {string}
+ */
+DND5E.sourceLicenses = {
```
Fyorl, line 2076: I've thought about it some more and it just doesn't feel like a list of licenses is something that really belongs in the system implementation. Are we going to add to this? Are modules expected to add their own licenses to this list? Since CC-BY-4.0 is the only one included by default, is that going to be taken as some sort of implicit recommendation that people use this license for their content?

I don't think it's a huge deal but, unless you have strong objections, I might suggest we remove this and replace it with a plain text field for now. We can very easily add it back in later if we need to since it would not require a migration.

Fyorl (changes requested):
Looks good to me, just one further point for discussion, then we can merge.

Fyorl (approved):

Issue #2350, opened 2023-06-19
NPC sheets can activate their items despite their actor being in a compendium


krbz999:
This is not the case for `character`-type actors.

unsoluble:
Bit more nuanced than that: `character` actors' Feature items can also be rolled.

PR #2577, opened 2023-11-09
[#2350] Require being a non-compendium actor to make any rolls on the sheet

krbz999:
Also moved the contextmenu into the 'isEditable' section since it's relevant only to editing, currently.

### DIFF
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -661,9 +661,6 @@ export default class ActorSheet5e extends ActorSheetMixin(ActorSheet) {
       // Active Effect management
       html.find(".effect-control").click(ev => ActiveEffect5e.onManageActiveEffect(ev, this.actor));
       this._disableOverriddenFields(html);
-
-      // Item Context Menu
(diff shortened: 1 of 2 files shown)


module/applications/actor/base-sheet.mjs
```diff
@@ -661,10 +661,13 @@ export default class ActorSheet5e extends ActorSheetMixin(ActorSheet) {
       // Active Effect management
       html.find(".effect-control").click(ev => ActiveEffect5e.onManageActiveEffect(ev, this.actor));
       this._disableOverriddenFields(html);
+
+      // Item Context Menu
+      new ContextMenu(html, ".item-list .item", [], {onOpen: this._onItemContext.bind(this)});
```
Fyorl, line 666: This change makes it more awkward for downstream sheet subclasses to add context menu items for observers only, for example. We should consider instead checking for `isEditable` when we are adding items to the context menu (or using the `condition` property on context menu items), in order to leave open the possibility of adding menu items in other contexts.

Fyorl (changes requested):

Fyorl (approved):

PR #2578, opened 2023-11-09
Improve scroll creation so that effects and flags from the original spell are included in the created scroll

tavernaDelBeerholder:
Thanks to tposney for highlighting this need


module/documents/item.mjs
```diff
@@ -2164,6 +2164,9 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     // Get spell data
     const itemData = (spell instanceof Item5e) ? spell.toObject() : spell;
 
+    // improve scroll creation so that effects and flags from the original spell are included in the created scroll
+    options = mergeObject(options, {flags: itemData.flags, effects: itemData.effects}, {overwrite: false});
```
krbz999, line 2168: This seems better handled down on line 2218.

p4535992: Are you sure ?  I mean you're Zhell so you're probably right, but in line 2218 I only see property management for  'name', 'image' and 'system' no ?

Fyorl: I agree with Zhell that this is better handled in the section of code where we create the spell scroll data, since we are already including information from the item data there. The item data can still be modified in the `preCreateScrollFromSpell` hook, leading to the same outcome as if they had modified `options` instead.

Fyorl:
Do we really not already include effects from the original spell in the spell scroll? I wonder why we are migrating spell scrolls in that case in #2965 ?

krbz999:
We do, as of an earlier PR of mine to fix `system.properties` (after that change in the then "2.5").

We do not transfer `flags`.

arbron:
Needed some changes so resolved this separately

PR #2581, opened 2023-11-13
Make the item description edit buttons slightly easier to hit

krbz999:
Adding some padding so you won't need to be so precise to edit one part of the description.

### DIFF
--- a/less/items.less
+++ b/less/items.less
@@ -138,6 +138,14 @@
         .description-edit {
           margin-inline-start: auto;
           opacity: 20%;
+
+          .fa-solid.fa-edit {
+            position: absolute;
+            top: 0;
+            right: 0;
+            bottom: 0;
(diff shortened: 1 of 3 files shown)


less/items.less
```diff
@@ -138,6 +138,14 @@
         .description-edit {
           margin-inline-start: auto;
           opacity: 20%;
+
+          .fa-solid.fa-edit {
+            position: absolute;
+            top: 0;
+            right: 0;
+            bottom: 0;
```
arbron, line 146: I've been trying to switch to using more logical properties where possible, so these should be `inset`:

```suggestion
            inset-block: 0;
            inset-inline-end: 0;
```

arbron (changes requested):

arbron (approved):

PR #2584, opened 2023-11-14
[#342] Change race to apply senses & movement dynamically

arbron:


module/data/item/race.mjs
```diff
@@ -113,25 +113,6 @@ export default class RaceData extends SystemDataModel.mixin(ItemDescriptionTempl
...
-      if ( this.senses[key] > attributes.senses[key] ) {
-        updates[`system.attributes.senses.${key}`] = this.senses[key];
-      }
-    }
-    if ( this.senses.special ) {
-      updates[system.attributes.senses.special] = [attributes.senses.special, this.senses.special].filterJoin(";");
-    }
-    updates["system.attributes.senses.units"] = this.senses.units;
 
     if ( this.type.value ) updates["system.details.type.value"] = this.type.value;
```
Fyorl, line 117: Should this be derived from the race item as well? We have existing movement and senses editors that we probably don't want to remove, but creature type on characters is new, so there's nothing to take away there. Perhaps it should not be editable at all, nor exist on the character data model, and should come purely from the race item? Anyone wanting to modify their creature type should do it via the race item, in the same way that editing level is done via the class item.

module/migration.mjs
```diff
@@ -522,6 +523,29 @@ function _migrateActorAC(actorData, updateData) {
...
 
+/**
+ * Migrate the actor movement & senses to replace `0` with `null`.
+ * @param {object} actorData   Actor data being migrated.
+ * @param {object} updateData  Existing updates being applied to actor. *Will be mutated.*
+ * @returns {object}           Modified version of update data.
+ * @private
+ */
+function _migrateActorMovementSenses(actorData, updateData) {
+  if ( foundry.utils.isNewerVersion("2.4.0", actorData._stats.systemVersion) ) {
```
Fyorl, line 534: Let's just guard this in case we're sent a partial actor delta or something.

```suggestion
  if ( actorData?._stats?.systemVersion && foundry.utils.isNewerVersion("2.4.0", actorData._stats.systemVersion) ) {
```

Fyorl (changes requested):
Looks pretty good to me, thanks for doing this work, Jeff. I had one point of discussion about inheriting creature type. It would be good to decide now before we release as it will be harder to change later.

arbron:
@Fyorl I've made it so type is only defined by race, and I've linked the configure button on the character sheet so it edits the value directly on the race item, so it is still easy to modify.

templates/apps/actor-type.hbs
```diff
@@ -4,6 +4,14 @@
...
         <label>{{ localize "DND5E.CreatureType" }}</label>
         <ul class="trait-list">
+            {{#if automatic}}
+              <li>
+                <label class="radio">
+                  <input type="radio" name="value" value="" {{checked type.chosen}}>
+                  {{automatic}}
+                </label>
+              </li>
+            {{/if}}
```
Fyorl, line 14: Seems this is not used? I can't find a reference to `automatic` in `ActorTypeConfig` anywhere.

Fyorl (changes requested):

templates/apps/actor-type.hbs
arbron: Whoops, leftover from another approach I was attempting.

Fyorl (approved):
Looks good, this work is coming together nicely, thanks Jeff

PR #2592, opened 2023-11-14
[#2591] Implement 'attack.flat' property

### LINKED ISSUE #2591: Allow for a flat attack bonus on an item, which ignores all modifiers except any value placed in `system.attackBonus`
Spell scrolls as well as spells from magic items have a flat modifier that does not incorporate the user's proficiency bonus or ability scores in any way but simply have, e.g., a +7 to hit. Same as we have a way to set a flat saving throw DC, we should more easily be able to set a flat attack bonus.

A current workaround is to add an attack bonus of `7 - @prof` in the `system.attackBonus` field and then also set `system.ability` to `none`, where a simple checkbox next to the attack bonus field would make this much simpler and more intuitive.

krbz999:
Closes #2591.

Coincidentally, attack rolls with spell scrolls were adding proficiency bonuses unintentionally. This fixes that, too, as a side effect.


lang/en.json
```diff
@@ -863,6 +863,8 @@
 "DND5E.ItemActivationType": "Activation Type",
 "DND5E.ItemAmmoProperties": "Ammunition Properties",
 "DND5E.ItemAttackBonus": "Attack Roll Bonus",
+"DND5E.ItemAttackFlat": "Flat Bonus",
+"DND5E.ItemAttackFlatHint": "If checked, all bonuses except the value input here will be ignored.",
```
arbron, line 867: Probably need to make this a bit more explicit as excluding proficiency and ability modifiers:

```suggestion
"DND5E.ItemAttackFlatHint": "If checked, don't add proficiency, ability modifier, or any other bonuses to attack roll except what is included here.",
```

templates/items/parts/item-action.hbs
```diff
@@ -28,7 +28,11 @@
 <div class="form-group">
     <label>{{ localize "DND5E.ItemAttackBonus" }}</label>
     <div class="form-fields">
-        <input type="text" name="system.attackBonus" value="{{system.attackBonus}}" data-formula-editor/>
+        <input type="text" name="system.attack.bonus" value="{{system.attack.bonus}}" data-formula-editor/>
```
arbron, line 31: Gotta git rid of these pesky closing slashes :/

```suggestion
        <input type="text" name="system.attack.bonus" value="{{system.attack.bonus}}" data-formula-editor>
```

arbron (changes requested):

arbron (approved):

Fyorl:
I think I agree that handling this via the drop-down is not particularly intuitive because the drop-down is labelled 'ability modifier', and this setting controls proficiency bonus and global modifiers too. I wonder, then, if there is any point in the 'None' option anymore? Or are all cases where we might want to set 'Ability Modifier: None' the same cases we we use the 'flat bonus' flag?

Fyorl:
> I suppose you could theoretically want to add proficiency but not any ability modifier to a roll. Nothing I can think of does that though.

Yeah, ok, well I suppose it doesn't really hurt to have both.

Fyorl (approved):

PR #2593, opened 2023-11-15
[#2579] Fix creature type still being editable despite being overridden by an Active Effect.

### LINKED ISSUE #2579: (2.4) The creature type config on characters and npcs does not have inputs disabled when an effect is applied to its fields

Fyorl:
- Closes #2579

### DIFF
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -652,6 +652,7 @@ h5 {
   font-size: 12px;
   line-height: 20px;
   font-weight: normal;
+  align-items: baseline;
 }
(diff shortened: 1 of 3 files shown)


dnd5e.css
```diff
@@ -652,6 +652,7 @@ h5 {
   font-size: 12px;
   line-height: 20px;
   font-weight: normal;
+  align-items: baseline;
```
Fyorl, line 655: I don't really know what changed here, but this fixed some alignment issues I spotted when looking at this.

arbron (approved):

PR #2601, opened 2023-11-17
Added missing translated label

adgranger:
The label didn't use the localization function (so we are not able to translate it in our translation modules).


p4535992:
@adgranger point the pr from your 2.4.x branch to the 2.4.x branch of the project

arbron (approved):

Issue #2596, opened 2023-11-16
[2.4] Bug: Forced Expertise Trait Advancement not working for Tools


Zanderaf:
![dwdw](https://github.com/foundryvtt/dnd5e/assets/65571170/943d9784-654f-4d75-bdae-ec3e15b99ce8)
![1111111](https://github.com/foundryvtt/dnd5e/assets/65571170/25f96232-9fb1-4976-9fb4-62dfb8901522)

Trying to create a feature that allows a player to choose a tool from the options available and gives them expertise with it, but when advancement is going through, no options show up

krbz999:
Seems like the choice only shows up when 'Mode' is 'default'.

PR #2611, opened 2023-11-17
[#2596] Fix expertise modes not working with tools on TraitAdvancement

arbron:


module/documents/actor/select-choices.mjs
```diff
@@ -63,6 +63,25 @@ export default class SelectChoices {
 
   /* -------------------------------------------- */
 
+  /**
+   * Find key and value for the provided key or key suffix.
+   * @param {string} key  Full prefixed key (e.g. `tool:art:alchemist`) or just the suffix (e.g. `alchemist`).
+   * @returns {Array}     An array with the first value being the matched key, and the second being the value.
```
Fyorl, line 69:
```suggestion
   * @returns {[string, SelectChoicesEntry]}     An array with the first value being the matched key, and the second being the value.
```

Fyorl (approved):
One JSDoc adjustment, then this can be merged.

Issue #2607, opened 2023-11-17
Contents of collapsed descriptions don't expand after editing any description


MotoMoto1234:
When you edit the identified/unidentified fields while they are collapsed, the toggle to view fails to update the text.  

[Item editor bug 2.4 dnd.webm](https://github.com/foundryvtt/dnd5e/assets/116308065/ac79a417-fb3d-404e-9669-7d7967b4c8ba)

PR #2613, opened 2023-11-17
[#2607] Fix collpased accordion sections not opening after editing description

arbron:
~~This fix at least ensures that the contents of the accordion are always visible, but it doesn't animate for the first expanding after changing another property. I would welcome thoughts as to a better fix.~~

Found a better fix that properly stores the height.

> [!Note]
> I switched from using `setTimeout` to `requestAnimationFrame` for the accordion timing here because it felt more appropriate, but that isn't part of the actual bug fix.

### DIFF
--- a/module/applications/accordion.mjs
+++ b/module/applications/accordion.mjs
@@ -48,7 +48,7 @@ export default class Accordion {
    * Augment the given markup with an accordion effect.
    * @param {HTMLElement} root  The root HTML node.
    */
-  async bind(root) {
+  bind(root) {
     const firstBind = this.#sections.size < 1;
     if ( firstBind ) this.#collapsed = [];
     this.#sections = new Map();
@@ -102,7 +102,7 @@ export default class Accordion {
    * @param {boolean} [options.animate=true]  Whether to animate the expand effect.
    * @protected
    */
-  async _onExpandSection(heading, content, { animate=true }={}) {
+  _onExpandSection(heading, content, { animate=true }={}) {
     this.#cancelOngoing(heading);
 
     if ( this.#config.collapseOthers ) {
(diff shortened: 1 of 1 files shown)


Fyorl:
I tracked this to a regression caused by the changes here: https://github.com/foundryvtt/dnd5e/pull/2573/files#diff-48f5cd8733c53432fef055aab117363d87bf159c8a3088a567bca60dbf4af90e

If we remove those lines, the issue goes away. I am having trouble reproducing the issue that those lines were meant to fix, so perhaps we could just remove them? Or perhaps address the issue they are fixing in a different way?

arbron:
> I tracked this to a regression caused by the changes here: https://github.com/foundryvtt/dnd5e/pull/2573/files#diff-48f5cd8733c53432fef055aab117363d87bf159c8a3088a567bca60dbf4af90e
> 
> If we remove those lines, the issue goes away. I am having trouble reproducing the issue that those lines were meant to fix, so perhaps we could just remove them? Or perhaps address the issue they are fixing in a different way?

If I remove those lines this is what I get:

https://github.com/foundryvtt/dnd5e/assets/19979839/19d09c6c-6cd2-4f3c-9edf-ada8645c8848

Might be a browser-specific issue.

Fyorl:
Yeah I don't experience any flash when I tested it, but it is deliberately allowing them to render at full size in order to get the correct height, so I can see why it would happen. It might be related to browsers, or just general FPS and monitor refresh rate. It's fine to solve it with some extra calls to `getBoundingClientRect` then if it avoids that issue.

Fyorl (changes requested):
We should consider if `bind`, `_onExpandSection` and `_onCollapseSection` still need to be `async`. Might be best that they're not in order to avoid implying that the sections are collapsed or expanded when the Promise resolves, because that actually only happens in `_onEnd`.

Fyorl (approved):

PR #2617, opened 2023-11-18
Scale spells with slot level if they consume slot as resource, rather than amount of consumed slots

krbz999:
This adds in some changes to resource consumption when a spell is set as innate/atwill but has a spell slot as its resource.

The change is that the spell is upcast with the level of the slot, rather than (in 2.4.0) the amount of slots consumed.

This logic should eventually be able to carry over for other item types that should scale with slots.


templates/apps/ability-use.hbs
```diff
@@ -10,7 +10,15 @@
         <label>{{ localize "DND5E.ConsumeScalingLabel" }}</label>
         <div class="form-fields">
             <select name="resourceAmount">
+                {{#if resourceArray}}
+                {{#select item.system.consume.target}}
+                {{#each resourceOptions as |l|}}
+                <option value="{{l.key}}" {{disabled (not l.canCast)}}>{{l.label}}</option>
+                {{/each}}
```
Fyorl, line 17:
```suggestion
                {{#each resourceOptions}}
                <option value="{{key}}" {{disabled (not canCast)}}>{{label}}</option>
                {{/each}}
```

module/applications/item/ability-use-dialog.mjs
```diff
@@ -60,6 +60,7 @@ export default class AbilityUseDialog extends Dialog {
       slotOptions: config.consumeSpellSlot ? this._createSpellSlotOptions(item.actor, item.system.level) : [],
       summoningOptions: this._createSummoningOptions(item),
       resourceOptions: this._createResourceOptions(item),
+      resourceArray: Array.isArray(resourceOptions),
```
Fyorl, line 63: Seems `resourceOptions` isn't defined anywhere.

module/documents/item.mjs
```diff
@@ -1165,10 +1167,12 @@ export default class Item5e extends SystemDocumentMixin(Item) {
 
     // Consume Spell Slots
     if ( config.consumeSpellSlot ) {
-      const level = this.actor?.system.spells[config.slotLevel];
+      const spellData = this.actor?.system.spells ?? {};
+      const level = spellData[config.slotLevel];
       const spells = Number(level?.value ?? 0);
       if ( spells === 0 ) {
-        const labelKey = config.slotLevel === "pact" ? "DND5E.SpellProgPact" : `DND5E.SpellLevel${this.system.level}`;
+        const isLeveled = /spell[0-9]+/.test(config.slotLevel || "");
```
Fyorl, line 1174: I know a lot of these regexes are from existing code but might as well try to clean up a bit as we go along:

```suggestion
        const isLeveled = /spell\d+/.test(config.slotLevel || "");
```

Fyorl (changes requested):

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -194,10 +194,10 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
 
     const largestSlot = Object.entries(spells).reduce((slot, [key, data]) => {
       if ( data.max === 0 ) return slot;
-      const level = parseInt(key.replace("spell", ""));
-      if ( !Number.isNaN(level) && level > slot ) return level;
+      const level = parseInt(data.level);
+      if ( !Number.isNaN(level) && (level > slot) ) return level;
```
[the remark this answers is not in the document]
krbz999: It can be undefined, though I suppose that would still mean `data.level > slot` is falsy.

Fyorl (approved):

PR #2626, opened 2023-11-18
Custom Enricher for rolling Items

### LINKED ISSUE #1546: [Feature Request] Add Roll button to inline roll links
Originally in GitLab by @MaxPat931

Include a 'Roll' button at the end of inline roll links for Items and Rolltables

image

This would allow a user to roll for an item from the item sidebar/compendium as if they were the owner which could be useful in keeping an actor's sheet more concise for things like a weapon that allows casting of spells, or a journal with links to multiple roll tables

In v10 when an actor's embedded items can also be linked, this will also allow for functionally linking their different attacks in a multiattack chat card

image

MaxPat931:
Adds a custom enricher for rolling items on your assigned/selected actor.
[[/item Heavy Crossbow]]
[[/item Claw]]{Claws}

image

Related to my comment on #1865 and would be sufficient to close #1546


MaxPat931:
A suggestion from the Discord that would be an interesting addition: Allow for passing a UUID in the enriched text (e.g. `[[/item Compendium.dnd5e.spells.Item.JLTQyqXEaJDrTXyW]]`) that would roll the item as if the Actor owned the item. 
But I dont think I have the ability to implement that myself.

MaxPat931:
@Fyorl any shot this gets included in 2.4.1?

Fyorl:
> @Fyorl any shot this gets included in 2.4.1?

I don't think so, unfortunately. Enrichers are particularly sensitive since they are quite hard to migrate later, and so we would need to deliberate carefully before adding them. Adding them into a patch release is probably not appropriate.

From a brief look at the Multiattack example in particular, it's not clear why we would prefer to go the route of looking up an item via name and then via assigned character vs. using something like a relative UUID that should resolve to the sibling Item immediately and unambiguously.

MaxPat931:
> From a brief look at the Multiattack example in particular, it's not clear why we would prefer to go the route of looking up an item via name and then via assigned character vs. using something like a relative UUID that should resolve to the sibling Item immediately and unambiguously.

Using the name just seemed easy to implement it, kind of sidestepping dragging an item to the macro bar but having the same outcome. 
It's also easy for users to setup, as the name of an item is basically a known entity, but getting a relative ID isn't difficult of course and I can see some benefit to that method, although it is less flexible.

This is also how Kaelad's Inline Roll Commands implemented item rolls as well, so I was partially taking inspiration from there.

MaxPat931:
I was able to add support for `Actor.X.Item.Y` UUIDs, this is useful for maybe having a Trap Actor with different types of traps that might exist on a scene, and using and enricher in a journal such as `[[/item Actor.p26xCjCCTQm5fRN3.Item.amUUCouL69OK1GZU]]` to trigger it without having to open or reference the Trap Actor.

I couldnt figure out how to get the Parent Actor for a Relative UUID in the enrichItem function yet, so any guidance there would be awesome, but I'll take another crack at it later

module/enrichers.mjs
```diff
@@ -315,22 +315,39 @@ async function enrichSave(config, label, options) {
...
+ * ```html
+ * <a class="roll-action" data-type="item">
+ *   <i class="fa-solid fa-dice-d20"></i> Bite
+ * </a>
+ * ```
 */
 
 async function enrichItem(config, label) {
   const givenItem = config.values.join(' ');
+  const itemUuidMatch = itemName.match(/^Actor\.\w{16}\.Item\.\w{16}$/);
```
krbz999, line 339: A uuid can also support having the name instead of an item's id.

MaxPat931: oops, I did leave in the wrong variable here, should have been `givenItem` instead of `itemName`  
But, that's not what you're talking about here, however, while a document link like `@Actor[]` will take a Name or Id, from what I can tell a UUID only takes Ids, something like `@UUID[Actor.Adult Black Dragon.Item.Bite]` does not enrich by default.  
While it would be nice to support these "pseudo-UUIDs" with names, since it would make manually creating these enrichers easier, it would offer little payoff in my opinion, since I think this method is really only useful for the Journal method mentioned earlier, and for Linked tokens.  
Unlinked tokens would be better served by the Name method or if I can get the Relative UUID method working, since the token's items could be different than their parent Actor's and the token could have effects active that the Actor wouldnt.

MaxPat931: Added support for using actor and item names in the UUID format, so now all of the following will work:
`[[/item Bite]]`
`[[/item Actor.Adult Black Dragon.Item.Bite]]`
`[[/item Actor.p26xCjCCTQm5fRN3.Item.amUUCouL69OK1GZU]]`
can even mix and match ids and names
`[[/item Actor.Adult Black Dragon.Item.amUUCouL69OK1GZU]]`
`[[/item Actor.p26xCjCCTQm5fRN3.Item.Bite]]`

image

MaxPat931:
I have added support for relative Ids.
On Use the enricher will find the appropriate Token or Actor of the embedded Chat Card, Actor Sheet, or Item Sheet to execute the roll.  
I think the main disconnect was that I couldnt do the same for creating the enriched link itself, I instead check for an Actor in the World that owns an item of the given Id. 
This means there are two niche cases (I can think of) where the enrichment would either fail or be incorrect:
- A Token has an Item its parent Actor does not, the text will not enrich in this case. (e.g. Using a macro/module to embed an item temporarily onto a token? Using the Name method would solve this problem though)
- Multiple Actors have an item with the same Id but different names, this could cause the enriched name to be incorrect, but should resolve the item use correctly (I think this could only happen if you import the same actor from a compendium, then modify the item's name?)

MaxPat931:
Minor update to support copying an item's id and using that as a relative id, so that the leading `.` is not necessary, but also allows for the name method if an item's name is 16 characters.

image

MaxPat931:
Added support for a Relative Name Ids  

I think this makes all use cases I can think of available and function as one might expect

image

[itemEnrichersInAction.webm](https://github.com/foundryvtt/dnd5e/assets/86370342/5ae2061f-7005-4b60-88b1-a9482eeb43b9)

Issue #2628, opened 2023-11-19
When a leveled spell has no value set in the scaling formula despite scaling set to 'spell level', it does not scale


krbz999:
Makes sense that it would simply fall back to the first damage part.

PR #2629, opened 2023-11-19
[#2628] Fall back to first damage part for spell scaling

krbz999:


module/documents/item.mjs
```diff
@@ -1571,9 +1571,9 @@ export default class Item5e extends SystemDocumentMixin(Item) {
         else level = this.actor.system.details.spellLevel;
         rollConfigs.forEach(c => this._scaleCantripDamage(c.parts, scaling.formula, level, rollData));
       }
-      else if ( spellLevel && (scaling.mode === "level") && scaling.formula ) {
+      else if ( spellLevel && (scaling.mode === "level") ) {
```
Fyorl, line 1574: Does it break anything if we attempt scaling `undefined` or blank? Otherwise we can just guard it.
```suggestion
      else if ( spellLevel && (scaling.mode === "level") && (scaling.formula || c.parts.length) ) {
```

Fyorl (changes requested):

Fyorl (approved):

PR #2630, opened 2023-11-19
Create data + event listeners for embedded spells

krbz999:
Introducing the internal data for creating 'embedded' spells on magical items and handling item creation and deletion.


p4535992:
I think the solution used is smart enough to allow the functionality to be extended to other items such as Features to emulate other use cases (users of the Magic Item module would be pleased).

krbz999:
> I think the solution used is smart enough to allow the functionality to be extended to other items such as Features to emulate other use cases (users of the "Magic Item" module would be pleased).

Not part of this PR. There are (imo) better ways of handling items that have multiple 'actions' than adding features onto the actor sheet (and there are different issues open for just such a thing).

Issue #2473, opened 2023-09-13
[Request] Allow for weapons with the ammunition property to also roll the saving throw and description field of ammunition used


Ackerfe:
Currently the system handles ammunition by only adding its damage formula and its attack bonus field to the roll of the weapon that is set to use that ammunition.
Ammunition items themselves though have a field in their details tab to set things like saving throws and chat messages.
It would be useful for the system to be able to roll those fields when selecting that ammunition in the weapon
For example, if I want to make a custom arrow that has a saving throw and applies disadvantage to attacks if the saving throw is failed I currently would have to manually roll the ammunition item itself rather than being able to set that ammunition in the weapon (for example with Forien's Ammo Swapper) and rolling the weapon itself.

MaxPat931:
With v4's activities, it would be useful to be able to prompt for an ammunition's activities somewhere in the roll process, maybe in the attack prompt when the ammo is selected? 
![image](https://github.com/user-attachments/assets/afb5ef0f-039e-44c0-a4d1-515241a4c02d)

Or maybe triggering the activity-choice dialog if the selected ammunition has activities?

PR #2638, opened 2023-11-21
[#2473] Add ammo description, save, and template (if applicable) to item messages

krbz999:
image

This checks the ammo of an item being used for whether it has either a saving throw or a measured template. If it has either, some buttons are appended as well as the ammo's description.


MaxPat931:
Might be nice to add the remaining quantity of the ammunition next to its name, e.g. `Explosive Arrows (6)` or maybe appended to the end of the description?

PR #2640, opened 2023-11-22
Remove Trait. in ToolIndex

### LINKED ISSUE #2639: "Trait is not defined" preventing Tool checks from being enriched
image

https://github.com/foundryvtt/dnd5e/blob/afe245632ddd40c350d4c44dda223525e1c029bb/module/enrichers.mjs#L139C32-L139C37

MaxPat931:
Closes #2639 
Allows Tool checks to be enriched

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -136,7 +136,7 @@ async function enrichCheck(config, label, options) {
   }
 
   const toolUUID = CONFIG.DND5E.enrichmentLookup.tools[config.tool];
-  const toolIndex = toolUUID ? Trait.getBaseItem(toolUUID, { indexOnly: true }) : null;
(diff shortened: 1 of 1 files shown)


arbron:
I'm confused as to what you are doing here, `Trait` is already imported at the top of the file and `getBaseIndex` isn't exposed globally, so removing `Trait` will just cause an error, not fix it.

https://github.com/foundryvtt/dnd5e/blob/afe245632ddd40c350d4c44dda223525e1c029bb/module/enrichers.mjs#L3

MaxPat931:
Sorry, seems to be an error on my part.
Is the `Trait.` in this Enrichers.js file supposed to drop off during bundling into dnd5e.mjs? It seems so, but for whatever reason the system I had installed last night that `Trait.` was still there causing the error, but on reinstall of the system it is gone, and I am not receiving this error anymore.

enrichers.js
image

dnd5e.mjs
image

PR #2647, opened 2023-11-24
Amend warning on AbilityUseDialog for pact slots

krbz999:
When a spell is set to "pact," a warning is always shown despite there being pact slots available if the actor has no regular slots of the spell's level.
This changes it so warnings are not shown for spells set to "pact" if there are slots of the spell's level _or_ there are pact slots available.

### DIFF
--- a/module/applications/item/ability-use-dialog.mjs
+++ b/module/applications/item/ability-use-dialog.mjs
@@ -245,15 +245,16 @@ export default class AbilityUseDialog extends Dialog {
   static _getAbilityUseWarnings(data) {
     const warnings = [];
     const item = data.item;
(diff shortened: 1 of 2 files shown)


Fyorl (approved):
LGTM. Can we also downgrade the display of the warning from red to yellow? I could be mistaken but I feel like it was yellow before and not having a slot of the exact level of the spell available is not the end of the world since you can always cast it with a higher level slot. I think the warning only really exists so that you don't accidentally skip through the dialog without realising that the default spell level selected has 0 slots available.

krbz999:
> LGTM.

I've done this globally rather than for just the single 'error'. This makes sense to me since none of them are actually *errors*.

Fyorl (approved):

PR #2670, opened 2023-12-10
Update creatureTypes to be object

krbz999:
Changes `CONFIG.DND5E.creatureTypes` to be nested objects, each with `label` and optionally `detectAlignment`.

I opted not to do something like `alignment: "good"`, since the official rules seem to be shying away from such things.

[Related](https://github.com/foundryvtt/dnd5e/pull/2269#discussion_r1269791096)


Fyorl (approved):
Good change, thanks

PR #2676, opened 2023-12-19
[#2168] Standardize item system type format

Ikaguia:
Draft of #2168 fix, with the changes requested by @Fyorl [here](https://github.com/foundryvtt/dnd5e/pull/2242#issuecomment-1771601242).

### DIFF
--- a/module/applications/item/ability-use-dialog.mjs
+++ b/module/applications/item/ability-use-dialog.mjs
@@ -216,7 +216,7 @@ export default class AbilityUseDialog extends Dialog {
       if ( uses.value > 1 ) str = "DND5E.AbilityUseConsumableChargeHint";
       else if ( quantity > 1 ) str = "DND5E.AbilityUseConsumableQuantityHint";
       return game.i18n.format(str, {
-        type: game.i18n.localize(`DND5E.Consumable${item.system.consumableType.capitalize()}`),
+        type: game.i18n.localize(`DND5E.Consumable${item.system.type.value.capitalize()}`),
         value: uses.value,
         quantity: quantity,
         max: uses.max,
@@ -280,7 +280,7 @@ export default class AbilityUseDialog extends Dialog {
 
     // Display warnings that the item or its resource item will be destroyed.
     if ( item.type === "consumable" ) {
-      const type = game.i18n.localize(`DND5E.Consumable${consumableType.capitalize()}`);
+      const type = game.i18n.localize(`DND5E.Consumable${consumableType.value.capitalize()}`);
       if ( this._willLowerQuantity(item) && (quantity === 1) ) {
         warnings.push(game.i18n.format("DND5E.AbilityUseConsumableDestroyHint", {type}));
       }
--- a/module/applications/item/ability-use-dialog.mjs
+++ b/module/applications/item/ability-use-dialog.mjs
@@ -245,7 +245,7 @@ export default class AbilityUseDialog extends Dialog {
   static _getAbilityUseWarnings(data) {
     const warnings = [];
     const item = data.item;
-    const { quantity, level, consume, consumableType, preparation } = item.system;
+    const { quantity, level, consume, preparation } = item.system;
     const scale = item.usageScaling;
     const levels = (preparation?.mode === "pact") ? [level, item.actor.system.spells.pact.level] : [level];
 
--- a/module/applications/item/ability-use-dialog.mjs
+++ b/module/applications/item/ability-use-dialog.mjs
@@ -280,7 +280,7 @@ export default class AbilityUseDialog extends Dialog {
 
     // Display warnings that the item or its resource item will be destroyed.
     if ( item.type === "consumable" ) {
-      const type = game.i18n.localize(`DND5E.Consumable${consumableType.value.capitalize()}`);
+      const type = game.i18n.localize(`DND5E.Consumable${item.system.type.value.capitalize()}`);
       if ( this._willLowerQuantity(item) && (quantity === 1) ) {
         warnings.push(game.i18n.format("DND5E.AbilityUseConsumableDestroyHint", {type}));
       }
--- a/module/applications/item/item-sheet.mjs
+++ b/module/applications/item/item-sheet.mjs
@@ -124,7 +124,7 @@ export default class ItemSheet5e extends ItemSheet {
       isCrewed: item.system.activation?.type === "crew",
 
       // Armor Class
-      hasDexModifier: item.isArmor && (item.system.armor?.type !== "shield"),
+      hasDexModifier: item.isArmor && (item.system.type.value !== "shield"),
 
       // Advancement
       advancement: this._getItemAdvancement(item),
@@ -235,7 +235,7 @@ export default class ItemSheet5e extends ItemSheet {
     const baseIds = CONFIG.DND5E[`${type}Ids`];
     if ( baseIds === undefined ) return {};
 
-    const typeProperty = type === "armor" ? "armor.type" : `${type}Type`;
+    const typeProperty = "type.value";
     const baseType = foundry.utils.getProperty(this.item.system, typeProperty);
 
     const items = {};
@@ -263,7 +263,7 @@ export default class ItemSheet5e extends ItemSheet {
     // Ammunition
     if ( consume.type === "ammo" ) {
       return actor.itemTypes.consumable.reduce((ammo, i) => {
-        if ( i.system.consumableType === "ammo" ) ammo[i.id] = `${i.name} (${i.system.quantity})`;
+        if ( i.system.type.value === "ammo" ) ammo[i.id] = `${i.name} (${i.system.quantity})`;
         return ammo;
       }, {});
     }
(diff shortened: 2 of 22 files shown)


module/documents/actor/actor.mjs
```diff
@@ -470,8 +470,8 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const armorTypes = new Set(Object.keys(CONFIG.DND5E.armorTypes));
     const {armors, shields} = this.itemTypes.equipment.reduce((obj, equip) => {
       const armor = equip.system.armor;
```
arbron, line 472: This isn't needed anymore.

```suggestion
```

arbron (changes requested):
Looking good!

In terms of your question of `type` vs `category`, I am inclined to stay with `type` for now because it has already been adopted by the `feat` data model.

module/applications/item/ability-use-dialog.mjs
```diff
@@ -280,7 +280,7 @@ export default class AbilityUseDialog extends Dialog {
 
     // Display warnings that the item or its resource item will be destroyed.
     if ( item.type === "consumable" ) {
-      const type = game.i18n.localize(`DND5E.Consumable${consumableType.capitalize()}`);
+      const type = game.i18n.localize(`DND5E.Consumable${consumableType.value.capitalize()}`);
```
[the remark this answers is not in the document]
Ikaguia: I think it's best to do the second option here

module/applications/item/item-sheet.mjs
```diff
@@ -235,7 +235,7 @@ export default class ItemSheet5e extends ItemSheet {
     const baseIds = CONFIG.DND5E[`${type}Ids`];
     if ( baseIds === undefined ) return {};
 
-    const typeProperty = type === "armor" ? "armor.type" : `${type}Type`;
+    const typeProperty = "type.value";
     const baseType = foundry.utils.getProperty(this.item.system, typeProperty);
```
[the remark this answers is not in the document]
Ikaguia: Would a good compromise be doing the 3rd option, adding a deprecation warning on `getBaseItem` if the old fields are detected on an item?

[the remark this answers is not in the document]
Ikaguia: Should be resolved in my latest commit: https://github.com/foundryvtt/dnd5e/pull/2676/commits/d4c2f1b700756ed5063d09102f62711a2cf45da2

module/documents/actor/trait.mjs
```diff
@@ -269,8 +268,19 @@ export function getBaseItem(identifier, { indexOnly=false, fullItem=false }={})
...
 
   // Build the extended index and return a promise for the data
-  const promise = packObject.getIndex({ fields: traitIndexFields() }).then(index => {
+  const fields = traitIndexFields();
+  const promise = packObject.getIndex({ fields }).then(index => {
     const store = index.reduce((obj, entry) => {
+      for (const field in fields) {
+        const val = foundry.utils.getProperty(entry, field);
+        if (field !== "system.type.value" && val !== undefined) {
+          foundry.utils.setProperty("system.type.value", val);
```
krbz999, line 277: Missing an argument here in setProperty.

Ikaguia: Should be fixed in my latest commit

module/applications/item/item-sheet.mjs
```diff
@@ -235,8 +235,7 @@ export default class ItemSheet5e extends ItemSheet {
     const baseIds = CONFIG.DND5E[`${type}Ids`];
     if ( baseIds === undefined ) return {};
 
-    const typeProperty = type === "armor" ? "armor.type" : `${type}Type`;
-    const baseType = foundry.utils.getProperty(this.item.system, typeProperty);
+    const baseType = this.item.system.type.value;
 
     const items = {};
     for ( const [name, id] of Object.entries(baseIds) ) {
```
arbron, line 241: Line 243 below here will also need to change now that `typeProperty` has been removed:

```javascript
if ( baseType !== baseItem?.system.type?.value ) continue;
```

module/documents/actor/trait.mjs
```diff
@@ -269,8 +268,19 @@ export function getBaseItem(identifier, { indexOnly=false, fullItem=false }={})
   if ( !packObject ) return;
 
   // Build the extended index and return a promise for the data
-  const promise = packObject.getIndex({ fields: traitIndexFields() }).then(index => {
+  const fields = traitIndexFields();
+  const promise = packObject.getIndex({ fields }).then(index => {
     const store = index.reduce((obj, entry) => {
+      for (const field in fields) {
```
arbron, line 274: Wrong iterator used here, you'll want to use `of`:

```suggestion
      for ( const field of fields ) {
```

module/documents/actor/trait.mjs
```diff
@@ -269,8 +268,19 @@ export function getBaseItem(identifier, { indexOnly=false, fullItem=false }={})
...
+        const val = foundry.utils.getProperty(entry, field);
+        if (field !== "system.type.value" && val !== undefined) {
+          foundry.utils.setProperty(entry, "system.type.value", val);
+          foundry.utils.logCompatibilityWarning(
+            `The '${field}' property has been deprecated in favor of a standardized \`system.type.value\` property.`,
+            { since: "DnD5e 2.5.0", until: "DnD5e 2.6" }
```
arbron, line 280: We usually give at least 2 versions for changes. I've also added the `once` option that was introduced in v12 to reduce these warnings in the future.

```suggestion
            { since: "DnD5e 2.5", until: "DnD5e 2.7", once: true }
```

module/documents/actor/trait.mjs
```diff
@@ -269,8 +268,19 @@ export function getBaseItem(identifier, { indexOnly=false, fullItem=false }={})
   if ( !packObject ) return;
 
   // Build the extended index and return a promise for the data
-  const promise = packObject.getIndex({ fields: traitIndexFields() }).then(index => {
+  const fields = traitIndexFields();
+  const promise = packObject.getIndex({ fields }).then(index => {
     const store = index.reduce((obj, entry) => {
+      for (const field in fields) {
+        const val = foundry.utils.getProperty(entry, field);
+        if (field !== "system.type.value" && val !== undefined) {
```
arbron, line 276:
```suggestion
        if ( (field !== "system.type.value") && (val !== undefined) ) {
```

module/documents/actor/trait.mjs
```diff
@@ -289,7 +299,7 @@ export function getBaseItem(identifier, { indexOnly=false, fullItem=false }={})
  * @protected
  */
 export function traitIndexFields() {
-  const fields = [];
+  const fields = [ "system.type.value" ];
```
arbron, line 302:
```suggestion
  const fields = ["system.type.value"];
```

arbron (changes requested):

arbron (approved):

Ikaguia:
Just as a reminder, PR does not include any changes to the compendium files

module/data/item/templates/item-type.mjs
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Convert old types into the new standard.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static #migrateType(source) {
+    if ( !source.type ) source.type = { value: "", subtype: "", baseItem: "" };
+    const oldType = source.consumableType ?? source.armor?.type ?? source.toolType ?? source.weaponType;
+    if ( ![undefined, null].includes(oldType) ) source.type.value = oldType;
+    const oldBase = source.baseItem;
+    if ( ![undefined, null].includes(oldBase) ) source.type.baseItem = oldBase;
```
Fyorl, line 45: These migrations are a bit tricky because they are also called on the update delta when an item is updated, so we must take care to avoid adding things to `source` indiscriminately.

```suggestion
    if ( foundry.utils.getType(source.type) === "Object" ) return;
    const oldType = source.consumableType ?? source.armor?.type ?? source.toolType ?? source.weaponType;
    if ( (oldType !== null) && (oldType !== undefined) ) foundry.utils.setProperty(source, "type.value", oldType);
    if ( "baseItem" in source ) foundry.utils.setProperty(source, "type.baseItem", source.baseItem);
```

Fyorl (changes requested):
These are good changes, thank you. In addition to my comment on the migration, I had two more general comments:

1. `LootData` might also potentially mixin `ItemTypeTemplate` (meaning it would not need a `defineSchema` override at all).
2. It seems weapons and consumables are assumed to always have some type. Technically the schema allowed blank types for these but our UI does not, so I'm assuming that was the intention. It might break some things if this assumption is not respected, but I'm not sure. Having default types in the `template.json` is not sufficient since these are ignored now that we use a `DataModel`. My best suggestion at the moment is to have something like `static _defaultSystemType = "simpleM";` in `WeaponData` (and something similar in `ConsumableData`). Then in `ItemTypeTemplate` we can have:

```js
value: new foundry.data.fields.StringField({required: true, blank: true, initial: this._defaultSystemType ?? "", label: "DND5E.Type"}),
```

module/data/item/templates/item-type.mjs
Ikaguia: Should be resolved in my latest commit.

Fyorl (approved):
It seems my `static _defaultSystemType` idea didn't work when I tested it. Instead I implemented a shared `ItemTypeField` which does seem to work. I kept the `ItemTypeTemplate` as a centralised place to perform the migration (and potentially any data preparation in the future). Thanks very much for your help on this.

PR #2677, opened 2023-12-20
Apply critical/fumble class only to the correct elements

### LINKED ISSUE #2671: Highlighting critical success/failure with multiple rolls applies the result of the first to all roll results
I wrote a macro to make multiple attack rolls for the Scorching Ray spell. That way I could have one chat message with all the attacks and my DM could tell me how many hit before I rolled damage. There seems to be a problem how the critical hit/failure/success highlighting works.

I saw the code that handles that, in `chat-message.mjs`, specifically the `highlightCriticalSuccessFailure` function. It grabs the first d20 roll out of the chat message and then uses it to test for `isCritical`, `isFumble`, and whether the result was `>= target`. Then that is applied to all of the resulting rolls. I'd be nice if they could be evaluated separately.

Here's the macro:

```js
const item = actor.items.getName("Scorching Ray");
const rolls = [];
for (let i = 0; i < 3; i++) {
  const roll = await item.rollAttack({ chatMessage: false });
  rolls.push(roll);
}
toMessage(rolls);

async function toMessage(rolls, messageData = {}, { rollMode, create = true } = {}) {
  for (const roll of rolls) {
    if (!roll._evaluated) await roll.evaluate({ async: true });
  }

  // Prepare chat data
  messageData = foundry.utils.mergeObject(
    {
      user: game.user.id,
      type: CONST.CHAT_MESSAGE_TYPES.ROLL,
      sound: CONFIG.sounds.dice,
    },
    messageData
  );
  messageData.rolls = rolls;

  // Either create the message or just return the chat data
  const cls = getDocumentClass("ChatMessage");
  const msg = new cls(messageData);

  // Either create or return the data
  if (create) return cls.create(msg.toObject(), { rollMode });
  else {
    if (rollMode) msg.applyRollMode(rollMode);
    return msg.toObject();
  }
}
```

And here's an example of what I saw when the first result was a crit:
image

Ikaguia:
Fixes #2671 
image

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -8,23 +8,25 @@ export function highlightCriticalSuccessFailure(message, html, data) {
   if ( !message.isRoll || !message.isContentVisible || !message.rolls.length ) return;
 
   // Highlight rolls where the first part is a d20 roll
-  for (let [index, d20Roll] of message.rolls.entries()) {
+  for ( let [index, d20Roll] of message.rolls.entries() ) {
 
     const d0 = d20Roll.dice[0];
-    if ((d0?.faces !== 20) || (d0?.values.length !== 1)) continue;
+    if ( (d0?.faces !== 20) || (d0?.values.length !== 1) ) continue;
 
     d20Roll = dnd5e.dice.D20Roll.fromRoll(d20Roll);
     const d = d20Roll.dice[0];
 
     const isModifiedRoll = ("success" in d.results[0]) || d.options.marginSuccess || d.options.marginFailure;
-    if ( isModifiedRoll ) return;
+    if ( isModifiedRoll ) continue;
 
     // Highlight successes and failures
-    if ( d20Roll.isCritical ) $(html.find(".dice-total")[index]).addClass("critical");
-    else if ( d20Roll.isFumble ) $(html.find(".dice-total")[index]).addClass("fumble");
+    const total = html.find(".dice-total")[index];
+    if ( !total ) continue;
+    if ( d20Roll.isCritical ) total.classList.add("critical");
+    else if ( d20Roll.isFumble ) total.classList.add("fumble");
     else if ( d.options.target ) {
-      if ( d20Roll.total >= d.options.target ) $(html.find(".dice-total")[index]).addClass("success");
-      else $(html.find(".dice-total")[index]).addClass("failure");
+      if ( d20Roll.total >= d.options.target ) total.classList.add("success");
+      else total.classList.add("failure");
     }
   }
 }
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -8,23 +8,24 @@ export function highlightCriticalSuccessFailure(message, html, data) {
   if ( !message.isRoll || !message.isContentVisible || !message.rolls.length ) return;
 
   // Highlight rolls where the first part is a d20 roll
-  let d20Roll = message.rolls.find(r => {
-    const d0 = r.dice[0];
-    return (d0?.faces === 20) && (d0?.values.length === 1);
-  });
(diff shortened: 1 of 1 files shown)


module/documents/chat-message.mjs
```diff
@@ -8,23 +8,24 @@ export function highlightCriticalSuccessFailure(message, html, data) {
...
-  const isModifiedRoll = ("success" in d.results[0]) || d.options.marginSuccess || d.options.marginFailure;
-  if ( isModifiedRoll ) return;
-
-  // Highlight successes and failures
-  if ( d20Roll.isCritical ) html.find(".dice-total").addClass("critical");
-  else if ( d20Roll.isFumble ) html.find(".dice-total").addClass("fumble");
-  else if ( d.options.target ) {
-    if ( d20Roll.total >= d.options.target ) html.find(".dice-total").addClass("success");
-    else html.find(".dice-total").addClass("failure");
+  for (let [index, d20Roll] of message.rolls.entries()) {
```
Fyorl, line 11:
```suggestion
  for ( let [index, d20Roll] of message.rolls.entries() ) {
```

module/documents/chat-message.mjs
```diff
@@ -8,23 +8,24 @@ export function highlightCriticalSuccessFailure(message, html, data) {
...
+
+    const d0 = d20Roll.dice[0];
+    if ((d0?.faces !== 20) || (d0?.values.length !== 1)) continue;
```
Fyorl, line 14:
```suggestion
    if ( (d0?.faces !== 20) || (d0?.values.length !== 1) ) continue;
```

module/documents/chat-message.mjs
```diff
@@ -8,23 +8,24 @@ export function highlightCriticalSuccessFailure(message, html, data) {
...
+
+    const isModifiedRoll = ("success" in d.results[0]) || d.options.marginSuccess || d.options.marginFailure;
+    if ( isModifiedRoll ) return;
+
+    // Highlight successes and failures
+    if ( d20Roll.isCritical ) $(html.find(".dice-total")[index]).addClass("critical");
+    else if ( d20Roll.isFumble ) $(html.find(".dice-total")[index]).addClass("fumble");
+    else if ( d.options.target ) {
+      if ( d20Roll.total >= d.options.target ) $(html.find(".dice-total")[index]).addClass("success");
+      else $(html.find(".dice-total")[index]).addClass("failure");
```
Fyorl, line 27: We can avoid having to re-wrap these in jQuery objects:

```suggestion
    const total = html.find(".dice-total")[index];
    if ( !total ) continue;
    if ( d20Roll.isCritical ) total.classList.add("critical");
    else if ( d20Roll.isFumble ) total.classList.add("fumble");
    else if ( d.options.target ) {
      if ( d20Roll.total >= d.options.target ) total.classList.add("success");
      else total.classList.add("failure");
```

module/documents/chat-message.mjs
```diff
@@ -8,23 +8,24 @@ export function highlightCriticalSuccessFailure(message, html, data) {
...
+  for (let [index, d20Roll] of message.rolls.entries()) {
+
+    const d0 = d20Roll.dice[0];
+    if ((d0?.faces !== 20) || (d0?.values.length !== 1)) continue;
+
+    d20Roll = dnd5e.dice.D20Roll.fromRoll(d20Roll);
+    const d = d20Roll.dice[0];
+
+    const isModifiedRoll = ("success" in d.results[0]) || d.options.marginSuccess || d.options.marginFailure;
+    if ( isModifiedRoll ) return;
```
Fyorl, line 20:
```suggestion
    if ( isModifiedRoll ) continue;
```

Fyorl (changes requested):

Fyorl (approved):

PR #2681, opened 2023-12-25
[#2636] Migrate spell `components` into `properties`, add `properties` to feature items, create `DND5E.itemProperties` and `DND5E.validProperties`.

### LINKED ISSUE #1448: [Suggestion] Add tags to features

### LINKED ISSUE #2636: Generalise `system.properties` for all item types
* Generalise the concept of `properties` for _all_ items by adding the field to the model for all item types. We can use `MappingField` and `initialKeys` to customise the properties on a per-type basis.
* Migrate spell `system.components` to be `system.properties` instead.
* Add 'Magical' and 'Concentration' properties to `feature` items, per #1448.

As per https://github.com/foundryvtt/dnd5e/pull/2356#issuecomment-1631444990

krbz999:
Does not close #2636.
Closes #1448.

This does not add properties to all item types, merely migrates spells, and adds properties to features (with 'magical' and 'concentration').

Adds `CONFIG.DND5E.itemProperties` and a second object of sets, `CONFIG.DND5E.validProperties` which shows what properties any item type has.


module/config.mjs
```diff
@@ -897,6 +897,22 @@ preLocalize("featureTypes.class.subtypes", { sort: true });
...
+ */
+DND5E.featureProperties = {
+  concentration: {
+    label: "DND5E.FeaturePropertyCon"
+  },
+  mgc: {
+    label: "DND5E.FeaturePropertyMgc"
+  }
+};
+preLocalize("featureProperties", {keys: ["label"]});
```
krbz999, line 912: I feel like this should just be a temporary solution. Maybe we should look into having a single `DND5E.itemProperties` object, and each property having an `itemTypes` array to show which items it shows up on. I looked into it only briefly, but ammo makes it more complicated, since it should depend on both item *type* (consumable) but then also item *subtype* (ammo).

arbron:
@krbz999 I've submitted a PR to convert properties to a set, rather than an object (https://github.com/foundryvtt/dnd5e/pull/2690). Assuming it is accepted, that will require a bit of retooling on this PR.

arbron (changes requested):
Be sure to check `_getItemProperties` which has a bunch of type-specific code for handling properties that can be simplified.

One issues I see with merging properties into a single `DND5E.itemProperties` in the current solution is sorting. The old method we could individually sort each config object, alphabetically sorting `weaponProperties` and not sorting `spellComponents`. We can either have a config option indicating which properties should be sorted, or just special-case spells in particular (realistically, for spells, components should always appear first in the specified order and tags should appear second in alphabetical order, but right now we don't sort any of those and it is fine).

krbz999:
> One issues I see with merging properties into a single `DND5E.itemProperties` in the current solution is sorting. The old method we could individually sort each config object, alphabetically sorting `weaponProperties` and not sorting `spellComponents`. We can either have a config option indicating which properties should be sorted, or just special-case spells in particular (realistically, for spells, components should always appear first in the specified order and tags should appear second in alphabetical order, but right now we don't sort any of those and it is fine).

I am thinking it could be some follow-up work to get rid of/deprecate the other config objects (weaponProperties, spellTags, etc), and handle some sorting method there. For now, I added `sort: true`, and there seems to be no issue.

module/config.mjs
```diff
@@ -897,6 +897,133 @@ preLocalize("featureTypes.class.subtypes", { sort: true });
...
+    label: "DND5E.Item.PropertyTwoHanded"
+  },
+  ver: {
+    label: "DND5E.Item.PropertyVersatile"
+  },
+  vocal: {
+    label: "DND5E.Item.PropertyVerbal"
+  }
+};
+preLocalize("itemProperties", {keys: ["label", "abbr"], sort: true});
```
arbron, line 979:
```suggestion
preLocalize("itemProperties", { keys: ["label", "abbr"], sort: true });
```

module/data/item/feat.mjs
```diff
@@ -12,10 +12,11 @@ import ItemTypeField from "./fields/item-type-field.mjs";
  * @mixes ActivatedEffectTemplate
  * @mixes ActionTemplate
  *
- * @property {string} requirements       Actor details required to use this feature.
- * @property {object} recharge           Details on how a feature can roll for recharges.
- * @property {number} recharge.value     Minimum number needed to roll on a d6 to recharge this feature.
- * @property {boolean} recharge.charged  Does this feature have a charge remaining?
+ * @property {Set<string>} properties                    General properties of a feature item.
+ * @property {string} requirements                  Actor details required to use this feature.
```
arbron, line 16:
```suggestion
 * @property {Set<string>} properties               General properties of a feature item.
 * @property {string} requirements                  Actor details required to use this feature.
```

module/data/item/feat.mjs
```diff
@@ -24,6 +25,7 @@ export default class FeatData extends SystemDataModel.mixin(
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
       type: new ItemTypeField({}, { label: "DND5E.ItemFeatureType" }),
+      properties: new foundry.data.fields.SetField(new foundry.data.fields.StringField(), {label: "DND5E.ItemFeatureProperties"}),
```
arbron, line 28:
```suggestion
      properties: new foundry.data.fields.SetField(new foundry.data.fields.StringField(), {
        label: "DND5E.ItemFeatureProperties"
      }),
```

module/data/item/spell.mjs
```diff
@@ -141,4 +132,19 @@ export default class SpellData extends SystemDataModel.mixin(
...
+  /* -------------------------------------------- */
+
+  /**
+   * Provide a backwards compatible getter for accessing `components`.
+   * @deprecated since v2.5.
+   * @type {Set<string>}
+   */
+  get components() {
+    foundry.utils.logCompatibilityWarning(
+      `The 'system.components' property has been deprecated in favor of a standardized \`system.properties\` property.`,
```
arbron, line 145:
```suggestion
      "The `system.components` property has been deprecated in favor of a standardized \`system.properties\` property.",
```

module/applications/item/item-sheet.mjs
```diff
@@ -135,10 +135,12 @@ export default class ItemSheet5e extends ItemSheet {
...
-      context.properties = Object.entries(validProperties).reduce((obj, [k, label]) => {
-        obj[k] = { label, selected: item.system.properties.has(k) };
+    if ( ("properties" in item.system) && (item.type in CONFIG.DND5E.validProperties) ) {
+      context.properties = CONFIG.DND5E.validProperties[item.type].reduce((obj, k) => {
+        obj[k] = {
+          label: CONFIG.DND5E.itemProperties[k].label,
+          selected: item.system.properties.has(k)
+        };
         return obj;
       }, {});
```
arbron, line 145: This should resolve the sorting issue:

```suggestion
      }, {});
      if ( item.type !== "spell" ) context.properties = sortObjectEntries(context.properties, "label");
```

(just remember to import `sortObjectEntries` with `filteredKeys` at the top)

arbron (changes requested):
Looking pretty good. For the sorting issue, adding `sort` to the CONFIG isn't the full solution because right now the order is being defined by the order in `validProperties`, rather than the `itemProperties` object. You'll have to sort the properties on sheet preparation (I've indicated that change below).

arbron (approved):

module/data/item/spell.mjs
```diff
@@ -141,4 +134,19 @@ export default class SpellData extends SystemDataModel.mixin(
...
+   * Provide a backwards compatible getter for accessing `components`.
+   * @deprecated since v2.5.
+   * @type {Set<string>}
+   */
+  get components() {
+    foundry.utils.logCompatibilityWarning(
+      "The `system.components` property has been deprecated in favor of a standardized \`system.properties\` property.",
+      { since: "DnD5e 2.5", until: "DnD5e 2.7", once: true }
+    );
+    return this.properties;
```
Fyorl, line 150: I think the caller would expect back an object of booleans if they're still using the old `components` property.

```suggestion
    return this.properties.reduce((obj, p) => obj[p] = true, {});
```

Don't forget to adjust the JSDoc if you incorporate this suggestion.

Fyorl (changes requested):
Good stuff, thanks Zhell. I only spotted one thing, but otherwise this is good to go

Fyorl (approved):

MaxPat931:
What impact might this have to the World Scripts wiki page?
I assume the Weapon Properties examples will need to be updated - https://github.com/foundryvtt/dnd5e/wiki/Modifying-Your-Game-with-Scripts#add-a-new-weapon-property

Would we want to include any additional examples when the time comes to update it?

krbz999:
Any new properties would need to be added to `CONFIG.DND5E.itemProperties`, and then they should be assigned per item type in `CONFIG.DND5E.validProperties`.

MaxPat931:
Thank you! Just gonna drop this here for future reference
Old
```js
Hooks.once("init", () => {
  CONFIG.DND5E.weaponProperties.laser = "Laser";
  CONFIG.DND5E.physicalWeaponProperties.laser = "Laser";
});
```
New
```js
Hooks.once("init", () => {
  CONFIG.DND5E.itemProperties.lsr = {
  label: "Laser",
  isPhysical: "true",
  };
  CONFIG.DND5E.validProperties.weapon.add("Laser");
});
```

PR #2688, opened 2023-12-27
[#2168] Tweak item types to allow certain fields to not be included

arbron:
Modified how the new type fields are created so that we can exclude the fields that aren't used on certain types (e.g. no `baseItem` on consumables or `subtype` on weapons). Also cleaned up a few comment and export issues.

### DIFF
--- a/module/data/item/consumable.mjs
+++ b/module/data/item/consumable.mjs
@@ -21,12 +21,13 @@ import ItemTypeField from "./fields/item-type-field.mjs";
  * @property {boolean} uses.autoDestroy  Should this item be destroyed when it runs out of uses.
  */
 export default class ConsumableData extends SystemDataModel.mixin(
-  ItemDescriptionTemplate, ItemTypeTemplate, PhysicalItemTemplate, EquippableItemTemplate, ActivatedEffectTemplate, ActionTemplate
+  ItemDescriptionTemplate, ItemTypeTemplate, PhysicalItemTemplate, EquippableItemTemplate,
+  ActivatedEffectTemplate, ActionTemplate
 ) {
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      type: new ItemTypeField({value: "potion"}, {subtype: false, baseItem: false, label: "DND5E.ItemConsumableType"}),
+      type: new ItemTypeField({value: "potion", subtype: false, baseItem: false}, {label: "DND5E.ItemConsumableType"}),
       properties: new MappingField(new foundry.data.fields.BooleanField(), {
         required: false, label: "DND5E.ItemAmmoProperties"
       }),
--- a/module/data/item/consumable.mjs
+++ b/module/data/item/consumable.mjs
@@ -26,7 +26,7 @@ export default class ConsumableData extends SystemDataModel.mixin(
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      type: new ItemTypeField({ value: "potion" }, { label: "DND5E.ItemConsumableType" }),
+      type: new ItemTypeField({value: "potion"}, {subtype: false, baseItem: false, label: "DND5E.ItemConsumableType"}),
       properties: new MappingField(new foundry.data.fields.BooleanField(), {
         required: false, label: "DND5E.ItemAmmoProperties"
       }),
--- a/module/data/item/equipment.mjs
+++ b/module/data/item/equipment.mjs
@@ -36,7 +36,7 @@ export default class EquipmentData extends SystemDataModel.mixin(
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      type: new ItemTypeField({value: "light"}, {subtype: false, label: "DND5E.ItemEquipmentType"}),
+      type: new ItemTypeField({value: "light", subtype: false}, {label: "DND5E.ItemEquipmentType"}),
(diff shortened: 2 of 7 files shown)


module/data/item/loot.mjs
```diff
@@ -16,7 +16,7 @@ export default class LootData extends SystemDataModel.mixin(
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      type: new ItemTypeField({}, {label: "DND5E.ItemLootType"})
+      type: new ItemTypeField({}, {subtype: false, baseItem: false, label: "DND5E.ItemLootType"})
```
krbz999, line 19: The system doesn't provide any subtypes for loot, but we do support it in the sheet if any were added via a module or world script.
https://github.com/foundryvtt/dnd5e/blob/master/module/applications/item/item-sheet.mjs#L150

module/data/item/fields/item-type-field.mjs
```diff
@@ -1,25 +1,29 @@
 /**
  * A field for storing Item type data.
+ *
+ * @param {object} [defaults={}]                   Options to configure this field's behavior.
+ * @param {string} [defaults.value]                An initial value for the Item's type.
+ * @param {string} [defaults.subtype]              An initial value for the Item's subtype.
+ * @param {string} [defaults.baseItem]             An initial value for the Item's baseItem.
+ * @param {DataFieldOptions} [schemaOptions]       Options forwarded to the SchemaField.
+ * @param {boolean} [schemaOptions.subtype=true]   Include the subtype field.
+ * @param {boolean} [schemaOptions.baseItem=true]  Include the baseItem field.
```
Fyorl, line 10: Rather than mix in these extra options with the `DataFieldOptions`, could we keep them as part of the first argument, which is more intended to be special options for this field in particular. Perhaps we could change the types to `{string|false}` where explicit `false` indicates the field should be omitted from the schema?

Fyorl (changes requested):
Looks good, just one comment, and let's keep loot subtypes for now, unless there's any strong objection.

module/data/item/consumable.mjs
```diff
@@ -26,7 +26,7 @@ export default class ConsumableData extends SystemDataModel.mixin(
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      type: new ItemTypeField({ value: "potion" }, { label: "DND5E.ItemConsumableType" }),
+      type: new ItemTypeField({value: "potion"}, {subtype: false, baseItem: false, label: "DND5E.ItemConsumableType"}),
```
Fyorl, line 29:
```suggestion
      type: new ItemTypeField({value: "potion", subtype: false, baseItem: false}, {label: "DND5E.ItemConsumableType"}),
```

module/data/item/equipment.mjs
```diff
@@ -36,11 +36,11 @@ export default class EquipmentData extends SystemDataModel.mixin(
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      type: new ItemTypeField({ value: "light" }, { label: "DND5E.ItemEquipmentType" }),
+      type: new ItemTypeField({value: "light"}, {subtype: false, label: "DND5E.ItemEquipmentType"}),
```
Fyorl, line 39:
```suggestion
      type: new ItemTypeField({value: "light", subtype: false}, {label: "DND5E.ItemEquipmentType"}),
```

module/data/item/loot.mjs
```diff
@@ -16,7 +16,7 @@ export default class LootData extends SystemDataModel.mixin(
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      type: new ItemTypeField({}, {label: "DND5E.ItemLootType"})
+      type: new ItemTypeField({}, {baseItem: false, label: "DND5E.ItemLootType"})
```
Fyorl, line 19:
```suggestion
      type: new ItemTypeField({baseItem: false}, {label: "DND5E.ItemLootType"})
```

module/data/item/tool.mjs
```diff
@@ -24,7 +24,7 @@ export default class ToolData extends SystemDataModel.mixin(
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      type: new ItemTypeField(),
+      type: new ItemTypeField({}, {subtype: false, label: "DND5E.ItemToolType"}),
```
Fyorl, line 27:
```suggestion
      type: new ItemTypeField({subtype: false}, {label: "DND5E.ItemToolType"}),
```

module/data/item/weapon.mjs
```diff
@@ -29,7 +29,7 @@ export default class WeaponData extends SystemDataModel.mixin(
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      type: new ItemTypeField({ value: "simpleM" }, { label: "DND5E.ItemWeaponType" }),
+      type: new ItemTypeField({value: "simpleM"}, {subtype: false, label: "DND5E.ItemWeaponType"}),
```
Fyorl, line 32:
```suggestion
      type: new ItemTypeField({value: "simpleM", subtype: false}, {label: "DND5E.ItemWeaponType"}),
```

Fyorl (changes requested):
Don't forget to move the properties into the first argument

Fyorl (approved):

Issue #2689, opened 2023-12-28
Convert `system.properties` to set


arbron:
Currently the data is stored as an object of booleans, which is a much more verbose data structure than a set but conveys the exact same information. Converting this data to a set would reduce the amount of data stored for every single item in the system and make accessing the property data a bit more logical.

Since we are already performing some breaking changes to properties with #2636, this would be a good time to sneak in this change.

arbron:
> I suppose they're not strictly equivalent. If I think about generating a set of checkboxes then I can know to even create a checkbox based on the existence of an entry in the object, and then I know the checkbox's value based on the entry's value. With a set we have to get the first piece of information externally. Are we already doing that maybe? In which case I suppose they're pretty equivalent.

We are effectively doing that by using the `initialKeys` option on `MappingField` to ensure the properties array always has the properties defined in config. This would just move that process from the data model to the sheet.

PR #2690, opened 2023-12-28
[#2689] Convert properties into a set

arbron:


module/data/item/consumable.mjs
```diff
@@ -17,25 +18,46 @@ import ItemTypeField from "./fields/item-type-field.mjs";
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Migrate the properties object into a set.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static #migratePropertiesData(source) {
+    if ( foundry.utils.getType(source.properties) !== "Object" ) return;
+    source.properties = filteredKeys(source.properties);
```
krbz999, line 58: Doesn't seem like `filteredKeys` is imported.

module/applications/item/item-sheet.mjs
```diff
@@ -428,6 +437,11 @@ export default class ItemSheet5e extends ItemSheet {
     const damage = formData.system?.damage;
     if ( damage ) damage.parts = Object.values(damage?.parts || {}).map(d => [d[0] || "", d[1] || ""]);
 
+    // Handle properties
+    if ( foundry.utils.hasProperty(formData, "system.properties") ) {
+      formData.system.properties = filteredKeys(formData.system.properties);
```
Fyorl, line 442: We can preserve any currently unconfigured properties fairly easily I think. It's not clear whether this presents an opportunity to clean up old data, or whether we should preserve it, but I think we should err on the side of preserving it for now until it becomes clear that it would be better to do otherwise.

```suggestion
      const keys = new Set(Object.keys(formData.system.properties));
      const preserve = this.object.system.properties.difference(keys);
      formData.system.properties = [...filteredKeys(formData.system.properties), ...preserve];
```

Fyorl (changes requested):
Looks good to me. I left one suggestion for potential discussion.

Fyorl (approved):

PR #2701, opened 2023-12-30
[#2698] Move getRollData to system models

### LINKED ISSUE #2698: Direct calls of `getRollData` in Actor5e to the system model
Currently we just skip over `group` actor specifically in `Actor5e`, but this fails for any other actor types added by modules.

If `getRollData` instead returned `system.getRollData` each added actor type could change behaviour any way it pleases.

Currently a new actor type is forced to have `system.attributes.prof`, otherwise they cannot even render owned items' sheets.

In fact, a whole slew of methods fail if an added actor type does not have `system.attributes`.

krbz999:
This moves the getRollData to a system method (if it has one).
Also adds `flags` to getRollData, which is a common ask.
Also moves `attributes.spellmod` to data prep where it probably should always have been. \:-)

Closes #2698

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -1,3 +1,4 @@
+import Proficiency from "../../documents/actor/proficiency.mjs";
 import { FormulaField, LocalDocumentField } from "../fields.mjs";
 import CreatureTypeField from "../shared/creature-type-field.mjs";
 import AttributesFields from "./templates/attributes.mjs";
@@ -158,6 +159,21 @@ export default class CharacterData extends CreatureTemplate {
     this.attributes.senses.units ??= raceData.senses.units;
 
     this.details.type = raceData.type;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  getRollData({ deterministic=false }={}) {
+    const data = {...this};
+    data.prof = new Proficiency(this.attributes.prof, 1);
+    if ( deterministic ) data.prof = data.prof.flat;
+    data.classes = {};
+    for ( const [identifier, cls] of Object.entries(this.parent.classes) ) {
+      data.classes[identifier] = {...cls.system};
+      if ( cls.subclass ) data.classes[identifier].subclass = cls.subclass.system;
+    }
+    return data;
   }
 }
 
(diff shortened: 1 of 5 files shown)


thatlonelybugbear:
Consider adding `effects` to the getRollData method, alongside `flags` and you are gonna make many users happy :)

module/data/actor/character.mjs
```diff
@@ -159,6 +160,21 @@ export default class CharacterData extends CreatureTemplate {
 
     this.details.type = raceData.type;
   }
+
+  /* -------------------------------------------- */
```
arbron, line 164:
```suggestion
  /* -------------------------------------------- */
  /*  Helpers                                     */
  /* -------------------------------------------- */
```

module/data/actor/npc.mjs
```diff
@@ -190,4 +191,14 @@ export default class NPCData extends CreatureTemplate {
       source.type.custom = original;
     }
   }
+
+  /* -------------------------------------------- */
```
arbron, line 195:
```suggestion
  /* -------------------------------------------- */
  /*  Helpers                                     */
  /* -------------------------------------------- */
```

module/data/actor/templates/common.mjs
```diff
@@ -111,4 +111,17 @@ export default class CommonTemplate extends SystemDataModel.mixin(CurrencyTempla
     const s = original.split(" ");
     if ( s.length > 0 ) source.attributes.movement.walk = Number.isNumeric(s[0]) ? parseInt(s[0]) : 0;
   }
+
+  /* -------------------------------------------- */
```
arbron, line 115:
```suggestion
  /* -------------------------------------------- */
  /*  Helpers                                     */
  /* -------------------------------------------- */
```

module/data/actor/vehicle.mjs
```diff
@@ -141,6 +142,16 @@ export default class VehicleData extends CommonTemplate {
       source.details.source = { custom: source.details.source };
     }
   }
+
+  /* -------------------------------------------- */
```
arbron, line 146:
```suggestion
  /* -------------------------------------------- */
  /*  Helpers                                     */
  /* -------------------------------------------- */
```

arbron (changes requested):

arbron (approved):

PR #2705, opened 2023-12-31
Add placeTemplate hooks

### LINKED ISSUE #2684: Feature Request: Expose a way to modify the `MeasuredTemplate when created from an Item
When casting a spell, like _fireball_, dnd5e runs through a series of steps to link the item (e.g., the _fireball_ spell) to the template. But it is very difficult for modules to modify the template that is created at that point.

In `item.mjs`, this line of code is particularly problematic:
```js
templates = await (dnd5e.canvas.AbilityTemplate.fromItem(item))?.drawPreview();
```

There is not, as far as I can tell, a straightforward way to modify the created template prior to its preview being drawn. I could not find anywhere that `AbilityTemplate` is exposed, so I don't think I can use libWrapper. There are hooks before and after this step, which are nice, but they are all either too early or too late. I need to take information from the item and use it to modify the template parameters. (My current workaround is to hook the `MeasuredTemplate` drawing, but that is not great because it is overinclusive and because I don't really care about the template drawing per se, I care about getting information from the item upon template creation.)

So, my request is to add something to let modules modify the template upon creation. Something like:
```js
const template = dnd5e.canvas.AbilityTemplate.fromItem(item);
Hooks.call("dnd5e.itemTemplateCreation," template, item, options);
await template.drawPreview();
```

Ikaguia:
Closes #2684

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1801,27 +1801,7 @@ export default class Item5e extends SystemDocumentMixin(Item) {
         await item.rollToolCheck({event}); break;
       case "placeTemplate":
         try {
-          /**
-           * A hook event that fires before a template is placed for an Item.
-           * @function dnd5e.prePlaceTemplate
-           * @memberof hookEvents
-           * @param {Item5e} item                     Item for which the template is being placed.
-           * @returns {boolean}                       Explicitly return false to prevent the roll from being performed.
-           */
-          if ( Hooks.call("dnd5e.prePlaceTemplate", this) === false ) break;
-          const template = dnd5e.canvas.AbilityTemplate.fromItem(item, {"flags.dnd5e.spellLevel": spellLevel});
-
-          /**
-           * A hook event that fires after a template is placed for an Item.
-           * @function dnd5e.placeTemplate
-           * @memberof hookEvents
-           * @param {Item5e} item                Item for which the template is being placed.
-           * @param {AbilityTemplate} template   The template being placed.
-           */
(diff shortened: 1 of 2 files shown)


module/documents/item.mjs
```diff
@@ -1801,7 +1801,27 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
+          /**
+           * A hook event that fires after a template is placed for an Item.
+           * @function dnd5e.placeTemplate
+           * @memberof hookEvents
+           * @param {Item5e} item                Item for which the template is being placed.
+           * @param {AbilityTemplate} template   The template being placed.
+           */
+          if ( template ) {
+            Hooks.callAll("dnd5e.placeTemplate", this, template);
+            await template.drawPreview();
```
krbz999, line 1823: Should probably use `item` rather than `this` in both hooks (as well as change the comments).

arbron:
At the moment, this hook is only called if you use the "Place Template" button in the chat card, but not if the template is placed automatically when the item is used. You'd need to define this again in `Item5e#use`. To avoid code duplication it would probably be better to place these hooks in `AbilityTemplate#fromItem`.

I'd say you should also pass in the `options` to the `prePlaceTemplate` hook in case consumers need those.

Ikaguia:
@arbron makes sense, handled those issues in my recent commits

arbron (approved):

Fyorl (approved):

PR #2709, opened 2023-12-31
[#2636] Add properties to all item types, deprecate config enums

### LINKED ISSUE #2636: Generalise `system.properties` for all item types

krbz999:
Closes #2636.


templates/items/backpack.hbs
```diff
@@ -73,7 +83,7 @@
             </div>
 
             <div class="form-group stacked">
-                <label>{{localize 'DND5E.ItemContainerProperties'}}</label>
+                <label>{{localize 'DND5E.ItemContentProperties'}}</label>
```
arbron, line 86: I think weightless contents should be merged into properties. We can add a migration to add it to the properties set if it is checked and then define a getter to return the old property.

Something like this can be defined in `prepareDerivedData` on `ContainerData`:

```javascript
Object.defineProperty(this.capacity, "weightless", {
  value: this.properties.has("weightlessContents"),
  writable: false
});
```

module/config.mjs
```diff
@@ -985,10 +985,20 @@ preLocalize("itemProperties", { keys: ["label", "abbr"], sort: true });
  * @enum {object}
  */
 DND5E.validProperties = {
+  backpack: new Set([
+    "mgc"
+  ]),
+  equipment: new Set([
```
arbron, line 991: Consumables should be able to be marked as magical also.

module/config.mjs
```diff
@@ -1634,6 +1648,7 @@ preLocalize("weaponTypes");
 /**
  * A subset of weapon properties that determine the physical characteristics of the weapon.
  * These properties are used for determining physical resistance bypasses.
+ * @deprecated
```
arbron, line 1651:
```suggestion
 * @deprecated since DnD5e 2.5, available until DnD5e 2.7
```

module/config.mjs
```diff
@@ -1647,6 +1662,7 @@ preLocalize("physicalWeaponProperties", { sort: true });
 
 /**
  * The set of weapon property flags which can exist on a weapon.
+ * @deprecated
```
arbron, line 1665:
```suggestion
 * @deprecated since DnD5e 2.5, available until DnD5e 2.7
```

arbron (changes requested):

module/config.mjs
krbz999: Not sure what you would suggest we do here. The set for `consumable` is at the bottom (should probably be alphabetical, yes), but the properties apply only if the subtype is `ammo`.

templates/items/backpack.hbs
krbz999: I may be missing how we can both define the property yet be rid of it if it is `true`.

module/config.mjs
arbron: Aw, that's why I didn't see it when I was looking at potions in particular. Specifically I was thinking it would be good to be able to mark potions as magical, but because ammo needs additional properties that is tough with this structure. Maybe display properties for all consumable types, but filter out the `isPhysical` properties unless the consumable type is ammo?

templates/items/backpack.hbs
arbron: We should be able to:
1. Remove `weightless` from the schema
2. Add a migration that checks to see if `source.capacity.weightless` is `true` and add that to `source.properties` if so, otherwise do nothing
3. Then use `defineProperty` in `prepareDerivedData` to create a fallback value (though you'd probably want to rewrite what I put above to use `get() { }` rather than `value:` so you can log a compatibility warning)

krbz999: One issue I am seeing, just doing steps 1 and 2, is that if `source.capacity.weightless === true`, then `source.properties === undefined`.

module/config.mjs
krbz999: Alternatively we could come up with a more generalized approach for subtypes and in the meantime suffer being able to make potions made of adamantine. This would require that all item types that have `system.properties` also have `system.type` in their schema.

krbz999: Afraid I'm not seeing a good solution here since `mgc` is also `isPhysical`.

templates/items/backpack.hbs
krbz999: After thinking on it some, I am agreeing it should be part of `properties`. It is entirely specific to containers and the contents that it holds.

templates/items/consumable.hbs
```diff
@@ -73,7 +73,6 @@
                 </label>
             </div>
 
-            {{#if (eq system.type.value "ammo")}}
             {{!-- Consumable Properties --}}
             <div class="form-group stacked weapon-properties">
                 <label>{{ localize "DND5E.ItemAmmoProperties" }}</label>
```
arbron, line 78: This bit still says ammunition properties even on other consumable types.

templates/items/backpack.hbs
```diff
@@ -51,6 +51,16 @@
...
 
+            {{!-- Container Properties --}}
+            <div class="form-group stacked container-properties">
+                <label>{{ localize "DND5E.ItemContainerProperties" }}</label>
+                {{#each properties}}
+                <label class="checkbox">
+                    <input type="checkbox" name="system.properties.{{@key}}" {{checked selected}}> {{ label }}
+                </label>
+                {{/each}}
+            </div>
```
arbron, line 62: Then if we don't add `weightless` to the properties, how about we still show it in this list to avoid too many large sections with just a single checkbox:

```suggestion
            <div class="form-group stacked">
                <label>{{localize 'DND5E.ItemContainerProperties'}}</label>
                {{#each properties}}
                <label class="checkbox">
                    <input type="checkbox" name="system.properties.{{@key}}" {{checked selected}}> {{ label }}
                </label>
                {{/each}}
                <label class="checkbox">
                    <input type="checkbox" name="system.capacity.weightless" {{checked system.capacity.weightless}}>
                    {{localize 'DND5E.ItemContainerWeightless'}}
                </label>
            </div>
```

arbron (changes requested):

templates/items/backpack.hbs
krbz999: I've gone ahead with an alternative that is also much more in line with how we handle these related inputs on other item sheets, merging all capacity-related fields into one row. :)

image

arbron: I'm not so sure, because "Weightless Contents" doesn't really have to do with the carrying capacity of the backpack.

Also, by just changing it to "Weightless" it isn't clear that that only applies to the contents and not the whole container.

krbz999: We may be splitting hairs but I'm not sure I agree; all three inputs are related to the contents of the container; how much, what kind, and if it weighs.

We could relabel the form group from 'Capacity' to 'Contents'.

arbron: Then we'd lose the clarity that those values represent the maximum container capacity.

Time to call in a 3rd party! @Fyorl Any thoughts?

krbz999: Lives may be at stake.

krbz999: A second idea I had was to keep it the way it originally was - the issue seemed to be that the form group for container contents had only a single checkbox. We could, if need be, come up with more properties to fill it with. I'm sure future backpack management features could make use of some.

Fyorl: Feels like it should be part of the container's `properties` to me. What's the argument against that? That it's specific to containers? Aren't most weapon properties also only specific to weapons?

krbz999: We (or at least I) do not appear able to migrate it. See here: https://github.com/foundryvtt/dnd5e/pull/2709#discussion_r1439092290

Fyorl: Do you foresee any issues with the following migration?
```js
#migrateWeightless(source) {
  if ( foundry.utils.getProperty(source, "capacity.weightless") === true ) {
    source.properties ??= [];
    source.properties.push("weightless");
  }
}
```

krbz999: I may be incorrect, but would that not cause an issue if doing `item.update({"system.capacity.weightless": true})` regardless of if the property exists?

Fyorl: Yes, it would wipe all the other properties (in this case there's only one other property to wipe). But our system code will not be performing such updates, and if there's a module doing that, then it's really no different to any number of problems an out-of-date module might cause.

If we wanted a more robust deprecation path, I suppose we could perform this update as part of `migration.mjs`, rather than an in-memory migration (with all of the associated caveats that entails).

arbron: Haven't fully tested this, but it seems like we could just do this in `migrateItemData`:

```javascript
  // Migrate container properties
  if ( (item.type === "backpack") && (foundry.utils.getProperty(item, "system.capacity.weightless") === true) ) {
    updateData["system.properties"] = foundry.utils.getProperty(item, "system.properties") ?? [];
    updateData["system.properties"].push("weightlessContents");
  }
```

arbron: Okay, that one didn't fully work, but here is a diff for a change that uses a similar system to how we handled the `TraitAdvancement` migration on classes.

[changes.diff.txt](https://github.com/foundryvtt/dnd5e/files/13814688/changes.diff.txt)

arbron (changes requested):
This migration should work:

[changes.diff.txt](https://github.com/foundryvtt/dnd5e/files/13814690/changes.diff.txt)

(also fixed an unrelated migration bug)

templates/items/backpack.hbs
Fyorl: Can you elaborate on why the solution mentioned here didn't work: https://github.com/foundryvtt/dnd5e/pull/2709#discussion_r1439942453

I think I'd prefer it over the solution presented in the diff which feels overly complicated for what should be a very simple migration that we have done many times before in `migration.mjs`.

arbron (approved):
I think this will work.

I'll do a followup PR to re-write `#migrateComponentData` in the same way as `_migrateWeightlessData` to avoid losing properties with a partial spell update.

PR #2713, opened 2024-01-04
[#2636] Modify spell component migration to avoid losing properties

arbron:
Swap spell components over to using the new properties migration system added to avoid potential data loss.

### DIFF
--- a/module/data/item/spell.mjs
+++ b/module/data/item/spell.mjs
@@ -63,13 +63,12 @@ export default class SpellData extends SystemDataModel.mixin(
   }
 
   /* -------------------------------------------- */
-  /*  Migrations                                  */
+  /*  Data Migrations                             */
   /* -------------------------------------------- */
 
   /** @inheritdoc */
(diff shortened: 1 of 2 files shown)


module/data/item/spell.mjs
```diff
@@ -79,9 +78,9 @@ export default class SpellData extends SystemDataModel.mixin(
    * Migrate the component object to be 'properties' instead.
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
-  static #migrateComponentData(source) {
-    if ( !source.components || ("properties" in source)) return;
-    source.properties = filteredKeys(source.components);
+  static _migrateComponentData(source) {
+    const components = filteredKeys(source.system.components ?? {});
```
Fyorl, line 82: It's possible `system` might not be in `source`. If only the Item's `name` is updated, for example.

```suggestion
    const components = filteredKeys(source.system?.components ?? {});
```

PR #2717, opened 2024-01-05
Add 'concentration' to scroll properties, fix createScrollFromSpell

krbz999:
Adds `concentration` as a valid item property if the consumable subtype is `scroll`, and fixes a few issues with `Item5e#createScrollFromSpell`.

### DIFF
--- a/module/data/item/consumable.mjs
+++ b/module/data/item/consumable.mjs
@@ -102,6 +102,7 @@ export default class ConsumableData extends SystemDataModel.mixin(
     if ( this.type.value === "ammo" ) Object.entries(CONFIG.DND5E.itemProperties).forEach(([k, v]) => {
       if ( v.isPhysical ) valid.add(k);
     });
+    else if ( (this.type.value === "scroll") ) valid.add("concentration");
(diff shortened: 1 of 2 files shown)


module/data/item/consumable.mjs
```diff
@@ -102,6 +102,7 @@ export default class ConsumableData extends SystemDataModel.mixin(
     if ( this.type.value === "ammo" ) Object.entries(CONFIG.DND5E.itemProperties).forEach(([k, v]) => {
       if ( v.isPhysical ) valid.add(k);
     });
+    else if ( this.type.value === "scroll" ) valid.add("concentration");
```
arbron, line 105: We should also add the other spell components other than material, because scrolls still might require verbal or somatic components.

```suggestion
    else if ( this.type.value === "scroll" ) CONFIG.DND5E.validProperties.spell
      .filter(p => p !== "material").forEach(p => valid.add(p));
```

arbron (changes requested):

arbron (approved):

PR #2720, opened 2024-01-06
Add `isConcentration` getter

krbz999:


module/documents/item.mjs
```diff
@@ -33,6 +33,19 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     return this.system.isActive ?? false;
   }
 
+  /* --------------------------------------------- */
+
+  /**
+   * Does this item require concentration?
+   * @type {boolean}
+   */
+  get isConcentration() {
```
arbron, line 42: `requiresConcentration` might be a better name.

arbron: And maybe shift it down below `get isVersatile()` so things are bit closer to alphabetical order.

arbron (approved):

PR #2722, opened 2024-01-06
[#2716] Show AC only on armor/shield/vehicle

### LINKED ISSUE #2716: When an `equipment` item is set to nothing, 'clothing', or 'trinket', the AC tag still shows on the actor sheet and in chat messages
image

krbz999:
Closes #2716.


arbron (approved):
Might also be valuable to add the AC tag to mountable weapons which also have armor in a followup PR.

PR #2723, opened 2024-01-06
Add AC tag to siege weaponry

krbz999:


module/data/item/weapon.mjs
```diff
@@ -80,7 +80,10 @@ export default class WeaponData extends SystemDataModel.mixin(
    * @type {string[]}
    */
   get chatProperties() {
-    return [CONFIG.DND5E.weaponTypes[this.type.value]];
+    return [
+      this.isMountable ? (this.parent.labels?.armor ?? null) : null,
+      CONFIG.DND5E.weaponTypes[this.type.value]
```
arbron, line 85: To match the ordering in equipment, this should probably be type then AC.

```suggestion
      CONFIG.DND5E.weaponTypes[this.type.value],
      this.isMountable ? (this.parent.labels?.armor ?? null) : null
```

arbron (changes requested):

arbron (approved):

PR #2725, opened 2024-01-07
Patch damage and healing types config, add color, label, isPhysical properties

krbz999:


arbron (changes requested):
Looks good, but pull out the colors for now to handle in a separate PR with a proposed usage. Let's just do the change to an object in this one.

Fyorl:
> Bruh.

Well, for the separate PR we can consider using the `Color` type (:

arbron (approved):

arbron:
<too-many-cooks.gif>

module/config.mjs
```diff
@@ -1186,35 +1187,54 @@ preLocalize("physicalDamageTypes", { sort: true });
 
 /**
  * Types of damage the can be caused by abilities.
- * @enum {string}
+ * @enum {{label: string, color: string}}
```
Fyorl, line 1190:
```suggestion
 * @enum {{label: string, [isPhysical]: boolean}}
```

Fyorl (approved):

PR #2727, opened 2024-01-08
Expose ItemUseConfiguration in AbilityUseDialog

krbz999:
Modules can use `dnd5e.preUseItem` to add their own properties to the configuration, but this is not be available in the `AbilityUseDialog`.

Currently we can only use a render hook, and *assume* that the key must be `true` if the item is able to do the related thing (and otherwise do nothing), but it would be good to be able to make use of the pre-use hook as well in case a user wants to programmatically skip the prompt; i.e., set it to `null` or `false`.

With this, a module can read the key in `abilityUseDialog.configuration` and set the default value accordingly (or not at all if `null`).

### DIFF
--- a/module/applications/item/ability-use-dialog.mjs
+++ b/module/applications/item/ability-use-dialog.mjs
@@ -1,13 +1,14 @@
 /**
  * A specialized Dialog subclass for ability usage.
  *
- * @param {Item5e} item                   Item that is being used.
- * @param {ItemUseConfiguration} config   The ability use configuration's values.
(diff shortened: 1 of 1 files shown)


Fyorl:
This looks fine to me, but we could probably make it non-breaking by including the item config in the Application options instead, since we are just assigning it to an instance property and not otherwise doing anything with it. Then it would be accessible via `app.options.itemUseConfiguration` or something. Does that sound reasonable?

Fyorl (approved):

PR #2736, opened 2024-01-11
Add finesse to flying snake bite attack.

### LINKED ISSUE #2735: Flying snake bite attack is missing finesse property
https://5thsrd.org/gamemaster_rules/monsters/flying_snake/
> Melee Weapon Attack:+6 to hit

Attack is calculated with -1 to hit (-3 str +2 hd). Adding the finesse property results in +6 (+4 dex + 2hd).

I've made and tested this change, and will open a PR shortly.

chuckhazard:
Fixes #2735

Add finesse property to flying snake bite attack.

### DIFF
--- a/packs/src/monsters/beast/flying-snake.json
+++ b/packs/src/monsters/beast/flying-snake.json
@@ -645,7 +645,7 @@
         "baseItem": "",
         "properties": {
           "amm": false,
-          "fin": false,
+          "fin": true,
           "fir": false,
           "foc": false,
           "hvy": false,


Fyorl:
Rather than setting this one weapon to finesse, perhaps we would be better served by implementing some of the suggestions from #2735? I'd propose that for any weapon of type 'Natural' with the attack ability modifier set to 'Default' that it should use the highest of Str/Dex automatically (min 0). Exceptions could be handled by setting an explicit ability, or by using `attack.flat` proposed in #2592.

We'd have to evaluate any disruption this might cause for pre-existing Actors using natural weapons but I'm hoping that (a) natural weapons are rare on non-NPCs, and (b) that it will produce the correct result in most cases.

Fyorl:
Should now be resolved with #5935

Issue #2744, opened 2024-01-20
Adding an Ability Score via the World Script Section on the Wiki doesn't work as intended.


TouchfuzzyGH:
There's an issue with the Wiki section "Modifying Your Game with World Scripts."  Specifically under "Add a New Ability."  If you add an ability with a short name, the CSS seems to wig out and ability boxes merge together:

![image](https://github.com/foundryvtt/dnd5e/assets/128764006/99cae861-a1a6-4d5c-b150-d55c61e3f953)

After asking about it, this was fixed by adding a styles.css and a css rule, but I feel as though this is an issue with the system, and until fixed, the guide should either include altering css, or mention that the new ability name can't be shorter than the problem-causing amount.

PR #2745, opened 2024-01-20
[#2744] Add min-width to .ability-scores .ability

etiquettestartshere:
[#2744] Prevent short ability score names from being improperly displayed by adding `display: grid` property to `.dnd5e.sheet.actor .ability-scores`

thanks to Zhell for this cleaner solution, my original idea was:
```css
.dnd5e.sheet.actor h4.box-title {
  min-width: 36px
}
```

### DIFF
--- a/dnd5e.css
+++ b/dnd5e.css
@@ -940,7 +940,6 @@ h5 {
   white-space: nowrap;
 }
 .dnd5e.sheet.actor .ability-scores {
-  display: grid;
   flex: 0 0 100px;
   align-content: flex-start;
   list-style: none;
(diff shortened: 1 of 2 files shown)


krbz999:
Tested this myself, and adding just `display: grid` seems to add some margin on the right of each ability:

image

If the `flexrow` class is removed from the same `div` in the template, it looks better.

image

etiquettestartshere:
Thanks, don't really know what i'm doing 😅 . I should have caught the gap, especially considering `min-width: 36px` on `h4.box-title` didn't cause it. Implemented your suggestions, sorry for the trouble

arbron (approved):

arbron:
Fix pushed, thanks for figuring it out

PR #2746, opened 2024-01-20
Fix armor class, allow for shield subtypes

krbz999:
The 2.5 branch broke AC due to a missing word.

This also adds the ability to add shield 'subtypes', in line with how other armor proficiency categories work. To clarify, the issue is that adding anything to `CONFIG.DND5E.shieldIds` does not allow a user to set an item to be of this newly added shield type.

### DIFF
--- a/module/applications/item/item-sheet.mjs
+++ b/module/applications/item/item-sheet.mjs
@@ -231,7 +231,8 @@ export default class ItemSheet5e extends ItemSheet {
    */
   async _getItemBaseTypes() {
     const type = this.item.type === "equipment" ? "armor" : this.item.type;
-    const baseIds = CONFIG.DND5E[`${type}Ids`];
+    const {armorIds, shieldIds} = CONFIG.DND5E;
+    const baseIds = this.item.type === "equipment" ? {...armorIds, ...shieldIds} : CONFIG.DND5E[`${type}Ids`];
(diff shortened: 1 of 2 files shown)


arbron (approved):

module/applications/item/item-sheet.mjs
```diff
@@ -231,7 +231,10 @@ export default class ItemSheet5e extends ItemSheet {
    */
   async _getItemBaseTypes() {
     const type = this.item.type === "equipment" ? "armor" : this.item.type;
```
arbron, line 233:
```suggestion
```

module/applications/item/item-sheet.mjs
```diff
@@ -231,7 +231,10 @@ export default class ItemSheet5e extends ItemSheet {
...
-    const baseIds = CONFIG.DND5E[`${type}Ids`];
+    const baseIds = this.item.type === "equipment" ? {
+      ...CONFIG.DND5E.armorIds,
+      ...CONFIG.DND5E.shieldIds
+    } : CONFIG.DND5E[`${type}Ids`];
```
arbron, line 237:
```suggestion
    } : CONFIG.DND5E[`${this.item.type}Ids`];
```

arbron (changes requested):

arbron (approved):

Issue #2747, opened 2024-01-21
Advancements cannot be used on module types


arbron:
Currently advancements define which items they can be used with in their metadata, which is immutable. This limits their ability to be used on new types defined by modules. Moving this property into `CONFIG.advancementTypes` would allow modules to configure what advancement types can be used on their new items.

PR #2748, opened 2024-01-21
[#2747] Move advancement valid item types into config so module subtypes can change it

arbron:
Changes `advancementTypes` into an object and moved `validItemTypes` from `Advancement#metadata` into the configuration object, so module custom types can define what advancement types they accept.

The deprecations should ensure everything continues working for any custom advancement types.

### DIFF
--- a/module/data/item/race.mjs
+++ b/module/data/item/race.mjs
@@ -95,8 +95,9 @@ export default class RaceData extends ItemDataModel.mixin(ItemDescriptionTemplat
       { type: "Trait", configuration: { grants: ["languages:standard:common"] } }
     ];
     this.parent.updateSource({"system.advancement": toCreate.map(c => {
-      const AdvancementClass = CONFIG.DND5E.advancementTypes[c.type];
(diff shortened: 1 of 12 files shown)


module/data/item/race.mjs
```diff
@@ -95,8 +95,8 @@ export default class RaceData extends SystemDataModel.mixin(ItemDescriptionTempl
       { type: "Trait", configuration: { grants: ["languages:standard:common"] } }
     ];
     this.parent.updateSource({"system.advancement": toCreate.map(c => {
-      const AdvancementClass = CONFIG.DND5E.advancementTypes[c.type];
-      return new AdvancementClass(c, { parent: this.parent }).toObject();
+      const cls = CONFIG.DND5E.advancementTypes[c.type].documentClass;
```
Fyorl, line 98:
```suggestion
      const config = CONFIG.DND5E.advancementTypes[c.type];
      const cls = config.documentClass ?? config;
```

PR #2757, opened 2024-01-28
Fix the v2.5 `type` migration

krbz999:
Since the embedded `type` field always exists, the source data migration was never performed.

### DIFF
--- a/module/data/item/templates/item-type.mjs
+++ b/module/data/item/templates/item-type.mjs
@@ -28,9 +28,8 @@ export default class ItemTypeTemplate extends SystemDataModel {
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
   static #migrateType(source) {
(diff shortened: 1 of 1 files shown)


Fyorl:
> Since the embedded `type` field always exists, the source data migration was never performed.

What do you mean by this?

arbron:
> What do you mean by this?

Well, one place the current migration won't work is if someone is calling `item.update({"consumableType": "blah"})` if type already exists. These changes should allow for those kind of legacy updates.

Fyorl:
> Well, one place the current migration won't work is if someone is calling `item.update({"consumableType": "blah"})` if type already exists. These changes should allow for those kind of legacy updates.

If they perform such an update, then `source` will only be the update delta, i.e. `{"consumableType": "blah"}`, so there will be no `type` property, and it won't be an object. I'd expect the migration to proceed correctly in that case.

arbron:
Resolved by 3.0 release

PR #2785, opened 2024-02-01
[#1546] Add enricher that rolls items

### LINKED ISSUE #1546: [Feature Request] Add Roll button to inline roll links

MaxPat931:
A refresh of #2626 for 3.0
and closes #1546 

Updated how the token/actor is gathered when given a Relative Id when the link is in the new Chat Cards as well as the Tooltips on the new Actor sheets.

image

## Item Enrichers
Item enrichers allow you to use an item from an enriched link. There are several different methods to create an Item enricher, which will determine how the item is used:  
**By Item Name:**  
`[[/item Bite]]`  
This functions similarly to a system macro, as if you have dragged that item to the macro hotbar. When clicked on it will check for a selected token, or your assigned actor, if a the found token or actor has an item of a matching name, it will be used, if the found token or actor does not have an item of that name, a warning will be displayed.  

**By UUID:**  
`[[/item Actor.p26xCjCCTQm5fRN3.Item.amUUCouL69OK1GZU]]`  
A UUID contains references to an Actor and an Item it owns. When clicked on, the enricher will find the specified Actor and use the specified Item.  

**By Relative ID:**  
`[[/item amUUCouL69OK1GZU]]`  
`[[/item .amUUCouL69OK1GZU]]`  
`[[/item .Bite]]`  
A Relative ID can contain a reference to an owned Item either by an Id, a relative Id (note the preceding `.`), or by name (again, note the preceding `.`). When clicked on the enricher will use its location (either in an Actor Sheet, Item Sheet, or Chat Card) to determine the Token or Actor that owns that card or sheet to use the specified item from that owner.


module/enrichers.mjs
```diff
@@ -823,6 +917,87 @@ async function rollAction(event) {
...
+              const sceneToken = parentScene.collections.tokens.get(locatedToken);
+              const tokenItem = sceneToken.delta.collections.items.get(target.dataset.rollRelativeItemId) || sceneToken.delta.collections.items.getName(target.dataset.rollRelativeItemName);
+              if (tokenItem) {
+                if (sceneToken.actor.testUserPermission(game.user, "OWNER")) {
+                  return tokenItem.use();
+                } else {
+                  return ui.notifications.warn(`You do not have ownership of ${sceneToken.name}, and cannot use this item!`);
+                }
+              }
+              else return ui.notifications.warn(`Item ${target.dataset.rollRelativeItemId || target.dataset.rollRelativeItemName} not found on Token ${sceneToken.name} in Scene ${parentScene.name}!`);
```
krbz999, line 994: Should localize this, display it, and then return `null`.

module/enrichers.mjs
```diff
@@ -636,6 +637,99 @@ async function enrichSave(config, label, options) {
 
 /* -------------------------------------------- */
 
+/**
+ * Enrich an item use link to roll an item on the selected token.
+ * @param {string[]} config            Configuration data.
+ * @param {string} [label]             Optional label to replace default text.
+ * @returns {HTMLElement|null}         An HTML link if the item link could be built, otherwise null.
```
krbz999, line 644:
```suggestion
 * @returns {Promise<HTMLElement|null>}         An HTML link if the item link could be built, otherwise null.
```

Fyorl:
I don't think we should support this custom UUID syntax. Mixing names with IDs complicates this code considerably, and also does not handle things like compendium UUIDs or synthetic Actor UUIDs. By not using `fromUuid` it also means it will not benefit from future improvements such as foundryvtt/foundryvtt#10301, or any changes made to UUIDs in general.

I would suggest we just support a single level name lookup that simply called `actor.items.getName`, as well as UUIDs and relative UUIDs via `fromUuid`.

MaxPat931:
Mixing the names with IDs does complicate the code a bit, but it is more user friendly since it's easy to see at a glance the actor and item names, not all users know how to get the UU/IDs, even though its just a quick press of the button  

Im not sure why we would want to support Compendium UUIDs, as there isnt a current method to use an item from a compendium (there was even an issue where use from the compendium was possible and it was closed as a bug #2350 )  

I dont really know what a synthetic actor UUID is used for tbh, so, how would a user actually get/use that to create one of these enrichers on their actor/token actor?  

Why use `actor.items.getName` instead of `dnd5e.documents.macro.rollItem`?   

I havent really been able to figure out how to get `fromUuid` to work with a relative id at all and I couldnt find anything in the discord that worked. So for instance, importing an Aboleth, it's Tentacle should have a UUID of `Actor.rfsMPNRV4EcZL5Gj.Item.sgauK8Lyt8qxsxOH` could you give me an example of how to use fromUuid with the item's relative UUID of `.sgauK8Lyt8qxsxOH`?

Fyorl:
> Mixing the names with IDs does complicate the code a bit, but it is more user friendly since it's easy to see at a glance the actor and item names, not all users know how to get the UU/IDs, even though its just a quick press of the button

I appreciate it's more user-friendly but I don't think we should support an entirely separate UUID scheme of our own invention, and only in one place. Let's start with actual UUIDs for now.

> I dont really know what a synthetic actor UUID is used for tbh, so, how would a user actually get/use that to create one of these enrichers on their actor/token actor?

Unlinked tokens have UUIDs that look like this: `Scene.IbFluxE7rYKZ6qXr.Token.ICIMDrbXo6qGD7xE.Actor.4IkQE3RWi7nkzVKD`. It can be obtained in exactly the same way as any other Actor UUID, just from the sheet of an unlinked token, rather than a linked one.

> Why use `actor.items.getName` instead of `dnd5e.documents.macro.rollItem`?

Doesn't `macro.rolIItem` require having a token selected? I may be misunderstanding, but if I have one of these Item enrichers on an Actor sheet, shouldn't clicking it roll it with the Actor whose sheet it is, rather than with the selected token?

> I havent really been able to figure out how to get `fromUuid` to work with a relative id at all and I couldnt find anything in the discord that worked. So for instance, importing an Aboleth, it's Tentacle should have a UUID of `Actor.rfsMPNRV4EcZL5Gj.Item.sgauK8Lyt8qxsxOH` could you give me an example of how to use fromUuid with the item's relative UUID of `.sgauK8Lyt8qxsxOH`?

The UUID of `".sgauK8Lyt8qxsxOH"` would be relative to one of the Aboleth's other Items. The UUID of the Tentacle Item relative to the Aboleth is `".Item.sgauK8Lyt8qxsxOH"`. You can generate the UUID of some Document relative to another Document with `Document#getRelativeUUID`.

```js
// This returns the Tentacle Item.
fromUuid(".Item.sgauK8Lyt8qxsxOH", { relative: aboleth });
```

MaxPat931:
Ok, I was able to take another pass at it, removing "Name UUID" support and adding better UUID handling including synthetic actor items UUIDs.
I did keep in the two methods for Names, the "Relative Name" `[[/item .Bite]]` which always resolves to the parent Token/Actor when in an owned item, and the regular "Name" `[[/item Bite]]` which uses the `macro.rollItem` because I do find it useful to be able to use one enricher in a journal or popped out chat card to Bite with all 12 zombies in an encounter. and it doesnt add much to the code.

module/enrichers.mjs
```diff
@@ -49,6 +49,7 @@ async function enrichString(match, options) {
     case "skill":
     case "tool": return enrichCheck(config, label, options);
     case "save": return enrichSave(config, label, options);
+    case "item": return enrichItem(config, label);
     case "embed": return enrichEmbed(config, label, options);
```
arbron, line 53: Just trying to keep some semblance of alphabetical order, though obviously that is tough given the way the switch fall-throughs are set up here :).

```suggestion
    case "embed": return enrichEmbed(config, label, options);
    case "item": return enrichItem(config, label);
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
 
 /* -------------------------------------------- */
 
+/**
+ * Enrich an item use link to roll an item on the selected token.
+ * @param {string[]} config            Configuration data.
+ * @param {string} [label]             Optional label to replace default text.
+ * @returns {Promise<HTMLElement|null>}         An HTML link if the item link could be built, otherwise null.
```
arbron, line 644:
```suggestion
 * @param {string[]} config              Configuration data.
 * @param {string} [label]               Optional label to replace default text.
 * @returns {Promise<HTMLElement|null>}  An HTML link if the item link could be built, otherwise null.
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+ *
+ * @example Use an item from a Name:
+ * ```[[/item Heavy Crossbow]]```
+ * becomes
+ * ```html
+ * <a class="roll-action" data-type="item">
```
arbron, line 650:
```suggestion
 * <a class="roll-action" data-type="item" data-roll-item-name="Heavy Crossbow">
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+ *   <i class="fa-solid fa-dice-d20"></i> Mace
+ * </a>
+ * ```
+ *
+ * @example Use an Item from a Relative UUID:
+ * ```[[/item .amUUCouL69OK1GZU]]```
+ * becomes
+ * ```html
+ * <a class="roll-action" data-type="item">
```
arbron, line 677:
```suggestion
 * <a class="roll-action" data-type="item" data-roll-relative-item-id="amUUCouL69OK1GZU">
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+ * ```
+ *
+ * @example Use an Item from a "Name UUID":
+ * ```[[/item Actor.Akra (Dragonborn Cleric).Item.Mace]]```
+ * becomes
+ * ```html
+ * <a class="roll-action" data-type="item">
+ *   <i class="fa-solid fa-dice-d20"></i> Mace
+ * </a>
+ * ```
```
arbron, line 671: Not a supported format any more.

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+ * <a class="roll-action" data-type="item">
+ *   <i class="fa-solid fa-dice-d20"></i> Bite
+ * </a>
+ * ```
+ */
+async function enrichItem(config, label) {
+  const givenItem = config.values.join(" ");
+  // If config is a UUID
+  const itemUuidMatch = givenItem.match(/^(?<synthid>Scene\.\w{16}\.Token\.\w{16}\.)?(?<actorid>Actor\.\w{16})(?<itemid>\.?Item(?<relativeId>\.\w{16}))$/);
+  if (itemUuidMatch) {
```
arbron, line 686:
```suggestion
  if ( itemUuidMatch ) {
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+ * ```html
+ * <a class="roll-action" data-type="item">
+ *   <i class="fa-solid fa-dice-d20"></i> Bite
+ * </a>
+ * ```
+ */
+async function enrichItem(config, label) {
+  const givenItem = config.values.join(" ");
+  // If config is a UUID
+  const itemUuidMatch = givenItem.match(/^(?<synthid>Scene\.\w{16}\.Token\.\w{16}\.)?(?<actorid>Actor\.\w{16})(?<itemid>\.?Item(?<relativeId>\.\w{16}))$/);
```
arbron, line 685: Bit too long of a line.

```suggestion
  const itemUuidMatch = givenItem.match(
    /^(?<synthid>Scene\.\w{16}\.Token\.\w{16}\.)?(?<actorid>Actor\.\w{16})(?<itemid>\.?Item(?<relativeId>\.\w{16}))$/
  );
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+  if (itemUuidMatch) {
+    const ownerActor = itemUuidMatch.groups.actorid.trim();
+    if (!label) {
```
arbron, line 688:
```suggestion
    if ( !label ) {
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+      if ((await fromUuid(givenItem)) == null) {
+        console.warn(`Item not found while enriching ${givenItem}.`);
+        return null;
+      } else label = (await fromUuid(givenItem)).name;
```
arbron, line 692: We shouldn't be calling `fromUuid` here twice. There is also no need for `else` when we always return within the `if` statement.

```suggestion
      const item = await fromUuid(givenItem);
      if ( !item ) {
        console.warn(`Item not found while enriching ${givenItem}.`);
        return null;
      }
      label = item.name;
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+    }
+    return createRollLink(label, { type: "item", rollItemActor: ownerActor, rollItemUuid: givenItem });
+  }
+
+  // If config is a relative Id/Name
+  const relativeIdMatch = givenItem.match(/^\.\w{16}$/);  // .amUUCouL69OK1GZU
+  const copiedIdMatch = givenItem.match(/^\w{16}$/);      // amUUCouL69OK1GZU
+  const relativeNameMatch = givenItem.startsWith(".");    // .Bite
```
arbron, line 700:
```suggestion
  const relativeIdMatch = givenItem.match(/^\.\w{16}$/);  // [[/item .amUUCouL69OK1GZU]]
  const copiedIdMatch = givenItem.match(/^\w{16}$/);      // [[/item amUUCouL69OK1GZU]]
  const relativeNameMatch = givenItem.startsWith(".");    // [[/item .Bite]]
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+  if (relativeIdMatch || copiedIdMatch || relativeNameMatch) {
```
arbron, line 701:
```suggestion
  if ( relativeIdMatch || copiedIdMatch || relativeNameMatch ) {
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+    const relativeId = relativeIdMatch ? givenItem.substr(1) : givenItem;
+    const foundActor = game.actors.find(actor => actor.items.get(relativeId));
+    if (foundActor) {
```
arbron, line 704:
```suggestion
    if ( foundActor ) {
```

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+    return createRollLink(label, { type: "item", rollItemActor: ownerActor, rollItemUuid: givenItem });
+  }
+
+  // If config is a relative Id/Name
+  const relativeIdMatch = givenItem.match(/^\.\w{16}$/);  // .amUUCouL69OK1GZU
+  const copiedIdMatch = givenItem.match(/^\w{16}$/);      // amUUCouL69OK1GZU
+  const relativeNameMatch = givenItem.startsWith(".");    // .Bite
+  if (relativeIdMatch || copiedIdMatch || relativeNameMatch) {
+    const relativeId = relativeIdMatch ? givenItem.substr(1) : givenItem;
+    const foundActor = game.actors.find(actor => actor.items.get(relativeId));
```
arbron, line 703: We shouldn't be doing an actor search like this. Relative IDs should always be relative to the item being enriched. So if I include `[[/item .amUUCouL69OK1GZU]]` it should only check the actor upon which the document is defined, and display an error otherwise.

If we pass the enrichment options into this method we can use `options.relativeTo` to get the document to which the description belongs, and then work up from there until we find an actor. If we don't find one, or an item with that ID doesn't exist on the actor, we return `null` with an error message, otherwise we use that item for the rest of the process.

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+        console.warn(`Item not found while enriching ${givenItem}.`);
+        return null;
+      } else label = (await fromUuid(givenItem)).name;
+    }
+    return createRollLink(label, { type: "item", rollItemActor: ownerActor, rollItemUuid: givenItem });
+  }
+
+  // If config is a relative Id/Name
+  const relativeIdMatch = givenItem.match(/^\.\w{16}$/);  // .amUUCouL69OK1GZU
+  const copiedIdMatch = givenItem.match(/^\w{16}$/);      // amUUCouL69OK1GZU
```
arbron, line 699: This doesn't seem to be used anywhere other than the `if` statement and `[[/item amUUCouL69OK1GZU]]` isn't a valid format for the enricher.

module/enrichers.mjs
```diff
@@ -636,6 +637,95 @@ async function enrichSave(config, label, options) {
...
+        label = relativeName;
+        return createRollLink(label, { type: "item", rollRelativeItemName: relativeName });
+      }
+    }
+  }
+
+  // Finally, if config is an item name
+  if (!label) {
+    label = givenItem;
+  }
```
arbron, line 723:
```suggestion
  if ( !label ) label = givenItem;
```

module/enrichers.mjs
```diff
@@ -748,7 +838,7 @@ function createRollLink(label, dataset) {
   span.insertAdjacentElement("afterbegin", link);
 
   // Add chat request link for GMs
-  if ( game.user.isGM && (dataset.type !== "damage") ) {
+  if ( game.user.isGM && (dataset.type !== "damage" && dataset.type !== "item") ) {
```
arbron, line 841:
```suggestion
  if ( game.user.isGM && (dataset.type !== "damage") && (dataset.type !== "item") ) {
```

Fyorl:
> and the regular "Name" `[[/item Bite]]` which uses the `macro.rollItem` because I do find it useful to be able to use one enricher in a journal or popped out chat card to Bite with all 12 zombies in an encounter.

Ah I think I just didn't understand the use-case for this. But if it's in chat or a journal with no parent Actor context, then that makes sense. If you invoke `macro.rollItem` with 12 selected zombies does it roll 12 attacks, or just 1? It will be important to tweak the functionality Jeff is proposing below:

> What I'm thinking is we have a single style without the leading period, and when it is clicked we first look at the selected token(s) and search them for an item of that name, activating the first one found. If no item exists on the selected actors, we search the actor upon which the item resides to see if the item exists there (this could also be reversed to check the actor upon which the description exists first, and then check selected tokens).

This seems like a reasonable idea. We get the best of both worlds without having to 'hard-code' whether the reference is relative or not.

MaxPat931:
> If you invoke `macro.rollItem` with 12 selected zombies does it roll 12 attacks, or just 1?

It will attempt the roll from the most recently selected token, which is how the macro functions if you drag an item to the hotbar, and also consistent with the other enrichers

> we first look at the selected token(s) and search them for an item of that name, activating the first one found. If no item exists on the selected actors, we search the actor upon which the item resides to see if the item exists there (this could also be reversed to check the actor upon which the description exists first, and then check selected tokens).

Which of these is the "right" order here? I think the first one makes more sense to me, but I am not sure

module/enrichers.mjs
MaxPat931: Why wouldnt that be a valid format? If a user clicks on the book icon of an item to copy its id, then this is the format they get so I thought it was important to account for. 
Essentially, if the `givenItem` is 16 characters, first check it as an id, if it doesnt match then it is treated as a name

module/enrichers.mjs
MaxPat931: ok cool, didn't know about `options.relativeTo` I'll try to search up on it, but happy to take any tips on implementing it ;)

arbron: Should be a simple matter of passing `options` into the method like the other enrichers, and then examining `relativeTo` to see if it is an actor (search for that actor for items), an item (search the item's parent if it exists for items), or something else (return `null`, cause there is no way to know where to search).

MaxPat931: The other enrichers that use `relativeTo` are the Embeds using `config.docs`, but this enricher doesn't have that, so I'm not sure what to pass into `options = { ...options, relativeTo: ??};` (or if that is even what I should be doing)

arbron: This doesn't have anything to do with `config`. Relative to is part of the `EnrichmentOptions` that core passes in when the enrichment is happening, so all you have to do is in `enrichString` pass in the third `options` parameter and you'll have access to it.

MaxPat931: Oh. yea I overcomplicated it, or was doing something wrong when I tried that earlier, but yep I'm seeing it now, should be able to make some adjustments today, thanks!

MaxPat931:
Ok I was able to update the Relative ID methods to using `relativeTo` and refactor the Name method to function as discussed, first attempt to use an item of the given name on the selected token, then if no token selected falling back to the assigned actor, then finally the Actor that owns the item the enricher is in (if it's in an item or chat card of an item)

module/enrichers.mjs
```diff
@@ -636,6 +637,85 @@ async function enrichSave(config, label, options) {
...
+      }
+      label = item.name;
+    }
+    return createRollLink(label, { type: "item", rollItemActor: ownerActor, rollItemUuid: givenItem });
+  }
+
+  // If config is a relative Id/Name
+  const relativeIdMatch = givenItem.match(/^\.\w{16}$/);  // [[/item .amUUCouL69OK1GZU]]
+  const copiedIdMatch = givenItem.match(/^\w{16}$/);      // [[/item amUUCouL69OK1GZU]]
+  if ( relativeIdMatch || copiedIdMatch ) {
```
arbron, line 696: Can be simplified into a single regex since we don't use either of these values later:

```suggestion
  if ( givenItem.match(/^\.?\w{16}$/) ) {
```

module/enrichers.mjs
```diff
@@ -636,6 +637,85 @@ async function enrichSave(config, label, options) {
...
+    if ( options.relativeTo?.documentName === "Item") {
+      const foundActor = options.relativeTo?.parent;
```
arbron, line 698: It would also be good to support actors here, in case you want to link to an item from an actor's description:

```suggestion
    const foundActor = options.relativeTo instanceof Item ? options.relativeTo.parent
      : options.relativeTo instanceof Actor ? options.relativeTo : null;
    if ( foundActor ) {
```

module/enrichers.mjs
```diff
@@ -636,6 +637,85 @@ async function enrichSave(config, label, options) {
...
+      const relativeId = givenItem.startsWith(".") ? givenItem.substr(1) : givenItem;
+      let foundItem = foundActor?.items.get(relativeId);
```
arbron, line 700: With the above change we can't get here unless there is a `foundActor`, so need for the conditional.

```suggestion
      const foundItem = foundActor.items.get(relativeId);
```

module/enrichers.mjs
```diff
@@ -636,6 +637,85 @@ async function enrichSave(config, label, options) {
...
+      if ( !item ) {
+        console.warn(`Item not found while enriching ${givenItem}.`);
+        return null;
+      }
+      label = item.name;
+    }
+    return createRollLink(label, { type: "item", rollItemActor: ownerActor, rollItemUuid: givenItem });
+  }
+
+  // If config is a relative Id/Name
```
arbron, line 693: No longer handles name path.

```suggestion
  // If config is a relative ID
```

module/enrichers.mjs
```diff
@@ -823,6 +903,36 @@ async function rollAction(event) {
...
           options.ability = ability;
           return await actor.rollToolCheck(tool, options);
+        case "item":
+          if ( target.dataset.rollItemUuid ) {
+            const gameActor = await fromUuid(target.dataset.rollItemActor);
+            if ( gameActor.testUserPermission(game.user, "OWNER") ) {
+              return (await fromUuid(target.dataset.rollItemUuid)).use();
+            } else {
+              return ui.notifications.warn(`You do not have ownership of ${gameActor?.name}, and cannot use this item.`);
+            }
```
arbron, line 913: This permissions check should probably be handled by `Item5e#use` method rather than here.

```suggestion
            return (await fromUuid(target.dataset.rollItemUuid)).use();
```

(in `item.mjs`)
```javascript
  async use(config={}, options={}) {
    if ( !this.isOwner ) {
      ui.notifications.error("DND5E.DocumentUseWarn", { localize: true });
      return null;
    }
```

(in `en.json`)
```json
"DND5E.DocumentUseWarn": "You lack permission to use an item on this document.",
```

module/enrichers.mjs
```diff
@@ -823,6 +903,36 @@ async function rollAction(event) {
...
+                  return ui.notifications.warn(`${gameActor.name} does not have an Item with name ${target.dataset.rollItemName}.`);
+                }
+              } else {
+                return ui.notifications.warn(`You do not have ownership of ${gameActor?.name}, and cannot use this item.`);
+              }
+            } else {
+              return ui.notifications.warn(`Your controlled actor ${canvas.tokens.controlled[0]?.actor.name} does not have an Item with name ${target.dataset.rollItemName}.`);
+            }
+          }
+
```
arbron, line 935: I find this logic a bit hard to follow, I've taken a pass in what I hope maintains the functionality. I've skipped going through the `rollItem` macro because I don't think it adds much here.

```suggestion
            actor = canvas.tokens.controlled[0]?.actor ?? game.user.character
              ?? await fromUuid(target.dataset.rollItemActor);
            if ( !actor ) {
              ui.notifications.error("No token or selected actor for which to use {item}.", { localize: true });
              return;
            }

            const item = actor?.items.getName(target.dataset.rollItemName);
            if ( !item ) {
              ui.notifications.error("{actor} does not have an Item with name {item}.", { localize: true });
              return;
            }

            return item.use();
```

Basically it sets the actor to be the first of (1) first controlled token on scene, (2) assigned actor, or (3) actor within which the enricher is embedded.

The notification messages will need to be converted to properly localized strings.

arbron (changes requested):

module/enrichers.mjs
MaxPat931: It looks like with this change non-owners can use an item enricher built with the UUID or Relative ID
It is used from the proper owning actor, but with the player as the Speaker. They can't use the Attack or Damage buttons, but yea it's not preventing them from using it
image

module/enrichers.mjs
MaxPat931: > I find this logic a bit hard to follow  

Same, tbh.
This is mostly for handling enrichers in chat cards, and while I think it would be odd for Player1 to use an `[[/item Shortsword]]` enricher from an NPC's Multiattack chat card - to use their own Shortsword, not the NPC's - but I dont see why it should necessarily be prevented?  

With the proposed change, if a user does not own the actor the enricher came from, and does not have a token selected,  but the actor assigned to them does have an item of a matching name, the item from the actor that owns the enricher will be used, I dont think the owned actor will ever get used, so I would swap around line 916 to 
`?? game.user.character ?? await fromUuid(target.dataset.rollItemActor);`  

Then, if the selected token does not have an item of that name, it will just throw an error, and not fall back to the owner actor, which I think is fine but was a part of what I think was discussed earlier, and is the use of the warning in line 922

module/enrichers.mjs
MaxPat931: I remember putting this in because some enrichers in chat cards that wouldnt enrich would not have `relativeTo` data, but I'm not seeing that now... Perhaps because if relative ids are not found, they are treated as names, since we cant really know if a 16 character string is an id or item name

module/enrichers.mjs
MaxPat931: I totally did not think about putting these into the Actor bio, makes sense though

MaxPat931:
I've made some changes due to the feedback, the logic for the Name rolling I hope is a bit simpler:
For Enrichers in a Sheet owned by an actor (actor sheet, tooltip, item sheet, item chat card)
- If used by the Owner of the parent Actor it will check if a token is selected and has an item of that name, if so use that item, else use the parent's item of that name (and throw a warning if the Parent does not have an item of that name, or selected token was bypassed for not having an item of that name)
- If used by a Non-Owner, use the macro method  (this prevents the enricher from acting differently for non-owners depending on if they have their token selected or not)

For enrichers not in an Actor owned Sheet (Journal, Rolltable result,  typed into chat)
- Use the Macro method

I know you wanted to get away from the macro method, but using it saved me from identifying token/actor and repeating the warnings

However, I did keep in the `testUserPermission` checks, as getting rid of them leads to non-owners being able to use the items for some reason

Oh, and these should also be excluded from the work in #3356 imo

module/enrichers.mjs
```diff
@@ -823,6 +903,36 @@ async function rollAction(event) {
...
+            const gameActor = await fromUuid(target.dataset.rollItemActor);
+            if ( gameActor.testUserPermission(game.user, "OWNER") ) {
+              return (await fromUuid(target.dataset.rollItemUuid)).use();
+            } else {
+              return ui.notifications.warn(`You do not have ownership of ${gameActor?.name}, and cannot use this item.`);
+            }
+          } else if ( target.dataset.rollItemName ) {
+            if ( target.dataset.rollItemActor ) {
+              const gameActor = await fromUuid(target.dataset.rollItemActor);
+              if ( gameActor.testUserPermission(game.user, "OWNER") ) {
```
krbz999, line 917: When we just check for the current user, this can be simplified:
```suggestion
              if ( gameActor.isOwner ) {
```
This (and some other lines of code) is repeated elsewhere; it would probably be good to move stuff further up the chain and just call them once, like using `fromUuid` up at the top just once.

module/enrichers.mjs
```diff
@@ -823,6 +903,36 @@ async function rollAction(event) {
...
+                if ( canvas.tokens.controlled[0]?.actor.items.getName(target.dataset.rollItemName) ) {
```
krbz999, line 918: If you defined `let item;` up top, you could simplify these, and just call `item.use()` at the very end (assuming `item` by that point has been defined).

module/enrichers.mjs
```diff
@@ -823,6 +903,36 @@ async function rollAction(event) {
...
+                  return canvas.tokens.controlled[0]?.actor.items.getName(target.dataset.rollItemName).use();
+                } else if ( gameActor.items.getName(target.dataset.rollItemName) ) {
+                  await gameActor.items.getName(target.dataset.rollItemName).use();
+                  if (canvas.tokens.controlled.length > 0 ) {
+                    ui.notifications.warn(`Your controlled token ${canvas.tokens.controlled[0].name} does not have an Item with name ${target.dataset.rollItemName} using Item Enricher owner`);
```
krbz999, line 923: This warning seems misplaced?

MaxPat931: We already checked if the user has a token selected, but without the Named item, so this is to give feedback that we arent rolling the item from that selected token but from the enricher owner, and its only needed if they actually have a token selected, as this is also the method that would be used if they dont have anything selected. Where else should it go?

krbz999: I would expect a warning that my selected token's actor simply doesn't have the item, and then nothing more happens.

MaxPat931: This is what we discussed up here https://github.com/foundryvtt/dnd5e/pull/2785#issuecomment-1965119618

module/enrichers.mjs
```diff
@@ -967,6 +1044,30 @@ async function rollAction(event) {
...
+                  }
+                  return;
+                } else {
+                  return ui.notifications.warn(`${gameActor.name} does not have an Item with name ${target.dataset.rollItemName}.`);
+                }
+              }
+              return dnd5e.documents.macro.rollItem(target.dataset.rollItemName);
+            } else {
+              return dnd5e.documents.macro.rollItem(target.dataset.rollItemName);
+            }
```
arbron, line 1069: Okay, I've gone over this logic to make the flow a bit clearer, add some comments, and remove some levels of indentation. The functionality should be the same, but let me know if you notice any issues.

```suggestion
          let item;
          // If UUID is provided, always roll that item directly
          if ( target.dataset.rollItemUuid ) item = await fromUuid(target.dataset.rollItemUuid);
          else if ( target.dataset.rollItemName ) {
            const actor = target.dataset.rollItemActor ? await fromUuid(target.dataset.rollItemActor) : null;

            // If no actor is specified or player isn't owner, fall back to the macro rolling logic
            if ( !actor?.isOwner ) return rollItem(target.dataset.rollItemName);

            // If a token is controlled and it has an item with the correct name, activate it
            if ( canvas.tokens.controlled[0]?.actor.items.getName(target.dataset.rollItemName) ) {
              item = canvas.tokens.controlled[0]?.actor.items.getName(target.dataset.rollItemName);
            }

            // Otherwise check the specified actor for the item
            else {
              item = actor.items.getName(target.dataset.rollItemName);

              // Display a warning to indicate the item wasn't rolled from the controlled actor
              if ( item && (canvas.tokens.controlled.length > 0) ) ui.notifications.warn(
                game.i18n.format("MACRO.5eMissingTargetWarn", {
                  actor: canvas.tokens.controlled[0].name, item: target.dataset.rollItemName
                })
              );
            }

            // If no item could be found at all, display a warning
            if ( !item ) ui.notifications.warn(game.i18n.format("EDITOR.DND5E.Inline.Warning.NoItemOnActor", {
              actor: actor.name, item: target.dataset.rollItemName
            }));
          }
          if ( item ) item.use();
          return;
```

This also requires adjusting in a few other places:

1) At the top of the file, I imported `rollItem` rather than refering to it using the whole path:

```javascript
import { rollItem } from "./documents/macro.mjs";
```

2) I adjusted a localization name for the warning, so the line below the shows `NoActorWarning` should be adjusted:

```javascript
        ui.notifications.warn("EDITOR.DND5E.Inline.Warning.NoActor", { localize: true });
```

3) New localization entries should be added to `EDITOR.DND5E.Inline`:

```json
  "Warning": {
    "NoActor": "No selected or assigned actor could be found to execute this roll.",
    "NoItemOnActor": "{actor} does not have an Item with name {item}.",
  }
```

And the "NoActorWarning" line should be removed.

arbron (changes requested):
Some reworking of the logic to make the flow clearer, but it shouldn't change the functionality. I've attached a patch file in case that is easier to apply:

[item-rolling.patch](https://github.com/foundryvtt/dnd5e/files/15118867/item-rolling.patch)

module/enrichers.mjs
MaxPat931: Looks good to me, updated. 
Thank you Jeff, I appreciate it

module/enrichers.mjs
```diff
@@ -967,11 +1045,45 @@ async function rollAction(event) {
...
+	              item = canvas.tokens.controlled[0]?.actor.items.getName(target.dataset.rollItemName);
+	            }
+	
+	            // Otherwise check the specified actor for the item
+	            else {
+	              item = actor.items.getName(target.dataset.rollItemName);
+	
+	              // Display a warning to indicate the item wasn't rolled from the controlled actor
+	              if ( item && (canvas.tokens.controlled.length > 0) ) ui.notifications.warn(
+	                game.i18n.format("MACRO.5eMissingTargetWarn", {
```
MaxPat931, line 1069: Looks like this warning doesnt get the type or name passed to it
image

arbron: Oh yeah, you'll want to change the `item` key to `name`, and pass in the localized "Item" for `type`.

MaxPat931: Since this was an existing warning, should I add a new one to `EDITOR.DND5E.Inline` instead?

module/enrichers.mjs
```diff
@@ -967,11 +1045,45 @@ async function rollAction(event) {
     try {
       switch ( type ) {
         case "damage": return await rollDamage(event);
+        case "item":
+          let item;
+	          // If UUID is provided, always roll that item directly
+	          if ( target.dataset.rollItemUuid ) item = await fromUuid(target.dataset.rollItemUuid);
```
Fyorl, line 1051: Check indentation

module/enrichers.mjs
```diff
@@ -967,11 +1045,45 @@ async function rollAction(event) {
...
+	          else if ( target.dataset.rollItemName ) {
```
Fyorl, line 1052: Could we split this branch out into its own function, maybe `findItem` or something?

Fyorl (changes requested):
Thanks Max and Jeff for getting this over the finish line. Only one major comment. If you agree that the relative UUIDs are a bit confusing then I think just removing support for them should be a pretty easy change. And we should make sure the issue with the warning text is fixed.

MaxPat931:
Thanks for the feedback Kim, unfortunately I dont have the bandwidth to continue with this at the moment, @arbron would you mind finishing this off?

Fyorl:
> Thanks for the feedback Kim, unfortunately I dont have the bandwidth to continue with this at the moment, @arbron would you mind finishing this off?

No problem, Max, thanks for your hard work on this. I can make these changes and get this merged.

Fyorl (approved):

MaxPat931:
Awesome, thanks Kim, looking forward to it!

PR #2789, opened 2024-02-01
Custom Enricher for Toggling Status Effects

### LINKED ISSUE #2788: Create a new custom Enricher for toggling Status Effects
An enricher to easily toggle the new status effects on a selected token would be really convenient

MaxPat931:
Closes #2788  
Allows you to toggle a status from `CONFIG.statusEffects` on the selected Token, e.g. `[[/status poisoned]]`

image

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -557,11 +557,17 @@ async function embedRollTable(config, label, options) {
  * ```@Reference[condition=unconscious]{Label}```
  * becomes
  * ```html
- * <a class="content-link" draggable="true"
- *    data-uuid="Compendium.dnd5e.rules.JournalEntry.w7eitkpD7QQTB6j0.JournalEntryPage.UWw13ISmMxDzmwbd"
- *    data-type="JournalEntryPage" data-tooltip="Text Page">
- *   <i class="fas fa-file-lines"></i> Label
- * </a>
+ * <span class="reference-link">
+ *   <a class="content-link" draggable="true"
+ *      data-uuid="Compendium.dnd5e.rules.JournalEntry.w7eitkpD7QQTB6j0.JournalEntryPage.UWw13ISmMxDzmwbd"
+ *      data-type="JournalEntryPage" data-tooltip="Text Page">
+ *     <i class="fas fa-book-open"></i> Label
+ *   </a>
+ *   <a class="enricher-action" data-action="apply" data-status="unconscious"
+ *      data-tooltip="EDITOR.DND5E.Inline.ApplyStatus" aria-label="Apply Status to Selected Tokens">
+ *     <i class="fas fa-fw fa-reply-all fa-flip-horizontal"></i>
+ *   </a>
+ * </span>
  * ```
  */
 async function enrichReference(config, label, options) {


thatlonelybugbear:
Would you say that `CONFIG.statusEffects` is the proper place to look for them statuses?

From a quick look, there are:
`CONFIG.DND5E.conditionTypes`
`CONFIG.DND5E.statusEffects`
`CONFIG.statusEffects`
`CONFIG.specialStatusEffects`

Granted not all are the same, but which one would the system devs suggest using?

Fyorl:
> Granted not all are the same, but which one would the system devs suggest using?

`CONFIG.statusEffects` is the canonical source for token HUD status effects. Use `CONFIG.DND5E.conditionTypes` if you want D&D conditions only.

Fyorl:
Should this functionality be added to the `&Reference` enricher instead? Rather than having two ways of referencing the same condition, i.e. `&Reference[restrained]` and `[[/status restrained]]`?

arbron:
We could add it to the reference enricher and just append an "Apply" button after like we do with the "Post to Chat" button for check/save enrichers.

Fyorl:
> We could add it to the reference enricher and just append an "Apply" button after like we do with the "Post to Chat" button for check/save enrichers. That would make this functionality available to all existing uses of the `&Reference` enricher. We know when looking up the UUID for the reference enricher what the type of rule referenced is, so it would be easy to add something special for condition references.

I was going to suggest maybe it only applies if you have a token selected, otherwise it opens the journal page, but I think your suggestion is better.

MaxPat931:
I was thinking recently, perhaps a little stuck in my current method (because I actually started this in 2.4.1, I was just waiting for the system status effects to be implemented), trying to add a Reference on to these, but just having one enricher and tacking this functionality definitely makes more sense.
Is this something you would like me to continue work on, or have one of you take over?

arbron:
> I was thinking recently, perhaps a little stuck in my current method (because I actually started this in 2.4.1, I was just waiting for the system status effects to be implemented), trying to add a Reference on to these, but just having one enricher and tacking this functionality definitely makes more sense. Is this something you would like me to continue work on, or have one of you take over?

If you feel okay going down that path, feel free to keep working on it.

MaxPat931:
Ok, I'll see what I can do this weekend!

MaxPat931:
Unfortunately, I havent been able to crack this one, I think you'll need to take over 😔

Fyorl:
> Unfortunately, I havent been able to crack this one, I think you'll need to take over 😔

No problem, I think can make the necessary adjustments

arbron (approved):
Looks good, just need to update the example in the JSDocs for `enrichReference`:

```
 * <span class="reference-link">
 *   <a class="content-link" draggable="true"
 *      data-uuid="Compendium.dnd5e.rules.JournalEntry.w7eitkpD7QQTB6j0.JournalEntryPage.UWw13ISmMxDzmwbd"
 *      data-type="JournalEntryPage" data-tooltip="Text Page">
 *     <i class="fas fa-file-lines"></i> Label
 *   </a>
 *   <a class="enricher-action" data-action="apply" data-status="unconscious"
 *      data-tooltip="EDITOR.DND5E.Inline.ApplyStatus" aria-label="Apply Status to Selected Tokens">
 *     <i class="fas fa-fw fa-reply-all fa-flip-horizontal"></i>
 *   </a>
 * </span>
```

Issue #2798, opened 2024-02-01
Allow Modified Encumbrance Thresholds to Affect Encumbrance Pips


Larkinabout:
When modifying the encumbrance thresholds in `CONFIG.DND5E.encumbrance.threshold`, it would be helpful if the pips on the encumbrance tracker of the 3.0 character sheet would move to be indicative of the modified thresholds.

As an example, `CONFIG.DND5E.encumbrance.threshold.encumbered` may be increased resulting in the stages between states no longer being equidistant. We do this in our game to make allowances for starting equipment, while still maintaining some level of weight management.

PR #2808, opened 2024-02-01
[#2798] Support adjustable encumbrance arrows

Larkinabout:
Allows for the encumbrance arrows on the encumbrance card to adjust based on thresholds:
image

### DIFF
--- a/less/v2/inventory.less
+++ b/less/v2/inventory.less
@@ -34,6 +34,8 @@
     width: 230px;
 
     .meter {
+      --encumbrance-low: 33%;
+      --encumbrance-high: 66%;
       border-radius: 3px 3px 0 0;
       height: 25px;
       border: none;
@@ -51,8 +53,6 @@
       }
 
       .breakpoint {
-        --encumbrance-low: 33%;
-        --encumbrance-high: 66%;
         display: block;
         position: absolute;
         block-size: 0;
--- a/less/v2/inventory.less
+++ b/less/v2/inventory.less
@@ -51,13 +51,15 @@
       }
 
       .breakpoint {
+        --encumbrance-low: 33%;
+        --encumbrance-high: 66%;
         display: block;
         position: absolute;
         block-size: 0;
         inline-size: 0;
         border-inline: 3px solid transparent;
-        &.encumbrance-33 { inset-inline-start: 33% }
-        &.encumbrance-66 { inset-inline-start: 66% }
+        &.encumbrance-low { inset-inline-start: var(--encumbrance-low); }
+        &.encumbrance-high { inset-inline-start: var(--encumbrance-high); }
         &.arrow-up {
           inset-block-end: 0;
           border-block-end: 3px solid var(--dnd5e-color-gold);
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -198,6 +198,19 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
       path: showTokenPortrait ? this.actor.isToken ? "" : "prototypeToken.texture.src" : "img"
     };
 
+    // Encumbrance
+    const encumberedThreshold = game.settings.get("dnd5e", "metricWeightUnits") 
+      ? CONFIG.DND5E.encumbrance.threshold.encumbered.metric 
+      : CONFIG.DND5E.encumbrance.threshold.encumbered.imperial;
+    const heavilyEncumberedThreshold = game.settings.get("dnd5e", "metricWeightUnits") 
+      ? CONFIG.DND5E.encumbrance.threshold.heavilyEncumbered.metric
+      : CONFIG.DND5E.encumbrance.threshold.heavilyEncumbered.imperial;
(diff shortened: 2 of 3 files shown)


Larkinabout:
Closes #2798.

module/applications/actor/character-sheet-2.mjs
```diff
@@ -198,6 +198,19 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
...
+      : CONFIG.DND5E.encumbrance.threshold.encumbered.imperial;
+    const heavilyEncumberedThreshold = game.settings.get("dnd5e", "metricWeightUnits") 
+      ? CONFIG.DND5E.encumbrance.threshold.heavilyEncumbered.metric
+      : CONFIG.DND5E.encumbrance.threshold.heavilyEncumbered.imperial;
+    const maximumThreshold = game.settings.get("dnd5e", "metricWeightUnits")
+      ? CONFIG.DND5E.encumbrance.threshold.maximum.metric
+      : CONFIG.DND5E.encumbrance.threshold.maximum.imperial;
+    context.encumbrance.low = (encumberedThreshold / maximumThreshold) * 100;
+    context.encumbrance.high = (heavilyEncumberedThreshold / maximumThreshold) * 100;
+
```
arbron, line 213: We are already calculating the stops in `system.attributes.encumbrance.stops`, so you can skip this whole section.

```suggestion
```

less/v2/inventory.less
```diff
@@ -51,13 +51,15 @@
       }
 
       .breakpoint {
+        --encumbrance-low: 33%;
+        --encumbrance-high: 66%;
```
arbron, line 55: If you move this up and define these variables on `.meter` instead, then you can set them in the template at the same declaration where we set `--bar-percentage` and avoid setting them twice within the template.

templates/shared/inventory2.hbs
```diff
@@ -14,10 +14,10 @@
...
                     <span class="max">{{ max }}</span>
                 </div>
-                <i class="breakpoint encumbrance-33 arrow-up" role="presentation"></i>
-                <i class="breakpoint encumbrance-33 arrow-down" role="presentation"></i>
-                <i class="breakpoint encumbrance-66 arrow-up" role="presentation"></i>
-                <i class="breakpoint encumbrance-66 arrow-down" role="presentation"></i>
+                <i class="breakpoint encumbrance-low arrow-up" role="presentation" style="--encumbrance-low: {{ low }}%"></i>
+                <i class="breakpoint encumbrance-low  arrow-down" role="presentation" style="--encumbrance-low: {{ low }}%"></i>
+                <i class="breakpoint encumbrance-high arrow-up" role="presentation" style="--encumbrance-high: {{ high }}%"></i>
+                <i class="breakpoint encumbrance-high arrow-down" role="presentation" style="--encumbrance-high: {{ high }}%"></i>
```
arbron, line 20: Set these above above on the meter div to avoid setting them twice:

```hbs
            <div class="meter progress" role="meter" aria-valuemin="0" aria-valuenow="{{ pct }}"
                 aria-valuetext="{{ value }}" aria-valuemax="{{ max }}" style="--bar-percentage: {{ pct }}%;
                 --encumbrance-low: {{ stops.encumbered }}%; --encumbrance-high: {{ stops.heavilyEncumbered }}%;">
```

```suggestion
                <i class="breakpoint encumbrance-low arrow-up" role="presentation"></i>
                <i class="breakpoint encumbrance-low  arrow-down" role="presentation"></i>
                <i class="breakpoint encumbrance-high arrow-up" role="presentation"></i>
                <i class="breakpoint encumbrance-high arrow-down" role="presentation"></i>
```

arbron (changes requested):

module/applications/actor/character-sheet-2.mjs
Larkinabout: It looks like the stops are only updated with an actor update, rather than on opening the character sheet or following an up to the encumbrance thresholds config. Definitely worthwhile using the existing calculation, but think it needs extra development to make the sheet instantly reflective of the config to avoid confusion for users.

templates/shared/inventory2.hbs
Larkinabout: I've moved this as suggested. Only thing different is I've put the style string all on one line as otherwise it gets a line break in the HTML and looks messy from that side. I had a look but I couldn't find a fancy handlerbar method to remove line breaks.

arbron (approved):

templates/shared/inventory2.hbs
arbron: If you do it within a handlebars element, you can use tilde:

```hbs
<div>   {{~ localize "blah" }}</div> <!-- Removes any space before -->
<div>{{ localize "blah" ~}}   </div> <!-- Removes any space after -->
<div>   {{~ localize "blah" ~}}   </div> <!-- Removes space on both sides -->
```

module/applications/actor/character-sheet-2.mjs
arbron: I'm not sure how important it is to do this dynamically, and if a module adds it to a UI they can just manually call `reset` on any actor when the value is changed.

PR #2811, opened 2024-02-01
Added Dark Mode support for 3.0.x

ZeroXNoxus:
This overhaul is aimed to easily add dark mode to the new sheets by using the "filter: invert(1);" styling on certain elements in the sheet and added some QOL stylings to the sidebar.

Also added a game setting to enable dark mode (requires reload).

May need some tweaking, I don't have much experience with LESS, though as I understand it, this should work as intended.


phenomen:
Inverting colors is not really a good way to handle dark mode...

ZeroXNoxus:
Let me get a preview ready, as this is using "invert", yes, but in a regulated way.

ZeroXNoxus:
Starting page:

image

Inventory page:

image

Feature page:

image

Spell page:

image

This is the current styling for the character sheet with dark mode active. I've taken some time and thought into creating this PR and I think a dark mode was strangely absent in the 5th Edition system.

There certainly are more robust ways to emulate the dark mode, sure, but I feel like this is the least intrusive way to establish it and, IMHO, this looks really good.

I can also publish this as a seperate module, but I feel like this would belong here.

ZeroXNoxus:
Chat changes:

image

Scene directory:

image

Actor directory:

image

Note: The height of the folders, actor sheets and scenes has been modified as the previous display seems very clunky.

dnd5e.mjs
ZeroXNoxus, line 1: Automatically set "dark" class to body and sheet if dark mode was enabled in the settings.

lang/en.json
ZeroXNoxus, line 1: Text for the new Dark Mode setting

less/v1/actors.less
ZeroXNoxus, line 1: Dark Mode Styling

less/v2/apps.less
ZeroXNoxus, line 1: Dark Mode Styling

less/v2/character.less
ZeroXNoxus, line 1: Dark Mode Styling

less/v2/chat.less
ZeroXNoxus, line 1: Dark Mode Styling

less/v2/inventory.less
ZeroXNoxus, line 1: Dark Mode Styling

less/v2/tooltips.less
ZeroXNoxus, line 1: Dark Mode Styling

module/settings.mjs
ZeroXNoxus, line 1: New Setting called "darkMode". Boolean to determine if the client wants Dark Mode displayed or not.

ZeroXNoxus (commented):
Everything seems in order from my side, the Dark Mode could be easily implemented and would liekly not be intrusive.

ZeroXNoxus:
Ah, I see. I haven't noticed this in the release notes and I just Put this together on an evening of free time. Will this dark mode include a dark mode for the chat, as well? 
And the redesign for folders, actor and scene entries could be included from this PR, as they look a lot cleaner this way.

Is there an ETA for the dark mode Release?

Fyorl:
> Will this dark mode include a dark mode for the chat, as well?

Eventually.

> And the redesign for folders, actor and scene entries could be included from this PR, as they look a lot cleaner this way.

Since these are part of the core UI, I would say that we should not include dark mode styles in the system, and they should come from core instead. It would be different if the system restyled the sidebar in some way in general, but it presently does not.

> Is there an ETA for the dark mode Release?

There is not.

PR #2815, opened 2024-02-01
[#2796] Align encumbrance arrows on centre

### LINKED ISSUE #2796: Encumbrance Pips Not Quite Matching Encumbrance Effects
With Variant Encumbrance enabled, the encumbrance bar and pips on the 3.0 character sheet do not appear to match up with when the Encumbered and Heavily Encumbered effects are applied.

Here the character has the Encumbered effect, but the first pip is slightly further along:
image

Likewise, this is when the Heavily Encumbered effect is applied:
image

No modules are enabled.

Larkinabout:
Align encumbrance arrows on their centres. Position of arrow in image is just to show an easy reference point.
image

Partially closes #2796 depending on the handling of the bar's border width.


Larkinabout:
Closes #2796.

arbron (approved):

PR #2836, opened 2024-02-02
Prevent chat card rendering from failing if sender has been removed from message

### LINKED ISSUE #2835: _enrichChatCard assumes message-sender exists
As an example, my module [Change Log](https://github.com/Larkinabout/fvtt-change-log/) has an option to remove the `message-sender` from the chat message using Foundry's `renderChatMessage` hook. This is for the purpose of reducing the chat card when the sender info isn't important. 
image

However, dnd5e 3.0.0's `_enrichChatCard` subsequently assumes the content of the chat message, specifically that `message-sender` exists.

It's possible that other modules would wish to modify the chat message using `renderChatMessage` rather than dnd5e's own hook where the module is intended to be system-agonistic, therefore I'd request that the system handles for missing content.

Larkinabout:
Closes #2835.

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -163,7 +163,7 @@ export default class ChatMessage5e extends ChatMessage {
     else if ( this.alias !== this.user?.name ) subtitle.innerText = this.user?.name ?? "";
     name.appendChild(subtitle);
     const sender = html.querySelector(".message-sender");
-    sender.replaceChildren(avatar, name);
+    if (sender) { sender.replaceChildren(avatar, name); }
     html.querySelector(".whisper-to")?.remove();
 
     // Context menu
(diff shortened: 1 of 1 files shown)


module/documents/chat-message.mjs
```diff
@@ -163,7 +163,7 @@ export default class ChatMessage5e extends ChatMessage {
     else if ( this.alias !== this.user?.name ) subtitle.innerText = this.user?.name ?? "";
     name.appendChild(subtitle);
     const sender = html.querySelector(".message-sender");
-    sender.replaceChildren(avatar, name);
+    if (sender) { sender.replaceChildren(avatar, name); }
```
arbron, line 166:
```suggestion
    sender?.replaceChildren(avatar, name);
```

arbron (changes requested):

arbron (approved):

PR #2846, opened 2024-02-03
[#2809, #2858, #2859] Add user configuration for status conditions and exhaustion

### LINKED ISSUE #2809: Add a property in the config object for conditions to allow users to disable the forced automation on an individual basis
eg `CONFIG.DND5E.conditionTypes.grappled.disabled` (or some better named key) which when explicitly `true` would not change any data on the actor automatically.

### LINKED ISSUE #2858: If an actor is immune to a condition, enabling said condition still modifies their attributes
(Fixed in #2846)

### LINKED ISSUE #2859: When changing the level of exhaustion, the name is not localized
(Fixed in #2846)

krbz999:
Adding in some much needed user agency, a quick one-liner can now modify which condition does what.

Doesn't touch exhaustion.

Closes #2809.
Closes #2858.
Closes #2859.

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2458,7 +2458,7 @@ patchConfig("conditionTypes", "label", { since: "DnD5e 3.0", until: "DnD5e 3.2" 
 DND5E.conditionEffects = {
   noMovement: new Set(["grappled", "paralyzed", "petrified", "restrained", "stunned", "unconscious"]),
   halfMovement: new Set(["prone"]),
-  crawl: new Set(["prone", "exceedingCarryingCapacity"]),
+  crawl: new Set(["prone"]),
   petrification: new Set(["petrified"])
 };
 
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2384,7 +2384,8 @@ DND5E.conditionTypes = {
   exhaustion: {
     label: "DND5E.ConExhaustion",
     icon: "systems/dnd5e/icons/svg/statuses/exhaustion.svg",
-    reference: "Compendium.dnd5e.rules.JournalEntry.w7eitkpD7QQTB6j0.JournalEntryPage.cspWveykstnu3Zcv"
+    reference: "Compendium.dnd5e.rules.JournalEntry.w7eitkpD7QQTB6j0.JournalEntryPage.cspWveykstnu3Zcv",
+    maximum: 6
   },
   frightened: {
     label: "DND5E.ConFrightened",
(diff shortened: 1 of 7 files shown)


arbron:
Would you be able to split the exhaustion changes from the other condition changes? I have a thought on the whole exhaustion design that I'd like to riff on that might take a bit longer but it would be good to get the other changes in the next version.

krbz999:
> I have a thought on the whole exhaustion design that I'd like to riff on

Would it conflict?

module/documents/actor/actor.mjs
```diff
@@ -209,6 +209,20 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * Is this actor under the effect of this property from some status?
+   * @param {string} key      A key in `DND5E.conditionProperties`.
+   * @returns {boolean}       Whether the actor is affected.
+   */
+  hasCondition(key) {
```
arbron, line 219: I'm going to vote on a slightly longer name here because this is checking for specific effects imposed by conditions, not the conditions themselves.

```suggestion
  hasConditionEffect(key) {
```

module/config.mjs
```diff
@@ -2451,6 +2451,19 @@ patchConfig("conditionTypes", "label", { since: "DnD5e 3.0", until: "DnD5e 3.2"
 
 /* -------------------------------------------- */
 
+/**
+ * Various effects of conditions and which conditions apply it.
+ * @enum {object}
+ */
+DND5E.conditionProperties = {
```
arbron, line 2458: Just to synchronized this with my suggested name for the function.

```suggestion
DND5E.conditionEffects = {
```

arbron (changes requested):

arbron (approved):

module/data/actor/templates/attributes.mjs
```diff
@@ -105,17 +105,17 @@ export default class AttributesFields {
...
-      .intersection(statuses).size || (this.attributes.exhaustion >= 5);
-    const halfMovement = statuses.has("prone") || (this.attributes.exhaustion >= 2);
+    const noMovement = this.parent.hasConditionEffect("noMovement") || (this.attributes.exhaustion >= 5);
+    const halfMovement = this.parent.hasConditionEffect("halfMovement") || (this.attributes.exhaustion >= 2);
     const reduction = statuses.has("heavilyEncumbered")
       ? CONFIG.DND5E.encumbrance.speedReduction.heavilyEncumbered
       : statuses.has("encumbered") ? CONFIG.DND5E.encumbrance.speedReduction.encumbered : 0;
-    const crawl = statuses.has("prone") || statuses.has("exceedingCarryingCapacity");
+    const exceeding = statuses.has("exceedingCarryingCapacity");
+    const crawl = this.parent.hasConditionEffect("crawl") || exceeding;
```
arbron, line 114: I'd vote to go back to the previous version, since that would allow customization of how `exceedingCarryingCapacity` behaves beyond simply enabling and disabling it in the settings.

krbz999: Not that I feel strongly about it, I simply didn't think that the subsystems of statuses and encumbrance should have overlapped as such.

PR #2861, opened 2024-02-03
Add a `dnd5e.calculateDamage` hook

krbz999:
Adds a hook to see what calculations have been made in the applyDamage function.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1050,7 +1050,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
      * @param {DamageDescription[]} damages       Damage descriptions.
      * @param {DamageApplicationOptions} options  Additional damage application options.
      * @returns {boolean}                         Explicitly return `false` to prevent damage application.
-     * @function dnd5e.preCalculateDamage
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -1021,17 +1021,15 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
     const rollData = this.getRollData({deterministic: true});
 
-    let amount = damages.reduce((total, d) => {
+    damages.forEach(d => {
       // Skip damage types with immunity
-      if ( !ignore("immunity", d.type) && hasEffect("di", d.type, d.properties) ) return total;
-
-      let value = d.value;
+      if ( !ignore("immunity", d.type) && hasEffect("di", d.type, d.properties) ) return;
```
arbron, line 1026: This will end up applying the full damage for any immunities.

```suggestion
      if ( !ignore("immunity", d.type) && hasEffect("di", d.type, d.properties) ) {
        d.value = 0;
        return;
      }
```

arbron (changes requested):

arbron (approved):

PR #2882, opened 2024-02-04
[#2838] Don't collapse chat card if the user clicked a link or button in the collapsible region

### LINKED ISSUE #2838: Clicking on the request roll from an item's chat card description collapses the description
I was testing out enrichers on features and realized that the "request roll" icon shows up in those too, not just journal entries. So I created a new Feature item and put this in the description:

```
I wonder if it's possible to have enrichers here?
Make a [[/skill Athletics 15]] check to escape
```

I clicked on the feature to send it to chat and the chat card had the "request roll" icon, that chat bubble. I clicked on it and it created another chat message with the roll request just as expected. What wasn't expected is that it collapsed the feature's description. It appears that clicking anywhere in the description will collapse it, not just clicking on the arrow. That would mean that both listeners are kicking in: the one to request a roll and the one to collapse the description. Personally, I don't think it should collapse when clicking anywhere in the description, just the arrow or maybe that header, but not the text itself.

schultzcole:
Closes #2838

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1988,6 +1988,10 @@ export default class Item5e extends SystemDocumentMixin(Item) {
    * @private
    */
   static _onChatCardToggleContent(event) {
+    // If the user is clicking on a link in the collapsible region, don't collapse
+    const clickTarget = event.target.closest(':is(.item-name, .collapsible) :is(a, button)')
+    if ( clickTarget) return;
+
     event.preventDefault();
     const header = event.currentTarget;
     const card = header.closest(".chat-card");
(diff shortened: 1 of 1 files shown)


module/documents/item.mjs
```diff
@@ -1988,6 +1988,10 @@ export default class Item5e extends SystemDocumentMixin(Item) {
    * @private
    */
   static _onChatCardToggleContent(event) {
+    // If the user is clicking on a link in the collapsible region, don't collapse
+    const clickTarget = event.target.closest(':is(.item-name, .collapsible) :is(a, button)')
+    if ( clickTarget) return;
```
arbron, line 1993:
```suggestion
    if ( event.target.closest(":is(.item-name, .collapsible) :is(a, button)") ) return;
```

arbron (approved):

PR #2893, opened 2024-02-04
[#2869] Added Limited Sheets, Partial sheets, and Modified logic in character sheet 2 if permission is limited

### LINKED ISSUE #2869: Limited Ownership setting for Player Character sheets is no longer working in v3.0.0
Since i migrated to 3.0, the selected "limited" Option for the Players Character sheets seem to have no affect. All players can see the full character sheets of other characters, even when their visibility should be "limited".

Hoppyhob:
-Added limited-sheet-2.hbs created by paring down the character-sheet-2.hbs
-Added limited-biography.hbs created by paring down the character-biography.hbs
-Updated utils.mjs to add limited-biography.hbs partial to preload
-Updated character.less to add css for limited class under .tab.biography
-Modified character-sheet-2.mjs:
--Added template() check if user is not GM and has Limited permission return limited template 
--Added check around rendering tabs, so that tabs are only rendered when user is GM or not Limited 
--Modified activeTab in getData to force biography if not GM user or limited (this is used to set cssClass which is used in handlebars and css styling).

Closes #2869

New Limited Sheet:
Screenshot 2024-02-04 125846
Old Limited Sheet
Screenshot 2024-02-04 125928


Fyorl:
I have some (unused) alternative header designs. Obviously we don't need the class, level, and XP bar in the limited sheets but maybe moving the portrait art up into the header could work, leaving the biography to be a full page. We could also decrease the `min-width` of the limited sheet then.

image

Fyorl:
> If we want to use the alternate header, I can work toward adding that to this PR. If we have a jumping off point that might help since it seems like there was a prototype, and that might help prevent me from trying to redo work to re-create it, unless @Fyorl wants to add it.

It's only a little bit of styling and template adjustment so it's probably easiest if I just tack it onto the end of this PR.

templates/actors/limited-sheet-2.hbs
```diff
@@ -0,0 +1,47 @@
...
+                <div class="tab biography active" data-group="primary" data-tab="biography">
+
+                    {{!-- Biography --}}
+                    <div class="bottom">
+                        <h3 class="icon">
+                            <i class="fas fa-feather"></i>
+                            <dnd5e-icon src="systems/dnd5e/icons/svg/ink-pot.svg"></dnd5e-icon>
+                            <span class="roboto-upper">{{ localize "DND5E.Biography" }}</span>
+                        </h3>
+                        {{ editor biographyHTML target="system.details.biography.value" button=false editable=@root.editable
```
arbron, line 35:
```suggestion
                        {{ editor biographyHTML target="system.details.biography.value" button=false
                                  editable=@root.editable engine="prosemirror" collaborate=false }}
```

arbron (approved):

PR #2898, opened 2024-02-04
[#2812] Allow favoriting tool categories

### LINKED ISSUE #2812: Tool categories cannot be favorited, only individual tool profs

schultzcole:
Closes #2812

Work in progress:
- [x] Add full tool type proficiency configs so that we can associate tooltip references and icons with tool types.
- [x] Handle drag/drop of tool types.
- [x] Handle using tool type proficiency config to populate favorite data.
- [ ] Add tool type description journal entry pages and use them to populate tooltips for tool types.
- [ ] Find an icon for vehicle tool proficiency??
- [ ] Enable favoriting specific vehicle proficiencies.

### DIFF
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -686,6 +686,7 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
     let type;
     if ( key in CONFIG.DND5E.skills ) type = "skill";
     else if ( key in CONFIG.DND5E.toolIds ) type = "tool";
+    else if ( key in CONFIG.DND5E.toolTypeProficiencies ) type = "tool";
     else if ( preparationMode && (level !== "0") && isSlots ) type = "slots";
     if ( !type ) return super._onDragStart(event);
     const dragData = { dnd5e: { action: "favorite", type } };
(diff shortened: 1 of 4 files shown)


schultzcole:
@Fyorl just wanted to check in and see if you think this approach is a good start before getting too far.

Fyorl:
I question the usefulness of being able to favourite tool categories. There is no way to be proficient in an entire tool category (in the rules that is, I know we allow it in the system), and similarly there are no rules for calling for an ability check with a tool category. Given that this PR is already 100+ LoC I think we need a more compelling use case.

krbz999:
It seemed intended to be able to favorite tool categories since they could be dragged and dropped (with no effect). I think it would be fine to just remove the ability to drag-start tool categories, however if a character is proficient with, say, all Artisan's Tools, they would actually not be able to mark themselves as proficient with an individual one.

schultzcole:
I'm fine to abandon this, although I do think there's some value in supporting tooltips for tool categories aside from being able to favorite them.

Fyorl:
> if a character is proficient with, say, all Artisan's Tools, they would actually not be able to mark themselves as proficient with an individual one and thus not able to favorite an individual one either.

I think this is a case of it being less effort to have proficiency in tool categories because they use the same underlying mechanism as weapons and armour proficiencies, where you *can* be proficient in an entire category. Obviously it's *more* effort now to allow tool categories to be favourited.

Maybe allowing them to be favourited is the best way forward then. Or we prevent tool categories from being marked as proficient. That would be simple enough but it unfortunately would come with a migration too at this stage so I'm not sure.

krbz999:
> Maybe allowing them to be favourited is the best way forward then. Or we prevent tool categories from being marked as proficient. That would be simple enough but it unfortunately would come with a migration too at this stage so I'm not sure.

We could change how proficient tool categories (read: *only tools*) are rendered on the sheet; proficiency in a category listing each entry in that category individually.

PR #2899, opened 2024-02-04
Fix linter warnings

schultzcole:


module/data/item/templates/activated-effect.mjs
```diff
@@ -336,7 +336,7 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
   get activatedEffectChatProperties() {
     foundry.utils.logCompatibilityWarning("ActivatedEffectTemplate#activatedEffectChatProperties is deprecated. "
       + "Please use ActivatedEffectTemplate#activatedEffectCardProperties.",
-      { since: "DnD5e 3.0", until: "DnD5e 3.2", once: true });
+    { since: "DnD5e 3.0", until: "DnD5e 3.2", once: true });
```
arbron, line 339: I know the linter says this is correct, but it doesn't feel like the correct indentation. I'd say we just move the first line of the warning down and have everything indented:

```javascript
      "ActivatedEffectTemplate#activatedEffectChatProperties is deprecated. "
      + "Please use ActivatedEffectTemplate#activatedEffectCardProperties.",
      { since: "DnD5e 3.0", until: "DnD5e 3.2", once: true });
```

module/data/item/templates/equippable-item.mjs
```diff
@@ -79,7 +79,7 @@ export default class EquippableItemTemplate extends SystemDataModel {
   get equippableItemChatProperties() {
     foundry.utils.logCompatibilityWarning("EquippableItemTemplate#equippableItemChatProperties is deprecated. "
       + "Please use EquippableItemTemplate#equippableItemCardProperties.",
-      { since: "DnD5e 3.0", until: "DnD5e 3.2", once: true });
+    { since: "DnD5e 3.0", until: "DnD5e 3.2", once: true });
```
arbron, line 82:
```javascript
      "EquippableItemTemplate#equippableItemChatProperties is deprecated. "
      + "Please use EquippableItemTemplate#equippableItemCardProperties.",
      { since: "DnD5e 3.0", until: "DnD5e 3.2", once: true });
```

module/data/item/templates/activated-effect.mjs
schultzcole: Yeah, sounds reasonable

PR #2922, opened 2024-02-06
[#2911] Privatize item roll cards

### LINKED ISSUE #2911: Private GM Rolls Appearing for Players
Just ran into this issue today on version 3.0.1.

I had a hidden token with chat set to "Private GM Roll". Upon rolling a (private) attack roll, the creature's tile, name, and weapon appeared in chat.

image

This is what I see: 
image

This is what my players see:
image

In <3.0 versions, this would be completely hidden from my player's views, as intended with the "Private GM Roll".

Hoping this is either a bug or something that can be toggled. Thanks for all your work!

schultzcole:
Closes #2911

Observer view on the left, speaker view on the right

Screenshot_20240205_180750

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -186,7 +186,7 @@ export default class ChatMessage5e extends ChatMessage {
     // Enriched roll flavor
     const roll = this.getFlag("dnd5e", "roll");
     const item = fromUuidSync(roll?.itemUuid);
-    if ( item ) {
+    if ( this.isContentVisible && item ) {
       const isCritical = (roll.type === "damage") && this.rolls[0]?.options?.critical;
       const subtitle = roll.type === "damage"
         ? isCritical ? game.i18n.localize("DND5E.CriticalHit") : game.i18n.localize("DND5E.DamageRoll")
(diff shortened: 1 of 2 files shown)


IonPickaxe43:
You all rock!

module/documents/chat-message.mjs
```diff
@@ -150,18 +150,32 @@ export default class ChatMessage5e extends ChatMessage {
...
     const name = document.createElement("span");
     name.classList.add("name-stacked");
-    name.innerHTML = `<span class="title">${this.alias}</span>`;
+    name.innerHTML = `<span class="title">${nameText}</span>`;
+
     const subtitle = document.createElement("span");
     subtitle.classList.add("subtitle");
     if ( this.whisper.length ) subtitle.innerText = html.querySelector(".whisper-to")?.innerText ?? "";
-    else if ( this.alias !== this.user?.name ) subtitle.innerText = this.user?.name ?? "";
+    if ( nameText !== this.user?.name && !subtitle.innerText.length ) subtitle.innerText = this.user?.name ?? "";
```
arbron, line 175:
```suggestion
    if ( (nameText !== this.user?.name) && !subtitle.innerText.length ) subtitle.innerText = this.user?.name ?? "";
```

arbron (changes requested):
I've also noticed that if you click on the total in a hidden message you get an error because it tries to expand a non-existent tooltip. If you move the event listener on `.dice-roll` up a line into the block that should prevent the error.

We should also modify it so it doesn't appear to be clickable. I'd recommend adding a `secret` class to the `roll-total` block if content isn't visible and then tweak the styling to remove the disclosure arrow and use the default cursor rather than the pointer.

arbron (approved):

PR #2926, opened 2024-02-06
[#2895] convert spell to scroll when dropping in container

### LINKED ISSUE #2895: Dropping a Spell into an owned Container adds the spell to the owner's Spellbook
I would expect a spell scroll to be created. Dropping a spell into an unowned container doesnt seem to do anything.

https://github.com/foundryvtt/dnd5e/assets/86370342/1dd32e29-d827-442c-82ec-9e55c666d6ae

schultzcole:
Closes #2895


arbron (changes requested):
Looks good, but the same change should be made above in `_onDropFolder` to ensure it works with folders of spells.

arbron (approved):

PR #2945, opened 2024-02-07
[#2931] Fallback to default item type label

### LINKED ISSUE #2931: Invalid `system.type.value` on certain item types can cause the item sheet to fail rendering and data preparation to fail for both the item and its owner
```js
await Item.create({type: "feat", name: "Steve", system: {type: {value: "steve"}}}, {parent: _token.actor});
```

schultzcole:
Closes #2931

### DIFF
--- a/module/data/item/equipment.mjs
+++ b/module/data/item/equipment.mjs
@@ -139,7 +139,8 @@ export default class EquipmentData extends ItemDataModel.mixin(
   /** @inheritDoc */
   prepareDerivedData() {
     super.prepareDerivedData();
-    this.type.label = CONFIG.DND5E.equipmentTypes[this.type.value];
+    this.type.label = CONFIG.DND5E.equipmentTypes[this.type.value]
+      ?? game.i18n.localize(CONFIG.Item.typeLabels.equipment);
   }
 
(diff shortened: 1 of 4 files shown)


module/data/item/equipment.mjs
```diff
@@ -215,6 +216,7 @@ export default class EquipmentData extends ItemDataModel.mixin(
     if ( actor.type === "npc" ) return 1; // NPCs are always considered proficient with any armor in their stat block.
     const config = CONFIG.DND5E.armorProficienciesMap;
     const itemProf = config[this.type.value];
+    if ( !itemProf ) return 0;
```
arbron, line 219: Other than skipping 2 lines of code, does this give us anything? The code seems to work otherwise even if the config isn't found (`undefined === true` and `actorProfs.has(undefined)` are still valid pieces of code that return false, which ultimately results in `Number(false)` returning `0`.

schultzcole: Yeah, I suppose if item prof is undefined it could still get a value from `actorProfs[this.type.baseitem]` so it's technically incorrect to return early. I'll remove it.

arbron (approved):

Issue #2954, opened 2024-02-07
Add `getData` Hooks


Larkinabout:
Call hooks within `getData` to allow data to be added prior to sheet rendering. It would allow modules to easily do stuff like this without needing to edit the html:

![dnd5e-getdata](https://github.com/foundryvtt/dnd5e/assets/105953297/89f3b853-ea04-492e-88b5-0ec75319ef7b)

arbron:
I feel like this would be much better handled by the core software as a counterpart to the existing `renderApplication` hooks. A `preRenderApplication` set of hooks that provides access to the already prepared data. I've filed an issue on core about that: https://github.com/foundryvtt/foundryvtt/issues/10397

arbron:
Coming with core in Application V2, so this will probably be implemented whenever we go V12-only and switch over to that.

PR #2955, opened 2024-02-07
[#2954] Add 'getData' hooks

Larkinabout:
Add hook to `ActorSheet5eCharacter2.getData`.

Mostly an example to see if it's viable. If so, I can look to see where else it might be appropriate, such as for the legacy sheets.


Larkinabout:
Seems to work well applied to all config and sheet apps. If viable, I'm happy to do the development as I'm working on a module that will put the functionality to good use.

PR #2959, opened 2024-02-07
Allow for exhaustion to be configurable

krbz999:
Allows for exhaustion to be configured in the `CONFIG` object (or entirely disabled) and generalizes some code surrounding it to match.

### DIFF
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -200,7 +200,7 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
     };
 
     // Exhaustion
-    const max = CONFIG.DND5E.conditionTypes.exhaustion.maximum;
+    const max = CONFIG.DND5E.conditionTypes.exhaustion.levels;
     context.exhaustion = {
       pips: Array.fromRange(max, 1).map(n => {
         const label = game.i18n.format("DND5E.ExhaustionLevel", { n });
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -200,14 +200,18 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
     };
 
     // Exhaustion
-    context.exhaustion = Array.fromRange(6, 1).map(n => {
-      const label = game.i18n.format("DND5E.ExhaustionLevel", { n });
-      const classes = ["pip"];
-      const filled = attributes.exhaustion >= n;
-      if ( filled ) classes.push("filled");
-      if ( n === 6 ) classes.push("death");
-      return { n, label, filled, tooltip: label, classes: classes.join(" ") };
-    });
+    const max = CONFIG.DND5E.conditionTypes.exhaustion.maximum;
+    context.exhaustion = {
+      pips: Array.fromRange(max, 1).map(n => {
+        const label = game.i18n.format("DND5E.ExhaustionLevel", { n });
+        const classes = ["pip"];
+        const filled = attributes.exhaustion >= n;
+        if ( filled ) classes.push("filled");
+        if ( n === max ) classes.push("death");
+        return { n, label, filled, tooltip: label, classes: classes.join(" ") };
+      }),
+      mid: max / 2
+    };
 
     // Speed
     context.speed = Object.entries(CONFIG.DND5E.movementTypes).reduce((obj, [k, label]) => {
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -201,17 +201,18 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
 
     // Exhaustion
     const max = CONFIG.DND5E.conditionTypes.exhaustion.levels;
-    context.exhaustion = {
-      pips: Array.fromRange(max, 1).map(n => {
-        const label = game.i18n.format("DND5E.ExhaustionLevel", { n });
-        const classes = ["pip"];
-        const filled = attributes.exhaustion >= n;
-        if ( filled ) classes.push("filled");
-        if ( n === max ) classes.push("death");
-        return { n, label, filled, tooltip: label, classes: classes.join(" ") };
-      }),
-      mid: max / 2
-    };
+    context.exhaustion = Array.fromRange(max, 1).reduce((acc, n) => {
+      const label = game.i18n.format("DND5E.ExhaustionLevel", { n });
+      const classes = ["pip"];
+      const filled = attributes.exhaustion >= n;
+      if ( filled ) classes.push("filled");
+      if ( n === max ) classes.push("death");
+      const pip = { n, label, filled, tooltip: label, classes: classes.join(" ") };
+
+      if ( n <= max / 2 ) acc.left.push(pip);
+      else acc.right.push(pip);
+      return acc;
+    }, { left: [], right: [] });
 
     // Speed
     context.speed = Object.entries(CONFIG.DND5E.movementTypes).reduce((obj, [k, label]) => {
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2385,7 +2385,8 @@ DND5E.conditionTypes = {
   exhaustion: {
     label: "DND5E.ConExhaustion",
     icon: "systems/dnd5e/icons/svg/statuses/exhaustion.svg",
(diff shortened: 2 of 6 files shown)


module/config.mjs
```diff
@@ -2385,7 +2385,8 @@ DND5E.conditionTypes = {
   exhaustion: {
     label: "DND5E.ConExhaustion",
     icon: "systems/dnd5e/icons/svg/statuses/exhaustion.svg",
-    reference: "Compendium.dnd5e.rules.JournalEntry.w7eitkpD7QQTB6j0.JournalEntryPage.cspWveykstnu3Zcv"
+    reference: "Compendium.dnd5e.rules.JournalEntry.w7eitkpD7QQTB6j0.JournalEntryPage.cspWveykstnu3Zcv",
+    maximum: 6
```
arbron, line 2389: What about calling this `levels`, like "6 levels of exhaustion"? You should also update the typedef for `ConditionConfiguration` above to mention this property and note it is only used for exhaustion.

```suggestion
    levels: 6
```

module/config.mjs
```diff
@@ -2453,14 +2454,16 @@ patchConfig("conditionTypes", "label", { since: "DnD5e 3.0", until: "DnD5e 3.2"
...
- * Various effects of conditions and which conditions apply it.
+ * Various effects of conditions and which conditions apply it. Either keys for the conditions,
+ * or a number for a level of exhaustion.
  * @enum {object}
  */
 DND5E.conditionEffects = {
-  noMovement: new Set(["grappled", "paralyzed", "petrified", "restrained", "stunned", "unconscious"]),
-  halfMovement: new Set(["prone"]),
+  noMovement: new Set([5, "grappled", "paralyzed", "petrified", "restrained", "stunned", "unconscious"]),
+  halfMovement: new Set([2, "prone"]),
```
arbron, line 2463: Not a big fan of just these numbers here, how do you feel about `"exhaustion-5"` instead?

```suggestion
  noMovement: new Set(["exhaustion-5", "grappled", "paralyzed", "petrified", "restrained", "stunned", "unconscious"]),
  halfMovement: new Set(["exhaustion-2", "prone"]),
```

module/config.mjs
```diff
@@ -2453,14 +2454,16 @@ patchConfig("conditionTypes", "label", { since: "DnD5e 3.0", until: "DnD5e 3.2"
...
   crawl: new Set(["prone", "exceedingCarryingCapacity"]),
-  petrification: new Set(["petrified"])
+  petrification: new Set(["petrified"]),
+  halfHealth: new Set([4])
```
arbron, line 2466:
```suggestion
  halfHealth: new Set(["exhaustion-4"])
```

module/documents/actor/actor.mjs
```diff
@@ -212,15 +212,19 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
    */
   hasConditionEffect(key) {
     const props = CONFIG.DND5E.conditionEffects[key] ?? new Set();
+    const level = this.system.attributes?.exhaustion ?? null;
     const imms = this.system.traits?.ci?.value ?? new Set();
     const statuses = this.statuses;
-    return statuses.difference(imms).intersects(props);
+    return props.some(k => {
+      return (statuses.has(k) && !imms.has(k))
+        || (!imms.has("exhaustion") && (level !== null) && Number.isInteger(k) && (level >= k));
```
arbron, line 226: If you modify how exhaustion is listed in `conditionEffects`.

```suggestion
      const l = Number(k.split("-").pop());
      return (statuses.has(k) && !imms.has(k))
        || (!imms.has("exhaustion") && (level !== null) && Number.isInteger(l) && (level >= l));
```

module/documents/actor/actor.mjs
```diff
@@ -212,15 +212,19 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
   /* -------------------------------------------- */
 
   /**
-   * Is this actor under the effect of this property from some status?
+   *Is this actor under the effect of this property from some status or possibly due to its level of exhaustion?
```
arbron, line 215:
```suggestion
   * Is this actor under the effect of this property from some status or due to its level of exhaustion?
```

module/documents/active-effect.mjs
```diff
@@ -285,11 +285,12 @@ export default class ActiveEffect5e extends ActiveEffect {
    * @protected
    */
   _prepareExhaustionLevel() {
+    const config = CONFIG.DND5E.conditionTypes.exhaustion;
     let level = this.getFlag("dnd5e", "exhaustionLevel");
     if ( !Number.isFinite(level) ) level = 1;
-    this.icon = `systems/dnd5e/icons/svg/statuses/exhaustion-${level}.svg`;
+    this.icon = `${config.icon.split(".")[0]}-${level}.svg`;
```
arbron, line 291: Since we are going this logic twice, I'd say split it out into a `_getExhaustionIcon` helper method. It would also be a good idea to be cognizant of the extension passed in in case someone wasn't using an SVG or the path includes other dots.

```javascript
  /**
   * Get the image used to represent exhaustion at this level.
   * @param {number} level
   * @returns {string}
   */
  _getExhaustionImage(level) {
    const split = CONFIG.DND5E.conditionTypes.exhaustion.icon.split(".");
    const ext = split.pop();
    const path = split.join(".");
    return `${path}-${level}.${ext}`;
  }

File: module/documents/active-effect.mjs [hunk 7]
```diff
@@ -285,11 +285,12 @@ export default class ActiveEffect5e extends ActiveEffect {
...
     this.name = `${game.i18n.localize("DND5E.Exhaustion")} ${level}`;
-    if ( level >= 6 ) this.statuses.add("dead");
+    if ( level >= config.maximum ) this.statuses.add("dead");
```
arbron, line 293:
```suggestion
    if ( level >= config.levels ) this.statuses.add("dead");
```

module/documents/active-effect.mjs
```diff
@@ -387,7 +389,8 @@ export default class ActiveEffect5e extends ActiveEffect {
     event.stopPropagation();
     if ( event.button === 0 ) level++;
     else level--;
-    actor.update({ "system.attributes.exhaustion": Math.clamped(level, 0, 6) });
+    const max = CONFIG.DND5E.conditionTypes.exhaustion.maximum;
```
arbron, line 392:
```suggestion
    const max = CONFIG.DND5E.conditionTypes.exhaustion.levels;
```

module/applications/actor/character-sheet-2.mjs
```diff
@@ -200,14 +200,18 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
...
     // Exhaustion
-    context.exhaustion = Array.fromRange(6, 1).map(n => {
-      const label = game.i18n.format("DND5E.ExhaustionLevel", { n });
-      const classes = ["pip"];
-      const filled = attributes.exhaustion >= n;
-      if ( filled ) classes.push("filled");
-      if ( n === 6 ) classes.push("death");
-      return { n, label, filled, tooltip: label, classes: classes.join(" ") };
-    });
+    const max = CONFIG.DND5E.conditionTypes.exhaustion.maximum;
```
arbron, line 203:
```suggestion
    const max = CONFIG.DND5E.conditionTypes.exhaustion.levels;
```

templates/actors/character-sheet-2.hbs
```diff
@@ -148,8 +148,8 @@
                         <div class="top">
 
                             <div class="pips" data-prop="system.attributes.exhaustion">
-                                {{#each exhaustion}}
-                                {{#if (lt @index 3)}}
+                                {{#each exhaustion.pips}}
+                                {{#if (lt @index ../exhaustion.mid)}}
```
arbron, line 152: Rather than doing this check in handlebars, maybe do something similar to what is done for the abilities on the new sheet and reduce them into a `left` and `right` array during data preparation. That might leave a bit more room for more tweaking of layout (when we get the ability to modify rendering context with Application V2).

arbron (changes requested):

module/documents/active-effect.mjs
```diff
@@ -364,15 +365,30 @@ export default class ActiveEffect5e extends ActiveEffect {
     const actor = app.object.actor;
     const level = foundry.utils.getProperty(actor, "system.attributes.exhaustion");
     if ( Number.isFinite(level) && (level > 0) ) {
+      const img = this._getExhaustionImage(level);
```
arbron, line 368: Since this method is a hook handler, `this` is the `Hooks` singleton rather than `ActiveEffect5e`, so we'll need to manually refer to that.

```suggestion
      const img = ActiveEffect5e._getExhaustionImage(level);
```

arbron (approved):

PR #2962, opened 2024-02-08
Added compatibility with v12 condition filters.

Feu-Secret:
- Urgent: fixed a *bug* in v12 token ring batched shader to avoid crashing into a v11 environment.
- Added compatibility with v12 filter effects (and thus, invisibility that needed a special treatment)


Feu-Secret:
@arbron @Fyorl I'll merge this PR because it is critical.

PR #2965, opened 2024-02-08
Don't apply effects directly from owned items to actors

schultzcole:
WIP:

- [x] add a migration that does the following:
  - for any actors that have spells or spell scrolls with a `disabled: false`, `transfer: true` active effect, create a copy of that effect directly on the actor with the original effect as the source.
  - for all effects on a spell or spell scroll, disable the effect, disable transfer, and reset startTime/startRound/startTurn
- [ ] when applying an effect to an actor from the apply effect button in the chat, copy the effect to the actor, even if the actor owns the effect's item and `transfer == true`.
- [ ] when applying an effect to an actor, reset the effect's startTime to right now.
- [ ] don't show transfer effects in the actor effects tab.


module/migration.mjs
```diff
@@ -740,3 +794,10 @@ export async function purgeFlags(pack) {
   }
   await pack.configure({locked: true});
 }
+
+/* -------------------------------------------- */
+
+function isSpellOrScroll(item) {
+  if ( item.type === "consumable" && item.system.type.value === "scroll" ) return true;
```
Fyorl, line 801:
```suggestion
  if ( (item.type === "consumable") && (item.system.type.value === "scroll") ) return true;
```

module/migration.mjs
```diff
@@ -740,3 +794,10 @@ export async function purgeFlags(pack) {
   }
   await pack.configure({locked: true});
 }
+
+/* -------------------------------------------- */
+
+function isSpellOrScroll(item) {
```
Fyorl, line 800: Missing JSDoc

Fyorl:
This looks good to me. I think 3.1 is the right target for this since the migration has a fairly large blast radius. We have some other AE-related things coming in 3.1 that should help contextualise this PR for users.

system.json
```diff
@@ -215,7 +215,7 @@
   "primaryTokenAttribute": "attributes.hp",
   "background": "systems/dnd5e/ui/official/dnd5e-background.webp",
   "flags": {
-    "needsMigrationVersion": "3.0.0",
+    "needsMigrationVersion": "3.0.2",
```
Fyorl, line 218:
```suggestion
    "needsMigrationVersion": "3.1.0",
```

schultzcole:
> Thanks for taking care of this. I've approved it but it looks like it needs a rebase. Let me know if you're happy to do that, otherwise I can take care of it.

I can rebase it tonight!

PR #2974, opened 2024-02-08
[#2845] Display temp max on character sheet HP bar

### LINKED ISSUE #2845: Add visual indication to character sheet when Temp HP Max is modified
The Token Bar has some indicators that the Max HP has been modified, it would be great to have that also on the Actor Sheet

Positive Temp HP Max
image

Negative Temp HP Max
image

Additional coloration like the token would be nice, but even appending the Temp Max HP would be useful
image

arbron:
Screenshot 2024-02-08 at 10 34 43

Closes #2845

### DIFF
--- a/less/v2/apps.less
+++ b/less/v2/apps.less
@@ -358,12 +358,23 @@
     }
 
     .label {
-      color: var(--color-text-light-0);
+      width: 100%;
+      display: flex;
+      align-items: center;
+      gap: .25rem;
       position: relative;
-      padding-left: .5rem;
+      padding-inline-start: .5rem;
+      padding-inline-end: .75rem;
+      color: var(--color-text-light-0);
 
+      &[hidden] { display: none; }
       .value, .max { text-shadow: 0 0 4px var(--dnd5e-color-gold); }
       .separator { color: var(--dnd5e-color-gold); }
+      .bonus {
+        margin-inline-start: auto;
+        font-size: var(--font-size-14);
+        opacity: .8;
+      }
     }
   }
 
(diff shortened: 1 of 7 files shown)


Fyorl:
I think we should show the total hit points on the meter rather than having the base hit points with a (+/- X). Part of the advantage of using a VTT is that it does these sorts of calculations for you, so that you can see exactly how many max hit points your character currently has. If we have an ActiveEffect adding +5 to an ability score, we don't display it as 15 (+5), for example.

If we're worried about people forgetting they have a temporary adjustment applied, I might suggest something like this which shows a bonus indicator. There's the danger that a user thinks they have 25 HP + another 5 of course, so it's not perfect, but the best I've been able to come up with for now. Any further suggestions/iteration are welcome.

image

We add an additional `<span class="bonus">+5</span>` and the styles are just:
```css
.dnd5e2 .meter .label {
  color: var(--color-text-light-0);
  position: relative;
  padding: 0 0.75rem 0 0.5rem;
  display: flex;
  gap: 0.5rem;
  width: 100%;
  align-items: center;
}

.dnd5e2 .meter .label .base {
  margin-left: auto;
  font-size: var(--font-size-14);
  opacity: .8;
}
```

krbz999:
> Any further suggestions/iteration are welcome.

When affected by `hp.tempmax`, additionally changing the maximum to a blueish color could help indicate that it is not the default maximum. (In addition to the inserted span.)

arbron:
Restyled and added `hp.effectiveMax` to avoid having to recalculate that value throughout the code like we were before.

Modified the color based on whether the temp is positive or negative, here's how that looks:

Screenshot 2024-02-09 at 09 48 55
Screenshot 2024-02-09 at 09 49 04

Fyorl (approved):

PR #2989, opened 2024-02-09
[#2918] Add modifiers Set to dice scale values

### LINKED ISSUE #2918: If an active effect modifies a dice scale value or any of its specific properties (.number or .faces) by appending a string, the remaining properties are completely lost
For example, if you have a scale value `@scale.steve.dice`, and an active effect with `system.scale.steve.dice.faces | Add | min2`, then you can no longer refer to its number, i.e., `@scale.steve.dice.number`. It seems like the full derived scale value on the actor is replaced with just a string.

Note: numbers work fine, this seems to only be an issue with strings such as `min2` or `r<3`, etc.

krbz999:
Closes #2918.

This adds a new property, `modifiers`, to dice scale values, intended for properties like `min2` or `r<3`.

Using this ensures that scale values won't get pummeled into strings by appending such onto the scale value itself, or onto `.faces` or `.number`.

Adds the `dnd5e.utils.isValidDiceModifier` method, and the `denom` and `mods` getters to dice scale values.

image

### DIFF
--- a/module/data/advancement/scale-value.mjs
+++ b/module/data/advancement/scale-value.mjs
@@ -226,9 +226,11 @@ export class ScaleValueTypeCR extends ScaleValueTypeNumber {
 export class ScaleValueTypeDice extends ScaleValueType {
   /** @inheritdoc */
   static defineSchema() {
+    const fields = foundry.data.fields;
     return {
-      number: new foundry.data.fields.NumberField({nullable: true, integer: true, positive: true}),
-      faces: new foundry.data.fields.NumberField({required: true, integer: true, positive: true})
+      number: new fields.NumberField({nullable: true, integer: true, positive: true}),
+      faces: new fields.NumberField({required: true, integer: true, positive: true}),
+      modifiers: new fields.SetField(new fields.StringField({required: true}))
     };
   }
 
@@ -275,7 +277,7 @@ export class ScaleValueTypeDice extends ScaleValueType {
    */
   get die() {
     if ( !this.faces ) return "";
-    return `d${this.faces}`;
+    return `d${this.faces}${Array.from(this.modifiers).filterJoin("")}`;
   }
 
   /* -------------------------------------------- */
--- a/module/data/advancement/scale-value.mjs
+++ b/module/data/advancement/scale-value.mjs
@@ -272,12 +272,36 @@ export class ScaleValueTypeDice extends ScaleValueType {
   /* -------------------------------------------- */
 
   /**
-   * The die value to be rolled with the leading "d" (e.g. "d4").
+   * The entire die, with leading "d" and any modifiers, e.g., "d4" or "d4r1".
    * @type {string}
    */
   get die() {
     if ( !this.faces ) return "";
-    return `d${this.faces}${Array.from(this.modifiers).filterJoin("")}`;
+    return `d${this.faces}${this.mods}`;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * The die modifiers.
+   * @type {string}
+   */
+  get mods() {
+    if ( !this.modifiers ) return "";
+    return this.modifiers.reduce((acc, mod) => {
+      return acc + (dnd5e.utils.isValidDieModifier(mod) ? mod : "");
+    }, "");
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * The die value to be rolled with the leading "d" (e.g. "d4").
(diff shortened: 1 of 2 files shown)


module/data/advancement/scale-value.mjs
```diff
@@ -275,7 +277,7 @@ export class ScaleValueTypeDice extends ScaleValueType {
    */
   get die() {
     if ( !this.faces ) return "";
-    return `d${this.faces}`;
+    return `d${this.faces}${Array.from(this.modifiers).filterJoin("")}`;
```
Fyorl, line 280: Maybe modifiers should only be added on the `formula` getter? I think it's useful to be able to get the `die` value without any associated modifiers. Perhaps this breaks some use case for the modifiers though?

krbz999: This can be achieved just by using `.faces`. I think the current approach is more correct since these are modifiers on the die, not modifiers on the formula. At least that is the intent.

krbz999: Even so, the idea is to be able to modify `@scale.steve.pool.die` which is in use in, say, a damage formula, without users always having to always write something like `@scale.steve.pool.die@scale.steve.pool.modifiers`.

Fyorl: It sounds like what you actually want is some getter that returns `formula` but with `number` hard-coded to '1'. Maybe something like `one`, e.g. `@scale.steve.pool.one`?

krbz999: Not quite. There are also cases where one would want two of the dice, not just a single one. I believe there is a monk subclass ki feature that deals damage equal to two of their Martial Arts dice. Conversely, is there a use case for getting the die without a prepended number and excluding modifiers as opposed to getting the die without a prepended number and including modifiers?

krbz999: To hopefully clarify the issue:

Using a dice scale value is a good way of having both a "pool" and a scaling die size. It's a very nice way of using only a single scale value for, e.g., Battle Master Superiority Dice; you can use `.dice.number` for your size of the pool, and `.dice.die` for the actual damage (or other utility) roll that they are used for.

We allow for `ActiveEffect`s to modify scale values, and this works without issues on dice scale values if the effect value is numeric and targets `.faces` or `.number`.

The issue is when the effect value is not numeric. I believe we should still allow them to be non-numeric, since they can be targeting `.dice.die` and `.dice`, which are already strings. The problem becomes pretty clear then, because this turns the entire scale value (on the actor) into a singular string, and you then lose access to `.dice.number` and `.dice.faces`.

This approach here says to target `.dice.modifiers` which then are more gracefully appended and show in both `.dice.die` and thus also in `.dice` itself (the full formula). I think that would be the expected behavior, rather than using a different scale value such as `dice.one` or a combination of several scale values.

Fyorl: Is it really expected behaviour that `dice.die` returns something like `d4r=1`? Because, unless I'm misunderstanding, that looks very much like a formula expression to me, and not a die denomination.

krbz999: To me it looks like a die with modifiers, and to me modifiers are part of the die, but we may just be splitting hairs at this point.

krbz999: @Fyorl How would you feel about introducing a `@scale.steve.dice.denom` getter instead, which returns just the die, with no number and no modifiers?

Fyorl: > @Fyorl How would you feel about introducing a `@scale.steve.dice.denom` getter instead, which returns just the die, with no number and no modifiers?

`denom` better matches `DiceTerm#denomination`, but the issue is still that `dice.die` will return a die with modifiers attached. Modifiers are part of the overall dice term expression, not part of the individual die. You cannot roll a `d4r=1` or a `d6dl`. You are rolling a d4 or a d6, with some additional rules on resolving the value you roll, some of which are meaningless without knowing the number of dice being rolled.

I understand the desire to avoid writing `@scale.dice.die@scale.dice.modifiers` each time, and having this change 'just work' with existing formulae that have been written. I do not have a better suggestion though. `dieWithModifiers` or `numberless` are not really hitting the mark so I will solicit some additional suggestions and come back to this PR in a little while, and if no-one has any better ideas then we'll go with `die`.

Fyorl (approved):

Issue #2992, opened 2024-02-09
Use a different icon to indicate a spell is 'always prepared'


Fyorl:
The new sheets currently show no indication of this, which is an issue as it looks like the preparation state can be toggled off when it can't. Previously it used to show as red, but we should use a different icon entirely for accessibility.

schultzcole:
Yeah, I think different icons for "always prepared", "prepared", and "not prepared" would be best for accessibility. Maybe also an icon for "known/inherent"?

etiquettestartshere:
I'll throw in my own personal little wish to this one: make the attunement icon different from any of the above.

PR #2996, opened 2024-02-09
[#2992] Use different icons for unprepared & always prepared states

arbron:
Showing "Always Prepared", "Prepared", and "Unprepared" spells:

Screenshot 2024-02-09 at 10 24 16

Here is an alternative icon:

Screenshot 2024-02-09 at 10 20 52


Fyorl (approved):
I think I prefer the alternative, with the icons progressing from 'filled' to 'empty'.

Issue #2947, opened 2024-02-07
`Item5e#deleteDialog` fails for module actor types without a `system.details` property.


krbz999:
![image](https://github.com/foundryvtt/dnd5e/assets/50169243/b43c283a-dabc-444f-8896-102003d8c87a)

System has a check only for the actor type being `group`. There should probably more simply be a check for `"details" in this.actor.system`.

PR #2997, opened 2024-02-09
[#2947] Add metadata.supportsAdvancement, fix issue with deleteDialog failing

arbron:
Move singleton property from `SystemDataModel#metadata` to `ItemDataModel#metadata` and reworked data model metadata as frozen object with proper default values.

Added `supportsAdvancement` metadata option to `ActorDataModel#metadata` to indicate actor types that can have advancement applied. Using the new metadata property in `deleteDialog` should ensure no errors for module-provided actor types.

### DIFF
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -949,21 +949,20 @@ export default class ActorSheet5e extends ActorSheetMixin(ActorSheet) {
     const stacked = this._onDropStackConsumables(itemData);
     if ( stacked ) return false;
 
-    // Ensure that this item isn't violating the singleton rule
-    // TODO: When v10 support is dropped, this will only need to be handled for items with advancement
-    const dataModel = CONFIG.Item[dnd5e.isV10 ? "systemDataModels" : "dataModels"][itemData.type];
-    const singleton = dataModel?.metadata.singleton ?? false;
-    if ( singleton && this.actor.itemTypes[itemData.type].length ) {
-      ui.notifications.error(game.i18n.format("DND5E.ActorWarningSingleton", {
-        itemType: game.i18n.localize(CONFIG.Item.typeLabels[itemData.type]),
-        actorType: game.i18n.localize(CONFIG.Actor.typeLabels[this.actor.type])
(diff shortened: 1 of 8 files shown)


module/data/abstract.mjs
```diff
@@ -69,15 +69,16 @@ export default class SystemDataModel extends foundry.abstract.TypeDataModel {
...
-   * @property {boolean} [singleton]                  Should only a single item of this type be allowed on an actor?
    * @property {typeof DataModel} [systemFlagsModel]  Model that represents flags data within the dnd5e namespace.
    */
 
   /**
    * Metadata that describes this DataModel.
    * @type {SystemDataModelMetadata}
    */
-  static metadata = {};
+  static metadata = Object.freeze({
```
Fyorl, line 79: Do we know of any modules making use of `SystemDataModel`? If they add their own sub-types they might, right? This could be a breaking change in that case.

Fyorl (changes requested):

module/applications/actor/base-sheet.mjs
```diff
@@ -962,7 +962,8 @@ export default class ActorSheet5e extends ActorSheetMixin(ActorSheet) {
     }
 
     // Bypass normal creation flow for any items with advancement
-    if ( itemData.system.advancement?.length && !game.settings.get("dnd5e", "disableAdvancements") ) {
+    if ( this.actor.system.metadata?.supportsAdvancement && itemData.system.advancement?.length
```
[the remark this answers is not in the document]
arbron: Aw yeah, that code can be moved down into the advancement handling block.

module/data/abstract.mjs
arbron: I can skip freezing this data for now, though maybe we can do that in `3.1.0` to avoid module subtypes accidentally modifying the metadata of other types. I've left the metadata on the actual concrete data models frozen because I doubt anyone is inheriting from `BackgroundData` or `RaceData`.

Fyorl: Yeah that sounds good to me

Fyorl (approved):

PR #2999, opened 2024-02-09
[#2867] Add context menus to race, background, classes, & subclasses

### LINKED ISSUE #2867: Add context menus to class and subclass items on new sheet
On the new sheet that 3.0 introduced, I am unable to edit the class or subclass items. It doesn't seem to matter whether I'm in Play or Edit mode. With all the other items in the Features tab, I can right-click on them or use the new vertical ellipsis to edit them. With the class and subclass items, it appears that all I can do is delete them in Edit mode by clicking on the icon.

arbron:
Adds a context menu on ActorSheet5eCharacter2 targeting any element with `data-item-id` that is inside `.pills-lg`, which allows it to target all the concept items that exist outside the normal inventory elements.

In order to properly activate the context menus on these elements it fetches the `InventoryElement` on the features tab and uses its `_onOpenContextMenu`.

Closes #2867

### DIFF
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -1,8 +1,9 @@
-import ActorSheet5eCharacter from "./character-sheet.mjs";
-import * as Trait from "../../documents/actor/trait.mjs";
-import Tabs5e from "../tabs.mjs";
-import { simplifyBonus, staticID } from "../../utils.mjs";
 import CharacterData from "../../data/actor/character.mjs";
+import * as Trait from "../../documents/actor/trait.mjs";
+import { simplifyBonus, staticID } from "../../utils.mjs";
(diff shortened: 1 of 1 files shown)


module/applications/actor/character-sheet-2.mjs
```diff
@@ -607,6 +608,12 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
       }
     });
 
+    // Apply special context menus for items outside inventory elements
+    const featuresElement = html[0].querySelector(`[data-tab="features"] ${this.options.elements.inventory}`);
```
Fyorl, line 612: Awkward, but I don't see a better way if we want to retain agnosticism for the `InventoryElement` being placed in either an ActorSheet or ItemSheet.

Fyorl (approved):

PR #3001, opened 2024-02-09
[#2876] Fix context menus being displayed off the bottom of the screen

### LINKED ISSUE #2876: Unable to delete items/spells that are at the bottom of the character sheet at its default size
When a feature, inventory item, or spell is at the very bottom of the character sheet, the context menu descends outside the window, making the lower buttons unclickable.

image

This can be avoided by resizing the character sheet to be smaller vertically, but on a 1080p monitor the sheet takes up the entire window height by default.

Running Foundry build 11.315, D&D5e version 3.0.1

arbron:
Previously the tooltip sizing wasn't available because it didn't yet exist in the DOM. Moving the `appendChild` call to the top of the method resolved that.

This also required anchor the tooltip by its bottom edge when expanding up to ensure the display/hiding animation occurred in the correct direction.

Closes #2876

### DIFF
--- a/module/applications/context-menu.mjs
+++ b/module/applications/context-menu.mjs
@@ -5,6 +5,7 @@
 export default class ContextMenu5e extends ContextMenu {
   /** @override */
   _setPosition([html], [target]) {
+    document.body.appendChild(html);
     const { clientWidth, clientHeight } = document.documentElement;
     const { width, height } = html.getBoundingClientRect();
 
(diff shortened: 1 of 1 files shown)


Fyorl (approved):
The PR says 'tooltip' but this only addresses context menus. Is that just a mistake, or are we missing some other code in this PR maybe? Code for context menus looks good to me though.

arbron:
> The PR says 'tooltip' but this only addresses context menus. Is that just a mistake, or are we missing some other code in this PR maybe? Code for context menus looks good to me though.

Just miswrote the title :)

Issue #2979, opened 2024-02-08
Applying healing to a selected token from a chat message triggers scrolling damage numbers despite their being at full hp


krbz999:
Repro:
1. apply damage via chat message
2. apply same as healing via same chat message
3. repeat step 2 (scrolling number should not appear, but does)

Seems to be only possible to do once, step 3 can't be repeated.

Alternative repro steps:
1. Give actor tempmax bonus.
2. Put actor back to new max hp.
3. Remove the tempmax bonus.
4. Apply healing.

PR #3002, opened 2024-02-09
[#2979] Clamp damage/healing amount so it doesn't exceed HP bounds

arbron:
Previously the scrolling damage numbers and token flashes would be calculated based on the damage amount, without regards to how much HP the actor currently has. This results in showing floating numbers even when nothing about the actor's HP has been modified.

This change clamps the change to HP based on their current HP (for damage) or the negative of their current damage (for healing). To assist with this `hp.damage` has been added during hit points preparation for easy reference.

One limitation of this is that the damage numbers reflect only the damage applied, not the total damage taken, which might underrepresent massive hits. An alternative solution is to do the same clamping, but only check if `deltaTemp + deltaHP` is zero and don't pass `dhp` to the update function if nothing was change (or just skip the update all together).

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1129,7 +1129,6 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       "system.attributes.hp.temp": hp.temp - deltaTemp,
       "system.attributes.hp.value": hp.value - deltaHP
     };
-    amount = deltaTemp + deltaHP;
 
     /**
      * A hook event that fires before damage is applied to an actor.
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -674,6 +674,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
     hp.effectiveMax = hp.max + (hp.tempmax ?? 0);
     hp.value = Math.min(hp.value, hp.effectiveMax);
+    hp.damage = hp.effectiveMax - hp.value;
     hp.pct = Math.clamped(hp.effectiveMax ? (hp.value / hp.effectiveMax) * 100 : 0, 0, 100);
   }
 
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -673,6 +673,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     }
 
     hp.value = Math.min(hp.value, hp.max + (hp.tempmax ?? 0));
+    hp.damage = hp.max + (hp.tempmax ?? 0) - hp.value;
```
krbz999, line 676: Is `hp.temp` left out on purpose?

arbron: Well I suppose that depends on what you see this damage number as. The way it is currently being used in `applyDamage` it needs to know how many HP the player's current hit points are below their maximum for calculating healing.

`15 (max) - 10 (current) = 5 (damage able to be healed)`
`15 (max) + 2 (temp max) - 10 (current) = 7 (damage able to be healed)`

Temp HP is something that is sort of orthogonal to this. While it can absorb damage, it shouldn't affect healing in any way.

`15 (max) - 10 (current) - 5 (temp) = 0 (you're totally fine! nothing to heal here!)`

Fyorl:
> An alternative solution is to do the same clamping, but only check if `deltaTemp + deltaHP` is zero and don't pass `dhp` to the update function if nothing was change (or just skip the update all together).

This seems like the best of both worlds to me: We fix the issue but retain full information otherwise. I think the update will already be skipped with this clamping in place due to the `diff: true` default.

arbron:
> This seems like the best of both worlds to me: We fix the issue but retain full information otherwise. I think the update will already be skipped with this clamping in place due to the `diff: true` default.

Okay, I've swapped back to the full `amount` for the reported DHP and it works to prevent displaying the numbers if nothing has actually changed.

Fyorl (approved):

Issue #2885, opened 2024-02-04
player effects page leaks unidentified items


jtracey:
If a player is given an unidentified item with effects, the item's name and effects are masked in the Inventory page of their character sheet, but are all completely visible in the effects page.

Fyorl:
The relationship between attuning and identifying needs more work. There are also (very rarely) items that confer magical properties on wielders while they are not attuned, and there's no way to distinguish at the moment. That will probably have to wait for the work on activities. For now, hiding effects of unidentified items probably gets us most of the way there and is consistent with the item sheet.

PR #3004, opened 2024-02-10
[#2885] Hide active effect names when from unidentified item

arbron:
Just changes the names of active effects that belong to items that are unidentified.

Screenshot 2024-02-09 at 16 33 38

The details of the active effect can still be determined by opening its sheet, but modifying that behavior is a much larger change.

### DIFF
--- a/module/applications/components/effects.mjs
+++ b/module/applications/components/effects.mjs
@@ -104,6 +104,7 @@ export default class EffectsElement extends HTMLElement {
 
     // Iterate over active effects, classifying them into categories
     for ( const e of effects ) {
+      if ( (e.parent.system?.identified === false) && !game.user.isGM ) continue;
       if ( e.isSuppressed ) categories.suppressed.effects.push(e);
       else if ( e.disabled ) categories.inactive.effects.push(e);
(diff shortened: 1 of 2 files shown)


Fyorl:
The logic for entirely hiding AEs from unidentified (owned) Items should be fairly straightforward, no?

arbron:
> The logic for entirely hiding AEs from unidentified (owned) Items should be fairly straightforward, no?

Well if we are talking about conditionally hiding parts of the Active Effect configuration interface the same way we do with the item sheet that would probably involve a template replacement or a `render` hook to modify it after the fact.

arbron:
Or do you mean completely hiding the effect on the sheet? That would be easier.

Fyorl:
> Or do you mean completely hiding the effect on the sheet? That would be easier.

Yeah, just hiding them altogether. It's not a perfect solution for various reasons, but maybe it's good enough for 90% of cases.

arbron:
> Yeah, just hiding them altogether. It's not a perfect solution for various reasons, but maybe it's good enough for 90% of cases.

Now completely hidden.

Fyorl (approved):

Issue #2995, opened 2024-02-09
Wildshape / polymorph error with dnd5e 3.0.1


neovatar:
Wildshape / polymorph throws errors on every second try.

Steps to reproduce:

1. Create new campaign with dnd5e 3.0.1 ruleset and no modules
2. Drag starter hero druid and wolf from the SRD compendiums to actors
3. Drag druid token on canvas
4. Open druid character sheet
5. Drag the wolf actor on the character sheet
6. Click "Wild Shape" on the polymorph dialogue
7. Druid Token on canvas is transformed - actor for transformed druid is created
8. Click the little arrows to "Restore Transformation" on the transformed actor character sheet
9. Druid token on canvas is transformed back to druid - actor for transformed druid is deleted
10. Drag the wolf actor on the character sheet a second time
11. Druid Token on canvas is **NOT** transformed - actor for transformed druid is created
12. Try again - it works again - it errors on every second try

The following error is logged in console:
```
foundry.js:68326 Cannot add property originalActor, object is not extensible
fetch @ foundry.js:68326
notify @ foundry.js:68223
error @ foundry.js:68260
#preUpdateDocumentArray @ foundry.js:13659
_updateDocuments @ foundry.js:13449
update @ commons.js:8677
await in update (async)
updateDocuments @ commons.js:8001
updateEmbeddedDocuments @ commons.js:8233
transformInto @ actor.mjs:2793
await in transformInto (async)
callback @ base-sheet.mjs:811
submit @ foundry.js:65213
_onClickButton @ foundry.js:65141
dispatch @ jquery.min.js:2
y.handle @ jquery.min.js:2
foundry.js:753 TypeError: Cannot add property originalActor, object is not extensible
    at _mergeInsert (commons.js:2787:17)
    at mergeObject (commons.js:2761:12)
    at _mergeUpdate (commons.js:2803:14)
    at mergeObject (commons.js:2760:41)
    at #updateField (commons.js:7354:9)
    at #updateData (commons.js:7302:26)
    at TokenDocument5e.updateSource (commons.js:7258:30)
    at TokenDocument5e.updateSource (commons.js:12431:26)
    at #preUpdateDocumentArray (foundry.js:13657:20)
    at ClientDatabaseBackend._updateDocuments (foundry.js:13449:73)
onError @ foundry.js:753
#preUpdateDocumentArray @ foundry.js:13660
_updateDocuments @ foundry.js:13449
update @ commons.js:8677
await in update (async)
[...]

PR #3005, opened 2024-02-10
[#2995] Fix transformation reversion errors

arbron:
Adds the flags that track transformation status into `TokenSystemFlags` so they can be properly tracked and merged during transformation.


Fyorl (approved):
Good job getting to the bottom of this.

PR #3010, opened 2024-02-10
Fix: `ActiveEffect` changes are persisted when submitting the `TraitSelector` sheet

dev7355608:
* Disable choices that have been removed by an override.
* Disable `custom` if overridden.
* Disable `bypasses` if overridden.

### DIFF
--- a/module/applications/actor/trait-selector.mjs
+++ b/module/applications/actor/trait-selector.mjs
@@ -94,14 +94,29 @@ export default class TraitSelector extends BaseConfigSheet {
   /** @inheritdoc */
   _getActorOverrides() {
     const overrides = super._getActorOverrides();
-    const path = Trait.changeKeyPath(this.trait);
-    const src = new Set(foundry.utils.getProperty(this.document._source, path));
-    const current = foundry.utils.getProperty(this.document, path);
-    const delta = current.difference(src);
-    for ( const choice of delta ) {
-      overrides.push(`choices.${choice}`);
-    }
+    const path = Trait.actorKeyPath(this.trait);
+    this.#addOverriddenChoices("choices", Trait.changeKeyPath(this.trait), overrides);
+    this.#addOverriddenChoices("bypasses", `${path}.bypasses`, overrides);
+    const pathCustom = `${path}.custom`;
+    const sourceCustom = foundry.utils.getProperty(this.document._source, pathCustom);
+    const currentCustom = foundry.utils.getProperty(this.document, pathCustom);
+    if ( sourceCustom !== currentCustom ) overrides.push(pathCustom);
     return overrides;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Add choices that have been overridden.
+   * @param {string} prefix       The initial form prefix under which the choices are grouped.
+   * @param {string} path         Path in actor data.
+   * @param {string[]} overrides  The list of fields that are currently modified by Active Effects. *Will be mutated.*
+   */
+  #addOverriddenChoices(prefix, path, overrides) {
+    const source = new Set(foundry.utils.getProperty(this.document._source, path) ?? []);
+    const current = foundry.utils.getProperty(this.document, path) ?? new Set();
+    const delta = current.symmetricDifference(source);
+    for ( const choice of delta ) overrides.push(`${prefix}.${choice}`);
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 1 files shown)


module/applications/actor/trait-selector.mjs
```diff
@@ -94,18 +94,33 @@ export default class TraitSelector extends BaseConfigSheet {
...
+  /**
+   * Add choices that have been overridden.
+   * @param {string} prefix       The initial form prefix under which the choices are grouped.
+   * @param {string} path         Path in actor data.
+   * @param {string[]} overrides  The list of fields that are currently modified by Active Effects. *Will be mutated.*
+   */
+  #addOverriddenChoices(prefix, path, overrides) {
+    const source = new Set(foundry.utils.getProperty(this.document._source, path) ?? []);
+    const current = foundry.utils.getProperty(this.document, path) ?? new Set();
+    const delta = current.symmetricDifference(source);
```
dev7355608, line 118: Was `difference`, but needs to `symmetricDifference`: an override my remove a choice.

module/applications/actor/trait-selector.mjs
```diff
@@ -94,18 +94,33 @@ export default class TraitSelector extends BaseConfigSheet {
...
-    for ( const choice of delta ) {
-      overrides.push(`choices.${choice}`);
-    }
+    const path = Trait.actorKeyPath(this.trait);
+    this.#addOverriddenChoices("choices", Trait.changeKeyPath(this.trait), overrides);
+    this.#addOverriddenChoices("bypasses", `${path}.bypasses`, overrides);
+    const pathCustom = `${path}.custom`;
+    const sourceCustom = foundry.utils.getProperty(this.document._source, pathCustom);
+    const currentCustom = foundry.utils.getProperty(this.document, pathCustom);
+    if ( sourceCustom !== currentCustom ) overrides.push(pathCustom);
```
dev7355608, line 103: `bypasses` and `custom` wasn't handled here before, and so overrides of those are currently persisted.

module/applications/actor/trait-selector.mjs
```diff
@@ -154,7 +175,7 @@ export default class TraitSelector extends BaseConfigSheet {
     const path = Trait.actorKeyPath(this.trait);
     const data = foundry.utils.getProperty(this.document, path);
 
-    this._prepareChoices("choices", `${path}.value`, formData);
+    this._prepareChoices("choices", Trait.changeKeyPath(this.trait), formData);
```
dev7355608, line 178: Seems like this another potential bug that `${path}.value` was used there instead of `Trait.changeKeyPath(this.trait)` as above.

module/applications/actor/trait-selector.mjs
```diff
@@ -133,18 +148,24 @@ export default class TraitSelector extends BaseConfigSheet {
...
-      if ( formData[key] ) chosen.push(key.replace(`${prefix}.`, ""));
+      if ( formData[key] ) chosen.add(key.replace(`${prefix}.`, ""));
       delete formData[key];
     }
-    formData[path] = chosen;
+
+    // Add choices from the source that have been removed by an override: if we didn't, the override would be persisted
+    const source = new Set(foundry.utils.getProperty(this.document._source, path));
+    const current = foundry.utils.getProperty(this.document, path);
+    for ( const choice of source.difference(current) ) chosen.add(choice);
```
dev7355608, line 166: We need to make sure selected, overridden choices remain selected.

module/applications/actor/trait-selector.mjs
```diff
@@ -133,18 +148,24 @@ export default class TraitSelector extends BaseConfigSheet {
...
+
+    formData[path] = Array.from(chosen).sort((a, b) => a.localeCompare(b, "en"));
```
Fyorl, line 168: Should this be `game.i18n.lang`?

```suggestion
    formData[path] = Array.from(chosen).sort((a, b) => a.localeCompare(b, game.i18n.lang));
```

Fyorl (approved):
Looks good, just one comment

module/applications/actor/trait-selector.mjs
arbron: Since this is sorting the `keys`, I'm not sure it would make a difference. Could probably even just use a simple `Array.from(chosen).sort()` here since the specific order is unimportant, only the consistency of the sorting.

PR #3012, opened 2024-02-10
[#2224] Only automatically set item sheet height on tab change

### LINKED ISSUE #2224: Details tab in Item sheets persistently resizes the sheet window
The Details tab in Item sheets persistently resizes the sheet window when it cannot fit the Details tab content into the window. The new window size is kept when switching between tabs.

[Desktop 2023.02.26 - 03.27.26.01.webm](https://user-images.githubusercontent.com/73297245/221389027-932bbb93-3c12-4848-ba7a-f41880f5057e.webm)

Foundry 10.291
DnD5e 2.1.5
Edge latest, self-hosted server

schultzcole:
Previously we forced the height of the item sheet to certain values whenever the app was moved (due to overriding `setPosition`). The approach in this PR only automatically changes the height when the tab changes, rather than any time the app moves.

We can also just rely on the standard auto height sizing to adjust the size of the sheet for the content (like for a class or subclass sheet). This results in fewer non-standard cases and is more consistent UX IMO.

Closes #2224

[changing tabs on the item sheet](https://github.com/foundryvtt/dnd5e/assets/1226888/fe0e96b1-e346-44f6-8559-de1835d9e345)

### DIFF
--- a/module/applications/item/item-sheet.mjs
+++ b/module/applications/item/item-sheet.mjs
@@ -16,15 +16,6 @@ export default class ItemSheet5e extends ItemSheet {
   constructor(...args) {
     super(...args);
 
-    // Expand the default size of the class sheet
-    if ( this.object.type === "class" ) {
-      this.options.width = this.position.width = 600;
-      this.options.height = this.position.height = 680;
-    }
-    else if ( this.object.type === "subclass" ) {
-      this.options.height = this.position.height = 540;
-    }
-
     this._accordions = this._createAccordions();
   }
 
@@ -34,7 +25,6 @@ export default class ItemSheet5e extends ItemSheet {
   static get defaultOptions() {
     return foundry.utils.mergeObject(super.defaultOptions, {
       width: 560,
-      height: 500,
       classes: ["dnd5e", "sheet", "item"],
       resizable: true,
       scrollY: [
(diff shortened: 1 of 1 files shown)


Fyorl (approved):
It seems like there would be some tension between a sheet being `resize: true`, but then setting the the `height` to `auto` whenever the tab is changed. I suppose horizontal resizing is still supported.

This behaviour seems like an improvement to what we have currently though. At least until item sheets are given a v2 update.

schultzcole:
> It seems like there would be some tension between a sheet being `resize: true`, but then setting the the `height` to `auto` whenever the tab is changed. I suppose horizontal resizing is still supported.
> 
> This behaviour seems like an improvement to what we have currently though. At least until item sheets are given a v2 update.

Interestingly, doing this:

```js
this.setPosition({ height: "auto" });
// can manually resize afterward
```

doesn't conflict with `resize: true` at all as far as I can tell. It sets the height automatically but allows resizing afterwards. On the other hand, if you set height to auto in the options, then it would prevent resizing:

```js
this.options.height = "auto";
this.setPosition();
// can't manually resize afterward
```

Fyorl:
> Interestingly, doing this:
> 
> ```js
> this.setPosition({ height: "auto" });
> // can manually resize afterward
> ```
> 
> doesn't conflict with `resize: true` at all as far as I can tell. It sets the height automatically but allows resizing afterwards.

I meant more as in, what is the point of allowing resizing (vertically) if changing the tab is just going to snap back to a fixed size anyway? If manually resizing the tab would be remembered when switching back to the tab then I think that would make sense, but that's not what I observed.

It's fine though, as I said, I think it's still an improvement. We don't have to solve everything here, especially since item sheets will (eventually) be getting a redesign.

Issue #1941, opened 2022-11-24
Feature Request: add a dnd5e.preRollDeathSave hook or change abort behavior


itteerde:
`Actor5e.rollDeathSave` starting with 

```
    // Display a warning if we are not at zero HP or if we already have reached 3
    if ( (this.system.attributes.hp.value > 0) || (death.failure >= 3) || (death.success >= 3) ) {
      ui.notifications.warn(game.i18n.localize("DND5E.DeathSaveUnnecessary"));
      return null;
    }
```

prevents me from having player continue rolling Death Saves (I have a slightly modified handlebars and am extending the CharacterSheet, so hiding the numbers is covered in my case) when they are stabilized or dead (which my GM likes for maximum tension until the healer arrives). With the abort being before 
```
if ( Hooks.call("dnd5e.preRollDeathSave", this, rollData) === false ) return;
````

I need to pretty much reimplement rolling death saves. Adding a hook that fires at the very top (which I think a preSomething should do) would make my useCase very simple. Moving the pre Hook up is probably not an option anymore as I guess people will be relying on how it works now (actually another change we have for ToA with setting the DC to 15 already does even for us, so please consider doing the true pre hook as prePreRollDeathSave).

In general I think it is a good idea to have many preSomething hooks that are availble before any condition could lead to a return in that Something. It might be a good rule for the future to never write a preSomething after a return. If the existing stuff makes this seem to break user expectations maybe add something like calledSomething hooks right at the method's start just checking if the hooks user wants to modify what is there (for whatever reasons) before the method takes over (checking if there is something registered for the hook isn't expensive, while every case where you can't use the hook and need to overwrite the method is cause of future maintainance terror).

![image](https://user-images.githubusercontent.com/8367477/203838750-ca00730f-6b80-41c8-aa9f-167cdf94deb7.png)

itteerde:
... and preSomethings (or because already established expectations require another name whatever atTheRealStartOfSomething) should also be before any side effect of the Something (the notification in this case would invalidate our use even without the return for example).

PR #3018, opened 2024-02-11
Status effect improvements

dev7355608:
- Renamed status effect IDs:
  - `hidden` -> `hiding` (a creature may be hiding, but that doesn't mean it is actually hidden from other creatures)
  - `silence` -> `silenced` (for consistency)
- Added new status effects:
  - Ethereal (`ethereal`) (for *Etherealness* spell, *Blink* spell, and ghosts)
  - Hovering (`hovering`)
  - Stabilized (`stable`)
- Added special status effects:
  - `burrowing`: `BURROW`
  - `flying`: `FLY`
  - `hovering`: `HOVER`
- Added `DND5E.Status*` locale strings for each status effect in `DND5E.statusEffects`, so that we don't rely on core for labels.
- Added icons in `systems/dnd5e/icons/svg/statuses` for each status effect in `DND5E.statusEffects`, replacing core icons for the few effect that are still using these. The advantage is that if we change an icon in the future and all existing effects of theses status aren't stuck with the old core icon.
- Changes to status effect icons:
  - Replaced the icon of `hiding` in order to avoid confusion with the GM hidden state, which uses the same icon. Also could have caused problems with the default token icon as an overlay status effect (two mystery mans on top of each other).
  - Replaced the icon for `burrowing` with an icon of a spade (previously an icon of a cave).
- Changes to implicit statuses:
  - `sleeping` now implies `unconscious` (which implies  `incapacitated` and `prone`) (DMG 248; XGtE 77).


module/applications/actor/character-sheet-2.mjs
```diff
@@ -355,7 +355,7 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
         name, reference,
         id: k,
         icon: img ?? icon,
-        disabled: existing ? disabled : !this.actor.statuses.has(k)
```
dev7355608, line 358: Rendering conditions as enabled that are only implied by another is confusing because it gives the impression that the implied conditions can be removed. Perhaps we should have a different section on the sheet where we show the list of all status effects (not only RAW conditions) (plus encumbrance statuses) that are affecting the actor.

module/documents/active-effect.mjs
```diff
@@ -299,7 +299,10 @@ export default class ActiveEffect5e extends ActiveEffect {
     if ( !Number.isFinite(level) ) level = 1;
     this.icon = this.constructor._getExhaustionImage(level);
     this.name = `${game.i18n.localize("DND5E.Exhaustion")} ${level}`;
-    if ( level >= config.levels ) this.statuses.add("dead");
+    if ( level >= config.levels ) {
+      this.statuses.add("dead");
+      CONFIG.DND5E.statusEffects.dead.statuses?.forEach(s => this.statuses.add(s));
```
dev7355608, line 304: In case we in the future or some module add implied statuses to `dead`.

module/config.mjs
```diff
@@ -3234,7 +3262,7 @@ function patchConfig(key, fallbackKey, options) {
 
   Object.values(DND5E[key]).forEach(o => {
     if ( foundry.utils.getType(o) !== "Object" ) return;
-    o.toString = toString;
+    Object.defineProperty(o, "toString", {value: toString});
```
dev7355608, line 3265: Adding `toString` as an non-enumerable property. Otherwise `toString` appears in `CONFG.statusEffects` entries.

MotoMoto1234:
Midi has already done this on its end but there are 5 conditions that also apply incapacitated.  Petrified, stunned, unconscious, paralyzed, and dead.

krbz999:
I feel pretty strongly that there should be a level of disconnect between what the character is affected by and what the player is affected by. Attaching the blinded condition to *burrowing, petrified, sleeping, and unconscious* will in most cases result in not only the player's character being taken out of the game, but also the player entirely.

This may be technically correct, but it should not be the default; if players at a table feel strongly enough about it that a burrowing or petrified player character should also not have their player be able to tell what is happening at their table (given whichever permissions they have of other token actors for vision), then they should add that status to those conditions themselves.

dev7355608:
> I feel pretty strongly that there should be a level of disconnect between what the character is affected by and what the player is affected by. Attaching the blinded condition to _burrowing, petrified, sleeping, and unconscious_ will in most cases result in not only the player's character being taken out of the game, but also the player entirely.
> 
> This may be technically correct (I don't truly believe it is in all four cases here), but it should not be the default; if players at a table feel strongly enough about it that a burrowing or petrified player character should also not have their player be able to tell what is happening at their table (given whichever permissions they have of other token actors for vision), then they should add that status to those conditions themselves.

I think you can make this argument about the blinded condition already. A GM might not want this condition to actually blind the token so that the player can still enjoy what's happening around them. Of course in this case the GM could never use the blinded condition at all and track it somewhere else, but this wouldn't be ideal.

Not blinding the token while petrified/unconscious/dead isn't doesn't solve the problem you describe entirely. After all the encounter might move to an adjacent room for example and the petrified player would see as much of it as if blinded. I think a better solution would be to temporarily grant the petrified/unconscious/dead player observer permissions of their party members (or play with observer permission always, which might be the right call if the table doesn't want to be so strict about visibility in general).

krbz999:
> I think you can make this argument about the blinded condition already. A GM might not want this condition to actually blind the token so that the player can still enjoy what's happening around them. Of course in this case the GM could never use the blinded condition at all and track it somewhere else, but this wouldn't be ideal.

That is 1 condition, which this change would make 5 conditions instead.

(I don't see the relevance of the second paragraph.)

In any case, _unconscious_ - at the very least - should not imply blinded, deafened, and silenced; a loud enough sound can wake someone sleeping, sleep-talking is a thing, and bright lights can wake a sleeping character (and `sleeping` here is implying `unconscious`, so this would end up being incorrect).

I was not able to find any official ruling that implied that 'unaware of your surroundings' equated to being blind, deaf, and mute.

Fyorl:
> 1. `burrowing` now implies `blinded`.
> 2. `paralyzed` now implies `silenced` (_[...] can't [...] speak._)
> 3. `petrified` now implies `blinded`, `deafened`, and `silenced` (_[...] can't [...] speak, and is unaware of its surroundings._).
> 4. `sleeping` now implies `unconscious` (which implies `blinded`, `deafened`, `incapacitated`, `prone`, and `silenced`) (DMG 248; XGtE 77).
> 5. `unconcious` now implies `blinded`, `deafened`, `silenced` (_[...] can't [...] speak, and is unaware of its surroundings._).

Since `silenced` is a pseudo-condition invented by us, we should use our own judgement in applying it. I think the proposed uses here make sense to me. Sleeping is explicitly defined in the rules as has having the unconscious condition, so 4 should not be controversial. For the others, I'm personally ok with 1-3 but it's less clear cut since the rules on conditions are explicit about which other conditions also apply, and we are adding ones that are not mentioned. Since they're not mentioned in the rules, and are controversial, I would propose that we do not add them ourselves and leave it up to DMs to apply based on their own interpretation.

txm3278:
I don't think it makes sense to limit the application of implied conditions because of a potential vision issue. It makes more sense to me to have the core blinded/invisible vision effects be toggleable(defaulted on), than to avoid automated applying effects

krbz999:
> One additional comment is that, instead of a `stabilized` status, perhaps we should have a `dying` status instead? A `stabilized` creature is the same as an `unconscious` creature, but a `dying` creature is `unconscious` while also making death saves.

This could help with https://github.com/foundryvtt/dnd5e/issues/1941, acting as an alternative approach to "hiding" from the player when/if a player character is dying. Simply remove the "dying" status when the GM tells you to.

dev7355608:
I found [this](https://www.youtube.com/watch?v=u67o2kE6InA&t=1352s) which in fact confirms that "unaware of your surroundings" is not intended to imply the deafened condition. I assume the same would be true for the blinded condition.

So by RAW the GM cannot apply the blinded condition to an unconscious actor. But I think we can agree that while sleeping/unconscious we do not see anything, and the token's vision should reflect that. So I think we need to integrate unconscious/sleeping as a special status effect that suppresses token vision like blinded does.

dev7355608:
> One additional comment is that, instead of a `stabilized` status, perhaps we should have a `dying` status instead? A `stabilized` creature is the same as an `unconscious` creature, but a `dying` creature is `unconscious` while also making death saves.

The PHB mentions "stable" as a status but not "dying": you're dying if you are at 0 hit points and not stable. So I think it is more true to the rules if we have stable as a status instead of dying.

Re-reading it again I noticed that we need to rename `stabilized` to `stable`: the PHB writes "stable creature" instead of "stabilized creature" (in *Stabilizing a Creature*, *Knocking a Creature Out*, and *Spare the Dying*).

krbz999:
> But I think we can agree that while sleeping/unconscious we do not see anything, and the token's vision should reflect that.

An ideal alternative might be a condition (using unconscious/sleeping as you suggest) that reduces vision rather than removes it, blurs the canvas, or removes all tokens from the affected actor's field of view.

module/data/actor/templates/traits.mjs
```diff
@@ -124,6 +124,8 @@ export default class TraitsField {
   static prepareResistImmune() {
     if ( this.parent.hasConditionEffect("petrification") ) {
       this.traits.dr.custom = game.i18n.localize("DND5E.DamageAll");
+      Object.keys(CONFIG.DND5E.damageTypes).forEach(type => this.traits.dr.value.add(type));
```
Fyorl, line 127: I think this results in spamming the character sheet with a huge amount of damage immunity pills, which is why we opted for the 'All' custom damage immunity instead. It might be preferable to defer this until we can add special-case handling for an 'all' type in the system as part of the resistances/immunities/vulnerability automation.

It's not very often that you are petrified though so I don't feel strongly.

Fyorl (approved):
Looks good, thanks, just one comment for your consideration. Happy to go with `stable` over `dying` if that's how it's framed in the rules.

module/data/actor/templates/traits.mjs
dev7355608: I added special handling for petrification to the Actor sheet to avoid the spam.

PR #3024, opened 2024-02-11
Movement speed calculation fixes

dev7355608:
- Fixed: *Prone* condition halved the movement speed incorrectly.
- Changed: Apply flat speed reductions of encumbrance after speed halving of *Exhaustion 2*.
- Fixed: Halved movement speed was rounded (floored).
- Added: Support for metric units in movement speed calculations.

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2461,7 +2461,7 @@ patchConfig("conditionTypes", "label", { since: "DnD5e 3.0", until: "DnD5e 3.2" 
  */
 DND5E.conditionEffects = {
   noMovement: new Set(["exhaustion-5", "grappled", "paralyzed", "petrified", "restrained", "stunned", "unconscious"]),
-  halfMovement: new Set(["exhaustion-2", "prone"]),
+  halfMovement: new Set(["exhaustion-2"]),
   crawl: new Set(["prone", "exceedingCarryingCapacity"]),
   petrification: new Set(["petrified"]),
   halfHealth: new Set(["exhaustion-4"])
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -1627,7 +1627,7 @@ preLocalize("distanceUnits");
  * @property {Record<string, number>} threshold.encumbered
  * @property {Record<string, number>} threshold.heavilyEncumbered
  * @property {Record<string, number>} threshold.maximum
- * @property {Record<string, number>} speedReduction     Speed reduction caused by encumbered status effects.
+ * @property {Record<string, {ft: number, m: number}>} speedReduction  Speed reduction caused by encumbered status.
  * @property {Record<string, number>} vehicleWeightMultiplier  Multiplier used to determine vehicle carrying capacity.
  */
 
@@ -1669,8 +1669,18 @@ DND5E.encumbrance = {
     }
   },
   speedReduction: {
-    encumbered: 10,
-    heavilyEncumbered: 20
+    encumbered: {
+      ft: 10,
+      m: 3
+    },
+    heavilyEncumbered: {
+      ft: 20,
+      m: 6
+    },
+    exceedingCarryingCapacity: {
+      ft: 5,
+      m: 1.5
+    }
   },
   vehicleWeightMultiplier: {
     imperial: 2000, // 2000 lbs in an imperial ton
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -162,6 +162,8 @@ export default class CharacterData extends CreatureTemplate {
   prepareEmbeddedData() {
     const raceData = this.details.race?.system;
     if ( !raceData ) {
+      this.attributes.movement.units ??= Object.keys(CONFIG.DND5E.movementUnits)[0];
+      this.attributes.senses.units ??= Object.keys(CONFIG.DND5E.movementUnits)[0];
       this.details.type = new CreatureTypeField({ swarm: false }).initialize({ value: "humanoid" }, this);
       return;
     }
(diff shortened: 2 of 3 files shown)


module/config.mjs
```diff
@@ -1669,8 +1669,18 @@ DND5E.encumbrance = {
...
+      m: 3
+    },
+    heavilyEncumbered: {
+      ft: 20,
+      m: 6
+    },
+    exceedingCarryingCapacity: {
+      ft: 5,
+      m: 1.5
+    }
```
dev7355608, line 1683: Perhaps we should have a `CONFIG.DND5E.feetToMeters` conversion factor instead and have only values in feet in your config.

module/config.mjs
```diff
@@ -2461,7 +2471,7 @@ patchConfig("conditionTypes", "label", { since: "DnD5e 3.0", until: "DnD5e 3.2"
  */
 DND5E.conditionEffects = {
   noMovement: new Set(["exhaustion-5", "grappled", "paralyzed", "petrified", "restrained", "stunned", "unconscious"]),
-  halfMovement: new Set(["exhaustion-2", "prone"]),
```
dev7355608, line 2464: Prone does not half the speed. Crawling makes 1 foot movement cost 1 extra foot. So in difficult terrain crawling 1 foot costs 3 feet of movement (see PHB). So a prone creature with a speed 60 feet can crawl 20 feet in difficult terrain. If instead the rule was that the speed was halved while prone, a creature with a speed of 60 could only crawl 15 feet (30 feet after halving, then 2 feet per 1 foot makes 15 feet).

module/data/actor/templates/attributes.mjs
```diff
@@ -107,15 +107,28 @@ export default class AttributesFields {
...
+      if ( noMovement || (crawl && (type !== "walk")) ) speed = 0;
+      else {
+        if ( halfMovement ) speed *= 0.5;
+        if ( heavilyEncumbered ) {
+          speed = Math.max(0, speed - (CONFIG.DND5E.encumbrance.speedReduction.heavilyEncumbered[units] ?? 0));
+        } else if ( encumbered ) {
+          speed = Math.max(0, speed - (CONFIG.DND5E.encumbrance.speedReduction.encumbered[units] ?? 0));
+        }
+        if ( exceedingCarryingCapacity ) {
+          speed = Math.min(speed, CONFIG.DND5E.encumbrance.speedReduction.exceedingCarryingCapacity[units] ?? 0);
```
dev7355608, line 128: `min` there so that the speed cannot be increased by exceeding the carrying capacity if it already less that then 5 feet.

module/data/actor/templates/attributes.mjs
```diff
@@ -107,15 +107,26 @@ export default class AttributesFields {
...
-      if ( (crawl && (k !== "walk")) || noMovement ) this.attributes.movement[k] = 0;
-      else if ( statuses.has("exceedingCarryingCapacity") ) this.attributes.movement[k] = 5;
-      else if ( halfMovement ) this.attributes.movement[k] = Math.floor(this.attributes.movement[k] * 0.5);
-    });
+    const units = this.attributes.movement.units;
+    for ( const type in CONFIG.DND5E.movementTypes ) {
+      let speed = this.attributes.movement[type];
+      if ( noMovement || (crawl && (type !== "walk")) ) speed = 0;
+      else {
+        if ( halfMovement ) speed *= 0.5;
```
dev7355608, line 119: Applying halving before the flat reduction. I don't think there's actually a rule for the order to resolve exhaustion 2 and encumbrance and other flat speed reductions. I think this case usually the effect that is applied first is calculated first. But I don't think that makes sense here, because dropping all your items to remove the encumbrance status and then picking them back up shouldn't change your speed. Anyway, I think the encumbrance reduction should be applied after exhaustion 2, but opinions might differ here.

Another problem I see here handling the movement speed calculation here is that they are always applied after Active Effects, which will surely lead to incorrect speed calculation if for example slow effects from spells are applied to the actor, which should probably be applied after encumbrance and exhaustion 2.

module/data/actor/character.mjs
```diff
@@ -170,13 +172,13 @@ export default class CharacterData extends CreatureTemplate {
...
     if ( raceData.movement.hover ) this.attributes.movement.hover = true;
-    this.attributes.movement.units ??= raceData.movement.units;
+    this.attributes.movement.units ??= raceData.movement.units ?? Object.keys(CONFIG.DND5E.movementUnits)[0];
 
     for ( const key of Object.keys(CONFIG.DND5E.senses) ) {
       if ( raceData.senses[key] ) this.attributes.senses[key] ??= raceData.senses[key];
     }
     this.attributes.senses.special = [this.attributes.senses.special, raceData.senses.special].filterJoin(";");
-    this.attributes.senses.units ??= raceData.senses.units;
+    this.attributes.senses.units ??= raceData.senses.units ?? Object.keys(CONFIG.DND5E.movementUnits)[0];
```
dev7355608, line 181: I've seen `Object.keys(CONFIG.DND5E.movementUnits)[0]` elsewhere in the code.. Perhaps we should have a `CONFIG` entry for the default distance units.

module/data/actor/templates/attributes.mjs
Fyorl: Yeah this is an area where the rules do not give clear guidance as far as I know. I don't think we really have a good way to apply effects in the order they are received, and implementing a way of tracking that seems like it would be far more effort than it's worth. So we should just pick an order (at least for now), and this proposed order seems reasonable enough to me.

Issue #2864, opened 2024-02-03
Add theme picker interface for changing between light, dark, and other themes


arbron:
- Auto (changes between light/dark mode depending on OS settings)
- Light
- Dark
- Colorblindness (might make sense as a separate system setting so it can apply to light and dark mode)

This will also support user-definable themes defined in CSS (any HTML changes would have to be handled by separate sheet modules).

schultzcole:
Just adding to the wishlist here, it might be useful or nice to be able to apply multiple themes at once (light + colorblind, dark + colorblind, etc.), but have a way to denote that certain themes are mutually exclusive (light and dark are mutually exclusive, but not exclusive with colorblind, for instance).

This would allow for useful combinations of themes (potentially provided by modules or core dnd5e).

PR #3025, opened 2024-02-12
[#2864] Add theme option to sheet config & system settings

arbron:
Replaces the standard sheet configuration dialog with one that includes a theme dropdown:

Screenshot 2024-02-12 at 10 03 54

It also adds the same theme dropdown into the system settings.

Doesn't implement any themes itself, but will add `data-theme="<name>"` to the sheet if selected on the sheet, or the body if set in the global settings.

### DIFF
--- a/module/settings.mjs
+++ b/module/settings.mjs
@@ -3,7 +3,7 @@ import { ModuleArtConfig } from "./module-art.mjs";
 /**
  * Register all of the system's settings.
  */
-export default function registerSystemSettings() {
+export function registerSystemSettings() {
   // Internal System Migration Version
   game.settings.register("dnd5e", "systemMigrationVersion", {
     name: "System Migration Version",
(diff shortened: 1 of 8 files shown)


Fyorl (approved):
Looks good to me, just one comment: Can we add a CSS class to body and sheet as well as the `data-theme`? Something like `dnd5e-theme-dark` maybe (this may require some restrictions/post-processing of theme keys)?

arbron:
> Looks good to me, just one comment: Can we add a CSS class to body and sheet as well as the `data-theme`? Something like `dnd5e-theme-dark` maybe (this may require some restrictions/post-processing of theme keys)?

I can, though it does make the logic for changing the theme more complicated since I would have to remove the old theme class before setting the new one, rather than just modifying the data value.

Is that just cause you want to use `.dnd5e-theme-dark` in the CSS rather than `[data-theme="dark"]`?

module/settings.mjs
```diff
@@ -315,3 +315,41 @@ class PrimaryPartyData extends foundry.abstract.DataModel {
...
+ * Set the theme on an element, removing the previous theme class in the process.
+ * @param {HTMLElement} element
+ * @param {string} theme
+ */
+export function setTheme(element, theme) {
+  element.dataset.theme = theme ?? "";
+  const oldTheme = Array.from(element.classList.values()).find(v => v.startsWith("dnd5e-theme-"));
+  if ( theme && oldTheme ) element.classList.replace(oldTheme, `dnd5e-theme-${theme.slugify()}`);
+  else if ( theme ) element.classList.add(`dnd5e-theme-${theme.slugify()}`);
+  else element.classList.remove(oldTheme);
```
Fyorl, line 354: Perhaps slightly terser:

```suggestion
  element.className = element.className.replace(/\bdnd5e-theme-\w+/g, "");
  element.classList.add(`dnd5e-theme-${theme.slugify()}`);
```

Fyorl:
> Is that just cause you want to use `.dnd5e-theme-dark` in the CSS rather than `[data-theme="dark"]`?

Ah, yeah, just a bit more idiomatic to check for classes rather than `data-` attributes in CSS. It's not a big though, but I see you added it anyway, so thank you.

Fyorl (approved):
Thanks for making those adjustments, I had one additional comment which is a suggestion only.

Issue #1019, opened 2021-03-03
Add a long rest/ short rest button to Group actor


aaclayton:
Originally in GitLab by @thingscouldgetdicey1

i was wondering if there could be a global button to short rest and long rest pcs/ npcs. built into the actors tab for 5e. that only gms could see that would quickly rest or long rest pcs/ npcs and auto prompt players with the short rest screen. this could just speed up traveling and resting for parties.

PR #3026, opened 2024-02-12
[#1019] Add resting buttons to group actor for GMs

arbron:
Adds resting buttons for GMs on the group sheet.

Group Rests

These will bring up slightly modified versions of the existing rest dialogs:

Screenshot 2024-02-12 at 11 19 41

And then initiate a rest for all members of the group.

### Future Development
- Add interface for only resting individual members
- Combine resting chat messages into one big group message
- Automatically advance game time by rest amount

### DIFF
--- a/module/applications/actor/group-sheet.mjs
+++ b/module/applications/actor/group-sheet.mjs
@@ -53,6 +53,7 @@ export default class GroupActorSheet extends ActorSheetMixin(ActorSheet) {
     context.system = this.actor.system;
     context.items = Array.from(this.actor.items);
     context.config = CONFIG.DND5E;
+    context.isGM = game.user.isGM;
 
     // Membership
     const {sections, stats} = this.#prepareMembers();
@@ -276,6 +277,8 @@ export default class GroupActorSheet extends ActorSheetMixin(ActorSheet) {
       inputs.focus(ev => ev.currentTarget.select());
       inputs.addBack().find('[type="text"][data-dtype="Number"]').change(this._onChangeInputDelta.bind(this));
       html.find(".action-button").click(this._onClickActionButton.bind(this));
+    } else {
+      html.find(".rest-button").click(this._onClickActionButton.bind(this));
     }
   }
 
@@ -298,12 +301,18 @@ export default class GroupActorSheet extends ActorSheetMixin(ActorSheet) {
         const removeMemberId = button.closest("li.group-member").dataset.actorId;
         this.object.system.removeMember(removeMemberId);
         break;
-      case "rollQuantities":
-        this.object.system.rollQuantities();
+      case "longRest":
+        this.object.longRest();
         break;
       case "movementConfig":
         const movementConfig = new ActorMovementConfig(this.object);
         movementConfig.render(true);
+        break;
+      case "rollQuantities":
+        this.object.system.rollQuantities();
+        break;
+      case "shortRest":
+        this.object.shortRest();
         break;
     }
   }
--- a/module/applications/actor/short-rest.mjs
+++ b/module/applications/actor/short-rest.mjs
@@ -36,26 +36,29 @@ export default class ShortRestDialog extends Dialog {
 
   /** @inheritDoc */
   getData() {
-    const data = super.getData();
+    const context = super.getData();
+    context.isGroup = this.actor.type === "group";
 
-    // Determine Hit Dice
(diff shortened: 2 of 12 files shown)


module/applications/actor/short-rest.mjs
```diff
@@ -36,26 +36,29 @@ export default class ShortRestDialog extends Dialog {
 
   /** @inheritDoc */
   getData() {
-    const data = super.getData();
+    const context = super.getData();
+    context.isGroup = this.actor.type === "group";
```
krbz999, line 40: hi sorry to butt in.
im working on a module subtype that is an extension of the group actor. Could these properties be moved to a set in the application's class such that a module could determine which actors are valid rather than hardcoding it to only `"group"`?

arbron: I suppose, though the only place this is used is to determine a variation of the resting message, nothing functionality-wise is dependent on it.

module/applications/actor/group-sheet.mjs
```diff
@@ -276,6 +277,8 @@ export default class GroupActorSheet extends ActorSheetMixin(ActorSheet) {
       inputs.focus(ev => ev.currentTarget.select());
       inputs.addBack().find('[type="text"][data-dtype="Number"]').change(this._onChangeInputDelta.bind(this));
       html.find(".action-button").click(this._onClickActionButton.bind(this));
+    } else {
+      html.find(".rest-button").click(this._onClickActionButton.bind(this));
```
Fyorl, line 281: These buttons only exists for GMs, right? Seems like this would do nothing then.

module/config.mjs
```diff
@@ -1787,6 +1787,36 @@ DND5E.hitDieTypes = ["d4", "d6", "d8", "d10", "d12"];
...
+/**
+ * Configuration data for rest types.
+ *
+ * @typedef {object} RestConfiguration
+ * @property {Record<string, number>} duration  Duration of different rest variants in minutes.
+ */
+
+/**
+ * Types of rests.
+ * @type {RestConfiguration}
```
Fyorl, line 1799: Should this be `@enum` or `Record<string, RestConfiguration>`? It doesn't look quite right to me.

module/documents/actor/actor.mjs
```diff
@@ -2137,32 +2157,56 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+        "Actor5e._rest now takes a config object and a results object as parameters.",
+        { since: "DnD5e 3.1", until: "DnD5e 3.3" }
+      );
+      const [longRest, dhd, dhp] = args;
+      config = { chat: config, newDay: result };
+      config.type = longRest ? "long" : "short";
+      result = { dhd, dhp };
+    }
+
+    if ( foundry.utils.getType(this.system.rest) === "function" ) return this.system.rest(config, result);
```
Fyorl, line 2201: Interesting idea, but wouldn't this mean that each data model that implements this has to be separately responsible for advancing the game time and firing the hooks?

Fyorl (changes requested):
Good stuff, thanks Jeff, just had a few comments.

module/applications/actor/group-sheet.mjs
arbron: Technically it would allow using rests for group actors in a compendium, but probably not a useful use case.

module/documents/actor/actor.mjs
arbron: Only if they want some logic vastly different than the default actor resting logic, I figure most wouldn't bother and just call the normal rest code. Its not really any different than adding an conditional statement to `Actor#_rest` with special logic for groups, but keeps from filling that method with a bunch of type-specific logic.

Fyorl: Well the group actor is kind of a special case where it calls `longRest`/`shortRest` on each group member. But for a more 'normal' Actor, how would that work? If they add a `rest` function to perform some special logic during a rest, they can't call the normal rest code, can they? Because the normal rest code calls `system.rest`, creating an infinite loop?

arbron: How about this, I've made it so if you return `false` from `ActorDataModel#rest` it will stop the rest of the resting process, but otherwise it will continue.

Fyorl: That seems like a good solution to me

Issue #2934, opened 2024-02-06
AEs explode on migration to 3.0.1


itteerde:
AEs explode on migration to 3.0.1

https://discord.com/channels/170995199584108546/670336046164213761/1204435692977918092

Fyorl:
Can you post the Actor's UUID (right-click the book icon next to the character's name in the title bar), as well as the `origin` values for each of the Comprehend Languages and Disguise Self effects?

itteerde:
`Actor.XbezijMrVTRuOmFc`
```js
[
    {
        "name": "Comprehend Languages",
        "origin": "Actor.XbezijMrVTRuOmFc.Item.ozjZadHlZOnEWP4P"
    },
    {
        "name": "Lamp",
        "origin": "Actor.XbezijMrVTRuOmFc.Item.YSqIuD4NIO3jRJmt"
    },
    {
        "name": "Comprehend Languages",
        "origin": "Actor.XbezijMrVTRuOmFc.Item.ozjZadHlZOnEWP4P"
    },
    {
        "name": "Detect Magic",
        "origin": "Actor.XbezijMrVTRuOmFc.Item.YbmATedSdUh1PxqQ"
    },
    {
        "name": "Disguise Self",
        "origin": "Actor.XbezijMrVTRuOmFc.Item.p27xtZCilZ12ixm5"
    },
    {
        "name": "Expeditious Retreat",
        "origin": "Actor.XbezijMrVTRuOmFc.Item.y1eNMIMfG5vx4C5C"
    },
    {
        "name": "Jump",
        "origin": "Actor.XbezijMrVTRuOmFc.Item.LPhQ9s5WsYXRrR1j"
    }
]```

arbron:
I wonder if this is an issue with the times on the effects. The migration script that detects duplicates does a diff between the effect on the item and on the actor, so if the times have changed then it might result in them not being identified as duplicates.

MotoMoto1234:
[...]

PR #3061, opened 2024-02-15
[#2934] Fix issue with active effect de-duplication

arbron:
First fix was a bug that was thrown if the `origin` on an effect was `null`. The existing check assumed all origins were strings, which would throw an error on `null` origins and prevent checks against subsequent effects from occuring.

The second fix was due to the way the effect de-duplicate worked with the `persistSourceMigration`. The de-duplication operated by calling `deleteEmbeddedDocuments` on the document being migrated to remove the effects, but when `persistSourceMigration` was set it would perform an update on the actor data without diffing resulting in the effects being added back. To fix this we swap to removing the effects from the source data if that flag is set. This fix is only necessary for compendium migrations because world migrations perform the update before the de-duplication.

### DIFF
--- a/module/migration.mjs
+++ b/module/migration.mjs
@@ -23,7 +23,7 @@ export const migrateWorld = async function() {
         }
         await actor.update(updateData, {enforceTypes: false, diff: valid && !flags.persistSourceMigration});
       }
-      if ( actor.effects && actor.items && foundry.utils.isNewerVersion("3.0.0", actor._stats.systemVersion) ) {
+      if ( actor.effects && actor.items && foundry.utils.isNewerVersion("3.0.3", actor._stats.systemVersion) ) {
         const deleteIds = _duplicatedEffects(actor);
         if ( deleteIds.size ) await actor.deleteEmbeddedDocuments("ActiveEffect", Array.from(deleteIds));
       }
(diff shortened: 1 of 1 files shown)


module/migration.mjs
```diff
@@ -560,7 +563,7 @@ function _duplicatedEffects(parent) {
       if ( !effect.transfer ) continue;
       const match = parent.effects.find(t => {
         const diff = foundry.utils.diffObject(t, effect);
-        return t.origin.endsWith(`Item.${item._id}`) && !("changes" in diff) && !deleteIds.has(t._id);
+        return t.origin?.endsWith(`Item.${item._id}`) && !diff.changes?.length && !deleteIds.has(t._id);
```
Fyorl, line 566: I don't think checking `diff.changes?.length` is what we want.

```js
console.log(foundry.utils.diffObject({ changes: ['foo'] }, { changes: ['foo'] })); // {}
console.log(foundry.utils.diffObject({ changes: ['foo'] }, { changes: [] })); // { changes: [] }
```

Fyorl (changes requested):

Fyorl (approved):
Thanks for getting to the bottom of this, Jeff

PR #3064, opened 2024-02-16
Handle missing size when grabbing config

Larkinabout:
Fallback for when a custom actor size is added to the config, actors are assigned the custom size, then the custom config is not loaded. In these instances, it will fallback to the medium actor size data. This could alternatively be a placeholder, such as 'Unassigned' as the size isn't actually reassigned when the sheet is opened. Reloading the custom config is likely preferable to having every actor with a custom size actually reverted to medium, but this at least avoids a breaking error when opening the sheet.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1223,6 +1223,8 @@
 "DND5E.SizeSmallAbbr": "Sm",
 "DND5E.SizeTiny": "Tiny",
 "DND5E.SizeTinyAbbr": "Tn",
+"DND5E.SizeUnassigned": "Unassigned",
+"DND5E.SizeUnassignedAbbr": "Un",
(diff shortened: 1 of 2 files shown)


Larkinabout:
Changed to an Unassigned label as the previous commit didn't account for removal of core actor sizes, such as Medium.

Fyorl (approved):

Fyorl:
Needs a rebase before merge, if you don't mind.

arbron:
@Larkinabout Missed this PR and fixed the bug separately myself here: https://github.com/foundryvtt/dnd5e/commit/fcfcbdb97127933a39b8a2e9988f710f27e1da9b

Issue #2892, opened 2024-02-04
Allow for the list of targets to be collapsed in attack roll messages and to determine which token is which target


krbz999:
Would be useful if massive lists like these, rare as they are, could be collapsed or each individual entry be a bit smaller. Likewise would be good if there was a way to determine which target is which (eg click to pan or click to select). The names seem to be the actor name, rather than token names.

![image](https://github.com/foundryvtt/dnd5e/assets/50169243/036aec83-116f-45c0-8dfe-a22884d88b91)

GamerFlix:
The ability to select the token would further allow for easy damage application, especially if a way of selecting the token without dropping previous selections is available.

MaxPat931:
Being able to select the speaker of a Saving Throw chat card would also be helpful

![image](https://github.com/foundryvtt/dnd5e/assets/86370342/2f5a87ff-4e00-4cde-8478-cb66f6f01387)

krbz999:
Going a step further, the ability to apply damage (or healing or temp hp) with a right-click on an *attack* roll should be removed. This could be substituted with new context menu options that select all targets that were hit or missed (two options).

Fyorl:
The initial implementation was just for attack rolls in order to allow this design work to be postponed, but it does need to be solved eventually for things like the targets of AoE spells. Some combination of a scrollable container + collapsing is the most straightforward but we could look at some other solutions such as utilising popout chat cards, or even tooltips maybe.

arbron:
Partially solved by Zhell's PR, remainder is being solved with https://github.com/foundryvtt/dnd5e/pull/3405

PR #3066, opened 2024-02-16
[#2892] Quality of life changes to attack/damage roll messages

krbz999:
Related to #2892.

- Prevent damage context menu options in roll types that are not damage. (If a message has no roll type, nothing is changed.)
- Add context menu options for selecting hit or missed targets in attack roll messages.
- Sort the targets of an attack roll by hit first, then misses.
- Add hover-in, hover-out, and click event handlers for the targeted tokens in attack roll messages. (With visible inspiration from `CombatTracker`.)


Fyorl (approved):
Good additions

PR #3070, opened 2024-02-16
[#2094] Move actor base data preparation into data models

arbron:
Moves the base data preparation methods in `Actor5e` into their respective data models.


Fyorl (approved):
Nice work, looks good to me

Issue #3058, opened 2024-02-15
Altered core font size causes many issues on new sheets


etiquettestartshere:
**Font size 7:** 
![image](https://github.com/foundryvtt/dnd5e/assets/148253744/99d67f5a-af23-433f-b410-efe50b8f2813)

**Font size 6:**
![image](https://github.com/foundryvtt/dnd5e/assets/148253744/3142b2dc-47da-405b-a959-d6e7d90bfbd9)

**Font size 4:**
![image](https://github.com/foundryvtt/dnd5e/assets/148253744/dd624078-2c11-48c8-aab5-23a1ad09e93c)

It would be nice if these font sizes, which do not seem extreme (two and one above default, and one below default respectively) could be accommodated for accessibility and preference. I personally use font size 6, which is the least disruptive, but still somewhat breaking. If they could result in the abilities organized like font size 5, that would be optimal. Minor nitpick would be to adjust things so that all the default skills, like Animal Handling, only take one line on fonts 6 and 7.

**Default, font size 5**
![image](https://github.com/foundryvtt/dnd5e/assets/148253744/e10cd5de-d67b-4ac7-809a-460d7a50443e)

PR #3082, opened 2024-02-19
[#3058] Fix bad spacing with alternate font sizes

arbron:
Swapped a bunch of relative units used for spacing and positioning with absolute ones to fix the worst of the layout issues at alternate font sizes. Font size can go down as low as 1 and everything still lays out fine, but it still has some issues going much higher than 7.

### Font Size 3

Character Sheet - Font Size 3 (before)
Character Sheet - Font Size 3 (after)

### Font Size 7

Character Sheet - Font Size 7 (before)
Character Sheet - Font Size 7 (after)

### DIFF
--- a/less/v2/character.less
+++ b/less/v2/character.less
@@ -278,7 +278,7 @@
       display: flex;
       justify-content: end;
       align-items: center;
-      padding-right: .75rem;
+      padding-right: 12px;
       background: var(--dnd5e-color-black);
       color: var(--dnd5e-color-gold);
       --icon-fill: var(--dnd5e-color-gold);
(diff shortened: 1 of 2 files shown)


less/v2/character.less
```diff
@@ -694,6 +694,10 @@
               }
             }
           }
+
+          .death-tab > i {
+            font-size: 12px;
+          }
```
Fyorl, line 700: Should this be `.card-tab > i`? We have some general `.card-tab` styles around line  367.

Fyorl (approved):
Thanks for taking care of this, just one comment

Issue #3029, opened 2024-02-12
Extend the actor data model with a new object `attributes.concentration` for concentration related properties


krbz999:
Related to #3028.

Akin to initiative, I strongly believe a new property in the system model would be the way forward rather than using `flags.dnd5e`.

Needed properties include:
- a field for bonuses to concentration.
- a boolean property for whether the actor has advantage.
- like initiative, include a string field to decide what ability the actor uses for concentration.
- a boolean property for whether the actor can concentrate on an item at all (some effects leave an actor unable to concentrate, for example Rage). This could also be a property in `DND5E.conditionEffects` instead.
- A property (preferrably a string) that determines the minimum the actor rolls on the d20 (see: Circle of Stars druid, Dragon Constellation).

PR #3086, opened 2024-02-20
[#3029] Add 'concentration' object in actor system data

krbz999:
This introduces a `RollConfig` field which shares some common data that all similar fields can have (eg initiative, skills, abilities).

The `concentration` property is for `creatures` only.

(Also fixes the documentation for the `init` property.)

```js
  /*
   * @property {object} concentration
   * @property {string} concentration.ability   The ability used for concentration saving throws.
   * @property {string} concentration.bonus     The bonus provided to concentration saving throws.
   * @property {number} concentration.limit     The amount of items this actor can concentrate on.
   * @property {number} concentration.mode      The default advantage mode for this actor's concentration saving throws.
   * @property {object} concentration.roll
   * @property {number} concentration.roll.min  The minimum the d20 can roll.
   * @property {number} concentration.roll.max  The maximum the d20 can roll.
   */
```

### DIFF
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -2,6 +2,7 @@ import { FormulaField } from "../../fields.mjs";
 import MovementField from "../../shared/movement-field.mjs";
 import SensesField from "../../shared/senses-field.mjs";
 import ActiveEffect5e from "../../../documents/active-effect.mjs";
+import RollConfigField from "../../shared/roll-config-field.mjs";
 
 /**
  * Shared contents of the attributes schema between various actor types.
@@ -12,8 +13,8 @@ export default class AttributesFields {
    *
    * @type {object}
    * @property {object} init
-   * @property {number} init.value       Calculated initiative modifier.
-   * @property {number} init.bonus       Fixed bonus provided to initiative rolls.
+   * @property {string} init.ability     The ability used for initiative rolls.
+   * @property {string} init.bonus       The bonus provided to initiative rolls.
    * @property {object} movement
    * @property {number} movement.burrow  Actor burrowing speed.
    * @property {number} movement.climb   Actor climbing speed.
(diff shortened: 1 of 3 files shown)


module/data/shared/roll-config-field.mjs
```diff
@@ -0,0 +1,24 @@
...
+const { StringField, NumberField, SchemaField } = foundry.data.fields;
+
+/**
+ * Field for storing data for a specific type of roll.
+ */
+export default class RollConfigField extends foundry.data.fields.SchemaField {
+  constructor({roll={}, ...fields}={}, options={}) {
+    const opts = { initial: null, nullable: true, min: 1, max: 20, integer: true };
+    fields = {
+      ability: new StringField({required: true, label: "DND5E.AbilityModifier"}),
```
Fyorl, line 12: I think we need a way to pass in a default value for the ability. Maybe something like this:

```suggestion
  constructor({ability="", roll={}, ...fields}={}, options={}) {
    const opts = { initial: null, nullable: true, min: 1, max: 20, integer: true };
    fields = {
      ability: new StringField({required: true, initial: ability, label: "DND5E.AbilityModifier"}),
```

Fyorl (changes requested):
Looks pretty good to me, just one comment

module/data/actor/templates/attributes.mjs
```diff
@@ -63,7 +73,10 @@ export default class AttributesFields {
       }),
       exhaustion: new foundry.data.fields.NumberField({
         required: true, nullable: false, integer: true, min: 0, initial: 0, label: "DND5E.Exhaustion"
-      })
+      }),
+      concentration: new RollConfigField({
+        limit: new foundry.data.fields.NumberField({integer: true, min: 0, initial: 1, label: "DND5E.AttrConcentration.Limit"})
```
Fyorl, line 78:
```suggestion
        ability: "con",
        limit: new foundry.data.fields.NumberField({integer: true, min: 0, initial: 1, label: "DND5E.AttrConcentration.Limit"})
```

krbz999: Woopsie doodle!

Fyorl (approved):
All good, thanks very much Zhell

PR #3095, opened 2024-02-21
[#2849] Display damage modification on sheet, add configuration app

### LINKED ISSUE #2849: Display damage modification on sheet & design interface for manually configuring it
Damage modification was added to `3.0`, but not exposed visibly anywhere on the sheet. A new application for configuring its values similar to the damage resistance app but with inputs for formulas should be implemented as well as a nicely formatted output on the sheet.

arbron:
Displays damage modifications on the sheet beneath vulnerabilities:

Damage Modification Display

And provides a new configuration application for setting them:

Damage Modification Config

Closes #2849

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -563,6 +563,11 @@
 "DND5E.DamageFire": "Fire",
 "DND5E.DamageForce": "Force",
 "DND5E.DamageLightning": "Lightning",
+"DND5E.DamageModification": {
+  "Label": "Damage Modification",
+  "Hint": "Formulas for amount that will be added to typed damage applied to this actor. Negative values will reduce the damage taken.",
+  "BypassHint": "These weapon properties will bypass damage modification for physical damage."
+},
 "DND5E.DamageNecrotic": "Necrotic",
 "DND5E.DamagePiercing": "Piercing",
 "DND5E.DamagePhysical": "Non-Magical Physical",
--- a/lang/en.json
+++ b/lang/en.json
@@ -565,7 +565,7 @@
 "DND5E.DamageLightning": "Lightning",
 "DND5E.DamageModification": {
   "Label": "Damage Modification",
-  "Hint": "Formulas for amount that will be added to typed damage applied to this actor. Negative values will reduce the damage taken.",
+  "Hint": "Formulas for amounts that will be added to typed damage applied to this actor. Negative values will reduce the damage taken.",
   "BypassHint": "These weapon properties will bypass damage modification for physical damage."
 },
 "DND5E.DamageNecrotic": "Necrotic",
--- a/less/v1/apps.less
+++ b/less/v1/apps.less
@@ -567,7 +567,7 @@ dnd5e-effects {
 /* ----------------------------------------- */
 
 .trait-selector {
-  max-height: 700px;
+  max-height: 80vh;
 
   form {
     overflow: hidden;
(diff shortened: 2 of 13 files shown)


krbz999:
Does it make sense to extend this to healing as well?

lang/en.json
```diff
@@ -563,6 +563,11 @@
 "DND5E.DamageFire": "Fire",
 "DND5E.DamageForce": "Force",
 "DND5E.DamageLightning": "Lightning",
+"DND5E.DamageModification": {
+  "Label": "Damage Modification",
+  "Hint": "Formulas for amount that will be added to typed damage applied to this actor. Negative values will reduce the damage taken.",
```
Fyorl, line 568: Either 'amounts', or 'the amount', I'm not sure.

less/v1/apps.less
```diff
@@ -602,6 +602,18 @@ dnd5e-effects {
   }
 }
 
+.damage-modification {
+  .modifications label {
```
Fyorl, line 606: This may be an issue in all the trait selectors, but from the screenshot, it looked like this could do with some spacing between sections:

```suggestion
  .modifications {
    padding-bottom: 6px;
    margin-bottom: 6px;
  }
  .modifications label {
```

module/applications/actor/character-sheet-2.mjs
```diff
@@ -439,12 +439,32 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
...
+    // Prepare damage modifications
+    const dm = this.actor.system.traits?.dm;
+    if ( dm ) {
+      const rollData = this.actor.getRollData({ deterministic: true });
+      const values = Object.entries(dm.amount).map(([k, v]) => {
+        const total = simplifyBonus(v, rollData);
+        if ( !total ) return null;
+        const value = {
+          label: `${CONFIG.DND5E.damageTypes[k]?.label ?? key} ${formatNumber(total, { signDisplay: "always" })}`,
+          color: total < 0 ? "maroon" : "green"
```
Fyorl, line 459: Should this be the other way around? Reducing damage taken is a good thing, right?

module/applications/actor/damage-modification-config.mjs
```diff
@@ -0,0 +1,70 @@
...
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _getSubmitData(updateData) {
+    const data = foundry.utils.expandObject(super._getSubmitData(updateData));
+    const formData = {};
+    for ( const [type, formula] of Object.entries(foundry.utils.getProperty(data, "system.traits.dm.amount")) ) {
+      if ( formula ) formData[`system.traits.dm.amount.${type}`] = formula;
+      else formData[`system.traits.dm.amount.-=${type}`] = "";
```
Fyorl, line 63: I think this is technically supposed to be equal to `null` when indicating a property should be deleted. It probably still works because it's falsy.

```suggestion
      else formData[`system.traits.dm.amount.-=${type}`] = null;
```

Fyorl (approved):
LGTM, only some minor comments

module/applications/actor/damage-modification-config.mjs
arbron: I specifically tried with `null` and it didn't remove the property, only using `""` worked. Could it be something funky with `MappingField`?

Issue #3033, opened 2024-02-12
Provide an actor method `rollConcentration`, with associated `dnd5e.preRollConcentration` and `dnd5e.rollConcentration` hooks


krbz999:
Related to #3028.

kaelad02:
Please make sure the `dnd5e.preRollConcentration` hook is called before the roll dialog is shown. This wasn't done with initiative and... well made things difficult for modules to really change things.

krbz999:
> Please make sure the `dnd5e.preRollConcentration` hook is called before the roll dialog is shown. This wasn't done with initiative and... well made things difficult for modules to really change things.

I swear it on my life.

PR #3097, opened 2024-02-21
[#3033] Add concentration actor method and hooks

krbz999:
Adding the actor method to roll a concentration saving throw, allowing for a DC and other properties to be provided.

Adds two standard hooks.

Making use of `concentration.roll.min` and `.max` for modifying the D20Roll is left as follow-up work since this is [blocked by the roll refactor](https://github.com/foundryvtt/dnd5e/pull/1760#issuecomment-1866519625).

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1632,17 +1632,16 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const parts = [];
 
     // Concentration bonus
-    if ( conc.bonus ) {
-      parts.push(conc.bonus);
-    }
+    if ( conc.bonus ) parts.push(conc.bonus);
 
     const ability = (conc.ability in config.abilities) ? conc.ability : config.defaultAbilities.concentration;
 
     options = foundry.utils.mergeObject({
+      ability: ability,
       isConcentration: true,
       targetValue: 10,
-      advantage: options.advantage || (conc.roll.mode === modes.ADVANTAGE),
-      disadvantage: options.disadvantage || (conc.roll.mode === modes.DISADVANTAGE),
+      advantage: options.advantage || (conc.mode === modes.ADVANTAGE),
+      disadvantage: options.disadvantage || (conc.mode === modes.DISADVANTAGE),
     }, options);
     options.parts = parts.concat(options.parts ?? []);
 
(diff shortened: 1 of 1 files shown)


arbron (approved):

module/documents/actor/actor.mjs
```diff
@@ -1709,6 +1709,64 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    if ( !conc ) throw new Error("You may not make a Concentration Saving Throw with this Actor.");
+
+    const config = CONFIG.DND5E;
+    const modes = CONFIG.Dice.D20Roll.ADV_MODE;
+    const parts = [];
+
+    // Concentration bonus
+    if ( conc.bonus ) {
+      parts.push(conc.bonus);
+    }
```
Fyorl, line 1729:
```suggestion
    if ( conc.bonus ) parts.push(conc.bonus);
```

module/documents/actor/actor.mjs
```diff
@@ -1709,6 +1709,64 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+
+    const ability = (conc.ability in config.abilities) ? conc.ability : config.defaultAbilities.concentration;
```
Fyorl, line 1731: Do we want to allow for a way to override this via some options passed to this method?

module/documents/actor/actor.mjs
```diff
@@ -1709,6 +1709,64 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+
+    options = foundry.utils.mergeObject({
+      isConcentration: true,
+      targetValue: 10,
+      advantage: options.advantage || (conc.roll.mode === modes.ADVANTAGE),
+      disadvantage: options.disadvantage || (conc.roll.mode === modes.DISADVANTAGE),
```
Fyorl, line 1737: I might have missed something but I think `mode` is a top level property of `RollConfigField`:

```suggestion
      advantage: options.advantage || (conc.mode === modes.ADVANTAGE),
      disadvantage: options.disadvantage || (conc.mode === modes.DISADVANTAGE),
```

Maybe it should be under `roll`?

module/documents/actor/actor.mjs
```diff
@@ -1709,6 +1709,64 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    /**
+     * A hook event that fires before a saving throw to maintain concentration is rolled for an Actor.
+     * @function dnd5e.preRollConcentration
+     * @memberof hookEvents
+     * @param {Actor5e} actor                   Actor for which the saving throw is being rolled.
+     * @param {D20RollConfiguration} options    Configuration data for the pending roll.
+     * @param {string} abilityId                ID of the ability being rolled as defined in `DND5E.abilities`.
+     * @returns {boolean}                       Explicitly return `false` to prevent the save from being performed.
+     */
+    if ( Hooks.call("dnd5e.preRollConcentration", this, options, ability) === false ) return;
```
Fyorl, line 1750: We could consider allowing this hook to modify which ability is used for the concentration save also, but I do not feel strongly.

Fyorl (changes requested):
This is great, thank you for taking care of this work. I have left some comments that need clarification. With regards to overriding the ability used, I do not feel particularly strongly. It's ok if we defer purely to whatever is in the data model at the time of rolling.

module/documents/actor/actor.mjs
krbz999: We could maybe discuss that on my PR to add the `RollConfig` field to other areas of the actor model, but I'm not opposed. The original idea was `roll` being *modifiers* for the roll, but I suppose that advantage mode very much is that as well.

module/documents/actor/actor.mjs
krbz999: That's OK I have the strength for both of us.

Fyorl (approved):

Issue #3099, opened 2024-02-22
More Details in ChatMessage does not work, collapses details instead


itteerde:
The `More Details` in the `ChatMessage` rendering does not seem to work anymore (3.0.3) (which is somewhat bad as one cannot read the text in the CharacterSheet either (no expand there anymore [bad btw] and the tooltip has the `More Details` label for no functional reason), so only when editing the item, and no way to show it to others). It does instead collapse the Details.
![image](https://github.com/foundryvtt/dnd5e/assets/8367477/d3b46180-c69f-4007-99e9-ce3c1f407718)

PR #3105, opened 2024-02-22
[#3099] Fix chat messages collapsing when clicking on details element

arbron:
Would love another set of eyes on this one. It fixes all the issues with using `<details>` elements in chat card descriptions, except when you collapse the details section the chat card doesn't reclaim the height properly:

Screenshot 2024-02-22 at 11 18 07

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -2007,7 +2007,14 @@ export default class Item5e extends SystemDocumentMixin(Item) {
    */
   static _onChatCardToggleContent(event) {
     // If the user is clicking on a link in the collapsible region, don't collapse
-    if ( event.target.closest(":is(.item-name, .collapsible) :is(a, button)") ) return;
+    const closest = event.target.closest(":is(.item-name, .collapsible) :is(a, button, summary)");
+    if ( closest ) {
+      if ( closest.tagName === "SUMMARY" ) requestAnimationFrame(() => {
+        const details = event.currentTarget.querySelector(".collapsible-content");
+        details.style.height = `${details.scrollHeight}px`;
+      });
+      return;
+    }
 
     event.preventDefault();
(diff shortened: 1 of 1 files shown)


Fyorl:
I think to reclaim the space we need to clear the fixed height so the container can recalculate its `scrollHeight` appropriately, then set the container to the `scrollHeight` size if we want to show a CSS transition.

Looking at this code though, there are some things I would like to clean up:
1. We shouldn't consume click events on the details at all really. It should really only be collapsible from the summary (that's what has the cursor style implying interactivity). Then we don't need to check for links and things at the top, we can just let those have their normal click behaviour.
2. In order to handle things that might resize the details container, we should clear the fixed `height` once the CSS transition finishes. We only use fixed heights because CSS transitions (rather annoyingly) do not work on auto-sized containers. So we should use `transitionend` listeners with `once: true` to clear the fixed height when expanding is completed, and then re-add the fixed height just before collapsing so that the transition to `0` animates correctly. Such code is a bit messy, and there are other places we have to do this dance (like the accordion) so it might be nice to build some generic helpers that let us do this easily.

I'll raise an issue for the above, and if you don't mind, we could mark this issue as blocked and bump it to 3.1? I don't think this issue is so urgent that it needs to be fixed immediately, just a little annoying.

Fyorl:
> There is a somewhat inconvenient/non-intuitive, but solid work around using grid that would allow us to avoid having to manually set the height: [keithjgrant.com/posts/2023/04/transitioning-to-height-auto/#with-grid](https://keithjgrant.com/posts/2023/04/transitioning-to-height-auto/#with-grid)
> 
> I've used it at work successfully and it should be supported on all of the browser versions we already require.

That's a great tip; if we can do it purely with CSS that would really be ideal.

PR #3109, opened 2024-02-22
[#1787] Add custom journal page for spell lists

### LINKED ISSUE #1787: Add custom journal entry page for spell lists

arbron:
Introduces a new journal entry page type for representing lists of spells. These spells are listed as links and automatically grouped using a variety of grouping modes:

Spell List Modes

In the configuration page, a default grouping mode can be set. The mode can also be set by viewers of the page (though this isn't persisted).

Spell List Config Empty

Spell lists are configured by dragging a spell or a folder of spells into the config window.

Spell List Config Full

Spell lists also have a Type and an Identifier, which allows a spell list to be associated with a specific class for example for future features that can combine multiple spell lists together for gaining spells.

Closes #1787 

### Followup PRs
- [ ] Add way to represent potentially unavailable spells (spells that aren't in the current module or the SRD)
- [ ] Add a method for embedding spell lists
- [ ] Replace SRD spell listing with a spell list


module/applications/journal/spell-list-page-sheet.mjs
```diff
@@ -0,0 +1,177 @@
...
+    let fields;
+    switch ( grouping ) {
+      case "level": fields = ["system.level"]; break;
+      case "school": fields = ["system.school"]; break;
+      default: fields = []; break;
+    }
+
+    let collections = new Collection();
+    for ( const uuid of this.document.system.spells ) {
+      const { collection } = foundry.utils.parseUuid(uuid);
```
Fyorl, line 108: A non-compendium Item has a `collection`, but you can't call `collection.getIndex` on it, so it throws an error and prevents the sheet from rendering. It's probably not a good idea for a user to use spells that are not in a compendium, but we could fix the error and allow it. Alternatively we could check in `_onDrop` and emit a warning, preventing the Item from being added if it's not in a compendium.

Fyorl (changes requested):
Please see my comments but this looks good, thanks Jeff

module/data/journal/_module.mjs
```diff
@@ -1,18 +1,21 @@
 import ClassJournalPageData from "./class.mjs";
 import MapLocationJournalPageData from "./map.mjs";
 import RuleJournalPageData from "./rule.mjs";
+import SpellListJournalPageData from "./spell-list.mjs";
```
Fyorl, line 4: Should the filename also be `spells.mjs` for consistency?

module/applications/journal/spell-list-page-sheet.mjs
```diff
@@ -0,0 +1,179 @@
...
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _onDrop(event) {
+    const data = TextEditor.getDragEventData(event);
+    let spells;
+    switch ( data?.type ) {
+      case "Folder":
+        spells = (await Folder.implementation.fromDropData(data))?.contents;
```
Fyorl, line 163: I noticed this only retrieves the immediate children of the folder rather than any spells that are contained at some level under the dropped folder. I don't feel strongly on whether that should change but wanted to mention it in case you intended for the latter behaviour.

Fyorl (approved):
A couple more comments for your consideration, but otherwise this is good to merge

module/applications/journal/spell-list-page-sheet.mjs
arbron: This does match the current behavior of dropping a folder anywhere else which only handles one level, so I'd prefer to keep everywhere matching for the moment unless we decide to adjust other locations where we also accept dropped folders.

PR #3114, opened 2024-02-23
[#2841, #3110, #3111, #3113] Update build script to fix issues & remove deprecations

### LINKED ISSUE #2841: The release includes the pack's _source directory
I'm not sure if it's intentional or not, but I noticed that the release includes the `packs/_source` directory that has the separate JSON files. It's not a show-stopper, but I don't see any reason for the release to include those since they're already packaged up.

### LINKED ISSUE #3110: Remove deprecated `set-output` command from build script
The current build script uses the `set-output` command for passing variables, which has build deprecated. The script should switch to setting those variables as environmental variables instead.

Current:
```
echo ::set-output name=TAG_NAME::$TAG

${{ steps.get_vars.outputs.TAG_NAME }}
```

New:
```
echo "TAG_NAME=$TAG" >> $GITHUB_ENV

${{ env.TAG_NAME }}
```

### LINKED ISSUE #3111: Allow build script to output draft releases, don't delete existing release body
The current build script forces `draft` and `prerelease` to `false` for all newly built releases, forcing any new release to be the "Latest" release. It should be simple to modify the build script to pass through the setting from the Create Release page.

It should also avoid replacing the existing release body when it adds the manifest link, and instead append it after any existing text.

### LINKED ISSUE #3113: Update build script to remove deprecated Node.js 16 actions
`Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, ActionsTools/read-json-action@main, microsoft/variable-substitution@v1, actions/setup-node@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.`

`The following actions uses node12 which is deprecated and will be forced to run on node16: microsoft/variable-substitution@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/`

arbron:
Closes #2841 
Closes #3110 
Closes #3111 
Closes #3113

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,6 +99,16 @@ jobs:
     - run: zip ${{ env.ZIP_NAME }} -r fonts icons lang json packs templates tokens ui dnd5e.css dnd5e.mjs dnd5e-compiled.mjs.map LICENSE.txt README.md system.json template.json --exclude "packs/_source/*"
 
 
+    # Fetch the body from the release
+    - name: Fetch Release Body
+      id: release
+      uses: cardinalby/git-get-release-action@v1
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      with:
+        tag: ${{ env.TAG_NAME }}
+
+
     # Create a release for this specific version
     - name: Update Release with Files
       id: create_version_release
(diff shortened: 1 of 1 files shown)


Fyorl (approved):
Looks good to me, thanks for tidying this up. I had one question, it relates to this:

> The current build script forces `draft` and `prerelease` to `false` for all newly built releases, forcing any new release to be the "Latest" release. It should be simple to modify the build script to pass through the setting from the Create Release page.
> 
> It should also avoid replacing the existing release body when it adds the manifest link, and instead append it after any existing text.

Since we trigger the release on push of a tag, I don't think there are any settings from the 'Create Release' page to 'pass through' here, nor would there be any existing release body. Am I misunderstanding something about our release pipeline perhaps?

arbron:
> Since we trigger the release on push of a tag, I don't think there are any settings from the 'Create Release' page to 'pass through' here, nor would there be any existing release body. Am I misunderstanding something about our release pipeline perhaps?

Oh, well I've always gone through the Create Release page and let that create the tag for me so it has whatever data I enter there to work with. I'll have to give this a test to see what happens if you create a tag without first creating the release.

arbron:
@Fyorl Okay, I've updated the script slightly to work without going through the Create Release page. Now it will default all new releases created via tag to `prerelease`, but preserve the settings and body if going through Create Release.

PR #3115, opened 2024-02-23
Improve compatibility of localized token resources.

DJ4ddi:
The goal of this PR is to simplify the usage of the new item resources without fully replicating the system's behavior. Working around the current implementation of `TokenConfig._prepareResourceLabels` is very difficult because it runs after the entire rendering pipeline. This is resolved by moving the logic to the earliest possible place. More specifically:
- Moved the attribute grouping to `TokenDocument.getTrackedAttributeChoices` to allow modules to use the localized groups.
- Moved the item attributes to `TokenConfig.getData` to allow dealing with them in the regular rendering pipeline.
- Moved the item name resolution to `getHumanReadableAttributeLabel` to allow resolving relative item names globally.
- Changed the selector in `TokenConfig._prepareResourceLabels` to `select.bar-attribute` to cover additional inputs.
- Slightly optimized the sorting process to avoid an unnecessary loop.

### DIFF
--- a/module/applications/token-config.mjs
+++ b/module/applications/token-config.mjs
@@ -92,6 +92,8 @@ export default class TokenConfig5e extends TokenConfig {
     if ( !this._minimized ) this.setPosition();
   }
 
+  /* -------------------------------------------- */
+
   /**
    * Adds charge based items as attributes for the current token.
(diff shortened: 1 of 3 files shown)


Fyorl (approved):
Nice simplification, thank you.

PR #3118, opened 2024-02-23
[#3116] Add ability to embed actor & item descriptions

### LINKED ISSUE #3116: Expand embed enricher to support embedding item/actor descriptions
This will be more flexible when we can support V12 only, but for now we can simply check actors for `system.details.biography.value` and items for `system.description.value` and display those for the embeds.

arbron:
Actor:Item Embeds

Did a bit of re-ordering of enrichers to keep things grouped together for easier code perusal.

Closes #3116

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -3,8 +3,6 @@ import Award from "./applications/award.mjs";
 import { damageRoll } from "./dice/_module.mjs";
 import * as Trait from "./documents/actor/trait.mjs";
 
-const MAX_EMBED_DEPTH = 5;
-
 const slugify = value => value?.slugify().replaceAll("-", "");
 
 /**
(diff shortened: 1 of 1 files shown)


module/enrichers.mjs
```diff
@@ -345,6 +357,75 @@ async function enrichDamage(config, label, options) {
...
+ * @returns {HTMLElement|null}         An HTML figure containing the image, caption from the image page or a custom
+ *                                     caption, and a link to the source if it could be built, otherwise null.
+ */
+async function embedDocument(config, label, options) {
+  options = { ...options, _embedDepth: options._embedDepth + 1, relativeTo: config.doc };
+  config.inline ??= config.values.includes("inline");
+
+  const keyPath = (config.doc instanceof Actor) ? "system.details.biography.value" : "system.description.value";
+  const description = foundry.utils.getProperty(config.doc, keyPath);
+  if ( description === undefined ) return null;
```
Fyorl, line 423: Is it desirable that a blank description be rendered as such? Or should this be:

```suggestion
  if ( !description ) return null;
```

module/enrichers.mjs
```diff
@@ -545,6 +603,45 @@ async function embedRollTable(config, label, options) {
 
 /* -------------------------------------------- */
 
+/**
+ * Wrap embeds in containing elements.
+ * @param {string} enriched  Enriched text content to include.
+ * @param {object} config              Configuration data.
+ * @param {string} label               Optional label to use as the table caption.
```
Fyorl, line 610:
```suggestion
 * @param {string} label               Optional label to replace the default caption.
```

module/enrichers.mjs
```diff
@@ -545,6 +603,45 @@ async function embedRollTable(config, label, options) {
 
 /* -------------------------------------------- */
 
+/**
+ * Wrap embeds in containing elements.
+ * @param {string} enriched  Enriched text content to include.
```
Fyorl, line 608:
```suggestion
 * @param {string} enriched            Enriched text content to include.
```

module/enrichers.mjs
```diff
@@ -345,6 +357,75 @@ async function enrichDamage(config, label, options) {
...
+ * @param {string} [label]             Optional label to replace the default caption.
+ * @param {EnrichmentOptions} options  Options provided to customize text enrichment.
+ * @returns {HTMLElement|null}         An HTML figure containing the image, caption from the image page or a custom
+ *                                     caption, and a link to the source if it could be built, otherwise null.
+ */
+async function embedDocument(config, label, options) {
+  options = { ...options, _embedDepth: options._embedDepth + 1, relativeTo: config.doc };
+  config.inline ??= config.values.includes("inline");
+
+  const keyPath = (config.doc instanceof Actor) ? "system.details.biography.value" : "system.description.value";
```
Fyorl, line 421: Should we handle unidentified items?

```suggestion
  const keyPath = (config.doc instanceof Actor) ? "system.details.biography.value" : game.user.isGM || (config.doc.system.identified === false) ? "system.description.value" : "system.unidentified.description";
```

Fyorl (changes requested):

module/enrichers.mjs
arbron: If we go this way then a blank enricher will render as the original enricher text. I figured it we better to have a blank space for a blank description rather than showing that and leave the enricher text fallback as an option for people trying to embed documents that don't support this format.

Fyorl (approved):

PR #3135, opened 2024-02-26
Add bleeding, cursed, silenced, surprised, and transformed to conditions

dev7355608:
We already have *diseased* as a condition even though it technically isn't one so that it shows up nicely under *Condition Immunities* for paladins and monks. So we might as well add some of the other non-RAW conditions, I think. Probably controversial.
- Bleeding: Perhaps useful considering there's blood magic that is only able to target creatures with blood.
- Cursed: I couldn't find a creature that is immune against curses, but it might be possible, not sure.
- Silenced: If you can be immune against blinded and deafened, I imagine you could possibly be immune against being silenced.
- Surprised: Some creatures are immune to being surprised unless incapacitated (e.g. Canoloth). There's also the Helm of Awareness (Artificer Infusion).
- Transformed: Some creatures are immune to any spell or effect that would alter its form (e.g. Clay Golem).

### DIFF
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -353,7 +353,7 @@ export default class ActorSheet5eCharacter2 extends ActorSheet5eCharacter {
     // Effects & Conditions
     const conditionIds = new Set();
     context.conditions = Object.entries(CONFIG.DND5E.conditionTypes).reduce((arr, [k, c]) => {
-      if ( k === "diseased" ) return arr; // Filter out diseased as it's not a real condition.
+      if ( !c.raw ) return arr; // Filter out non-RAW conditions.
       const { label: name, icon, reference } = c;
       const id = staticID(`dnd5e${k}`);
       conditionIds.add(id);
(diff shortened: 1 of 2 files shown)


Fyorl:
I think we actually want to *remove* diseased from conditions, rather than go the other way. It's got a special-case exception to omit it from the character sheet, for example.

krbz999:
I think removing Diseased would be even more controversial.

We could introduce `system.traits.ailments` or some such, if a split is really needed.

Fyorl:
If it's the simplest way to implement some of these features like Immutable Form, then that's fine. I can see we have a `raw` flag in this PR which would probably have been my only real concern.

Fyorl:
> Might be a bit nicer to invert `raw`, add `special` or something that is only for conditions that are not official conditions rather than tagging every official condition. Also not sure how clear `raw` is without the JSDocs, given it isn't clear it is an acronym.

Yeah, I think I agree that marking the non-official ones is preferable. Maybe we can flag them as `pseudo: true`? It's reasonably terse and things like 'surprised' are sometimes referred to as 'pseudo conditions' because they behave similarly without actually being in the conditions appendix.

Fyorl (approved):

PR #3145, opened 2024-02-27
Make icon use currency abbreviation

cstby:
Currently, the currency icons depend on the object key in `CONFIG.DND5E.currenices`. This is not ideal because it means these icons will be mismatched if a module patches the currencies object.

This pull requests updates the inventory template that that it will use icons based on the currency abbreviations, which can be patched by https://github.com/cstby/foundryvtt-world-currency-5e.

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -1404,27 +1404,32 @@ DND5E.currencies = {
   pp: {
     label: "DND5E.CurrencyPP",
     abbreviation: "DND5E.CurrencyAbbrPP",
-    conversion: 0.1
+    conversion: 0.1,
+    icon: "pp"
   },
   gp: {
     label: "DND5E.CurrencyGP",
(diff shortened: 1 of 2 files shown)


arbron:
Not a feasible change, because those abbreviations might be localized, and we don't want to force localization modules to update a whole bunch of CSS just to change those abbreviations.

cstby:
That's totally understandable.

What would you think about providing the icon specifications in the currencies config object? That would allow modules to remap the icon without running into the localization issue you mentioned above.

cstby:
Updated the pull request to give you a concrete idea of the proposal above.

Larkinabout:
Could the config contain the path to the image itself instead, so that they could be swapped out with other images without further messing with the CSS? I can only think of replacing the classes with a style tag with the url populated in the template.

arbron:
Closed in favor of https://github.com/foundryvtt/dnd5e/pull/3155

Issue #2138, opened 2023-01-30
Refactor roll pipeline


Fyorl:
There are two main goals for this refactor:

1. Separate out the various 'options groups' that are passed around as part of a roll. At the moment these are all bundled into a single, huge `options` parameter. These options fall into three main groups:
    - Options that configure the intermediary roll dialog that prompts users for advantage/normal/disadvantage.
    - Options that are passed to the `D20Roll` constructor to configure the roll itself.
    - Options that control the chat message that is created as part of the roll.

2. Iron out the roll pipeline to accommodate a user changing fundamental aspects of a roll in the intermediary roll dialog (e.g. #1530). The current pipeline operates as follows:
    - Dialog data, roll data, and message data is prepared in the roll method, e.g. `Actor5e#rollSkill`.
    - We instantiate a `D20Roll` instance with the configuration from the previous step.
    - We check whether the roll is fast-forwarded, and roll immediately if it is, otherwise continue.
    - We spawn an intermediary roll dialog with `D20Roll#configureDialog` that prompts a user for a situational bonus, advantage or disadvantage, or certain other choices specific to the roll like the ability score to use for the roll.
    - We then perform surgery on the roll terms themselves in `D20Roll#_onDialogSubmit` in order to splice in any changes that were made in the dialog (like changing the `@mod` term if the ability score was changed). This is missing some roll-specific logic, like with the remarkable athlete bug above.

There are two draft PRs that tackle this refactor in different ways, so we can decide which one we prefer to proceed with (or none of them). (1) is common to both PRs and is more or less optional, but I feel it makes the code easier to follow and improves the API surface. (2) is where each of the PRs diverge.

#2139 is the simpler solution that straightens out the pipeline in a more understandable way, in my view, but has some downsides such as making the formula preview less useful (or removing it entirely).

#2140 is a more involved solution that still solves the problem but makes the pipeline a bit harder to follow, in my view.

JPMeehan:
Want to express support for resolving this issue. Example use case from a module I'm working on that's implementing the Level Up ruleset by EN Publishing: Characters can gain an "Expertise die" as a bonus to skills, ranging from d4 to d8. Some of these benefits are passive, while others are situational. I would like to have a `select` field ranging from blank to d8, with the initial value based on the passive configuration from the actor's flags. Unfortunately, there's no way to have that data translate either INTO the roll dialog (by smuggling it as the starting value of the situational bonus) or OUT of the roll dialog (with a render-inserted select field)

PR #3146, opened 2024-02-27
[#2138] Add BasicRoll type as part of the roll refactor

### LINKED ISSUE #1813: Add optional hit dice roll configuration window
In certain cases it would be useful to be able to show a roll configuration window for hit dice rolls to add bonuses, but with the change of hit dice from `DamageRoll` to plain `Roll` that is no longer possible.

arbron:
First part of refactoring the rolling system. This adds a new `BasicRoll` type that replaces core's `Roll` adding a number of methods to help with building one or more rolls, displaying a roll prompt, and posting a message to chat.

Splits roll configuration into three configuration objects, one for the rolls, one for the prompt, and one for the chat message. This separation should make it clear what options control what part of the process and make it easy to introduce new options in the future or to subclasses without a single massive configuration object like is currently used for the `d20Roll` function.

Adds a custom dialog application for roll prompts rather than just customizing the contents of a standard dialog to allow for more dynamic interface. Every time changes are made in the dialog it applies those changes to the roll configuration object and then re-creates the rolls from those configs. The default dialog class can be specified for each roll type and by individual calls to the `build` factory method, allowing very custom implementations for specific rolls (such as adding an ability dropdown only for skill rolls).

Currently I've only converted the `Actor5e#rollHitDie` method to use this new system as an example, but this `BasicRoll` will also be used for things like rolling hit points and item recharges. The changes to `rollHitDie` include deprecation paths to ensure that existing use of that function or its hooks don't break, though this required adding a new version of the hooks so nothing breaks for people using the existing ones.

Closes #1813


krbz999:
> Every time changes are made in the dialog it applies those changes to the roll configuration object and then re-creates the rolls from those configs.

Would it be possible to add a method on the application to programmatically push a new roll? For example:
```js
dialog.addRoll("2d8", "radiant");
```
which then causes it to rebuild. A way to achieve the same result as using the situational bonus field essentially.

arbron:
> Would it be possible to add a method on the application to programmatically push a new roll? For example:
> 
> ```js
> dialog.addRoll("2d8", "radiant");
> ```
> 
> which then causes it to rebuild. A way to achieve the same result as using the situational bonus field essentially.

Rather than adding a specific `addRoll` method, I've added a public `rebuild()` method that will rebuild the rolls and render the dialog after changes have been made to the config. So that would be:

```javascript
dialog.config.rolls.push({ parts: ["1d6"], options: { type: "radiant" } });
dialog.rebuild();
```

module/applications/dice/roll-configuration-dialog.mjs
```diff
@@ -0,0 +1,295 @@
...
+  /**
+   * Handle submission of the dialog using the form buttons.
+   * @param {Event|SubmitEvent} event    The form submission event.
+   * @param {HTMLFormElement} form       The submitted form.
+   * @param {FormDataExtended} formData  Data from the dialog.
+   * @private
+   */
+  static async #handleFormSubmission(event, form, formData) {
+    const rolls = this._finalizeRolls(event.submitter?.dataset?.action);
+    await this.close({ dnd5e: { rolls } });
```
Fyorl, line 257: We could avoid having to propagate these rolls via `options` if we change the contract for `_finalizeRolls` to expect subclasses to modify `this.rolls` in-place, rather than returning new rolls. That would mean that after `_finalizeRolls` is called, the dialog's `this.rolls` is the final version of the rolls, which is perhaps useful.

```suggestion
    this._finalizeRolls(event.submitter?.dataset?.action);
    this.close();
```

module/applications/dice/roll-configuration-dialog.mjs
```diff
@@ -0,0 +1,295 @@
...
+   * A helper to handle displaying and responding to the dialog.
+   * @param {BasicRollProcessConfiguration} [config]        Initial roll configuration.
+   * @param {BasicRollConfigurationDialogOptions} [dialog]  Dialog configuration options.
+   * @param {BasicRollMessageConfiguration} [message]       Message configuration.
+   * @returns {Promise}
+   */
+  static async configure(config={}, dialog={}, message={}) {
+    return new Promise((resolve, reject) => {
+      new this(config, message, { ...(dialog.options ?? {}), resolve, reject }).render(true);
+    });
```
Fyorl, line 293: If you take my suggestion above to change the contract for `_finalizeRolls` and not propagate the final rolls via the `close` options, then we could write this in a slightly more idiomatic AppV2 style, removing the need for `resolve` and `reject` options in `BasicRollConfigurationDialogOptions`, and the `_onClose` override.

```suggestion
    return new Promise(resolve => {
      const app = new this(config, message, dialog);
      app.addEventListener("close", () => resolve(app.rolls), { once: true });
      app.render({ force: true });
    });
```

Fyorl (changes requested):
This looks really good, thank you Jeff. Please see my comments for consideration.

Fyorl (changes requested):
This looks really good, thank you Jeff. Please see my comments for consideration.

templates/dice/roll-formulas.hbs
```diff
@@ -0,0 +1,14 @@
+<fieldset class="roll-formulas">
+    <legend>{{ localize "DND5E.RollConfiguration.Rolls" }}</legend>
+    {{#each rolls}}
+    <div class="form-group">
+        <label>{{ localize "DND5E.Formula" }} {{#if type}}({{ type }}){{/if}}</label>
+        <input type="text" name="formula" value="{{ formula }}" disabled>
```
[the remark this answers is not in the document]
arbron: Was mainly to get the height and formatting consistent between rows, but if we do a final UI pass that might be less important:

Screenshot 2024-07-24 at 16 22 19
Screenshot 2024-07-24 at 16 22 09

module/applications/dice/roll-configuration-dialog.mjs
```diff
@@ -0,0 +1,295 @@
...
+    this.message = message;
+    this.rolls = this.#buildRolls(foundry.utils.deepClone(this.config));
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static DEFAULT_OPTIONS = {
+    classes: ["roll-configuration"],
+    tag: "form",
```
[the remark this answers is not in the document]
arbron: My one concern would be players who begin a roll and then want to look up something on their sheet to see if they have advantage or some sort of situational bonus.

It might be good to be a `popover` instead of modal, so it will always stay on top but still allow for interacting with other applications. Unfortunately core's window dragging doesn't seem to interact very well with popovers.

```javascript
  /** @inheritDoc */
  async _onFirstRender(context, options) {
    await super._onFirstRender(context, options);
    this.element.setAttribute("popover", "");
  }
```

module/applications/dice/roll-configuration-dialog.mjs
arbron: Yeah, I can update that. Those main reason I was using closing options here is to know that the closing occurred via the submission versus occurring through the close button so I can know what to send to the `resolve` method.

Since the `close` method already takes `ApplicationClosingOptions`, it would be fantastic if core provided some more details indicating what triggered the closing.

module/applications/dice/roll-configuration-dialog.mjs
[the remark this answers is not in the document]
Fyorl: Yeah that makes sense. Probably more trouble than it's worth at this stage to use popover or dialog.

module/applications/dice/roll-configuration-dialog.mjs
Fyorl: That is a good idea. Maybe we can add some properties to the close event.

module/dice/basic-roll.mjs
```diff
@@ -0,0 +1,183 @@
...
+   * @param {BasicRollMessageConfiguration} [message={}]  Configuration for message creation.
+   * @returns {BasicRoll[]}
+   */
+  static async build(config={}, dialog={}, message={}) {
+    this.applyKeybindings(config, dialog, message);
+
+    let rolls;
+    if ( dialog.configure === false ) rolls = config.rolls?.map(config => this.create(config)) ?? [];
+    else {
+      let DialogClass = dialog.applicationClass ?? this.DefaultConfigurationDialog;
```
Fyorl, line 93:
```suggestion
      const DialogClass = dialog.applicationClass ?? this.DefaultConfigurationDialog;
```

Fyorl (approved):

Issue #3147, opened 2024-02-27
Add magical bonus to weapons & armor


arbron:
Currently turning a weapon into a `+1` version involves adding `1` to the "Attack Bonus" field and ` + 1` to the first damage part. This doesn't give any indication that the value in those fields is from the magical part of the weapon and not something else. This comes into play for weapons that require attunement, which shouldn't confer the magical bonus unless they are actually attuned. It is also an issue if we ever implement something like an anti-magic field, which should prevent the magical bonuses from being used.

A new "Magical Bonus" field on the weapon, possible only displayed when the "Magical" property is checked, should allow us to add these bonuses to attack and damage automatically as long as the proper conditions are met.

thatlonelybugbear:
The attunement part for the Magic bonus to attack/damage is not a prerequisite in some cases. There are some weapons that are magical and confer their base bonuses to attack/damage to the wielder while not attuned and other abilities when attuned.
Hammer of Thunderbolts is one such example

krbz999:
> The attunement part for the Magic bonus to attack/damage is not a prerequisite in some cases. There are some weapons that are magical and confer their base bonuses to attack/damage to the wielder while not attuned and other abilities when attuned. Hammer of Thunderbolts is one such example

"Hazirawn" is another special case. The attack bonus is +1, but +2 if attuned, and the damage bonus is +1d6 or +2d6 if attuned.

arbron:
> The attunement part for the Magic bonus to attack/damage is not a prerequisite in some cases. There are some weapons that are magical and confer their base bonuses to attack/damage to the wielder while not attuned and other abilities when attuned. Hammer of Thunderbolts is one such example

The general rule is that for any item that requires attunement must be attuned in order to get any of its magical effects, that includes the +1 to attack and damage. So for the Hammer of Thunderbolts, you wouldn't get the +1 to attack or damage if you aren't attuned.

> "Hazirawn" is another special case. The attack bonus is +1, but +2 if attuned, and the damage bonus is +1d6 or +2d6 if attuned.

That is a special case. In that instance I imagine it would be implemented with a static +1 in the attack bonus & first damage field like we use currently, and a +1 in the new magical bonus field.

For the extra necrotic damage, I think that will have to wait until the activities system which will have more control over what damage parts are added under certain conditions.

PR #3148, opened 2024-02-28
[#3147] Add magical bonus to weapons & armor

arbron:
Introduces `system.magicalBonus` to weapons and `system.armor.magicalBonus` to equipment to represent bonuses to attack, damage, and AC from the magical part of these items.

Also introduces `system.magicAvailable` to physical items that indicates whether the magical effects of these items is available. Currently this is only based on current attunement, but could be expanded later to other things that might make magic unavailable.

### Followup Work
- [ ] Modify SRD weapons & armor to use these new properties
- [ ] Add magical bonus to ammunition

### DIFF
--- a/module/data/item/equipment.mjs
+++ b/module/data/item/equipment.mjs
@@ -143,6 +143,7 @@ export default class EquipmentData extends ItemDataModel.mixin(
   /** @inheritDoc */
   prepareDerivedData() {
     super.prepareDerivedData();
+    this.armor.value = (this._source.armor.value ?? 0) + (this.magicAvailable ? (this.armor.magicalBonus ?? 0) : 0);
     this.type.label = CONFIG.DND5E.equipmentTypes[this.type.value]
       ?? game.i18n.localize(CONFIG.Item.typeLabels.equipment);
   }
@@ -184,16 +185,6 @@ export default class EquipmentData extends ItemDataModel.mixin(
       (this.isArmor || this.isMountable) ? (this.parent.labels?.armor ?? null) : null,
       this.properties.has("stealthDisadvantage") ? game.i18n.localize("DND5E.Item.Property.StealthDisadvantage") : null
     ];
-  }
-
-  /* -------------------------------------------- */
-
-  /**
-   * Effective armor class, taking any magical bonus into account.
-   * @type {number}
-   */
-  get effectiveAC() {
-    return this.armor.value + (this.magicAvailable ? (this.armor.magicalBonus ?? 0) : 0);
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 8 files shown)


module/data/item/equipment.mjs
```diff
@@ -38,21 +41,22 @@ export default class EquipmentData extends ItemDataModel.mixin(
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
       type: new ItemTypeField({value: "light", subtype: false}, {label: "DND5E.ItemEquipmentType"}),
-      armor: new foundry.data.fields.SchemaField({
-        value: new foundry.data.fields.NumberField({required: true, integer: true, min: 0, label: "DND5E.ArmorClass"}),
-        dex: new foundry.data.fields.NumberField({required: true, integer: true, label: "DND5E.ItemEquipmentDexMod"})
+      armor: new SchemaField({
+        value: new NumberField({required: true, integer: true, min: 0, label: "DND5E.ArmorClass"}),
+        magicalBonus: new NumberField({min: 0, integer: true, label: "DND5E.MagicalBonus"}),
```
krbz999, line 46: Are there no items that grant a bonus equal to some roll data like your Charisma modifier? I feel like there must be. In which case it would need to preferably be a FormulaField.

arbron: I have never seen anything that adds a variable bonus equally to both attack and damage.

templates/items/equipment.hbs
```diff
@@ -129,6 +129,15 @@
                     {{numberInput system.armor.value name="system.armor.value"}}
                 </div>
             </div>
+
+            {{#if properties.mgc.selected}}
+            <div class="form-group">
+                <label>{{ localize "DND5E.MagicalBonus" }}</label>
+                <div class="form-fields">
+                    {{ numberInput system.armor.magicalBonus name="system.armor.magicalBonus" placeholder=0 }}
```
Fyorl, line 137:
```suggestion
                    {{ numberInput system.armor.magicalBonus name="system.armor.magicalBonus" placeholder="0" }}
```

templates/items/weapon.hbs
```diff
@@ -128,6 +128,15 @@
             {{> "dnd5e.item-mountable"}}
             {{/if}}
 
+            {{#if properties.mgc.selected}}
+            <div class="form-group">
+                <label>{{ localize "DND5E.MagicalBonus" }}</label>
+                <div class="form-fields">
+                    {{ numberInput system.magicalBonus name="system.magicalBonus" placeholder=0 }}
```
Fyorl, line 135:
```suggestion
                    {{ numberInput system.magicalBonus name="system.magicalBonus" placeholder="0" }}
```

Fyorl (changes requested):
Looking pretty good, I've got some comments for discussion. Doesn't have to be done in this PR, but do we anticipate adding a magical bonus to consumables as well? There's a bit of a gap with magical ammunition here.

module/data/item/equipment.mjs
krbz999: This is on armor, for AC bonus. :)

Fyorl (approved):

Issue #997, opened 2021-02-11
Add actor data templating for description fields


schultzcole:
This proposal is for a feature by which you could use some syntax in a actor or item description field to reference properties of that document (or it's parent document).

While editing the description, it may look like this (example comes from the "Acid Spray" SRD monster feature):

> The {parentData.name} spits acid in a line that is 30 ft. long and 5 ft. wide. Each creature in that line must make a Dexterity saving throw.

When not editing the description (and when shown in a chat card, etc.), this text would be formatted to replace these templates:

> The Ankheg spits acid in a line that is 30 ft. long and 5 ft. wide. Each creature in that line must make a Dexterity saving throw.

This would make it much easier to build up a custom actor based on the provided monster features and items compendiums, or create your own items which can generically be used for multiple different actors without needing to manually adjust all of the items one by one. Additionally, in cases where a user takes an existing SRD monster from the compendium and changes it's name, the item features also change for "free".

PR #3152, opened 2024-02-29
[#997] Add lookup enricher for including non-roll data

arbron:
Adds a new enricher that allows for looking up values within roll data and displaying them as is:

Screenshot 2024-02-28 at 17 06 01

This includes an extra `extended` option for `getRollData` that includes the creature's name. This could be extended in the future to include additional data, like perhaps the localized creature type or anything else that would be useful in this context but not in standard roll formulas.

### DIFF
--- a/module/data/abstract.mjs
+++ b/module/data/abstract.mjs
@@ -341,10 +341,9 @@ export class ActorDataModel extends SystemDataModel {
    * @param {object} [options]
    * @param {boolean} [options.deterministic] Whether to force deterministic values for data properties that could be
    *                                          either a die term or a flat term.
-   * @param {boolean} [options.extended]      Include extra values like "name" that aren't useful in formulas.
    * @returns {object}
    */
-  getRollData({ deterministic=false, extended=false }={}) {
+  getRollData({ deterministic=false }={}) {
     const data = { ...this };
     data.prof = new Proficiency(this.attributes?.prof ?? 0, 1);
     if ( deterministic ) data.prof = data.prof.flat;
@@ -498,12 +497,11 @@ export class ItemDataModel extends SystemDataModel {
    * @param {object} [options]
    * @param {boolean} [options.deterministic] Whether to force deterministic values for data properties that could be
    *                                          either a die term or a flat term.
-   * @param {boolean} [options.extended]      Include extra values like "name" that aren't useful in formulas.
    * @returns {object}
    */
-  getRollData({ deterministic=false, extended }={}) {
+  getRollData({ deterministic=false }={}) {
     if ( !this.parent.actor ) return null;
-    const actorRollData = this.parent.actor.getRollData({ deterministic, extended });
+    const actorRollData = this.parent.actor.getRollData({ deterministic });
     const data = { ...actorRollData, item: { ...this } };
     return data;
   }
--- a/module/data/abstract.mjs
+++ b/module/data/abstract.mjs
@@ -341,9 +341,10 @@ export class ActorDataModel extends SystemDataModel {
    * @param {object} [options]
    * @param {boolean} [options.deterministic] Whether to force deterministic values for data properties that could be
    *                                          either a die term or a flat term.
+   * @param {boolean} [options.extended]      Include extra values like "name" that aren't useful in formulas.
    * @returns {object}
    */
-  getRollData({ deterministic=false }={}) {
+  getRollData({ deterministic=false, extended=false }={}) {
     const data = { ...this };
     data.prof = new Proficiency(this.attributes?.prof ?? 0, 1);
     if ( deterministic ) data.prof = data.prof.flat;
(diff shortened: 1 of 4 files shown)


Fyorl:
I think we had some convention around whether to use a forward slash or not, right? If it's something interactive, we use the slash, otherwise we don't? The passive skill check enricher is a bit of an exception because it's reusing most of the `[[/check]]` logic, which is normally interactive.

Is there still an issue in v11 preventing us from using it without a slash, e.g. `[[lookup @name]]`? What if we used the `@` symbol instead and omitted the `lookup`, e.g. `[[@name capitalize]]`?

arbron:
> Is there still an issue in v11 preventing us from using it without a slash, e.g. `[[lookup @name]]`? What if we used the `@` symbol instead and omitted the `lookup`, e.g. `[[@name capitalize]]`?

I like the `[[@name]]` format, nice and simple.

Fyorl:
> I like the `[[@name]]` format, nice and simple.

Now that I've said this, I wonder, does this not already exist? The inline roll syntax should actually already work with roll data. For example, this already works:

```
[[@abilities.cha.mod]]
```

So I suppose we maybe can't use it after all? What would happen to existing inline rolls that actually have non-deterministic terms like `[[@abilities.cha.mod + 2d6]]`?

arbron:
> > I like the `[[@name]]` format, nice and simple.
> 
> Now that I've said this, I wonder, does this not already exist? The inline roll syntax should actually already work with roll data. For example, this already works:
> 
> ```
> [[@abilities.cha.mod]]
> ```
> 
> So I suppose we maybe can't use it after all? What would happen to existing inline rolls that actually have non-deterministic terms like `[[@abilities.cha.mod + 2d6]]`?

Indeed. How would you feel about `{{@name}}` (`{{@abilities.cha.mod}}`)? If you're not a fan of the curly brackets we can just use `[[lookup @name]]` and that will work fine.

Fyorl:
> Indeed. How would you feel about `{{@name}}` (`{{@abilities.cha.mod}}`)? If you're not a fan of the curly brackets we can just use `[[lookup @name]]` and that will work fine.

I believe Atro prefers the double square brackets for all of these styles of enrichers for consistency, so let's go with the `[[lookup @name]]` style.

MaxPat931:
What about borrowing `&` from `&Reference`, since these are kinda used to reference character data and differentiates it from rollable data? eg `[[&name]]`

arbron:
@Fyorl Adjusted to use `[[lookup @name]]` style

module/data/abstract.mjs
```diff
@@ -341,9 +341,10 @@ export class ActorDataModel extends SystemDataModel {
    * @param {object} [options]
    * @param {boolean} [options.deterministic] Whether to force deterministic values for data properties that could be
    *                                          either a die term or a flat term.
+   * @param {boolean} [options.extended]      Include extra values like "name" that aren't useful in formulas.
```
Fyorl, line 344: It felt a little off to me to have an option that adds data to `getRollData` that is not useful for rolls, so I was going to suggest maybe `getReferenceData` that calls `getRollData` and then adds supplementary data on top. But actually, there's probably a lot of stuff under `system` that is not useful for *rolls* specifically, and we include it all.

I don't really know what else we would include as `extended` data. There are only fields that exist at the top level, and there aren't many of those (some we already include, like `flags` now), as well as some derived properties like `statuses` which we may end up including anyway (#3122).

So maybe we should just get rid of the `extended` option and always include `name`?

module/documents/item.mjs
```diff
@@ -1837,15 +1837,19 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
-    if ( this.system.getRollData ) data = this.system.getRollData({ deterministic });
+    if ( this.system.getRollData ) data = this.system.getRollData({ deterministic, extended });
     else {
       if ( !this.actor ) return null;
-      data = { ...this.actor.getRollData({ deterministic }), item: { ...this.system } };
+      data = { ...this.actor.getRollData({ deterministic, extended }), item: { ...this.system } };
+    }
+    if ( data?.item ) {
+      data.item.flags = { ...this.flags };
+      if ( extended ) data.item.name = this.name;
```
Fyorl, line 1851: Do we want to include some logic for unidentified items here?

```suggestion
      if ( extended ) data.item.name = game.user.isGM || (this.system.identified !== false) ? this.name : this.system.unidentified.name;
```

Fyorl (changes requested):
Looks good, thank you Jeff. Please see the comments for discussion.

module/documents/item.mjs
arbron: The name is already swapped during data preparation to use the unidentified name, so this is handled automatically (unless we wanted to handle this differently and have it always show the normal name to GMs, currently we show the unidentified name to everyone if it is unidentified).

Fyorl: Oh that's right, I had forgotten about that. Maybe always showing the GM the real name is desirable, I don't really know, so I guess we just leave it for now.

Fyorl (approved):

PR #3154, opened 2024-02-29
Move trackable attributes to setup

### LINKED ISSUE #3153: `_configureTrackableAttributes` Ignores Custom Configs
`_configureTrackableAttributes` in `dnd5e.mjs` occurs before modules can customise the `CONFIG.DND5E.abilities` resulting in the abilities in the trackable attributes not reflecting the state of `CONFIG.DND5E.abilities`. Can `_configureTrackableAttributes` be moved to update `CONFIG.Actor.trackableAttributes` as it's needed rather than on init, so that it can pull the most recent state of  `CONFIG.DND5E.abilities`.

Larkinabout:
Closes #3153.

Moved `_configureTrackableAttributes` as well as `_configureConsumableAttributes` and `_configureStatusEffects` for good measure to the `setup` hook to allow modules to update `CONFIG.DND5E` on `init` before these are generated. Saves having to update these objects as they'd most likely need to be rebuilt or looped through to remove attributes that no longer exist in the config.

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -86,12 +86,12 @@ Hooks.once("init", function() {
   // Configure tooltips
   game.dnd5e.tooltips = new Tooltips5e();
 
+  // Set up status effects
+  _configureStatusEffects();
+
   // Remove honor & sanity from configuration if they aren't enabled
   if ( !game.settings.get("dnd5e", "honorScore") ) delete DND5E.abilities.hon;
   if ( !game.settings.get("dnd5e", "sanityScore") ) delete DND5E.abilities.san;
-
-  // Patch Core Functions
-  Combatant.prototype.getInitiativeRoll = documents.combat.getInitiativeRoll;
 
   // Register Roll Extensions
   CONFIG.Dice.rolls.push(dice.D20Roll);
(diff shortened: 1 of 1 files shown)


dnd5e.mjs
```diff
@@ -320,6 +313,13 @@ function _configureStatusEffects() {
  * Prepare attribute lists.
  */
 Hooks.once("setup", function() {
+  // Configure trackable & consumable attributes.
+  _configureTrackableAttributes();
+  _configureConsumableAttributes();
+
+  // Set up status effects
+  _configureStatusEffects();
```
krbz999, line 321: This would prevent users from pushing to `CONFIG.statusEffects` on the `init` hook unless they add their new status to `CONFIG.DND5E.statusEffects`.

Larkinabout: Good point. Neither is a good option as `CONFIG.DND5E.statusEffects` is probably the right place to add for the system based on what it's doing, but not for everyone currently pushing to `CONFIG.statusEffects`. Equally, having it in the `init` means subsequently adding into `CONFIG.DND5E.statusEffects` isn't reflected in `CONFIG.statusEffects`. I'll have a look to see what instances are using `CONFIG.statusEffects` to see if it's an issue like with trackable attributes.

Larkinabout: Looks like `CONFIG.DND5E.statusEffects` is only used in this instance to populate `CONFIG.statusEffects`, so the answer for users/modules seems to be to push to `CONFIG.statusEffects` instead of `CONFIG.DND5E.statusEffects`. I'll move `_configureStatusEffects` back to the `init`.

Fyorl (approved):

PR #3155, opened 2024-02-29
Allow custom currency icons

Larkinabout:
Allow for users/modules to customise the currency icons by adding the icon to the config and replacing the hardcoded CSS with a style tag for the background url. Related to #3145.


Fyorl:
This is a bit more complicated, unfortunately, because there are journals out there that use these currency icon styles, as well as the award enricher I think.

There might be less complicated/ugly options, but we could make these styles entirely dynamic.

```js
Hooks.on("setup", () => {
  const style = document.createElement("style");
  style.innerHTML = `
    :is(.dnd5e2, .dnd5e2-journal) :is(i, span).currency {
      ${Object.entries(CONFIG.DND5E.currencies).reduce((rules, [k, { icon }]) => {
        return `${rules}\n.${k} { background-image: url("${icon}"); }`;
      }, "")}
    }
  `;
  document.head.append(style);
});
```

Larkinabout:
Sounds good to me.

less/v2/apps.less
```diff
@@ -664,12 +664,6 @@
   height: 1em;
   display: inline-block;
   background: transparent no-repeat center / contain;
-
-  &.cp { background-image: url("icons/currency/copper.webp"); }
-  &.sp { background-image: url("icons/currency/silver.webp"); }
-  &.ep { background-image: url("icons/currency/electrum.webp"); }
-  &.gp { background-image: url("icons/currency/gold.webp"); }
-  &.pp { background-image: url("icons/currency/platinum.webp"); }
```
arbron, line 672: We use these in some journals, such as in Phandelver & Below, for explaining currencies, so they shouldn't be removed.

module/applications/award.mjs
```diff
@@ -57,8 +57,13 @@ export default class Award extends DialogMixin(FormApplication) {
...
     context.CONFIG = CONFIG.DND5E;
-    context.currency = Object.entries(CONFIG.DND5E.currencies).reduce((obj, [k, { label }]) => {
-      obj[k] = { label, value: this.options.currency ? this.options.currency[k] : this.object?.system.currency[k] };
+    context.currency = Object.entries(CONFIG.DND5E.currencies).reduce((obj, [k, { label, icon }]) => {
+      obj[k] = {
+        label,
+        icon,
+        value: this.options.currency
+          ? this.options.currency[k]
+          : this.object?.system.currency[k] };
```
arbron, line 66:
```suggestion
    context.currency = Object.entries(CONFIG.DND5E.currencies).reduce((obj, [k, { label, icon }]) => {
      obj[k] = {
        label, icon,
        value: this.options.currency ? this.options.currency[k] : this.object?.system.currency[k]
      };
```

module/config.mjs
```diff
@@ -1404,27 +1404,32 @@ DND5E.currencies = {
   pp: {
```
arbron, line 1404: Be sure to update the type definition above to document the added icon:

```
 * @property {string} icon          Icon representing the currency in the interface.
```

templates/apps/currency-manager.hbs
```diff
@@ -21,7 +21,7 @@
                 {{#each currency}}
                 <label>
                     {{#with (lookup (lookup @root.CONFIG.currencies @key) "label") as |label|}}
-                    <i class="currency {{@key}}" data-tooltip="{{label}}" aria-label="{{label}}"></i>
+                    <i class="currency" style="background-image: url('{{ lookup (lookup @root.CONFIG.currencies @key) "icon" }}');" data-tooltip="{{label}}" aria-label="{{label}}"></i>
                     {{/with}}
```
arbron, line 25:
```javascript

                    {{#with (lookup @root.CONFIG.currencies @key) as |currency|}}
                    <i class="currency" style="background-image: url('{{ lookup currency "icon" }}');"
                       data-tooltip="{{ lookup currency 'label' }}" aria-label="{{ lookup currency 'label' }}"></i>
                    {{/with}}
```

arbron (changes requested):
You'll also need to fix the award enricher to use the correct icons.

Larkinabout:
So, forget making them dyanmic, leave the styles as is and replace the references within the dnd5e system to use the config icon?

It should mean that if currencies are removed from the config, everything that points to those styles will still work. With the understanding that a journal won't reflect homebrew currencies, which I don't think anyone would expect anyway.

less/v2/apps.less
Larkinabout: Added back in.

arbron (approved):

PR #3157, opened 2024-02-29
[#3156] Add setting to show attack roll result (AC check) card to players

### LINKED ISSUE #3156: Option to show the attack result (AC check) to players
Right now only DM can see the AC check card that is automatically created after the attack hits or doesn't.
image

For DMs that play "in the open" (and my players instantly remember the AC after the first hit lands anyway) it would be nice to have an option to display this card to players as well.
That would considerably speed up the combat, since players won't have to wait for the DM to say "it hits" to make a damage roll.

### LINKED ISSUE #3868: Challenge Visibility setting not taken into account for attack success/failure if player initiates item-use chat card
Currently, `ChatMessage5e`'s `shouldDisplayChallenge` getter will return `true` no matter what for a user if the owner of the chat card is the user in question. In the `_highlightCriticalSuccessFailure` function, the originating message is checked for `shouldDisplayChallenge`, which means that for saving throws, players will almost always not be the owners of the cards, so all is well. For _attack_ rolls however, the players will almost always be owners of the originating message, so the "Challenge Visibility" setting is bypassed completely in that case:
```js
get shouldDisplayChallenge() {
  if ( game.user.isGM || (this.user === game.user) ) return true; // <-- will always evaluate to `true` on a normal player attack flow
  switch ( game.settings.get("dnd5e", "challengeVisibility") ) {
    case "all": return true;
    case "player": return !this.user.isGM;
    default: return false;
  }
}
```
It seems to me that that element of the check can be removed without any issue (I'm not sure what the intended use case that covers is). As far as I can tell, the only difference would be that if the setting is "Hide All", it would in fact hide _all_ highlighting for a user; which strikes me as the correct implementation.

pbasov:
- Closes #3156
- Closes #3868

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2119,6 +2119,13 @@
 "SETTINGS.5eAutoCollapseCardN": "Collapse Item Cards in Chat",
 "SETTINGS.5eAutoSpellTemplateL": "When a spell is cast, defaults to begin the process to create the corresponding Measured Template if any (requires TRUSTED or higher player role)",
 "SETTINGS.5eAutoSpellTemplateN": "Always place Spell Template",
+"SETTINGS.5eAttackRollCheckVisibility": {
+  "Name": "Show attack roll result to players",
+  "Hint": "Control visibility of AC check cards.",
+  "GM": "Show only to GM",
+  "Players": "Show to players",
+  "PlayersHideAC": "Show to players, but hide AC"
+},
 "SETTINGS.5eChallengeVisibility": {
   "Name": "Challenge Visibility",
   "Hint": "Control what roll DCs are visible to the players and whether successes/failures are highlighted.",
--- a/lang/en.json
+++ b/lang/en.json
@@ -2119,12 +2119,12 @@
 "SETTINGS.5eAutoCollapseCardN": "Collapse Item Cards in Chat",
 "SETTINGS.5eAutoSpellTemplateL": "When a spell is cast, defaults to begin the process to create the corresponding Measured Template if any (requires TRUSTED or higher player role)",
 "SETTINGS.5eAutoSpellTemplateN": "Always place Spell Template",
-"SETTINGS.5eAttackRollCheckVisibility": {
-  "Name": "Show attack roll result to players",
-  "Hint": "Control visibility of AC check cards.",
-  "GM": "Show only to GM",
-  "Players": "Show to players",
-  "PlayersHideAC": "Show to players, but hide AC"
+"SETTINGS.5eAttackRollVisibility": {
+  "Name": "Attack Result Visibility",
+  "Hint": "Control visibility of attack roll results in chat cards for players.",
+  "All": "Show results & target ACs",
+  "HideAC": "Show only results",
+  "None": "Hide all"
 },
 "SETTINGS.5eChallengeVisibility": {
   "Name": "Challenge Visibility",
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -359,13 +359,9 @@ export default class ChatMessage5e extends ChatMessage {
   _enrichAttackTargets(html) {
     const attackRoll = this.rolls[0];
     const targets = this.getFlag("dnd5e", "targets");
-    const visibility = game.settings.get("dnd5e", "attackRollCheckVisibility");
-
-    const visibilityCondition = game.user.isGM ||
-    (visibility === "players" && !game.user.isGM) ||
-    (visibility === "players_hideac" && !game.user.isGM);
-
-    if (!visibilityCondition || !(attackRoll instanceof dnd5e.dice.D20Roll) || !targets?.length) return;
+    const visibility = game.settings.get("dnd5e", "attackRollVisibility");
+    const isVisible = game.user.isGM || (visibility !== "none");
+    if ( !isVisible || !(attackRoll instanceof dnd5e.dice.D20Roll) || !targets?.length ) return;
     const tray = document.createElement("div");
     tray.classList.add("dnd5e2");
     tray.innerHTML = `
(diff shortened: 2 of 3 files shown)


module/settings.mjs
```diff
@@ -28,6 +28,16 @@ export function registerSystemSettings() {
...
   });
 
+  // Attack AC check visibility
+  game.settings.register("dnd5e", "attackRollCheckVisibility", {
+    name: "SETTINGS.5eAttackRollCheckVisibility.Name",
+    hint: "SETTINGS.5eAttackRollCheckVisibility.Hint",
+    scope: "world",
+    config: true,
+    default: false,
+    type: bool
```
krbz999, line 38:
```suggestion
    type: Boolean
```

pbasov:
@krbz999 Thanks for quick review (were hoping to grab a built minjs, but you don't build them for PRs)
Fixed the code, tested on my own instance, works as intended without reload, PTAL.

Player (left) side, GM side (right) False/True:
imageimage

Can be improved by an addional logic to hide/show precise AC number for players, but i think it's a separate commit.

module/settings.mjs
```diff
@@ -28,6 +28,20 @@ export function registerSystemSettings() {
     }
   });
 
+  game.settings.register("dnd5e", "attackRollCheckVisibility", {
+    name: "SETTINGS.5eAttackRollCheckVisibility.Name",
+    hint: "SETTINGS.5eAttackRollCheckVisibility.Hint",
+    scope: "world", // Applies to the entire world, not just individual users
+    config: true, // Shows this setting in the Configure Settings menu
+    default: "gm", // Sets "gm" as the default value
+    type: String, // Specifies that the setting is a string choice
```
arbron, line 37: Probably don't need these comments.

```suggestion
    scope: "world",
    config: true,
    default: "none",
    type: String,
```

lang/en.json
```diff
@@ -1647,6 +1647,13 @@
 "SETTINGS.5eAutoCollapseCardN": "Collapse Item Cards in Chat",
 "SETTINGS.5eAutoSpellTemplateL": "When a spell is cast, defaults to begin the process to create the corresponding Measured Template if any (requires TRUSTED or higher player role)",
 "SETTINGS.5eAutoSpellTemplateN": "Always place Spell Template",
+"SETTINGS.5eAttackRollCheckVisibility": {
+  "Name": "Show attack roll result to players",
+  "Hint": "Control visibility of AC check cards.",
```
arbron, line 1652: Just to make the name match closer to other settings we can move more information into the hint:

```suggestion
  "Name": "Attack Result Visibility",
  "Hint": "Control visibility of attack roll results in chat cards for players.",
```

lang/en.json
```diff
@@ -1647,6 +1647,13 @@
...
+  "GM": "Show only to GM",
+  "Players": "Show to players",
+  "PlayersHideAC": "Show to players, but hide AC"
```
arbron, line 1655: Tweaking the wording of these to match the challenge visibility setting more.

```suggestion
  "All": "Show results & target ACs",
  "HideAC": "Show only results",
  "None": "Hide all"
```

module/settings.mjs
```diff
@@ -28,6 +28,20 @@ export function registerSystemSettings() {
...
+    choices: {
+      gm: "SETTINGS.5eAttackRollCheckVisibility.GM",
+      players: "SETTINGS.5eAttackRollCheckVisibility.Players",
+      players_hideac: "SETTINGS.5eAttackRollCheckVisibility.PlayersHideAC"
```
arbron, line 41: To match the wording changes mentioned above and keep the order similar to `Challenge Visibility`:

```suggestion
      all: "SETTINGS.5eAttackRollCheckVisibility.All",
      hideAC: "SETTINGS.5eAttackRollCheckVisibility.HideAC",
      none: "SETTINGS.5eAttackRollCheckVisibility.None"
```

module/documents/chat-message.mjs
```diff
@@ -274,10 +274,17 @@ export default class ChatMessage5e extends ChatMessage {
...
     const attackRoll = this.rolls[0];
     const targets = this.getFlag("dnd5e", "targets");
-    if ( !game.user.isGM || !(attackRoll instanceof dnd5e.dice.D20Roll) || !targets?.length ) return;
+    const visibility = game.settings.get("dnd5e", "attackRollCheckVisibility");
+
+    const visibilityCondition = game.user.isGM ||
+    (visibility === "players" && !game.user.isGM) ||
+    (visibility === "players_hideac" && !game.user.isGM);
+    
+    if (!visibilityCondition || !(attackRoll instanceof dnd5e.dice.D20Roll) || !targets?.length) return;
```
arbron, line 283: That check can be simplified a bunch.

```suggestion
    const visibility = game.settings.get("dnd5e", "attackRollCheckVisibility");
    const isVisible = game.user.isGM || (visiblity !== "none");
    if ( !isVisible || !(attackRoll instanceof dnd5e.dice.D20Roll) || !targets?.length ) return;
```

module/documents/chat-message.mjs
```diff
@@ -274,10 +274,17 @@ export default class ChatMessage5e extends ChatMessage {
...
     const evaluation = document.createElement("ul");
     evaluation.classList.add("dnd5e2", "evaluation");
     evaluation.innerHTML = targets.reduce((str, { name, img, ac, uuid }) => {
+      const displayedAC = (visibility === "players_hideac" && !game.user.isGM) ? "" : ac;
```
arbron, line 287: Might be clearer to just change `ac` in place:

```suggestion
      if ( !game.user.isGM && (visibility !== "all") ) ac = "";
```

module/documents/chat-message.mjs
```diff
@@ -291,7 +298,7 @@ export default class ChatMessage5e extends ChatMessage {
           </div>
           <div class="ac">
             <i class="fas fa-shield-halved"></i>
-            <span>${ac}</span>
+            <span>${displayedAC}</span>
           </div>
```
arbron, line 302: It would be good to remove this whole block that way players don't see just a random shield icon:

```javascript
          ${ac ? `
          <div class="ac">
            <i class="fas fa-shield-halved"></i>
            <span>${ac}</span>
          </div>
          ` : ""}
```

arbron (changes requested):

Fyorl (approved):

PR #3159, opened 2024-02-29
Convert several attribute fields to use `RollConfigField`

krbz999:
Removed `bonus` field from `RollConfigField`. This would have conflicted with those properties, like `skills`, that have `bonus` as derived data. Instead individual properties define `bonuses` with 'check' and/or 'save'.

Seems the only property that has `bonus` as an actual field is `attributes.init`, but it doesn't - in my opinion - seem worth it to migrate this to `attributes.init.bonuses.check`.


Fyorl:
Where did we land on `mode` vs. `roll.mode`. I think I'm leaning towards the latter just because `foo.mode` is a little bit general and ambiguous, whereas `foo.roll.mode` is pretty clear what the mode refers to.

Fyorl (approved):

PR #3164, opened 2024-03-01
Add dark mode styles for character sheets.

Fyorl:
This adds the overall infrastructure for including dark mode styles and having them work depending on whether dark mode is applied via OS styles, via system override, or via individual sheet theme override. It also contains enough v2 styles to fully style the character sheet in dark mode.


arbron (changes requested):
Great! The system seems to be working well, noticed a few minor visual issues:

Empty Race & Background pills are black text on very dark gray:

Screenshot 2024-03-01 at 09 22 52

The buttons in Prosemirror inherit the transition style from the other buttons, leading to a weird transition from light to dark styling when they first appear:

https://github.com/foundryvtt/dnd5e/assets/19979839/eb9d7b87-2708-489f-90e6-7dc3ce7f65e0

Context menus are still in light styling. Perhaps in our `_setPosition` method on our new context menus we can use the target of the event, find the closest `[data-theme]`, and set the theme on the context menu based on that.

Screenshot 2024-03-01 at 09 25 12

arbron:
Just going to throw screenshots from high contrast mode here just to make sure we handle it before release:

Screenshot 2024-03-01 at 09 33 12

Screenshot 2024-03-01 at 09 33 20

Fyorl:
The text at least should be an easy fix. What do you think of using a similar technique with `matchMedia` that adds a `.dnd5e-hc` class to the body? Then we could layer these styles more easily, rather than having media queries playing havoc with specificity.

```css
.dnd5e-theme-dark .dnd5e2 {
  --color-text-dark-primary: var(--dnd5e-color-blue-white);
}

.dnd5e-hc.dnd5e-theme-dark .dnd5e2 {
  --color-text-dark-primary: white;
}
```

arbron (approved):

Issue #3037, opened 2024-02-12
When using an item that requires concentration, supply appropriate context in the `AbilityUseDialog` for if, when, and why its use would end or replace concentration


krbz999:
Related to #3028.

Circumstances in which an item may affect concentration include:
- It requires concentration, but you are not concentrating; using the item *will initiate* concentration.
- It requires concentration, but you are concentrating on a different item; using the item *will replace* concentration.
- It requires concentration, but you are concentrating on the same item at a different level; the user should be warned that concentration will end *unless* they cast the spell at the same level.

Fyorl:
> It requires concentration, but you are concentrating on the same item at a different level; the user should be warned that concentration will end _unless_ they cast the spell at the same level.

Could you clarify this? It's not a rule I'm aware of.

krbz999:
> > It requires concentration, but you are concentrating on the same item at a different level; the user should be warned that concentration will end _unless_ they cast the spell at the same level.
> 
> Could you clarify this? It's not a rule I'm aware of.

Any concentration spell being cast would end previous concentration. When the second item is the same item, it may be worthwhile considering not breaking concentration if it gets cast at the same level.

So to clarify, this is not a rule, but a convenience feature. In case a user may want to simply re-display the item, let's say they forgot to pop out the message for example.

krbz999:
Agreed

Issue #3034, opened 2024-02-12
Allow for modules to extend or duplicate the system's concentration automation


krbz999:
Related to #3028.

It would be good if whatever method is created to track concentration was able to be extended such that, for example, an actor could
- track concentration on two items at the same time.
- have one 'type' of concentration that is disconnected from regular concentration (example: the Witch class from Valda's Spire of Secrets, which tracks concentration on their hexes separately from spells).

While not part of any official material, it would be good if we made an effort not to directly *prevent* a module's ability to do this, or better yet exposed the relevant API such that a module could duplicate the automation.

etiquettestartshere:
While it may not be official official, the Moon Domain Cleric sublcass from [Tal’Dorei Campaign Setting Reborn](https://www.dndbeyond.com/marketplace/sourcebooks/taldorei-campaign-setting-reborn) (available on dndbeyond) has "Mind of Two Moons", a divinity feature that allows for a second 'type' of concentration. It would be nice if this feature was supported, or at least not disallowed.

JPMeehan:
MCDM talent is another example of multi-concentration, as is the official Niv-Mizzet creature.

Issue #3028, opened 2024-02-12
Concentration automation


Fyorl:
### Must Haves
- [x] When casting a spell, create an AE that can act as the origin of any effects applied by that spell.
  - Eagerly created for concentration spells.
  - Lazily created (i.e. if one of the spell's effects is applied to any target) for non-concentration spells that have a duration.
- [x] If the spell requires concentration, the AE should apply the 'concentrating' status.
- [x] Deleting this AE should delete all AEs that were connected to it.
- [x] Create an extra field on the Actor data model for a bonus to concentration saves specifically. Probably in special traits?
- [x] When an Actor takes damage, create a roll request chat message to roll a Constitution (concentration) saving throw with an auto-calculated DC. Send this message to any owner of the Actor.
- [x] Display a warning message in the spell use dialog that reminds a player if their character is already concentrating on something.
- [x] Automatically end concentration if another spell requiring concentration is cast regardless.
- [x] The spell use dialog should allow for an option to not have this instance of the spell require concentration.

### Ideal
- [x] Extend this functionality to any item with the `concentration` property.

### Additional Notes
- Auto-calculating concentration save DC is dependant on us enabling the resistances/immunities/vulnerabilities automation. Not too much work left to do there though. Mostly UI/UX.

schultzcole:
When the origin effect on the casting actor ends, do we also end any effects on other actors that have the origin effect as their origin?

krbz999:
The chat message prompt for saving throws should also have utility buttons for:
- ending concentration easily (perhaps store the uuid of the concentration effect).
- removing templates created from the item on which the actor was concentrating (easy comparison due to `templateDocument.flags.dnd5e.origin`).

> Extend this functionality to any item with the concentration property.

We should make use of `Item5e#requiresConcentration`. There should not be an issue extending this to all item types from the very beginning.

Fyorl:
> When the origin effect on the casting actor ends, do we also end any effects on other actors that have the origin effect as their origin?

Good catch, that is a glaring omission from the requirements. Just to be clear though, we are not intending on handling AE expiration with this epic. Handling manual deletion of AEs should be fine though.

JPMeehan:
For automatically ending concentration when casting another spell with concentration: Making that somehow configurable or its own function would really help with trying to manage the MCDM talent, which is designed around maintaining multiple concentrations. There's also at least one official monster with multiple concentrations, Niv-Mizzet.

PR #3166, opened 2024-03-01
[#3028, #3034, #3037] Concentration prompts

krbz999:
Related to #3028, #3034, #3037.

Adds prompt for concentration to any item that requires it.

image

Using the item creates an effect with the name of the item appended, and the 'concentrating' status.

image

Trying to use a second, different item *while not at the limit of how many items you can concentrate on* will prompt you to _optionally_ replace the effect. The example actor, Steve, here has a limit of 2.

image

Trying to use a third item while at the limit of how many items you can concentrate on will prompt you to replace one. There won't be a blank option this time.

image

The concentration effect(s) can be removed in the token HUD, with extra functionality added.

https://github.com/foundryvtt/dnd5e/assets/50169243/a853f9ee-a414-4896-87a4-55521dadc3a6

If you click the concentration effect while you have only 1, it will remove it. If you have more than 1, you will be prompted to pick. The dialog will show the name of the item being concentrated on. If no such item exists (like if you toggled on the default concentration effect), it will display as 'Concentration with no item'.

image

If the actor has a 'limit' of zero, then this warning will appear:

image


tposney:
Would it be possible to call some hooks when adding/removing concentration so modules could do their own stuff as necessary when concentration is applied/removed?

krbz999:
> Would it be possible to call some hooks when adding/removing concentration

There's likely going to be more work in that department, but I've added two hooks in the creation method. Worst case, the core hooks still apply and you can detect whether it is a concentration effect by looking at its `statuses`.

module/applications/item/ability-use-dialog.mjs
```diff
@@ -42,11 +42,19 @@ export default class AbilityUseDialog extends Dialog {
...
+
     const data = {
       item,
       ...config,
       slotOptions,
       resourceOptions,
+      concentration: {
+        show: concentrationOptions.length > 0,
+        options: concentrationOptions,
+        optional: (concentrationOptions.length < limit) ? "-" : null
```
Fyorl, line 56: Typically use an em dash for blank values.

```suggestion
        optional: (concentrationOptions.length < limit) ? "—" : null
```

module/documents/active-effect.mjs
```diff
@@ -354,7 +354,83 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+      return (d === item.id) || (d._id === item.id);
+    });
+
+    let effectData;
+    // Replace an existing effect if already concentrating on this item or if another effect has been chosen.
+    if ( existing ) {
+      const data = existing.flags.dnd5e?.itemData ?? {};
+      const sameItem = (data === item.id) || (data._id === item.id);
+      if ( sameItem ) {
+        effectData = foundry.utils.mergeObject({...existing.toObject(), duration: {}}, baseData);
```
Fyorl, line 397: I don't think I quite understand this logic. What do we gain from inheriting the existing effect's data vs. just creating a new effect that would be (seemingly) the same as the existing one if the origin Items are the same too, as we're asserting that they are here?

Fyorl (changes requested):
This is shaping up nicely, thank you Zhell.

module/applications/item/ability-use-dialog.mjs
```diff
@@ -89,6 +97,24 @@ export default class AbilityUseDialog extends Dialog {
...
   /* -------------------------------------------- */
 
+  /**
+   * Create an array of options for which concentration effect to end or replace.
+   * @param {Item5e} item     The item being used.
+   * @returns {object[]}      Array of concentration options.
+   * @private
+   */
+  static _createConcentrationOptions(item) {
+    const effects = item.actor.concentration?.effects ?? new Set();
```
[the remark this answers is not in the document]
krbz999: Oh I didn't see the `Group` actor was actually an `Actor5e` now!

module/documents/active-effect.mjs
```diff
@@ -399,24 +475,83 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+    } else if ( target.dataset?.statusId === "concentrating" ) {
+      const effects = actor.concentration?.effects ?? new Set();
+      if ( effects.size < 1 ) return;
+      event.preventDefault();
+      event.stopPropagation();
+      if ( effects.size === 1 ) {
+        effects.first().delete();
+        return;
+      }
+      const choices = effects.reduce((acc, effect) => {
```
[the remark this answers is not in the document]
krbz999: I think it's clever, but I'm not sure how I feel about using a dialog that is historically used only for item *usage* for just a prompt to delete an effect.

I suspect this can cause a lot of confusion for modules relying on it.

module/documents/item.mjs
```diff
@@ -1020,6 +1022,12 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     // Prepare card data & display it if options.createMessage is true
     const cardData = await item.displayCard(options);
 
+    // Initiate concentration.
+    let effects;
+    if ( config.beginConcentrating ) {
+      effects = await item.actor.beginConcentratingOn(item, { deleteId: config.endConcentration });
+    }
```
Fyorl, line 1029: Perhaps we should split the functionality for ending concentration out of the `beginConcentratingOn` method. I'm thinking ahead to when creatures fail a concentration saving throw, we probably want an easy way to simply end their concentration. Something like `Actor5e#endConcentrationOn(effect)` that takes an effect ID or instance.

```suggestion
    const effects = [];
    if ( config.beginConcentrating ) {
      effects.push(await item.actor.beginConcentratingOn(item));
      if ( config.endConcentration ) effects.push(await item.actor.endConcentrationOn(config.endConcentration));
    }
```

It allows the return value for `beginConcentratingOn` to be a bit nicer also, as it will always simply return the concentration effect that was created, rather than sometimes returning two effects.

Fyorl (changes requested):

module/documents/actor/actor.mjs
```diff
@@ -1105,6 +1133,88 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /* -------------------------------------------- */
 
+  /**
+   * Initiate concentration on an item.
+   * @param {Item5e} item                     The item on which to being concentration.
+   * @param {object} [effectData]             Effect data to merge into the created effect.
+   * @returns {Promise<ActiveEffect5e[]>}     A promise that resolves to the created and deleted effects.
```
Fyorl, line 1140:
```suggestion
   * @returns {Promise<ActiveEffect5e>}     A promise that resolves to the created effect.
```

module/documents/actor/actor.mjs
```diff
@@ -1105,6 +1133,88 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+
+    return effect;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * End concentration on an item.
+   * @param {Item5e|ActiveEffect5e|string} target     An item or effect to end concentration on, or id of an effect.
+   * @returns {Promise<ActiveEffect5e>}               A promise that resolves to the deleted effect.
```
Fyorl, line 1176:
```suggestion
   * @returns {Promise<ActiveEffect5e|void>}               A promise that resolves to the deleted effect.
```

Fyorl (approved):
Looks good, just two minor JSDoc things I caught.

Having had more time to read over these methods, the 'on' part of their names feels a bit clunky to me, we could consider:
* `beginConcentratingOn` -> `beginConcentrating` with a required argument of what to concentrate on of course.
* `endConcentrationOn` -> `endConcentration` with an optional first argument, otherwise it ends concentration on *all* effects.

Having a required argument for `endConcentationOn` does mean that calling code is more aware of multi-concentration creatures, but as this is an entirely homebrew concept, I don't think our API should be optimised around it.

I'll have a better sense of it when we get around to ending concentration on Actors as a result of taking damage, so these points are just for discussion/consideration for now.

module/documents/actor/actor.mjs
krbz999: > but as this is an entirely homebrew concept

Not true, as per an example in #3034, a creature from Guildmaster's Guide to Ravnica evidently can conc on two effects.

module/documents/actor/actor.mjs
krbz999: I changed this into always returning an array instead. I think that makes a lot of sense.

chrisk123999:
> Having a required argument for `endConcentationOn` does mean that calling code is more aware of multi-concentration creatures, but as this is an entirely homebrew concept, I don't think our API should be optimised around it.

[Niv-Mizzet](https://www.dndbeyond.com/monsters/175958-niv-mizzet) from Guildmasters' Guide to Ravnica is a non- hombrew creature with that ability.

Issue #893, opened 2020-12-21
[Feature Request] Provide a mechanism for actors like summons / familiars / companions / mounts to provide a reference back to another actor which owns or summoned them.


aaclayton:
Originally in GitLab by @thingscouldgetdicey1

lots of classes involved in dnd 5e need to have the abilty to be able to reference a players abilities. such as an artificer who needs the cannon to have 5x the hp of the lvl in artificer. or the beast master who needs to alter the AC according to the proficiency bonus of the player character. and other npcs using the actors proficiency for attack bonuses as with the wildfire spirit for wildfire druids. echo knights echo calculates the AC based off the proficiency level of the echo knight. there are lots more examples i could find where this is applicable. 
if these could be systemically added it would be less book keeping. as this could be added via a ddb importer to be automated if there was a mechnical systemic way of this to be referenced.  maybe this could be a feature if npcs are made friendly. 
if we could reference spell DC of the original actor that would be great too. please let me know if this would be possible.

PR #3167, opened 2024-03-01
[#893] Add summoning to item activation workflow

arbron:
Adds summoning configuration to the item usage configuration and the ability use dialog. Players will see a checkbox to indicate whether summoning should occur and if the item has more than one summoning profile, a dropdown for selecting which profile to use:

Summoning Ability Use Dialog

This also introduces a `prompt` boolean similar to the one used for measured templates and limited uses to control whether it is done by default or not. If the `"Place Summons"` checkbox is checked, then the selected summoning profile ID is stored in flags in the chat message.

A new `"Summon"` button has been added to the chat message to perform the summoning if not done automatically through the ability use process:

Summoning Chat Card

If the button is clicked it will try to summon whatever profile is specified in the flags, or if none is specified it will throw up the ability use dialog again to ask which should be used.

### DIFF
--- a/module/data/item/templates/action.mjs
+++ b/module/data/item/templates/action.mjs
@@ -2,7 +2,7 @@ import { ItemDataModel } from "../../abstract.mjs";
 import { FormulaField } from "../../fields.mjs";
 
 const {
-  ArrayField, BooleanField, DocumentIdField, IntegerSortField, NumberField, SchemaField, StringField
+  ArrayField, BooleanField, DocumentIdField, NumberField, SchemaField, StringField
 } = foundry.data.fields;
 
 /**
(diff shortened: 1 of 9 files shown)


module/data/item/templates/action.mjs
```diff
@@ -271,6 +275,16 @@ export default class ActionTemplate extends ItemDataModel {
 
   /* -------------------------------------------- */
 
+  /**
+   * Does this Item implement summoning as part of its usage?
+   * @type {boolean}
+   */
+  get hasSummoning() {
+    return (this.actionType === "summ") && this.summons.profiles.length;
```
krbz999, line 283: This can return an integer.

module/documents/item.mjs
```diff
@@ -2005,6 +2030,42 @@ export default class Item5e extends SystemDocumentMixin(Item) {
 
   /* -------------------------------------------- */
 
+  /**
+   * Handle summoning from a chat card.
+   * @param {ChatMessage5e} message  The message that was clicked.
+   * @param {Item5e} item            The item from which to summon.
+   */
+  static async _onChatCardSummon(message, item) {
+    let summonsProfile = message.getFlag("dnd5e", "use.summonsProfile");
```
Fyorl, line 2039: We should consider perhaps not storing the chosen summon profile on the card. I imagine one of the use-cases for clicking the chat card summon button will be that you chose the wrong summon, or want to make a different choice, and it won't be possible to do so with this design, if I understand it correctly.

Fyorl (changes requested):
Nice, just one comment for discussion, let me know what you think.

Fyorl (approved):

PR #3169, opened 2024-03-01
Adapt high contrast mode to dark mode

arbron:
This uses `matchMedia` to set a high contrast flag. This is handled as a flag list in preparation for additional theme flags.

Moves all the high contrast changes into separate folder to match dark mode separation.

### DIFF
--- a/module/settings.mjs
+++ b/module/settings.mjs
@@ -368,7 +368,7 @@ export function setTheme(element, theme="", flags=new Set()) {
   else delete element.dataset.theme;
 
   // Additional Flags
-  if ( matchMedia("(prefers-contrast: more)").matches ) flags.add("high-contrast");
+  if ( (element === document.body) && matchMedia("(prefers-contrast: more)").matches ) flags.add("high-contrast");
   for ( const flag of flags ) element.classList.add(`dnd5e-flag-${flag.slugify()}`);
   element.dataset.themeFlags = Array.from(flags).join(" ");
 }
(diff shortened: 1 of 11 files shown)


Fyorl (changes requested):
Thanks for taking care of this, Jeff. I noticed some typos which I've flagged. There are a couple of areas on the dark mode sheet that are still a bit low contrast, but they can be iterated on, they don't have to be resolved in this PR:

The subtitle colour:
image

Inactive control colour:
image

The tooltip styles do probably need resolving though, at least until we add some dark tooltip styles like we did with the context menu:
image

module/settings.mjs
```diff
@@ -348,16 +351,24 @@ export function registerDeferredSettings() {
...
   }
-  element.className = element.className.replace(/\bdnd5e-theme-\w+/g, "");
   if ( theme ) {
     element.classList.add(`dnd5e-theme-${theme.slugify()}`);
     element.dataset.theme = theme;
   }
   else delete element.dataset.theme;
+
+  // Additional Flags
+  if ( matchMedia("(prefers-contrast: more)").matches ) flags.add("high-contrast");
```
Fyorl, line 371: Noticed this accidentally just now but I think this also needs to go into the `element === document.body` block, otherwise sheets that toggle their themes back to 'Automatic' get stuck with this flag on.

Fyorl (approved):
Thanks very much, Jeff. I just noticed one additional issue, that in dark mode these inputs get the high-contrast border colour even without high contrast on.

image

But it's good to go after that.

PR #3172, opened 2024-03-02
Allow stored item data in chat cards to contain active effects

aaclayton:
I was working on an project using `dnd5e` which required me to store temporary item data in chat cards rather than relying on an actual item that exists on an actual actor.

This worked fine because of our chat card framework supporting the possibility of `storedData = message.getFlag("dnd5e", "itemData")` on the chat message itself, but this broke down if that item contains active effects which should apply to its usage because the `case "applyEffect"` handler required the effect to be retrieved by `uuid` rather than using a local `id` within the `item`.

This change prefers an effect local to the item of the chat card rather than a UUID lookup.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1910,7 +1910,9 @@ export default class Item5e extends SystemDocumentMixin(Item) {
       let targets;
       switch ( action ) {
         case "applyEffect":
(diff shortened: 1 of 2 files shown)


arbron (approved):

krbz999:
Technically this fixes a bug since the current implementation prevents applying effects from a consumed scroll.

Issue #1280, opened 2021-09-05
Add a spell preparation mode for ritual only spells


aaclayton:
Originally in GitLab by @Pjb518

Certain characters have spells that can only be cast as rituals. Two examples are Warlocks with the Book of Ancient Secrets Invocation and characters with the Ritual Caster feat.

For ease of organisation, I suggest we add a "Ritual Book" or "Ritual Only" preparation mode so that these spells can be kept separate from any existing spellcasting sources the character currently possesses.

JPMeehan:
One comment here: It would be nice if this could also handle the Wizard's spellbook, as that allows you to cast unprepared spells IFF they are rituals. It's possible the current features are sufficient.

etiquettestartshere:
> @etiquettestartshere Can you comment on the original issue for this so I can assign it to you? (#1280)

#:wave:

PR #3173, opened 2024-03-02
[#1280] Add Ritual Spell Preparation Mode

etiquettestartshere:
From my original PR:
> As per [#1280], add a Spell Preparation Mode for spells intended to only be cast as rituals and not as leveled spells, such spells obtained through the Warlock's Eldritch Invocation: Book of Ancient Secrets, the Ritual Spellcaster Feat, and the Totem Warrior Barbarian's Spirit Seeker and Spirit Walker.

I had a little trouble rebasing this (just because it was my first time and I barely know how to use git, not because it was complicated), so please let me know if there are any issues.

I believe the changes I made before resulted in ritual being in between atwill and innate, but now because of the new method for sorting sections, ritual ends up on the top. I do not have an opinion about this, it just seemed to be the least number of cuts to add ritual at order: -30 rather than changing the others. I assume order: -5 would also work (or reordering the whole object).

edit: The more I think about it the more I think ritual should appear at the bottom of this list rather than above At-Will, but I would appreciate some feedback.
image


Fyorl:
> From my original PR:
> 
> > As per [#1280], add a Spell Preparation Mode for spells intended to only be cast as rituals and not as leveled spells, such spells obtained through the Warlock's Eldritch Invocation: Book of Ancient Secrets, the Ritual Spellcaster Feat, and the Totem Warrior Barbarian's Spirit Seeker and Spirit Walker.

I think I have a slight preference for 'Ritual Book' from the original issue, rather than 'Ritual Spellcasting'. I think the latter runs the risk of confusing users into thinking they have to use this mode for all their ritual spells, rather than it being a special 'ritual only' mode.
> 
> I believe the changes I made before resulted in ritual being in between atwill and innate, but now because of the new method for sorting sections, ritual ends up on the top. I do not have an opinion about this, it just seemed to be the least number of cuts to add ritual at order: -30 rather than changing the others. I assume order: -5 would also work (or reordering the whole object).

No strong opinion from me at least where this category appears in the spells tab. I imagine everyone will have their own individual preferences so we should just pick one.

> edit: The more I think about it the more I think ritual should appear at the bottom of this list rather than above At-Will, but I would appreciate some feedback. image

No strong opinion, at the bottom is fine.

I don't see it in this PR but we should ensure that casting a spell from the 'ritual only' section doesn't prompt to consume a slot, the same way that At-Will and Innate Spellcasting don't.

krbz999:
> I don't see it in this PR but we should ensure that casting a spell from the 'ritual only' section doesn't prompt to consume a slot, the same way that At-Will and Innate Spellcasting don't.

`Item5e#usageScaling` combined with the lack of `upcast: true` in the config object takes care of this.

Fyorl (approved):

arbron:
@etiquettestartshere Can you comment on the original issue for this so I can assign it to you? (https://github.com/foundryvtt/dnd5e/issues/1280)

PR #3190, opened 2024-03-05
[#2920, #3147] Add magical bonus to ammo, fix attack bonus data

### LINKED ISSUE #2920: Formula data in attack bonuses do not work on ammunition
I was fiddling around trying to setup a custom ammunition in my campaign, trying to hack my way around making the "Magic Stone" cantrip work as-is without having to fiddle with the ammo or duplicate my player's weapon 😄 So I used the following process to make it as simple as possible:
`@mod - @abilities.dex.mod`
But then I was greeted with the following error: 
image
So, I thought that maybe using `@mod` could be the culprit for some reason, so I tried expanding it into `@abilities.wis.mod`
image
No luck yet... I made a bunch more tests, including *only* adding a single attribute in the attack bonus to check if making calculations was an issue in this specific scenario, but it was still refusing to work :(

I checked to see if the "summary" bulletpoints on the left were perhaps glitching (which would have given me a clue), but... the equation was actually calculating the bonus correctly!
image
So my conclusion was, "If the computation on the item's card is correct, it means the equation is supposedly parsed as it should, so could it be a bug from one of the latest releases?"

I made one last test, which is to put the same equation onto a weapon, NOT onto a piece of ammunition. And it worked!

So if my testing is correct, interpolation equations do not work on attack rolls for ammunition exclusively - it still works on weapons though

EDIT: I forgot to mention, but of course the bug was replicated on a fresh table with no module, Foundry v11 Build 315 and DnD5 3.0.1

### LINKED ISSUE #3147: Add magical bonus to weapons & armor

arbron:
Calls `Roll.replaceFormulaData` on the ammo attack bonus before it is added to the weapon's to hit formula to avoid errors with string terms not resolving.

Adds `magicalBonus` to ammunition that acts like that on weapons. This adds to the attack bonus for the weapon when it is being used with the ammunition. The damage bonus is appended to the first damage formula on the weapon, and the damage properties gain any physical properties provided by the ammunition so silvered or magical ammo properly carries through to the weapon damage.

This also changes how additional ammo damage is added to make use of the new system for multiple rolls on damage. Ammunition damage is now constructed as separate rolls with their own type data rather than being appended to the base weapon damage formula.

Closes #2920 
Closes #3147

### DIFF
--- a/module/data/item/consumable.mjs
+++ b/module/data/item/consumable.mjs
@@ -9,7 +9,7 @@ import ItemTypeTemplate from "./templates/item-type.mjs";
 import PhysicalItemTemplate from "./templates/physical-item.mjs";
 import ItemTypeField from "./fields/item-type-field.mjs";
 
-const { BooleanField, SetField, StringField } = foundry.data.fields;
+const { BooleanField, NumberField, SetField, StringField } = foundry.data.fields;
 
 /**
  * Data definition for Consumable items.
@@ -21,6 +21,7 @@ const { BooleanField, SetField, StringField } = foundry.data.fields;
  * @mixes ActivatedEffectTemplate
  * @mixes ActionTemplate
  *
+ * @property {number} magicalBonus       Magical bonus added to attack & damage rolls by ammunition.
  * @property {Set<string>} properties    Ammunition properties.
  * @property {object} uses
  * @property {boolean} uses.autoDestroy  Should this item be destroyed when it runs out of uses.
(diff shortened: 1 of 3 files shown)


Fyorl (approved):
Nice work, thank you

PR #3196, opened 2024-03-06
Deprecate `DND5E.LimitedUseFormulaPeriods`

### LINKED ISSUE #3264: Deprecate 'LimitedUseFormulaPeriods' in the config
Would help to consolidate this into 'LimitedUsePeriods' for easier configuration as well.

krbz999:
Closes #3264


Fyorl (approved):
Can we have an associated issue (if there isn't already one) for the patch notes.

PR #3197, opened 2024-03-06
Allow for options in `Item5e#rollRecharge`

krbz999:
I saw someone ask how to roll recharge without displaying a chat message. This allows for that, and further configuration.


arbron:
Rather than adding specific options for `rollRecharge`, it would probably be better to use the changes in the roll refactor so it is consistent across roll types (https://github.com/foundryvtt/dnd5e/pull/3146). That way we wouldn't have to worry about migrating these later.

PR #3201, opened 2024-03-06
[#3200] Fix available uses displaying as `null` in ability use dialog

### LINKED ISSUE #3200: Empty values in the `Available Uses` field currently show as `null`
Example Appearance:
image

krbz999:
Closes #3200.


arbron:
@krbz999 Can you comment on the original issue so I can assign it to you?

arbron (approved):

Issue #2935, opened 2024-02-06
AE duration is not only changed on Round change on the AE on the Actor but also on the Spell Item


itteerde:
3.0.1 see comments, because the 3.0 change to AE transferrals the SRD Bless (guess more?) Spell Item needs to be changed.

AE duration is not only changed on Round change on the AE on the Actor but also on the Spell Item

All modules deactivated, dropped Bless from SRD Compendium to be sure to not have corrupted data.

etiquettestartshere:
As far as I can tell, this means that the timers for effects are useless once an effect expires once. I could find no way to reset the timer aside from getting a new copy of the spell. I believe in 2.4 that this could be accomplished by deleting the effect from the actor and dragging it back from the spell, but now it seems when I delete the bless effect from the actor, it is removed from the spell also. Would love to be proven wrong 😅.

krbz999:
> but now it seems when I delete the bless effect from the actor, it is removed from the spell also. Would love to be proven wrong 😅.

If an effect is set to transfer, then in v3 you will see it on the actor despite it being on the item. Hence what you were deleting was likely the effect on the item.

itteerde:
> If an effect is set to transfer, then in v3 you will see it on the actor despite it being on the item. Hence what you were deleting was likely the effect on the item. 

Yes. But that still leads to what he wrote. Currently as the duration on the Item goes down, not just on the AE on the Actor, one would have to drop a new Spell Item for every casting or reset the duration on the Spell Item.

etiquettestartshere:
> If an effect is set to transfer, then in v3 you will see it on the actor despite it being on the item. Hence what you were deleting was likely the effect on the item.

Ah, okay, I can confirm checking Effect Suspended and unchecking Apply Effect to Actor works much better: I can apply the effect from the chat card, it ticks down, then I can delete it and reapply from the chat card. However this is only possible if I modify Bless. Maybe SRD items that should be applied in this way should have Apply Effect to Actor unticked by default.

itteerde:
ah, true. So this is not a code bug, but compendium

Fyorl:
[...]

PR #3202, opened 2024-03-06
[#2935] Prevent effect migrations from running more than once

arbron:
Should do the job to gate those active effect migrations, but I haven't tested it on a world that requires those migrations yet.

### DIFF
--- a/module/migration.mjs
+++ b/module/migration.mjs
@@ -498,7 +498,7 @@ export const migrateEffectData = function(effect, migrationData, { parent }={}) 
   const updateData = {};
   _migrateDocumentIcon(effect, updateData, {...migrationData, field: "icon"});
   _migrateEffectArmorClass(effect, updateData);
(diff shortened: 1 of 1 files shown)


module/migration.mjs
```diff
@@ -496,7 +498,9 @@ export const migrateEffectData = function(effect, migrationData, { parent }={})
   const updateData = {};
   _migrateDocumentIcon(effect, updateData, {...migrationData, field: "icon"});
   _migrateEffectArmorClass(effect, updateData);
-  _migrateTransferEffect(effect, parent, updateData);
+  if ( foundry.utils.isNewerVersion("3.1.0", effect._stats?.systemVersion) ) {
```
Fyorl, line 501: It's unfortunate, but I don't believe ActiveEffects have `_stats` in v11. We have the `parent` here, so maybe that's good enough?

Fyorl (changes requested):
Thanks for taking care of this, Jeff. Just check my one comment and then I think we can merge it and do some testing.

Fyorl (approved):

Issue #1871, opened 2022-10-12
Advancement Type: Starting Equipment


arbron:
The equipment advancement type would be responsible for presenting a list of equipment that a player can choose or gets automatically from their class or background.

## Level Up Interface
The advancement is not responsible for actually adding the equipment, only holding data on what equipment should be added. The adding of the equipment will be handled by a separate character creation workflow.

~~The player would be presented with a list of choices for each group with options. For A/B options, there would be checkboxes for which the player wants to choose. Within each option additional choices like "A simple weapon" would either have a dropdown or a drop area for selecting a specific weapon type.~~

## Configuration
The `pool` property contains a list of options of various types to build up the equipment list, and the `wealth` property contains a formula used to determine starting wealth if the equipment isn't taken.

```javascript
{
  "pool": [
    {
      "_id": "OJNDsndRmYsq7Eph",
      "sort": 0,
      "type": "OR"
    },
    {
      "_id": "OJFFk7CEeqIdvO8e",
      "sort": 10000,
      "type": "OR"
    },
    {
      "_id": "XUvtgs0KtyowcJ6N",
      "sort": 20000,
      "type": "AND"
    },
    {
      "_id": "c7lqmQA2HFFjKTyl",
      "group": "OJNDsndRmYsq7Eph",
      "sort": 30000,
      "type": "linked",
      "key": "Compendium.dnd5e.items.Item.1Lxk6kmoRhG8qQ0u"
    },
    {
      "_id": "f2UnHjj5S6hjWUdb",
      "group": "OJNDsndRmYsq7Eph",
      "sort": 40000,
      "type": "weapon",
      "key": "martialM"
    },
    {
      "_id": "hgYMehnwxzzdSU98",
      "group": "OJFFk7CEeqIdvO8e",
      "sort": 50000,
      "type": "linked",
      "count": 2,
      "key": "Compendium.dnd5e.items.Item.eO7Fbv5WBk5zvGOc"
    },
    {
      "_id": "aphK30SQ2eCetYdd",
      "group": "OJFFk7CEeqIdvO8e",
      "sort": 60000,
      "type": "weapon",
      "key": "sim"
    },
    {
      "_id": "rQAu5VtTtJQia35h",
      "group": "XUvtgs0KtyowcJ6N",
      "sort": 70000,
      "type": "linked",
      "key": "Compendium.dnd5e.items.Item.8KWz5DJbWUpNWniP"
    },
    {
[...]

PR #3203, opened 2024-03-06
[#1871] Implement configuration UI for starting equipment

arbron:
Screenshot 2024-03-06 at 14 21 49

### DIFF
--- a/module/applications/item/starting-equipment-config.mjs
+++ b/module/applications/item/starting-equipment-config.mjs
@@ -1,3 +1,6 @@
+import PhysicalItemTemplate from "../../data/item/templates/physical-item.mjs";
+import { EquipmentEntryData } from "../../data/item/templates/starting-equipment.mjs";
+
 /**
  * Configuration application for Starting Equipment.
  */
@@ -7,9 +10,9 @@ export default class StartingEquipmentConfig extends DocumentSheet {
   static get defaultOptions() {
     return foundry.utils.mergeObject(super.defaultOptions, {
       classes: ["dnd5e", "starting-equipment"],
-      dragDrop: [{ dropSelector: "form" }],
+      dragDrop: [{ dragSelector: ".drag-bar", dropSelector: "form" }],
       template: "systems/dnd5e/templates/apps/starting-equipment-config.hbs",
-      width: 400,
+      width: 480,
       height: "auto",
       sheetConfig: false,
       closeOnSubmit: false,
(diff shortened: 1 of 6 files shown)


module/applications/item/starting-equipment-config.mjs
```diff
@@ -18,23 +22,268 @@ export default class StartingEquipmentConfig extends DocumentSheet {
...
+
+    // If drag entry & drop entry are the same, or drop entry is drag entry's group, do nothing
+    if ( (dropEntry?._id === dragEntry._id) || (dropEntry?._id === dragEntry.group) ) return;
+
+    let updateData;
+    let sortBefore;
+    let target;
+
+    // If drop entry is a group, move drag entry into it
+    if ( dropEntry?.type in EquipmentEntryData.GROUPING_TYPES ) {
```
Fyorl, line 252: We need to check for depth here, I was able to exceed the maximum depth when I was testing by dropping a group into another group.

module/applications/item/starting-equipment-config.mjs
```diff
@@ -18,23 +22,268 @@ export default class StartingEquipmentConfig extends DocumentSheet {
...
+        getDeleteIds(this.document.system.startingEquipment.find(i => i._id === data.entryId));
+        data.startingEquipment = data.startingEquipment.filter(e => !deleteIds.has(e._id));
+        break;
+      case "toggle-proficiency":
+        const entry = data.startingEquipment.find(e => e._id === data.entryId);
+        if ( entry ) entry.requiresProficiency = !entry.requiresProficiency;
+        break;
+    }
+
+    return { "system.startingEquipment": data.startingEquipment, "system.wealth": data.wealth };
```
Fyorl, line 166: Doesn't seem like we configure wealth in this dialog anymore.

Fyorl (approved):
This sort of nested AND/OR grouping UI/UX can be quite complicated, but this looks really great. I have only really some high level comments, but they are suggestions only.

Design:
1. I think the buttons looks pretty reasonable when not 'raised', and maybe with the v2 styles. When they are on the same level as the group they are adding to, I think it's still obvious enough that they are adding an entry to that group, and what the result will be.

image

Code:
1. I think I understand why we are hiding the proficiency flag behind the context menu, because it's so vanishingly rare (only Cleric uses it, right?), but I think we have enough room and it looks ok on the main configuration. If we moved it there, then I think there wouldn't be much reason to need a context menu since adding and deleting entries are quite prominent and discoverable already. So we could potentially eliminate the context menu entirely and save ourselves some complexity.
2. The drag-drop functionality and sorting is another source of potential bugs and complexity that I'm not sure how much demand there would be for this feature. You would probably know better than me having setup all of the SRD starting equipment, but I think it's worth considering.
image

arbron:
> 1. The drag-drop functionality and sorting is a source of potential bugs and complexity that I'm not sure how much demand there would be for this feature. You would probably know better than me having setup all of the SRD starting equipment, but I think it's worth considering.

It definitely came in handy doing the SRD stuff, since the equipment needs to be in such a particular order. Nothing more annoying then doing a whole class and realizing you made a mistake earlier and have to delete and redo everything just to get the order right.

PR #3208, opened 2024-03-07
[#2094, #3185] Move ability & hp prep, fix advancement HP on NPCs

### LINKED ISSUE #3185: Modify `HitPointsAdvancement` to allow changing NPC HP max
Allow for `HitPointsAdvancement` to increase the max hit points on an NPC.

arbron:
Moves most of the logic from `Actor5e._prepareHitPoints` into a new `AttributesFields.prepareHitPoints` method with changes to allow different types to customize how their hit points are prepared.

In order to make that change it also required the moving of `Actor5e._prepareAbilities` to `CommonTemplate.prepareAbilities` so that ability modifiers could still be calculated before the hit points.

`NPCData` now fetches `HitPointsAdvancement`s from its classes and adds those to the hit points it calculates.

Closes #3185

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -1,4 +1,5 @@
 import Proficiency from "../../documents/actor/proficiency.mjs";
+import { simplifyBonus } from "../../utils.mjs";
 import { FormulaField, LocalDocumentField } from "../fields.mjs";
 import CreatureTypeField from "../shared/creature-type-field.mjs";
 import AttributesFields from "./templates/attributes.mjs";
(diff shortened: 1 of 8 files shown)


Fyorl (approved):
Code looks good, I did have one question about this line in #3185:

> This should also allow for that advancement to change max hit points on PCs that have their maximum overridden.

Should that say NPCs? For PCs it seems that we still have HP override ignore advancements, which I think is desirable, because there would be no way to specify a flat HP override otherwise. For NPCs, this becomes a problem, but I think what we have with this PR is a good iteration regardless.

arbron:
> Code looks good, I did have one question about this line in #3185:
> 
> > This should also allow for that advancement to change max hit points on PCs that have their maximum overridden.
> 
> Should that say NPCs? For PCs it seems that we still have HP override ignore advancements, which I think is desirable, because there would be no way to specify a flat HP override otherwise. For NPCs, this becomes a problem, but I think what we have with this PR is a good iteration regardless.

Aw yeah, that issue was originally written when I was going to take a different approach which would have also applied to PCs with the max override. That is no longer the case and this one shouldn't affect PCs at all.

PR #3209, opened 2024-03-07
[#2668] Derive NPC details from race item if present

### LINKED ISSUE #2668: Feature Request: Add Race to NPCs
The various books from DnD contain many humanoids that are "any race". Having the ability to add a race to them would simplify creating unique variants that contain the racial modifiers, such as darkvision and resistances.

arbron:
Introduces a new `AbilitiesFields#prepareRace` method that takes a race item and applies its stats to the actor, either complementing the default values or replacing them depending on whether the `force` option is set.

This then replaces the existing code in `CharacterData` that used to apply race data and is added to `NPCData`.

Closes #2668

### DIFF
--- a/module/data/actor/npc.mjs
+++ b/module/data/actor/npc.mjs
@@ -252,10 +252,9 @@ export default class NPCData extends CreatureTemplate {
    * Prepare movement & senses values derived from race item.
    */
   prepareEmbeddedData() {
-    const race = this.parent.items.find(i => i.type === "race");
-    if ( race ) {
-      AttributesFields.prepareRace.call(this, race, { force: true });
-      this.details.type = race.system.type;
+    if ( this.details.race instanceof Item ) {
+      AttributesFields.prepareRace.call(this, this.details.race, { force: true });
+      this.details.type = this.details.race.system.type;
     }
   }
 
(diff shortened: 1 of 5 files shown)


module/data/actor/npc.mjs
```diff
@@ -248,6 +248,19 @@ export default class NPCData extends CreatureTemplate {
...
   /* -------------------------------------------- */
 
+  /**
+   * Prepare movement & senses values derived from race item.
+   */
+  prepareEmbeddedData() {
+    const race = this.parent.items.find(i => i.type === "race");
+    if ( race ) {
+      AttributesFields.prepareRace.call(this, race, { force: true });
+      this.details.type = race.system.type;
```
Fyorl, line 258: Seems like this could also be done in `prepareRace`, unless I am missing something?

Fyorl (approved):
Just had one question, but looks good to go

module/data/actor/npc.mjs
arbron: Mainly a pedantic reason because `prepareRace` is a method on `AttributesFields`, but the type is within `details` not `attributes` like the rest of the changed values.

PR #3210, opened 2024-03-07
Add concentration button on character sheet, show concentration effects

krbz999:
... ... ... ...

image


Zanderaf:
Would it be any clearer if this was maybe in the spellbook tab instead? I get the feeling that most things that need concentration are gonna be in the spell book / reminding people that they are concentrating on something while searching through their spellbook

PR #3211, opened 2024-03-08
[#3035] Remove various dependencies on "dhp" and "dtemp" properties and add damage/healing hooks

### LINKED ISSUE #3035: Provide damage and healing hooks for when an actor is updated
Related to #3028.

The hooks that currently best serve to track when and if an actor receives damage (provided by the system) are found in the `applyDamage` method. However this does not account for the times a user may simply apply damage (or healing) directly through the character sheet, or via a direct update.

It would be good to track the differences in current hit points and call an appropriate hook, which may likely be required regardless to determine the DC of a saving throw to maintain concentration.

krbz999:
Closes #3035.

This adds the 'dnd5e.healActor' and 'dnd5e.damageActor' hooks (I don't feel strongly that pre-hooks are necessary).

An additional benefit is that this removes the odd dependencies on a "dhp" or "dtemp" property being passed along in updates to get scrolling damage/healing numbers; the event listeners simply take care of it, and the scrolling damage numbers will also display when damage is applied through the sheet, not just via the chatlog or other hardcoded methods.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -3228,8 +3228,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       };
       changes.total = changes.hp + changes.temp;
 
-
-      if ( Number.isInteger(changes.total) && parseInt(changes.total) !== 0 ) {
+      if ( Number.isInteger(changes.total) && (changes.total !== 0) ) {
         this._displayTokenEffect(changes);
 
         /**
@@ -3237,10 +3236,10 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
          * of the hook will depend on the change in hit points.
          * @function dnd5e.damageActor
          * @memberof hookEvents
-         * @param {Actor5e} actor     The actor that had their hit points reduced.
-         * @param {number} changes    The changes to hit points.
-         * @param {object} update     The triggering updated performed.
-         * @param {string} userId     Id of the user that performed the update.
+         * @param {Actor5e} actor                                       The actor that had their hit points reduced.
+         * @param {{hp: number, temp: number, total: number}} changes   The changes to hit points.
+         * @param {object} update                                       The original update delta.
+         * @param {string} userId                                       Id of the user that performed the update.
          */
         Hooks.callAll(`dnd5e.${changes.total > 0 ? "heal" : "damage"}Actor`, this, changes, data, userId);
       }
@@ -3287,7 +3286,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
   _displayTokenEffect(changes) {
     let key;
     let change;
-    if ( !parseInt(changes.hp) && parseInt(changes.temp) ) key = "temp";
+    if ( !changes.hp && changes.temp ) key = "temp";
     else {
       key = "hp";
       change = changes.total;
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -3285,26 +3285,30 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
    */
   _displayTokenEffect(changes) {
     let key;
-    let change;
-    if ( !changes.hp && changes.temp ) key = "temp";
-    else {
-      key = "hp";
-      change = changes.total;
+    let value;
+    if ( changes.hp < 0 ) {
+      key = "damage";
+      value = changes.total;
+    } else if ( changes.hp > 0 ) {
+      key = "healing";
+      value = changes.total;
+    } else if ( changes.temp ) {
+      key = "temp";
+      value = changes.temp;
     }
-    if ( !key ) return;
-    change ??= changes[key];
+    if ( !key || !value ) return;
 
     const tokens = this.isToken ? [this.token] : this.getActiveTokens(true, true);
     if ( !tokens.length ) return;
 
-    const pct = Math.clamped(Math.abs(change) / this.system.attributes.hp.max, 0, 1);
-    const fill = CONFIG.DND5E.tokenHPColors[(key === "hp") ? (change < 0 ? "damage" : "healing") : "temp"];
+    const pct = Math.clamped(Math.abs(value) / this.system.attributes.hp.max, 0, 1);
+    const fill = CONFIG.DND5E.tokenHPColors[key];
 
     for ( const token of tokens ) {
-      if ( !token.object.visible || !token.object.renderable ) continue;
-      token.flashRing({ [key]: change });
+      if ( !token.object?.visible || !token.object?.renderable ) continue;
+      token.flashRing(key);
       const t = token.object;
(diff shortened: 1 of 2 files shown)


module/documents/actor/actor.mjs
```diff
@@ -3254,28 +3278,42 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+  _displayTokenEffect(changes) {
+    let key;
+    let change;
+    if ( !parseInt(changes.hp) && parseInt(changes.temp) ) key = "temp";
+    else {
+      key = "hp";
+      change = changes.total;
+    }
+    if ( !key ) return;
+    change ??= changes[key];
```
krbz999, line 3296: If the change is only to temp hp (up or down), then only that should be displayed. If there is any change at all to hp.value, then show healing or damage.

module/documents/actor/actor.mjs
```diff
@@ -1007,7 +1008,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       isBar: true
     }, updates) === false ) return this;
```
krbz999, line 1009: Relatedly, I'm not sure why we have this hook call. Far as I can tell, it does nothing at all.

arbron: As far as I understand it, that is so that anyone monitoring for changes to attributes on the token bar will see changes to HP made elsewhere, but it has been around forever so I don't really know `¯\_(ツ)_/¯`

module/documents/actor/actor.mjs
```diff
@@ -3216,11 +3214,37 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    if ( hp && (!options.isRest && !options.isAdvancement) ) {
+      const curr = this.system.attributes.hp;
+      const changes = {
+        hp: curr.value - hp.value,
+        temp: curr.temp - hp.temp
+      };
+      changes.total = changes.hp + changes.temp;
+
+
+      if ( Number.isInteger(changes.total) && parseInt(changes.total) !== 0 ) {
```
Fyorl, line 3232: What is the `parseInt` in aid of?

```suggestion
      if ( Number.isInteger(changes.total) && (changes.total !== 0) ) {
```

module/documents/actor/actor.mjs
```diff
@@ -3216,11 +3214,37 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+        this._displayTokenEffect(changes);
+
+        /**
+         * A hook event that fires when an actor is damaged or healed by any means. The actual name
+         * of the hook will depend on the change in hit points.
+         * @function dnd5e.damageActor
+         * @memberof hookEvents
+         * @param {Actor5e} actor     The actor that had their hit points reduced.
+         * @param {number} changes    The changes to hit points.
```
Fyorl, line 3241:
```suggestion
         * @param {{hp: number, temp: number, total: number}} changes    The changes to hit points.
```

module/documents/actor/actor.mjs
```diff
@@ -3216,11 +3214,37 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+         * @param {object} update     The triggering updated performed.
```
Fyorl, line 3242:
```suggestion
         * @param {object} update     The original update delta.
```

module/documents/actor/actor.mjs
```diff
@@ -3254,28 +3278,42 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+   * @param {number} changes.temp     The change to `hp.temp`.
+   * @param {number} changes.total    The total change to hit points.
    * @protected
    */
-  _displayTokenEffect(options) {
-    const dhp = Number(options.dhp);
+  _displayTokenEffect(changes) {
+    let key;
+    let change;
+    if ( !parseInt(changes.hp) && parseInt(changes.temp) ) key = "temp";
```
Fyorl, line 3290: Still unclear what the `parseInt` is for:

```suggestion
    if ( !changes.hp && changes.temp ) key = "temp";
```

module/documents/actor/actor.mjs
```diff
@@ -3254,28 +3278,42 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-          anchor: CONST.TEXT_ANCHOR_POINTS.TOP,
-          // Adapt the font size relative to the Actor's HP total to emphasize more significant blows
-          fontSize: 16 + (32 * pct), // Range between [16, 48]
-          fill: CONFIG.DND5E.tokenHPColors[dhp < 0 ? "damage" : "healing"],
-          stroke: 0x000000,
-          strokeThickness: 4,
-          jitter: 0.25
-        });
-      }
+      token.flashRing({ [key]: change });
```
Fyorl, line 3306: It seems `flashRing` still expects `dhp` or `dtemp` arguments.

Fyorl (changes requested):
Thanks for doing this. It should mean we are able to prompt for concentration saves now when receiving damage in any way.

We could consider continuing to send the `dhp` and `dtemp` options in order to make the change no longer breaking. It's unclear to me how disruptive it would be to lose these.

module/documents/actor/actor.mjs
```diff
@@ -1007,7 +1008,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       isBar: true
     }, updates) === false ) return this;
 
-    await this.update(updates, {dhp: -amount});
+    await this.update(updates, { dnd5e: { damages } });
```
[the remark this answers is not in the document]
krbz999: It is not. I included it to provide some context in the `updateActor` hook. It's not for the system itself though, and I don't feel too strongly. Might use it later if we feel like reviving the colored damage numbers PR I had.

module/documents/actor/actor.mjs
krbz999: I considered it less breaking to leave the arguments for this method as is, and only getting rid of the update options. I suspect those would have been much less used (they were never quite reliable if one wanted to use those to check for hp changes).

One thing we could do is change it to a single argument here that is just the string "temp", "damage", or "healing", but lemme know your thoughts on that.

module/documents/actor/actor.mjs
krbz999: Let's leave it since we're all emotionally attached to it.

module/documents/actor/actor.mjs
Fyorl: I'm sorry, I'm not following. What update options? The issue was that this line, as-written, was doing nothing because it was passing the incorrect arguments to `flashRing`.

krbz999: Oh I see what you mean now, yeah that must have been an oversight. Regardless perhaps the change I made makes the function's use a bit clearer?

Fyorl (approved):

PR #3217, opened 2024-03-08
[#893] Add summoning actor updates & overall workflow

arbron:
Moves the summoning data into an `EmbeddedDataField` to allow for the summoning methods to be defined there rather than taking up a bunch of space in `ActionTemplate`.

Introduces `SummonsData#summon` method which takes the ID of a summoning profile and proceeds through the process of summoning the actor:

1) Fetches the actor to summon based on the profile. This will eventually need some extra work to handle summoning from compendiums.

2) Minimizes the actor sheet (if open) and determine placement on scene. Currently the placement is a placeholder method that just places the summons in a fixed position.

3) Iterates over all token placements, requesting updates that will be made to the actor and its contained items for each placement.
 - Match proficiency is applied using an active effect
 - AC bonus is applied to `ac.flat` value if calc is `flat`, otherwise it uses an active effect to apply to `ac.bonus`
 - HP bonus is applied to `hp.max` unless summoned creature is a PC without an overriden maximum HP, in which case it uses an effect
 - Match attacks calculates the to hit based on the ability specified on the summoning item. It fetches attack bonuses from the summoner, but it uses spell bonuses in place of weapon bonuses
 - Match save simply swaps to flat scaling mode with the DC from the item or the summoner's spell DC.

4) Synthesizes token data with the updates.

5) Once all tokens are ready, maximizes the sheet if it was previously minimized and creates token documents on the scene.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1424,6 +1424,9 @@
       "Hint": "Modify to saving throw DCs on the summoned creature's abilities to match that of the summoner."
     }
   },
+  "Placement": {
+    "GenericError": "Failed to place summoned actor"
+  },
   "Profile": {
     "Label": "Summons Profile",
     "LabelPl": "Summons Profiles",
--- a/module/applications/item/summoning-config.mjs
+++ b/module/applications/item/summoning-config.mjs
@@ -27,7 +27,7 @@ export default class SummoningConfig extends DocumentSheet {
    * @type {object[]}
    */
   get profiles() {
-    return this.document.system.summons.profiles;
+    return this.document.system.summons?.profiles ?? [];
   }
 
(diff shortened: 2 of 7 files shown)


krbz999:
> This will eventually need some extra work to handle summoning from compendiums.

Let's not.

arbron:
> > This will eventually need some extra work to handle summoning from compendiums.
> 
> Let's not.

We gotta, otherwise there is no way to deliver these pre-configured. Players that add a spell to their sheet shouldn't then have to go into a specific compendium, import the actors, and then drag them onto their spells before they can actually use them.

Fyorl (approved):
Very nice, thank you Jeff

PR #3219, opened 2024-03-08
[#893] Improve permissions checks & allow compendium summoning

arbron:
Adds a `SummonsData#canSummon` property that determines whether a scene is open and whether the user has permissions to summon. If not, then no summoning control is displayed in the `AbilityUseDialog` and the chat button is removed.

If user is trying to summon and actor that they do not have ownership of the will be shown an error.

Also introduces a check for file browser permission when creating token data if the token is set to use a wildcard image. If the user lacks permission here, then the token will fall back to using the portrait image and a warning will be displayed in the console.

Adds a new `SummonsData#fetchActor` data which handles locating the actor to be summoned. If the actor is in the world, then it just returns that version. If the associated actor is in a compendium, it will first search through world actors to find one with a matching `sourceId` and the `dnd5e.summonedCopy` flag set, summoning from the pre-imported version if available. Only then will it fall back on importing the actor from the compendium so long as the user has permission to do so.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1491,9 +1491,9 @@
     "Hint": "Disable the summoning prompt when item is used. Players will still be able to summon from the chat card."
   },
   "Warning": {
-    "CreateActor": "You must have 'Create New Actors' permission in order to summon directly a compendium.",
+    "CreateActor": "You must have 'Create New Actors' permission in order to summon directly from a compendium.",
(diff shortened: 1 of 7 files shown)


Fyorl (changes requested):
Thanks very much, Jeff

Fyorl (approved):

PR #3225, opened 2024-03-09
#3098 Removed armor check for Stealth Disadvantage

### LINKED ISSUE #3098: 'Stealth disadvantage' shows as a tag on equipment if ticked and unavailable
- make equipment item on an actor.
- make it armor / shield.
- tick 'Stealth Disadvantage'.
- see tooltip in inventory has tag for Stealth Disadvantage.
- make it clothing.
- notice that Stealth Disadv is no longer a valid option (should it be?).
- see tooltip still has tag for Stealth Disadvantage regardless.

Would simplify things if this was just a property for any 'equipment' regardless of being armor or shield. It's easy to imagine a trinket or clothing that grants disadvantage on stealth.

Hoppyhob:
Removed Armor check from validProperties for stealthDisadvantage for equipment.
This allows all equipment to be marked with stealthDisadvantage.
Closes #3098

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -1343,7 +1343,8 @@ DND5E.validProperties = {
   ]),
   equipment: new Set([
     "concentration",
-    "mgc"
+    "mgc",
+    "stealthDisadvantage"
   ]),
   feat: new Set([
     "concentration",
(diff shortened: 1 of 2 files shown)


Fyorl (changes requested):
If we're adding this property to all equipment, not just armour, then we don't need the `EquipmentData#validProperties` override at all, it can just be deleted, and we can add the property directly to `CONFIG.DND5E.validProperties.equipment`.

Fyorl (approved):

PR #3226, opened 2024-03-09
[#2100] Implement HitDice helper class

### LINKED ISSUE #1074: Auto select next highest hit dice denomination when rolling hit dice from Short Rest window
Currently, when rolling hit dice on a short rest, if you run out of a large denomination of hit dice and you click "Roll" again it will present an error saying your have run out of hit dice of that denomination. A more intuitive behavior would be to automatically select the next-highest hit dice denomination in the dropdown when the current denomination is exhausted so players can continue rolling if needed.

### LINKED ISSUE #2100: Allow largest and smallest available hit dice to be used in damage formulas

krbz999:
Redo of #2519.

Closes #2100.
Closes #1074.

Saves people from https://github.com/foundryvtt/dnd5e/issues/2100#issuecomment-1986962845.

This moves most hit dice related logic into a single class, which simplifies a lot of code throughout.

### DIFF
--- a/module/documents/actor/HitDice.mjs
+++ b/module/documents/actor/HitDice.mjs
@@ -1,49 +1,63 @@
 /**
  * Object describing the hit dice for an actor.
- *
- * @param {Actor5e} actor
  */
 export default class HitDice {
+  /**
+   * Object describing the hit dice for an actor.
+   * @param {Actor5e} actor     The actor whose hit dice this document describes.
+   */
   constructor(actor) {
-    /**
-     * Store a reference to the actor.
-     * @type {Actor5e}
-     */
     this.actor = actor;
 
-    /**
-     * Remaining hit dice.
-     * @type {number}
-     */
-    this.remaining = 0;
-
-    /**
-     * The actor's total amount of hit dice.
-     * @type {number}
-     */
-    this.total = 0;
-
-    /**
-     * All valid die sizes derived from all classes.
-     * @type {Set<number>}
-     */
-    this.sizes = new Set();
-
-    /**
-     * Store valid class items.
-     * @type {Set<Item5e>}
-     */
-    this._classes = new Set();
-
-    for (const item of Object.values(this.actor.classes)) {
-      if ( /^d[0-9]+$/.test(item.system.hitDice) ) {
-        this._classes.add(item);
-        this.remaining += item.system.levels - item.system.hitDiceUsed;
-        this.total += item.system.levels;
-        this.sizes.add(parseInt(item.system.hitDice.replace("d", "")));
+    for ( const item of Object.values(actor.classes) ) {
+      if ( /^d\d+$/.test(item.system.hitDice) ) {
+        this.classes.add(item);
+        this.value += item.system.levels - item.system.hitDiceUsed;
+        this.max += item.system.levels;
+        this.sizes.add(parseInt(item.system.hitDice.slice(1)));
       }
     }
   }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Store a reference to the actor.
+   * @type {Actor5e}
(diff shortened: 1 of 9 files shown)


revilowaldow:
Yes please!

module/documents/actor/HitDice.mjs
```diff
@@ -0,0 +1,172 @@
...
+     * All valid die sizes derived from all classes.
+     * @type {Set<number>}
+     */
+    this.sizes = new Set();
+
+    /**
+     * Store valid class items.
+     * @type {Set<Item5e>}
+     */
+    this._classes = new Set();
```
Fyorl, line 36: I suppose this is marked as an underscore to indicate that it shouldn't be mutated externally? There are ways we can guarantee that, but they're all a bit slow. I think we should remove the underscore since this seems to be used publicly, and reserve the underscore for `@protected` or `@internal` fields only.

module/documents/actor/HitDice.mjs
```diff
@@ -0,0 +1,172 @@
...
+
+    for (const item of Object.values(this.actor.classes)) {
```
Fyorl, line 38:
```suggestion
    for ( const item of Object.values(actor.classes) ) {
```

module/documents/actor/HitDice.mjs
```diff
@@ -0,0 +1,172 @@
...
+      if ( /^d[0-9]+$/.test(item.system.hitDice) ) {
```
Fyorl, line 39:
```suggestion
      if ( /^d\d+$/.test(item.system.hitDice) ) {
```

module/documents/actor/HitDice.mjs
```diff
@@ -0,0 +1,172 @@
...
+        this._classes.add(item);
+        this.remaining += item.system.levels - item.system.hitDiceUsed;
+        this.total += item.system.levels;
+        this.sizes.add(parseInt(item.system.hitDice.replace("d", "")));
```
Fyorl, line 43:
```suggestion
        this.sizes.add(parseInt(item.system.hitDice.slice(1)));
```

module/documents/actor/HitDice.mjs
```diff
@@ -0,0 +1,172 @@
...
+  /**
+   * Return an object of remaining hit dice categorized by size.
+   * @returns {object}
+   */
+  get bySize() {
+    const hd = {};
+    this._classes.forEach(cls => {
+      const d = cls.system.hitDice;
+      const remaining = cls.system.levels - cls.system.hitDiceUsed;
+      hd[d] = ( d in hd ) ? hd[d] + remaining : remaining;
```
Fyorl, line 129:
```suggestion
      hd[d] = (hd[d] ?? 0) + remaining;
```

Fyorl (changes requested):
This will make dealing with hit dice a lot more ergonomic, thanks

module/documents/actor/HitDice.mjs
```diff
@@ -0,0 +1,172 @@
+/**
+ * Object describing the hit dice for an actor.
+ *
+ * @param {Actor5e} actor
+ */
+export default class HitDice {
+  constructor(actor) {
```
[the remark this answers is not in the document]
krbz999: > and move these initializers out of the constructor and into class fields instead.

I'm not entirely sure what you mean this.

Fyorl (approved):

PR #3235, opened 2024-03-11
[#893] Add system for selecting summoning placement

### LINKED ISSUE #893: [Feature Request] Provide a mechanism for actors like summons / familiars / companions / mounts to provide a reference back to another actor which owns or summoned them.

arbron:
The `TokenPlacement` class right now is small and only delegates to the `TokenPlacementTemplate` which is responsible for rendering and returning placement information. With the V12 improvements to placeables hopefully we'll be able to shift more logic out of the `MeasuredTemplate` subclass and into the `TokenPlacement` class.

The `TokenPlacementConfiguration` data structure currently just includes the prototype token information, but will eventually have quantity, origin, and range values to handle multiple summons and restricting range from the summoner.

Works with tokens of any size or scale on the square grid. On hex grids, it handles 1x1 tokens pretty good with only a bit of offset on the final token placement. Stranger token sizes lead to some issues with positioning, but I'm not sure how important those are to fix before the other grid improvements in V12.

Closes #893

### DIFF
--- a/module/canvas/token-placement.mjs
+++ b/module/canvas/token-placement.mjs
@@ -34,53 +34,12 @@ export default class TokenPlacement {
   config;
 
   /* -------------------------------------------- */
-  /*  Placement                                   */
-  /* -------------------------------------------- */
-
-  /**
-   * Perform the placement, asking player guidance when necessary.
-   * @returns {PlacementData[]}
-   */
-  async place() {
-    const template = TokenPlacementTemplate.create(this.config);
-    const placement = await template.place();
-    return placement;
-  }
-}
-
-
-/**
- * A MeasuredTemplate used to visualize token placement in the scene.
- */
-export class TokenPlacementTemplate extends MeasuredTemplate {
-
-  /**
-   * Configuration information for this placement.
-   * @type {TokenPlacementConfiguration}
-   */
-  config;
-
-  /* -------------------------------------------- */
 
   /**
    * Track the bound event handlers so they can be properly canceled later.
    * @type {object}
    */
   #events;
-
-  /* -------------------------------------------- */
-
-  /**
-   * The initially active CanvasLayer to re-activate after the workflow is complete.
-   * @type {CanvasLayer}
-   */
-  #initialLayer;
-
-  /**
-   * Placements that have been generated.
-   * @type {PlacementData[]}
-   */
-  placements;
 
   /* -------------------------------------------- */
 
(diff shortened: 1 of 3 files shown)


module/canvas/token-placement.mjs
```diff
@@ -0,0 +1,410 @@
...
+
+  /**
+   * Interval for snapping token to the grid.
+   * @type {number}
+   */
+  get snapInterval() {
+    // TODO: Figure out proper snapping sized based on token size
+    return canvas.grid.type === CONST.GRID_TYPES.GRIDLESS ? 0
+      : Math.max(this.config.tokens[0].width, this.config.tokens[0].height) > 0.5 ? 1 : 2;
+  }
```
arbron, line 250:
```suggestion
```

module/canvas/token-placement.mjs
```diff
@@ -0,0 +1,259 @@
...
+    switch ( canvas.grid.type ) {
+      case CONST.GRID_TYPES.SQUARE:
+        return {
+          x: token.width % 2 === 0 ? Math.round(size * 0.5) : 0,
+          y: token.height % 2 === 0 ? Math.round(size * 0.5) : 0
+        };
+      default:
+        return { x: 0, y: 0 };
+    }
+  }
```
Fyorl, line 155: We should be able to use `canvas.grid.getSnappedPosition` (`getSnappedPoint` in v12) for this. I don't think we need to handle this ourselves.

module/canvas/token-placement.mjs
```diff
@@ -0,0 +1,259 @@
...
+    const point = event.data.getLocalPosition(canvas.tokens);
+    const center = canvas.grid.getCenter(point.x - adjustment.x, point.y - adjustment.y);
+    preview.updateSource({
+      x: center[0] + adjustment.x - Math.round((this.config.tokens[0].width * canvas.dimensions.size) / 2),
+      y: center[1] + adjustment.y - Math.round((this.config.tokens[0].height * canvas.dimensions.size) / 2)
+    });
+    this.#placements[0].x = preview.x;
+    this.#placements[0].y = preview.y;
+    preview.object.refresh();
+    this.#throttle = true;
```
Fyorl, line 218: Handling throttling via `requestAnimationFrame` is probably preferable I think. `#throttle` can become a simple boolean.

```suggestion
    if ( this.#throttle ) return;
    this.#throttle = true;
    const preview = this.#previews[0];
    const adjustment = this.#getSnapAdjustment(preview);
    const point = event.data.getLocalPosition(canvas.tokens);
    const center = canvas.grid.getCenter(point.x - adjustment.x, point.y - adjustment.y);
    preview.updateSource({
      x: center[0] + adjustment.x - Math.round((this.config.tokens[0].width * canvas.dimensions.size) / 2),
      y: center[1] + adjustment.y - Math.round((this.config.tokens[0].height * canvas.dimensions.size) / 2)
    });
    this.#placements[0].x = preview.x;
    this.#placements[0].y = preview.y;
    preview.object.refresh();
    requestAnimationFrame(() => this.#throttle = false);
```

Fyorl (changes requested):
Please see my comments, but this looks great.

module/canvas/token-placement.mjs
arbron: I find that using `getSnappedPosition` works fine for 1x1 tokens, but results in larger or stranger shaped tokens being offset in rather than centered under the token. I tried this:

```javascript
const center = canvas.grid.getSnappedPosition(
  point.x - Math.round(canvas.dimensions.size * 0.5),
  point.y - Math.round(canvas.dimensions.size * 0.5),
  1,
  { token: this.config.tokens[0].object }
);
```

For 2x2 tokens for example, this would result in the token being places with the top-left quarter centered over the square that I am hovering which makes it non-intuitive where to move the mouse to place the token.

Fyorl: Ah, that's fair enough. I think this has all been improved in v12 so I will experiment with checking `game.release` here and calling `getSnappedPoint` instead, but this seems good enough for now.

Fyorl (approved):

PR #3236, opened 2024-03-11
Remove unneeded effects from races

krbz999:
- Removed effect that granted poison resistance from dwarf feature. Dwarf race already had this.
- Removed effect that granted Tinker's Tools proficiency (poorly) from gnome feature.
- Added Tinker's Tools proficiency in gnome race.
- Removed effect that set the Halfling Lucky special trait from halfling feature. This is inline with how the Half-Orc is set up, and the feature already included a note on how to set it up manually.
- Removed effect that granted fire resistance from tiefling feature. Tiefling race already had this.

### DIFF
--- a/packs/_source/races/gnome/rock-gnome.json
+++ b/packs/_source/races/gnome/rock-gnome.json
@@ -90,6 +90,23 @@
         },
         "level": 0,
         "title": "",
+        "icon": null,
+        "value": {}
+      },
+      {
+        "_id": "3iZH8gdahsVwLPbQ",
+        "type": "Trait",
+        "configuration": {
+          "mode": "default",
+          "allowReplacements": false,
+          "grants": [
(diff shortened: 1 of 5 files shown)


packs/_source/races/gnome/rock-gnome.json
```diff
@@ -90,6 +90,23 @@
...
+          "mode": "default",
+          "allowReplacements": false,
+          "grants": [
+            "tool:art:tinker"
+          ],
+          "choices": [],
+          "hint": ""
+        },
+        "level": 0,
+        "title": "",
```
arbron, line 109: We should give this a name to match the feature, same way the description lists it.

```suggestion
        "title": "Tinker",
```

arbron (changes requested):

arbron (approved):

PR #3237, opened 2024-03-12
#3065 Modify Deterministic Removal in simplifyRollFormula

### LINKED ISSUE #3065: Roll command errors when using multiplication or division
Using a `/roll` command involving multiplication or division results in an error. For example, `/roll 1d6 * 5` yields

```
Uncaught (in promise) Error: Roll.safeEval produced a non-numeric result from expression " * 5"
    safeEval foundry.js:10437
    simplifyRollFormula simplify-roll-formula.mjs:42
    _enrichRollTooltip chat-message.mjs:251
    _enrichChatCard chat-message.mjs:233
    _enrichChatCard chat-message.mjs:232
    getHTML chat-message.mjs:58
```

Made sure to test in a fresh world with no modules enabled. Also tested in a fresh world with pf2 system, but got no errors. Putting parentheses around the roll avoids the issue.

Hoppyhob:
There was an issue where our simplifyRollFormula Deterministic check would loop through the roll terms and remove any term marked as Deterministic. This caused an issue if a *, /, or % operator was used on a Deterministic term, as the Deterministic term was removed but not the operator and other term that can not be evaluated as those operators are dependent on the Deterministic term. The solution was to treat the operator and other term as Deterministic as well, and remove them. This does not affect *, /, and % used on constants.
Additionally modified _enrichRollTooltip in chat-message.mjs to pass _formula as this was the roll's original formula before evaluation where parenthesis are collapsed, providing a more accurate constant. 
Closes #3065

### DIFF
--- a/module/dice/simplify-roll-formula.mjs
+++ b/module/dice/simplify-roll-formula.mjs
@@ -17,17 +17,39 @@ export default function simplifyRollFormula(formula, { preserveFlavor=false, det
 
   // Optionally strip flavor annotations.
   if ( !preserveFlavor ) roll.terms = Roll.parse(roll.formula.replace(RollTerm.FLAVOR_REGEXP, ""));
+
   if ( deterministic ) {
-    // Remove non-deterministic terms and their preceding operators.
+    // Perform arithmetic simplification to simplify parsing through the terms.
+    roll.terms = _simplifyOperatorTerms(roll.terms);
+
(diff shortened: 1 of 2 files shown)


Fyorl (approved):
I did a small reshuffle/code style pass and I think it should all still be in order but feel free to raise another PR if I made any errors, thanks.

Issue #3183, opened 2024-03-04
`rollDamage` no longer returns the correct result when the `returnMultiple` parameter is false or undefined


Javalar:
# Description
Before version 3.0, when a weapon or spell had multiple damage rolls (say for example, 2d8 Fire + 2d6 Radiant) the rolls where combined, such that the `rollDamage` function would return the result of a combined roll of `2d8 + 2d6`.

I see that it is now possible to pass the `returnMultiple` option in order to receive multiple roll results in an array. In the example above, the result would be an array of two rolls, one for the `2d8`, and the other for the `2d6`.

*But*, if `returnMultiple` is false or omitted, the resulting roll is only for the `2d8` rather than a combining of the rolls as before. This had led to one of my critical modules, not surprisingly, breaking.

I can fix the module (in fact being able to receive an array of rolls *greatly* simplifies the code. Seriously, thank you for that.), but this would still remain a bug as far as retro-compatibility is concerned.

# Expected Behaviour
In order to replicate the behaviour in version 2.x, when the `returnMultiple` option is false or undefined, the `rollDamage` function should combine rolls together and return a single combined result.

For example, if the `Flame Strike` spell is cast, the result of the `rollDamage` function is a single roll of `4d6 + 4d6`.

# Actual Behaviour
Only the first roll is returned. So in the case of `Flame Strike`, the result of a single roll of `4d6` is returned.

arbron:
Probably not worth fixing at this point because the `returnMultiple` flag is being removed completely in the next version after being deprecated since `4.0`.

PR #3241, opened 2024-03-12
[#3183] Adjust rollDamage to return merged roll, rather than first

arbron:
In `3.0.0` we modified `Item5e#rollDamage` to support separate rolls for multiple damage parts and added `returnMultiple` option for backward compatibility. The issue with that initial implementation is that if `returnMultiple` is `false`, it returns the first roll made rather than matching previously behavior of having all the damage parts merged into a single roll.

This change alters the bahavior of `damageRoll` if `returnMultiple` is `false` and there are more than one roll to create a new merged roll from the evaulated rolls and return that.

The separate rolls are still stored in the chat message and used by the system, but the combined rolls are now available to modules that rely on the `dnd5e.rollDamage` hook or call `Item5e#rollDamage` directly.

### DIFF
--- a/module/dice/dice.mjs
+++ b/module/dice/dice.mjs
@@ -217,7 +217,24 @@ export async function damageRoll({
 
   // Create a Chat Message
   if ( rolls?.length && chatMessage ) await CONFIG.Dice.DamageRoll.toMessage(rolls, messageData, { rollMode });
-  return returnMultiple ? rolls : rolls[0];
+  if ( returnMultiple ) return rolls;
+  if ( rolls?.length <= 1 ) return rolls[0];
+
+  const mergedRoll = new CONFIG.Dice.DamageRoll();
+  mergedRoll._total = 0;
+  for ( const roll of rolls ) {
+    if ( mergedRoll.terms.length ) {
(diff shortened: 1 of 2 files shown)


tposney:
Shouldn't the returned roll be a DamageRoll rather than a Roll?

The options.damageType will end up being that of the first roll which means you can't recover the damage type for each of the terms in the merged roll. I realise there is no solution for that with the current damageRoll, but perhaps consideration for each term optionally having a damageType to allow for that.

arbron:
> Shouldn't the returned roll be a DamageRoll rather than a Roll?

Good catch!

> The options.damageType will end up being that of the first roll which means you can't recover the damage type for each of the terms in the merged roll. I realise there is no solution for that with the current damageRoll, but perhaps consideration for each term optionally having a damageType to allow for that.

Since damage type is a new option that wasn't available with the legacy single-roll format and this is a backwards-compatibility feature, we probably don't need to handle that case. Anyone who wants to take advantage of the roll-specific damage types can set `returnMultiple` to be `true` in the pre-roll hook and get the data they need.

Fyorl (approved):

Issue #3250, opened 2024-03-14
Aggregate damage types before passing to `applyDamage`


arbron:
Right now damage passed to `applyDamage` is grouped based on its originating rolls, but this causes issues with resistance when roll flavor is used to add damage types. A piercing weapon that does `1d6 + 5` damage combined with a player's damage bonus of `2d6[fire]` will all be treated as piercing damage when applied.

`1d6 + 5 + 2d6[fire]` => `1d6 + 5` (piercing) & `2d6` (fire)

PR #3251, opened 2024-03-14
[#3250] Aggregate damage types before passing to `applyDamage`

arbron:
Adds a new dice helper function `aggregateDamageRolls` which splits and re-combines damage rolls based on the roll's type, term flavor, and optionally any roll properties. This function is then used by `ChatMessage5e#applyChatCardDamage` before rolls are totaled to ensure that damage types are properly split.

The function starts by using a `chunkTerms` helper function which goes through the terms in a roll grouping into chunks based on the operators. Addition and subtraction cause new chunks to be created while multiplication and division are included in the same chunk. This is so that a damage type defined in flavor of a term can spread to other terms that it is multiplied or divided with. So if you have a formula `1d6 + 5 * 1d4[fire]` that has the overall type of piercing, the `1d6` will remaining piercing while the `5 * 1d4` will all be treated as fire.

Once the chunking is complete then chunks from all different rolls are combined into lists of terms grouped by type and properties (so long as the `respectProperties` option is set, otherwise they will only be grouped by type). These sets are then converted into new `DamageRoll`s with options set appropriately.

### DIFF
--- a/module/dice/_module.mjs
+++ b/module/dice/_module.mjs
@@ -1,3 +1,4 @@
+export {default as aggregateDamageRolls} from "./aggregate-damage-rolls.mjs";
 export {default as D20Roll} from "./d20-roll.mjs";
 export {default as DamageRoll} from "./damage-roll.mjs";
 export * from "./dice.mjs";
--- a/module/dice/aggregate-damage-rolls.mjs
+++ b/module/dice/aggregate-damage-rolls.mjs
@@ -1,0 +1,77 @@
+/**
+ * Parse the provided rolls, splitting parts based on damage types & properties, taking flavor into account.
+ * @param {DamageRoll[]} rolls                   Evaluated damage rolls to aggregate.
+ * @param {object} [options={}]
+ * @param {boolean} [options.respectProperties]  Should damage properties also affect grouping?
+ * @returns {DamageRoll[]}
+ */
+export default function aggregateDamageRolls(rolls, { respectProperties }={}) {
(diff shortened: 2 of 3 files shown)


krbz999:
Does it make sense to use this somewhere early in `Item5e#rollDamage`?

arbron:
> Does it make sense to use this somewhere early in `Item5e#rollDamage`?

Maybe…though this was written without taking into account things like unresolved string terms so it might lead to issues if we try it before the rolls occur. For example if the formula is `1d6 + @mod + @attackBonus` (piercing) and then the data includes `{ attackBonus: "2d8[radiant]" }`, we don't have the information available to properly split damages until those terms are resolved.

If we wait until after the terms are resolved during the rolling process, then we are creating brand new rolls before they are saved to the chat message, so the saved rolls are different than what was actually rolled.

Perhaps something we can look at in the future, but for now it feels better to save the data as-rolled and do this processing after the fact when we actually need the aggregation.

Fyorl (commented):
How does this overlap with `ChatMessage5e#_aggregateDamageRoll`? It seems they do fairly similar things. Can we simplify that method at all by delegating some of the work to these helpers?

arbron:
> How does this overlap with `ChatMessage5e#_aggregateDamageRoll`? It seems they do fairly similar things. Can we simplify that method at all by delegating some of the work to these helpers?

This should be able to be used in concert with that, but since that method has some extra work to do I didn't want to work on it in the same PR.

Fyorl (approved):
Just one potential issue, then it looks good to merge.

Issue #3040, opened 2024-02-13
Add ability to bypass or downgrade certain damage resistances or immunities


arbron:
Some abilities like "Elemental Adept" bypass resistance to a certain type of damage. There are also certain abilities that treat an immunity to a certain damage type as a resistance instead.

arbron:
Current damage application options work well for ignoring and resistances, vulnerabilities, or immunities, but we'll need something extra for downgrading:

```javascript
/**
 * @typedef {object} DamageApplicationOptions
 * …
 * @property {boolean|Set<string>} [downgrade] Should this actor's damage immunity be downgraded to resistance?
 *                                             A set of damage types to be downgraded or `true` to downgrade all.
 *                     
 */
```

In terms of how to represent this on the actor, that is a bit more complicated because a feature like "Elemental Adept" only works for spells that do that type of damage, but you could conceivably want this to also apply to all damage of that type. One possibility is just adding a `flags.dnd5e.elementalAdept` that contains a list of damage types. That would allow the system to send the appropriate data to damage rolls and to implement the necessary changes to tread any 1s as 2s.

PR #3253, opened 2024-03-15
[#3040] Add `downgrade` option to damage calculating

arbron:
This new option will take any immunities and downgrade them to resistances. Downgrade will be ignore if that immunity is ignored, but not if the resistance is ignored. This allows for features that ignore resistance to a certain damage type and treat immunity as resistance.

This should allow the implementation of features like Fiery Soul from the Pyromancy origin in Magic the Gathering - Plane Shift: Kaladesh:
Screenshot 2024-03-15 at 10 46 29


Fyorl:
Feels like this logic might be simpler if our definition of 'downgrade' were more general, i.e. immunity -> resistance, resistance -> normal. It would cover the case of Fiery Soul in your example since resistance would be downgraded to normal, effectively being ignored.

krbz999:
May as well allow for 'normal' to be downgraded to 'vulnerability' while we're at it, then.

Fyorl:
> May as well allow for 'normal' to be downgraded to 'vulnerability' while we're at it, then.

Well then we no longer implement the example feature correctly.

krbz999:
> > May as well allow for 'normal' to be downgraded to 'vulnerability' while we're at it, then.
> 
> Well then we no longer implement the example feature correctly.

True. Should it be more configurable? E.g., only allowing to downgrade resistance but not an immunity on a per-roll or per-feature basis?

Admittedly not aware of anything that ignores resistance only while not ignoring immunity.

Fyorl:
> Admittedly not aware of anything that ignores resistance only while not ignoring immunity.

Most features that do this sort of thing actually ignore resistance but not immunity. The Elemental Adept feat for example, does nothing if the target has immunity to the chosen element. But in those cases you would just pass `ignore`, rather than `downgrade`. With `ignore` + `downgrade` you have a decent amount of configurability, I was hoping that we could simplify the code a bit for `downgrade`.

arbron:
> Feels like this logic might be simpler if our definition of 'downgrade' were more general, i.e. immunity -> resistance, resistance -> normal. It would cover the case of Fiery Soul in your example since resistance would be downgraded to normal, effectively being ignored.

My only concern is that this would make it impossible for something that downgrades immunity but still keeps resistances, but I'm not sure I've actually seen that feature.

Fyorl:
> My only concern is that this would make it impossible for something that downgrades immunity but still keeps resistances, but I'm not sure I've actually seen that feature.

Yeah, me neither. In the absence of such a feature, I'd rather we take the complexity win if we can.

arbron:
@Fyorl Okay now `downgrade` reduces immunity to resistance and resistance to normal.

PR #3255, opened 2024-03-15
Migrate ItemGrant & Choice to store items as objects rather than UUIDs

arbron:
Necessary migration to allow for additional per-object configuration of granted items or choices. Now granted & chosen items are represented by and object with a `uuid` value, rather than a straight UUID string.

May be a breaking change for any custom migrations relying on the drop functionality built in to `AdvancementConfig` which now assumes objects rather than strings as the target value. This could be made to not be a breaking change if we added some extra handling to detect the field type of the target (about a 3 line change if desired), but I'm not sure anyone is relying on this at the moment.

Had to remove `AdvancementDataField#_cleanType` which was interfering with the migrations, but I don't think it is necessary because cleaning is handled during the sub-model initialization.

### DIFF
--- a/module/applications/advancement/advancement-config.mjs
+++ b/module/applications/advancement/advancement-config.mjs
@@ -68,7 +68,7 @@ export default class AdvancementConfig extends FormApplication {
   /** @inheritdoc */
   async close(options={}) {
     await super.close(options);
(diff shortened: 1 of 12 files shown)


Fyorl (approved):
LGTM

PR #3256, opened 2024-03-15
[#3165] Prevent template placement if no scene is open

### LINKED ISSUE #3165: Do not offer to place measured template if no scene is open
Currently the item activation dialog will try to place a measured template by default even if no scene is open, leading immediately to an error as it fails to place the template. We should disable the option in that dialog if no scene is open to prevent this issue.

arbron:
Adds a check when preparing `ItemUseConfiguration` to ensure scene is open before presenting the option to place a template.

Also adds a check when rendering chat messages to ensure a scene is open and the user has permission to place template. The one limitation of the chat card change is that the button will not re-appear when a scene is opened, requiring a refresh to see the button. An alternate approach would be adding a style that hides the button if no scene is open, but that would require a class or property somewhere up the DOM that it could check, which isn't currently available.

Closes #3165

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -91,8 +91,11 @@ export default class ChatMessage5e extends ChatMessage {
 
       // If the user is the message author or the actor owner, proceed
       let actor = game.actors.get(this.speaker.actor);
-      if ( actor && actor.isOwner ) return;
-      else if ( game.user.isGM || (this.user.id === game.user.id)) return;
+      if ( game.user.isGM || actor?.isOwner || (this.user.id === game.user.id) ) {
+        const template = chatCard[0].querySelector('button[data-action="placeTemplate"]');
+        if ( template && !game.user.can("TEMPLATE_CREATE") ) template.style.display = "none";
+        return;
+      }
 
       // Otherwise conceal action buttons except for saving throw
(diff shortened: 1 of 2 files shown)


Fyorl (changes requested):
Removing the prompt from the `AbilityUseDialog` if there is no scene active makes sense. I think for the chat card, the least bad option is probably to just throw up a warning notification when attempting to click the button with no scene active.

module/documents/chat-message.mjs
```diff
@@ -91,8 +91,11 @@ export default class ChatMessage5e extends ChatMessage {
 
       // If the user is the message author or the actor owner, proceed
       let actor = game.actors.get(this.speaker.actor);
-      if ( actor && actor.isOwner ) return;
-      else if ( game.user.isGM || (this.user.id === game.user.id)) return;
+      if ( game.user.isGM || actor?.isOwner || (this.user.id === game.user.id) ) {
+        const template = chatCard[0].querySelector('button[data-action="placeTemplate"]');
+        if ( template && !game.user.can("TEMPLATE_CREATE") ) template.style.display = "none";
```
Fyorl, line 96: This also has the issue that the chat card will not re-render if the player's template creation permissions are granted/revoked. I suppose that pretty rarely changes though.

Fyorl (approved):

PR #3261, opened 2024-03-17
Update roll data and active effect wikis for 3.x

### LINKED ISSUE #2755: Spell save DC bonus is missing from the AE guide
`system.bonuses.spell.dc` (+number/formula)

krbz999:
Closes #2755.

### DIFF
--- a/wiki/Roll-Formulas.md
+++ b/wiki/Roll-Formulas.md
@@ -167,9 +167,11 @@
 
 `@details.level` - Overall character level.
 
-`@details.xp.value` - Actor's total XP earned.
+`@details.cr` - Challenge rating.
 
-`@details.xp.min` & `.max` - XP range for the actor's current level.
+`@details.xp.value` - Actor's total XP earned (or the XP a monster is worth).
+
+`@details.xp.min` and `.max` - XP range for the actor's current level.
 
 `@details.xp.pct` - Progress towards the next level.
 
--- a/wiki/Roll-Formulas.md
+++ b/wiki/Roll-Formulas.md
@@ -75,7 +75,7 @@
 
 `@attributes.concentration.limit` - Maximimum number of effects the actor can sustain concentration on.
 
-`@attributes.concentration.roll.min` and `.max` - The minimum and maximum values of the d20 for concentration-related saving throws.
+`@attributes.concentration.save` - The final bonus an actor has to concentration saving throws.
 
 #### Death Saves
 
--- a/wiki/Roll-Formulas.md
+++ b/wiki/Roll-Formulas.md
@@ -296,3 +296,19 @@
 `@spells.*.override` - A value that overrides the calculated `max` slots.
 
 `@spells.*.level` - The level of the relevant spell slots (e.g., the level of the pact slots, or the '3' in 'spell3').
+
+## Item Properties
+
+Items have roll data same as actors, but only while said item is owned by an actor (i.e., is placed on their sheet). The roll data of item is an extension of the actor's roll data, adding `@item` as an additional property.
+
+The most common properties are:
+
+`@item.uses.value` - The current Limited Uses available on the item.
+
+`@item.uses.max` - The max Limited Uses of this item.
+
+`@item.save.dc` - The saving throw DC of this item.
+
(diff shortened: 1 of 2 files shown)


wiki/Roll-Formulas.md
```diff
@@ -52,10 +48,6 @@
 
 #### Armor Class
 
-`@attributes.ac.calc` - Calculation mode that its used for determining `attributes.ac.base`
-
-`@attributes.ac.formula` - Custom formula that will be used to determine `attributes.ac.base` if `calc` is set to `custom`
```
Fyorl, line 57: Why are these removed?

wiki/Roll-Formulas.md
```diff
@@ -79,201 +71,163 @@
 
 `@attributes.attunement.max` - Maximum number of attunement slots
 
+#### Concentration
+
+`@attributes.concentration.limit` - Maximimum number of effects the actor can sustain concentration on.
+
+`@attributes.concentration.roll.min` and `.max` - The minimum and maximum values of the d20 for concentration-related saving throws.
```
Fyorl, line 78: These don't do anything currently, do they? Probably shouldn't mention them on the wiki in that case.

wiki/Roll-Formulas.md
```diff
@@ -79,201 +71,163 @@
...
-
-`@bonuses.rsak` - Ranged spell attack
-
-`@bonuses.rwak` - Ranged weapon attack
-
-***Note:** Replace the `*` in the following formulas with the four-letter code from above.*
-
-`@bonuses.*.attack` - Global bonus to attack rolls
-
-`@bonuses.*.damage` - Global bonus to damage rolls
```
Fyorl, line 181: Why are we removing these? Did they get moved somewhere else?

wiki/Roll-Formulas.md
```diff
@@ -315,40 +269,46 @@
 
 ***Note:** Replace the `*` in the following formulas with the three-letter code from above.*
 
-`@skills.*.ability` - Three letter code for the ability associated with this skill by default
```
Fyorl, line 318: Unclear why this is and the bonuses were removed.

Fyorl (changes requested):

wiki/Roll-Formulas.md
krbz999: They can't be used. Unsure why they were ever included; you can't construct a roll formula with the word "default", and we don't (off the top of my head) double-evaluate roll data to allow for use of `ac.formula`.

wiki/Roll-Formulas.md
krbz999: Same as before, they can't be used as roll data. But they are in the AE wiki.

wiki/Roll-Formulas.md
krbz999: Agree. Replacing with `.save`, which is a number (added during data prep).

wiki/Roll-Formulas.md
Fyorl: Hmm, I see. Well it's probably fine so long as they're in the AE wiki, but some of them, like `@bonuses.mwak` certainly could be used in roll data so long as they don't also contain roll data references themselves. I suppose I don't really know when you would want to do that though.

wiki/Roll-Formulas.md
Fyorl: `roll.min` and `roll.max` appear in a few other places in this changeset as well. I don't know if perhaps I'm getting confused with the overlap with AEs maybe, but I don't think those fields can even be configured in the UI (yet) so I'm not sure about including them in the wiki at this stage.

krbz999: Just me being focused on conc. They should not be there if they are not functioning yet.

wiki/Roll-Formulas.md
krbz999: I think excluding properties that can lead to issues in that way is a good rule of thumb to use (else we could or should technically include `@name` if you name your character "5").

Fyorl (approved):
Need to wait to merge this until 3.1 is released.

Issue #1372, opened 2021-12-12
Provide buttons in chat message for options to apply damage and healing


aaclayton:
Originally in GitLab by @epsilon7430

As a user, it is valuable to have buttons available in a chat message to be able to quickly apply damage or healing, so that I do not have to first right-click to open a context menu and then left-click to select a menu option to apply.  

Buttons improve a repetitive workflow, which is beneficial to both game play and physical health.  

Buttons would appear on mouse pointer hover over the damage box.

Buttons would only be available to players permitted to apply damage/healing.  The availability of buttons therefore indicates to the player that they can do this, versus having to right-click first to see if the context options are available.

Option for users to enable/disable buttons would be most flexible.

A working example are the buttons added by the BetterRolls5e module (compatibility with Foundry v9+ and future DnD5e system is unknown).

![image](https://gitlab.com/foundrynet/dnd5e/uploads/54466d9298ca2b4c9ef8aea9c660db04/buttons.PNG)

PR #3265, opened 2024-03-18
[#1015, #1372] Add damage application tray to chat cards

### LINKED ISSUE #1015: Add Quarter Damage to Apply Damage Dropdown
Originally in GitLab by @jpmeehan5

As a user, when I right click a damage roll, I would like the option to select "Apply quarter damage" in addition to the existing 2x, 1/2x options. This would be used when a creature resistant to a damage type succeeds on a saving throw, such as a tiefling caught in a fireball.

arbron:
Introduces a new `DamageApplicationElement` that is inserted into chat cards to handle applying damage to actors. It features two targeting modes, one that uses the tokens that were targeted when the damage was rolled and another that uses the currently selected tokens on the scene. For players the application will only display if one of their tokens was targeted by the damage. They will only see targeted selection mode and only the tokens of theirs that were targeted will be listed.

Damage Application V2

Each target lists the amount of damage calculated after resistances, immunities, and vulnerabilities are taken into account. Beneath that are a list of multiplier controls that appear for each target, make it easy to apply damage changes outside the automatic ones. Changes to the damage from these buttons is automatically reflected in the total number.

Closes #1015

### Future Work
- [ ] Display resistances, immunities, and vulnerabilities that are affecting the damage total
- [ ] Add ability to bypass certain resistances/immunities/vulnerabilities

### DIFF
--- a/module/applications/components/damage-application.mjs
+++ b/module/applications/components/damage-application.mjs
@@ -50,18 +50,24 @@ export default class DamageApplicationElement extends HTMLElement {
    * @type {"targeted"|"selected"}
    */
   get targetingMode() {
-    if ( !game.user.isGM ) return "targeted";
     if ( this.targetSourceControl.hidden ) return "selected";
     return this.targetSourceControl.querySelector('[aria-pressed="true"]')?.dataset.mode ?? "targeted";
   }
 
   set targetingMode(mode) {
+    if ( this.targetSourceControl.hidden ) mode = "selected";
     const toPress = this.targetSourceControl.querySelector(`[data-mode="${mode}"]`);
-    if ( this.targetSourceControl.hidden || (toPress?.ariaPressed === "true") ) return;
     const currentlyPressed = this.targetSourceControl.querySelector('[aria-pressed="true"]');
     if ( currentlyPressed ) currentlyPressed.ariaPressed = false;
     toPress.ariaPressed = true;
+
     this.buildTargetsList();
+    if ( (mode === "targeted") && (this.selectedTokensHook !== null) ) {
+      Hooks.off("controlToken", this.selectedTokensHook);
+      this.selectedTokensHook = null;
+    } else if ( (mode === "selected") && (this.selectedTokensHook === null) ) {
+      this.selectedTokensHook = Hooks.on("controlToken", foundry.utils.debounce(() => this.buildTargetsList(), 50));
+    }
   }
 
   /* -------------------------------------------- */
@@ -85,7 +91,7 @@ export default class DamageApplicationElement extends HTMLElement {
    * @param {string} uuid  UUID of the targeted token.
    * @returns {DamageApplicationOptions}
    */
-  targetOptions(uuid) {
+  getTargetOptions(uuid) {
     if ( !this.#targetOptions.has(uuid) ) this.#targetOptions.set(uuid, { multiplier: 1 });
     return this.#targetOptions.get(uuid);
   }
@@ -102,7 +108,7 @@ export default class DamageApplicationElement extends HTMLElement {
   /*  Rendering                                   */
   /* -------------------------------------------- */
 
-  async connectedCallback() {
+  connectedCallback() {
     // Fetch the associated chat message
     const messageId = this.closest("[data-message-id]")?.dataset.messageId;
     this.chatMessage = game.messages.get(messageId);
@@ -120,7 +126,7 @@ export default class DamageApplicationElement extends HTMLElement {
         </label>
         <div class="collapsible-content">
           <div class="target-source-control">
-            <button type="button" class="unbutton" data-mode="targeted" aria-pressed="true">
+            <button type="button" class="unbutton" data-mode="targeted" aria-pressed="false">
               <i class="fa-solid fa-bullseye" inert></i> ${game.i18n.localize("DND5E.Tokens.Targeted")}
             </button>
             <button type="button" class="unbutton" data-mode="selected" aria-pressed="false">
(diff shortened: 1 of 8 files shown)


krbz999:
Since the GM's view of this can dynamically change depending on their selection, could similar not be done for players? There are games where the GM simply does not make use of targeting, and I don't feel that this should be required by the GM to allow for the players to make use of this.

module/applications/components/damage-application.mjs
```diff
@@ -0,0 +1,334 @@
...
+    switch ( this.targetingMode ) {
+      case "targeted":
+        targetedTokens = (this.chatMessage.getFlag("dnd5e", "targets") ?? []).map(t => t.uuid);
+        if ( this.selectedTokensHook ) {
+          Hooks.off("controlToken", this.selectedTokensHook);
+          this.selectedTokensHook = null;
+        }
+        break;
+      case "selected":
+        targetedTokens = canvas.tokens.controlled.map(t => t.document.actor.uuid);
```
krbz999, line 176: Did some quick fooling around, and this lead to some weirdness with multiple tokens of the same actor being duplicated (as well as tokens with no actor throwing errors).

arbron:
> Since the GM's view of this can dynamically change depending on their selection, could similar not be done for players? There are games where the GM simply does not make use of targeting, and I don't feel that this should be required by the GM to allow for the players to make use of this.

Main reason I didn't go that way is to avoid adding the application control to every single damage message for players, rather just showing the ones applicable to their characters. Could go either way I suppose.

krbz999:
> Main reason I didn't go that way is to avoid adding the application control to every single damage message for players, rather just showing the ones applicable to their characters. Could go either way I suppose.

I think not directly requiring targeting, by the GM, for damage rolls would be helpful. Besides the tray is initially collapsed and does not take up much space.

Fyorl:
I don't think this should be available to players at all really, for similar reasons to why the effect tray is also (mostly) unavailable to players. I think there is a lot of merit to having a clear and consistent UX of:

1. Creatures take their actions.
2. The DM applies the results.

For several reasons:
1. The DM is the only one that can always apply anything to any token. There's no confusion about what options appear when and why, and no confusion about under what circumstances they work and do not.
2. It runs parallel to the most basic rules in the 'How To Play' section of the PHB, i.e. 'The players describe what they want to do. The DM narrates the results.' So it should be intuitive to anyone who has played the game.
3. It allows us to reach a point in the 'action workflow' (from the [roadmap](https://github.com/foundryvtt/dnd5e/releases/tag/release-3.0.0#user-content-whats-next)) where an action can be considered fully 'resolved'. If players are piecemeal applying parts of a damage card to themselves and not other targets (because they lack ownership), then determining action resolution requires more work.
4. It's simpler to implement.

module/applications/components/damage-application.mjs
Fyorl: When you say 'multiple tokens of the same actor', I assume you mean multiple *linked* tokens of the same actor? Unlinked token actors should have unique UUIDs from each other. We should de-dupe those though since applying damage to any one of a linked actor's tokens applies damage to all of them. And we should fix the error for actor-less tokens of course.

module/applications/components/damage-application.mjs
```diff
@@ -0,0 +1,334 @@
...
+    if ( collapsible.style.height !== "0px" ) collapsible.style.height = "auto";
+
+    let targetedTokens;
+    switch ( this.targetingMode ) {
+      case "targeted":
+        targetedTokens = (this.chatMessage.getFlag("dnd5e", "targets") ?? []).map(t => t.uuid);
+        if ( this.selectedTokensHook ) {
+          Hooks.off("controlToken", this.selectedTokensHook);
+          this.selectedTokensHook = null;
+        }
```
Fyorl, line 173: Managing the hooks feels more appropriate to perform in the `targetingMode` setter. Perhaps we could move it there and then have `connectedCallback` invoke `this.targetingMode = this.targetingMode` instead of `this.buildTargetsList()` (and we would also need to make both controls `aria-pressed="false"` initially)? Now that I've written it, it looks a bit silly, but perhaps worth considering at least.

module/applications/components/damage-application.mjs
```diff
@@ -0,0 +1,334 @@
...
+        targetedTokens = canvas.tokens.controlled.map(t => t.document.actor.uuid);
+        this.selectedTokensHook ??= Hooks.on("controlToken", foundry.utils.debounce(() => this.buildTargetsList(), 50));
+        break;
+    }
+    const targets = await Promise.all(targetedTokens.map(t => this.buildTargetListEntry(t)));
+    if ( targets.length ) this.targetList.replaceChildren(...targets.filter(e => e));
+    else {
+      const li = document.createElement("li");
+      li.classList.add("none");
+      li.innerText = game.i18n.localize("DND5E.Tokens.NoneTargeted");
```
Fyorl, line 185: I guess for polish this should probably say 'no tokens targeted' or 'no tokens selected', depending on the current target mode.

module/applications/components/damage-application.mjs
```diff
@@ -0,0 +1,334 @@
...
+      this.targetList.replaceChildren(li);
+    }
+
+    // Reset collapsible height to allow for animation
+    requestAnimationFrame(() => {
```
Fyorl, line 190: I am going to try to revisit these sometime today or tomorrow to see if we can use the grid technique that cole mentioned, which will hopefully allow us to clean up all our auto-height collapsing code.

Fyorl (approved):
Please see my comments, but this looks great. Really nice UI.

module/applications/components/damage-application.mjs
arbron: Our linter complains about assigning a variable to itself, but I can easily determine what the initial mode should be based on the current state.

module/applications/components/damage-application.mjs
krbz999: Yeah sorry, that is what I meant. I fixed an identical issue [here](https://github.com/foundryvtt/dnd5e/pull/3090).

arbron:
@Fyorl I've adjusted so the damage application UI only appears for GM users for now (we can always loosen that in a followup PR).

Issue #3039, opened 2024-02-12
Provide various methods to manually prompt for saving throws to maintain concentration


krbz999:
Related to #3028.

According to official rules, a GM can request a saving throw to maintain concentration for reasons other than a character taking damage, such as being tossed around on a ship at sea, etc.

Some nice-to-haves:

- [x] The character sheet should display a button to roll concentration without being prompted.
- [x] A way to request with an inline roll or chat command (with a DC).
- [x] Relevant API methods.
- [x] Enrichers.

Fyorl:
Some of these are done, the rest will have to wait for 3.1.1 or 3.2

PR #3270, opened 2024-03-19
[#3039] Add concentration interactivity to the character sheet.

Fyorl:
Adds some indication to the sheet when the character is concentrating on an item:

image

Adds a context menu option to break concentration on any given item:

image

Adds a concentration save to the saving throws box that is visible when in edit mode or when actively concentrating on something:

image

Follow-ups:
- [ ] Configuration application for concentration bonuses
- [ ] Find somewhere to put ephemeral items that are being concentrated on (e.g. spells cast from scrolls)


arbron (approved):
Looks good, just a few thoughts:
- It would be good to have some additional indication on what the spinning icon represents. Could be some text after name that says `(concentrating)` if we want it available to all users or a `aria-label="Concentrating"` on just the icon
- The concentration icon in the saving throws section doesn't feel quite aligned with the proficiency icons. Perhaps there could be a pixel or two of extra padding to that entry to make them line up a bit more.

arbron:
One more thing I notice that warrants some thought but doesn't need to be solved now is how we handle the effect. Namely, right now you can disable the active effect that represents the concentration, but that has no effect on the icon. Do we want to perhaps have it so the concentration effect can't be disabled, but only broken?

krbz999:
One additional thing is where (or whether) to place ephemeral items being conc'd on on the sheet (e.g., if you are conc'ing on a spell from a scroll, which was consumed).

Fyorl:
> * It would be good to have some additional indication on what the spinning icon represents. Could be some text after name that says `(concentrating)` if we want it available to all users or a `aria-label="Concentrating"` on just the icon

Personal preference but I think it's ok if this is just learned. I would not prefer to have '(concentrating)' tacked onto spell names once I have learned what the icon represents. I've opted for the `aria-label` only for now.

> One more thing I notice that warrants some thought but doesn't need to be solved now is how we handle the effect. Namely, right now you can disable the active effect that represents the concentration, but that has no effect on the icon. Do we want to perhaps have it so the concentration effect can't be disabled, but only broken?

Good idea. I've made these effects untoggleable and removed the delete and disable options from their context menus, replacing them with a 'break concentration' option. This funnels all (user-interactable) ways of breaking concentration via the `Actor5e#endConcentration` method, which is probably desirable.

> One additional thing is where (or whether) to place ephemeral items being conc'd on on the sheet (e.g., if you are conc'ing on a spell from a scroll, which was consumed).

A good question. I've added it to the follow-ups. I don't have any good ideas off the top of my head. My first instinct would be a special category on the Effects tab (like the conditions category) since that tab usually has a fair bit of room (in my experience). I don't know how intuitive this is though since it violates the expectation of seeing ActiveEffects in the Effects tab.

Maybe add a new category to the Features, Inventory, & Spells tabs called 'ephemeral'. Some care needs to be taken with how this interacts with grouping and sorting, and general interactivity like drag-and-drop. Firmly in the nice-to-have category I think.

PR #3273, opened 2024-03-20
[#1372] Add display of resistances, etc. in damage application

### LINKED ISSUE #1372: Provide buttons in chat message for options to apply damage and healing

arbron:
Displays what resistances, immunities, vulnerabilities, and modifications are modifying the damage to be applied. Each of these can be clicked on to ignore that source of change or, if it is an immunity, downgrade it. The total damage that will be applied is updated as you click these buttons just like the multiplier buttons below.

Damage Application - Immunity
Damage Application - Bypassing Immunity

Closes #1372

### DIFF
--- a/module/applications/components/damage-application.mjs
+++ b/module/applications/components/damage-application.mjs
@@ -223,8 +223,8 @@ export default class DamageApplicationElement extends HTMLElement {
         <button class="change-source unbutton" type="button" data-type="${type}" data-change="${change}"
                 data-tooltip="${label}" aria-label="${label}" aria-pressed="${pressed}">
           <dnd5e-icon src="${icon}" inert></dnd5e-icon>
-          <i class="fa-solid fa-slash inert"></i>
-          <i class="fa-solid fa-chevron-down" inert></i>
+          <i class="fa-solid fa-slash" inert></i>
+          <i class="fa-solid fa-arrow-turn-down" inert></i>
         </button>
       `;
       return acc;
--- a/module/applications/components/damage-application.mjs
+++ b/module/applications/components/damage-application.mjs
@@ -206,7 +206,29 @@ export default class DamageApplicationElement extends HTMLElement {
 
     // Calculate damage to apply
     const targetOptions = this.getTargetOptions(uuid);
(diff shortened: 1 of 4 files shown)


module/applications/components/damage-application.mjs
```diff
@@ -206,7 +206,29 @@ export default class DamageApplicationElement extends HTMLElement {
...
+      }
+    }
+    const changeSources = types.reduce((acc, {type, change, icon}) => {
+      const { label, pressed } = this.getChangeSourceOptions(type, change, targetOptions);
+      acc += `
+        <button class="change-source unbutton" type="button" data-type="${type}" data-change="${change}"
+                data-tooltip="${label}" aria-label="${label}" aria-pressed="${pressed}">
+          <dnd5e-icon src="${icon}" inert></dnd5e-icon>
+          <i class="fa-solid fa-slash inert"></i>
+          <i class="fa-solid fa-chevron-down" inert></i>
```
Fyorl, line 227: There's not a lot of room to work with and I think you've done a great job with what is available. I found the chevron a little hard to see at first that it was a downwards chevron with some of the damage types I tested with (some were easier to see than others). We could try `angles-down` rather than `chevron-down` maybe. See what you think, if it doesn't get in the way of the icon too much. Otherwise we can stick with the chevron since it's good enough.

Fyorl (approved):
Nice, thank you Jeff

PR #3277, opened 2024-03-20
Add blank option in Concentration config

krbz999:
Just like initiative, it needs to be possible to set the ability to a blank option to default to the default.

### DIFF
--- a/templates/apps/concentration-config.hbs
+++ b/templates/apps/concentration-config.hbs
@@ -4,7 +4,7 @@
     <div class="form-group">
         <label>{{ localize "DND5E.AbilityModifier" }}</label>
         <select name="system.attributes.concentration.ability">
-            {{ selectOptions abilities selected=ability labelAttr="label" }}
+            {{ selectOptions abilities selected=ability blank="" labelAttr="label" }}
         </select>
     </div>
     <div class="form-group">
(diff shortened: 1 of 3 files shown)


templates/apps/concentration-config.hbs
```diff
@@ -4,7 +4,7 @@
     <div class="form-group">
         <label>{{ localize "DND5E.AbilityModifier" }}</label>
         <select name="system.attributes.concentration.ability">
-            {{ selectOptions abilities selected=ability labelAttr="label" }}
+            {{ selectOptions abilities selected=ability blank="" labelAttr="label" }}
```
Fyorl, line 7: Why? There's no affordance in the rules for not having an ability associated with a concentration save, is there?

krbz999: To allow to fall back to the default set in `CONFIG.DND5E.defaultAbilities` rather than having to adjust for each actor. This does not force the actor to make init/concentration without an ability.

Fyorl: Ah ok, that makes sense

Fyorl (changes requested):
Seems to apply the correct modifier when rolled, but shows as +0 on the character sheet.

Fyorl (approved):

Issue #3280, opened 2024-03-20
Expose the damage-application element for use by modules


etiquettestartshere:
In case modules want to expand the functionality of the GM only damage application tray, it could be hidden rather than removed to allow for this without, hopefully, too much hassle.

arbron:
For now we will only be adding it to GM chat messages, but it should be easy for modules to add it themselves to other messages using the same code we use:

```javascript
      const damageApplication = document.createElement("damage-application");
      damageApplication.classList.add("dnd5e2");
      damageApplication.damages = dnd5e.dice.aggregateDamageRolls(rolls, { respectProperties: true }).map(roll => ({
        value: roll.total,
        type: roll.options.type,
        properties: new Set(roll.options.properties ?? [])
      }));
      html.querySelector(".message-content").appendChild(damageApplication);
```

Fyorl:
Closing in favour of #3300

PR #3281, opened 2024-03-20
[#3280] expose damageApplication for use by modules

etiquettestartshere:
[#3280]
A small change to hide, instead of never create, the damage-application element on damage chat cards for non GMs. The intent of this is to expose this feature to modules, should they want to make it available to users other than the GM and/or modify it in some capacity. 

I am unsure if this is all that would need to change to make this usable by modules, but I tested simply removing the 'hidden' class used by this PR on a render hook with a module, and it worked as imagined.

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -359,16 +359,15 @@ export default class ChatMessage5e extends ChatMessage {
     `;
     html.querySelector(".message-content").appendChild(roll);
 
-    if ( game.user.isGM ) {
-      const damageApplication = document.createElement("damage-application");
-      damageApplication.classList.add("dnd5e2");
-      damageApplication.damages = aggregateDamageRolls(rolls, { respectProperties: true }).map(roll => ({
-        value: roll.total,
-        type: roll.options.type,
-        properties: new Set(roll.options.properties ?? [])
-      }));
-      html.querySelector(".message-content").appendChild(damageApplication);
-    }
+    const damageApplication = document.createElement("damage-application");
+    damageApplication.classList.add("dnd5e2");
+    if ( !game.user.isGM ) damageApplication.classList.add("hidden");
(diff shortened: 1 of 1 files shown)


arbron:
I'd probably lean towards just letting modules add the damage application element themselves to chat messages if they need it. This method means that the element is being initialized for all users even if they can't see it. Everything that should be required for modules to add it themselves exists (`<damage-application>` element registered, `aggregateDamageRolls` exported, and the rolls on the chat message themselves are available).

krbz999:
The change is very small and doesn't add any complexity. But perhaps it warrants adding a setting instead?

etiquettestartshere:
> I'd probably lean towards just letting modules add the damage application element themselves to chat messages if they need it. This method means that the element is being initialized for all users even if they can't see it. Everything that should be required for modules to add it themselves exists (`<damage-application>` element registered, `aggregateDamageRolls` exported, and the rolls on the chat message themselves are available).

That's more or less how I did it in effectivetray, though without the help of a similar registered element like `<damage-application>`. If something like this PR isn't acceptable, I think as you say, as far as I can tell nothing needs to be added for modules to construct it themselves. Personally, I don't think I will use this either way, as I'm more or less fine with damage application being GM only, but I do think potentially opening more things up to users is something the system should consider, given very common module choices and requests, etc. A setting sounds like a good idea.

Fyorl:
Agreed that initialising this element for all users indiscriminately on every damage card is unnecessary work.

etiquettestartshere:
I was encouraged to make this PR, though it was not my natural inclination. I did not see a middleground between initializing but not displaying this element 'indiscriminately' on every damage card and not making it available to modules at all. While arbron's suggested approach has some limitations (it is, for instance, difficult to access any element of the damage tray on render hooks due to the fact that everything below the damage-application element is injected at a later time--though this change wouldn't alter that, obviously), it is probably good enough, barring further breaks. With no comment about potential settings, I'll close this and refrain from making PRs based on suggestions from others in the future. Sorry to bother you.

PR #3294, opened 2024-03-21
[#2780] Allow default concept compendiums to be defined

### LINKED ISSUE #2780: Add setting for defining default Compendiums when clicking Add X
Add an option to define which compendiums should be opened when clicking Add Race/Add Background/etc so that users can default to their own personal compendiums if they want

arbron:
Takes advantage of the existing `CONFIG.DND5E.sourcePacks` object to define packs for backgrounds, classes, and races.

Closes #2780

### DIFF
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -1,6 +1,5 @@
 import CharacterData from "../../data/actor/character.mjs";
 import * as Trait from "../../documents/actor/trait.mjs";
-import { setTheme } from "../../settings.mjs";
 import { formatNumber, simplifyBonus, staticID } from "../../utils.mjs";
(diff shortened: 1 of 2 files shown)


Fyorl:
It doesn't seem like these have a use beyond `_onFindItem`, and any changes users or modules make to these variables will stop working once we adjust `_onFindItem` to launch the compendium browser instead.

arbron:
> It doesn't seem like these have a use beyond `_onFindItem`, and any changes users or modules make to these variables will stop working once we adjust `_onFindItem` to launch the compendium browser instead.

Probably true, but it's a fairly minor change to fix an annoyance people are having. I doubt any proper modules will use it (we could even add a comment to discourage that), but I could see myself using it for now with my own private content module while waiting for the compendium browser to come along.

PR #3302, opened 2024-03-22
[#3301] Remove everything related to conc if the item does not require conc at all

### LINKED ISSUE #3301: Don't show the concentration warning when casting a spell that doesn't require concentration
When concentrating on a spell, casting a spell that doesn't require concentration should not warn the user that they cannot maintain concentration on more effects.

Example: Akra is concentrating on Bless.
image

Attempting to cast Cure Wounds warns the user that they cannot maintain concentration on more effects. Cure Wounds doesn't require concentration, so this warning is irrelevant.
image

Running D&D5e 3.1.0 with no modules enabled.

krbz999:
Closes #3301


arbron:
@krbz999 Comment on the original issue for me so I can give it to you

arbron (approved):

PR #3303, opened 2024-03-22
[#3297] Prevent magical bonuses from applying if the item is not magical

### LINKED ISSUE #3297: Magical bonus to AC applies regardless of the item being magical
When 'Magical' is toggled off, the AC bonus should not be applying, but it do be doing that.

repro:
- make shield
- give +2 ac
- make magical
- give 2 magical bonus
- equip it
- gain +4 AC on actor
- toggle magical off on shield
- still have +4 AC

krbz999:
Closes #3297

### DIFF
--- a/module/data/item/templates/equippable-item.mjs
+++ b/module/data/item/templates/equippable-item.mjs
@@ -75,7 +75,8 @@ export default class EquippableItemTemplate extends SystemDataModel {
    * @type {boolean}
    */
   get magicAvailable() {
-    return this.attunement !== CONFIG.DND5E.attunementTypes.REQUIRED;
+    const attunement = this.attunement !== CONFIG.DND5E.attunementTypes.REQUIRED;
+    return attunement && this.properties.has("mgc") && this.validProperties.has("mgc");
   }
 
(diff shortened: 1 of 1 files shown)


arbron:
Any situation you think that an item will have the `mgc` property but that property is not in the valid properties list?

krbz999:
> Any situation you think that an item will have the `mgc` property but that property is not in the valid properties list?

We have support for adjusting these properties, so one could technically make it an invalid property for one of the system's item types, or a module subtype.

arbron:
> > Any situation you think that an item will have the `mgc` property but that property is not in the valid properties list?
> 
> We have support for adjusting these properties, so one could technically make it an invalid property for one of the system's item types, or a module subtype.

Still not convinced it is necessary, but if you keep it then you should probably do two `has` checks rather than an intersection, because the monkey-patched set intersection method has to iterate over all entries in the set.

```javascript
// O(1) time instead of O(n) time
const isMagical = this.validProperties.has("mgc") && this.properties.has("mgc");
```

arbron (approved):

PR #3304, opened 2024-03-22
[#3263] Preserve dice modifiers when simplifying formulas

### LINKED ISSUE #3263: `dnd5e.dice.simplifyRollFormula` removes dice modifiers
image

Expected: `6d4x`

krbz999:
Closes #3263

image

### DIFF
--- a/module/dice/simplify-roll-formula.mjs
+++ b/module/dice/simplify-roll-formula.mjs
@@ -166,7 +166,6 @@ function _simplifyDiceTerms(terms) {
       ? new Coin({ number: number })
       : new Die({ number, faces: parseInt(key.slice(1)), modifiers: [...new Set(modifiers)] })
   ]));
-  console.warn({diceQuantities, simplified});
   return [...simplified, ...annotated];
 }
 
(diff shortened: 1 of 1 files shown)


module/dice/simplify-roll-formula.mjs
```diff
@@ -150,17 +150,19 @@ function _simplifyDiceTerms(terms) {
...
+    obj[key].number = (obj[key].number ?? 0) + curr.number;
+    obj[key].modifiers = (obj[key].modifiers ?? []).concat(curr.modifiers);
     return obj;
   }, {});
 
   // Add new die and operator terms to simplified for each die size and sign
-  const simplified = Object.entries(diceQuantities).flatMap(([key, number]) => ([
+  const simplified = Object.entries(diceQuantities).flatMap(([key, {number, modifiers}]) => ([
     new OperatorTerm({ operator: key.charAt(0) }),
     key.slice(1) === "c"
```
Fyorl, line 163: If the `key` now contains the modifiers, I think this test will not correctly detect a `Coin` term with modifiers.

Fyorl (changes requested):

module/dice/simplify-roll-formula.mjs
Fyorl: Counting successes is probably a very common modifier for coin flips, so I'm not sure if it's *the* solution necessary. But certainly nothing in official content that I know of, and modifiers for coin flips were ignored before, so we haven't made anything *worse* here I suppose.

Fyorl (approved):

Issue #3308, opened 2024-03-22
Summoning without actor changes fail to retain item changes


arbron:
If the summoning occurring doesn't make any direct changes to the actor data but does to item data then the changes to items aren't applied properly.

PR #3309, opened 2024-03-22
[#3308] Fix item changes not being retained on summons

arbron:
Moved the stored origin flag from the token to the actor, which gets around the bug that seems to exist in core with updating `ActorDelta`.

### DIFF
--- a/module/data/item/fields/summons-field.mjs
+++ b/module/data/item/fields/summons-field.mjs
@@ -242,7 +242,7 @@ export class SummonsData extends foundry.abstract.DataModel {
    * @returns {object}                Changes that will be applied to the actor & its items.
    */
   async getChanges(actor, profile) {
-    const updates = { actor: {}, effects: [], items: [] };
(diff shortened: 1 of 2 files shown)


module/data/item/fields/summons-field.mjs
```diff
@@ -390,11 +390,9 @@ export class SummonsData extends foundry.abstract.DataModel {
       ui.notifications.warn("DND5E.Summoning.Warning.Wildcard", { localize: true });
     }
 
+    actorUpdates["flags.dnd5e.summon.origin"] = this.item.uuid;
```
Fyorl, line 393: Is this used anywhere? Do we need to update those references?

Fyorl (approved):

module/data/item/fields/summons-field.mjs
arbron: We haven't referenced them yet, intended for a potential future feature but probably works better on the synthetic actor anyway.

PR #3329, opened 2024-03-23
Add kh and kl as valid dice modifiers in Dice Scale Values

krbz999:
An oversight (who ever uses kh/kl, amirite?)

### DIFF
--- a/module/utils.mjs
+++ b/module/utils.mjs
@@ -50,7 +50,8 @@ export function isValidDieModifier(mod) {
     explode: /xo?([0-9]+)?([<>=]+)?([0-9]+)?/i,
     minimum: /(?:min)([0-9]+)/i,
     maximum: /(?:max)([0-9]+)/i,
-    keep: /k([hl])?([0-9]+)?/i
+    dropKeep: /[dk]([hl])?([0-9]+)?/i,
+    count: /(?:c[sf])([<>=]+)?([0-9]+)?/i
   };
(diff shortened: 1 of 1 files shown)


module/utils.mjs
```diff
@@ -49,7 +49,8 @@ export function isValidDieModifier(mod) {
     reroll: /rr?([0-9]+)?([<>=]+)?([0-9]+)?/i,
     explode: /xo?([0-9]+)?([<>=]+)?([0-9]+)?/i,
     minimum: /(?:min)([0-9]+)/i,
-    maximum: /(?:max)([0-9]+)/i
+    maximum: /(?:max)([0-9]+)/i,
+    keep: /k([hl])?([0-9]+)?/i
```
Fyorl, line 53: Are we missing 'drop lowest' modifiers as well in that case?

Fyorl (changes requested):

module/utils.mjs
krbz999: `unsoluble-discord-avatar.jpg` 😩

krbz999: I added drop as well as count (for "cs" and "cf"). That should be more than what is needed for most standard games of dnd. It would be handy if these regular expressions were exposed directly somewhere in the Roll or Term classes.

Fyorl: Well, core doesn't really match individual modifiers with regular expressions like this.

Fyorl (approved):

PR #3330, opened 2024-03-23
[#3321] Allow spell scrolls to be created at increased level

### LINKED ISSUE #3321: Support creating spell scrolls that are cast at a higher level than base
The Master Scrivener for Order of Scribes allows for crafting spells scrolls that count as one level higher than normal.

arbron:
The `SpellScrollConfiguration` object now has a `level` parameter that causes the created spell scroll to include a spell at a higher level than normal. This stores the increase level information in `flags.dnd5e.spellLevel`, which contains the original leve, the casting level, and the scaling information.

`Item5e#rollDamage` has been modified to now check for the new flag before scaling damage, and use that information if present. The scaling information is stored in the flag because otherwise it would be lost when converting to a consumable type.

Closes #3321

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1615,8 +1615,8 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     let scaling = this.system.scaling;
     const levelingFlag = this.getFlag("dnd5e", "spellLevel");
     if ( !spellLevel && levelingFlag ) {
-      spellLevel = parseInt(levelingFlag.value);
-      originalLevel = parseInt(levelingFlag.base);
+      spellLevel = levelingFlag.value;
+      originalLevel = levelingFlag.base;
       scaling = levelingFlag.scaling;
     }
 
@@ -2701,11 +2701,11 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     const itemData = (spell instanceof Item5e) ? spell.toObject() : spell;
     if ( Number.isNumeric(config.level) ) {
       flags.dnd5e = { spellLevel: {
-        value: parseInt(config.level),
+        value: config.level,
         base: spell.system.level,
         scaling: spell.system.scaling
       } };
-      itemData.system.level = parseInt(config.level);
+      itemData.system.level = config.level;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


krbz999:
A "base level" such as this would be good to have for the Tiefling's Hellish Rebuke as well as any magic items that grant spells cast at a level higher than the base. It would be fitting to have this on the Spell data model in general.

arbron:
> A "base level" such as this would be good to have for the Tiefling's Hellish Rebuke as well as any magic items that grant spells cast at a level higher than the base. It would be fitting to have this on the Spell data model in general.

Probably true, but since created spells scrolls are consumables they don't have level at all so I'll have to do it with flags.

module/documents/item.mjs
```diff
@@ -1609,6 +1609,16 @@ export default class Item5e extends SystemDocumentMixin(Item) {
   async rollDamage({critical, event=null, spellLevel=null, versatile=false, options={}}={}) {
     if ( !this.hasDamage ) throw new Error("You may not make a Damage Roll with this Item.");
 
+    // Fetch level from tags if not specified
+    let originalLevel = this.system.level;
+    let scaling = this.system.scaling;
+    const levelingFlag = this.getFlag("dnd5e", "spellLevel");
+    if ( !spellLevel && levelingFlag ) {
+      spellLevel = parseInt(levelingFlag.value);
+      originalLevel = parseInt(levelingFlag.base);
```
Fyorl, line 1618: Do we need to be lenient with these flag types? Is there any reason they wouldn't necessarily be the correct type already?

```suggestion
      spellLevel = levelingFlag.value;
      originalLevel = levelingFlag.base;
```

module/documents/item.mjs
```diff
@@ -2637,6 +2647,15 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
 
     // Get spell data
+    const flags = {};
+    if ( Number.isNumeric(config.level) ) {
+      flags.dnd5e = { spellLevel: {
+        value: parseInt(config.level),
+        base: spell.system.level,
+        scaling: spell.system.scaling
+      } };
+      spell.system.level = parseInt(config.level);
```
Fyorl, line 2657: `config.level` is annotated as `number`

```suggestion
      flags.dnd5e = { spellLevel: {
        value: config.level,
        base: spell.system.level,
        scaling: spell.system.scaling
      } };
      spell.system.level = config.level;
```

Fyorl (changes requested):
Just a couple of comments. Maybe we're just being defensive here, but it's possible to pass any garbage for any value of `SpellScrollConfiguration` that `parseInt` won't necessarily protect against, so I think we should just let it fail if bad input is provided.

Fyorl (approved):

PR #3337, opened 2024-03-25
Spell sorting

absymylyar:
Add spell sort within each levels, first by name, then by "always prepared" status and finally by "prepared" status


arbron:
New sheets already have the option of switching between alphabetic sorting and manual sorting, and this change makes it impossible to use manual sorting on the legacy sheets, so unless you can give a good reason to take away that ability I'll probably say this won't be changed.

absymylyar:
Maybe this kind of sorting can be proposed as an additional automatic sorting (or in place of alphabetic sorting) to highlight spells by priority.
This diplay is far better
2024-03-26_11h28_00

Fyorl:
We can add a new sorting method, but it will need the associated UI to go along with it.

absymylyar:
I add this sorting as an option

Fyorl:
Seems like this got rebased on top of one of your other branches rather than the `3.2.x` or `3.1.x` branch? Could you rebase it onto `3.2.x` so that this PR only contains the changes for spell sorting?

PR #3344, opened 2024-03-25
[#1856, #3342] Allow UUIDs to be used as consumption targets

### LINKED ISSUE #1856: Allow item consumption targets to be configurable on non-embedded items
Currently, item consumption targets can only be configured when embedded in an actor. This is an understandable limitation, but means that certain items need a note specifically instructing users to update them once added to their sheet, such as the various Ki features on monks that should be able to specify a shared consumption target for the `"Ki"` item.

- `ammo`, `material`, `charges`: ~~These are all based off other items on the actor. I propose allowing these to be set up using item identifiers that would be auto-matched when used. So a `"Stunning Strike"` item could be configured with `ki` as its consumption target, and then when added to the actor it would be linked to the `"Ki"` item that shares that identifier.~~ See https://github.com/foundryvtt/dnd5e/issues/3342
- `attribute`: This can be configured by entering the consumption key path just like in active effects, which is what is stored under the surface anyway.
- `hitDice`: This doesn't reference any actor data, so there is not reason this shouldn't work now.

### LINKED ISSUE #3342: Allow configuring UUID for Item Uses resource consumption type
Currently there is no way to pre-configure items that consume another item's uses, requiring additional configuration for various class features when they are added. For example, to get a monk properly working the player must open each of the monk's ki abilities and set them to consume uses of their "Ki" item. This is a pain and might catch new players unaware when their features don't track uses properly.

To solve this, we could allow setting the UUID in the consumption target. This would only be settable on non-owned items, items on the actor would get the same dropdown they have now. During data preparation for the item on the actor, we can check to see if its target is a UUID and if so, search through the other items on the actor for one whose `sourceId` matches the UUID. The UUID then will be replaced in memory with that item's ID and everything else should continue working as normally.

Related to #1856

arbron:
Items with the consumption modes of ammo, item uses, and resource will now be display their target as a text input field which accepts an item UUID. When the item is then added to an actor, it will search for an item on the actor with a matching `sourceId` and set that item's ID as its consumption target. This should allow for pre-configuring consumption in compendium items.

Closes #1856 
Closes #3342

### DIFF
--- a/module/data/item/templates/activated-effect.mjs
+++ b/module/data/item/templates/activated-effect.mjs
@@ -108,6 +108,21 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
   }
 
   /* -------------------------------------------- */
+
+  /**
+   * Prepare activated effect data.
+   */
+  prepareActivatedEffectData() {
+    if ( !this.parent.isEmbedded ) return;
+    if ( ["ammo", "charges", "material"].includes(this.consume.type) && this.consume.target.includes(".") ) {
+      const item = this.parent.actor.items.find(i =>
+        (i.flags.core?.sourceId === this.consume.target) || (i.flags.dnd5e?.sourceId === this.consume.target)
+      );
+      if ( item ) this.consume.target = item.id;
+    }
+  }
+
+  /* -------------------------------------------- */
   /*  Migrations                                  */
   /* -------------------------------------------- */
 
--- a/module/data/item/templates/activated-effect.mjs
+++ b/module/data/item/templates/activated-effect.mjs
@@ -110,14 +110,12 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
   /* -------------------------------------------- */
 
   /**
-   * Prepare activated effect data.
+   * Prepare activated effect data, should be called during `prepareFinalData` stage.
    */
-  prepareActivatedEffectData() {
+  prepareFinalActivatedEffectData() {
     if ( !this.parent.isEmbedded ) return;
     if ( ["ammo", "charges", "material"].includes(this.consume.type) && this.consume.target.includes(".") ) {
-      const item = this.parent.actor.items.find(i =>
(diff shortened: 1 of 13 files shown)


module/data/item/templates/activated-effect.mjs
```diff
@@ -107,6 +107,21 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
...
+
+  /**
+   * Prepare activated effect data.
+   */
+  prepareActivatedEffectData() {
+    if ( !this.parent.isEmbedded ) return;
+    if ( ["ammo", "charges", "material"].includes(this.consume.type) && this.consume.target.includes(".") ) {
+      const item = this.parent.actor.items.find(i =>
+        (i.flags.core?.sourceId === this.consume.target) || (i.flags.dnd5e?.sourceId === this.consume.target)
+      );
```
Fyorl, line 120: I think we need to build up a `sourceId` map when doing the initial `prepareEmbeddedItems` pass, and then perform these lookups in `prepareFinalAttributes` probably. I don't think we can afford to scan over the entire Item list for multiple Items (though I'm sure it's entirely possible that we are inadvertently doing this elsewhere).

Alternatively, we could perform the lookup only once when the Item is added to the Actor for the first time (in `Item5e#_onCreate` maybe) and set the reference at that point. It has the disadvantage that Items can't be added out-of-order, (i.e. you must add Ki before you add Flurry of Blows) but I would lean fairly strongly towards this solution if it doesn't pose any immediate issues (like with Advancement addition/removal/modification).

Fyorl (changes requested):
We could potentially give the UUID box more room. 

image

(Tangentially related, but the limited uses 'maximum' field could do with similar treatment)

A text UUID field is definitely the simplest to implement, but it would be ideal if we could move to a drag-and-drop version at some point. Perhaps it should wait for the v2 item sheet styles.

image

Fyorl (approved):

PR #3348, opened 2024-03-25
Reinstate advancement data cleaning and ensure migrations are run before cleaning.

Fyorl:
Can go in 3.1.2 or 3.2, whichever you have a better feeling for.

This was a change we made in core to allow `DataField`s to recursively apply their own migrations/cleaning, and I never followed-up in the dnd5e repo. These `migrateSource` changes are in v11, it's not just a v12 thing, for the avoidance of doubt.


arbron (approved):
Looks like it works.

We might also want to apply this to `ScaleValueEntryField` in `data/advancement/scale-value.mjs` since that follows a similar pattern.

PR #3349, opened 2024-03-25
[#3345] Add Place Members buttons to group to place into scene

### LINKED ISSUE #3345: Add ability to place group members in scene
A new "Place Members" button should be added to the group actor for GMs that will use the new summoning placement system to place each of the group members quickly in a scene.

arbron:
Includes modifications to the `TokenPlacement` class to fully support multiple token placements. The functionality of that system has been changed slightly so right clicking will now skip the current placement, but not cancel the whole process. This should allow skipping certain group members when placing the group.

There is a current limitation that prevents the already placed tokens from displaying in the scene.

Closes #3345

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -845,6 +845,7 @@
     "one": "Member",
     "other": "Members"
   },
+  "PlaceMembers": "Place Members",
   "Primary": {
     "Remove": "Remove as Primary Party",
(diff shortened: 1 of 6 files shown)


Fyorl (approved):
Some very minor comments, and I pushed a fix for the disappearing previews issue.

PR #3360, opened 2024-03-26
[#3358] Properly handle temporary hit points in `applyDamage`

### LINKED ISSUE #3358: Applying temp HP roll using damage application system damages token
Currently the only way to apply temp HP is to use the "Apply Temporary HP" context menu option. If the damage application system is used with temp HP then it applies as damage. The system should be smart enough to separate out temp HP during calculations and apply it properly.

arbron:
Any damage rolls with the `temphp` type are now not included in the total damage and instead split off. Then, after any damage is applied to current & temporary HP, that temp value is used to set the new temporary HP floor.

The damage application interface has been adjusted to display temp HP separately from other changes. The display of normal damage has also changed to add subtle colors (red for damage, green for healing, blue for temp HP). The signs have also been adjusted, so healing will always have a "+", damage will always have a "-", and temp HP will be unsigned.

These changes should make it more clear what each of the numbers represents, but just in case this also adds tooltips to each number.

Closes #3358

### DIFF
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -546,10 +546,14 @@
       }
     }
 
-    .calculated-damage {
+    .calculated {
       padding-inline-end: 4px;
       font-size: var(--font-size-14);
       font-weight: bold;
+      color: var(--dnd5e-color-application-damage);
+
+      &.healing { color: var(--dnd5e-color-application-healing); }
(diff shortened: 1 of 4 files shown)


Fyorl (approved):
Good changes

PR #3364, opened 2024-03-26
[#3353, #3361, #3362] Add support for NPC hit dice based on HP formula

### LINKED ISSUE #3353: Add a numeric value to DND5E.actorSizes to facilitate size comparison in macros
By adding a numeric value to CONFIG.DND5E.actorSizes (e.g.: order, value) it would facilitate doing size comparison in macro code. Many features are only available when a target is maximum one size larger or have a different behavior depending on the size difference.

### LINKED ISSUE #3361: HitPointAdvancement on NPCs should use NPC hit die size, rather than class's
Both the "Monsters with Classes" rules in the DMG and "Sidekick" rules in Tasha's specify that NPCs with classes roll hit points based on the hit dice defined in their stat block, rather than defined in the class, so `HitPointsAdvancement` should reflect that when on a class on an NPC.

### LINKED ISSUE #3362: Calculate NPC hit dice based on their HP formula or size
Generally a NPC's hit dice quantity and denomination is based off their HP formula, so a creature with `12d8 + 24` hit points would have `12` hit dice of a `d8` denomination. We should calculate that information from the HP formula automatically, and fall back to using the NPC's size to determine the denomination if the formula isn't available.

arbron:
Calculates the hit dice max and denomination based on a NPC's HP formula with a fallback based on their size if necessary. On NPCs `HitPointAdvancement` will now determine the die size based on the NPC's hit die denomination rather than what is specified in the class.

Adjust the `Actor#rollHitDie` method to be able to handle rolling hit dice on NPCs and the resting methods to be able to properly recover NPC hit dice.

Closes #3353
Closes #3361 
Closes #3362

### DIFF
--- a/module/data/actor/npc.mjs
+++ b/module/data/actor/npc.mjs
@@ -8,6 +8,8 @@ import CreatureTemplate from "./templates/creature.mjs";
 import DetailsFields from "./templates/details.mjs";
 import TraitsFields from "./templates/traits.mjs";
 
+const { BooleanField, NumberField, SchemaField, StringField } = foundry.data.fields;
+
 /**
  * System data definition for NPCs.
  *
@@ -16,6 +18,8 @@ import TraitsFields from "./templates/traits.mjs";
  * @property {number} attributes.ac.flat         Flat value used for flat or natural armor calculation.
  * @property {string} attributes.ac.calc         Name of one of the built-in formulas to use.
  * @property {string} attributes.ac.formula      Custom formula to use.
+ * @property {object} attributes.hd
+ * @property {number} attributes.hd.spent        Number of hit dice spent.
  * @property {object} attributes.hp
  * @property {number} attributes.hp.value        Current hit points.
  * @property {number} attributes.hp.max          Maximum allowed HP value.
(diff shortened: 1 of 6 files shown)


module/data/actor/npc.mjs
```diff
@@ -272,6 +285,9 @@ export default class NPCData extends CreatureTemplate {
     AttributesFields.prepareConcentration.call(this, rollData);
     TraitsFields.prepareResistImmune.call(this);
 
+    // Hit Dice
+    this.attributes.hd.value = Math.max(0, this.attributes.hd.max - this.attributes.hd.spent);
```
Fyorl, line 289: Could we consider changing this the other way around? So `hd.value` is the field we store, and `hd.spent` is the derived field? It matches our other `value`/`max` fields better, even if it doesn't match the `class` Item's `hitDice`/`hitDiceUsed`.

Fyorl (changes requested):

module/data/actor/npc.mjs
arbron: The disadvantage of that approach is we have to make a more deliberate effort when the HP formula or class levels is changed. By storing `spent` then the `value` automatically increases alongside `max`.

Fyorl: I see, that sounds like a good improvement then

Fyorl (approved):

PR #3366, opened 2024-03-27
Miscellaneous CSS tweaks

Fyorl:
- Moved object fit styles to common gold-icon class.
 - Fixed necrotic SVG path.
 - Bumped up damage icons in application tray from 11px to 13px.

### DIFF
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -340,8 +340,6 @@
       > img {
         width: 32px;
         height: 32px;
-        object-fit: cover;
-        background-color: var(--dnd5e-color-light-gray);
       }
 
       .name-stacked {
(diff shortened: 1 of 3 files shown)


less/v2/chat.less
```diff
@@ -527,9 +525,16 @@
...
+      display: flex;
+      gap: 4px;
+    }
+
     .change-source {
-      width: 16px;
-      padding: 2px;
+      width: 13px;
+      height: 13px;
+      --icon-size: 13px;
```
Fyorl, line 536: This was as large as I could make it without increasing the height of the overall line, but we could set the `--icon-size: 16px` and keep the button size at 13px which allows the icons to overhang the line without increasing the height of the line itself. I'm not sure myself so it's just 13px for now.

arbron (approved):

PR #3373, opened 2024-03-27
[#1549] Allow NPCs to get spellcasting details from classes

### LINKED ISSUE #1549: Add proper support for spellcasting progression on NPCs
Currently, NPCs have only the most basic support for spellcasting by setting spellcaster level, which simulates a full-caster with normal, leveled progression. There is no support for NPC pact casters, nor is there proper support for half or third casters (you can manually divide the spellcaster level by two or three, but then cantrips don't scale properly).

Rather than just specifying a flat spellcaster level, NPCs should be able to define a progression type like can be added on classes.

arbron:
NPCs will now use classes to calculated spellcasting levels allowing for them to use any spellcasting type. If the existing `spellLevel` is set on the class then that number is added to the spellcaster level calculated from their classes.

Closes #1549

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -581,27 +581,21 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const progression = { slot: 0, pact: 0 };
     const types = {};
 
-    // NPCs don't get spell levels from classes
-    if ( this.type === "npc" ) {
-      progression.slot = this.system.details.spellLevel ?? 0;
-      types.leveled = 1;
-    }
+    // Grab all classes with spellcasting
+    const classes = this.items.filter(cls => {
+      if ( cls.type !== "class" ) return false;
+      const type = cls.spellcasting.type;
+      if ( !type ) return false;
+      types[type] ??= 0;
+      types[type] += 1;
+      return true;
+    });
 
-    else {
-      // Grab all classes with spellcasting
-      const classes = this.items.filter(cls => {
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -581,27 +581,21 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-      });
+    for ( const cls of classes ) this.constructor.computeClassProgression(
+      progression, cls, { actor: this, count: types[cls.spellcasting.type] }
+    );
 
-      for ( const cls of classes ) this.constructor.computeClassProgression(
-        progression, cls, { actor: this, count: types[cls.spellcasting.type] }
-      );
-    }
+    if ( this.type === "npc" ) progression.slot += this.system.details.spellLevel ?? 0;
```
Fyorl, line 598: Maybe this should act as an override rather than an addition?

arbron: Hmm, taking a look back at the rules I'd probably vote the other way around, that any classes with spellcasting completely overwrite the `spellLevel` value. While the DMG "Monsters with Classes" rules don't mention existing spellcasting levels, Tasha's "Spellcaster" sidekick class says it replaces and existing spellcasting on the stat block. If we have class levels take priority over this level then things would work according to the rules by simply adding a class to the NPC, rather than having to explicitly remove the original `spellLevel` value first.

Fyorl: Yeah, I'd agree with that

Fyorl (approved):

PR #3378, opened 2024-03-28
Spell preparation management

absymylyar:
Manage the spell preparation for spell caster classes which need to prepare spells.

- Add a new field for class definition to set the spell preparation limit formula
2024-03-28_19h35_03

- Add a new select for spell items to bind the spell to one of the spell casting class from the character class list
2024-03-28_19h31_39

- now can display the number of prepared spells separatly for each class and manage "opacity" according to the "active" spell caster class
2024-03-28_19h30_20

- warning message when the user try to prepare more spells than the limit and cancel the preparation
2024-03-28_19h30_59


absymylyar:
TODO:
- Bind the spell datas (attack bonus, damages, etc...) according to the bound character class, and not to the active spell caster class (I don't know how to do that)
- Bind new, or dropped, spells directly to the active spell caster class

arbron:
This should definitely be split up to make it more reviewable, I'd say have:
- PR that adds the ability for a spell to be associated with a specific class that fetches the appropriate spellcasting modifier
- PR that adds the ability to specify whether a class is a prepared caster and calculates the total number of spells they can prepare (but not look at how many spells are prepared, so this can be reviewed in parallel with the first PR)
- PR that builds upon the first two to display current prepared spells per-class

These new PRs should be constructed atop the `3.2.x` branch and you should use rebasing rather than merging when new system updates need to be incorporated so that you don't end up with all of these extra commits that aren't relevant.

arbron:
In terms of the content, players shouldn't have to specify a custom preparation formula for each class because all of the standard classes use the same formula: class level * multiplier + spellcasting modifier (min 1). An override should be possible, but the default should be able to handle all of the official classes.

I also don't think we should have a hard limit on the spells that you can prepare, but instead just add a little warning next to the class section. That would be more friendly if you are preparing a new spell you can temporarily go over the list and then unprepare something else to get back down to the proper number.

absymylyar:
Okay, I'm working on it.

Instead of a hard limit, I can put the number in strong and red, but I think that the game masters should be noticed that a character player doesn't have a valid spell preparation, maybe with a toast and/or a chat message.
Does it sound nice ?

absymylyar:
Instead of a formula field, I will just add a boolean checkbox "Prepared spell class" and the limit will be calculated as follow :
Full spell caster = class level + spellcasting modifier
Half spell caster = (class level / 2) to closest upper integer value + spellcasting modifier

I think that those formula should be put in the "config" so that it could be changed quite easily for people who wish to create their own system.

absymylyar:
I think also that the race should have a spellcasting definition on its own which should appear in the spell tab.

absymylyar:
> Races also are starting to do the whole "Choose your spellcasting stat" for most of their spells. Good example would be Hexblood's Hex Magic feature.

I'm willing to take this in charge if this project is still free.

arbron:
> Races also are starting to do the whole "Choose your spellcasting stat" for most of their spells. Good example would be Hexblood's Hex Magic feature.

Since races don't offer whole spellcasting progression, but rather just innate spells, then this should be handled by the ability selection in `ItemGrantAdvancement` and `ItemChoiceAdvancement` which already allows for setting which ability score is associated with a given spell. We should make some changes to allow for more than one ability score to be specified. I can take a look at that because it might require some migration to get the data structures set up properly.

PR #3383, opened 2024-03-29
[#1833] Add level prerequisite to feature items used by item choice

### LINKED ISSUE #1833: Add level to `feat` items
Feature items should be able to get a level assigned to them for things like sorcerer invocations that might have a certain level requirement, or the upcoming level-restricted feats in D&D One. This could be used in the future by the Choose Items advancement (#1401) to restrict what can be chosen based on level.

arbron:
Adds a new `prerequisites.level` field to feature items which indicates which class or character level at which the item can be taken.

This data is then used by `ItemChoiceAdvancement` to only display or support drops from features if their meet the level requirement.

Closes #1833

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1144,7 +1144,7 @@
     "prerequisites": {
       "level": {
         "label": "Required Level",
-        "hint": "Character or class level required to select this feature as part of a choice advancement."
+        "hint": "Character or class level required to select this feature when levelling up."
       }
     }
   }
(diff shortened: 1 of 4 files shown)


lang/en.json
```diff
@@ -1137,6 +1138,17 @@
 "DND5E.PolymorphWildShape": "Wild Shape",
 "DND5E.Portrait": "Portrait",
 "DND5E.Prepared": "Prepared",
+"DND5E.Prerequisites": {
+  "Header": "Feature Prerequisites",
+  "FIELDS": {
+    "prerequisites": {
+      "level": {
+        "label": "Required Level",
+        "hint": "Character or class level required to select this feature as part of a choice advancement."
```
Fyorl, line 1147: I'm not sure how meaningful 'choice advancement' is to a non-power-user DM who is not manually configuring advancements themselves. I know we don't want to run the risk of this being misunderstood as preventing players from adding these feats to their sheet via plain drag-and-drop, so maybe something like:

```suggestion
        "hint": "Character or class level required to select this feature when levelling up."
```

Fyorl (approved):
Looks good, just one comment on wording, thanks

PR #3397, opened 2024-04-02
[#3322] Support individual optional ItemGrant features

### LINKED ISSUE #3322: Support individual optional features in `ItemGrantAdvancement`
Currently the optional control in item grant is an all or nothing endeavor, either everything is option or nothing is. It would be good to be able to configure it so that some features are required while others are optional.

arbron:
Individual items in an `ItemGrant` advancement can now be marked as optional:

Screenshot 2024-04-02 at 12 05 28

During the advancement process, only those features will get a checkbox:

Screenshot 2024-04-02 at 12 05 17

The existing "optional" setting for the whole advancement still exists, and acts like before, but if that is set then the optional checkbox controls whether an individual feature starts checked or not.

Closes #3322


templates/advancement/item-grant-flow.hbs
```diff
@@ -1,17 +1,18 @@
 <form id="{{appId}}" data-level="{{level}}" data-id="{{advancement.id}}" data-type="{{type}}">
     <h3>{{{title}}}</h3>
 
+    {{ log this }}
```
Fyorl, line 4: Errant logging statement

Fyorl (approved):
Let's make sure to update the optional hint to really make it clear how the advancement's optional checkbox functions. I think we have to be a bit verbose to make it clear which 'optional' we're talking about. Something like:

> If the whole advancement is marked optional, players may opt out of any of the following items, otherwise all non-optional items are granted.

PR #3405, opened 2024-04-03
Attack targets iteration

Fyorl:
Some small pieces of iteration on attack chat cards, and chat cards in general.

- Added icon indicators to success and failure styles for roll results and added some high-contrast tweaks to address some colour-blindness concerns.
- If an attack roll has only one target (the majority of cases I'd expect), we set a target value of the target's AC, just like with saving throws. This is controlled by the saving throw challenge visibility, but we could split it out into a separate permission. I think there's an open PR with something along those lines already.
- Placed the attack target results into a tray. I think we need to do more work here but for now this is easy to do and unifies the styles that we seem to have aligned on wrt to the chat card trays.
- Condensed attack targets information.
- Prevented the roll result from showing as a success or failure if it was a crit or fumble for roll types that can't critically succeed or critically fail.
- Added bespoke crit hit and fumble styles to help differentiate between success and failure indicators.

image

image

image

image

### DIFF
--- a/less/dnd5e.less
+++ b/less/dnd5e.less
@@ -33,4 +33,3 @@
 @import "v2/high-contrast/apps.less";
 @import "v2/high-contrast/character.less";
 @import "v2/high-contrast/inventory.less";
-@import "v2/high-contrast/chat.less";
--- a/less/dnd5e.less
+++ b/less/dnd5e.less
@@ -33,3 +33,4 @@
 @import "v2/high-contrast/apps.less";
 @import "v2/high-contrast/character.less";
 @import "v2/high-contrast/inventory.less";
+@import "v2/high-contrast/chat.less";
--- a/less/v2/high-contrast/chat.less
+++ b/less/v2/high-contrast/chat.less
@@ -1,0 +1,3 @@
+.dnd5e-flag-high-contrast {
+  .dice-result .dice-total > i { opacity: unset; }
+}
--- a/less/v2/high-contrast/chat.less
+++ b/less/v2/high-contrast/chat.less
@@ -1,3 +1,0 @@
-.dnd5e-flag-high-contrast {
-  .dice-result .dice-total > i { opacity: unset; }
-}
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -181,12 +181,14 @@ export default class ChatMessage5e extends ChatMessage {
       // Highlight successes and failures
       const total = html.find(".dice-total")[index];
       if ( !total ) continue;
+      // Only attack rolls can crit or fumble.
+      const canCrit = this.getFlag("dnd5e", "roll.type") === "attack";
       if ( d.options.target && displayChallenge ) {
         if ( d20Roll.total >= d.options.target ) total.classList.add("success");
         else total.classList.add("failure");
       }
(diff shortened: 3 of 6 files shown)


less/v2/high-contrast/chat.less
```diff
@@ -0,0 +1,3 @@
+.dnd5e-flag-high-contrast {
+  .dice-result .dice-total > i { opacity: unset; }
+}
```
arbron, line 3: We should also increase the contrast for the AC indicator which is pretty low:

```suggestion
  :is(.chat-popout, #chat-log) .message .dnd5e2.evaluation .target.hit .ac { opacity: .8; }
}
```

module/documents/item.mjs
```diff
@@ -1527,6 +1527,9 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     const elvenAccuracy = (flags.elvenAccuracy
       && CONFIG.DND5E.characterFlags.elvenAccuracy.abilities.includes(this.abilityMod)) || undefined;
 
+    // Targets
+    const targets = this.constructor._formatAttackTargets();
```
arbron, line 1531: I notice that `formatAttackTargets` still uses the actor name rather than the token name. It would be good to use the token name in cases where DMs are using auto-numbering (`Goblin #1` & `Goblin #2`, rather than just `Goblin` & `Goblin`).

arbron (approved):
Looks good, just a few minor comments.

less/dnd5e.less
```diff
@@ -33,3 +33,4 @@
 @import "v2/high-contrast/apps.less";
 @import "v2/high-contrast/character.less";
 @import "v2/high-contrast/inventory.less";
+@import "v2/high-contrast/chat.less";
```
arbron, line 36: This file isn't included in the PR.

module/documents/chat-message.mjs
```diff
@@ -331,19 +353,27 @@ export default class ChatMessage5e extends ChatMessage {
     const attackRoll = this.rolls[0];
     const targets = this.getFlag("dnd5e", "targets");
     if ( !game.user.isGM || !(attackRoll instanceof dnd5e.dice.D20Roll) || !targets?.length ) return;
-    const evaluation = document.createElement("ul");
-    evaluation.classList.add("dnd5e2", "evaluation");
-    evaluation.innerHTML = targets.map(({ name, img, ac, uuid }) => {
+    const tray = document.createElement("div");
+    tray.classList.add("dnd5e2");
+    tray.innerHTML = `
+      <div class="card-tray targets-tray collapsible collapsed">
```
arbron, line 359: With the auto-expanding chat trays, this results in the tray being expanded for a single target in addition to the highlighting of the total result.

arbron (changes requested):

module/documents/chat-message.mjs
Fyorl: Do you think we should make a special exception for the attack targets tray? With the new condensed display this is probably not a problem anymore. It perhaps obviates the need to even show success/failure on the attack roll anymore but it feels reasonable to me to have consistency across attack rolls, ability checks and saving throws now with an appropriate target. What do you think?

Fyorl:
Bumping to 3.3 as too significant a change to make in a patch release.

Issue #3374, opened 2024-03-27
Add new active effect type that can modify an item's details, rather than actor


arbron:
Core to the enchantment system (#3305) is a new active effect type (specified by flag for now, using a proper active effect type in V12) that allows the active effect changes to apply to the item rather than to the containing actor.

- This effect would never be transferred to the actor or offered up in the current effect application UI
- Dropping this active effect onto an item works, but dropping it onto an actor throws an error
- Would be skipped during the normal actor active effect application
- Would be applied during new item active effect application
- Only works for physical items

See also: https://github.com/foundryvtt/foundryvtt/issues/10185

PR #3409, opened 2024-04-04
[#3374] Add enchantment active effect that modifies an item

arbron:
Active Effects created with `flags.dnd5e.type` of `enchantment` will now apply to the item in which they are included, not the actor.

The existing item sheets will now identify fields that have been modified by an enchantment effect and disable them, similar to how field are disabled in actor configs.

Enchantment Effect

### Followup Work
- [ ] Distinguish enchantments from other active effects in an item's effect list
- [ ] Add ability to create enchantments without manually setting the flag

### DIFF
--- a/templates/items/consumable.hbs
+++ b/templates/items/consumable.hbs
@@ -110,8 +110,7 @@
                 {{/if}}
                 {{#each properties}}
                 <label class="checkbox">
-                    <input type="checkbox" name="system.properties.{{ @key }}" {{ checked selected }} {{#if disabled~}}
-                           disabled data-tooltip="{{ localize 'DND5E.Enchantment.Warning.Override' }}"{{/if}}>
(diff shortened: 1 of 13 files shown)


templates/items/consumable.hbs
```diff
@@ -110,7 +110,9 @@
                 {{/if}}
                 {{#each properties}}
                 <label class="checkbox">
-                    <input type="checkbox" name="system.properties.{{@key}}" {{checked selected}}> {{ label }}
+                    <input type="checkbox" name="system.properties.{{ @key }}" {{ checked selected }} {{#if disabled~}}
```
Fyorl, line 113: Does the code in `ItemSheet5e#activateListeners` not do this automatically?

arbron: Indeed

Fyorl (approved):

PR #3413, opened 2024-04-05
[#3412] Allow specifying creature size during summoning

### LINKED ISSUE #3412: Add `traits.size` multi-select to summoning config
Similar to the "Creature Type" multi-select, it would be convenient to also be able to modify the final actor size (and resulting token size) in lieu of creating multiple actors with only differing sizes.

SRD examples of use:
* `Animate Dead` allows targeting a Small or Medium dead creature in order to create a particular undead. Strict reading of the spell would imply that the created creature (skeleton or zombie) is their "normal" creature size, but creating a size small zombie from a small creature seems like a harmless interpretation.
* `Create Undead` follows the same reasoning as above.

Non-SRD example: Artificers can create their cannons as small or tiny creatures.

arbron:
Similar to the recent change adding selection of creature types during summoning, this adds creature size to the options. When a size is selected both the actor details and the token will be adjusted to the appropriate size.

Creature Size - Dialog

Creature Size - Config

Closes #3412

### DIFF
--- a/module/applications/item/summoning-config.mjs
+++ b/module/applications/item/summoning-config.mjs
@@ -52,6 +52,10 @@ export default class SummoningConfig extends DocumentSheet {
       (lhs.name || lhs.document?.name || "").localeCompare(rhs.name || rhs.document?.name || "", game.i18n.lang)
     );
     context.summons = this.document.system.summons;
+    context.creatureSizes = Object.entries(CONFIG.DND5E.actorSizes).reduce((obj, [k, c]) => {
+      obj[k] = { label: c.label, selected: context.summons?.creatureSizes.has(k) ? "selected" : "" };
+      return obj;
+    }, {});
     context.creatureTypes = Object.entries(CONFIG.DND5E.creatureTypes).reduce((obj, [k, c]) => {
       obj[k] = { label: c.label, selected: context.summons?.creatureTypes.has(k) ? "selected" : "" };
       return obj;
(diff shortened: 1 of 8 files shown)


module/applications/item/summoning-config.mjs
```diff
@@ -85,6 +89,7 @@ export default class SummoningConfig extends DocumentSheet {
   /** @inheritDoc */
   _getSubmitData(...args) {
     const data = foundry.utils.expandObject(super._getSubmitData(...args));
+    data.creautreSizes ??= [];
```
Fyorl, line 92:
```suggestion
    data.creatureSizes ??= [];
```

module/data/item/fields/summons-field.mjs
```diff
@@ -252,23 +255,24 @@ export class SummonsData extends foundry.abstract.DataModel {
   /* -------------------------------------------- */
 
   /**
-   * Prepare the updates to apply to the summoned actor.
+   * Prepare the updates to apply to the summoned actor and its token.
    * @param {Actor5e} actor             Actor that will be modified.
    * @param {SummonsProfile} profile    Summoning profile used to summon the actor.
    * @param {SummoningOptions} options  Additional summoning options.
-   * @returns {object}                  Changes that will be applied to the actor & its items.
+   * @returns {{actorChanges: object, tokenChanges, object}}  Changes that will be applied to the actor & its items.
```
Fyorl, line 262:
```suggestion
   * @returns {Promise<{actorChanges: object, tokenChanges: object}>}  Changes that will be applied to the actor, its items, and its token.
```

Fyorl (approved):
Spotted a couple of things, otherwise looks good

Issue #1315, opened 2021-10-01
Use new `Roll.isDeterministic` API to ensure item max uses and skill passive modifiers don't include dice terms


arbron:
The new `isDeterministic` property on Rolls should allow us to improve handling of formulas that should never include dice terms. There are two places we use these formulas now that I can think of:

- **Item Max Uses**: There is currently no checking when an item is saved whether is max uses formula contains rolls or not. Adding this check when an item data is saved should make it clearer to players what is an isn't allowed for max uses.
- **Skill Passive Bonus**: Currently we are limiting these bonuses to strictly numeric values. The ability to add your proficiency or some other attribute to your passive value would be nice. Using `isDeterministic` when skills are configured should allows us to ensure whatever formula is entered doesn't contain any dice while still allowing the player to use math formulas and select attributes.

PR #3414, opened 2024-04-05
[#1315] Allow spells to be linked to a specific class

absymylyar:
Add a field in spell data to store the spell origin.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1356,6 +1356,7 @@
 "DND5E.SpeedConditions": "Speed Conditions",
 "DND5E.SpeedSpecial": "Special Movement",
 "DND5E.SpellAbility": "Spellcasting Ability",
+"DND5E.needToPrepareSpells": "Need to Prepare Spells",
 "DND5E.SpellAbilitySet": "Set as Primary Spellcasting Ability",
 "DND5E.SpellAdd": "Add Spell",
 "DND5E.SpellCantrip": "Cantrip",
@@ -1372,6 +1373,7 @@
 "DND5E.SpellComponents": "Spell Components",
 "DND5E.SpellCreate": "Create Spell",
 "DND5E.SpellDC": "Spell DC",
+"DND5E.spellPreparationLimit": "Prep.",
 "DND5E.SpellDetails": "Spell Details",
 "DND5E.SpellEffects": "Spell Effects",
 "DND5E.SpellHeader": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -1410,6 +1410,7 @@
 "DND5E.SpellPrepAlways": "Always Prepared",
 "DND5E.SpellPreparation": "Spell Preparation",
 "DND5E.SpellPreparationMode": "Spell Preparation Mode",
+"DND5E.BoundSpellOrigin": "Bound spellcasting origin",
 "DND5E.SpellPrepared": "Prepared",
 "DND5E.SpellProgArt": "Artificer",
 "DND5E.SpellProgAvailable": "Available slots",
--- a/lang/en.json
+++ b/lang/en.json
@@ -310,6 +310,8 @@
 "DND5E.Armor": "Armor",
 "DND5E.ArmorClass": "Armor Class",
 "DND5E.ArmorClassEquipment": "Equipped Armor",
+"DND5E.ActiveSpellcastingClass" : "Active spell casting class",
+"DND5E.SpellPreparationLimitWarning" : "<p style=\"text-align: justify;\">The character <strong><em>{name}</em></strong> has an invalid spell preparation of {count}/{limit} for the class \"{class}\"</p>",
 "DND5E.ArmorClassFlat": "Flat",
 "DND5E.ArmorClassUnarmored": "Unarmored",
 "DND5E.ArmorClassNatural": "Natural Armor",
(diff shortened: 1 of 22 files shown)


lang/en.json
```diff
@@ -1410,6 +1410,7 @@
 "DND5E.SpellPrepAlways": "Always Prepared",
 "DND5E.SpellPreparation": "Spell Preparation",
 "DND5E.SpellPreparationMode": "Spell Preparation Mode",
+"DND5E.BoundSpellOrigin": "Bound spellcasting origin",
```
arbron, line 1413: I think it should be something a bit more specific referencing the class:

```suggestion
"DND5E.SpellSourceClass": "Source Class",
```

(also it should be shifted below "DND5E.SpellSlotsConfig" to keep some semblance of alphabetical order)

module/data/item/spell.mjs
```diff
@@ -36,6 +36,7 @@ export default class SpellData extends ItemDataModel.mixin(
         required: true, integer: true, initial: 1, min: 0, label: "DND5E.SpellLevel"
       }),
       school: new foundry.data.fields.StringField({required: true, label: "DND5E.SpellSchool"}),
+      boundOrigin: new foundry.data.fields.StringField({required: false, label: "DND5E.BoundSpellOrigin"}),
```
arbron, line 39: Required defaults to `false`, so no need to set it explicitly here.

```suggestion
      sourceClass: new foundry.data.fields.StringField({label: "DND5E.SpellSourceClass"}),
```

templates/items/spell.hbs
```diff
@@ -104,6 +104,22 @@
                     </select>
                 </div>
             </div>
+            {{!-- Bound Spellcasting Origin --}}
+            {{#if document.parent}}
+            <div class="form-group input-select">
+                <label>{{ localize "DND5E.BoundSpellOrigin" }}</label>
```
arbron, line 110:
```suggestion
                <label>{{ localize "DND5E.SpellSourceClass" }}</label>
```

module/documents/actor/actor.mjs
```diff
@@ -47,6 +47,17 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
   /* -------------------------------------------- */
 
+  /**
+   * Get all origins which have spellcasting ability.
+   * @type {Array<Item5e>}
+   */
+  get spellCastingClasses() {
+    return Object.values(this.classes).filter(({system}) => system.spellcasting
+    && system.spellcasting.progression !== "none" );
+  }
```
arbron, line 57: Rather than returning an array of items, it would probably be more useful to match the structure of the `classes` method and return a similar object:

```suggestion
  /**
   * Get all classes which have spellcasting ability.
   * @type {Record<string, Item5e>}
   */
  get spellcastingClasses() {
    return Object.entries(this.classes).reduce((obj, [identifier, cls]) => {
      if ( cls.spellcasting && (cls.spellcasting.progression !== "none") ) obj[identifier] = cls;
      return obj;
    }, {});
  }
```

This structure would allow the value to be read directly in the template rather than having to undergo additional processing.

We should also make sure we refer to `item.spellcasting` rather than `item.system.spellcasting`, because that allows us to take subclass spellcasting into account also.

templates/items/spell.hbs
```diff
@@ -104,6 +104,22 @@
                     </select>
                 </div>
             </div>
+            {{!-- Bound Spellcasting Origin --}}
```
arbron, line 107:
```suggestion

            {{!-- Source Class --}}
```

templates/items/spell.hbs
```diff
@@ -104,6 +104,22 @@
...
+            {{#if document.parent}}
```
arbron, line 108: We have `isEmbedded` here which is what we generally use:

```suggestion
            {{#if isEmbedded}}
```

templates/items/spell.hbs
```diff
@@ -104,6 +104,22 @@
...
+                <label>{{ localize "DND5E.BoundSpellOrigin" }}</label>
+                <div class="form-fields">
+                    <select name="system.boundOrigin">
+                        {{#select system.boundOrigin}}
+                        <option value=""></option>
+                        {{#each document.availableSpellCastingOrigins as |scs|}}
+                        <option value="{{scs.value}}">{{scs.name}}</option>
+                        {{/each}}
+                        {{/select}}
+                    </select>
```
arbron, line 119: Taking advantage of the suggested changes to `Actor5e#spellCastingClasses` allows us to simplify this:

```suggestion
                    <select name="system.sourceClass">
                        {{selectOptions document.parent.spellcastingClasses selected=system.sourceClass
                                        labelAttr="name" blank=""}}
                    </select>
```

module/documents/item.mjs
```diff
@@ -344,6 +344,17 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
+   * Get available spellcasting origins.
+   * @type {Array<Item5e>|null}
+   */
+  get availableSpellCastingOrigins() {
+    if (this.type !== "spell" | !this.parent) return;
+    return this.parent.spellCastingClasses.map(({name, system}) => ({name, value: system.identifier}));
+  }
+
+  /* -------------------------------------------- */
+
```
arbron, line 357: Won't be needing this if the above suggestions are taken.

```suggestion
```

arbron (changes requested):

absymylyar:
I'm sorry for my mistakes, I'm not so used with this technology (I'm C#/Angular dev.), but I learn fast.
Thanks for review

absymylyar:
Sorry, I don't know what happened, but it merged other branch, I try to rebase but I'm failling

PR #3431, opened 2024-04-08
[#1351] Allow spells to be linked to a specific class

### LINKED ISSUE #1351: Assign Spells by Class
Originally in GitLab by @danz5000

Might it be possible to assign spells to class, and thus assign spell attacks and DCs according to the class? That way we could organize how many spells we have according to which class, which class it's assigned to when multiclassing, and other quality of life improvements.

It would also allow us to differentiate having a Bloodwell Vial (applies only to sorcerer spells) vs. a Moon Sickle (applies only to druid spells) vs. a Wand of the Warmage (applies to all spell attacks).

absymylyar:
Closes #1351

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1410,6 +1410,7 @@
 "DND5E.SpellPrepAlways": "Always Prepared",
 "DND5E.SpellPreparation": "Spell Preparation",
 "DND5E.SpellPreparationMode": "Spell Preparation Mode",
+"DND5E.BoundSpellOrigin": "Bound spellcasting origin",
 "DND5E.SpellPrepared": "Prepared",
 "DND5E.SpellProgArt": "Artificer",
 "DND5E.SpellProgAvailable": "Available slots",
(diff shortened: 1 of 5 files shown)


arbron (approved):
I'm going to tentatively mark this in for `3.3.x` because I think we want to be able to handle it with the followup changes and not leave it partially implemented, but I'm not sure if we'll have time to review all the changes for `3.2.x`. I'll keep glances at the other PRs in this set so we can have everything in as good a state as possible before then.

absymylyar:
That's fine, it will give me time to test those changes with my players.
By the way, is there a Discord ?

arbron:
> That's fine, it will give me time to test those changes with my players. By the way, is there a Discord ?

The official Foundry Discord, you can find a link on the Foundry site.

arbron:
@absymylyar Could you pop over to the issue (https://github.com/foundryvtt/dnd5e/issues/1351) and give it a comment so I can assign it to you?

absymylyar:
Done

PR #3433, opened 2024-04-08
[#3300] Add setting to allow players to use Damage Application methods

### LINKED ISSUE #3300: Players cannot see or use the new damage application card
As it says in the title, players cant see the damage application card, making applying damage to their own tokens more difficult. it would be nice to have it as an option to see it and to apply damage to targeted tokens. this would reduce the burden on the dm and give some more control to players
image
player left, gm right

### LINKED ISSUE #3432: Allow for players to apply the correct damage to their own characters as easily as the GM can apply it
It is nonsensical and only leads to confusion that both players and GMs have an option to apply damage, but only the GM-exclusive method applies it correctly.

The context menu options entirely ignore damage traits meaning users are going to get confused why damage is sometimes correctly applied and sometimes (most often) not.

I see no reason that players should be excluded from having the ability to manage their own hit points more easily.

krbz999:
Adds a setting that lets a GM toggle whether players can also use the damage application options.

Intentionally enabled by default.

On the player side, this will only allow for Selected Tokens to have damage applied. The GM side is unchanged.

Closes #3300.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -6872,6 +6872,10 @@
       "Hint": "Make critical hits more deadly by multiplying non-dice modifiers in addition to rolled dice."
     }
   },
+  "DAMAGE": {
+    "Hint": "Allow players to use the damage application method in damage rolls.",
+    "Name": "Player Damage Application"
+  },
   "DEFAULTSKILLS": {
     "Name": "Default Skills",
     "Hint": "The default skills that appear on NPC sheets regardless of level of proficiency."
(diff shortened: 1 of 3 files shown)


Padhiver:
3.3 Milestone ? That would be a awesome option.

Fyorl:
Same as #5998 I think I want to land chat card changes before looking at this because we can potentially improve the UX. Please remind me if it looks like 6.0 is going to ship and we've forgotten this.

Fyorl:
Manually merged in 69996c658a83a1d21220f297d8cb9f8233074636

PR #3435, opened 2024-04-08
[#3418] Add enchantment configuration & restrictions

### LINKED ISSUE #3418: Add configuration & restrictions on enchantments
A data structure will need to be built up to store information for enchantments (see #3305) and how they relate to items:

### Enchantment Itself
- How many items an enchantment can be bound to, if it is limited
- What item type it can be applied to (e.g. weapon, equipment, consumable, loot)
  - Initially based just on item's primary type
  - Future work can include additional restrictions (e.g. requires "Ammunition" property, must be worth at least 100 gp)

### Enchantment Source
(this is an item that provides information about a group of enchantments)
- Period in which enchantments can be changed (short rest, long rest, never, at will, etc)
- How many total enchantments of this type can be in use at a time (e.g. you can only infuse 2 items at a time)
  - This must accept scale values and other formulas so it can increase with time

arbron:
Adds a new `EnchantmentData` embedded data model onto `feat` & `spell` items that contains configuration data for enchantments. This includes a number of helper methods to determine if an item represents an enchantment or enchantment source.

Enchantment Config

For enchantments, restrictions are added for whether the enchantment can be applied to magical items and what types of items it can apply to. There is also a `canEnchant` method that takes an item and determines whether it meets the required criteria.

Enchantment Source Configuration

For enchantment sources there is a formula for the maximum number of enchantments in effect at a time and the period at which enchanted items can be re-bound.

Closes #3418

### DIFF
--- a/module/data/item/fields/enchantment-field.mjs
+++ b/module/data/item/fields/enchantment-field.mjs
@@ -1,0 +1,135 @@
+import { FormulaField } from "../../fields.mjs";
+
+const { BooleanField, EmbeddedDataField, SchemaField, StringField } = foundry.data.fields;
+
+/**
+ * A field for storing summons data.
+ */
+export default class EnchantmentField extends EmbeddedDataField {
+  /**
+   * Construct an enchantment field.
+   * @param {object} [options={}]  Options to configure this field's behavior.
+   */
+  constructor(options={}) {
+    super(EnchantmentData, foundry.utils.mergeObject({ required: false, nullable: true, initial: null }, options));
+  }
+}
+
+/**
+ * Data model for enchantment configuration.
+ *
+ * @property {object} items
+ * @property {string} items.max                   Maximum number of items that can have this enchantment.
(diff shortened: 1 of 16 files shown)


module/data/item/fields/enchantment-field.mjs
```diff
@@ -0,0 +1,135 @@
...
+    super(EnchantmentData, foundry.utils.mergeObject({ required: false, nullable: true, initial: null }, options));
+  }
+}
+
+/**
+ * Data model for enchantment configuration.
+ *
+ * @property {object} items
+ * @property {string} items.max                   Maximum number of items that can have this enchantment.
+ * @property {string} items.period                Frequently at which the enchantment be swapped.
```
Fyorl, line 23:
```suggestion
 * @property {string} items.period                Frequency at which the enchantment be swapped.
```

Fyorl (approved):

PR #3436, opened 2024-04-09
Add Icons to Class and Subclass Journal Pages

absymylyar:


arbron:
Since these journal pages are designed to mirror the layout of the books, I don't think adding icons is desired.

Issue #3242, opened 2024-03-12
Add summoning features to support different summoning types


arbron:
Continuing the work started with #893, this is a list of additional features the summoning system will require to support official summoning spells & features.

## Features Required
- [ ] 1. Summoning based on CR & creature type, rather than explicit UUIDs
- [ ] 2. Summon multiple creatures, increase summoned quantity based on spell level
- [ ] 3. Increase summoned CR based on spell level
- [x] 4. Modify creature type (with choice) (https://github.com/foundryvtt/dnd5e/pull/3372)
- [ ] 5. Modify other stats
- [x] 6. Bonuses to damage formulas (https://github.com/foundryvtt/dnd5e/pull/3244)
- [ ] 7. Profiles restricted by casting level, character, or class level
- [ ] 8. Reference summoned creature's stats in AC & HP bonuses

### SRD/Player's Handbook
- Animate Objects (1, 2)
- Arcane Eye
- Arcane Sword
- Conjure Animals (1, 2)
- Conjure Celestial (1, 3)
- Conjure Elemental (1, 3)
- Conjure Fey (1, 3)
- Conjure Minor Elementals (1, 2)
- Conjure Woodland Beings (1, 2)
- Create Undead (7)
- Dancing Lights (1)
- Find Familiar (4)
- Find Steed (4, 5)
- Giant Insect
- Guardian of Faith
- Mage Hand
- Unseen Servant

### Book of Many Things
- Spirit of Death (6)

### Fizban's Treasury of Dragons
- Summon Draconic Spirit (6)

### Icewind Dale: Rime of the Frostmaiden
- Blade of Disaster

### Tasha's Cauldron of Everything
- Summon Aberration (6)
- Summon Beast (6)
- Summon Celestial (6)
- Summon Construct (6)
- Summon Elemental (6)
- Summon Fey (6)
- Summon Fiend (6)
- Summon Shadowspan (6)
- Summon Undead (6)

### Xanathar's Guide to Everything
- Find Greater Steed (4, 5)
- Healing Spirit (5)
- Infernal Calling* (1, 3)
- Summon Greater Demon* (1, 3)
- Summon Lesser Demons* (1, 2)

* Since these are under the GM control rather than the player, perhaps they don't need direct support or perhaps there is an option in summoning config for who can perform the summoning

### Artificer
- Homunculus Servant
- [Artillerist] Eldritch Cannon (7)
- [Battle Smith] Steel Defender (7)

### Bard
[...]

PR #3439, opened 2024-04-09
[#3242] Add level limits to summoning profiles

arbron:
Adds a minimum and maximum level to each summoning profile hidden beneath an "Additional Settings" tray. These values determine whether the profile will be visible for a given spell, character, or class level.

Summoning Level Limit Config

For summoning on a spell item, it will always use the spell's level. On all other items it uses the character level by default unless a class identifier is entered, in which case it will use that class's level.

Summoning Level Limit Identifier

The ability use dialog has been updated to dynamically refresh the list of summoning profiles as the casting level is changed.

### DIFF
--- a/module/applications/item/ability-use-dialog.mjs
+++ b/module/applications/item/ability-use-dialog.mjs
@@ -185,12 +185,28 @@ export default class AbilityUseDialog extends Dialog {
     const summons = item.system.summons;
     if ( !summons?.profiles.length ) return null;
     const options = {};
-    if ( summons.profiles.length > 1 ) options.profiles = summons.profiles.reduce((obj, profile) => {
-      const doc = profile.uuid ? fromUuidSync(profile.uuid) : null;
-      if ( !profile.uuid || doc ) obj[profile._id] = profile.name ? profile.name : (doc?.name ?? "—");
-      return obj;
-    }, {});
-    else options.profile = summons.profiles[0]._id;
+    const rollData = item.getRollData();
+    const keyPath = item.type === "spell"
+      ? "item.level"
+      : summons.classIdentifier
+        ? `classes.${summons.classIdentifier}.levels`
+        : "details.level";
+    const level = foundry.utils.getProperty(rollData, keyPath) ?? 0;
+    options.profiles = Object.fromEntries(
+      summons.profiles
+        .map(profile => {
+          const doc = profile.uuid ? fromUuidSync(profile.uuid) : null;
+          const withinRange = ((profile.level.min ?? -Infinity) <= level) && (level <= (profile.level.max ?? Infinity));
+          if ( !doc || !withinRange ) return null;
+          const label = profile.name ? profile.name : (doc?.name ?? "—");
+          return [profile._id, label];
+        })
+        .filter(f => f)
+    );
+    if ( Object.values(options.profiles).length <= 1 ) {
+      options.profiles = null;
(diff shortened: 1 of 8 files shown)


module/applications/item/ability-use-dialog.mjs
```diff
@@ -185,12 +185,28 @@ export default class AbilityUseDialog extends Dialog {
...
+        ? `classes.${summons.classIdentifier}.levels`
+        : "details.level";
+    const level = foundry.utils.getProperty(rollData, keyPath) ?? 0;
+    options.profiles = Object.fromEntries(
+      summons.profiles
+        .map(profile => {
+          const doc = profile.uuid ? fromUuidSync(profile.uuid) : null;
+          const withinRange = ((profile.level.min ?? -Infinity) <= level) && (level <= (profile.level.max ?? Infinity));
+          if ( !doc || !withinRange ) return null;
+          let label = profile.name ? profile.name : (doc?.name ?? "—");
```
Fyorl, line 201:
```suggestion
          const label = profile.name ? profile.name : (doc?.name ?? "—");
```

module/applications/item/ability-use-dialog.mjs
```diff
@@ -410,4 +426,54 @@ export default class AbilityUseDialog extends Dialog {
...
+   */
+  _onChangeSlotLevel(event) {
+    const level = parseInt(event.target.value.replace("spell", ""));
+    const item = this.item.clone({ "system.level": level });
+    const summoningData = this.constructor._createSummoningOptions(item);
+    const originalInput = this.element[0].querySelector('[name="summonsProfile"]');
+    if ( !originalInput ) return;
+
+    // If only one profile, replace with hidden input
+    if ( !summoningData.profiles ) {
```
Fyorl, line 456: Prefer to reverse these branches so that the `else` condition doesn't read like a double-negative

templates/apps/summoning-config.hbs
```diff
@@ -28,12 +28,42 @@
...
+                <div class="collapsible-content">
+                    <div class="wrapper">
+                        <div class="form-group">
+                            <label>{{ localize "DND5E.Summoning.Level.Label" }}</label>
+                            <div class="form-fields">
+                                <input type="number" name="profiles.{{ id }}.level.min" min=1 step=1 placeholder="0"
+                                       value="{{ level.min }}" aria-label="{{ localize 'DND5E.Summoning.Level.Min' }}">
+                                <span class="sep">&ndash;</span>
+                                <input type="number" name="profiles.{{ id }}.level.max" min=1 step=1 placeholder="∞"
+                                       value="{{ level.max }}" aria-label="{{ localize 'DND5E.Summoning.Level.Max' }}">
```
Fyorl, line 46: Technically valid (or at least, accepted) HTML, but not a style we use I don't think

```suggestion
                                <input type="number" name="profiles.{{ id }}.level.min" min="1" step="1" placeholder="0"
                                       value="{{ level.min }}" aria-label="{{ localize 'DND5E.Summoning.Level.Min' }}">
                                <span class="sep">&ndash;</span>
                                <input type="number" name="profiles.{{ id }}.level.max" min="1" step="1" placeholder="∞"
                                       value="{{ level.max }}" aria-label="{{ localize 'DND5E.Summoning.Level.Max' }}">
```

Fyorl (approved):
Good stuff, thank you Jeff

PR #3458, opened 2024-04-15
Literally remove almost every instance of the word 'pact'

### LINKED ISSUE #3454: The images used for spell levels (in favorite section) are hardcoded
This cannot be changed via a script since it always points to a folder in `systems/dnd5e`. Since there is support for spell levels above 9, this should be supported as well.

Suggest simply adding it to the config.

### LINKED ISSUE #3455: The label (SR or LR) in the Favorites section is hardcoded
This is a minor inconvenience for new spell slot progressions that wish to restore slots on a short rest. There is only a special case exception for "pact" slots; this should be the other way around with leveled slots being treated as the exception.

Suggest also simply adding `shortRest: true` somewhere in the config.

### LINKED ISSUE #3456: Generalize the `preparePactSlots` method
By changing just a few variable names, this method can be used more generally for any progression beyond just pact slots.

Suggest keeping a convenience method for pact slots specifically that calls this more general method "prepareAltSlots".

### LINKED ISSUE #3457: Dropping a spell onto an actor sheet should sort the spell into its pre-set preparation mode if it matches with a preparation mode that is present on the actor

krbz999:
Closes #3454.
Closes #3455.
Closes #3456.
Closes #3457.


module/documents/actor/actor.mjs
```diff
@@ -2504,25 +2517,29 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-      for ( let [k, v] of Object.entries(spells) ) {
-        updates[`system.spells.${k}.value`] = Number.isNumeric(v.override) ? v.override : (v.max ?? 0);
+    if ( !spells ) return updates;
+
+    Object.entries(CONFIG.DND5E.spellPreparationModes).forEach(([k, v]) => {
+      if ( v.upcast && ((recoverShort && v.shortRest) || recoverLong) ) {
+        if ( k === "prepared" ) {
+          Array.fromRange(Object.keys(CONFIG.DND5E.spellLevels).length - 1, 1).forEach(level => {
+            updates[`system.spells.spell${level}.value`] = spells[`spell${level}`].max;
+          });
```
Fyorl, line 2537: Unclear why we prefer to loop over the configured `spellLevels` rather than the `spells` entries. If there is no `spells` entry corresponding to an entry in `spellLevels`, won't the ``spells[`spell${level}`].max`` be undefined?

Fyorl (changes requested):
Just some linting comments from me, otherwise it looks good

module/documents/actor/actor.mjs
krbz999: Other way around actually, if you at any point adjust the spellLevels config object up to, say, 20 levels, then down again later, the actors will all have these properties still in the final data. For example, we ensure throughout the code that `levels` is a property in the `spells` entries to know that this is a valid one added by the system, module, or world script.

It's probably cleaner and simpler to loop throughout the `spells` object for keys of the proper name syntax and ensuring there is a `levels` or `max` property in them, however.

arbron:
Just a thought, but it seems like it would make more sense to specify resting behavior on `spellcastingTypes` rather than `spellPreparationModes`. It's not `prepared` spells that recover on a long rest, but `leveled` spellcasting slots that do.

krbz999:
> > > It seems like it would make more sense to specify resting behavior on `spellcastingTypes` rather than `spellPreparationModes`. It's not `prepared` spells that recover on a long rest, but `leveled` spellcasting slots that do.
> > 
> > 
> > I agree, though it complicates the logic somewhat. Perhaps we should break things and combine the two?
> 
> Well, but `always` and `prepared` both represent `leveled` spells, so it isn't quite a 1-to-1 mapping.

It would be if we got rid of 'always' as a "preparation mode", which honestly is a bit odd. It should be a separate property. A spell that is "always prepared" is a spell that is prepared, always. But maybe in the future.

Fyorl (approved):

PR #3468, opened 2024-04-18
[#1537, #3450] Add `ActivatedEffectTemplate` to tool items

### LINKED ISSUE #1537: Feature request: Add Limited Uses to tools
Originally in GitLab by @krbz999

Several tools, mostly musical instruments, have charges to expend on specific features. It would be good if these item types supported limited uses.

Examples:
- Pipes of Haunting (DMG)
- Horn of Silent Alarm (XGE)
- Instrument of Scribing (XGE)
- Pipes of the Sewers (DMG)

### LINKED ISSUE #3450: Tool items have the 'Concentration' property but cannot actually require concentration as they cannot have a duration.
For concentration to function, tools would need to be able to have a duration as well.

Adding `ActivatedEffectTemplate` to tools would solve this as well as #1537.

krbz999:
Closes #1537 (at long last).
Closes #3450.


Fyorl (approved):
As some follow-up work, I think we need to amend the favourites code slightly to `use` the tool Item rather than roll a tool check if the tool Item has some activated effect.

PR #3479, opened 2024-04-21
Ensure placeMembers includes quantity

### LINKED ISSUE #3478: PlaceMembers for encounters only places one of each type of member
Reproduction:
- Create 'Encounter' group
- Add actor to group
- Set 'quantity' to 2
- Click "Place Members"
- Observe only 1 token of each type is placed.

image

NeilWhite:
Place Members button in encounter/group forms don't take into consideration the quantity of each actor to create tokens for.
Fixes #3478

### DIFF
--- a/module/data/actor/group.mjs
+++ b/module/data/actor/group.mjs
@@ -203,7 +203,10 @@ export default class GroupActor extends ActorDataModel.mixin(CurrencyTemplate) {
 
     try {
       const placements = await TokenPlacement.place({
-        tokens: Object.values(this.members).map(m => m.actor.prototypeToken)
+        tokens: Object.values(this.members).flatMap(({ actor, quantity }) => {
+          if (!(quantity?.value)) return [];
+          return quantity.value === 1 ? actor.prototypeToken : new Array(quantity.value).fill(actor.prototypeToken);
+        })
       });
       for ( const placement of placements ) {
         const actor = placement.prototypeToken.actor;


module/data/actor/group.mjs
```diff
@@ -203,7 +203,10 @@ export default class GroupActor extends ActorDataModel.mixin(CurrencyTemplate) {
 
     try {
       const placements = await TokenPlacement.place({
-        tokens: Object.values(this.members).map(m => m.actor.prototypeToken)
+        tokens: Object.values(this.members).flatMap(({ actor, quantity }) => {
+          if (!(quantity?.value)) return [];
+          return quantity.value === 1 ? actor.prototypeToken : new Array(quantity.value).fill(actor.prototypeToken);
+        })
```
arbron, line 209: I think if `quantity.value` is `null`, then we should treat it as `1` rather than `0`. It should require having an explicit `0` value to prevent summoning. The rest of the group code isn't super consistent about that at the moment (blank quantity skips the actor for total HP, but treats it as one for total XP). To make that work properly every we need to change line `145` of `group-sheet.mjs`:

```javascript
const multiplier = type === "encounter" ? (memberData.quantity.value ?? 1) : 1;
```

We can also simplify the logic here to just always return an array even for a single entry (you also don't need `new` with arrays). It should also check that the type is `encounter` so any leftover quantity data doesn't cause multiple summons in other types:

```suggestion
        tokens: Object.values(this.members).flatMap(({ actor, quantity }) =>
          Array(this.type.value === "encounter" ? (quantity.value ?? 1) : 1).fill(actor.prototypeToken)
        )
```

arbron (changes requested):

module/data/actor/group.mjs
NeilWhite: I thought about defaulting to 1 in a my first try, and talked myself out of it but can't remember how. I like this though

arbron (approved):

PR #3480, opened 2024-04-21
Remove features whose deprecation period ends with `3.2`

arbron:
This extends the deprecation period for `getBaseItem` indexes another 2 versions because fixing this requires a compendium migration and there are still people who haven't done that.

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -642,25 +642,6 @@ DND5E.actorSizes = {
   }
 };
 preLocalize("actorSizes", { keys: ["label", "abbreviation"] });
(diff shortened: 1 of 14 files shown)


module/config.mjs
```diff
@@ -868,19 +848,6 @@ preLocalize("itemRarity");
 
 /* -------------------------------------------- */
 
-/**
- * The limited use periods that support a recovery formula.
- * @deprecated since DnD5e 3.1, available until DnD5e 3.3
```
Fyorl, line 873: Seems like this has one more version to go.

Fyorl (approved):

PR #3485, opened 2024-04-23
[#2773] Display limited spell uses on sheet

### LINKED ISSUE #2773: Spells with limited uses do not show these anywhere on the actor sheet

arbron:
Adds the uses as the first column on the player's spellbook tab. These uses only display when the spellbook is wide enough and are editable like in the features tab.

Spell Uses - Wide

If the sheet it too narrow, then the uses are instead displayed in the spell's subtitle alongside and required components. In the narrow mode they are not editable.

Spell Uses - Narrow

Closes #2773

### DIFF
--- a/templates/actors/tabs/character-spells.hbs
+++ b/templates/actors/tabs/character-spells.hbs
@@ -61,12 +61,12 @@
             {{!-- Section Header --}}
             <div class="items-header header">
                 <h3 class="item-name spell-header">{{ localize label }}</h3>
-                <div class="item-header item-uses spell-uses">{{ localize "DND5E.Uses" }}</div>
                 <div class="item-header item-school">{{ localize "DND5E.SpellHeader.School" }}</div>
                 <div class="item-header item-usage">{{ localize "DND5E.SpellHeader.Time" }}</div>
                 <div class="item-header item-range">{{ localize "DND5E.SpellHeader.Range" }}</div>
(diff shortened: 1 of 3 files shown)


templates/actors/tabs/character-spells.hbs
```diff
@@ -120,6 +128,16 @@
                         </div>
                     </div>
 
+                    {{!-- Spell Uses --}}
+                    <div class="item-detail item-uses spell-uses {{#unless ctx.hasUses}}empty{{/unless}}">
+                        {{#if ctx.hasUses}}
+                        <input type="text" value="{{ item.system.uses.value }}" placeholder="0" data-dtype="Number"
+                               data-name="system.uses.value" inputmode="numeric" pattern="[0-9+=\-]*">
```
Fyorl, line 135: I made a mistake with these delta expressions and have been trying to slowly fix them up wherever I find them

```suggestion
                               data-name="system.uses.value" inputmode="numeric" pattern="[+=\-]?\d*">
```

Fyorl (changes requested):
Thanks for taking care of this. I have a fairly strong preference that the uses column be either the last column or between the roll and formula columns in order to be more visually consistent with the inventory tab. I know the legacy sheet has it appear first on the spells tab, but I believe that was tacked-on a lot later.

Fyorl (approved):

PR #3486, opened 2024-04-23
[#3071] Fix rendering issues with UP & DOWN item tooltips

### LINKED ISSUE #3071: System tooltips render incorrectly with `data-tooltip-direction="UP"` and `="DOWN"`
Edited manually in console to add the dataset properties.

DOWN renders identically to `LEFT`.

UP renders like so:
image

arbron:
Seems to be working for all cases I tested, but I'm not sure if there is some case that the original positioning code was written for that the normal core code doesn't take into account.

Closes #3071

### DIFF
--- a/module/tooltips.mjs
+++ b/module/tooltips.mjs
@@ -176,6 +176,23 @@ export default class Tooltips5e {
    * @protected
    */
   _positionItemTooltip(direction=TooltipManager.TOOLTIP_DIRECTIONS.LEFT) {
+    const pos = this.tooltip.getBoundingClientRect();
+    const dirs = TooltipManager.TOOLTIP_DIRECTIONS;
+    switch ( direction ) {
+      case dirs.UP:
+        if ( pos.y - TooltipManager.TOOLTIP_MARGIN_PX <= 0 ) direction = dirs.DOWN;
+        break;
+      case dirs.DOWN:
+        if ( pos.y + this.tooltip.offsetHeight > window.innerHeight ) direction = dirs.UP;
+        break;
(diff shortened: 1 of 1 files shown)


Fyorl:
If you drag a sheet to the left edge of the window and hover an item or a skill, the tooltip can appear underneath the cursor, when it should appear on the right-hand side of the target instead. That's one of the things handled by the old code, but I thought the core tooltip implementation also handled that case, so perhaps we just need to call a different method or pass some extra information.

arbron:
> If you drag a sheet to the left edge of the window and hover an item or a skill, the tooltip can appear underneath the cursor, when it should appear on the right-hand side of the target instead. That's one of the things handled by the old code, but I thought the core tooltip implementation also handled that case, so perhaps we just need to call a different method or pass some extra information.

I see. I looked through `TooltipManager` and it only seems to auto-select the top or bottom position and only if no other direction is specified. I've expanded that auto-selection code to match what we need.

Fyorl:
The changes fix item tooltips from what I can tell, but the skill tooltips, for example, still appear under the cursor if the sheet is moved to the left edge of the viewport.

arbron:
> The changes fix item tooltips from what I can tell, but the skill tooltips, for example, still appear under the cursor if the sheet is moved to the left edge of the viewport.

Ah yeah, that is because they don't have `data-tooltip-position` set, so they aren't getting the initial calculation to check. I've added an initial call to `_setAnchor` if the `direction` is `undefined` that fixes that.

Fyorl (approved):
Nice, thank you

PR #3491, opened 2024-04-24
[#3452] Don't replace `damage.active` set in `preCalculateDamage`

### LINKED ISSUE #3452: Feature request actor.calcluateDamage
After the call to dnd5e.preCalculateDamage, damage.active is force set to {} / {active: "Immunity", multiplier: 0).
In midi-qol I have occasion to want to configure some of the actives during the preCalculateDamage hook, so that multipliers can be applied prior to damage modification.

Would it be possible to have damage.active initialised to {} prior to the call to dnd5e.preCalculateDamage?

arbron:
Doesn't override a `active` object created during the `preCalculateDamage` hook and incorporates any multiplier into the auto-calculated multiplier.

Closes #3452

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1039,13 +1039,15 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const rollData = this.getRollData({deterministic: true});
 
     damages.forEach(d => {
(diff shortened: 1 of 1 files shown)


tposney:
That looks perfect - many thanks.

PR #3493, opened 2024-04-24
[#3492] Add tooltips to active effects

### LINKED ISSUE #3492: Add tooltips for active effects on the character sheet
Would be a good way to view effect description plus other properties that aren't immediately visible such as what status effects are associated with an effect.

arbron:
Effect Tooltip

Closes #3492

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -623,4 +623,34 @@ export default class ActiveEffect5e extends ActiveEffect {
     const delta = current.symmetricDifference(source);
     for ( const choice of delta ) overrides.push(`${prefix}.${choice}`);
   }
+
+  /* -------------------------------------------- */
+
+  /**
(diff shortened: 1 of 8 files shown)


module/documents/active-effect.mjs
```diff
@@ -629,6 +629,36 @@ export default class ActiveEffect5e extends ActiveEffect {
     for ( const choice of delta ) overrides.push(`${prefix}.${choice}`);
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * Render a rich tooltip for this effect.
+   * @param {EnrichmentOptions} [enrichmentOptions={}]  Options for text enrichment.
+   * @returns {{content: string, classes: string[]}}
```
Fyorl, line 637:
```suggestion
   * @returns {Promise<{content: string, classes: string[]}>}
```

Fyorl (approved):

PR #3499, opened 2024-04-25
[#3498] Add prone separately for sleeping/unconscious

### LINKED ISSUE #3498: When applying the Sleeping or Unconscious conditions, apply 'Prone' separately as its own condition rather than as part of the same effect
However unlikely, it is feasible that a character can be unconscious or sleeping while not being prone for the full duration of that condition.

krbz999:
Closes #3498.

Adds a `riders` property for add-on conditions that should be applied separately.

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2486,7 +2486,7 @@ DND5E.consumableResources = [
  * @property {string} icon            Icon used to represent the condition on the token.
  * @property {string} [reference]     UUID of a journal entry with details on this condition.
  * @property {string} [special]       Set this condition as a special status effect under this name.
- * @property {Set<string>} [riders]   Additional conditions, by id, to apply as part of this condition.
+ * @property {string[]} [riders]      Additional conditions, by id, to apply as part of this condition.
  */
 
 /**
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2483,9 +2483,10 @@ DND5E.consumableResources = [
 
 /**
  * @typedef {object} _StatusEffectConfig5e
- * @property {string} icon         Icon used to represent the condition on the token.
- * @property {string} [reference]  UUID of a journal entry with details on this condition.
- * @property {string} [special]    Set this condition as a special status effect under this name.
+ * @property {string} icon            Icon used to represent the condition on the token.
+ * @property {string} [reference]     UUID of a journal entry with details on this condition.
+ * @property {string} [special]       Set this condition as a special status effect under this name.
+ * @property {Set<string>} [riders]   Additional conditions, by id, to apply as part of this condition.
  */
 
 /**
@@ -2621,7 +2622,8 @@ DND5E.conditionTypes = {
     label: "DND5E.ConUnconscious",
     icon: "systems/dnd5e/icons/svg/statuses/unconscious.svg",
     reference: "Compendium.dnd5e.rules.JournalEntry.w7eitkpD7QQTB6j0.JournalEntryPage.UWw13ISmMxDzmwbd",
-    statuses: ["incapacitated", "prone"]
+    statuses: ["incapacitated"],
+    riders: new Set(["prone"])
   }
 };
 preLocalize("conditionTypes", { key: "label", sort: true });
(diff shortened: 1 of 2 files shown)


module/documents/active-effect.mjs
thatlonelybugbear, line 1: Shouldn't the rider be "renewed" in case it should be reapplied when `existing`?
Like maybe 
```js
if ( existing ) await existing.delete();
```

Fyorl:
Seems a bit niche for the added complexity. Are there any other places we would use this?

krbz999:
> Seems a bit niche for the added complexity. Are there any other places we would use this?

If the question is whether adding this is worth it, I personally find myself adding the prone condition separately, on top of applying Unconscious when hitting zero hp, because I know (but others might not know) that the Prone condition would otherwise (incorrectly) remove itself when someone slaps a healing potion in my face. I think it is a worthwhile QOL feature.

As to why I decided to generalize it in this way? I cannot think of reasons as to why not.

arbron:
Definitely something I found myself wanting when playing last weekend when I knocked my players unconscious about a dozen times :).

module/config.mjs
```diff
@@ -2616,7 +2617,8 @@ DND5E.conditionTypes = {
     label: "DND5E.ConUnconscious",
     icon: "systems/dnd5e/icons/svg/statuses/unconscious.svg",
     reference: "Compendium.dnd5e.rules.JournalEntry.w7eitkpD7QQTB6j0.JournalEntryPage.UWw13ISmMxDzmwbd",
-    statuses: ["incapacitated", "prone"]
+    statuses: ["incapacitated"],
+    riders: new Set(["prone"])
```
Fyorl, line 2621: Let's make `riders` an array also to be consistent with `statuses`. The `DataModel` will handle coercion to a `Set`, and therefore duplicates, when the riders are applied.

module/documents/active-effect.mjs
```diff
@@ -511,8 +511,23 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+    if ( !riders ) return;
+
+    const createRider = async id => {
+      const existing = actor.effects.get(staticID(`dnd5e${id}`));
+      if ( existing ) return;
+      const effect = await ActiveEffect.implementation.fromStatusEffect(id);
+      return ActiveEffect.implementation.create(effect, { parent: actor, keepId: true });
+    };
+
+    Promise.all(Array.from(riders).map(createRider));
```
Fyorl, line 530: This doesn't handle other avenues of condition application, such as via the character sheet or enrichers. It can probably be moved to `ActiveEffect5e#_onCreate` though, I think that should cover all cases.

Fyorl (changes requested):
> Unconscious when hitting zero hp, because I know (but others might not know) that the Prone condition would otherwise (incorrectly) remove itself when someone slaps a healing potion in my face. I think it is a worthwhile QOL feature.

That makes sense

module/documents/active-effect.mjs
krbz999: Shows how often I look at the Effects tab.

Fyorl (changes requested):

module/config.mjs
```diff
@@ -2693,7 +2695,7 @@ DND5E.statusEffects = {
   sleeping: {
     name: "EFFECT.DND5E.StatusSleeping",
     icon: "systems/dnd5e/icons/svg/statuses/sleeping.svg",
-    statuses: ["incapacitated", "prone", "unconscious"]
+    statuses: ["incapacitated", "unconscious"]
```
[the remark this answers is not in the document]
krbz999: I removed it, since it adds Unconscious. I felt it was more correct that only Unconscious adds it (sleeping will still add all its "children's" riders) since there are no explicit rules about Sleeping applying Prone directly.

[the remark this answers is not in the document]
Fyorl: Oh right, I did see the code that recursively adds riders, that makes sense

Fyorl (approved):

Issue #3500, opened 2024-04-26
Allow module sub-types to define whether they appear in the inventory section of character sheets


arbron:
Currently module sub-types aren't able to be displayed on the sheet at all without manual handling in `renderSheet` hooks. Part of the issue is that the list of items that appear in the Inventory tab is hardcoded to just the physical types offered by core. By adding an option to `ItemDataModel#metadata` we can let type declare that they want to be included in that list of inventory entries.

PR #3501, opened 2024-04-26
[#3500] Allow module item types to appear in inventory

arbron:
Adds `inventoryItem` to metadata for item data models that allows them to automatically be listed in the inventory tab, as opposed to the former hard-coded list of types.

### DIFF
--- a/module/data/abstract.mjs
+++ b/module/data/abstract.mjs
@@ -360,13 +360,15 @@ export class ItemDataModel extends SystemDataModel {
 
   /**
    * @typedef {SystemDataModelMetadata} ItemDataModelMetadata
-   * @property {boolean} enchantable  Can this item be modified by enchantment effects?
-   * @property {boolean} singleton    Should only a single item of this type be allowed on an actor?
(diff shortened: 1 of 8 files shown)


module/data/abstract.mjs
```diff
@@ -360,13 +360,15 @@ export class ItemDataModel extends SystemDataModel {
...
-   * @property {boolean} singleton    Should only a single item of this type be allowed on an actor?
+   * @property {boolean} enchantable    Can this item be modified by enchantment effects?
+   * @property {boolean} inventoryItem  Should this item be listed with an actor's inventory?
+   * @property {boolean} singleton      Should only a single item of this type be allowed on an actor?
    */
 
   /** @type {ItemDataModelMetadata} */
   static metadata = Object.freeze(foundry.utils.mergeObject(super.metadata, {
     enchantable: false,
+    inventoryItem: true,
```
Fyorl, line 371: Should this not be `false` by default?

Fyorl (changes requested):

Fyorl (approved):

Issue #3305, opened 2024-03-22
Add support for enchanting items


Fyorl:
### Must Haves
1. Add a new category to `feature` Items called 'Enchantment'. Alternatively add a new `enchantment` Item type.
2. Enchantments should specify whether they can be bound to a limited number of Items or not and, if so, how many.
   * These references to other Items *must* be UUIDs to allow enchanting other creatures' Items.
3. Enchantments should specify whether they can be changed on a short rest, a long rest, or at will.
4. Enchantments should specify which type of Item they can apply to: `weapon`, `equipment`, etc.
5. Add a new 'Enchantment' concept to `ActiveEffect5e`.
   * Implemented via flags until v12 where we switch to `ActiveEffect#system`.
6. Enchantment Items should house an Enchantment ActiveEffect as the blueprint for the ActiveEffects that will be bound to enchanted Items.
7. Allow for dragging-and-dropping this Enchantment ActiveEffect onto another Item to enchant it.
8. The Item sheet for Enchantments should:
   1. Show which Items the Enchantment is bound to.
   2. Allow unbinding from those Items.

### Ideal
1. Other Item sheets should:
   1. Indicate somewhere if the Item has an Enchantment bound to it.
2. Enchantment ActiveEffects should contain a JSON blob that defines a diff that can be applied to the Enchanted Item. The diffs are applied during Item preparation.
3. Using the Enchantment Item posts it to the chat where other users can drop their Items onto it to bind their Item to the Enchantment.

### Nice to Haves
1. When performing a rest, show a dialog allowing the player to change any Enchantments they have bound.
2. Provide a user-friendly configuration interface for configuring common Enchantments instead of having to edit JSON.
3. Enchantments should have a way to configure additional restrictions such as 'only two-handed weapon', or 'only a ranged weapon'.

JPMeehan:
If this is a new feature type, this seems like a good way to structure Artificer Infusions.

arbron:
## Example Usage
### Weapon +1
Applied to: Any Weapon
Changes:
- `name`: append `" +1"`
- `description`: append `<p><em>Whether through demonic blessing, celestial bequeathment, crazed experiment, or skillful craft, this weapon has been enhanced to bring more bloodshed by the bearer.</em></p>\n<p>You have a bonus to attack and damage rolls made with this magic weapon.</p>`
- `system.properties`: add `mgc`
- `system.magicalBonus`: override `1`

### Vicious Weapon
[...]

PR #3506, opened 2024-04-30
[#3305] Enforce enchantment restrictions when dropping onto item

arbron:
When adding an enchantment to an item that accepts enchantments, the restrictions specified in the enchantment source will now be enforced. It will also prevent the creation of an enchantment effect directly on an actor.

Enchanted items in the world will register themselves with a static registry of all enchanted items so the enchantment items can easily determine what items they have enchanted.

Support has been added for items that might offer enchantments or receive them (to support a weapon that can enchant other weapons for example). This uses the `origin` data to track whether the enchantment effect is applied to the item or not. To this end the default origin for new enchantments isn't set like for normal effets. The origin is only set when an enchantment is dragged to another item. This ensures that the enchantment origin always matches the item doing the enchantment, even if that item was duplicated.

### DIFF
--- a/module/data/item/fields/enchantment-field.mjs
+++ b/module/data/item/fields/enchantment-field.mjs
@@ -152,7 +152,7 @@ export class EnchantmentData extends foundry.abstract.DataModel {
    */
   async getItems() {
     return (await Promise.all(
-      Array.from(this.constructor.#appliedEnchantments.get(this.item.uuid) ?? [])
+      Array.from(EnchantmentData.#appliedEnchantments.get(this.item.uuid) ?? [])
         .map(uuid => fromUuid(uuid).then(ae => ae?.parent))
     )).filter(i => i);
   }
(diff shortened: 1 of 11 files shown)


module/data/item/fields/enchantment-field.mjs
```diff
@@ -123,6 +143,59 @@ export class EnchantmentData extends foundry.abstract.DataModel {
...
+   * @param {string} enchanted  UUID of the enchantment to track.
+   */
+  static trackEnchantment(source, enchanted) {
+    if ( enchanted.startsWith("Compendium.") ) return;
+    if ( !this.#appliedEnchantments.has(source) ) {
+      this.#appliedEnchantments.set(source, new Set());
+    }
+    const applied = this.#appliedEnchantments.get(source);
+    applied.add(enchanted);
+  }
```
krbz999, line 185: This is a neat way of doing it that would also serve concentration effects and other buffs in general well.

module/documents/active-effect.mjs
```diff
@@ -333,6 +347,49 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+
+    if ( !this.isAppliedEnchantment ) return;
+
+    // Otherwise validate against the enchantment's restraints on the origin item
+    const origin = await fromUuid(this.origin);
+    const errors = origin?.system.enchantment?.canEnchant(this.parent);
+    this.updateSource({ disabled: false });
+    if ( !errors?.length ) return;
+
+    errors.forEach(err => ui.notifications.error(err.message));
```
Fyorl, line 387: Don't feel great about putting UI notifications inside API code. These should probably be logged to the console, and we should spawn UI notifications at the place the user interaction happens. For enchantments I guess that would be in the Item sheet drop handler, and potentially additional places in the future when we add more ways to enchant Items.

Fyorl (changes requested):
Looks good, thank you Jeff. I noticed some issues around re-rendering the Enchantment Source Item sheet when Items are enchanted, but we can polish that later.

Fyorl (approved):

PR #3509, opened 2024-05-01
Add initializeSource hook

arbron:


Fyorl (approved):
Let's consider if we want a more bespoke hook or API that specifically handles merging of advancement data, but this is probably not any worse than what a module could do with prototype patching.

PR #3513, opened 2024-05-02
[#3305] Add enchantment usage dialog & application tray

arbron:
Introduces enchantment to the `AbilityUseDialog` using the `applyEnchantment` and `enchantmentProfile` options, with similar details to how summoning is handled.

Enchantment Use Dialog

When enchantment is performed it adds a new "Enchant" tray to chat cards. This tray has a place to drop an item which is then enchanted when the apply button is pressed.

Enchantment Tray - Empty
Enchantment Tray - Full

Enchantments are tracked with concentration, so if concentration on the enchanting spell ends then the enchantments will be removed.

### DIFF
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -630,6 +630,11 @@ enchantment-application {
       align-items: center;
       gap: 6px;
 
+      &:not(:last-child) {
+        border-block-end: 1px dotted var(--color-border-light-2);
+        padding-block-end: 4px;
+      }
+
       > img {
         block-size: 32px;
         inline-size: 32px;
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -591,10 +591,44 @@
     }
   }
 
-  .targets + button {
+  .wrapper > button {
     margin: 3px;
     width: calc(100% - 6px);
     margin-block-start: 6px;
+  }
+
+  &.enchantment-tray {
+    .drop-area {
+      display: grid;
+      min-block-size: 40px;
+      margin: 4px;
+      border: 1px dashed var(--dnd5e-color-crimson);
+      border-radius: 4px;
+      padding-inline: 4px;
+      background: var(--dnd5e-color-card);
+
+      p {
+        place-content: center;
+        text-align: center;
+      }
+
+      .preview {
+        display: flex;
+        align-items: center;
+        gap: 6px;
+
+        > img {
+          block-size: 32px;
+          inline-size: 32px;
+          flex: 0 0 32px;
+        }
+        > .name {
+          flex: 1;
+          font-family: var(--dnd5e-font-roboto-slab);
+          font-weight: bold;
+        }
+      }
+    }
   }
 }
 
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -596,40 +596,6 @@
     width: calc(100% - 6px);
     margin-block-start: 6px;
   }
-
-  &.enchantment-tray {
-    .drop-area {
-      display: grid;
-      min-block-size: 40px;
-      margin: 4px;
-      border: 1px dashed var(--dnd5e-color-crimson);
-      border-radius: 4px;
-      padding-inline: 4px;
-      background: var(--dnd5e-color-card);
-
-      p {
-        place-content: center;
-        text-align: center;
-      }
-
-      .preview {
-        display: flex;
-        align-items: center;
-        gap: 6px;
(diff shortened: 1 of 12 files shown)


krbz999:
> Enchantments are tracked with concentration, so if concentration on the enchanting spell ends then the enchantments will be removed.

Are enchantments intended to work only with items that require concentration?

arbron:
> > Enchantments are tracked with concentration, so if concentration on the enchanting spell ends then the enchantments will be removed.
> 
> Are enchantments intended to work only with items that require concentration?

Nope. That just means if you are enchanting from an item with concentration it will link them, but otherwise you can enchant normally. Eventually there will be a way to end non-concentration enchantments from the item that provided them, but that’s not implemented yet

Fyorl (changes requested):
Nice work, this looks good. In terms of UX, I have two suggestions:

1. Consider not using a tray for the enchantment. Since the enchantment is the action, it's the only thing that the Item can do. It's equivalent to the action buttons like attack or summon, so it should occupy the same place on the card as where the buttons would normally be I think.
2. Consider automatically applying the enchantment on-drop rather than having a separate apply button. It's quite hard to accidentally drag-and-drop an Item there, so we can assume it's a very deliberate user action. They might have accidentally grabbed the wrong item though, so we could add something like an X on the right of the dropped Item display to unenchant it and clear the box.

arbron:
@Fyorl I made a change so now the chat card remembers what enchantments were added using it, rather than losing them on re-render and gave them a remove enchantment button. This does result in the chat card becoming potentially longer if multiple items are enchanted with it, so I wonder if it might make sense to move the enchanted items back into a tray and maybe only have the drop area out of it?

less/v2/chat.less
```diff
@@ -605,6 +605,48 @@
...
+
+    p {
+      place-content: center;
+      text-align: center;
+    }
+
+    .preview {
+      display: flex;
+      align-items: center;
+      gap: 6px;
```
Fyorl, line 631: Perhaps some slight separation between the rows:

image

```css
.preview:not(:last-child) {
  border-bottom: 1px dotted var(--color-border-light-2);
  padding-bottom: 4px;
}
```

module/applications/components/enchantment-application.mjs
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Build a list of enchanted items.
+   */
+  async buildItemList() {
+    const enchantedItems = (await Promise.all(
+      (this.chatMessage.getFlag("dnd5e", "enchanted") ?? []).map(uuid => fromUuid(uuid))
```
Fyorl, line 64: Is it necessary to store these on the chat message itself? Each enchantment Item knows which other Items it has enchanted, so perhaps this information can be retrieved from there? I don't think it's particularly important that each chat message remembers which Items it enchanted specifically. It should be ok when posting an enchantment to chat that you see all its enchanted items and can add/remove them still. Maybe not, but for the enchantment use cases I can think of, it seems ok. It would mean we remove the requirement for an active GM to be present to orchestrate this.

Fyorl (changes requested):
> @Fyorl I made a change so now the chat card remembers what enchantments were added using it, rather than losing them on re-render and gave them a remove enchantment button. This does result in the chat card becoming potentially longer if multiple items are enchanted with it, so I wonder if it might make sense to move the enchanted items back into a tray and maybe only have the drop area out of it?

I'm not sure how common it is to have more than one Item enchanted at a time. For the use cases I can think of (Infusions, Runes, Magic Weapon, Holy Weapon, Elemental Weapon, etc.) they can only ever have one Item enchanted at a time. For those, just having the one drop 'slot' on the chat card feels pretty good.

If we think it will be quite common to have multiple enchanted Items then I think your idea of putting them behind a tray but keeping the drop zone outside of it is good. The main issue I'm trying to work around is if the Item also has an effects tray. In those cases you end up with two stacked trays which does not look very good and defeats the design of the tray somewhat. You'd want to merge the trays really in that case but that might be quite complicated to do.

So it depends on how common we think it will be to have multiple Items enchanted, and how common it will be to have an Item with both an enchantment and effects, but I'm leaning towards having the enchantments outside the tray being the simplest and easiest solution for now.

One additional thing, if we do keep the enchantments outside of a tray, then currently they are being rendered below the effects tray, so they need to be moved up.

PR #3518, opened 2024-05-02
[#3517] Add spell level to granted and dependent effects

### LINKED ISSUE #3517: Flag effects applied from the effects tray with spellLevel
Flag active effects that are applied via the tray with the spell's level. As per [#3497],
> Useful for later reference and in line with behavior elsewhere in the system (chat messages, templates, etc.).

with the added advantage of being usable on active effects that do not require concentration. It should respect upcasts, which could probably be done by getting the spell's level from the message dataset.

This would allow the system to reference this information in future if needed (required for smoother handling of some spells such as Aid), and would be accessible by modules and macros.

krbz999:
Closes #3517.

Something similar I feel should be done for enchantments as well.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -2124,9 +2124,10 @@ export default class Item5e extends SystemDocumentMixin(Item) {
           let effect = item.effects.get(li.dataset.effectId);
           if ( !effect ) effect = await fromUuid(li.dataset.uuid);
           const concentration = actor.effects.get(message.getFlag("dnd5e", "use.concentrationId"));
+          const effectData = { "flags.dnd5e.spellLevel": spellLevel };
           for ( const token of canvas.tokens.controlled ) {
             try {
-              await this._applyEffectToToken(effect, token, { concentration, spellLevel });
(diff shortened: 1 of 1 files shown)


module/documents/item.mjs
```diff
@@ -2193,11 +2196,12 @@ export default class Item5e extends SystemDocumentMixin(Item) {
    * @param {object} [options]
    * @param {ActiveEffect5e} [options.concentration]  An optional concentration effect to act as the applied effect's
    *                                                  origin instead.
+   * @param {number} [options.spellLevel]             The level of the spell that was cast.
    * @returns {Promise<ActiveEffect5e|false>}
    * @throws {Error}                                  If the effect could not be applied.
    * @protected
    */
-  static async _applyEffectToToken(effect, token, { concentration }={}) {
+  static async _applyEffectToToken(effect, token, { concentration, spellLevel }={}) {
```
Fyorl, line 2204: Maybe this should just accept a general `data` option to merge with the effect data.

Fyorl (changes requested):

Fyorl (approved):

PR #3520, opened 2024-05-02
[#3512] Allow enchantments to be limited by level

### LINKED ISSUE #3512: Add level restrictions to enchantments
Similar to the levels added to summoning profiles, enchantments should also be able to be restricted by spell/class/character level.

arbron:
Adds `min` and `max` level limits as flags to the enchantment effects that are used to limit when those enchantments appear. This is based off spell level if it is a spell, or either character or class level for other items depending on whether a class identifier is provided.

Enchantment Level Limits

If the "Enchantment Prompt" option is not checked, then the system will take level restrictions into account when selecting which enchantment to provide. This should make it easy to configure a feature like "Enhanced Defense" artificer infusion which swaps enchantments depending on artifier level, without requiring the player to view the ability use dialog everytime they want to use the infusion.

Closes #3512

### DIFF
--- a/module/applications/item/enchantment-config.mjs
+++ b/module/applications/item/enchantment-config.mjs
@@ -24,6 +24,14 @@ export default class EnchantmentConfig extends DocumentSheet {
   /*  Properties                                  */
   /* -------------------------------------------- */
 
+  /**
+   * Expanded states for each enchantment.
+   * @type {Map<string, boolean>}
+   */
+  expandedEnchantments = new Map();
+
+  /* -------------------------------------------- */
+
   /** @inheritDoc */
   get title() {
     return `${game.i18n.localize("DND5E.Enchantment.Configuration")}: ${this.document.name}`;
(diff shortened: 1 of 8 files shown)


module/applications/item/enchantment-config.mjs
```diff
@@ -41,9 +49,13 @@ export default class EnchantmentConfig extends DocumentSheet {
       return obj;
     }, {});
     context.enchantment = this.document.system.enchantment;
-    context.enchantments = this.document.effects.filter(e =>
-      (e.getFlag("dnd5e", "type") === "enchantment") && !e.isAppliedEnchantment
-    );
+    context.enchantments = this.document.effects.reduce((arr, e) => {
+      if ( (e.getFlag("dnd5e", "type") === "enchantment") && !e.isAppliedEnchantment ) arr.push({
+        id: e.id, uuid: e.uuid, flags: e.flags, collapsed: this.expandedEnchantments.get(e._id) ? "" : "collapsed"
+      });
```
Fyorl, line 55:
```suggestion
    context.enchantments = this.document.effects.reduce((arr, { id, uuid, flags }) => {
      if ( (e.getFlag("dnd5e", "type") === "enchantment") && !e.isAppliedEnchantment ) arr.push({
        id, uuid, flags, collapsed: this.expandedEnchantments.get(id) ? "" : "collapsed"
      });
```

Fyorl (approved):

PR #3523, opened 2024-05-02
[#3426] Add simple string interpolation to active effects

### LINKED ISSUE #3426: Add original string placeholder option in `override` active effect
With upcoming work on enchantment (https://github.com/foundryvtt/dnd5e/issues/3305) there is sometimes a need to add text to the beginning of an existing string or around an existing string. For example, for the "Returning Weapon" Artificer infusion you might want to change the name to indicate its infusion while still showing the original weapon type, so "Returning Dagger" or "Returning Handaxe". This currently isn't possible because there is no `prefix` active effect mode that would add to the beginning of a string in the same way `add` appends to the end.

Rather than trying to include a new change mode, which core makes difficult, I propose adding support for a placeholder value that can be used with the `override` mode that would be replaced with the original string. So that infusion could be written `Returning {}` and the text `{}` would be replaced with the name before the active effect was applied.

We should add a restriction to ensure this can only be applied to places we can identify as `StringField` or `HTMLField` using data models, because attempting to infer whether the replacement should happen elsewhere may cause issues with data that just looks like strings. We could also potentially make it more restrictive and require a specific option be set on the field like `interpolation` which must be set to `true` for this to work. That would allow us to further restrict this to only fields that we explicitly specify, preventing it from causing issues on stringy but not textual data.

arbron:
Adds the option when apply and OVERRIDE change to a string value to include the original string value by using `{{}}`. So an effect with the change value of `Enhanced {{}}, +1` might be interpolated as `Enhanced Breastplate, +1` when applied.

Closes #3426

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -126,9 +126,9 @@ export default class ActiveEffect5e extends ActiveEffect {
       return changes;
     }
 
-    else if ( (targetType === "string") && (change.mode === modes.OVERRIDE) && change.value.includes("{{}}") ) {
+    else if ( (targetType === "string") && (change.mode === modes.OVERRIDE) && change.value.includes("{}") ) {
       change = foundry.utils.deepClone(change);
-      change.value = change.value.replace("{{}}", current ?? "");
+      change.value = change.value.replace("{}", current ?? "");
     }
 
     let delta;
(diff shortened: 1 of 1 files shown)


module/documents/active-effect.mjs
```diff
@@ -123,6 +126,11 @@ export default class ActiveEffect5e extends ActiveEffect {
       return changes;
     }
 
+    else if ( (targetType === "string") && (change.mode === modes.OVERRIDE) && change.value.includes("{{}}") ) {
```
Fyorl, line 129: Maybe just `{}` à la python? It's a bit closer to the `{foo}` format used in localisation strings that we might be able to expand later.

Fyorl (changes requested):

Fyorl (approved):

PR #3529, opened 2024-05-03
[#3511] Add additional effects that ride along with enchantments

### LINKED ISSUE #3511: Add secondary effects that are added alongside enchantments
Enchantments themselves can modify the item to which they are added, but have no mechanism to modify the actor. There are some enchantments that require changes to the actor such as "Enhanced Arcane Focus" which needs to add a bonus to spells or "Helm of Awareness" which gives advantage on initiative. Rather than try to include both item and actor changes into a single effect, enchantments should be able to define "ride-along" effects that are added with the enchantment but apply to the actor. These would be dependent effects so if the main enchantment is removed they are also removed.

arbron:
Enchantments can now define one or more effects that will be added whenever the enchantment is added and removed when the enchantment is removed. These are configured in the enchantment config.

Enchantment Ride Along Effects

In order to prevent any effects used as ride along effects from being transfered to the actor when not necessary, they get the `rideAlong` flag which is removed when applied. This prevents these effects from being transfered to the actor or being offered in the chat card. This separate flag allows for transfer and non-transfer ride along effects.

Closes #3511

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -751,8 +751,8 @@
     "Hint": "Disable the enchantment prompt when item is used. First available enchantment will always be applied."
   },
   "Riders": {
-    "Label": "Ride Along Effects",
-    "Hint": "These additional effects will be added when this enchantment is added, and removed when it is removed."
+    "Label": "Additional Effects",
+    "Hint": "These additional effects will be added to the enchanted item when this enchantment is added, and removed when the enchantment is removed."
   },
   "Warning": {
     "ConcentrationEnded": "Cannot apply this enchantment because concentration has ended.",
--- a/lang/en.json
+++ b/lang/en.json
@@ -750,7 +750,7 @@
     "Label": "Enchant Prompt",
     "Hint": "Disable the enchantment prompt when item is used. First available enchantment will always be applied."
   },
-  "RideAlong": {
+  "Riders": {
     "Label": "Ride Along Effects",
     "Hint": "These additional effects will be added when this enchantment is added, and removed when it is removed."
   },
--- a/lang/en.json
+++ b/lang/en.json
@@ -750,6 +750,10 @@
     "Label": "Enchant Prompt",
     "Hint": "Disable the enchantment prompt when item is used. First available enchantment will always be applied."
   },
+  "RideAlong": {
+    "Label": "Ride Along Effects",
+    "Hint": "These additional effects will be added when this enchantment is added, and removed when it is removed."
+  },
   "Warning": {
     "ConcentrationEnded": "Cannot apply this enchantment because concentration has ended.",
     "NoMagicalItems": "Items that are already magical cannot be enchanted.",
--- a/module/applications/item/enchantment-config.mjs
+++ b/module/applications/item/enchantment-config.mjs
@@ -118,7 +118,7 @@ export default class EnchantmentConfig extends DocumentSheet {
         foundry.utils.setProperty(updates, "flags.dnd5e.enchantment.riders", []);
       }
       // End bug fix
-      riderIds.add(...foundry.utils.getProperty(updates, "flags.dnd5e.enchantment.riders", []));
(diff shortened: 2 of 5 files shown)


lang/en.json
```diff
@@ -750,6 +750,10 @@
     "Label": "Enchant Prompt",
     "Hint": "Disable the enchantment prompt when item is used. First available enchantment will always be applied."
   },
+  "Riders": {
+    "Label": "Ride Along Effects",
+    "Hint": "These additional effects will be added when this enchantment is added, and removed when it is removed."
```
Fyorl, line 755: Not a huge fan of the 'ride along' term. Maybe call these 'Additional Effects', or even 'Creature Effects' to be very clear, but I am open to other suggestions.

```suggestion
    "Label": "Creature Effects",
    "Hint": "These additional effects will be added to the creature when one of its items is enchanted, and removed when the enchantment is removed."
```

module/applications/item/enchantment-config.mjs
```diff
@@ -97,7 +110,22 @@ export default class EnchantmentConfig extends DocumentSheet {
...
-    const effectsChanges = Object.entries(effects ?? {}).map(([_id, changes]) => ({ _id, ...changes }));
+    const riderIds = new Set();
+    const effectsChanges = Object.entries(effects ?? {}).map(([_id, changes]) => {
+      const updates = { _id, ...changes };
+      // Fix bug with <multi-select> in V11
+      if ( !foundry.utils.hasProperty(updates, "flags.dnd5e.enchantment.riders") ) {
+        foundry.utils.setProperty(updates, "flags.dnd5e.enchantment.riders", []);
+      }
+      // End bug fix
+      riderIds.add(...foundry.utils.getProperty(updates, "flags.dnd5e.enchantment.riders", []));
```
Fyorl, line 121:
```suggestion
      riderIds.add(...foundry.utils.getProperty(updates, "flags.dnd5e.enchantment.riders"));
```

module/documents/active-effect.mjs
```diff
@@ -327,7 +327,7 @@ export default class ActiveEffect5e extends ActiveEffect {
   /** @inheritDoc */
   prepareDerivedData() {
     super.prepareDerivedData();
-    if ( this.getFlag("dnd5e", "type") === "enchantment" ) this.transfer = false;
+    if ( this.getFlag("dnd5e", "type") === "enchantment" || this.getFlag("dnd5e", "rider") ) this.transfer = false;
```
Fyorl, line 330:
```suggestion
    if ( (this.getFlag("dnd5e", "type") === "enchantment") || this.getFlag("dnd5e", "rider") ) this.transfer = false;
```

module/documents/active-effect.mjs
```diff
@@ -391,6 +391,26 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+
+  /**
+   * Create additional effects that are applied separately from an enchantment.
+   */
+  async createRiderEnchantments() {
+    const origin = await fromUuid(this.origin);
+    const riders = (this.getFlag("dnd5e", "enchantment.riders") ?? []).map(id => {
+      const effectData = origin.effects.get(id)?.toObject();
+      if ( effectData ) {
+        delete effectData._id;
```
Fyorl, line 404: Should we also set `disabled: false` here too, for good measure?

Fyorl (changes requested):
Nice work, and thank you for taking the time to consolidate it with the changes from the rider conditions PR. I've just got one very minor point of discussion for the UI label for these rider effects.

lang/en.json
arbron: I'll go with "Additional Effects", because these are applied to the enchanted item and not necessarily transferred to the item's owner.

module/documents/active-effect.mjs
arbron: Well, you might want to have effects that are added but aren't immediately enabled, like an enchanted item that you activate and then it applies an effect.

Fyorl (approved):

PR #3532, opened 2024-05-05
Propose resolution to synchronous UUID link generation bug

aaclayton:
While using the system in V12 (12.231) I encountered a bug where advancement features being pulled from a compendium pack would throw an error and fail to render. You can easily reproduce this by creating a character and attempting to add a class from the compendium pack.

The error was being thrown by the attempted call to `TextEditor._createLegacyContentLink`. I'm not sure why that was deemed necessary.

There is some friction here because now in V12 content link generation is most directly an async operation, but this PR proposes a way to approach the issue using a synchronous workaround that is confirmed to work in V12 and should (but I did not test) work in V11.

### DIFF
--- a/module/utils.mjs
+++ b/module/utils.mjs
@@ -175,33 +175,23 @@ export function indexFromUuid(uuid) {
 
 /**
  * Creates an HTML document link for the provided UUID.
+ * Try to build links to compendium content synchronously to avoid DB lookups.
  * @param {string} uuid               UUID for which to produce the link.
  * @param {object} [options]
  * @param {string} [options.tooltip]  Tooltip to add to the link.
  * @returns {string}                  Link to the item or empty string if item wasn't found.
  */
 export function linkForUuid(uuid, { tooltip }={}) {
-  let element;
-
-  if ( game.release.generation < 12 ) element = TextEditor._createContentLink(["", "UUID", uuid]);
-
-  // TODO: When v11 support is dropped we can make this method async and return to using TextEditor._createContentLink.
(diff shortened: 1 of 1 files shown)


aaclayton:
I've realized there are a couple of issues with this suggestion. I was moving fast and trying to unblock myself, this probably would need some more careful thought to get it to handle all cases properly. Closing the PR.

module/utils.mjs
```diff
@@ -175,33 +175,22 @@ export function indexFromUuid(uuid) {
...
-      attrs: { draggable: "true" }
-    };
-    TextEditor._createLegacyContentLink("Compendium", [scope, pack, id].join("."), "", data);
-    data.dataset.link = "";
-    element = TextEditor.createAnchor(data);
+  let doc = fromUuidSync(uuid);
+  if ( uuid.startsWith("Compendium.") && !(doc instanceof foundry.abstract.Document) ) {
+    const {collection} = foundry.utils.parseUuid(uuid);
+    const cls = collection.documentClass;
+    doc = new cls(doc, {pack: collection.metadata.id}); // Build a minimal "shell" of the document using index data
```
arbron, line 189: I get a `can't define property "uuid": Object is not extensible` error when loading up a class in the compendium in V12. Deep cloning the data before creating the shell document fixes that:

```suggestion
    // Build a minimal "shell" of the document using index data
    doc = new cls(foundry.utils.deepClone(doc), {pack: collection.metadata.id});
```

arbron (changes requested):

arbron (approved):

PR #3538, opened 2024-05-07
[#2094, #3537] Convert target value into formula, move prep logic

### LINKED ISSUE #3537: Allow `targets` value to accept a formula
Spells like "Bless" increase the number of targets as they are cast at higher levels, so allowing formulas in the targets field in the same way it can be used in the duration field would be good to support this.

arbron:
Converts the targets value into a formula field so spells can change their values as they upcast. In order to support this and reduce code duplication the formula resolution logic used by max uses & duration fields was moved into its own method. All of the activated effect template data preparation was also moved into the data model at the same time.

In order to ensure these formulas resolve properly using item data for non-embedded items (Bless should show targets even when not on an actor), this modifies `getRollData` to always return item data even for non-embedded items.

In addition, the `replaceFormulaData` method has been moved from `Item5e` into utils to make it more generally available.

Closes #3537

### DIFF
--- a/module/data/item/templates/activated-effect.mjs
+++ b/module/data/item/templates/activated-effect.mjs
@@ -202,7 +202,6 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
   /** @inheritdoc */
   static _migrateData(source) {
     super._migrateData(source);
-    if ( ("activation" in source) && !source.activation.type ) source.activation.type = "";
     ActivatedEffectTemplate.#migrateFormulaFields(source);
     ActivatedEffectTemplate.#migrateRanges(source);
     ActivatedEffectTemplate.#migrateTargets(source);
@@ -234,7 +233,7 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
   static #migrateRanges(source) {
     if ( !("range" in source) ) return;
     source.range ??= {};
-    if ( ["none", null].includes(source.range.units) ) source.range.units = "";
+    if ( source.range.units === "none" ) source.range.units = "";
     if ( typeof source.range.long === "string" ) {
       if ( source.range.long === "" ) source.range.long = null;
       else if ( Number.isNumeric(source.range.long) ) source.range.long = Number(source.range.long);
@@ -259,8 +258,7 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
     if ( !("target" in source) ) return;
     source.target ??= {};
     if ( source.target.value === "" ) source.target.value = null;
-    if ( source.target.units === null ) source.target.units = "";
-    if ( ["none", null].includes(source.target.type) ) source.target.type = "";
+    if ( source.target.type === "none" ) source.target.type = "";
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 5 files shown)


module/data/item/templates/activated-effect.mjs
```diff
@@ -158,7 +234,7 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
   static #migrateRanges(source) {
     if ( !("range" in source) ) return;
     source.range ??= {};
-    if ( source.range.units === null ) source.range.units = "";
+    if ( ["none", ""].includes(source.range.units) ) source.range.units = null;
```
Fyorl, line 237: Is this correct? `range.units` is `nullable: false` (`StringField`s are `nullable: false` by default).

module/data/item/templates/activated-effect.mjs
```diff
@@ -184,7 +260,7 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
     source.target ??= {};
     if ( source.target.value === "" ) source.target.value = null;
     if ( source.target.units === null ) source.target.units = "";
-    if ( source.target.type === null ) source.target.type = "";
+    if ( ["none", ""].includes(source.target.type) ) source.target.type = null;
```
Fyorl, line 263: Same comment as above

module/data/item/templates/activated-effect.mjs
```diff
@@ -121,12 +170,39 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
...
+
+  /**
+   * Prepare a specific final formula, passing resolution errors to actor if available.
+   * @param {string} keyPath           Path within system data to where the property can be found.
+   * @param {object} options
+   * @param {string} options.label     Localizable name for the property to display in warnings.
+   * @param {object} options.rollData  Roll data to use to evaluate the formula.
+   */
+  _prepareFinalFormula(keyPath, { label, rollData }) {
+    let value = foundry.utils.getProperty(this, keyPath);
```
Fyorl, line 182:
```suggestion
    const value = foundry.utils.getProperty(this, keyPath);
```

Fyorl (changes requested):

module/data/item/templates/activated-effect.mjs
arbron: Hmm, well the previous code was rather confusing because it was using a migration to set `null` to `""`, and then in data preparation was setting `""` to `null`, so I'm entirely sure what values we want to be using here.

Fyorl: I'm not sure either. Could have been a mistake.

module/data/item/templates/activated-effect.mjs
```diff
@@ -121,12 +170,39 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
...
+  }
+
+  /* -------------------------------------------- */
+  /*  Data Migration                              */
   /* -------------------------------------------- */
 
   /** @inheritdoc */
   static _migrateData(source) {
     super._migrateData(source);
+    if ( ("activation" in source) && !source.activation.type ) source.activation.type = "";
```
Fyorl, line 205: If you update only part of the activation, this will clear the type:

```js
await item.update({ "system.activation": { cost: 1, type: "special" } });
console.log(item.system.activation.type); // "special"
await item.update({ "system.activation.cost": 2 });
console.log(item.system.activation.type); // ""
```

Fyorl (changes requested):

Fyorl (approved):

Issue #3541, opened 2024-05-08
Add consume usage/resource button on chat cards


arbron:
When the "Uses Prompt" checkbox is unchecked, there is current no way to consume item uses after the fact. This makes it difficult for items that have optional uses because you either need to uncheck the "Consume Available Usage" checkbox every time you activate it or uncheck "Uses Prompt" and manually adjust the uses from the actor sheet. It would be nice to have a button  allowing for uses to be consumed from the chat card.

Whether or not usage has been consumed should be tracked by the chat card so this button only appears if no uses were already consumed, which should prevent the chat log from being filled up with unnecessary buttons during normal play and prevent confusion if the button appears even after uses have been consumed.

krbz999:
> which should prevent the chat log from being filled up with unnecessary buttons during normal play

Though this would create a button for those items or features that never consume their own uses, like Channel Divinity, Ki, or magic items that hold charges only for their granted spells. Makes more sense to simply let those alternative actions consume the item's uses somehow.

PR #3542, opened 2024-05-08
[#3541] Add Consume Usage and Consume Resource buttons to chat

arbron:
Whenever an item has usage or resources to consume but those are not consumed during normal activation, this adds a pair of new chat buttons to "Consume Usage" or "Consume Resource" which performs the normal consumption when clicked. These buttons are removed from the chat card once they have been used to indicate that consumption has already occured.

To track whether these consumptions have happened, new flags have been added to messages in `dnd5e.use`: `consumedUsage`, `consumedResource`, and `consumedSpellSlot` (though no matching button was added for spell slots at this time).


Fyorl (approved):

Fyorl (approved):
It might be convenient to be able to consume additional uses from the chat card on-demand (i.e. not have the button disappear). But not something that requires immediate consideration.

Issue #3291, opened 2024-03-21
Automatically open damage application tray for new rolls


arbron:
It would be great if the damage application tray would pop open by default for newly created damage rolls so GMs don't have to open it every time.

krbz999:
Same for effects.

PR #3546, opened 2024-05-09
[#3291] Add setting to control chat card tray expansion

arbron:
Adds a new setting allowing setting the initial expanded state of damage & effect application trays in chat cards. The settings are "Expand All" and "Collapse All" plus "Collapse Older", which expands newly created cards but collapses any cards more than five minutes old.

To better support the expanding trays, a new CardTrayElement has been added which adds support for the `open` attribute on card trays to explicitly control its behavior. This attribute matches the behavior of the `open` attribute on `<details>` elements, so adding or removing the attribute will open and close the tray, and toggling the `open` property will cause the attribute to be added or removed. The tray also fires off a `toggle` event that matches the one fired by `<details>`.

### DIFF
--- a/module/applications/components/chat-tray-element.mjs
+++ b/module/applications/components/chat-tray-element.mjs
@@ -26,6 +26,7 @@ export default class ChatTrayElement extends HTMLElement {
   /*  Event Handlers                              */
   /* -------------------------------------------- */
 
+  /** @override */
   attributeChangedCallback(name, oldValue, newValue) {
     if ( name === "open" ) this._handleToggleOpen(newValue !== null);
   }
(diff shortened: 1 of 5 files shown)


module/applications/components/chat-tray-element.mjs
```diff
@@ -0,0 +1,60 @@
...
+  set open(open) {
+    if ( open ) this.setAttribute("open", "");
+    else this.removeAttribute("open");
+  }
+
+  /* -------------------------------------------- */
+  /*  Event Handlers                              */
+  /* -------------------------------------------- */
+
+  attributeChangedCallback(name, oldValue, newValue) {
```
Fyorl, line 29:
```suggestion
  /** @override */
  attributeChangedCallback(name, oldValue, newValue) {
```

Fyorl (approved):

PR #3552, opened 2024-05-10
[#2111, #3551] Split attunement & attuned, add optional attunement

### LINKED ISSUE #2111: Add support for "optional attunement"
Items with optional attunement are items that have some features that can be used without attunement and others that require attunement. It can be found in the "Spell Gem" in core material and has been popping up more frequently in 3rd party material (Helena's Guide to Monster Hunting uses this extensively if you have that). Adding support for this feature would be useful, but it is not possible with how we currently only use a single field to track attunement.

I propose either 1) change attunement to an object with `type` and `value`, or 2) moved attuned tracking from item to actor so that the `attunement` field on the item can represent just the attunement type (e.g. `none`, `required`, `optional`).

### LINKED ISSUE #3551: Split current attunement state from whether item requires attunement
Currently items store their attunement data in a single property that can be "No Attunement", "Attunement Required", or "Attuned". The current logic for attuning or un-attuning just toggles between the "Attunement Required" and "Attuned" values. Combing these two values causes a few problems:
- Overriding attunement with an enchantment prevents the attuned state from being changed
- Optional attunement items (see https://github.com/foundryvtt/dnd5e/issues/2111) cannot be implemented due the ambiguity as to un-attuned state when attuned

If we were to split attunement requirements and attuned state into two separate values that would resolve those issues.

arbron:
Splits `system.attunement` into two values, `attunement` which takes a string to indicate what kind of attunement is required, and `attuned` which is a `boolean` indicating whether the item is currently attuned.

This change allows simplification of a lot of the logic for attuning/un-attuning and for resetting attunement for dropped items.

This also adds the "Optional Attunement" mode that is used for the Spell Gem items and some 3rd party content. If attunement is optional then active effects and magical bonuses won't be suppressed if the item isn't attuned.

Also fixes a bug causing `_prepareItems` from being called twice on the new character sheets.

Closes #2111
Closes #3551

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -462,7 +462,7 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     const requireEquipped = (this.type !== "consumable")
       || ["rod", "trinket", "wand"].includes(this.system.type.value);
     if ( requireEquipped && (this.system.equipped === false) ) return true;
-    return this.system.attuned || (this.system.attunement !== "required");
+    return !this.system.attuned && (this.system.attunement === "required");
   }
 
   /* -------------------------------------------- */
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -462,7 +462,7 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     const requireEquipped = (this.type !== "consumable")
       || ["rod", "trinket", "wand"].includes(this.system.type.value);
     if ( requireEquipped && (this.system.equipped === false) ) return true;
-    return this.system.attunement === CONFIG.DND5E.attunementTypes.REQUIRED;
+    return this.system.attuned || (this.system.attunement !== "required");
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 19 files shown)


module/documents/item.mjs
```diff
@@ -462,7 +462,7 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     const requireEquipped = (this.type !== "consumable")
       || ["rod", "trinket", "wand"].includes(this.system.type.value);
     if ( requireEquipped && (this.system.equipped === false) ) return true;
-    return this.system.attunement === CONFIG.DND5E.attunementTypes.REQUIRED;
+    return this.system.attuned || (this.system.attunement !== "required");
```
Fyorl, line 465: I think this needs to be the inverse since we're returning whether the effect is suppressed or not:

```suggestion
    return !this.system.attuned && (this.system.attunement === "required");
```

Fyorl (changes requested):

Fyorl (approved):

PR #3553, opened 2024-05-10
[#3305] Allow "Enchant Prompt" to still enchant if prompt is shown

arbron:
Unchecking "Enchant Prompt" is useful when there is only a single enchantment available at any given time to help players to avoid seeing the ability use dialog if no other configuration is required, but if the dialog is shown then it requires the player to click "Apply Enchantment" or else the enchantment area won't appear in chat. For a more streamlined approach it would probably be simpler to just remove that checkbox so the enchantment area is always shown in chat if there is enchantments available. In this case the ability use dialog is only necessary when there is more than one enchantment profile to select.

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -980,7 +980,6 @@ export default class Item5e extends SystemDocumentMixin(Item) {
    * Configuration data for an item usage being prepared.
    *
    * @typedef {object} ItemUseConfiguration
-   * @property {boolean} applyEnchantment           Should this item apply an enchantment?
    * @property {boolean} createMeasuredTemplate     Should this item create a template?
    * @property {boolean} createSummons              Should this item create a summoned creature?
    * @property {boolean} consumeResource            Should this item consume a (non-ammo) resource?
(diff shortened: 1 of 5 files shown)


Fyorl (changes requested):
> For a more streamlined approach it would probably be simpler to just remove that checkbox so the enchantment area is always shown in chat if there is enchantments available. In this case the ability use dialog is only necessary when there is more than one enchantment profile to select.

Agreed with this. Do we also need to update the code which determines if ability use configuration is required? If there is only one profile and 'enchant prompt' is checked then you end up spawning a blank dialog.

arbron:
> Agreed with this. Do we also need to update the code which determines if ability use configuration is required? If there is only one profile and 'enchant prompt' is checked then you end up spawning a blank dialog.

Okay, I've taken this a bit further and removed the "Enchant Prompt" button entirely, so now the dialog will always be shown if there is more than one enchantment profile available because they cannot be chosen after the fact, and not if there is only a single profile. Should prevent users from having to fret over that value.

module/documents/item.mjs
```diff
@@ -1206,16 +1204,16 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
 
     const config = {
-      applyEnchantment: null,
       createMeasuredTemplate: null,
       createSummons: null,
       consumeResource: null,
       consumeSpellSlot: null,
       consumeUsage: null,
       enchantmentProfile: null,
+      promptEnchantment: null,
```
Fyorl, line 1216: We should probably add this to the `ItemUseConfiguration` typedef if it's a property that can be returned from this method.

Fyorl (approved):
> Okay, I've taken this a bit further and removed the "Enchant Prompt" button entirely, so now the dialog will always be shown if there is more than one enchantment profile available because they cannot be chosen after the fact, and not if there is only a single profile. Should prevent users from having to fret over that value.

I think that's a good change, yeah

PR #3556, opened 2024-05-12
[#2930]Added recharge logic to character-features.hbs

### LINKED ISSUE #2930: Items with `uses.value` or `uses.max` filled in but not `uses.per` still show as limited uses on the character sheet. If a feature instead has 'recharge', this does not show at all under any circumstances
Tested with a feature item.

Hoppyhob:
Added ctx.isOnCooldown and item.system.recharge.value checks from actor-features.hbs 
Moved div into the checks and added else to handle the empty case. 
closes #2930

Recharge


Fyorl:
I mentioned in the issue that this could probably wait for the NPC sheet since recharge is an NPC-only mechanic. We could merge it earlier though.

I was thinking of using an icon for this that could be clicked to roll the recharge, and then an additional context menu item to manually toggle the charged state on or off. Something like this:

image

image

If we can find some font awesome free icons that have an 'unfilled' variant that could also work:

image

image

image

(I think the star is probably not the best option but there's no free unfilled bolt)

We could also consider not showing the recharge formula, and put it on a tooltip instead, which I think reduces some noise in the display. Personally I don't need to know at a glance the actual formula, just whether the ability is charged or not.

PR #3559, opened 2024-05-13
[#2111, #3259] Move encumbrance calcs & add multipliers & bonuses

### LINKED ISSUE #3259: Allow capacity multiplier to be customized & targeted by active effects
I was trying to create an effect for Bull's Strength (one of the choices for the Enhance Ability spell), but it seems the only way to modify the carrying capacity multiplier is by overriding `flags.dnd5e.powerfulBuild` to `true`, since [the capacity is based only on strength, that flag, and character size](https://github.com/foundryvtt/dnd5e/blob/3edc7796236bf89cc91963da8ccc306e66e6f01a/module/documents/actor/actor.mjs#L493-L510). This works in some cases as a workaround, but not for small creatures like Halflings,  creatures that already have Powerful Build like Loxodons, or gargantuan creatures.

The easiest way I see to fix this would be to have some attribute key defaulting to 0 that effects could modify and gets [added to mod](https://github.com/foundryvtt/dnd5e/blob/3edc7796236bf89cc91963da8ccc306e66e6f01a/module/documents/actor/actor.mjs#L496). Other modifiers could be implemented too, e.g. a constant addition or subtraction to total carrying capacity, but I don't know any official effects that do something like that.

arbron:
Moves the encumbrance calculation logic into `AttributesFields` to be called by individual data models.

Also adds bonuses and multiplier fields that can be used to modify encumbrance in a more fine-grained level using active effects. There are formulas targeting the three encumbrance thresholds and one that applies to all three thresholds.

Closes #3259

### DIFF
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -180,7 +180,7 @@ export default class AttributesFields {
     const sizeConfig = CONFIG.DND5E.actorSizes[
       keys[this.parent.flags.dnd5e?.powerfulBuild ? Math.min(index + 1, keys.length - 1) : index]
     ];
-    const mod = sizeConfig?.capacityMultiplier ?? sizeConfig?.token ?? 1;
+    const sizeMod = sizeConfig?.capacityMultiplier ?? sizeConfig?.token ?? 1;
     let maximumMultiplier;
 
     const calculateThreshold = threshold => {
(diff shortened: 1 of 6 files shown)


krbz999:
Should add those fields to `ActiveEffect5e.FORMULA_FIELDS` as well.

module/data/actor/templates/attributes.mjs
```diff
@@ -145,6 +149,83 @@ export default class AttributesFields {
...
+      weight += numCoins / currencyPerWeight;
+    }
+
+    // Determine the Encumbrance size class
+    const keys = Object.keys(CONFIG.DND5E.actorSizes);
+    const index = keys.findIndex(k => k === this.traits.size);
+    const sizeConfig = CONFIG.DND5E.actorSizes[
+      keys[this.parent.flags.dnd5e?.powerfulBuild ? Math.min(index + 1, keys.length - 1) : index]
+    ];
+    const mod = sizeConfig?.capacityMultiplier ?? sizeConfig?.token ?? 1;
```
Fyorl, line 183: I know this was existing code but calling this `sizeMod` might make things a bit clearer in the calculations.

Fyorl (approved):

PR #3561, opened 2024-05-13
Fix preparation of movement/senses ranges and units

dev7355608:
I noticed in certain cases we currently end up with prepared `null` range/units instead of `0`/`"ft"`.

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -204,10 +204,12 @@ export default class CharacterData extends CreatureTemplate {
       AttributesFields.prepareRace.call(this, this.details.race);
       this.details.type = this.details.race.system.type;
     } else {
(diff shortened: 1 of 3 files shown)


Fyorl:
Units should probably be initialised to some default, but `null` is a valid value for these `NumberField`s. We don't prepare every single `NumberField` in the schema to `0` so it's unclear why doing so for these particular fields is necessary.

Fyorl:
> The units need to be initialized to for `prepareMovement`. The movement units are `null`, some of the movement reductions have no effect. No `null` ranges would be nice to have so that downstream code doesn't have to deal with that. In the context of detection modes a range of `null` is infinity: so there are chances that things go wrong with `null` ranges and a missing `?? 0` if the detection modes are prepared from the actors senses.

Yeah, definitely for units, but that makes sense for movement and senses as well

Fyorl (approved):

Issue #3563, opened 2024-05-14
Respect the addition of new transformationPresets into the config


liam-edwards:
As it currently stands, adding a transformation preset to `CONFIG.DND5E.transformationPresets` does not add it to the transformation dialog, as the dialog hardcodes each transformation preset, completely disregarding the config. https://github.com/foundryvtt/dnd5e/blob/d77116b5a178ede00567cbb67b2564475194cfb5/module/applications/actor/base-sheet.mjs#L845-L883

A quick fix to support the addition of new presets would be to replace lines 860 to 883 with the addition of the following:
```js
...Object.fromEntries(
    Object.entries(transformationPresets).map(([k, v]) => [k, {
        icon: v.icon,
        label: v.label,
        callback: html => this.actor.transformInto(sourceActor, foundry.utils.mergeObject(
            v.options,
            { transformTokens: rememberOptions(html).transformTokens }
        ))
    }])
)
```

It loses sorting but that can be quickly fixed.

liam-edwards:
With the addition of an `order` property to each preset, you can maintain the order of the transformations.
The `order` property should be between `0` and `1`, with the lower option appearing first.
```js
Object.fromEntries(
    Object.entries(transformationPresets).sort((a, b) => a[1].order - b[1].order).map(([k, v]) => [k, {
        icon: v.icon,
        label: v.label,
        callback: html => this.actor.transformInto(sourceActor, foundry.utils.mergeObject(
            v.options,
            { transformTokens: rememberOptions(html).transformTokens }
        ))
    }])
)
```

Additionally, as a correction to my previous statement at replacing lines 860 through to 883 with this, that would not work as the rest statement needs to be last, and the Cancel button follows. To get around this, a variable would need to be set after the Accept button, merge in the presets, then add the Cancel button afterwards.

PR #3566, opened 2024-05-15
[#3563] Add support for config-added transformation presets

liam-edwards:
Adds support for additional transformation presets through the config.

An `order` property was added to each preset item in the config to preserve the existing order.


arbron:
Should probably handle this alongside the redesign of the transformation dialog in `4.1`.

arbron:
Closing because the redesigned transformation dialog handles this: https://github.com/foundryvtt/dnd5e/pull/5239

PR #3567, opened 2024-05-15
[#3305] Add ability for enchantments to add rider items

arbron:
Adds rider items referenced by UUID that are added to a parent actor when one of their items is enchanted and removed when it is disenchanted.

Includes a backported version of `<document-tags>` from V12 as the interface for this to avoid having to implement something new.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -749,8 +749,14 @@
     "Entry": "{item} on  <em>{actor}</em>"
   },
   "Riders": {
-    "Label": "Additional Effects",
-    "Hint": "These additional effects will be added to the enchanted item when this enchantment is added, and removed when the enchantment is removed."
+    "Effect": {
+      "Label": "Additional Effects",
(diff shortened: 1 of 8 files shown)


lang/en.json
```diff
@@ -749,8 +749,14 @@
...
   "Riders": {
-    "Label": "Additional Effects",
-    "Hint": "These additional effects will be added to the enchanted item when this enchantment is added, and removed when the enchantment is removed."
+    "Effect": {
+      "Label": "Additional Effects",
+      "Hint": "These additional effects will be added to the enchanted item when this enchantment is added, and removed when the enchantment is removed."
+    },
+    "Item": {
+      "Label": "Additional Items",
+      "Hint": "These additional items will be added to the actor when an item is enchanted and removed when it loses its enchantment."
```
Fyorl, line 758:
```suggestion
      "Hint": "These additional items will be added to the creature when one of its items is enchanted, and will be removed if the enchantment is ever removed."
```

Fyorl (approved):
Looks good, only one suggestion for clarity around the rider item hint text

PR #3577, opened 2024-05-17
Control hints & other UI tweaks

Fyorl:
- Adds a setting for control hints so we can potentially place some information around the ctrl/alt/shift fast-forward functionality and other various mouse/keyboard combos.
- Adds a locked tooltip hint.
- Bespoke resize handle.

image


less/v2/tooltips.less
```diff
@@ -37,6 +37,26 @@
...
+  .lock-hint {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    gap: 4px;
+    padding: 6px 0 0;
+    margin-bottom: -4px;
+    font-size: var(--font-size-11);
+    color: var(--color-text-dark-5);
+    opacity: .75;
```
arbron, line 49: I sorta feel like this is both too prominent and too hard to read on that background. I fiddled with it a bit, decreasing the text size, removing the transparency, and pushing it to the right edge:

Screenshot 2024-05-17 at 13 15 56

```suggestion
    justify-content: flex-end;
    align-items: center;
    gap: 4px;
    padding: 6px 0 0;
    margin-bottom: -4px;
    margin-inline-end: -4px;
    font-size: var(--font-size-10);
    color: var(--color-text-dark-5);
```

It might also be good to hide the hint once the tooltip is locked.

Fyorl: Not sold on the right-alignment but I've removed the transparency, decreased the font size, and also hidden the hint when the tooltip is locked.

arbron (approved):

PR #3581, opened 2024-05-18
[#3401] Modified actor-features.hbs input node for item-detail

### LINKED ISSUE #3401: The NPC sheet does not allow for changing limited uses on items
The input field seems to do nothing at all.

Hoppyhob:
Added min, max, inputmode, and pattern attributes that were used on other item-details on the v1 sheets to the input for actor features. It appears that the inputmode="numeric" is used to bind the _onChangeInputDelta event.
Closes #3401

### DIFF
--- a/module/applications/actor/sheet-mixin.mjs
+++ b/module/applications/actor/sheet-mixin.mjs
@@ -15,7 +15,8 @@ export default Base => class extends Base {
   _onChangeInputDelta(event) {
     const input = event.target;
     const target = this.actor.items.get(input.closest("[data-item-id]")?.dataset.itemId) ?? this.actor;
-    parseInputDelta(input, target);
+    const result = parseInputDelta(input, target);
+    if ( result !== undefined ) target.update({ [input.dataset.name]: result });
   }
 
   /* -------------------------------------------- */
--- a/templates/actors/parts/actor-features.hbs
+++ b/templates/actors/parts/actor-features.hbs
@@ -59,7 +59,7 @@
                 {{else if item.system.recharge.value}}
                     {{localize "DND5E.Charged"}}
                 {{else if ctx.hasUses}}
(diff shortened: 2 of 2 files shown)


Fyorl (changes requested):
> Added min, max, inputmode, and pattern attributes that were used on other item-details on the v1 sheets to the input for actor features.

`min` and `max` are not supported attributes for `<input type="text">` fields so we should not add them, see [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input)

> It appears that the inputmode="numeric" is used to bind the _onChangeInputDelta event.

Are you sure this is the root cause? Looking at `ActorSheet5e#activateListeners` we have this line:

```js
inputs.addBack().find('[type="text"][data-dtype="Number"]').change(this._onChangeInputDelta.bind(this));
```

Which suggests that all we need is `[data-dtype="Number"]`, which this input seems to have, as far as I can tell.

Hoppyhob:
@Fyorl Ah, I did miss that. It looks like we have two _onChangeInputDelta()s one in inventory.mjs that requires inputmode="numeric" and one in sheet-mixin.mjs that requires type="text" and data-dtype="Number". The latter was not complete and would parse the input delta but not apply it. It also looks like we use inventory.mjs version most other places in the v1 sheets and both in the v2 sheets. I pushed a change to revert the changes from actor-features.hbs and add the update to sheet-mixin.mjs _onChangeInputDelta().

Fyorl (approved):
> @Fyorl Ah, I did miss that. It looks like we have two _onChangeInputDelta()s one in inventory.mjs that requires inputmode="numeric" and one in sheet-mixin.mjs that requires type="text" and data-dtype="Number". The latter was not complete and would parse the input delta but not apply it. It also looks like we use inventory.mjs version most other places in the v1 sheets and both in the v2 sheets. I pushed a change to revert the changes from actor-features.hbs and add the update to sheet-mixin.mjs _onChangeInputDelta().

Yes, sorry, there was a bit of shuffling with moving things into the mixin and then out again into the inventory element, but it looks like you got to the bottom of it, thanks.

PR #3593, opened 2024-05-20
[#3461] Added Overridden _onClickRight in token.mjs

### LINKED ISSUE #3461: Canceling token placement via right-clicking activates the token hud.
Repro:

- place token
- select it
- `await dnd5e.canvas.TokenPlacement.place({tokens: [token.actor.prototypeToken]})`
- right-click on canvas
- see token hud activate

Hoppyhob:
There was an issue where during token placement if the user right clicked we expect the token placement to end, however since the token preview is under the cursor most of the time during placement we were activating the _onClickRight for the token which then does not activate the canvas.app.view.oncontextmenu. This was fix by overriding the Token5e _onClickRight so that if the canvas.app.view.oncontextmenu event is defined, it will call that event otherwise it will call the normal _onClickRight for Token which activates the hud.
Closes #3461

### DIFF
--- a/module/canvas/token.mjs
+++ b/module/canvas/token.mjs
@@ -167,4 +167,11 @@ export default class Token5e extends Token {
     }
     return super._configureFilterEffect(statusId, active);
   }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _onClickRight(event) {
+    return (canvas?.app?.view?.oncontextmenu) ? canvas.app.view.oncontextmenu(event) : super._onClickRight(event);
+  }
 }


Hoppyhob:
@Fyorl It does appear that the fix from https://github.com/foundryvtt/dnd5e/commit/14e1cb33ce7887a60cb2e18ff508e95fb2364708 fixes the issue.

I did have some issues with break points in dev tools and foundry recognizing mouse up events to end things like panning, moving tokens, or drag selecting. Which sounds like what you are talking about, but I could not recreate them without breaking in dev tools and figured the button up event wasn't being registered because the application was in a break. That being said whether or not the break point thing is happening, there were other oddities I saw after retesting in the original fix like holding down the confirm placement (left click) makes the system act like you are dragging a token, which does not happen in the committed fix. So long way to say, looks good to me.

Fyorl:
> @Fyorl It does appear that the fix from [14e1cb3](https://github.com/foundryvtt/dnd5e/commit/14e1cb33ce7887a60cb2e18ff508e95fb2364708) fixes the issue.
> 
> I did have some issues with break points in dev tools and foundry recognizing mouse up events to end things like panning, moving tokens, or drag selecting. Which sounds like what you are talking about, but I could not recreate them without breaking in dev tools and figured the button up event wasn't being registered because the application was in a break. That being said whether or not the break point thing is happening, there were other oddities I saw after retesting in the original fix like holding down the confirm placement (left click) makes the system act like you are dragging a token, which does not happen in the committed fix. So long way to say, looks good to me.

It's definitely plausible that I had a breakpoint that was causing this behaviour, I don't quite remember. Thanks for verifying though, I'll go ahead and close this PR in that case.

Issue #3592, opened 2024-05-20
Add Active Effect for a minimum possible roll for skill checks


liam-edwards:
I would like to see a Active Effect added that allows you to set the minimum possible roll on the d20 for specific skills. This is similar to how a Rogue's Reliable Talent works, setting their minimum possible in all proficient skills to a 10. What I'm requesting is this to be more configurable.

How I imagine this working is like this:

Attribute Key | Change Mode | Effect Value
-|-|-
`system.skills.[abbreviation].check.minimum` | Upgrade | `[number]`
`system.skills.[abbreviation].check.maximum` | Downgrade | `[number]`

I included a `maximum` as well because if you're adding a minimum, it makes sense to add the inverse, though I can't imagine how it would be used.

When a skill check is rolled, it will check for that and either do `max(forcedMinimumValue, rolledValue)` or `max(forcedMaximumValue, rolledValue)`

arbron:
Duplicate of https://github.com/foundryvtt/dnd5e/issues/1350, there is already a PR for this but it is blocked by the roll refactor

PR #3599, opened 2024-05-22
[#3592] Add active effects for min/max skill check rolls

liam-edwards:
Added support for two new active effects to allow for the minimum or maximum possible result on a d20 roll for skill checks. I considered replacing how Reliable Talent works with this but realised there would be cascading changes that would be out of scope.

I never included it in the issue, however, the intention behind this is to support many homebrew features that set the minimum possible roll of specific skill checks, regardless of proficiency. While maximum was never included by anything I've seen, it was a simple addition and I can't see how it would be a negative to add.

Attribute Key | Change Mode | Effect Value
-|-|-
`system.skills.[abbreviation].check.minimum` | Upgrade | `[number]`
`system.skills.[abbreviation].check.maximum` | Downgrade | `[number]`

### DIFF
--- a/module/dice/d20-roll.mjs
+++ b/module/dice/d20-roll.mjs
@@ -40,6 +40,10 @@ const { Die, NumericTerm, OperatorTerm } = foundry.dice.terms;
  * @param {number} [options.fumble]              The value of d20 result which represents a critical failure
  * @param {(number)} [options.targetValue]       Assign a target value against which the result of this roll should be
  *                                               compared
+ *
+ * @param {number} [options.minimumRoll]         The lowest allowed result of a roll
+ * @param {number} [options.maximumRoll]         The highest allowed result of a roll
+ *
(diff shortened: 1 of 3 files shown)


krbz999:
There are existing properties for these things already.

`system.skills.<key>.roll.min` and `.max`.

arbron:
There is an existing PR for this (https://github.com/foundryvtt/dnd5e/pull/1760), but it is currently blocked by the roll refactor.

liam-edwards:
> There are existing properties for these things already.
> 
> `system.skills.<key>.roll.min` and `.max`.

Where can these be found?

> There is an existing PR for this (#1760), but it is currently blocked by the roll refactor.

I'll admit I hadn't seen that, but there hasn't been a commit on that since December 2022. I think it has been abandoned.

krbz999:
> Where can these be found?

Not sure I understand the question. The key used in an ActiveEffect is (almost always) an existing property on the actor it affects.

PR #3615, opened 2024-05-24
Fix Token Ring issues

dev7355608:
Fixes the invisibility / combat / defeated effects in V12.

The ring color becoming black after flashing is a V12 core issue.

### DIFF
--- a/module/canvas/token.mjs
+++ b/module/canvas/token.mjs
@@ -10,21 +10,6 @@ export default class Token5e extends Token {
   /* -------------------------------------------- */
 
   /**
-   * Callback invoked when a status effect is applied on a token.
-   * @param {Token5e} token         The token whose status effect is applied.
-   * @param {string} statusId       The status effect ID being applied, from CONFIG.specialStatusEffects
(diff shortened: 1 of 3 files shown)


module/canvas/token.mjs
```diff
@@ -127,7 +112,7 @@ export default class Token5e extends Token {
   /** @inheritDoc */
   _onUpdate(data, options, userId) {
     super._onUpdate(data, options, userId);
-    if ( !CONFIG.Token.ringClass.enabled || (game.release.generation > 11) ) return;
+    if ( (game.release.generation >= 12) || !CONFIG.Token.ringClass.enabled ) return;
```
Fyorl, line 115: Are these not equivalent?

Fyorl (approved):
Nice, thank you for testing both versions, these changes look good

module/canvas/token.mjs
dev7355608: I just reversed it because `CONFIG.Token.ringClass.enabled` doesn't exist in V12. But it doesn't make a difference here.

PR #3642, opened 2024-05-27
Make `TokenDocument#subjectPath` return the correct path in V12

dev7355608:


module/documents/token.mjs
```diff
@@ -206,8 +207,7 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
   /** @inheritDoc */
   _onUpdate(data, options, userId) {
     const textureChange = foundry.utils.hasProperty(data, "texture.src");
-    const tokenRingChange = foundry.utils.hasProperty(data, "flags.dnd5e.tokenRings.enabled");
```
dev7355608, line 209: `this.#subjectPath` doesn't seem to depend on the enabled state, so I just removed it.

module/documents/token.mjs
```diff
@@ -29,6 +29,7 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
    * @type {string}
    */
   get subjectPath() {
+    if ( game.release.generation >= 12 ) return this.ring.subject.texture;
```
Fyorl, line 32: Is this full path guaranteed to exist? Or do we need some optional chaining?

Fyorl (approved):

module/documents/token.mjs
dev7355608: Guaranteed to exist.

PR #3656, opened 2024-05-28
[#3607, #3608, #3610] Improve Table of Contents application

### LINKED ISSUE #3607: Table of Contents custom compendium application doesn't respect page order
Individual pages displayed within the table of contents are displayed in storage order, rather than the sorting value set on the pages.

### LINKED ISSUE #3608: Allow individual pages in Table of Contents to be hidden
Right now the show pages option is applied to the whole journal entry, allowing for hiding all of the pages but not for individual pages. A new flag should be available on pages to hide them.

```json
"flags": {
  "dnd5e": {
    "tocHidden": true
  }
}
```

### LINKED ISSUE #3610: Allow "Special" entries in Table of Contents to display their own pages
If special entries are appended and have `showPages` set to `true`, then they should display their pages.

arbron:
Sorts all pages displayed in the table of contents using their `order` property, rather than just using their storage order.

Special pages appended to another chapter can now include an `order` property to determine where they are sorted relative to the other pages in the chapter. They can also explicitly set `showPages` to `true` to display their own pages.

A `title` flag has been added for journal entries which replaces the journal entry's normal title.

Closes #3607 
Closes #3608 
Closes #3610

### DIFF
--- a/module/applications/journal/table-of-contents.mjs
+++ b/module/applications/journal/table-of-contents.mjs
@@ -41,38 +41,55 @@ export default class TableOfContentsCompendium extends Compendium {
       const flags = entry.flags?.dnd5e;
       if ( !flags ) continue;
       const type = flags.type ?? "chapter";
+
       if ( type === "header" ) {
         const page = entry.pages.contents[0];
         context.header = {
-          title: page?.name,
+          title: flags.title ?? page?.name,
           content: page?.text.content
         };
-      } else if ( type === "special" ) {
-        specialEntries.push({
-          type,
-          ...entry.toObject(),
-          showPages: flags.showPages,
+        continue;
+      }
+
+      const data = {
+        type,
+        id: entry.id,
+        name: flags.title ?? entry.name,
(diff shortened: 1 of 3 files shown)


module/applications/journal/table-of-contents.mjs
```diff
@@ -41,41 +41,58 @@ export default class TableOfContentsCompendium extends Compendium {
...
+        pages: Array.from(entry.pages).map(({ flags, id, name, sort }) => ({
+          name: flags.dnd5e?.title ?? name,
+          id,
+          entryId: entry.id,
+          sort,
           flags
-        });
+        })),
+        flags
+      };
```
Fyorl, line 66: Just a minor preference to group some of the auto-named properties together.

```js
const data = {
  type, flags,
  id: entry.id,
  name: flags.title ?? entry.name,
  pages: Array.from(entry.pages).map(({ flags, id, name, sort }) => ({
    id, sort, flags,
    name: flags.dnd5e?.title ?? name,
    entryId: entry.id
  }))
};
```

Fyorl (approved):

PR #3665, opened 2024-05-29
[#3078] Allow special journal blocks to be used on other sheets

### LINKED ISSUE #3078: Allow for the styles `.notable`, `.narrative`, `.quest`, `.advice` to be used outside of journals
Would be nice to use in eg a character sheet's description or an item. It is currently restricted to journals only.

arbron:
Moves the styling for special blocks to allow them to be used on actor & item sheets as well as journals.

Closes #3078

### DIFF
--- a/less/v2/journal.less
+++ b/less/v2/journal.less
@@ -192,102 +192,104 @@
       tr:nth-child(odd) { background-color: var(--dnd5e-color-table-row-odd); }
       td { padding: .375rem; }
     }
+  }
+}
 
-    /* Narrative Blocks */
-    .fvtt.narrative {
-      position: relative;
-      margin: 1.5rem 1rem;
-      padding: 1rem 1.5rem;
-      background-color: var(--dnd5e-color-card);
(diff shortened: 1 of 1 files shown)


krbz999:
While that's good, this styling will be lost when viewing the item description anywhere else, such as in the chatlog, a tooltip, or if embedded anywhere that isn't a journal, actor, or other item sheet.

less/v2/journal.less
```diff
@@ -192,102 +192,102 @@
...
-        left: 0;
-        z-index: -1;
-      }
-
-      &::before {
-        top: -13px;
-        right: 0;
-      }
+/* Notable Callout */
+aside.notable {
```
Fyorl, line 260: We risk clashing with core software styles if they are ever implemented, since some of these classes have relatively generic names. Is it possible to at least put them under something very general like `:is(.dnd5e, .dnd5e2)` maybe?

Fyorl (changes requested):

Fyorl (approved):

PR #3666, opened 2024-05-29
[#3659] Added clear call for preview

### LINKED ISSUE #3659: Canceled token placements stick around when placing multiples
Right clicking when placing multiple tokens using the `TokenPlacement` system, right click cancels the placement of a specific token but the token preview sticks around on the canvas until the whole process is completed.

Hoppyhob:
Added clear call for current placement preview if placement was false.
Closes #3659

### DIFF
--- a/module/canvas/token-placement.mjs
+++ b/module/canvas/token-placement.mjs
@@ -111,7 +111,7 @@ export default class TokenPlacement {
         obj.eventMode = "none";
         const placement = await this.#requestPlacement();
         if ( placement ) placements.push(placement);
-        else this.#previews[this.#currentPlacement]?.object.clear();
+        else obj.clear();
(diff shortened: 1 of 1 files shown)


module/canvas/token-placement.mjs
```diff
@@ -111,6 +111,7 @@ export default class TokenPlacement {
         obj.eventMode = "none";
         const placement = await this.#requestPlacement();
         if ( placement ) placements.push(placement);
+        else this.#previews[this.#currentPlacement]?.object.clear();
```
arbron, line 114: Since we already have the object as a local variable above, we can just use that to clean things up.

```suggestion
        else obj.clear();
```

arbron (approved):

PR #3677, opened 2024-05-31
[#3676] Fix issue with summoning using pact slot not setting level

### LINKED ISSUE #3676: Using pact slots does not allow for a choice of summons
https://github.com/foundryvtt/dnd5e/assets/50169243/abcc8dd1-8584-4ec3-ba53-b043573b401e

The item (a spell) has two choices for what to summon when using a 4th level (or higher) spell slot. However the user gets no options when selecting pact slots of exactly 4th level.

arbron:
Properly adjusts the slot level when pact level is selected in the dropdown and creates an initial clone at the proper level before rendering the dialog for if a level other than the default is set.

Closes #3676

### DIFF
--- a/module/applications/item/ability-use-dialog.mjs
+++ b/module/applications/item/ability-use-dialog.mjs
@@ -58,10 +58,16 @@ export default class AbilityUseDialog extends Dialog {
     const concentrationOptions = this._createConcentrationOptions(item);
     const resourceOptions = this._createResourceOptions(item);
 
+    const slotOptions = this._createSpellSlotOptions(item.actor, item.system.level);
+    if ( item.type === "spell" ) {
+      const slot = slotOptions.find(s => s.key === config.slotLevel) ?? slotOptions.find(s => s.canCast);
+      if ( slot ) item = item.clone({ "system.level": slot.level });
+    }
+
     const data = {
       item,
       ...config,
-      slotOptions: config.consumeSpellSlot ? this._createSpellSlotOptions(item.actor, item.system.level) : [],
+      slotOptions: config.consumeSpellSlot ? slotOptions : [],
       enchantmentOptions: this._createEnchantmentOptions(item),
       summoningOptions: this._createSummoningOptions(item),
       resourceOptions: resourceOptions,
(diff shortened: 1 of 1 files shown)


module/applications/item/ability-use-dialog.mjs
```diff
@@ -470,8 +476,10 @@ export default class AbilityUseDialog extends Dialog {
    * @param {Event} event  Triggering change event.
    */
   _onChangeSlotLevel(event) {
-    const level = parseInt(event.target.value.replace("spell", ""));
-    const item = this.item.clone({ "system.level": level });
+    const level = event.target.value === "pact"
```
krbz999, line 479: Seems I commented this on the commit rather than the PR, so putting it here as well for consideration.

`this.item.actor.system.spells[event.target.value].level` *should* presumably work just fine here.

module/applications/item/ability-use-dialog.mjs
```diff
@@ -58,10 +58,16 @@ export default class AbilityUseDialog extends Dialog {
     const concentrationOptions = this._createConcentrationOptions(item);
     const resourceOptions = this._createResourceOptions(item);
 
+    const slotOptions = this._createSpellSlotOptions(item.actor, item.system.level);
+    if ( item.type === "spell" ) {
+      const slot = slotOptions.find(s => s.key === config.slotLevel) ?? slotOptions.find(s => s.canCast);
+      if ( slot ) item = item.clone({ "system.level": slot.level });
```
Fyorl, line 64: Why do we need to indiscriminately clone the item if it's a spell?

arbron: `SummonsData#relevantLevel` fetches the appropriate level from the item's roll data, so we need to have the level be accurate for it to determine the correct levels for summoning in order to display the correct profiles available.

Fyorl (approved):

module/applications/item/ability-use-dialog.mjs
Fyorl: OK, that makes sense

PR #3679, opened 2024-05-31
[#3647] Display enchanted item count & limit on enchantment cards

### LINKED ISSUE #3647: Display enchantment target limit & enchanted item count
It would be useful if a single enchantment could have a limited amount of objects it could enchant. Few things come to mind:
- A spell that lets you enchant one weapon, and an additional weapon for every higher level the spell is cast.
- PHB Eldritch Knight can have up to two bonded weapons.
- Multiple (up to an ammount) Pieces of Ammunition for poisons?

arbron:
Adds a small display beneath the enchantment area on chat cards that displays the number of items that can be enchanted and the number currently enchanted. The enchantment limit is based off the target field on the enchanting item and scales with spell level.

Screenshot 2024-05-31 at 16 27 29

Closes #3647

### DIFF
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -660,7 +660,7 @@ enchantment-application {
     margin-inline-end: 4px;
     font-size: var(--font-size-10);
     text-align: end;
-    font-variant: small-caps;
+    text-transform: uppercase;
   }
 }
 
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -613,11 +613,12 @@
 /* ---------------------------------- */
 
 enchantment-application {
+  margin: 4px;
+
   .drop-area {
     display: grid;
     gap: 4px;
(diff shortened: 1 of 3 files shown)


less/v2/chat.less
```diff
@@ -653,6 +654,14 @@ enchantment-application {
       }
     }
   }
+
+  .count-area {
+    margin-block-start: 2px;
+    margin-inline-end: 4px;
+    font-size: var(--font-size-10);
+    text-align: end;
+    font-variant: small-caps;
```
Fyorl, line 663: This makes the text the same height as the numbers which I think looks a little better, and is a little easier to read given that it's quite small already at size 10.

```suggestion
    text-transform: uppercase;
```

Fyorl (approved):
Looks good, just one suggestion

Issue #3530, opened 2024-05-04
Wrong Maximized Critical Damage for rolls with dice modifiers


Ikaguia:
With the configuration to maximize dice on critical hit damage, if your weapon has dice modifiers, like `2d8kh` (for things like Savage Attacker) the maximized damage will ignore the modifiers (16 in this example, instead of 8)

Hoppyhob:
It seems like instead of trying to calculate the maximum we should evaluate the maximum with foundry. I can push a PR where I made that change, however it works in v11 but in v12 evaluate({ maximize: true }) does not call _evaluateModifiers() (I have a work around that checks the version and calls _evaluateModifiers() if it is v12 but it requires calling private Die api so I am not sure where we stand on that) I have submitted a foundry issue for this https://github.com/foundryvtt/foundryvtt/issues/11093

PR #3698, opened 2024-06-04
[#3530] Evaluate die maximum in damage-roll.mjs

Hoppyhob:
Instead of calculating the max value for a die, we can create a new die and ask foundry to evaluate it at the maximum. This works in v11 however the issue still is occurring in v12 as it does not appear synchronous rolls are processing their modifiers.
Issue #3530


module/dice/damage-roll.mjs
```diff
@@ -130,7 +130,12 @@ export default class DamageRoll extends Roll {
 
           // Powerful critical - maximize damage and reduce the multiplier by 1
           if ( this.options.powerfulCritical ) {
-            let bonus = term.number * term.faces;
+            const maxterm = new Die({
+              number: term.number,
+              faces: term.faces,
+              modifiers: term.modifiers
+            }).evaluate({ maximize: true });
```
Hoppyhob, line 137: As mentioned in the issue there is a work around for v12 if we make the configureDamage async and add the following code
`const version = parseInt(game.version.substring(0, game.version.indexOf(".")));
if (version <= 12) await maxterm._evaluateModifiers();`

arbron: You can use `game.release.generation` to make this easier. Also, current version is V12 so what are you protecting against with this?

arbron: I'm not sure how feasible it is to make `configureDamage` async because it is called within the roll's constructor.

Fyorl:
It's possible, but probably unlikely, that we would change foundryvtt/foundryvtt#11093 in v12 at this stage. So to support this in the system we would have to do some pre-processing of the roll, calculate the maximum values ourselves, and replace those `DiceTerm`s with `NumericTerm`s of the appropriate result. I don't think we should attempt to handle every possible modifier either (some wouldn't even make sense), probably just `k`, `kh`, `kl` , `d`, `dl`, and `dh` are good enough.

Hoppyhob:
Added a Roll5e roll type that extends Roll and loops through and replaces DiceTerm with NumericTerms when evaluate or evaluateSync is called and minimize or maximize is set, as suggested by Fyorl.
Currently considers modifiers k, kh, kl, d, dh, dl, max, and min.
Modified the enrichers.mjs, damage-roll.mjs, and d20-roll.mjs to use or extend the Roll5e.
Added the Roll5e to Config roll types.
Added version check to maintain evaluate compatibility with v11 in roll.mjs.

arbron:
> Added a Roll5e roll type that extends Roll and loops through and replaces DiceTerm with NumericTerms when evaluate or evaluateSync is called and minimize or maximize is set, as suggested by Fyorl. Currently considers modifiers k, kh, kl, d, dh, dl, max, and min. Modified the enrichers.mjs, damage-roll.mjs, and d20-roll.mjs to use or extend the Roll5e. Added the Roll5e to Config roll types. Added version check to maintain evaluate compatibility with v11 in roll.mjs. Closes #3530

We are adding a `BasicRoll` to be the new base roll class as part of the roll refactor, so once that other PR is merged these changes should be integrated into that class (see https://github.com/foundryvtt/dnd5e/pull/3146).

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,100 @@ export default class BasicRoll extends Roll {
...
+    const validModifiers = {
+      k: "keep",
+      kh: "keep",
+      kl: "keep",
+      d: "drop",
+      dh: "drop",
+      dl: "drop",
+      max: "maximum",
+      min: "minimum"
+    };
```
Fyorl, line 252: I see we are processing this list inside `if`-statements inside the loop further on. It would be more efficient to preprocess them here.

```suggestion
    const keep = new Set(["k", "kh", "kl"]);
    const drop = new Set(["d", "dh", "dl"]);
    const validModifiers = new Set([...keep, ...drop, "max", "min"]);
```

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,100 @@ export default class BasicRoll extends Roll {
...
+
+    for ( let modifier of currentModifiers ) {
```
Fyorl, line 254:
```suggestion
    for ( const modifier of currentModifiers ) {
```

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,100 @@ export default class BasicRoll extends Roll {
...
+      const rgx = /([Mm][AaIi][XxNn]|[KkDd][HhLl]?)([\d]+)?/i;
```
Fyorl, line 255: Since we are using the `i` flag, we do not need to include both upper- and lower-case letters in the regex. The `[\d]` character set is redundant also.

```suggestion
      const rgx = /(m[ai][xn]|[kd][hl]?)(\d+)?/i;
```

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,100 @@ export default class BasicRoll extends Roll {
...
+      const match = modifier.match(rgx);
+      if ( !match ) continue;
+      if (match && match[0].length < match.input.length) currentModifiers.push(match.input.slice(match[0].length));
```
Fyorl, line 258:
```suggestion
      if ( match && (match[0].length < match.input.length) ) currentModifiers.push(match.input.slice(match[0].length));
```

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,100 @@ export default class BasicRoll extends Roll {
...
+      let [command, value] = match.slice(1);
```
Fyorl, line 259:
```suggestion
      let [, command, value] = match;
```

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,100 @@ export default class BasicRoll extends Roll {
...
+      command = command.toLowerCase();
+      const amount = parseInt(value) || (command === "max" || command === "min" ? -1 : 1);
+
+      if (command in validModifiers && amount > 0 ) {
```
Fyorl, line 263: Per above change to `Set`:

```suggestion
      if ( validModifiers.has(command) && (amount > 0) ) {
```

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,100 @@ export default class BasicRoll extends Roll {
...
+        if ( (command === "max" && minimum) || (command === "min" && !minimum) ) continue;
+        else if ((command === "max" || command === "min") && amount >= 0 ) {
+          face = amount > die.faces ? die.faces : amount;
+        }
```
Fyorl, line 267: `amount` is guaranteed to be `> 0` by the higher-level `if`-statement.
```suggestion
        else if ( (command === "max") || (command === "min") ) face = Math.min(die.faces, amount);
```

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,100 @@ export default class BasicRoll extends Roll {
...
+        else if ( Object.entries(validModifiers).reduce((arr, [k, v]) => {if (v === "keep") arr.push(k); return arr;}, []).includes(command) ) {
+          number = amount < number ? amount : number;
+        }
+        else if ( Object.entries(validModifiers).reduce((arr, [k, v]) => {if (v === "drop") arr.push(k); return arr;}, []).includes(command) ) {
+          number = number - amount < 1 ? 1 : number - amount;
+        }
```
Fyorl, line 273: Per above change to `Set`:

```suggestion
        else if ( keep.has(command) ) number = Math.min(number, amount);
        else if ( drop.has(command) ) number = Math.max(1, number - amount);
```

module/dice/damage-roll.mjs
```diff
@@ -173,7 +175,13 @@ export default class DamageRoll extends Roll {
...
           // Powerful critical - maximize damage and reduce the multiplier by 1
           if ( this.options.powerfulCritical ) {
-            let bonus = term.number * term.faces;
+            const bonus = BasicRoll.create({
+              parts: [term.formula],
+              data: this.data,
+              situational: false,
+              options: this.options
+            }).evaluateSync({ maximize: true })
+              .total ?? 0;
```
Fyorl, line 184: Since all we have is a simple `DiceTerm` here, I don't think we need to pass all these extra options to `BasicRoll`. Note that `Roll.create` will automatically redirect to `BasicRoll`, so we do not need to explicitly import it here.

```suggestion
            const bonus = Roll.create(term.formula).evaluateSync({ maximize: true }).total ?? 0;
```

Fyorl (changes requested):
Sorry that this has languished for a little while, I've been quite busy. I have taken care of the rebase for you though, which I have force-pushed to your fork, so you will need to run `git fetch` then `git reset --hard origin/3.2.x_#3530` locally to make sure you get those changes.

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,86 @@ export default class BasicRoll extends Roll {
...
+   * Replaces all dice terms that have modifiers with their maximum/minimum value.
+   *
+   * @param {object} [options={}]            Extra optional arguments which describe or modify the BasicRoll.
+   * @returns {RollTerm[]}                   Returns the array of RollTerms.
+   */
+  preCalculateDiceTerms(options={}) {
+    if ( !this._evaluated && ((options.maximize ?? false) || (options.minimize ?? false)) ) {
+      this.terms = this.terms.map(term => {
+        if ( term instanceof foundry.dice.terms.DiceTerm && term.modifiers.length ) {
+          const minimize = ( options.maximize ?? false ) ? false : ( options.minimize ?? false );
```
krbz999, line 216: This can be simplified.
```suggestion
    if ( !this._evaluated && (options.maximize || options.minimize) ) {
      this.terms = this.terms.map(term => {
        if ( (term instanceof foundry.dice.terms.DiceTerm) && term.modifiers.length ) {
          const minimize = !options.maximize;
```

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,86 @@ export default class BasicRoll extends Roll {
...
+    for ( const modifier of currentModifiers ) {
+      const rgx = /(m[ai][xn]|[kd][hl]?)(\d+)?/i;
+      const match = modifier.match(rgx);
+      if ( !match ) continue;
+      if ( match && (match[0].length < match.input.length) ) currentModifiers.push(match.input.slice(match[0].length));
+      let [, command, value] = match;
+      command = command.toLowerCase();
+      const amount = parseInt(value) || (command === "max" || command === "min" ? -1 : 1);
+
+      if (validModifiers.has(command) && amount > 0 ) {
```
krbz999, line 255:
```suggestion
      if ( validModifiers.has(command) && (amount > 0) ) {
```

module/dice/basic-roll.mjs
```diff
@@ -180,4 +180,86 @@ export default class BasicRoll extends Roll {
...
+        if ( (command === "max" && minimize) || (command === "min" && !minimize) ) continue;
+        else if ((command === "max" || command === "min")) face = Math.min(die.faces, amount);
```
krbz999, line 257:
```suggestion
        else if ( (command === "max" || command === "min") ) face = Math.min(die.faces, amount);
```

module/dice/damage-roll.mjs
```diff
@@ -173,7 +173,7 @@ export default class DamageRoll extends Roll {
 
           // Powerful critical - maximize damage and reduce the multiplier by 1
           if ( this.options.powerfulCritical ) {
-            let bonus = term.number * term.faces;
+            const bonus = Roll.create(term.formula).evaluateSync({ maximize: true }).total ?? 0;
```
krbz999, line 176: I don't believe it's possible for this to be null or undefined.
```suggestion
            const bonus = Roll.create(term.formula).evaluateSync({ maximize: true }).total;
```

Fyorl (approved):

PR #3720, opened 2024-06-10
[#3719] Make award enricher handle XP 'correctly'

### LINKED ISSUE #3719: [Bug] Award enricher doesn't handle XP Correctly
When using an [[/award 100xp 200gp]] enricher, the XP is treated as a string, and causes the pop-up form to fail to render.
Currency values are treated as rolls, so `"200"` gets parsed and rolled to `200`; `"100"` does not.

Note: This could be considered a bug in foundry as a whole, as it's assuming `value` is a number from calling `Number.isNumeric` which is `true` for strings that look like numbers.

image
> value is `"100"`

NeilWhite:
Fixes #3719

### DIFF
--- a/module/applications/award.mjs
+++ b/module/applications/award.mjs
@@ -315,7 +315,13 @@ export default class Award extends DialogMixin(FormApplication) {
     }
 
     try {
-      const { currency, xp, party, each } = this.parseAwardCommand(message);
+      let { currency, xp, party, each } = this.parseAwardCommand(message);
+
+      if (xp) {
+        const roll = new Roll(xp);
+        await roll.evaluate();
+        xp = roll.total;
+      }
 
       for ( const [key, formula] of Object.entries(currency) ) {
         const roll = new Roll(formula);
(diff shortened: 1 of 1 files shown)


module/applications/award.mjs
```diff
@@ -315,7 +315,13 @@ export default class Award extends DialogMixin(FormApplication) {
     }
 
     try {
-      const { currency, xp, party, each } = this.parseAwardCommand(message);
+      let { currency, xp, party, each } = this.parseAwardCommand(message);
+
+      if (xp) {
+        const roll = new Roll(xp);
```
Fyorl, line 321: I don't think evaluating XP as a roll is necessary to fix this. We are not intending to support variable or rolled XP amounts. All I think is required is to cast the XP value to a number on line 369 (375 in this PR) in `parseAwardCommand`.

```js
else if ( label === "xp" ) xp = Number(amount);
```

Fyorl (changes requested):

Fyorl (approved):

PR #3727, opened 2024-06-11
[#3717] Fix appended numbers not being incremented for placed tokens

### LINKED ISSUE #3717: Bug: dropping unlinked npc tokens from an Group-Encounter actor, while the prototype token for the npc is set to add numbers to the ends of token names, simply names all of the tokens NPC (1)
https://i.imgur.com/fKw8L5c.gif

Repo.
Create an NPC and set the prototype token to "Append Number to Name of Unlinked Tokens."
Create a Group Token.
Change Group Type to Encounter.
Add more than one of the NPC you just created.
Place them.
Notice that the amended number does not increase.

arbron:
Since the core's code for appending numbers in `getTokenDocument` looks at the tokens currently in the scene, and token placement creates tokens in bulk, this modifies the generated numbers after the call to `getTokenDocument` to adjust them based on the index of the placed token.

To that end, this adds a `index` value to `PlacementData` which has a total index across all placements and an index for each unique original token, to avoid numbering issues with placement of mixed groups of actors.

Closes #3717

### DIFF
--- a/module/canvas/token-placement.mjs
+++ b/module/canvas/token-placement.mjs
@@ -10,6 +10,9 @@
  *
  * @typedef {object} PlacementData
  * @property {PrototypeToken} prototypeToken
+ * @property {object} index
+ * @property {number} index.total             Index of the placement across all placements.
+ * @property {number} index.unique            Index of the placement across placements with the same original token.
  * @property {number} x
  * @property {number} y
  * @property {number} rotation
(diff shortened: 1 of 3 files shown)


module/canvas/token-placement.mjs
```diff
@@ -10,6 +10,9 @@
  *
  * @typedef {object} PlacementData
  * @property {PrototypeToken} prototypeToken
+ * @property {object} index
+ * @property {number} index.total             Index of the placement across all placements.
```
Fyorl, line 14: It seems we don't actually use this total?

module/canvas/token-placement.mjs
```diff
@@ -104,14 +107,19 @@ export default class TokenPlacement {
...
         this.#currentPlacement++;
         const obj = canvas.tokens.preview.addChild(this.#previews[this.#currentPlacement].object);
         await obj.draw();
         obj.eventMode = "none";
         const placement = await this.#requestPlacement();
-        if ( placement ) placements.push(placement);
-        else obj.clear();
+        if ( placement ) {
+          uniqueTokens.set(placement.prototypeToken, (uniqueTokens.get(placement.prototypeToken) ?? -1) + 1);
+          placement.index = { total: total++, unique: uniqueTokens.get(placement.prototypeToken) };
```
Fyorl, line 120: Is it important that we specifically track the number of unique tokens? Can we just keep a simple counter and add it to the number generated by core?

Fyorl (changes requested):

module/canvas/token-placement.mjs
arbron: No, but I figured since it is easy to make it might be nice for other places to have this bit of data in hooks, in places where they don't have the whole `placements` array to work with.

module/canvas/token-placement.mjs
Fyorl: OK, I wasn't sure how core was doing it, if it was counting unique tokens or not. It looks like it's checking based on the `actorId`, so do we want to maybe use `actorId` as the key to the `Map` rather than the `PrototypeToken` instance?

module/canvas/token-placement.mjs
Fyorl: Sure, that's no problem then

Fyorl (approved):

PR #3728, opened 2024-06-11
Don't display scrolling combat text if the token is secret

dev7355608:
Related core changes: https://github.com/foundryvtt/foundry-vtt/pull/2478

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -3341,7 +3341,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const fill = CONFIG.DND5E.tokenHPColors[key];
 
     for ( const token of tokens ) {
-      if ( !token.object?.visible || !token.object?.renderable ) continue;
+      if ( !token.object?.visible || token.isSecret ) continue;
       if ( token.hasDynamicRing ) token.flashRing(key);
       const t = token.object;
       canvas.interface.createScrollingText(t.center, value.signedString(), {


module/documents/actor/actor.mjs
```diff
@@ -3341,7 +3341,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const fill = CONFIG.DND5E.tokenHPColors[key];
 
     for ( const token of tokens ) {
-      if ( !token.object?.visible || !token.object?.renderable ) continue;
+      if ( !token.object?.visible || token.isSecret ) continue;
```
dev7355608, line 3344: `TokenDocument#isSecret` will be an addition with V12 Stable 4. Considering it is `undefined` in prior versions, we don't really need a version check here, I think.

module/documents/actor/actor.mjs
```diff
@@ -3341,7 +3341,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const fill = CONFIG.DND5E.tokenHPColors[key];
 
     for ( const token of tokens ) {
-      if ( !token.object?.visible || !token.object?.renderable ) continue;
```
dev7355608, line 3344: We removed `renderable` in core `ActiveEffect#_displayScrollingStatus` with V12 Stable 4. Removing `renderable` makes the combat scrolling text appear when the token config is open, which wasn't the case previously.

Fyorl (approved):

Issue #3743, opened 2024-06-17
Actor transformation does not override the Token Ring configuration of transformed tokens in V12


dev7355608:
Originally reported by nonchalantwombat: https://discord.com/channels/170995199584108546/1184176415139184731/1251140525831229531

ALL MODULES DISABLED? true
OS, Hosting, Browser (if applicable): Windows 11, Hosting locally with Foundry client, Foundry v12.
Short Description of bug: When using polymorph/token shapeshifting, the custom scaling and dynamic token ring fails to transfer
Simple steps to reproduce the bug: Make two actors with different sizes, art, and scaled dynamic tokens. Polymorph/wildshape one into the other. The dynamic token ring will scale incorrectly, and not be the size of the intended prototype token, instead retaining the initial token's settings.

unsoluble:
Does this PR also address this 3.2.1 behaviour? If you have a DTR Subject Path set for a token, then as the transformation completes, this original subject path pops back onto the token (it should not).
https://github.com/foundryvtt/dnd5e/assets/801582/443bc558-7ff0-466d-a316-98602fd86130

PR #3744, opened 2024-06-17
Fix polymorphing with Token Ring in V12

dev7355608:
Closes https://github.com/foundryvtt/dnd5e/issues/3743.


module/documents/actor/actor.mjs
```diff
@@ -3061,7 +3064,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
         }));
         return;
       }
-      const prototypeTokenData = await baseActor.getTokenDocument();
+      const prototypeTokenData = (await baseActor.getTokenDocument()).toObject();
```
dev7355608, line 3067: Unrelated to the issue. Spotted it in passing. We should be using source data here, not prepared data.

PR #3745, opened 2024-06-18
fix dynamic token flashing on targeting

Drental:
targetToken passes the token placeable, so we need to access the token.document to see if the dynamic token ring is enabled

### DIFF
--- a/module/canvas/token.mjs
+++ b/module/canvas/token.mjs
@@ -16,7 +16,10 @@ export default class Token5e extends Token {
    * @param {boolean} targeted    Is the token targeted or not?
    */
   static onTargetToken(user, token, targeted) {
-    if ( !targeted || !token.document.ring?.enabled ) return;
+    if ( !targeted ) return;
+    if ( game.release.generation < 12 ) {
+        if ( !token.ring?.enabled ) return;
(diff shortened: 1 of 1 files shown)


module/canvas/token.mjs
```diff
@@ -16,7 +16,7 @@ export default class Token5e extends Token {
    * @param {boolean} targeted    Is the token targeted or not?
    */
   static onTargetToken(user, token, targeted) {
-    if ( !targeted || !token.ring?.enabled ) return;
+    if ( !targeted || !token.document.ring?.enabled ) return;
```
dev7355608, line 19: We need to keep the this logic for V11.

```suggestion
    if ( !targeted ) return;
    if ( game.release.generation < 12 ) {
        if ( !token.ring?.enabled ) return;
    } else if ( !token.hasDynamicRing ) return;
```

dev7355608 (changes requested):

Drental:
as a note, I find supporting v11 with the latest release weird, it's more likely to cause issues with v11 versions of modules than it is to improve compatibility overall.

But this PR was just a drive by from trying to showcase uses of the dynamic ring to some other system devs, so I don't really mind the details anyway^^

dev7355608 (approved):

Issue #3730, opened 2024-06-12
Implement Compendium Browser


arbron:
A number of features in the system require a way to browse content across multiple compendiums.

## Required Features
- [x] Load content quickly from multiple compendiums
- [x] Variety of filters for narrowing down results
  - [x] Document class (e.g. Actor or Item)
  - [x] Types (e.g. Weapon, Equipment, Loot, Class, etc.)
  - [x] Additional filters defined by individual types (e.g. filtering by properties, categories & sub-types, etc.)
- [x] Preset filters that can be programmatically locked in (e.g. clicking the "Add Background" link on the character sheet only shows a list of available backgrounds)
- [x] Search for finding specific item
- [ ] Hiding/disabling specific items in the results (if selecting a class, disable classes that character already has)
- [x] Selecting items and performing an action based on selection

## Nice to Have Features
- [ ] Ability for GMs to control what compendiums are visible in the browser beyond normal permissions
- [ ] Sorting of results by other methods than alphabetical

krbz999:
It would be nice to be able to hide individual items as well.

For example, when a player browses for a class, hiding "GM-only" classes such as those found in the DMG would be good despite those maybe living in a big shared compendium of class items. (User-configurable of course).

Also if the table has decided they are using an unofficial variant of an existing class or subclass, being able to hide the standard one.

Zanderaf:
Yea, hiding stuff would absolutely be great, I remade most of my own stuff in my private compendium, and rarely use the SRD stuff so hiding that would be a real useful thing to be able to do.

arbron:
### Filters by Types
Useful additional filters for specific actor & item types (must be based on indexable data):

#### Class
- Is spellcaster

#### Subclass
- Associated class

#### Race
- Has darkvision

#### Feature
- Feature type & subtype
- Properties

#### Spell
- Level
- School
- Properties
- Spell List (when properly implemented)

#### Physical Items
- Properties
- Rarity
- Requires Attunement
- Category & sub-type

#### NPC
- Creature Size
- Creature Type
- CR (min & max)
- Has certain speed

mclemente:
Would it be interesting to have Casting Time and Action Type as filters for spells?

![image](https://github.com/user-attachments/assets/8f2b7ead-4e86-4207-ac98-3e758529ee1b)

Just add this to `data/item/spell.mjs`:
```js
["activation", {
  label: "DND5E.SpellCastTime",
  type: "set",
  config: {
[...]

PR #3746, opened 2024-06-19
[#3730] Implement initial version of the Compendium Browser

arbron:
Adds the initial structure of the compendium browser with the ability to filter by document class (e.g. Actor or Item) and by type (e.g. NPC, Spell, Weapon). The browser will search all available compendiums and display all matching documents based on the types selected.

Compendium Browser - Initial General

The compendium browser can be invoked using a button on the compendium packs sidebar tab, which opens a compendium browser with no restrictions, or through the Add Race/Background/Class buttons on the character sheet, which open a compendium browser instance restricted to those specific types.

When opening a compendium browser programatically, certain filters can be locked to allow for only viewing a subset of documents. This locking will be used when invoking the browser from character creation, from a Choose Items advancement, or when selecting spells.

Compendium Browser - Initial Classes

The compendium browser has a `fetch` method which is responsible for getting results from across compendiums. This method filters compendiums by the appropriate document class and if more specific types are provided, adds additional compendium filtering using a new `types` flag that can be defined on packs in the manifest indicating which types can be found in that compendium. If the flag isn't set, then that compendium will be searched regardless of type, but it should reduce the number of compendiums that need to be searched if these flags are used widely.

The `fetch` method then indexes each compendium using explicitly provided index fields as well as any fields that are defined in the filters. Once the index is returned for each compendium it uses the new `Filter` system to quickly narrow down the results. If full documents are required, only after this initial filtering will it fetch those to avoid loading more than necessary.

As a final step it flattens down the results into a single list and sorts using the provided sorting key path or method, falling back on alphabetical by name if no sorting is specified.

The browser is implemented using AppV2 which is only available in Foundry V12, so it is not available when running in V11, but stubs were used to ensure the rest of the system is still available in the older version.

#### Next Steps
- [x] Implement more specific filtering
- [x] Make entries drag & droppable
- [x] Allow for selecting one or more entry from the list and returning selected documents
- [x] Properly style using `dnd5e2` styles


module/applications/compendium-browser.mjs
```diff
@@ -0,0 +1,416 @@
...
+      const context = { entry };
+      rendered.push(
+        renderTemplate("systems/dnd5e/templates/compendium/browser-entry.hbs", context)
+          .then(html => {
+            const template = document.createElement("template");
+            template.innerHTML = html;
+            const element = template.content.firstElementChild;
+            element.dataset.tooltip = `
+              <section class="loading" data-uuid="${entry.uuid}">
+                <i class="fa-solid fa-splinner fa-spin-pulse"></i>
```
Fyorl, line 212:
```suggestion
                <i class="fa-solid fa-spinner fa-spin-pulse" inert></i>
```

module/applications/compendium-browser.mjs
```diff
@@ -0,0 +1,416 @@
...
+              </seciton>
```
Fyorl, line 213:
```suggestion
              </section>
```

module/applications/compendium-browser.mjs
```diff
@@ -0,0 +1,416 @@
...
+            `;
+            element.dataset.tooltipClass = "dnd5e2 dnd5e-tooltip item-tooltip";
+            element.dataset.tooltipDirection ??= "LEFT";
```
Fyorl, line 216: I found that having the tooltips on the left obscured the filter sidebar.
```suggestion
            element.dataset.tooltipDirection ??= "RIGHT";
```

module/applications/compendium-browser.mjs
```diff
@@ -0,0 +1,416 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Prepare the sidebar context.
+   * @param {string} partId                        The part being rendered.
+   * @param {ApplicationRenderContext} context     Shared context provided by _prepareContext.
+   * @param {HandlebarsRenderOptions} options      Options which configure application rendering behavior.
+   * @returns {Promise<ApplicationRenderContext>}  Context data for a specific part.
```
Fyorl, line 146: I'm guessing this is intended to be `@protected` since it's underscore-prefixed. The same for the methods below this one.

module/applications/compendium-browser.mjs
```diff
@@ -0,0 +1,416 @@
...
+   * Inject the compendium browser button into the compendium sidebar.
+   * @param {HTMLElement} html  HTML of the sidebar being rendered.
+   */
+  static injectSidebarButton(html) {
+    if ( game.release.generation < 12 ) return;
+    const button = document.createElement("button");
+    button.type = "button";
+    button.classList.add("open-compendium-browser");
+    button.innerHTML = `
+      <i class="fa-solid fa-book-open-reader"></i>
```
Fyorl, line 408:
```suggestion
      <i class="fa-solid fa-book-open-reader" inert></i>
```

Fyorl (changes requested):

module/filter.mjs
```diff
@@ -0,0 +1,362 @@
...
+/* <><><><> <><><><> <><><><> <><><><> <><><><> <><><><> */
+
+/**
+ * Currently supported comparison functions.
+ * @enum {Function}
+ */
+export const COMPARISON_FUNCTIONS = {
+  _: exact, exact, iexact, contains, icontains,
+  startswith, istartswith, endswith, iendswith,
+  has, hasAny, hasAll, in: in_, gt, gte, lt, lte
```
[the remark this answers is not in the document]
arbron: I'll stick with all lowercase to avoid using `iContains` which looks too much like an Apple product.

module/filter.mjs
```diff
@@ -0,0 +1,362 @@
...
+/* <><><><> <><><><> <><><><> <><><><> <><><><> <><><><> */
+
+/**
+ * Case-insensitive exact match.
+ * @param {*} data
+ * @param {*} value
+ * @returns {boolean}
+ */
+export function iexact(data, value) {
+  return exact(String(data).toLowerCase(), String(value).toLowerCase());
```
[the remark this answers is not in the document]
arbron: Since we aren't using any case insensitive filters at the moment, maybe I'll just pull them out for the moment and we can stick them back if we find a need for them and make the decision then.

module/filter.mjs
```diff
@@ -0,0 +1,362 @@
...
+ * @returns {boolean}
+ */
+export function XOR(data, filter) {
+  let currentResult = false;
+  for ( const { k, v, o } of filter ) {
+    if ( _check(data, k, v, o) ) {
+      if ( !currentResult ) currentResult = true;
+      else return false;
+    }
+  }
```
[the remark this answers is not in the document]
arbron: As far as I can tell, there isn't one definite answer to how a 3+ input XOR gate should behave:

`true` if one and only one input is `true`:
xor1

`true` if an odd number of inputs are `true`:
xor2

I suppose we could adapt whatever interpretation we think might be useful.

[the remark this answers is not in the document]
Fyorl: I'm just using the more typical definition in Computer Science, see [here](https://mathworld.wolfram.com/XOR.html). For example, this is the definition used in Javascript.

```js
console.log(true ^ true); // 0
console.log(true ^ true ^ true); // 1
```

Fyorl (approved):

PR #3753, opened 2024-06-23
Update staves

krbz999:
Updates various magical staves/quarterstaves in the Items srd pack.
- This does not add 'quarterstaff' as base item unless the staff explicitly states it can be wielded as a quarterstaff.

### DIFF
--- a/packs/_source/items/weapon/staff-of-charming.json
+++ b/packs/_source/items/weapon/staff-of-charming.json
@@ -52,9 +52,9 @@
     "uses": {
       "value": 10,
       "max": "10",
-      "per": "charges",
+      "per": "dawn",
       "recovery": "1d8 + 2",
-      "prompt": true
+      "prompt": false
     },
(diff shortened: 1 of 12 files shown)


Fyorl:
I don't think it's a given that the most common use of, say, a Staff of Fire, is to attack with it as a Quarterstaff, and thus have `prompt: false`. In the absence of any complaints/issues here, I think we should maintain the status quo and allow users to edit the Item themselves based on their preferences.

krbz999:
A staff that grants spells doesn't have any use of its own charges. The spells use those. I can't imagine ever wanting to use the staff to expend charges when this can also be done directly through the actor sheet.

Fyorl:
Right, that makes sense. Please provide this sort of context up-front or link to an issue so we're not left trying to guess what the PR is about.

Fyorl (approved):
Good changes

PR #3764, opened 2024-06-27
[#3763] Select spellcasting ability based on spell's associated class

### LINKED ISSUE #3763: Automatically select spellcasting ability based on class set on spell
With the merging of https://github.com/foundryvtt/dnd5e/pull/3431 you can now assign a spell to a specific class. For any spells that have a class specified and "Spellcasting" set as their ability, it should automatically use the ability on their assigned class rather than the character-wide spellcasting ability.

The spellcasting ability should still be supported to support older spells without the class field set and to support spellcasting on NPCs which probably don't have a spellcasting class.

arbron:
Modifies `_typeAbilityMod` property on `SpellData` to first check the associated spellcasting class if specified on the spell before falling back to the global spellcasting ability.

Also made a change to `Item5e#getSaveDC` to first ask the item for its ability to use for the DC before falling back to the global spell DC.

Closes #3763

### DIFF
--- a/module/data/item/spell.mjs
+++ b/module/data/item/spell.mjs
@@ -157,7 +157,9 @@ export default class SpellData extends ItemDataModel.mixin(
 
   /** @inheritdoc */
   get _typeAbilityMod() {
-    return this.parent?.actor?.system.attributes.spellcasting || "int";
+    return this.parent?.actor?.spellcastingClasses[this.sourceClass]?.spellcasting.ability
+      ?? this.parent?.actor?.system.attributes.spellcasting
+      ?? "int";
   }
 
   /* -------------------------------------------- */
--- a/module/data/item/spell.mjs
+++ b/module/data/item/spell.mjs
@@ -158,7 +158,7 @@ export default class SpellData extends ItemDataModel.mixin(
   /** @inheritdoc */
   get _typeAbilityMod() {
     return this.parent?.actor?.spellcastingClasses[this.sourceClass]?.spellcasting.ability
-      ?? this.parent?.actor?.system.attributes.spellcasting
+      ?? this.parent?.actor?.system.attributes?.spellcasting
       ?? "int";
   }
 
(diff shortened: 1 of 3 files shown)


module/data/item/spell.mjs
```diff
@@ -157,7 +157,9 @@ export default class SpellData extends ItemDataModel.mixin(
 
   /** @inheritdoc */
   get _typeAbilityMod() {
-    return this.parent?.actor?.system.attributes.spellcasting || "int";
+    return this.parent?.actor?.spellcastingClasses[this.sourceClass]?.spellcasting.ability
+      ?? this.parent?.actor?.system.attributes.spellcasting
```
krbz999, line 161: Doesn't this assume that the actor's data model has `attributes`? That may not be the case for a module subtype.

module/data/item/spell.mjs
```diff
@@ -157,7 +157,9 @@ export default class SpellData extends ItemDataModel.mixin(
 
   /** @inheritdoc */
   get _typeAbilityMod() {
-    return this.parent?.actor?.system.attributes.spellcasting || "int";
+    return this.parent?.actor?.spellcastingClasses[this.sourceClass]?.spellcasting.ability
```
Fyorl, line 160: `Actor5e#spellcastingClasses` should probably be cached in the same way `Actor5e#classes` is, since it's going to be called by every single spell Item during preparation, potentially more than once.

Fyorl (changes requested):

module/documents/actor/actor.mjs
```diff
@@ -18,10 +18,17 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /**
    * The data source for Actor5e.classes allowing it to be lazily computed.
-   * @type {Object<Item5e>}
+   * @type {Record<string, Item5e>}
    * @private
    */
-  _classes;
+  #classes;
```
Fyorl, line 24: A good idea, but because of some weirdness with how ActorDeltas are prepared, it seems to break preparation of unlinked token Actors.

Fyorl (changes requested):

Fyorl (approved):

PR #3768, opened 2024-06-27
[#3742] Replace safeEval with evaluateSync in V12

### LINKED ISSUE #3742: Replace all `Roll.safeEval` with `Roll#evaluateSync`
`CONFIG.Dice.functions` are not available in `Roll.safeEval`.

- Custom dice functions cannot be used in Armor Class formulas.
- `simplifyBonus` logs an error and returns 0 if it encounters a custom dice function that is deterministic in the formula.
- `simplifyRollFormula` fails if it encounters custom dice function that is deterministic in the formula.
- ...

arbron:
Closes #3742

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -499,12 +499,20 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
         else ac.dex = this.system.abilities.dex?.mod ?? 0;
 
         rollData.attributes.ac = ac;
-        const replaced = replaceFormulaData(formula, rollData, {
-          actor: this, property: game.i18n.localize("DND5E.ArmorClass")
-        });
-        ac.base = replaced ? game.release.generation < 12
-          ? Roll.safeEval(replaced) : new Roll(replaced).evaluateSync().total
-          : 0;
+        try {
+          const replaced = replaceFormulaData(formula, rollData, {
+            actor: this, missing: null, property: game.i18n.localize("DND5E.ArmorClass")
+          });
+          ac.base = replaced ? game.release.generation < 12
+            ? Roll.safeEval(replaced) : new Roll(replaced).evaluateSync().total
+            : 0;
+        } catch(err) {
+          this._preparationWarnings.push({
+            message: game.i18n.format("DND5E.WarnBadACFormula", { formula }), link: "armor", type: "error"
+          });
+          const replaced = Roll.replaceFormulaData(CONFIG.DND5E.armorClasses.default.formula, rollData);
+          ac.base = game.release.generation < 12 ? Roll.safeEval(replaced) : new Roll(replaced).evaluateSync().total;
+        }
         break;
     }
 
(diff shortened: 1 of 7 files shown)


module/documents/actor/actor.mjs
```diff
@@ -500,14 +500,24 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
         rollData.attributes.ac = ac;
         try {
-          const replaced = Roll.replaceFormulaData(formula, rollData);
-          ac.base = Roll.safeEval(replaced);
+          if ( game.release.generation < 12 ) {
+            const replaced = Roll.replaceFormulaData(formula, rollData);
+            ac.base = Roll.safeEval(replaced);
+          } else {
+            const roll = new Roll(formula, rollData);
```
dev7355608, line 507: There's a small difference there, I think. The replacement done by `new Roll()` is done with `{ missing: "0" }`, which is not done in the V11 case.
`const roll = new Roll(Roll.replaceFormulaData(formula, rollData));` would be equivalent to the V11 case, but maybe `{ missing: "0" }` should have been used in the V11 case. I'm not sure.

module/documents/actor/actor.mjs
```diff
@@ -500,14 +500,24 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
           this._preparationWarnings.push({
             message: game.i18n.localize("DND5E.WarnBadACFormula"), link: "armor", type: "error"
           });
-          const replaced = Roll.replaceFormulaData(CONFIG.DND5E.armorClasses.default.formula, rollData);
-          ac.base = Roll.safeEval(replaced);
+          if ( game.release.generation < 12 ) {
+            const replaced = Roll.replaceFormulaData(CONFIG.DND5E.armorClasses.default.formula, rollData);
+            ac.base = Roll.safeEval(replaced);
+          } else {
+            const roll = new Roll(CONFIG.DND5E.armorClasses.default.formula, rollData);
```
dev7355608, line 518: Same thing with the `{ missing: "0" }`.

module/documents/actor/actor.mjs
arbron: I think we don't want to do that so we can catch errors in AC formulas if they have missing data. But calling roll normally doesn't do that either, so maybe we'll have to use our custom `replaceFormulaData` method.

arbron:
@Fyorl Calling in Kim for review because this will result in a slight change in behavior for AC preparation. Previously formulas that failed to evaluate would fall back to the equipped armor formula, but now it falls back to setting the AC base to just `0`.

Fyorl:
> @Fyorl Calling in Kim for review because this will result in a slight change in behavior for AC preparation. Previously formulas that failed to evaluate would fall back to the equipped armor formula, but now it falls back to setting the AC base to just `0`.

Both sound like reasonable enough ways to handle the failure case there. What would be involved in restoring the previous behaviour?

dev7355608:
> What would be involved in restoring the previous behaviour?

`new Roll(Roll.replaceFormulaData(formula, data))` instead of `new Roll(formula, data)`.

arbron:
> Both sound like reasonable enough ways to handle the failure case there. What would be involved in restoring the previous behaviour?

I've made an additional change that makes it more similar to the old system, so missing terms in AC formulas will now fall back to the equipped armor formula. I had to add back the `try…catch` to handle cases where an invalid AC formula is set using an active effect, to avoid breaking data preparation.

Fyorl (approved):

PR #3770, opened 2024-06-27
[#2129] Add additional hint for dice scale value & <copyable-text>

### LINKED ISSUE #2129: Add information on additional formulas for Dice scale values
Currently dice scale values list and provide a copy button for the whole formula (`@scale.fighter.superiorityDice`) but not to the new formulas needed to access portions (`@scale.fighter.superiorityDice.number`, `@scale.fighter.superiorityDice.faces`, and `@scale.fighter.superiorityDice.die`) making those formulas difficult to discover.

arbron:
Adds an extended identifier hint for dice scale values to help accessing the individual parts of the die. In order to make this easier this introduces a `<copyable-text>` element which adds a button for copying the contents of the element to the clipboard.

Closes #2129


Fyorl (changes requested):
No real issues with the code, it looks good. From testing it out, I had some feedback:

1. Do we need to restrict the copy functionality to the appended button only? It seems reasonably ergonomic to be able to click anywhere on the `<copyable-text>` element to copy it.
2. The full hint text for dice scale values runs over several lines and forms a big block, I was finding it a bit difficult to visually parse.

image

I don't have any great ideas for how to format it more digestibly. Ideally we'd have some diagram like this:

```
   4      d      6
|      |      |      |
+------+      +------+
Number |       Faces |
       +-------------+
             Die
```

But obviously we can't do that easily here. My only other suggestion would be to omit it and leave this as information in the wiki, but that does kind of defeat the purpose of this PR in the first place.

MaxPat931:
Borrowing from the format of the wiki, perhaps something like:
```
The following formulas can be used to return certain values of the current level of this scale value:
@scale.bard.inspiration 📋- The full scale value (e.g. 4d6)
@scale.bard.inspiration.number 📋- The number of dice defined in the scale value (e.g. 4 of 4d6).
@scale.bard.inspiration.die 📋- The die defined in the scale value (e.g. d6 of 4d6).
@scale.bard.inspiration.faces 📋- The number of faces on the die defined in the scale value (e.g. 6 of 4d6).
```

Fyorl:
> Borrowing from the format of the wiki, perhaps something like:
> 
> ```
> The following formulas can be used to return certain values of the current level of this scale value:
> @scale.bard.inspiration 📋- The full scale value (e.g. 4d6)
> @scale.bard.inspiration.number 📋- The number of dice defined in the scale value (e.g. 4 of 4d6).
> @scale.bard.inspiration.die 📋- The die defined in the scale value (e.g. d6 of 4d6).
> @scale.bard.inspiration.faces 📋- The number of faces on the die defined in the scale value (e.g. 6 of 4d6).
> ```

Yeah, a list would work. It's just a bit awkward because it's hint text at the moment, so we would need some bespoke styling regardless if we're really committed to cramming in all of this information at this particular place.

arbron:
Screenshot 2024-08-16 at 10 45 44

lang/en.json
```diff
@@ -253,7 +253,8 @@
 "DND5E.AdvancementSaveButton": "Save Advancement",
 "DND5E.AdvancementScaleValueTitle": "Scale Value",
 "DND5E.AdvancementScaleValueHint": "A single value that changes as a class increases in level that is made available in roll formulas (such as a Monk's martial arts die).",
-"DND5E.AdvancementScaleValueIdentifierHint": "This scale value for the current level will be available using <strong>@scale.{class}.{identifier}</strong> in a formula.",
+"DND5E.AdvancementScaleValueIdentifierHint": "This scale value for the current level will be available in a formula using:<br><copyable-text>@scale.{class}.{identifier}</copyable-text>",
+"DND5E.AdvancementScaleValueIdentifierDiceHint": "- whole formula (<code>4d6</code>)<br><copyable-text>@scale.{class}.{identifier}.die</copyable-text> - whole die (<code>d6</code>)<br><copyable-text>@scale.{class}.{identifier}.number</copyable-text> - number of dice (<code>4</code>)<br><copyable-text>@scale.{class}.{identifier}.faces</copyable-text> - number of faces (<code>6</code>)",
```
Fyorl, line 257: I don't know how localisation-friendly this is, considering how word order changes a lot. Maybe it's fine since it's a list, but we could probably just repeat the whole of `AdvancementScaleValueIdentifierHint` in `AdvancementScaleValueIdentifierDiceHint` to be safe.

less/elements.less
```diff
@@ -1,3 +1,33 @@
+/* ---------------------------------- */
+/*  Copyable Text                     */
+/* ---------------------------------- */
+
+copyable-text {
+  text-decoration: underline;
+  text-decoration-style: dashed;
+  cursor: pointer;
+  &:hover, &:has(:focus-visible) {
```
Fyorl, line 9: Is `:focus-within` equivalent?

```suggestion
  &:hover, &:focus-within {
```

module/applications/components/copyable-text.mjs
```diff
@@ -0,0 +1,45 @@
...
+   * Controller for removing listeners automatically.
+   * @type {AbortController}
+   */
+  #controller;
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  disconnectedCallback() {
+    this.querySelector("button")?.remove();
```
Fyorl, line 29:
```suggestion
    this.#controller.abort();
    this.querySelector("button")?.remove();
```

Fyorl (approved):

less/elements.less
arbron: `:focus-within` in this case behaves more like `:focus`, so it will stay highlighted after clicking.

PR #3771, opened 2024-06-28
[#3761] Pass allowInteractive: false to blind rolls

### LINKED ISSUE #3761: Evaluate rolls without interaction if the roll mode is blind
At least the following functions need to be adjusted:
- `DamageRoll#toMessage`
- `d20Roll`
- `damageRoll`
- `rollFormula`

arbron:
Closes #3761

### DIFF
--- a/module/dice/dice.mjs
+++ b/module/dice/dice.mjs
@@ -207,7 +207,7 @@ export async function damageRoll({
 
   // Evaluate the configured roll
   for ( const roll of rolls ) {
-    const rollMode = roll.options.rollMode ?? defaultRollMode;
+    const rollMode = rolls.at(-1).options.rollMode ?? defaultRollMode;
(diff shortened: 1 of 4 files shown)


module/dice/dice.mjs
```diff
@@ -207,7 +207,8 @@ export async function damageRoll({
 
   // Evaluate the configured roll
   for ( const roll of rolls ) {
-    await roll.evaluate({async: true});
+    const rollMode = roll.options.rollMode ?? defaultRollMode;
```
dev7355608, line 210:
```suggestion
    const rollMode = rolls.at(-1).options.rollMode ?? defaultRollMode;
```

Should be the same logic as in `DamageRoll.toMessage`. But I think both are equivalent, because all rolls should have the same roll mode here anyway.

dev7355608 (changes requested):

dev7355608 (approved):

Issue #2623, opened 2023-11-18
[2.4] Racial ASI Advancement - Add option to assign points to fixed improvements.


drakonik2782:
I have a number of races I'm making that have one fixed +1 ASI, and two free-floating +1s that can be assigned to any ability (to a maximum of +2 per score). The new advancement feature does not allow this; if I set a +1 to Dexterity, for example, I can't boost that ability with any of my free points because it's a "Fixed Improvement."

Please add an option to allow the use of free points on top of pre-set points.
![dnd5e-2 4-issue-1](https://github.com/foundryvtt/dnd5e/assets/114558351/a576bd45-39aa-4379-8f6a-f53f4d085823)
![dnd5e-2 4-issue-2](https://github.com/foundryvtt/dnd5e/assets/114558351/fc471de1-48aa-4789-9cf3-411bbbb39bbd)

PR #3784, opened 2024-07-01
[#2623] Separate ability score locking from fixed improvements

arbron:
Previously if you granted an ability score a fixed improvement it also locked that score so it cannot be improved further. This moves the list of locked ability scores into a separate list so scores without fixed improvements can be locked, and scores with fixed improvements can still be improved further.

Ability Score Improvement Locking

This is a slightly breaking change because fixed improvements that were previously locked will no longer be so.


Fyorl (approved):
Spotted an issue with double-counting the fixed improvements, but I think I got it. The branch does need a rebase though.

PR #3789, opened 2024-07-02
[#2728] Add hint to all advancement types

### LINKED ISSUE #2728: Add Hint text box for all Advancement Types
Hint text boxes were added to the newer Advancement types to allow for custom text to be displayed in the Advancement workflow, these should be added to all Advancement Types.
Types that do not have Hints: 
ASI
Grant Items
Scale Values
Hit Points

arbron:
Adds `hint` to the top-level advancement data and migrates any existing hints from within the configuration data. Now if hints are set for any advancement it will display in the flow window.

Closes #2728


Fyorl (approved):
Looks good, I noticed that the `TraitConfig` is throwing a deprecation warning, and also has two hint textareas though.

PR #3804, opened 2024-07-05
Initial v2 NPC sheet integration.

Fyorl:
This includes registering the sheet, and the overall sheet layout with header matter, sidebar content, and the features tab. I've kept out the other tabs for follow-up commits in order to try to keep the size down. Unfortunately, it's still a fair amount of bespoke CSS and markup, even with a lot of code sharing with the actor sheet.


less/v2/npc.less
```diff
@@ -0,0 +1,610 @@
...
+      /* Sidebar Collapser */
+      .sidebar-collapser {
+        position: absolute;
+        inset: 0 auto 0 var(--sidebar-size);
+        width: 16px;
+        padding: 0;
+        color: var(--dnd5e-color-gold);
+        transition: left 450ms ease;
+
+        &:hover, &:focus {
```
arbron, line 467: Might want to use `focus-visible` for this otherwise it stays highlighted after you click it and move your mouse away.

```suggestion
        &:hover, &:focus-visible {
```

templates/actors/npc-sheet-2.hbs
```diff
@@ -0,0 +1,511 @@
...
+
+                        <div class="proficiency">
+                            <span>{{ localize "DND5E.Proficiency" }}</span>
+                            <strong>{{ dnd5e-formatModifier system.attributes.prof }}</strong>
+                        </div>
+
+                    </div>
+
+                    {{!-- CR --}}
+                    <div class="right cr-badge badge" aria-label="{{ localize "DND5E.CRLabel" cr=system.details.cr }}">
```
arbron, line 253: It would be good to have a tooltip here, at least when not in editing mode, to give some indication of what this number means for people new to the sheet.

templates/actors/npc-sheet-2.hbs
```diff
@@ -0,0 +1,511 @@
...
+                    </ul>
+                </div>
+
+                {{!-- Skills --}}
+                <div class="pills-group {{#unless skills}}empty{{/unless}}">
+                    <h3 class="icon">
+                        <i class="fas fa-briefcase"></i>
+                        <span class="roboto-upper">{{ localize "DND5E.Skills" }}</span>
+                        {{#if editable}}
+                        <a class="config-button" data-action="skills" data-tooltip="DND5E.SkillsConfig"
```
arbron, line 418: I'm assuming this is not implemented just cause we don't have a "Configure All Skills" application?

module/applications/actor/npc-sheet-2.mjs
```diff
@@ -0,0 +1,220 @@
...
+  ];
+
+  /* -------------------------------------------- */
+
+  get template() {
+    if ( !game.user.isGM && this.actor.limited ) return "systems/dnd5e/templates/actors/limited-sheet-2.hbs";
+    return "systems/dnd5e/templates/actors/npc-sheet-2.hbs";
+  }
+
+  /* -------------------------------------------- */
```
arbron, line 37:
```suggestion
    /* -------------------------------------------- */
    /*  Rendering                                   */
    /* -------------------------------------------- */
```

module/applications/actor/npc-sheet-2.mjs
```diff
@@ -0,0 +1,220 @@
...
+        { classes: "item-roll", label: "DND5E.SpellHeader.Roll", partial: "dnd5e.column-roll" },
+        { classes: "item-formula", label: "DND5E.SpellHeader.Formula", partial: "dnd5e.column-formula" },
+        { classes: "item-controls", partial: "dnd5e.column-feature-controls" }
+      ];
+    });
+    context.classes = classes;
+    context.hasClasses = classes.length;
+  }
+
+  /* -------------------------------------------- */
```
arbron, line 193:
```suggestion
    /* -------------------------------------------- */
    /*  Event Listeners & Handlers                  */
    /* -------------------------------------------- */
```

module/applications/actor/npc-sheet-2.mjs
```diff
@@ -0,0 +1,220 @@
...
+    });
+    context.hasLegendaries = resources.legact.max || resources.legres.max || resources.lair.initiative;
+
+    return context;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  activateListeners(html) {
```
arbron, line 140: Move down into the "Event Listeners" section to match the mixin grouping.

module/applications/actor/npc-sheet-2.mjs
```diff
@@ -0,0 +1,220 @@
...
+    const context = await super.getData(options);
+    const { attributes, details, resources } = this.actor.system;
+    context.encumbrance = attributes.encumbrance;
+
+    // CR
+    context.labels.cr = {
+      "0.125": "&frac18;",
+      "0.25": "&frac14;",
+      "0.5": "&frac12;"
+    }[details.cr] ?? details.cr;
```
arbron, line 81: We have the `formatCR` utility method now, maybe we could even make that a handlebars helper so you don't have to worry about it in the sheet preparation at all.

```suggestion
    context.labels.cr = formatCR(details.cr);
```

module/applications/actor/npc-sheet-2.mjs
```diff
@@ -0,0 +1,220 @@
+import ActorSheet5eNPC from "./npc-sheet.mjs";
+import ActorSheetV2Mixin from "./sheet-v2-mixin.mjs";
+import { simplifyBonus } from "../../utils.mjs";
```
arbron, line 3:
```suggestion
import { formatCR, simplifyBonus } from "../../utils.mjs";
```

(if you decide to keep CR formatting in the sheet)

module/config.mjs
```diff
@@ -38,28 +38,32 @@ DND5E.abilities = {
     abbreviation: "DND5E.AbilityStrAbbr",
     type: "physical",
     fullKey: "strength",
-    reference: "Compendium.dnd5e.rules.JournalEntry.NizgRXLNUqtdlC1s.JournalEntryPage.nUPv6C66Ur64BIUH"
+    reference: "Compendium.dnd5e.rules.JournalEntry.NizgRXLNUqtdlC1s.JournalEntryPage.nUPv6C66Ur64BIUH",
+    icon: "systems/dnd5e/icons/svg/abilities/strength.svg"
```
arbron, line 42: Will need to update the `AbilityConfiguration` typedef to include `[icon]`.

arbron (changes requested):
Only other note is that the header icons in the circles aren't very legible because of the transparent circles on the lighter background compared to the PC sheets, perhaps they just need less transparency for the circles.

templates/actors/npc-sheet-2.hbs
Fyorl: Yeah I left it out of this PR to reduce the line count and allow it to be reviewed separately.

module/applications/actor/npc-sheet-2.mjs
```diff
@@ -0,0 +1,226 @@
...
```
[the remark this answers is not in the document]
Fyorl: Yeah, it will be in another PR

PR #3808, opened 2024-07-08
Add inventory tab to NPC sheet v2.

Fyorl:


less/v2/npc.less
```diff
@@ -524,10 +524,40 @@
         width: 100%;
 
         .items-list { gap: 8px; }
+
+        .bottom {
+          width: 100%;
+          position: sticky;
+          bottom: 0;
```
arbron, line 531: It feels a bit off that these elements are flush with the bottom of the window when scrolled up. Perhaps setting the bottom to match the inventory margin would be better?

arbron (changes requested):
No weights other than the container weights are displaying. You'll need to set `totalWeight` on the item's context. You should be able to do this in the reducer in `ActorSheet5eNPC#_prepareItems`:

```javascript
ctx.totalWeight = item.system.totalWeight?.toNearest(0.1);
```

Other than that it is looking good.

less/v2/npc.less
Fyorl: Yeah I agree with you that it looks a little off. I did try setting `bottom: 8px` originally but seeing part of the inventory content peeking out in the 8px below the bar looked worse than having it flush (in my opinion). I also tried setting some margin on the scroll container so the whole content was no longer flush with the bottom but that didn't look great either. This was the least bad look I could come up with for now.

arbron (approved):

PR #3810, opened 2024-07-08
Add spells tab to NPC sheet v2.

Fyorl:


less/v2/npc.less
```diff
@@ -549,6 +549,40 @@
...
+          option { color: var(--color-text-dark-primary); }
+        }
+
+        .config-button {
+          flex: 1;
+          display: grid;
+          place-content: center;
+        }
+
+        .spellcasting-ability { color: var(--dnd5e-color-gold); }
```
arbron, line 584: Very difficult to read.

Screenshot 2024-07-08 at 15 20 28

module/applications/actor/npc-sheet-2.mjs
```diff
@@ -190,6 +193,41 @@ export default class ActorSheet5eNPC2 extends ActorSheetV2Mixin(ActorSheet5eNPC)
...
+   * Prepare spellcasting data for display.
+   * @param {object} context  The display context.
+   * @protected
+   */
+  _prepareSpellcasting(context) {
+    const { abilities, attributes, bonuses, details } = this.actor.system;
+    context.spellcasting = [];
+    const msak = simplifyBonus(bonuses.msak.attack, context.rollData);
+    const rsak = simplifyBonus(bonuses.rsak.attack, context.rollData);
+    // TODO: Consider if we want to handle multiclassing for NPC spellcasters.
```
arbron, line 208: Though we don't have anything to prevent it now, it would probably just be best to not allow multiclassing on NPCs.

arbron (changes requested):

less/v2/npc.less
Fyorl: Looks ok to me, but I have added a text shadow to hopefully make it a bit clearer.

PR #3816, opened 2024-07-09
Don't create system flags data model from the system flags in `_source`

dev7355608:


Fyorl (approved):
Thanks for taking care of this

PR #3820, opened 2024-07-12
Compendium browser UI & iteration

Fyorl:
- Added dnd5e2 styling.
- Added basic mode filtering.
- Added name search.
- Added result rendering batching.
- Added drag-drop.
- Added check for compendium visibility.

The basic mode filtering is geared more towards player options or finding monsters. The advanced mode retains full compendium searching features for power users,

I had to add some basic batched rendering because it got very laggy when rendering several hundred entries. We need to do more here by deleting elements that scroll out of view and replacing them with empty boxes of the same size (then repopulating them if the user scrolls back) in order to keep browser memory manageable, but it's fairly complex.

We're still missing several important features like adding additional columns, sorting, and filtering by module/book source, but I think it's in a good enough state that it can still be useful.

image

image

image


module/applications/actor/character-sheet-2.mjs
```diff
@@ -432,8 +432,7 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
         case "background": game.packs.get(CONFIG.DND5E.sourcePacks.BACKGROUNDS)?.render(true); break;
       }
     } else {
-      const result = await CompendiumBrowser.selectOne({ filters: { locked: { types: new Set([type]) } } });
-      if ( result ) this._onDropItemCreate(await fromUuid(result));
+      new CompendiumBrowser({ filters: { locked: { types: new Set([type]) } } }).render(true);
```
Fyorl, line 435: For these specifically, I think we want to stick with drag-drop workflow as dragging from a compendium is how it worked before. The selection API remains in the compendium browser though for integration into summoning.

module/applications/compendium-browser.mjs
```diff
@@ -301,13 +523,26 @@ export default class CompendiumBrowser extends HandlebarsApplicationMixin(Applic
...
 
     else if ( partId === "filters" ) {
-      context.additional = Array.from(context.filterDefinitions?.entries() ?? []).reduce((obj, [key, data]) => {
-        obj[key] = foundry.utils.mergeObject(data, {
+      context.additional = Array.from(context.filterDefinitions?.entries() ?? []).reduce((arr, [key, data]) => {
+        // Special case handling for 'Feats' tab in basic mode.
+        if ( (types[0] === "feat") && ((key === "category") || (key === "subtype")) ) return arr;
+
+        let sort = 0;
+        switch ( data.type ) {
```
arbron, line 531: Part of the reason I used `Map` for for the additional filter definitions is so it will retain ordering even after the intersection step, so I wonder if we want to handle ordering more explicitly in the filter definitions to give items more control rather than taking it all over here.

module/applications/compendium-browser.mjs
```diff
@@ -325,56 +560,127 @@ export default class CompendiumBrowser extends HandlebarsApplicationMixin(Applic
...
-    this.#results = CompendiumBrowser.fetch(
-      CONFIG[context.filters.documentClass].documentClass,
-      {
-        types: context.filters.types,
-        filters: CompendiumBrowser.applyFilters(context.filterDefinitions, context.filters.additional)
-      }
-    );
+    const filters = CompendiumBrowser.applyFilters(context.filterDefinitions, context.filters.additional);
+    // Add the name filter
+    if ( this.#filters.name?.length ) filters.push({ k: "name", o: "istartswith", v: this.#filters.name });
```
arbron, line 565: Do we want to use an `icontains` filter here instead, so searching for something like "fireball" will return both "Wand of Fireballs" and "Necklace of Fireballs" or searching for "warning" will produce all of the weapons of warning?

arbron (changes requested):
I haven't had a chance to dig into the reason yet, but changing the type in the sidebar frequently doesn't seem to change anything. It triggers the `#onSetType` event, but after re-render nothing is changed. I'll dig into it after my flight home if you haven't gotten to it yet.

module/applications/compendium-browser.mjs
Fyorl: Yeah you're probably right. It would be a bit silly if you couldn't search for 'dragon' and get all the dragons, for example.

module/applications/compendium-browser.mjs
```diff
@@ -325,56 +560,127 @@ export default class CompendiumBrowser extends HandlebarsApplicationMixin(Applic
...
-      );
+    const { documentClass } = this.currentFilters;
+    const results = await this.#results;
+    this.#results = results;
+    const batchEnd = Math.min(this.constructor.BATCHING.SIZE, results.length);
+    for ( let i = 0; i < batchEnd; i++ ) {
+      rendered.push(this._renderResult(results[i], documentClass));
     }
-    this.element.querySelector('[data-application-part="results"] tbody')
+    this.element.querySelector('.results-loading').hidden = true;
```
arbron, line 647:
```suggestion
    this.element.querySelector(".results-loading").hidden = true;
```

module/applications/compendium-browser.mjs
```diff
@@ -325,56 +560,127 @@ export default class CompendiumBrowser extends HandlebarsApplicationMixin(Applic
...
-        types: context.filters.types,
-        filters: CompendiumBrowser.applyFilters(context.filterDefinitions, context.filters.additional)
-      }
-    );
+    const filters = CompendiumBrowser.applyFilters(context.filterDefinitions, context.filters.additional);
+    // Add the name filter
+    if ( this.#filters.name?.length ) filters.push({ k: "name", o: "icontains", v: this.#filters.name });
+    this.#results = CompendiumBrowser.fetch(CONFIG[context.filters.documentClass].documentClass, {
+      filters,
+      types: context.filters.types,
```
arbron, line 568:
```suggestion
      types: context.filters.types
```

module/applications/compendium-browser.mjs
```diff
@@ -200,10 +371,37 @@ export default class CompendiumBrowser extends HandlebarsApplicationMixin(Applic
...
+   * The mode the browser is currently in.
+   * @type {CompendiumBrowser.MODES}
+   */
+  _mode = this.constructor.MODES.BASIC;
+
+  /* -------------------------------------------- */
+
+  /**
+   * The function to invoke when searching results by name.
+   * @type {function}
```
arbron, line 386:
```suggestion
   * @type {Function}
```

module/applications/compendium-browser.mjs
```diff
@@ -128,9 +147,145 @@ export default class CompendiumBrowser extends HandlebarsApplicationMixin(Applic
...
+    /**
+     * The number of pixels before reaching the end of the scroll container to begin loading additional entries.
+     */
+    MARGIN: 50,
+
+    /**
+     * The number of entries to load per batch.
+     */
+    SIZE: 50
+  }
```
arbron, line 279:
```suggestion
  };
```

templates/compendium/browser-sidebar-types.hbs
```diff
@@ -1,20 +1,23 @@
 {{#*inline "type-entry"}}
 <div class="type-entry">
-    <input type="checkbox" value="{{ @key }}" id="{{ @root.partId }}-{{ @key }}" data-action="setType"
-           {{ checked chosen }} {{ disabled @root.isLocked.types }}>
-    <label for="{{ @root.partId }}-{{ @key }}">{{ localize label }}</label>
+    <label>
+        <dnd5e-checkbox value="{{ @key }}" data-action="setType" {{ checked chosen }}
```
arbron, line 4:
```suggestion
        <dnd5e-checkbox value="{{ @key }}" data-action="setType" name="type" {{ checked chosen }}
```

templates/compendium/browser-sidebar-types.hbs
```diff
@@ -1,20 +1,23 @@
...
+<div class="flexcol" {{#unless showTypes}}hidden{{/unless}}>
     {{#each types}}
     {{#if children}}
     <div class="type-group collapsible">
         <div class="type-group-header">
-            <input type="checkbox" id="{{ @root.partId }}-{{ @key }}" data-action="setType"
-                   {{ checked chosen }} {{ disabled @root.isLocked.types }}>
-            <label for="{{ @root.partId }}-{{ @key }}">{{ localize label }}</label>
+            <label>
+                <dnd5e-checkbox data-action="setType" {{ checked chosen }}
```
arbron, line 17:
```suggestion
                <dnd5e-checkbox data-action="setType" name="type" {{ checked chosen }}
```

arbron (changes requested):
To fix the issue with responding to keyboard changes to the new checkbox, we can use the same strategy that `<filter-state>` elements use, giving them a name and then responding to that in `_onChangeForm` to trigger the `#onSetType` method:

```javascript
if ( event.target.name === "type" ) CompendiumBrowser.#onSetType.call(this, event, event.target);
```

arbron:
Noticed you can get into a state where the whole browser disappears with a rendering error if you set an additional filter and then uncheck all of the types in the "Items" tab. Can be fixed by falling back to an empty map for `context.filterDefinitions` in `_prepareContext`.

```javascript
    context.filterDefinitions = dataModels
      .map(([, d]) => d.compendiumBrowserFilters ?? new Map())
      .reduce((first, second) => {
        if ( !first ) return second;
        return CompendiumBrowser.intersectFilters(first, second);
      }, null) ?? new Map();
```

module/applications/components/checkbox.mjs
```diff
@@ -0,0 +1,217 @@
...
+    const { signal } = this._controller;
+    this.addEventListener("click", this._onClick.bind(this), { signal });
+    this.addEventListener("keydown", event => event.key === " " ? this._onClick(event) : null, { signal });
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _onClick(_event) {
+    this.checked = !this.checked;
```
arbron, line 215: Can't say why exactly, but this seems to fix the issue in Firefox.

```suggestion
    _event.preventDefault();
    this.checked = !this.checked;
```

arbron (changes requested):

module/applications/compendium-browser.mjs
```diff
@@ -506,7 +916,7 @@ export default class CompendiumBrowser extends HandlebarsApplicationMixin(Applic
     }
 
     else {
-      for ( const child of target.closest(".type-group").querySelectorAll('input[type="checkbox"][value]') ) {
+      for ( const child of target.closest(".type-group").querySelectorAll("dnd5e-checkbox[value]") ) {
```
arbron, line 919: Small fix to prevent the checkbox from getting stuck in an indeterminate state when you click on it.

```suggestion
      target.indeterminate = false;
      for ( const child of target.closest(".type-group").querySelectorAll("dnd5e-checkbox[value]") ) {
```

arbron (changes requested):

templates/compendium/browser-sidebar-types.hbs
Fyorl: Had to use a different solution as this seemed to cause an infinite loop of `_onChangeForm` -> `#adjustCheckboxStates` -> `set checked` -> `_onChangeForm` -> `#adjustCheckboxStates` -> `...`

arbron (approved):
Looks good, just pushed a small fix for an issue with the "Mundane" option in rarities.

PR #3826, opened 2024-07-17
[#3443] Holding shift as GM shows original delete icon for chat messages

### LINKED ISSUE #3443: Put back the delete (trash can) button in the chat cards
As a GM I am constantly deleting unneeded/unwanted messages and having to click the dots or right click the messages for the context menu is significantly slower than it was previously when there was a delete button directly on the chat cards. Can you please put back, or make it an option in the settings to put back, the delete button that was there in the DND5e V2.x versions of the chat cards?

roth-michael:
Closes #3443 
Happy to make any requested changes.

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -263,9 +263,8 @@ export default class ChatMessage5e extends ChatMessage {
 
     // Context menu
     const metadata = html.querySelector(".message-metadata");
-    let deleteButton = metadata.querySelector(".message-delete");
+    const deleteButton = metadata.querySelector(".message-delete");
     if ( !game.user.isGM ) deleteButton?.remove();
-    deleteButton?.toggleAttribute("hidden", true);
     const anchor = document.createElement("a");
     anchor.setAttribute("aria-label", game.i18n.localize("DND5E.AdditionalControls"));
     anchor.classList.add("chat-control");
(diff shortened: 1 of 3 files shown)


arbron:
@roth-michael Could you make a quick comment on the original issue (https://github.com/foundryvtt/dnd5e/issues/3443) so I can assign it to you?

roth-michael:
Done

roth-michael:
Dunno what the merge process looks like for the system so pinging again here to get that original issue assigned in case it's holding something up.

module/documents/chat-message.mjs
```diff
@@ -263,7 +263,9 @@ export default class ChatMessage5e extends ChatMessage {
 
     // Context menu
     const metadata = html.querySelector(".message-metadata");
-    metadata.querySelector(".message-delete")?.remove();
+    let deleteButton = metadata.querySelector(".message-delete");
+    if (!game.user.isGM) deleteButton?.remove();
+    if (deleteButton) deleteButton.style.display = "none";
```
Fyorl, line 268:
```suggestion
    if ( !game.user.isGM ) deleteButton?.remove();
    if ( deleteButton ) deleteButton.style.display = "none";
```

module/documents/chat-message.mjs
```diff
@@ -738,6 +740,25 @@ export default class ChatMessage5e extends ChatMessage {
...
+      }
+    });
+    document.addEventListener("keyup", event => {
+      if (event.key !== "Shift") return;
+      let deleteButtons = document.querySelectorAll(".chat-message .message-delete");
+      if (!deleteButtons.length) return;
+      if (deleteButtons[0].style.display === "") {
+        deleteButtons.forEach(btn => btn.style.display = "none");
+      }
+    });
```
Fyorl, line 759: Looks good, and I had some suggestions:

1. We should probably make these listeners passive given that they never call `preventDefault`, (i.e. `document.addEventListener("keydown", event => { /* [snip] */ }, { passive: true });`
2. We probably need to add a `blur` listener on the window for edge cases where the user holds shift, clicks outside the window, then releases shift.
3. We could consider using the `hidden` attribute rather than setting `display: none`.
4. If we do use `hidden`, we could use `toggleAttribute`, which allows us to refactor this to share code a bit.

```js
/** JSDoc */
static activateListeners() {
  document.addEventListener("keydown", event => {
    if ( event.key === "Shift" ) this.toggleDeleteButtons(false);
  }, { passive: true });
  document.addEventListener("keyup", event => {
    if ( event.key === "Shift" ) this.toggleDeleteButtons(true);
  }, { passive: true });
  window.addEventListener("blur", () => this.toggleDeleteButtons(true), { passive: true });
}

/** JSDoc */
static toggleDeleteButtons(hide) {
  const deleteButtons = document.querySelectorAll(".chat-message .message-delete");
  deleteButtons.forEach(btn => btn.toggleAttribute("hidden", hide));
}
```

module/documents/chat-message.mjs
```diff
@@ -738,6 +740,25 @@ export default class ChatMessage5e extends ChatMessage {
     }
   }
 
+  static activateListeners() {
```
Fyorl, line 743: Missing JSDoc

Fyorl (changes requested):

module/documents/chat-message.mjs
roth-michael: Implemented these changes & added the missing JSDoc. Also switched earlier code in `_enrichChatCard` to use `hidden` instead of `display: none`.

arbron:
I'm wondering if it might make sense to instead attach this listener to a custom `ChatLog` application. So when you hold down shift the chat log itself gets a class or `data-` attribute indicating that shift is held down. We would then be able to update the CSS on the chat messages to hide/show the delete button, plus if anything else wanted to respond to that button we could easily handle those too.

I know there has been some discussion of communicating on buttons the fast forward mode on buttons when the appropriate modifier is held down, so if we added `data-modifier="shift"` (or `alt` or `control`) then we could use that for other features.

Fyorl (approved):
> I'm wondering if it might make sense to instead attach this listener to a custom `ChatLog` application. So when you hold down shift the chat log itself gets a class or `data-` attribute indicating that shift is held down. We would then be able to update the CSS on the chat messages to hide/show the delete button, plus if anything else wanted to respond to that button we could easily handle those too.

I think adding attributes to the chat log based on the held keys makes sense, it's not clear why it would need to be a custom `ChatLog` application in order to achieve that though.

I've made these changes myself in order to not give our contributor here too much whiplash from the sudden pivot in design.

PR #3866, opened 2024-07-21
[#3865] Add setting for default skills.

### LINKED ISSUE #3865: [Feature Request] Allow for a setting when non-proficient skills are visible for a creature in the new NPC sheet
Currently, only proficient skills are visible for creatures, but it is very common for creatures to have no proficiency in Athlethics/Acrobatics, and have to roll for these skills (grapple/shove). The same goes for Stealth (!) Perception (!) Insight, Deception, and less often for Sleight of Hands,

krbz999:
Adds a setting for skills that appear on NPC sheets regardless of level of proficiency.

Using a `DataField` for the setting, so requires v12.

Closes #3865.

image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2132,8 +2132,6 @@
 "SETTINGS.5eCriticalModifiersL": "Make critical hits more deadly by multiplying non-dice modifiers in addition to rolled dice.",
 "SETTINGS.5eCurWtL": "Carried currency affects character encumbrance following the rules on PHB pg. 143.",
 "SETTINGS.5eCurWtN": "Apply Currency Weight",
-"SETTINGS.5eDefaultSkillsN": "Default Skills",
-"SETTINGS.5eDefaultSkillsH": "The default skills that appear on NPC sheets regardless of level of proficiency.",
 "SETTINGS.5eDiagEuclidean": "Euclidean (7.07 ft. Diagonal)",
 "SETTINGS.5eDiagL": "Configure which diagonal movement rule should be used for games within this system.",
 "SETTINGS.5eDiagN": "Diagonal Movement Rule",
@@ -2190,6 +2188,10 @@
     "Always": "Collapse All",
     "Older": "Collapse Older Trays",
     "Never": "Expand All"
+  },
+  "DEFAULTSKILLS": {
+    "Name": "Default Skills",
+    "Hint": "The default skills that appear on NPC sheets regardless of level of proficiency."
   },
   "THEME": {
     "Name": "Theme",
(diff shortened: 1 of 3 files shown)


module/settings.mjs
```diff
@@ -339,6 +339,22 @@ export function registerSystemSettings() {
...
+
+  if (game.release.generation >= 12) {
+    game.settings.register("dnd5e", "defaultSkills", {
+      name: "SETTINGS.5eDefaultSkillsN",
+      hint: "SETTINGS.5eDefaultSkillsH",
+      type: new foundry.data.fields.SetField(
+        new foundry.data.fields.StringField({
+          choices: () => CONFIG.DND5E.skills
+        }), {
+          type: "checkboxes"
```
krbz999, line 351: (This doesn't do anything. I left it here just to name and shame.)

module/settings.mjs
```diff
@@ -339,6 +339,22 @@ export function registerSystemSettings() {
     type: Boolean,
     default: true
   });
+
+  if (game.release.generation >= 12) {
```
arbron, line 343: If this does go in it probably won't be until `4.0` which is dropping V11 support, so no need to worry about backwards compatibility.

lang/en.json
```diff
@@ -2132,6 +2132,8 @@
 "SETTINGS.5eCriticalModifiersL": "Make critical hits more deadly by multiplying non-dice modifiers in addition to rolled dice.",
 "SETTINGS.5eCurWtL": "Carried currency affects character encumbrance following the rules on PHB pg. 143.",
 "SETTINGS.5eCurWtN": "Apply Currency Weight",
+"SETTINGS.5eDefaultSkillsN": "Default Skills",
+"SETTINGS.5eDefaultSkillsH": "The default skills that appear on NPC sheets regardless of level of proficiency.",
```
arbron, line 2136: We've been adding new settings localization to the `SETTINGS.DND5E` object:

```json
"SETTINGS.DND5E": {
  "DEFAULTSKILLS": {
    "Name": "Default Skills",
    "Hint": "The default skills that appear on NPC sheets regardless of level of proficiency."
  }
}
```

module/settings.mjs
arbron: Is there a core issue filed for this? Probably just a result of the Settings app still being Application V1.

That said, I'm not sure checkboxes would be right here since there are so many skills, it might result in a big wall of checkboxes.

arbron (changes requested):

arbron:
(also, comment on the original issue so I can assign it to you)

module/settings.mjs
krbz999: I had hopes it was relatively uncontroversial.

module/settings.mjs
krbz999: There is not. Though I agree in hindsight it is also best to leave it out until we know what the SettingsConfigurationV2 will look like.

module/settings.mjs
arbron: Well, you can leave these in for now and we'll see what Kim thinks. If it doesn't go in `3.3.1` then I imagine it will have to wait for `4.0` cause I think we want to move onto working on that very quickly due to the hard deadline.

arbron (approved):

Fyorl (approved):

PR #3872, opened 2024-07-22
[#2201, #2550, #2551] Uncap XP tracking & add epic boon calculation

### LINKED ISSUE #2201: Make Lvl 20 allow more XP accumulation
Currently if I assign XP via whichever means are available to characters that are Lvl20 or higher, the XP just gets reset to the value for Lvl20, which makes Epic campaigns challenging.

### LINKED ISSUE #2550: Remove maximum XP at level 20
Currently the XP tracker still displays the maximum value once you reach level 20 (or whatever the max level you have defined in config). It might be cleaner to replace that with `∞` to indicate there is no longer a next level to achieve.

### LINKED ISSUE #2551: Add epic boons system option with tracking based on XP
The DMG suggests a system of epic boons that are granted to level 20 characters for every 30,000 xp they earn beyond 355,000, the amount needed to get to level 20. I propose adding an option to settings to enable tracking of XP increments past the max to indicate how many epic boons a player might have earned.

arbron:
Removes the cap for maximum XP. When the player's level is maximum it will calculate the number of "boons earned", which means the number of 30,000 chunks of XP they character has earned above the maximum XP. This will be displayed on the sheet instead of the XP threshold for the next level.

Screenshot 2024-07-22 at 16 28 08

The XP required to earn an epic boon can be set using `CONFIG.DND5E.epicBoonInterval`.

Closes #2201 
Closes #2550
Closes #2551


krbz999:
I think it would be good to adjust the XP setting as well to be a selection between "XP", "No XP" (as is current) but also "XP + Boons".

Fyorl (changes requested):
Looks good, thank you. Please see my comments. I also updated the design with a boon badge:

image

Fyorl (approved):
I had a last-minute idea to format the boon levels as roman numerals and have pushed that change. We can take it or leave it, let me know what you think.

image

PR #3876, opened 2024-07-23
[#2010] Add bloodied status effect that auto-applies at 50% HP

### LINKED ISSUE #2010: Option to display wounded/bloodied state
The DMG suggests in chapter 8 Running the Game to indicate when monsters drop below 50% health:

> Players often ask how hurt a monster looks. Don’t ever feel as though you need to reveal exact hit points, but if a monster is below half its hit point maximum, it’s fair to say that it has visible wounds and appears beaten down. You can describe a monster taken to half its hit points as bloodied, giving the players a sense of progress in a fight against a tough opponent, and helping them judge when to use their most powerful spells and abilities.

There should be an option to dispaly that without modules. A step further would be options to have that for players, and for players uncoscious at 0hp, too, for group playing without the PCs knowing each others values precisely, but still making the healers' job playable without extensive talking about the players' characters values.

arbron:
Similar to the encumbered status effects, this adds a hidden status effect that is applied whenever an actor drops to half their hit points or lower.

Closes #2010


krbz999:
This should be optional.

kaelad02:
Is there a similar issue/PR for automatically marking tokens Dead/Unconscious when reduced to 0 HP? I'd volunteer to make a PR for that... probably easier to distinguish who should get Unconscious now that NPCs have that flag to enable death saving throws.

krbz999:
> Is there a similar issue/PR for automatically marking tokens Dead/Unconscious when reduced to 0 HP? I'd volunteer to make a PR for that... probably easier to distinguish who should get Unconscious now that NPCs have that flag to enable death saving throws.

Doesn't look like it. It could probably build on top of this but would require a setting to adjust which actors (per item type?) automatically get marked as dead/unconscious. I had a lot of success only doing such a thing automatically for NPCs (characters have too many options for dropping to 1 hit point instead), plus auto-death or auto-unconscious would be even more controversial than auto-bloodied.

module/documents/actor/actor.mjs
```diff
@@ -3453,6 +3454,29 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+   * Handle applying/removing the bloodied status.
+   * @param {DocumentModificationContext} options  Additional options supplied with the update.
+   * @returns {Promise<ActiveEffect>|void}
+   */
+  updateBloodied(options) {
+    const hp = this.system.attributes?.hp;
+    if ( !hp?.effectiveMax || (game.settings.get("dnd5e", "bloodied") === "none") ) return;
+
+    const effect = this.effects.get(ActiveEffect5e.ID.BLOODIED);
+    if ( hp.value > hp.effectiveMax * CONFIG.DND5E.bloodied.threshold ) return effect?.delete();
```
kaelad02, line 3467: I think we should add a check for 0 HP here too. Meaning if the actor has 0 HP then remove the bloodied effect. Doesn't make sense to me to be bloodied if you're already unconscious or dead.

module/documents/actor/actor.mjs
```diff
@@ -3453,6 +3454,29 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    if ( effect ) return;
+
+    return ActiveEffect.implementation.create({
+      _id: ActiveEffect5e.ID.BLOODIED,
+      name: game.i18n.localize(CONFIG.DND5E.bloodied.name),
+      icon: CONFIG.DND5E.bloodied.icon,
```
krbz999, line 3473:
```suggestion
      img: CONFIG.DND5E.bloodied.icon,
```

tposney:
It would be nice if the threshold is configurable, rather than being a fixed percentage. Currently you'd have to use a script to modify CONFIG.DND5E.bloodied.threshold.

kaelad02:
> It would be nice if the threshold is configurable, rather than being a fixed percentage. Currently you'd have to use a script to modify CONFIG.DND5E.bloodied.threshold.

I disagree. We're finding out this is coming from the 2024 rules and at least it's externalized in a `CONFIG.DND5E` variable. IMHO we don't need to make so many settings, the variable is good enough.

module/documents/actor/actor.mjs
Fyorl: Nothing about the definition of bloodied implies you would lose it at 0 HP. A character at 0 HP, whether they are stable or not, could very well be subjected to an effect where their bloodied condition may be relevant.

Fyorl (approved):

Issue #1192, opened 2021-07-06
Suggestion: hide Dexterity scores at Combat Tracker when Initiative Dexterity Tiebreaker is activated


aaclayton:
Originally in GitLab by @paulohsc

Although Initiative Dexterity Tiebreaker is a very useful feature, the fact that the Dexterity ability scores of all combatants (even non controlled ones) is displayed on the Combat Tracker for all player gives a spoiler that can lead to metagaming. For example, a player can choose to use (or avoid using) a spell/effect resisted by Dexterity on an opponent based on its Dexterity score, as showed at the Combat Tracker.

I would suggest that Dexterity ability scores were hidden by default at Combat Tracker; or, alternatively, that there was an option at the Combat Tracker Settings to hide such scores.

PR #3880, opened 2024-07-23
[#1192] Hide dex tiebreakers in combat tracker

arbron:
Modifies context preparation on CombatTracker5e to drop any decimal digits after the initiative value.


kaelad02:
I turn on the Initiative Dexterity Tiebreaker setting in my game but never liked the information leakage. I've been tempted to fix this myself but never liked the solutions I came up with. I was thinking of ways to change the sorting algorithm but this is much cleaner. Keep the underlying data the same but just don't show it by rounding the number when prepping the data.

Fyorl (approved):

PR #3884, opened 2024-07-24
Document the Enchantment active effects

kaelad02:
Similar to the Active Effect Guide, add a section to the Enchantment wiki page to document the active effects. It has some common sections (i.e. Common Examples, Common Usage Examples, and Common Action Examples) plus sections for those item types that have other things in their data model an enchantment could realistically change. Not surprisingly, the Loot type doesn't have its own section since it doesn't have much to begin with and is covered by the Common Examples section.

There are some parts of the data model I decided to leave out since I didn't think they made sense to change in an enchantment or were very unlikely to be changed:

- `system.source`
- `system.type`
- `system.identified` and `unidentified`
- `system.quantity`, `weight`, `price`, and `rarity`
- `system.attunement`, `attuned`, and `equipped`
- anything only shown on vehicle items (e.g. `system.armor` on weapons)

### DIFF
--- a/wiki/Enchantment.md
+++ b/wiki/Enchantment.md
@@ -39,3 +39,430 @@ Once configuration is complete, enchantment is very easy. Simply activate the fe
 ![Enchantment Chat Message](https://raw.githubusercontent.com/foundryvtt/dnd5e/publish-wiki/wiki/images/enchantment/enchantment-chat-message.jpg)
 
 In the activation chat message a drop area will appear. Any player may drop an item in this area to enchant that item and the enchanted items will be listed. GMs or the player who owns the item will have access to the "Remove Enchantment" button in the chat card which will quickly remove the enchantment. Enchantments can also be removed by breaking concentration if they are from a concentration effect, or by enter the "Effects" tab on the item and manually deleting the enchantment.
+
+## Enchantment Active Effects
+
+This describes how to create Active Effects used as an Enchantment. This differs from the [Active Effect Guide](Active-Effect-Guide) in that these are changes to items instead of the actors. If you're creating an Active Effect to add to the Additional Effects section when configuring an Enchantment, then use that guide.
+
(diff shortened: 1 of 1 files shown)


wiki/Enchantment.md
```diff
@@ -39,3 +39,380 @@ Once configuration is complete, enchantment is very easy. Simply activate the fe
 ![Enchantment Chat Message](https://raw.githubusercontent.com/foundryvtt/dnd5e/publish-wiki/wiki/images/enchantment/enchantment-chat-message.jpg)
 
 In the activation chat message a drop area will appear. Any player may drop an item in this area to enchant that item and the enchanted items will be listed. GMs or the player who owns the item will have access to the "Remove Enchantment" button in the chat card which will quickly remove the enchantment. Enchantments can also be removed by breaking concentration if they are from a concentration effect, or by enter the "Effects" tab on the item and manually deleting the enchantment.
+
+## Enchantment Active Effects
+
+This describes how to create Active Effects used as an Enchantment. This differs from the [Active Effect Guide](Active-Effect-Guide) in that these are changes to items instead of the actors. If you're creating an Active Effect to add to the Additional Effects section when configuring an Enchantment, then use that guide.
```
Fyorl, line 45:
```suggestion
This describes how to create Active Effects used as Enchantments. This differs from the [Active Effect Guide](Active-Effect-Guide) in that these are changes to items instead of the actors. If you're creating an Active Effect to add to the Additional Effects section when configuring an Enchantment, then use that guide.
```

Fyorl (changes requested):
Can we bump the 'up to date as of 3.2.0' at the top of the page as well?

wiki/Enchantment.md
kaelad02: done

Fyorl (approved):

Issue #1964, opened 2022-12-15
Proposal: Multiple Activities on Items


arbron:
For awhile there has been a need to represent multiple different actions that can be handled by a single item. Some of this is currently hardcoded into the system (aka Normal vs. Versatile damage), some of it has been handled by modules (such as [Roll Groups](https://foundryvtt.com/packages/rollgroups) and [Items with Spells](https://foundryvtt.com/packages/items-with-spells-5e)), and sometimes we just recommend users create two identical items with slightly different properties.

This proposal is a way to add different activities to an item in a way that is easy to use, powerful, and fully extensible by modules. It is build around a system similar to the way that the Advancement system is currently build, a series of pseudo-documents stored within the item representing different things the item can do. Some of these activity types would be things that are handled currently like *Attack*, *Saving Throw*, or *Healing* while others would add new functionality such as *Cast a Spell* or *Change State*. Modules could define their own activity types with their own data structures, interface, and chat actions. So my module [Arbron's Summoning](https://foundryvtt.com/packages/arbron-summoner) could define a "Summon" activity type that would behave very similar to the how the module does now, but be fully integrated into the item.

### Attack
This would handle all attack actions. It would have configuration defining whether it is a "melee" or "ranged" attack, and whether it is "unarmed", "weapon", or "spell". It could also have additional configuration two assist with automatically determining display name and attack values such as whether it is a "one handed", "two handed", "off hand", or "thrown" attack. Like the existing attack action the user will be able to select an associated ability (or set it to automatic or flat), define a to-hit bonus, and define multiple damage formulas.

The damage formula format would be modified to allow for richer data, including adding custom critical damage to specific damage types rather than to the attack as a whole. It would also potentially support damage conditions, which would allow for a weapon such as a Flame-tongue Longsword to only do its fire damage if it is in the "Burning" state.

### Saving Throw
This would allow for defining what ability is used to set the DC and what is needed to save and define multiple damage formulas. What occurs when the save succeeds would be specified in a way that modules like MidiQOL could easily read, but wouldn't force any sort of automation into core.

In addition to damage, it would also be able to specify a condition that is applied to the target when they fail the save.

### Ability Check
[...]

PR #3892, opened 2024-07-25
[#1964] Implement data structures for storing activities

arbron:
Introduces the templates and data structures for storing, fetching, and updating activities on items.

- `ActivityTemplate`: Adds activities and uses to items
- `ActivitiesField`: Mapping field that stores activities and initializes as a custom `Collection`.
- `ActivityField`: Field inside the mapping field that stores an individual activity, initializing as a data model of the correct type.
- `ActivityCollection`: Subclass of `Collection` that organizes activities and provides a `getByType` method to retrieve a list of activities of a certain type.

In the process this splits `fields.mjs` into separate files because the original file was getting quite long.

-----------------

This introduces the base activity type, `UtilityActivity`, which doesn't have much behavior of its own and can be used for when an item doesn't have a more specific activity it requires. For example the spell "Mage Armor" would have a `UtilityActivity` that is just responsible for applying the active effect to the caster.

In order to support the necessary inheritance structure, the base behavior of all activities is implemented in `ActivityMixin`. This mixin doesn't have much behavior now, but will eventually handle activing the activity and consuming uses.

-----------------

A few modifications have been made to `Item5e` in order to support CRUD operations on activities using the same API that is used for other documents (e.g. calling
`item.createEmbeddedDocuments("Activity", [{ type: "utility }])` to create a new activity). This involved adding some additional behavior to `getCollectionName` and `getEmbeddedCollection` to recognize special collections defined withing system data.

In addition, `ActivityMixin` inherits from `PseudoDocumentMixin`, a new set of base methods to implement these CRUD operations. Eventually `Advancement` should be modifies to use this new setup so that the `createAdvancement` and related methods on `Item5e` can be removed.


module/data/item/templates/activities.mjs
```diff
@@ -0,0 +1,71 @@
...
+    else if ( source.recharge?.value ) {
+      source.uses.recovery = [{ period: "recharge", formula: source.recharge.value }];
+      source.uses.max = 1;
+      source.uses.spent = source.recharge.charged ? 0 : 1;
+      return;
+    }
+
+    else return;
+
+    if ( ("value" in source) && Number.isNumeric(source.uses.max) ) {
```
krbz999, line 67: "value" in source.uses?

Fyorl (changes requested):
Before I continue with review, I wanted to comment on the high-level direction of this PR. I really don't think we should be going to all this effort to make activities behave like pseudo embedded documents. It seems like a lot of extra complexity and overhead, without obvious tangible benefits. Can we stick with a simple `CONFIG.DND5E`-based registration, and some `Item5e` helper methods like `createActivity`, `updateActivity`, and `deleteActivity`, until it becomes more clear that we would need more complex things like pre- and post- CRUD handlers.

arbron:
> Before I continue with review, I wanted to comment on the high-level direction of this PR. I really don't think we should be going to all this effort to make activities behave like pseudo embedded documents. It seems like a lot of extra complexity and overhead, without obvious tangible benefits. Can we stick with a simple `CONFIG.DND5E`-based registration, and some `Item5e` helper methods like `createActivity`, `updateActivity`, and `deleteActivity`, until it becomes more clear that we would need more complex things like pre- and post- CRUD handlers.

The main advantage is no having to figure out a bespoke API just for activities. If you know how to use the existing `createEmbeddedDocuments` API that core offers then you can use that to create and manipulate activities rather than having to find the `createActivity` API and figure out how to call it.

krbz999:
> rather than having to find the `createActivity` API and figure out how to call it.

Not gonna lie, that isn't a big ask.

Fyorl:
> The main advantage is no having to figure out a bespoke API just for activities. If you know how to use the existing `createEmbeddedDocuments` API that core offers then you can use that to create and manipulate activities rather than having to find the `createActivity` API and figure out how to call it.

It's an awful lot of plumbing just for that. We also have to keep the API surface in-sync whenever the core embedded Document API changes. It doesn't seem worth it in order to enable calling `item.createEmbeddedDocument("Activity")` vs. `item.createActivity()`.

I may not be understanding the full extent of the proposal or plans, but it doesn't seem like complex manipulation of activities is necessarily required in order to achieve our goals with them. Basic CRUD operations will be limited to the item sheet creating, editing, and deleting activities. Then the majority of their implementation complexity comes from what happens when  an activity is invoked, which seems mostly tangential to their CRUD workflow to me.

module/data/item/templates/activities.mjs
```diff
@@ -0,0 +1,71 @@
...
+      source.uses.spent = Number(source.uses.max) - source.uses.value;
```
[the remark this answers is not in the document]
arbron: What I'm thinking at this point is we do a concrete migration to handle translating `value` to `spent`, and leave the rest of the changes up to in-memory migration. That should allow us to keep the proper available uses for everything in the world where it matters most (since stuff in compendiums will almost always have `0` uses spent).

module/data/item/templates/activities.mjs
```diff
@@ -0,0 +1,71 @@
...
+  /** @override */
+  static LOCALIZATION_PREFIXES = ["DND5E.USES"];
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  static defineSchema() {
+    return {
+      activities: new ActivitiesField(),
+      uses: new UsesField()
```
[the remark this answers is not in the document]
arbron: Added back in `autoDestroy` and removed `recharge` with compatibility warnings.

I figure the rest of the data can be transitioned out once larger migrations are in place in future PRs.

module/documents/activity/mixin.mjs
```diff
@@ -0,0 +1,41 @@
...
+   * @property {string} title  Default title.
+   */
+
+  /**
+   * Configuration information for this PseudoDocument.
+   * @type {PseudoDocumentsMetadata}
+   */
+  static metadata = Object.freeze({
+    name: "Activity",
+    collection: "activities"
```
Fyorl, line 24: Probably not used anymore.

module/data/fields/activities-field.mjs
```diff
@@ -0,0 +1,140 @@
+import BaseActivityData from "../activity/base-activity.mjs";
+import MappingField from "./mapping-field.mjs";
+
+/**
+ * Field that stories activities on an item.
```
Fyorl, line 5:
```suggestion
 * Field that stores activities on an item.
```

module/data/fields/activities-field.mjs
```diff
@@ -0,0 +1,140 @@
...
+ */
+export class ActivitiesField extends MappingField {
+  constructor(options) {
+    super(new ActivityField(), options);
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static hierarchical = true;
```
Fyorl, line 15: Probably not needed anymore.

module/data/fields/activities-field.mjs
```diff
@@ -0,0 +1,140 @@
...
+    if ( cls ) return cls.cleanData(value, options);
+    return value;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  initialize(value, model, options = {}) {
+    const cls = this.getModel(value);
+    if (cls) return new cls(value, { parent: model, ...options });
```
Fyorl, line 62:
```suggestion
    if ( cls ) return new cls(value, { parent: model, ...options });
```

module/data/fields/activities-field.mjs
```diff
@@ -0,0 +1,140 @@
...
+  migrateSource(sourceData, fieldData) {
+    const cls = this.getModel(fieldData);
+    if ( cls ) cls.migrateDataSafe(fieldData);
+  }
+}
+
+/* -------------------------------------------- */
+
+/**
+ * Specialized collection type for storied activities.
```
Fyorl, line 82:
```suggestion
 * Specialized collection type for stored activities.
```

module/data/fields/activities-field.mjs
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Convert the ActivityCollection to an array of simple objects.
+   * @param {boolean} [source=true]  Draw data for contained Documents from the underlying data source?
+   * @returns {object[]}             The extracted array of primitive objects.
+   */
+  toObject(source = true) {
```
Fyorl, line 137:
```suggestion
  toObject(source=true) {
```

module/data/fields/activities-field.mjs
```diff
@@ -0,0 +1,140 @@
...
+export class ActivityCollection extends Collection {
+  constructor(model, entries) {
+    super();
+    this.#model = model;
+    for ( const [id, entry] of Object.entries(entries) ) {
+      if ( !(entry instanceof BaseActivityData) ) continue;
+      this.set(id, entry);
+      if ( !this.#types.has(entry.type) ) this.#types.set(entry.type, []);
+      this.#types[entry.type] ??= [];
+      this.#types.get(entry.type).push(entry);
```
Fyorl, line 95: Doing this in the constructor is not enough to keep `#types` in-sync, we'd need to override `set`, and `delete` as well. Or, if we don't intend the Collection to be used in such a way, then we could override them to throw Errors, but it's not really clear why we'd use a Collection in the first place if that's the case.

module/documents/mixins/pseudo-document.mjs
```diff
@@ -0,0 +1,314 @@
+/**
+ * A mixin which extends a DataModel to provide behavior shared between activities & advancements.
+ * @param {typeof DataModel} Base  The base DataModel to be mixed.
+ * @returns {typeof PseudoDocument}
+ * @mixin
+ */
+export default Base => class extends Base {
+  constructor(data, { parent = null, ...options } = {}) {
+    if (parent instanceof Item) parent = parent.system;
```
Fyorl, line 9:
```suggestion
  constructor(data, { parent=null, ...options }={}) {
    if ( parent instanceof Item ) parent = parent.system;
```

module/documents/mixins/pseudo-document.mjs
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Lazily obtain a ApplicationV2 instance used to configure this PseudoDocument, or null if no sheet is available.
+   * @type {ApplicationV2|null}
+   */
+  get sheet() {
+    // TODO: Implement when pseudo document sheet registration is set up
```
Fyorl, line 145: I don't think we need to implement the ability to configure Activity sheets.

module/documents/mixins/pseudo-document.mjs
```diff
@@ -0,0 +1,314 @@
...
+    this._apps.get(doc?.uuid)?.delete(app);
+  }
+
+  /* -------------------------------------------- */
+  /*  Editing Methods                             */
+  /* -------------------------------------------- */
+
+  /**
+   * Update this PseudoDocument.
+   * @param {string} id                  ID of the PseudoDocument to update.
```
Fyorl, line 191: This parameter is unused

Fyorl (changes requested):

module/data/fields/activities-field.mjs
```diff
@@ -0,0 +1,151 @@
...
+   * The parent DataModel to which this ActivityCollection belongs.
+   * @type {DataModel}
+   */
+  #model;
+
+  /* -------------------------------------------- */
+
+  /**
+   * Pre-organized arrays of activities by type.
+   * @type {Map<string, Activity[]>}
```
Fyorl, line 105: We could consider making this a `Map<string, Set<Activity>>` to make the `set` and `delete` implementations a little more efficient at the cost of making `getByType` a little less efficient (assuming it continues to return an array and not a `Set`).

Fyorl (approved):

PR #3904, opened 2024-07-30
[#1964] Add basic activity configuration sheet

arbron:
Adds a new ApplicationV2-based sheet for editing activities. This sheet is designed to work with the existing `UtilityActivity` and can be easily subclassed for other activty types.

Note: Due to the ongoing item sheet re-design, no work has been done to design this sheet to match the system design, only to get the underlying functionality and UX working.

The sheet consists of three primary tabs, the first of which is the Identity tab containing the activity name and icon. This tab will probably be eventually expanded to include a description and visibility settings once they are added.

Activity Sheet v1 - Identity Tab

The Activation tab contains the bulk of the details with three sub-tabs. The Time tab contains activation time and condition and the duration of the activity.

Activity Sheet v1 - Activation Time Tab

The Consumption tab contains a list of consumption targets, allowing any number of consumption targets to be configured and supporting scaling of consumption. The scaling section indicates whether scaling is allowed (beyond default scaling from spells) and allows for setting a scaling cap. Uses indicates the spent uses and a formula for the maximum uses. Use recovery is another list of recovery profiles, allowing for complex recovery such as recovering one charge on a short rest and all on a long rest.

Activity Sheet v1 - Activation Consumption Tab

Finally the Targeting tab contains details on the range, affected targets, and area of effect. For area of effect, the dimensions displayed and their labels will change depending on the selected area type so you only see the relevant sizes.

Activity Sheet v1 - Activation Targeting Tab

The final primary tab, Effect, isn't currently implemented. For the base activity sheet this tab will contain controls for selecting what active effects can be applied by this activity. For other sheets this will display details on the type of saving throw the must be made, attack to hit modifiers, the damage to apply, creatures to summon, or enchantments to apply.

### ToDo
- [ ] Implement active effects controls on Effects tab
- [ ] Add "Recharge" recovery period & dropdown for selecting recharge roll target
- [ ] Add options for "Attribute" consumption type when on actor
- [ ] Handle displaying options that have been inferred from the containing item

### DIFF
--- a/module/applications/activity/activity-sheet.mjs
+++ b/module/applications/activity/activity-sheet.mjs
@@ -1,0 +1,494 @@
+import Application5e from "../api/application.mjs";
+
+/**
+ * Default sheet for activities.
+ */
+export default class ActivitySheet extends Application5e {
+  constructor(options={}) {
+    super(options);
+    this.#activityId = options.document.id;
+    this.#item = options.document.item;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  static DEFAULT_OPTIONS = {
+    classes: ["activity", "sheet"],
+    tag: "form",
+    document: null,
+    viewPermission: CONST.DOCUMENT_OWNERSHIP_LEVELS.LIMITED,
+    editPermission: CONST.DOCUMENT_OWNERSHIP_LEVELS.OWNER,
+    window: {
+      icon: "fa-solid fa-gauge"
+    },
+    actions: {
+      addConsumption: ActivitySheet.#addConsumption,
+      addRecovery: ActivitySheet.#addRecovery,
+      deleteConsumption: ActivitySheet.#deleteConsumption,
+      deleteRecovery: ActivitySheet.#deleteRecovery
+    },
+    form: {
+      handler: ActivitySheet.#onSubmitForm,
+      submitOnChange: true
+    },
+    position: {
+      width: 540,
+      height: "auto"
+    }
+  };
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static PARTS = {
+    tabs: {
+      template: "templates/generic/tab-navigation.hbs"
+    },
+    identity: {
+      template: "systems/dnd5e/templates/activity/identity.hbs"
+    },
+    activation: {
+      template: "systems/dnd5e/templates/activity/activation.hbs",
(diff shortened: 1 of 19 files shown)


module/applications/activity/activity-sheet.mjs
```diff
@@ -0,0 +1,494 @@
...
+  /**
+   * Handle adding a new entry to the consumption list.
+   * @this {ActivityConfig}
+   * @param {Event} event         Triggering click event.
+   * @param {HTMLElement} target  Button that was clicked.
+   */
+  static #addConsumption(event, target) {
+    const types = this.activity.validConsumptionTypes;
+    const existingTypes = new Set(this.activity.consumption.targets.map(t => t.type));
+    const filteredTypes = types.difference(existingTypes);
```
Fyorl, line 391: Feels like this logic makes a bit less sense for consumption targets. I would have thought a pretty common use case would be to, for example, -1 charges from one item, and +1 charges to another item. I don't feel strongly though.

module/applications/activity/activity-sheet.mjs
```diff
@@ -0,0 +1,494 @@
...
+  /**
+   * Perform any pre-processing of the form data to prepare it for updating.
+   * @param {SubmitEvent} event          Triggering submit event.
+   * @param {FormDataExtended} formData  Data from the submitted form.
+   * @returns {object}
+   */
+  _prepareSubmitData(event, formData) {
+    const submitData = foundry.utils.expandObject(formData.object);
+    if ( foundry.utils.hasProperty(submitData, "consumption.targets") ) {
+      submitData.consumption.targets = Object.values(submitData.consumption.targets);
```
Fyorl, line 476: I think this might be handled automatically by the DataModel now, see `ArrayField#_cast`. But I'm not sure, so this is fine to leave as well.

Fyorl (approved):
Nice work, thank you Jeff. I have left some comments but they are all just suggestions.

module/applications/activity/activity-sheet.mjs
arbron: It only seemed to be partially working for me, I would be able to update the `<select>` inputs perfectly find, but as soon as I entered something into the associated text `<input>` the select value would get reset. I couldn't quite figure out what was happening but explicitly casting the data to arrays before saving fixed it.

PR #3907, opened 2024-07-31
[#1964] Add applied effects data to activities & sheet

arbron:
Adds list of effects for an activity to apply. These effects are stored as an array of objects with IDs in case the system needs something like level restrictions in the future (like with enchantments).

Configuring these applied effects is handled through the Effect tab of the activity sheet. There is a list of currently applied effects (unstyled currently) with controls for opening and deleting effects. There is a button for adding new effects directly rather than returning to the item sheet, and a `<multi-select>` allowing for effects that are already on the item to be selected.

### DIFF
--- a/module/applications/activity/activity-sheet.mjs
+++ b/module/applications/activity/activity-sheet.mjs
@@ -24,8 +24,10 @@ export default class ActivitySheet extends Application5e {
     },
     actions: {
       addConsumption: ActivitySheet.#addConsumption,
+      addEffect: ActivitySheet.#addEffect,
       addRecovery: ActivitySheet.#addRecovery,
       deleteConsumption: ActivitySheet.#deleteConsumption,
+      deleteEffect: ActivitySheet.#deleteEffect,
       deleteRecovery: ActivitySheet.#deleteRecovery
     },
     form: {
@@ -59,7 +61,10 @@ export default class ActivitySheet extends Application5e {
       ]
     },
     effect: {
-      template: "systems/dnd5e/templates/activity/effect.hbs"
+      template: "systems/dnd5e/templates/activity/effect.hbs",
+      templates: [
+        "systems/dnd5e/templates/activity/parts/activity-effects.hbs"
+      ]
     }
   };
 
@@ -277,6 +282,12 @@ export default class ActivitySheet extends Application5e {
    */
   async _prepareEffectContext(context) {
     context.tab = context.tabs.effect;
+
+    const appliedEffects = new Set(context.activity.effects.map(e => e.id));
+    context.allEffects = this.item.effects.map(effect => ({
+      value: effect.id, label: effect.name, selected: appliedEffects.has(effect.id)
+    }));
+
     return context;
   }
 
@@ -358,7 +369,7 @@ export default class ActivitySheet extends Application5e {
   /** @override */
   _canRender(options) {
     if ( !this.isVisible ) throw new Error(game.i18n.format("SHEETS.DocumentSheetPrivate", {
-      type: game.i18n.localize("DND5E.ACTIVITY.Label")
+      type: game.i18n.localize("DND5E.ACTIVITY.Title.one")
     }));
   }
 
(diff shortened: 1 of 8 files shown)


module/applications/activity/activity-sheet.mjs
```diff
@@ -399,6 +419,25 @@ export default class ActivitySheet extends Application5e {
...
+  /**
+   * Handle creating a new active effect and adding it to the applied effects list.
+   * @this {ActivityConfig}
+   * @param {Event} event         Triggering click event.
+   * @param {HTMLElement} target  Button that was clicked.
+   */
+  static async #addEffect(event, target) {
+    const effectData = {
+      name: this.item.name,
+      icon: this.item.img,
```
krbz999, line 431:
```suggestion
      img: this.item.img,
```

module/applications/activity/activity-sheet.mjs
```diff
@@ -399,6 +419,25 @@ export default class ActivitySheet extends Application5e {
...
+      origin: this.item.uuid,
```
Fyorl, line 432: It seems a bit strange to set the origin to the same Item that the ActiveEffect is already on, no?

module/applications/activity/activity-sheet.mjs
```diff
@@ -435,6 +474,23 @@ export default class ActivitySheet extends Application5e {
...
+  /**
+   * Handle deleting an active effect and removing it from the applied effects list.
+   * @this {ActivityConfig}
+   * @param {Event} event         Triggering click event.
+   * @param {HTMLElement} target  Button that was clicked.
+   */
+  static async #deleteEffect(event, target) {
+    const effectId = target.closest("[data-effect-id]")?.dataset.effectId;
+    const result = await this.item.effects.get(effectId)?.deleteDialog();
+    if ( result !== false ) {
```
Fyorl, line 486: If the dialog is closed without choosing yes or no, I believe it returns `null`, so this code will proceed in that case.

module/documents/mixins/pseudo-document.mjs
```diff
@@ -159,7 +159,9 @@ export default Base => class extends Base {
    * @param {RenderOptions} [options]  Rendering options.
    */
   render(options) {
-    for ( const app of this.constructor._apps.get(this.uuid) ?? [] ) app.render(options);
+    for ( const app of this.constructor._apps.get(this.uuid) ?? [] ) {
+      app.render({ window: { title: app.title }, ...options });
```
Fyorl, line 163: Found a bug?

templates/activity/parts/activity-effects.hbs
```diff
@@ -0,0 +1,20 @@
+<fieldset>
+    <legend>{{ localize "DND5E.ACTIVITY.FIELDS.effects.label" }}</legend>
+    <multi-select name="appliedEffects">
```
Fyorl, line 3: I just want to make sure I understand the UI here: If you choose an effect from this multi-select, the app re-renders and that effect appears immediately in the list below? Or it has to wait for some other re-render action before it appears?

Fyorl (changes requested):

module/applications/activity/activity-sheet.mjs
arbron: Well that's what we do when creating effects on items at the moment, though presumable this will be set again when the effect is applied through the chat message.

module/documents/mixins/pseudo-document.mjs
arbron: Seems like, the window title doesn't update unless you explicitly pass in `window` to the options (`_updateFrame` checks for that explicitly).

templates/activity/parts/activity-effects.hbs
arbron: Yep, the change event on the `<mutli-select>` triggers form submission causing the newly selected effect to appear below.

module/applications/activity/activity-sheet.mjs
Fyorl: Ah, I guess so that it's set to the correct Item if you drop it onto another one.

templates/activity/parts/activity-effects.hbs
Fyorl: OK great, that sounds good to me then

Fyorl (approved):

PR #3915, opened 2024-07-31
[#1964] Add attribute & material consumption & recharge recovery

arbron:
Adds full support for configuring attribute & material consumption including a list of targets when embedded. Also adjusts how item uses targest are displayed to list charges remaining or for a period in the same way as they are presented on the current item sheet.

Adds the recharge recovery period that displays a fixed list of recharge rolls (e.g. "Recharge 4–6") to select.

### DIFF
--- a/module/applications/item/item-sheet.mjs
+++ b/module/applications/item/item-sheet.mjs
@@ -146,7 +146,6 @@ export default class ItemSheet5e extends ItemSheet {
 
       concealDetails: !game.user.isGM && (this.document.system.identified === false)
     });
-    context.abilityConsumptionTargets = this._getItemConsumptionTargets();
     if ( !item.isEmbedded && foundry.utils.isEmpty(context.abilityConsumptionTargets) ) {
       context.abilityConsumptionHint = (this.item.system.consume?.type === "attribute")
         ? "DND5E.ConsumeHint.Attribute" : "DND5E.ConsumeHint.Item";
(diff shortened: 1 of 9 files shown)


module/data/activity/base-activity.mjs
```diff
@@ -155,6 +156,7 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
   prepareData() {
     this.name = this.name || game.i18n.localize(this.metadata?.title);
     this.img = this.img || this.metadata?.img;
+    UsesField.prepareData.call(this, this.getRollData({ determinstic: true }));
```
krbz999, line 159:
```suggestion
    UsesField.prepareData.call(this, this.getRollData({ deterministic: true }));
```

module/data/shared/uses-field.mjs
```diff
@@ -32,7 +33,22 @@ export default class UsesField extends SchemaField {
    * @param {object} rollData
    */
   static prepareData(rollData) {
+    console.log("UsesField#prepareData");
```
krbz999, line 36: Gezundheit.

module/applications/item/item-sheet.mjs
```diff
@@ -269,77 +268,6 @@ export default class ItemSheet5e extends ItemSheet {
...
-   * @returns {Object<string>}   An object of potential consumption targets
-   * @private
-   */
-  _getItemConsumptionTargets() {
-    const consume = this.item.system.consume || {};
-    if ( !consume.type ) return [];
-    const actor = this.item.actor;
-    if ( !actor && (consume.type !== "hitDice") ) return {};
-
-    // Ammunition
```
Fyorl, line 283: How are we planning on handling ammo?

module/data/activity/base-activity.mjs
```diff
@@ -206,16 +210,39 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
...
    */
   static validItemUsesTargets() {
+    const makeLabel = (name, item) => {
+      let label;
+      const uses = item.system.uses;
+      if ( uses.max && (uses.recovery?.length === 1) && (uses.recovery[0].type === "recoverAll") ) {
+        const per = CONFIG.DND5E.limitedUsePeriods[uses.recovery[0].period]?.abbreviation;
+        label = game.i18n.format("DND5E.AbilityUseConsumableLabel", { max: uses.max, per });
+      }
+      else if ( uses.value ) label = game.i18n.format("DND5E.AbilityUseChargesLabel", { value: uses.value });
```
Fyorl, line 220: We should probably still use the charges label even if there are no charges currently left.

module/applications/item/item-sheet.mjs
```diff
@@ -269,77 +268,6 @@ export default class ItemSheet5e extends ItemSheet {
 
   /* -------------------------------------------- */
 
-  /**
-   * Get the valid item consumption targets which exist on the actor
-   * @returns {Object<string>}   An object of potential consumption targets
-   * @private
-   */
-  _getItemConsumptionTargets() {
```
Fyorl, line 277: How do we populate consumption targets for uses that are on the Item itself ('base uses') and not on an Activity? Do we call a method on the uses field?

Fyorl (changes requested):

module/applications/item/item-sheet.mjs
arbron: Items don't have consumption targets any more. Only activities can consume things now and items just hold uses that can be consumed by activities.

module/applications/item/item-sheet.mjs
arbron: I believe ammunition should be handled specially by the `AttackActivity` on weapons, because it is consumed when attacks are made and now when the activation is performed.

Fyorl: That makes a lot of sense, yeah

module/applications/item/item-sheet.mjs
Fyorl: Got it, I think that answers some other questions I had too, thanks

Fyorl (approved):

Issue #3910, opened 2024-07-31
Activity Type: Attack


arbron:
The attack activity will be able to roll attacks and damage. It will contain information on whether it is a melee or ranged attack, weapon or spell attack, and any bonuses to hit. For damage it will be able to take multiple formulas of different damage types and be able to fetch the intrinsic damage from a weapon if it exists on one.

krbz999:
What about items that can attack but do not deal damage?

arbron:
Then you can just not fill in any of the damage parts and no damage button will appear, similar to the current behavior on items.

Issue #313, opened 2020-01-23
Some Spells needs more complicated Scaling formula


aaclayton:
Originally in GitLab by @ariakas81

Spells like Spiritual Weapon, Flame Blade scale by 1dX for every two slot levels above their basic level.
Right now they can't scale properly with the available choices.![image](https://gitlab.com/foundrynet/dnd5e/uploads/d5e956881a38b6b1558a4ebaf11651a1/Cattura.JPG)

Fyorl:
A lot of these use cases are handled by per-damage-part scaling, as well as the improved scaling functionality available in activities, so I am marking this as done for now to allow for more targeted issues to be raised for any remaining use cases.

PR #3921, opened 2024-08-01
[#3910] Add `AttackActivity` data and sheet

arbron:
Introduces the attack activity which contains data on attack types, to hit modifiers, and damage parts. This adds the sheet for the attack activity which includes some extra details on the Identity tab for the attack type & classification and the to hit and damage parts on the Effect tab.

Screenshot 2024-08-01 at 13 12 37
Screenshot 2024-08-01 at 13 13 15

Note: Does not contain any functionality for rolling attacks or damage.

### Todo
- [ ] Display proper default labels for type, classification, and ability from the item if possible
- [ ] Display damage part from the weapon if "Include Base Damage" is checked
- [ ] Move `rollAttack` and `rollDamage` over
- [ ] Refactor damage parts of sheet and `rollDamage` into separate classes to share with `SavingThrowActivity`


lang/en.json
```diff
@@ -850,6 +903,63 @@
...
+    "Action": {
+      "Create": "Create Damage Part",
+      "Delete": "Delete Damage Part"
+    }
+  },
+  "Scaling": {
+    "Half": "Every Other Level",
+    "None": "No Scaling",
+    "Whole": "Every Level"
+  }
```
krbz999, line 961: Wondering if we shouldn't simply bite the bullet and allow for manual input of damage at every level. [This issue](<https://github.com/foundryvtt/dnd5e/issues/313#issuecomment-1154098007>) had some good suggestions, as well as edge cases that half/none/every level doesn't cover.

module/data/shared/damage-field.mjs
```diff
@@ -0,0 +1,126 @@
...
+   * Produce the auto-generated formula from the `number`, `denomination`, and `bonus`.
+   * @param {number} [increase=0]  Amount to increase the die count.
+   * @returns {string}
+   * @protected
+   */
+  _automaticFormula(increase=0) {
+    let formula;
+    const number = (this.number ?? 0) + increase;
+    if ( number && this.denomination ) formula = `${number}d${this.denomination}`;
+    if ( this.bonus ) formula = formula ? `${formula} + ${bonus}` : bonus;
```
Fyorl, line 89:
```suggestion
    if ( this.bonus ) formula = formula ? `${formula} + ${this.bonus}` : this.bonus;
```

module/data/shared/damage-field.mjs
```diff
@@ -0,0 +1,126 @@
...
+      formula = this.custom.formula;
+      formula = formula.replace(/^(\d)+d/, (match, number) => `${Number(number) + dieIncrease}d`);
+    } else {
+      formula = this._automaticFormula(dieIncrease);
+    }
+
+    // If custom scaling included, modify to match increase and append for formula
+    if ( this.scaling.formula ) {
+      let roll = new Roll(this.scaling.formula);
+      roll = roll.alter(increase, 0, { multiplyNumeric: true });
```
Fyorl, line 120: Should we be double-dipping here? Normally when you scale a formula, you either scale the base formula, *or* you apply the scaling formula to it. Here we are doing both.

module/data/shared/damage-field.mjs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/* -------------------------------------------- */
+
+/**
+ * Data model that stores information on a single damage part.
+ *
+ * @property {number} number           Number of dice to roll.
+ * @property {number} denomination     Die denomination to roll.
+ * @property {string} bonus            Bonus added to the damage.
```
Fyorl, line 21: This is fine so long as we're absolutely clear it's just for ease-of-input purposes, and that we are only ever using `DamageData#formula` everywhere else in the system.

module/data/shared/damage-field.mjs
```diff
@@ -0,0 +1,126 @@
...
+ * @property {Set<string>} types       One or more damage types. If multiple are selected, then the user will be able to
```
Fyorl, line 22: I am ok with supporting multiple damage types here, but I do think we should really only show the single drop-down on the sheet. Multiple damage types will never be used unless the table is using some homebrew or 3rd-party content. You can leave that for me though in the UI pass if you want, unless you feel strongly otherwise.

lang/en.json
Fyorl: Some of the more awkward spells where individual damage parts scale, like *flame strike* or *ice storm*, should be handled now that scaling is attached to the damage part in the model. I don't much relish the idea of building a UI for per-level scaling. Isn't it already achievable with only the scaling formula and whatever the equivalent of `@item.level` will be?

Fyorl (changes requested):
Good stuff, thank you Jeff

module/data/shared/damage-field.mjs
arbron: This provides a selection of possible damage types, for a spell like Chromatic Orb or the upcast damage on Flame Strike.

lang/en.json
arbron: Yeah, per-type scaling like this handle basically 99% of official spells, and a could of edge cases can be handled by setting the formula explicitly and using the scaling value in roll data. There are a few edge cases outside of damage (like durations that change from `1 minute` => `10 minutes` => `1 hour` as you upcast), but none of those are super urgent.

module/data/shared/damage-field.mjs
arbron: I wanted to include the dice scaling because it is what most damage uses so it easier, but still offer the flexibility of custom scaling formulas. I generally you'd use one or the other, but it didn't seem like there was any reason to prevent both if someone had a damage that added dice and numbers (like if someone wanted to scale Magic Missile, they could put `1` in the dice box and `1` into the custom formula box so it increases by `1d4 + 1` for each scaling step).

lang/en.json
krbz999: Yeah there aren't many left. I am mainly thinking of spells like Shadow Blade that don't scale in any easy pattern. For example that spell stops scaling at 7th level.

module/data/shared/damage-field.mjs
Fyorl: Fair point about *chromatic orb*, but how would this work for *flame strike*? You add the fire and radiant damage parts with no scaling, then you add a 3rd fire/radiant damage part of 0d6 with scaling? I guess maybe I answered my own question.

module/data/shared/damage-field.mjs
Fyorl: I think I actually misunderstood what this code was doing, but your answer clarifies it for me, thank you. Looks like we are missing a `let formula;` somewhere though.

lang/en.json
Fyorl: `BaseActivityData` has a `consumption.scaling.max` which should handle those maximums now. If the per-level scaling is just for UX purposes, and not because we can't accommodate that scaling with a formula then yeah, we can consider it. It's a non-trivial amount of UI work though, so unlikely to make it into 4.0.

Fyorl (approved):
See that one comment I posted about a missing `let formula;`, but otherwise this looks good to go, thank you.

PR #3924, opened 2024-08-02
Initial refactor for V2 sheet code and CSS sharing.

Fyorl:
As part of splitting up the Item sheets review I have separated out all the code and CSS reorganisation and refactoring. The vast majority of this PR is not new code, it is mainly just existing code that has been shuffled around.

- I extended the V2 mixin hierarchy. There is now a `DocumentSheetV2Mixin` that shares code common to Actor and Item sheets like the header buttons and mode toggle, as well as some code for placing the source field in the sheet header. `ActorSheetV2Mixin` now extends `DocumentSheetV2Mixin` and contains all the code shared across Actor-sheet types, and there will be a corresponding `ItemSheetV2Mixin`. When we eventually retire the legacy sheets we will be able to organise this into a more coherent hierarchy, or perhaps just continue using mixins.
- Moved shared V2 Document sheet CSS to `sheets.less`.
- Moved V2 form CSS to `forms.less` (there's some new code here).
- Made the journal content styles more general so they can now apply to Actor biographies and Item descriptions.
- Some assorted miscellaneous CSS tweaks/polishing where I noticed it (including some tweaks for v13 styles).


module/data/item/templates/activated-effect.mjs
```diff
@@ -152,9 +156,9 @@ export default class ActivatedEffectTemplate extends SystemDataModel {
       this.parent.labels.range = range.filterJoin(" ");
     } else this.parent.labels.range = game.i18n.localize("DND5E.None");
 
-    if ( this.recharge ) this.parent.labels.recharge = `${game.i18n.localize("DND5E.Recharge")} [${
-      `${this.recharge.value}${parseInt(this.recharge.value) < 6 ? "+" : ""}`
-    }]`;
+    if ( this.recharge ) this.parent.labels.recharge = `${game.i18n.localize("DND5E.Recharge")} [
+      ${this.recharge.value}${parseInt(this.recharge.value) < 6 ? "+" : ""}
+    ]`;
```
arbron, line 161: Can't test it at the moment because activities broke recharge, but wouldn't this lead to extra space between the brackets (`Recharge [ 5+ ]`)?

less/v2/apps.less
```diff
@@ -442,6 +442,38 @@
...
+
+    &::before { background: linear-gradient(to left, var(--dnd5e-color-gold), transparent); }
+    &::after { background: linear-gradient(to right, var(--dnd5e-color-gold), transparent); }
+
+    > div {
+      width: 20px;
+      height: 20px;
+      background: url("ui/official/ampersand-gold.svg") no-repeat center / contain;
+    }
+  }
```
arbron, line 465: It might make sense to use an `<hr>` element so we can get the semantic separator in addition to the visual styling. We could even get rid of the second element by massaging the CSS a bit:

```html
<hr class="ampersand">
```

```suggestion
  hr.ampersand {
    position: relative;
    height: 20px;
    border: none;
    background: url("ui/official/ampersand-gold.svg") no-repeat center / contain;

    &::before, &::after {
      content: "";
      position: absolute;
      inset-block: 9.5px;
      inset-inline: 0;
    }
    &::before {
      background: linear-gradient(to left, var(--dnd5e-color-gold), transparent);
      inset-inline-end: calc(50% + 14px);
    }
    &::after {
      background: linear-gradient(to right, var(--dnd5e-color-gold), transparent);
      inset-inline-start: calc(50% + 14px);
    }
  }
```

arbron (approved):
Seems good

module/data/item/templates/activated-effect.mjs
Fyorl: My IDE was complaining but I think it is wrong and you are right.

less/v2/apps.less
Fyorl: Good idea

PR #3925, opened 2024-08-02
Item & Container sheet V2 skeleton & header matter

Fyorl:
This PR contains the Item sheet V2 skeleton structure, including the header matter and associated styling.


module/applications/fields.mjs
```diff
@@ -0,0 +1,12 @@
+/**
+ * Create a text input for a StringField.
+ * @param {StringField} field               The field.
+ * @param {FormInputConfig<string>} config  The input configuration.
+ * @returns {HTMLElement|HTMLCollection}
+ */
+export function createTextInput(field, config) {
```
Fyorl, line 7: Basically only using this to polyfill the core `createTextInput` and allow attaching a class to the input.

templates/items/item-sheet-2.hbs
```diff
@@ -0,0 +1,128 @@
...
+
+        </div>
+
+        {{!-- Weight, Price, Quantity, Level, & School --}}
+        <div class="right common-fields {{#if isPhysical}}physical{{/if}}">
+
+            {{#if isPhysical}}
+
+            {{!-- Quantity --}}
+            <div class="quantity">
```
arbron, line 51: Having the numbers on the left for weight & cost but on the right for quantity feels a bit off. I wonder if we can find an icon for quantity and use that instead:

Screenshot 2024-08-02 at 13 53 39

```html
<i class="fa-solid fa-cubes-stacked"></i> // for example
```

arbron (approved):
Noticed that minimized sheets are broken, but that might be in a future PR:

Screenshot 2024-08-02 at 13 54 15

templates/items/item-sheet-2.hbs
Fyorl: I see what you mean, but the quantity icon isn't quite cutting it for me. I will try some things out on the forward branch.

Fyorl:
> Noticed that minimized sheets are broken, but that might be in a future PR:
> 
> Screenshot 2024-08-02 at 13 54 15

Not in a future PR but I will fix it on the forward branch.

PR #3927, opened 2024-08-02
[#1964] Add method to localize schemas in `ArrayField`s

arbron:
Adds a static `localize` method to activities which performs pre-localization tasks. During the `i18nInit` hook the system iterates over all activities and calls this method.

This method calls the normal `localizeDataModel` method on the activity itself and then calls `Activity#_localizeSchema` to localize any `SchemaField`s in arrays (this method is necessary because core's `localizeSchema` method is private).

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -32,6 +32,32 @@ export default Base => class extends PseudoDocumentMixin(Base) {
   });
 
   /* -------------------------------------------- */
+
+  /**
+   * Perform the pre-localization of this data model.
+   */
+  static localize() {
+    Localization.localizeDataModel(this);
+    const fields = this.schema.fields;
+    if ( fields.damage?.fields.parts ) this._localizeSchema(fields.damage.fields.parts.element, ["DND5E.DAMAGE"]);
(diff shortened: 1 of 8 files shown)


module/documents/activity/mixin.mjs
```diff
@@ -31,6 +31,32 @@ export default Base => class extends PseudoDocumentMixin(Base) {
...
 
+  /* -------------------------------------------- */
+
+  /**
+   * Perform the pre-localization of this data model.
+   */
+  static localize() {
+    Localization.localizeDataModel(this);
+    const fields = this.schema.fields;
+    if ( fields.damage?.fields.parts ) this._localizeSchema(fields.damage.fields.parts.element, ["DND5E.DAMAGE"]);
```
Fyorl, line 42: Should this be a part of an `AttackActivity.localize` override?

Fyorl (approved):

module/documents/activity/mixin.mjs
arbron: I have it here because other activities like `SaveActivity` will also have damage in the same place, so it seemed simpler to have it here rather than have each activity with damage parts localize it on their own.

PR #3928, opened 2024-08-02
[#1964] Add arbitrary roll to `UtilityActivity`

### LINKED ISSUE #941: Name-able "Other Formula"
"Other formula" is often less than clear about what it means for each item. I think it would be good to provide a name field for the other formula which would show on the chat card for the item to make it more clear what the button does.

I'm willing to make a MR if this is a reasonable addition.

arbron:
Adds a rolling formula and name that can be rolled when a `UtilityActivity` is used, similar to the "Other Formula" field on items currently.

Closes #941

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2323,7 +2323,20 @@
 },
 
 "DND5E.UTILITY": {
-  "Title": "Utility"
+  "Title": "Utility",
+  "FIELDS": {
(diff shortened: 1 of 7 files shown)


krbz999:
Related: #941

Fyorl (approved):

PR #3929, opened 2024-08-02
[#1964, #2349] Allow placing templates from activities

### LINKED ISSUE #2349: Improve tile snapping of measured template previews on hex grids
Placing them, e.g., in the center of a hex grid is near-impossible when they are generated from an item, while the Core template tools have no issue at all.

arbron:
Adds a new `AbilityTemplate#fromActivity` method and deprecates the old `fromItem` method. This new method supports placing multiple templates at a time.

Also switches to using `getSnappedPoint` rather than the deprecated `getSnappedPosition` which has the added benefit of working properly on hex grids.

Closes #2349

### DIFF
--- a/module/canvas/ability-template.mjs
+++ b/module/canvas/ability-template.mjs
@@ -28,13 +28,13 @@ export default class AbilityTemplate extends MeasuredTemplate {
   /* -------------------------------------------- */
 
   /**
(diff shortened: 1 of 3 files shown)


Fyorl (approved):
Good improvements

Issue #3914, opened 2024-07-31
Activity Type: Enchant


arbron:
The enchantment activity will behave the same as the existing enchantment behavior on items. It will require a custom activity activation dialog to handle selecting the enchantment if there is more than one.

PR #3935, opened 2024-08-05
[#3914] Add `EnchantActivity` data and sheet

arbron:
Copies over the data structure from `EnchantmentField` and the UI from `EnchantmentConfig`.

Enchant Sheet V1 - Enchantments
Enchant Sheet V1 - Restrictions

Introduces a new `enchantment` Active Effect type and added a migration to automatically convert effects with the old flag to the new type.

The data that was once stored in the flags on the old enchantment effects has been moved to the `effects` object on the enchant activity, which seems like a more logical place to store this information than on each enchantment effect because it has to do with the application process, not the final enchantment. When activity migrations are performed the old data in flags will be moved to the newly created `EnchantActivity`.

**Note:** Does not contain the functionality for enchanting items.

### Todo
- [ ] Move enchantment activation into `EnchantActivity`
- [ ] Migrate enchantment data from item into activity
- [ ] Determine where "Enchanted Items" list will be displayed on new item sheets


Fyorl (approved):

module/applications/item/enchantment-config.mjs
```diff
@@ -1,151 +1,10 @@
...
-      case "delete-enchantment":
-        enchantment?.deleteDialog();
-        break;
-      case "edit-enchantment":
-        enchantment?.sheet.render(true);
-        break;
-    }
+  constructor() {
+    throw new Error(
+      "EnchantmentConfig has been deprecated. Configuring enchating should now be performed through the Enchant activity."
```
krbz999, line 7:
```suggestion
      "EnchantmentConfig has been deprecated. Configuring enchanting should now be performed through the Enchant activity."
```

PR #3936, opened 2024-08-05
Item sheets V2 Description tab

Fyorl:
image

image

image

### DIFF
--- a/less/v2/items.less
+++ b/less/v2/items.less
@@ -199,6 +199,166 @@
   form > hr.ampersand { margin-bottom: -4px; }
 
   /* ---------------------------------- */
+  /*  Sheet Body                        */
+  /* ---------------------------------- */
+
+  .sheet-body {
+    .tab {
+      max-height: calc(70vh - var(--dnd5e-sheet-header-height));
+      overflow: auto;
+      display: flex;
+      flex-direction: column;
+      gap: 8px;
+      margin: -6px -11px -6px -6px;
+      padding: 6px 2px 6px 6px;
+      scrollbar-width: thin;
+      scrollbar-color: var(--dnd5e-color-scrollbar) transparent;
+      scrollbar-gutter: stable;
+
+      &:not(.active) { display: none; }
+      &.effects { padding: 12px 2px 58px 6px; }
+
+      > .pills .pill {
+        font-size: var(--font-size-10);
+        padding: 3px 6px;
+        line-height: 1;
+      }
+
+      &.single-description {
+        padding-right: 5px;
+        scrollbar-gutter: auto;
+
+        > .editor {
+          padding-right: 6px;
+          scrollbar-width: thin;
+          scrollbar-color: var(--dnd5e-color-scrollbar) transparent;
+        }
+      }
+    }
+
+    .spell-block {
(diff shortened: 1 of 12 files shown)


templates/items/parts/item-description2.hbs
```diff
@@ -0,0 +1,95 @@
...
+            {{#if @root.owner}}
+            <button type="button" class="description-edit unbutton" data-target="{{ target }}"
+                    aria-label="{{ localize "DND5E.DescriptionEdit" description=(localize label) }}">
+                <i class="fas fa-feather"></i>
+            </button>
+            {{/if}}
+        </div>
+        <div class="details collapsible-content">
+            <div class="wrapper">
+                {{ editor enriched target=target button=false editable=false engine="prosemirror" }}
```
arbron, line 15: Right now nothing appears to happen when collapsing/expanding a section with no description. Perhaps some placeholder text that indicates no description is present would help.

templates/items/parts/item-description2.hbs
```diff
@@ -0,0 +1,95 @@
...
+            </div>
+        </div>
+    </div>
+{{/inline}}
+
+<div class="tab {{#if singleDescription}}single-description{{/if}}" data-group="primary" data-tab="description">
+
+    {{#if editingDescriptionTarget}}
+        {{ editor enriched.editing target=editingDescriptionTarget button=false editable=true engine="prosemirror"
```
arbron, line 24: Any reason for favoring this over the new `<prose-mirror>` element?

arbron (approved):

templates/items/parts/item-description2.hbs
Fyorl: We have an `activateEditor` override in `ItemSheet` that configures the menu plugin to set `editingDescriptionTarget = null` when the editor is saved. I couldn't figure out how to replicate that behaviour with the `<prose-mirror>` element (admittedly I didn't spend too much time on it). Listening to the `change` event almost works, but it only fires if the editor content actually changes. We could subclass the element or add mutation observers maybe, but I would rather just improve the element in core to fire a `close` event or something.

less/v2/items.less
```diff
@@ -198,6 +198,166 @@
...
+            &::before {
+              content: "\f078";
+              font-family: var(--font-awesome);
+              font-weight: bold;
+              color: var(--color-text-dark-5);
+              transition: all 250ms ease;
+            }
+          }
+
+          &.collapsed > .header::before { transform: rotate(-90deg); }
```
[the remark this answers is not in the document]
Fyorl: The whole tab is the scroll container for the descriptions, they don't need to individually scroll. Setting `overflow: hidden` seems to work nicely, and also fixes a weird issue that I assumed might have been a browser bug where the internal scrollbar would sometimes persist even when fully expanded, but would go away if you forced a re-paint (e.g. by hovering a content link).

templates/items/parts/item-description2.hbs
Fyorl: I've removed the suggestion of interactivity on these empty sections for now, but we can change it if ends up being a bit confusing (it might).

Issue #3912, opened 2024-07-31
Activity Type: Damage/Healing


arbron:
One or two activity types for applying damage/healing directly to a target without a save or attack required (e.g. Healing Word or Magic Missile). Should contain a list of roll parts and details on whether the critical damage is allowed.

PR #3945, opened 2024-08-06
[#1964, #3912] Add migrations from item data to activities, rename `HealActivity`

arbron:
Adds migrations to handle creating the default activity on items based on the old `ActionTemplate` and `ActivatedEffectTemplate` data. These migrations are performed during the item migration process so it has access to flags and effects data.

The item calls into `ActivitiesTemplate` which determines whether the migration should be called at all. It will only perform the migration if it is a full-data migration, if the item has either `actionType` or `activation.type` set, and `system.activities` is empty. This then determines the proper activity type based on the `actionType` and calls the `migrateInitialActivity` method on its data model.

`BaseActivityData` handles the base level migration moving all of the shared activity data over and then passes that object to `transformTypeData` so specific activity types can handle any data they need.

### Todo
- [ ] Remove `ActionTemplate` and `ActivatedEffectTemplate` data

----------

All the other activity types are named in imperative form (e.g. "to attack", "to damage", "to summon"), so this updates heal to match.


module/data/activity/base-activity.mjs
```diff
@@ -146,6 +146,244 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
...
+        formula: ""
+      },
+      scaling: {
+        mode: source?.system.scaling?.mode !== "none" ? "whole" : "",
+        number: null,
+        formula: source?.system.scaling?.formula ?? ""
+      }
+    };
+
+    const parsed = (formula ?? "").match(/^\s*(\d+)d(\d+)(?:\s*([+|-])\s*(@?[\w\d.]+))?\s*$/i);
```
Fyorl, line 267: Could probably simplify the regex by removing spaces from the formula first, e.g. `(formula ?? "").replaceAll(" ", "").match()`. Even so, how confident are we that this parsing is going to catch what we want and not accidentally drop any data? It seems like it would be safer to just migrate legacy damage parts straight to formula.

module/data/activity/base-activity.mjs
```diff
@@ -146,6 +146,244 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
...
+      data.number = Number(parsed[1]);
+      data.denomination = Number(parsed[2]);
+      if ( parsed[4] ) data.bonus = parsed[3] === "-" ? `-${parsed[4]}` : parsed[4];
+    } else {
+      data.custom.enabled = true;
+      data.custom.formula = formula;
+    }
+
+    // If scaling denomination matches the damage denomination, set scaling using number rather than formula
+    const scaling = data.scaling.formula.match(/^\s*(\d+)d(\d+)\s*$/i);
```
Fyorl, line 278: Same as above, it feels like unnecessary work/risk when we can just set a formula in both instances and be done. If you are confident that we won't incorrectly migrate something and lose data though, then that's fine.

module/data/item/templates/activities.mjs
```diff
@@ -32,7 +33,55 @@ export default class ActivitiesTemplate extends SystemDataModel {
...
+
+  /**
+   * Method to determine whether the activity creation migration should be performed. This migration should only be
+   * performed on whole item data rather than partial updates, so check to ensure all of the necessary data is present.
+   * @param {object} source  Candidate source data to migrate.
+   * @returns {boolean}
+   */
+  static #shouldMigrateActivity(source) {
+    // If source data doesn't have effects or system data, then is only a partial migration
+    if ( !source.effects || !source.system ) return false;
```
Fyorl, line 52: Rather than relying on this heuristic, couldn't we instead invoke the activity migration from the `migrations.mjs` `migrateItemData`? Then we can be sure that it's only being run on full source data.

Also, should we add a check against `_stats.systemVersion`? I'm thinking about certain 'base' data like damage or range on weapons that might be accidentally re-migrated in some cases. Maybe it's impossible, but I don't think it can hurt to add this guard.

module/documents/activity/heal.mjs
```diff
@@ -1,27 +1,27 @@
...
 
   /** @inheritDoc */
   static metadata = Object.freeze(
     foundry.utils.mergeObject(super.metadata, {
-      type: "healing",
-      img: "systems/dnd5e/icons/svg/activity/healing.svg",
-      title: "DND5E.HEALING.Title",
-      sheetClass: HealingSheet
+      type: "heal",
+      img: "systems/dnd5e/icons/svg/activity/heal.svg",
```
Fyorl, line 22: Did this SVG actually get renamed? I don't see it in the diff.

Fyorl (changes requested):

module/data/activity/base-activity.mjs
krbz999: Can't remove spaces since `@steve-1` is not the same as `@steve - 1`.

module/data/activity/attack-data.mjs
```diff
@@ -38,4 +38,27 @@ export default class AttackActivityData extends BaseActivityData {
...
+        bonus: source.system.attack?.bonus ?? "",
+        flat: source.system.attack?.flat ?? false,
+        type: {
+          value: source.system.actionType.startsWith("m") ? "melee" : "ranged",
+          classification: source.system.actionType.endsWith("wak") ? "weapon" : "spell"
+        }
+      },
+      damage: {
+        includeBase: true,
+        parts: source.system.damage?.parts?.map(part => this.transformDamagePartData(source, part)) ?? []
```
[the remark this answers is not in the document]
arbron: There will be some extra handling once the data structure for weapons is finalized in the next PR that should prevent that from happening, same with any data stored in spells.

module/data/activity/base-activity.mjs
arbron: The regex should be fairly narrow so it should fall back to custom formula for anything but the expected case:

```
1d4 // parsed
1d4 + @mod // parsed
1d4 - 5 // parsed
1d3 // custom
5 // custom
1d4 + @mod + 1 // custom
```

I don't think there is any chance of losing data because of the limited scope of the RegEx.

module/data/item/templates/activities.mjs
arbron: If we run during `migrateItemData`, then all modules have to be explicitly migrated by their creators before they can be used by players. This ensures that those modules are usable on day 1.

We can add a `systemVersion` check to be doubly sure, but since we only perform the migration if `system.activities` is absent or empty, it should ensure that the migration doesn't occur twice.

Fyorl: > If we run during `migrateItemData`, then all modules have to be explicitly migrated by their creators before they can be used by players. This ensures that those modules are usable on day 1.

That's always been the case though for any concrete migrations. I think there are a couple of things we'd need to watch out for with this in-memory migration:
1. It's not immune to false positives. It's possible (though unusual) to perform an update like `item.update({ "system.level": 9, effects: [{ /* ... */ }] })` which I think would result in spawning a new Activity because `source.activities` would be `undefined` for that update delta.
2. If you open an Item in a compendium that has not had `migrateCompendium` run on it, and then try to update it without importing (if it's in a personal shared compendium for example), then the delta sent to the server will not contain any of the migrated data that only exists in-memory at that time, it will contain only what changed. Upon refreshing, the previously-migrated activity data would be wiped, and you'd be left with only the fields that were updated.

We might be able to mitigate some of this by moving the migration to `_initializeSource`, which isn't called on update deltas, but I think (2) is always an issue.

> We can add a `systemVersion` check to be doubly sure, but since we only perform the migration if `system.activities` is absent or empty, it should ensure that the migration doesn't occur twice.

It's possible to create an Item with no activities on it though, no? Or delete any auto-generated ones?

module/data/item/templates/activities.mjs
```diff
@@ -32,7 +33,58 @@ export default class ActivitiesTemplate extends SystemDataModel {
...
+
+    // If item doesn't have an action type or activation, then it doesn't need an activity
+    if ( !source.system.actionType || !source.system.activation?.type ) return false;
+
+    // If item was updated after `4.0.0`, it shouldn't need the migration
+    if ( foundry.utils.isNewerVersion(source._stats.systemVersion ?? "0.0.0", "3.9.9") ) return false;
```
Fyorl, line 58: Prefer to use the negation here, rather than a string like `3.9.9`, since version numbers aren't decimals.

```suggestion
    if ( !foundry.utils.isNewerVersion("4.0.0", source._stats.systemVersion ?? "0.0.0") ) return false;
```

module/data/activity/base-activity.mjs
MaxPat931: There are also a ton of users that will have damage type annotations in the formulas, e.g. `1d6 + @mod[fire]`, looks like those will also be punted to `custom`, but figured it was worth a mention

module/data/item/templates/activities.mjs
arbron: > It's not immune to false positives. It's possible (though unusual) to perform an update like `item.update({ "system.level": 9, effects: [{ /* ... */ }] })` which I think would result in spawning a new Activity because `source.activities` would be `undefined` for that update delta.

True, we could resolve that by adding more edge data points that would be guaranteed to be in the full data but not in a partial update like `ownership`.

> If you open an Item in a compendium that has not had `migrateCompendium` run on it, and then try to update it without importing (if it's in a personal shared compendium for example), then the delta sent to the server will not contain any of the migrated data that only exists in-memory at that time, it will contain only what changed. Upon refreshing, the previously-migrated activity data would be wiped, and you'd be left with only the fields that were updated.

Hmm, that is part of the reason I didn't include a `systemVersion` check to start with.

> It's possible to create an Item with no activities on it though, no? Or delete any auto-generated ones?

Yes, but a newly created item would be missing `actionType` and `activation.type` and a world item migrated in that way would have them removed during `persistSourceMigration` (I believe). That would probably be an issue on a user's module compendium that never had `migrateCompendium` run on it if they try to delete the auto-generated activity.

Fyorl (approved):

PR #3946, opened 2024-08-06
Item Sheet V2: Container Contents tab

Fyorl:
image

### DIFF
--- a/less/v2/items.less
+++ b/less/v2/items.less
@@ -52,6 +52,20 @@
     }
   }
 
+  &.minimized {
+    box-shadow: 0 0 20px var(--color-shadow-dark);
+    background: url("../../ui/denim075.png");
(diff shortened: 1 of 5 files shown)


arbron (approved):

arbron:
One though, though not specific to containers, is that the weight column is displaying the cumulative weight for the whole stack of items but the price column is only displaying the cost for a single item.

Fyorl:
> One thought, though not specific to containers, is that the weight column is displaying the cumulative weight for the whole stack of items but the price column is only displaying the cost for a single item.

Ah you are right. I guess it's more useful to see the total since you can always hover an item to see its individual price and weight. I will have to add a little bit of code to handle currency conversion when multiplying the cost by the quantity so it will have to wait for a separate PR.

PR #3947, opened 2024-08-07
Item Sheet V2: Effects tab

Fyorl:
image


arbron (changes requested):
Whenever I click on the three dots to open the context menu I get an error (though the menu still opens):

Screenshot 2024-08-07 at 08 14 14

Also tried for a bit to use the search control on an item in a compendium only to finally notice it was disabled. That probably shouldn't be disabled even if the item isn't editable (same problem on any inventories on locked sheets), or if it needs to be for some reason the styling will need to be adjusted to make it obvious.

arbron (approved):

PR #3949, opened 2024-08-07
force all token displayBars and displayName to make sense

Drental:
Just force displayBars to OWNER_HOVER and displayName to OWNER_HOVER for monsters and HOVER for heroes.

I just did a simple issue reproduction and stumbled across this not being the literal same for all tokens in the compendium, I am shook and dismayed and this PR is fueled by outrage.

I can change the values to any that should be the default in dnd5e, I just want them to always be the same.


Fyorl:
> Just force displayBars to OWNER_HOVER and displayName to OWNER_HOVER for monsters and HOVER for heroes.

I can see `displayBars` set to `20` for some (all?) of the Starter Heroes, which is OWNER_HOVER, not HOVER.

What is the issue that this PR is trying to address? That some of the SRD monsters have different bars/display name settings in their prototype token configurations to others?

Drental:
Yes, I did not like that the settings were randomly different for different actors.

Sharing information about the HP of characters is considered metagaming in some groups, so I set the bars to HOVER_OWNER, the GM or Players can then reveal the bar if they decide that this is a level of metagame that they are fine with. I'd rather err on the side of hiding information than showing information, since sharing information when the need arises is not too much work, but hiding information that was already seen means that the damage is already done for those that want to hide this information.

But as I said, I don't care what the defaults are, I care that there are defaults^^

PR #3950, opened 2024-08-07
Item Sheet V2: Advancement tab

Fyorl:
image

The `legacyDisplay` stuff is an attempt to maintain the original legacy sheet display. I'm not 100% sold on the new icons, but maybe we can try them out for a while. The SVGs still look pretty decent.


templates/items/parts/item-advancement2.hbs
```diff
@@ -0,0 +1,112 @@
...
+                            {{!-- Deleting --}}
+                            <a class="item-control item-action" data-action="delete"
+                               data-tooltip="DND5E.AdvancementControlDelete"
+                               aria-label="{{ localize "DND5E.AdvancementControlDelete" }}">
+                                <i class="fas fa-trash"></i>
+                            </a>
+                            {{/if}}
+
+                            {{!-- Context Menu --}}
+                            <a class="item-control interface-only" data-context-menu
```
arbron, line 95: Nothing happening when clicking on this.

templates/items/parts/item-advancement2.hbs
```diff
@@ -0,0 +1,112 @@
+<div class="tab advancement" data-group="primary" data-tab="advancement">
+    <section class="items-list">
+        {{#each advancement}}
+
+        <div class="items-section card" data-level="{{ @key }}">
```
arbron, line 5: Modify choices isn't working at the moment because it is looking for `<li>` to find the level. You'll have to modify line `814` of `item-sheet.mjs` to look for the data instead:

```javascript
const level = target.closest("[data-level]").dataset.level;
```

arbron (changes requested):

arbron (approved):

PR #3951, opened 2024-08-08
[#1964] Remove Action & Activated Effect data & add shims

arbron:
All of the item types in the system that mix in `ActivityTemplate` no longer use `ActionTemplate` or `ActivatedEffectTemplate`. To avoid breakages they now call `_applyActivityShims` which adds shims to all of the old properties from those templates where possible.

This PR also includes changes to `WeaponData` that includes base damage & versatile damage as well as changes to `SpellData` that includes activation, duration, range, and target objects.

A few additional data changes had to be made to due to gaps:
- `cover` and `crewed` have moved from `ActivatedEffectTemplate` to `MountableTemplate`
- `chatFlavor` has been added to all activities and is now used in the created chat messages
- Critical bonus formula has been added to attack & damage activities, and critical threshold has been added to attack activity

Currently the preparation that was performed in
`prepareFinalActivatedEffectData` hasn't been re-implemented, but most of that will be handled by `BaseActivityData`. This means that a few labels won't display on legacy item sheets.

### Todo
- [ ] Implement data preparation from `prepareFinalActivatedEffectData` in `BaseActivityData`
- [ ] Migrate weapon damage & range
- [ ] Migrate spell activation, duration, range, and target

### DIFF
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -11,15 +11,15 @@ const { ArrayField, BooleanField, NumberField, SchemaField, StringField } = foun
  * @property {object} attack
  * @property {string} attack.bonus                Arbitrary bonus added to the attack.
  * @property {object} attack.critical
- * @property {number} attack.critical.threshold   Minimum value on the D20 needed to rolling a critical hit.
+ * @property {number} attack.critical.threshold   Minimum value on the D20 needed to roll a critical hit.
  * @property {boolean} attack.flat                Should the bonus be used in place of proficiency & ability modifier?
  * @property {object} attack.type
  * @property {string} attack.type.value           Is this a melee or ranged attack?
  * @property {string} attack.type.classification  Is this a unarmed, weapon, or spell attack?
  * @property {object} damage
  * @property {object} damage.critical
- * @property {string} damage.critical.bonus       Extra damage applied when a critical is rolled. Will be added to the
- *                                                base damage or first damage part.
+ * @property {string} damage.critical.bonus       Extra damage applied when a critical is rolled. Added to the base
+ *                                                damage or first damage part.
  * @property {boolean} damage.includeBase         Should damage defined by the item be included with other damage parts?
  * @property {DamageData[]} damage.parts          Parts of damage to inflict.
  */
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -2,7 +2,7 @@ import FormulaField from "../fields/formula-field.mjs";
 import DamageField from "../shared/damage-field.mjs";
 import BaseActivityData from "./base-activity.mjs";
 
-const { ArrayField, BooleanField, SchemaField, StringField } = foundry.data.fields;
+const { ArrayField, BooleanField, NumberField, SchemaField, StringField } = foundry.data.fields;
 
 /**
  * Data model for an attack activity.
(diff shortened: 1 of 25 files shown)


module/data/activity/attack-data.mjs
```diff
@@ -2,19 +2,24 @@ import FormulaField from "../fields/formula-field.mjs";
...
+const { ArrayField, BooleanField, NumberField, SchemaField, StringField } = foundry.data.fields;
 
 /**
  * Data model for an attack activity.
  *
  * @property {string} ability                     Ability used to make the attack and determine damage.
  * @property {object} attack
  * @property {string} attack.bonus                Arbitrary bonus added to the attack.
+ * @property {object} attack.critical
+ * @property {number} attack.critical.threshold   Minimum value on the D20 needed to rolling a critical hit.
```
Fyorl, line 14:
```suggestion
 * @property {number} attack.critical.threshold   Minimum value on the D20 needed to roll a critical hit.
```

module/data/item/templates/activities.mjs
```diff
@@ -134,4 +204,133 @@ export default class ActivitiesTemplate extends SystemDataModel {
...
+      },
+      chatFlavor: () => firstActivity.description?.chatFlavor ?? "",
+      critical: () => {
+        const activity = this.activities.getByType("attack")[0] || this.activities.getByType("damage")[0];
+        return {
+          threshold: activity?.attack?.critical?.threshold ?? null,
+          damage: activity?.damage?.critical?.bonus ?? ""
+        };
+      },
+      dammage: () => ({ parts: [], versatile: "" }),
```
Fyorl, line 264: There's a type, but also, should this not be pulling from the attack, damage, or save activities?
```suggestion
      damage: () => ({ parts: [], versatile: "" }),
```

module/data/item/templates/activities.mjs
```diff
@@ -134,4 +204,133 @@ export default class ActivitiesTemplate extends SystemDataModel {
...
+      enchantment: () => this.activities.getByType("enchant")[0],
+      formula: () => this.activities.getByType("utility")[0]?.roll?.formula ?? "",
+      hasAbilityCheck: () => false,
+      hasAttack: () => !!this.activities.getByType("attack").size,
+      hasDamage: () => !!this.activities.find(a => a.damage?.parts?.length),
+      hasSave: () => !!this.activities.getByType("save").size,
+      hasSummoning: () => !!this.activities.getByType("summon").size,
+      isEnchantment: () => !!this.activities.getByType("enchant").size,
+      isHealing: () => !!this.activities.getByType("heal").size,
+      isVersatile: () => this.actionType && this.properties?.has("ver"),
```
Fyorl, line 274: Some of these seem like they would still be useful. For example, checking if a weapon is versatile, or checking if an Item has an attack rather than calling `item.system.activities.getByType("attack").size` each time. Should they just be moved to `ActivitiesTemplate` rather than throwing a deprecation warning?

module/data/item/templates/activities.mjs
```diff
@@ -134,4 +204,133 @@ export default class ActivitiesTemplate extends SystemDataModel {
...
+        const consume = this.consume;
+        return this.isActive && !!consume.target && !!consume.type && this.hasAttack && (consume.type === "ammo");
+      },
+      hasAreaTarget: () => this.isActive && this.target.template?.type,
+      hasIndividualTarget: () => this.isActive && this.target.affects?.type,
+      hasResource: () => this.isActive && !!consume.target && !!consume.type && !this.hasAttack,
+      hasScalarDuration: () => this.duration.units in CONFIG.DND5E.scalarTimePeriods,
+      hasScalarRange: () => this.range.units in CONFIG.DND5E.movementUnits,
+      hasScalarTarget: () => this.target.template?.type || ![null, "", "self"].includes(this.target.affects?.type),
+      hasTarget: () => this.isActive && (this.target.template?.type || this.target.affects?.type),
```
Fyorl, line 322: I guess these might still be useful if the Item has one of a `DurationField`, `RangeField`, or `TargetField`.

Fyorl (changes requested):
This is really coming together, thank you, Jeff. Please see my comments for discussion.

module/data/item/templates/activities.mjs
arbron: We mainly have those for rendering the sheet changes which needs to be shared between item/activity sheets, so it would probably be best to re-implement these as necessary on the individual data fields instead.

Fyorl: Yeah makes sense

module/data/item/templates/activities.mjs
```diff
@@ -23,6 +23,127 @@ export default class ActivitiesTemplate extends SystemDataModel {
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Does this Item implement summoning as part of its usage?
+   * @type {boolean}
+   */
+  get hasSummoning() {
+    const act = this.activities.getByType("summon")[0];
+    return act && act?.profiles.length > 0;
```
Fyorl, line 114:
```suggestion
    return act && act.profiles.length > 0;
```

PR #3959, opened 2024-08-08
[#1964] Prepare final formulas and labels

arbron:
Implements that final formula resolution and other preparation for duration, range, and target fields on spells and activities, and in the uses field everywhere. Some behavior previously implemented in `ActivatedEffectTemplate` to resolve formulas is now in the new `prepareFormulaValue` utility function.

This also re-implements the label generation previously handled in `ActivatedEffectTemplate` for spells and weapons.

### DIFF
--- a/module/data/item/weapon.mjs
+++ b/module/data/item/weapon.mjs
@@ -32,12 +32,23 @@ const { NumberField, SchemaField, SetField, StringField } = foundry.data.fields;
  * @property {object} range
  * @property {number} range.value           Short range of the weapon.
  * @property {number} range.long            Long range of the weapon.
- * @property {string} range.units           Units used to measure the weapon's range.
+ * @property {number} range.reach           Reach of the weapon.
+ * @property {string} range.units           Units used to measure the weapon's range and reach.
  */
 export default class WeaponData extends ItemDataModel.mixin(
   ActivitiesTemplate, ItemDescriptionTemplate, IdentifiableTemplate, ItemTypeTemplate,
   PhysicalItemTemplate, EquippableItemTemplate, MountableTemplate
 ) {
+
+  /* -------------------------------------------- */
+  /*  Model Configuration                         */
+  /* -------------------------------------------- */
+
+  /** @override */
+  static LOCALIZATION_PREFIXES = ["DND5E.RANGE"];
+
+  /* -------------------------------------------- */
+
   /** @inheritdoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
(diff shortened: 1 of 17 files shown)


module/data/item/weapon.mjs
```diff
@@ -157,6 +157,16 @@ export default class WeaponData extends ItemDataModel.mixin(
   prepareFinalData() {
     this.prepareFinalActivityData(this.parent.getRollData({ deterministic: true }));
     this.prepareFinalEquippableData();
+
+    const labels = this.parent.labels ??= {};
+    if ( this.hasRange ) {
```
Fyorl, line 162: After considering this some more, I think I'm on-board with this logic. We shouldn't conflate reach and range which are not quite the same thing. That said, we do not really have any support for the vacuum this leaves behind, since the system has no concept of reach currently. How do we represent something like the Kraken's tentacle attack? Do we give its Attack Activity a range of 30 ft.?

module/data/shared/activation-field.mjs
```diff
@@ -17,4 +17,29 @@ export default class ActivationField extends SchemaField {
...
+  /* -------------------------------------------- */
+
+  /**
+   * Prepare data for this field. Should be called during the `prepareFinalData` stage.
+   * @this {ItemDataModel|BaseActivityData}
+   * @param {object} rollData  Roll data used for formula replacements.
+   * @param {object} [labels]  Object in which to insert generated labels.
+   */
+  static prepareData(rollData, labels) {
+    Object.defineProperty(this.activation, "scalar", {
```
Fyorl, line 32: What's the benefit of a getter here, vs. derived data? It doesn't look like there's anything dynamic happening in the getter.

Fyorl (changes requested):
Looks good to me, just had a couple of questions first.

module/data/item/weapon.mjs
arbron: I think this is something I have on BF that I forgot to carry over, which is a `range.reach` field on a weapon. So for Kraken you would create a natural (or unarmed) weapon with the Reach property and enter `30` into the reach field. Adding the extra field separate from the range fields would also allow us to implement thrown weapons with reach.

module/data/shared/activation-field.mjs
arbron: Nothing then having it not being enumerable, but that probably isn't a huge benefit.

Fyorl: Let's switch to derived values for now then, as a more understandable concept, unless we find a good reason to use getters.

module/data/item/weapon.mjs
Fyorl: Reach is a little more complicated as it's kind of a property of the creature itself, and weapons inherit the creature's reach. This is pretty much only an issue for bugbears though, which we don't have a good way of representing anyway since they only gain the extra reach on their turn. I think your proposal should work for all reach weapons and natural weapons, and bugbears are no worse off than they were before.

Fyorl (approved):

PR #3960, opened 2024-08-08
[#1964] Copy inferred data from items to activity, add overrides

arbron:
During activity preparation this copies data from the item over to the activity unless the `override` option is set. If the original data couldn't be found on the item, then the override checkbox won't appear and the activity values will be used.

After the copying occurs a copy of the activity data is saved in `_inferredSource` so that the activity sheet has a way to show inferred data before any final formula resolution occurs.

The sheet then selects the item data or activity data based on the override checkbox, and if override isn't set it will disable the whole `<fieldset>`.

### Todo
- [ ] Handle inferred data from weapon to `AttackActivity`

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -159,7 +159,7 @@
       },
       "override": {
         "label": "Override Activation",
-        "hint": "Override the activation information from the item with values for this activity."
+        "hint": "Use these activation values instead of the item's when using this activity."
       },
       "type": {
         "label": "Activation Type",
@@ -1057,7 +1057,7 @@
       "label": "Duration",
       "override": {
         "label": "Override Duration",
-        "hint": "Override the duration information from the item with values for this activity."
+        "hint": "Use these duration values instead of the item's when using this activity."
       },
       "special": {
         "label": "Special Duration",
@@ -1790,7 +1790,7 @@
       },
       "override": {
         "label": "Override Range",
-        "hint": "Override the range information from the item with values for this activity."
+        "hint": "Use these range values instead of the item's when using this activity."
       },
       "reach": {
         "label": "Reach"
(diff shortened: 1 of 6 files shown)


lang/en.json
```diff
@@ -157,6 +157,10 @@
         "label": "Activation Condition",
         "hint": "Condition required to activate this activity."
       },
+      "override": {
+        "label": "Override Activation",
+        "hint": "Override the activation information from the item with values for this activity."
```
krbz999, line 162: This is confusingly worded, does it override the values _on the activity_ with values from the item, or the other way around

Fyorl: It's the other way around, but I agree the wording might be difficult to parse for the average user. I think we could probably do with simplifying the language across the board on a lot of our forms.

For this one, maybe something like: `"Use these values instead of the item's when using this activity."`

module/data/activity/base-activity.mjs
```diff
@@ -441,6 +458,26 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
   /*  Helpers                                     */
   /* -------------------------------------------- */
 
+  /**
+   * Add an `canOverride` property to the provided object and, if `override` is `true`, replace the data on the
```
Fyorl, line 462: I'm becoming a little confused now, but I think this should be `false`, based on my understanding and the code below. I think the term 'override' makes sense from the UX perspective, because we want to essentially 'activate' a given block of data for an Activity. But from an implementation perspective, what's actually happening is that every Activity is, by default, *inheriting* the Item's properties, and we use the `override` flag to turn off that inheritance.

```suggestion
   * Add an `canOverride` property to the provided object and, if `override` is `false`, replace the data on the
```

Fyorl (changes requested):

Fyorl (approved):

PR #3961, opened 2024-08-09
[#1964] Add inferred attack data, fix bug in save sheet

arbron:
Infers attack type & classification from the containing item and displays a label indicating the inferred value if no overriding value is specific on the `AttackActivity`. Also displays the potential ability options that might be used for the attack.

When base damage is included, this adds the weapon's base damage to the list of parts but disables its form elements, so users can view the weapon's damage on the activity sheet.

Also fixed a bug on the `SaveSheet` causing the incorrect DC calculation value to be displayed.

### DIFF
--- a/module/applications/activity/attack-sheet.mjs
+++ b/module/applications/activity/attack-sheet.mjs
@@ -42,12 +42,28 @@ export default class AttackSheet extends ActivitySheet {
   async _prepareEffectContext(context) {
     context = await super._prepareEffectContext(context);
 
-    // TODO: Add better default label to indicate what ability will be chosen
-    // TODO: Add spellcasting option to automatically select spellcasting ability
+    const availableAbilities = this.activity.availableAbilities;
     context.abilityOptions = [
-      { value: "", label: "" },
-      ...Object.entries(CONFIG.DND5E.abilities).map(([value, config]) => ({ value, label: config.label }))
+      {
+        value: "", label: game.i18n.format("DND5E.DefaultSpecific", {
+          default: this.activity.type.classification === "spell"
+            ? game.i18n.localize("DND5E.Spellcasting").toLowerCase()
+            : availableAbilities.size
+              ? game.i18n.getListFormatter({ style: "short", type: "disjunction" }).format(
+                Array.from(availableAbilities).map(a => CONFIG.DND5E.abilities[a].label.toLowerCase())
+              )
+              : game.i18n.localize("DND5E.None").toLowerCase()
+        })
+      },
+      // TODO: Stick an <hr> here when possible
+      { value: "none", label: game.i18n.localize("DND5E.None") },
+      { value: "spellcasting", label: game.i18n.localize("DND5E.Spellcasting") },
+      ...Object.entries(CONFIG.DND5E.abilities).map(([value, config]) => ({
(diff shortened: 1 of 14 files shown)


module/applications/activity/attack-sheet.mjs
```diff
@@ -42,13 +42,29 @@ export default class AttackSheet extends ActivitySheet {
...
+      },
+      // TODO: Stick an <hr> here when possible
+      { value: "none", label: game.i18n.localize("DND5E.None") },
+      { value: "spellcasting", label: game.i18n.localize("DND5E.Spellcasting") },
+      ...Object.entries(CONFIG.DND5E.abilities).map(([value, config]) => ({
+        value, label: config.label, group: game.i18n.localize("DND5E.Abilities")
+      }))
     ];
 
+    context.hasBaseDamage = this.item.system.damage?.base;
```
Fyorl, line 66:
```suggestion
    context.hasBaseDamage = this.item.system.damage?.base.formula;
```

module/config.mjs
```diff
@@ -332,6 +334,15 @@ DND5E.weaponProficienciesMap = {
   martialR: "mar"
 };
 
+/**
+ * A mapping between `DND5E.weaponTypes` and `DND5E.attackClassifications`. Unlisted types are assumed to be
+ * of the "weapon" classification.
+ * @enum {string}
+ */
+DND5E.weaponClassificationMap = {
+  natural: "unarmed"
```
Fyorl, line 343: Not sure about this. Natural weapons are considered attacks with weapons, while unarmed strikes are not.

module/data/activity/attack-data.mjs
```diff
@@ -60,6 +60,30 @@ export default class AttackActivityData extends BaseActivityData {
...
+   * @type {Set<string>}
+   */
+  get availableAbilities() {
+    // Defer to item if available
+    if ( this.item.system.availableAbilities ) return this.item.system.availableAbilities;
+
+    // Spell attack not associated with a single class, use highest spellcasting ability on actor
+    if ( this.attack.type.classification === "spell" ) return new Set([
+      this.parent?.actor?.system.attributes?.spellcasting,
+      ...Object.values(this.parent?.actor?.spellcastingClasses ?? {}).map(c => c.spellcasting.ability)
```
Fyorl, line 77: This implies that we would use the highest of the default spellcasting modifier or any spellcasting class' spellcasting modifier. There's no point in setting a default spellcasting modifier if it's just going to be ignored in that case.

Fyorl (changes requested):

module/applications/activity/attack-sheet.mjs
arbron: I was thinking we'd always want the "Include Base Damage" checkbox to be shown on weapons, even if the user hasn't set up any damage on the weapon yet (it won't show up in the weapon parts list unless an actual formula is configured).

module/config.mjs
arbron: Okay, I can leave it blank for now until we add an actual "Unarmed" weapon type.

module/applications/activity/attack-sheet.mjs
Fyorl: Ah ok, makes sense

Fyorl (approved):

PR #3962, opened 2024-08-09
Item Sheet V2: Activities tab

Fyorl:
image

There's room for some supplementary information on these activity rows, so the design may change a bit, but for now I haven't thought much about what would go there.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -180,8 +180,10 @@
 },
 
 "DND5E.ACTIVITY": {
-  "Create": "Create Activity",
-  "Delete": "Delete Activity",
+  "Action": {
+    "Create": "Create Activity",
+    "Delete": "Delete Activity"
+  },
   "Title": {
     "one": "Activity",
     "other": "Activities"
--- a/lang/en.json
+++ b/lang/en.json
@@ -180,6 +180,8 @@
 },
 
 "DND5E.ACTIVITY": {
+  "Create": "Create Activity",
+  "Delete": "Delete Activity",
   "Title": {
     "one": "Activity",
     "other": "Activities"
--- a/module/applications/item/item-sheet-2.mjs
+++ b/module/applications/item/item-sheet-2.mjs
@@ -33,6 +33,7 @@ export default class ItemSheet5e2 extends ItemSheetV2Mixin(ItemSheet5e) {
   /** @inheritDoc */
   async getData(options) {
     const context = await super.getData(options);
+    const { activities } = this.item.system;
 
     // Effects
     for ( const category of Object.values(context.effects) ) {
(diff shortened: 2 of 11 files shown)


module/documents/item.mjs
```diff
@@ -90,10 +90,17 @@ export default class Item5e extends SystemDocumentMixin(Item) {
      */
     if ( options.pack || options.parent?.pack ) Hooks.callAll("dnd5e.initializeItemSource", this, data, options);
 
-    ActivitiesTemplate.initializeActivities(data);
```
Fyorl, line 93: I noticed there are some times when `_initializeSource` might receive partial source data, such as when creating a temporary Document, i.e. `new Actor({ name: 'foo', type: 'npc' });`. I moved this to `fromSource` instead, which generally expects to instantiate a Document from full source data. It's not foolproof, but it's the best I can think of right now.

lang/en.json
```diff
@@ -180,6 +180,8 @@
 },
 
 "DND5E.ACTIVITY": {
+  "Create": "Create Activity",
+  "Delete": "Delete Activity",
```
arbron, line 184: For some other localizations recently I've been putting these in an an Action object:

```suggestion
  "Action": {
    "Create": "Create Activity",
    "Delete": "Delete Activity"
  },
```

templates/items/parts/item-activities.hbs
```diff
@@ -0,0 +1,29 @@
+<div class="tab activities" data-group="primary" data-tab="activities">
+    {{#each system.activities}}
+    <div class="activity card" data-id="{{ _id }}">
+
+        {{!-- TODO: Add drag handle for re-ordering --}}
+
+        {{!-- Icon --}}
+        <div class="icon">
+            <dnd5e-icon src="{{ img }}"></dnd5e-icon>
```
arbron, line 9: You can give activities custom icons, so we'll probably want to handle that similar to the advancement icons.

module/applications/item/item-sheet-2.mjs
```diff
@@ -92,6 +92,55 @@ export default class ItemSheet5e2 extends ItemSheetV2Mixin(ItemSheet5e) {
         }));
       });
     }
+
+    if ( this.isEditable ) {
+      html.find(".activities .activity .name").on("click", this._onEditActivity.bind(this));
```
arbron, line 97: It would probably still be good to be able to open the activity sheet even if the item is locked.

arbron (changes requested):

module/documents/item.mjs
arbron: Seems to be working, but I'll keep an eye out to see if anything funky happens.

arbron (approved):

PR #3963, opened 2024-08-09
Fix handling of certain rolls that contain division or multiplication

### LINKED ISSUE #3895: System invalidating roll formulas
These work as expected:
- `/r 1d10 + (3/2)`
- `/r 1d10 + (floor(3/2))`

These throw an error:
- `/r 1d10 + 3/2`
- `/r 1d10 + floor(3/2)`

image

Tested similar rolls in a different system with zero issues.

eelpik:
Closes #3895. This change gets certain rolls that threw errors before (like 1d10 + 3/2 or 1d10 + 3*2) to work properly.

### DIFF
--- a/module/dice/simplify-roll-formula.mjs
+++ b/module/dice/simplify-roll-formula.mjs
@@ -66,14 +66,6 @@ export default function simplifyRollFormula(formula, { preserveFlavor=false, det
   // Perform arithmetic simplification on the existing roll terms.
   roll.terms = _simplifyOperatorTerms(roll.terms);
 
-  // If the formula contains multiplication or division we cannot easily simplify
-  if ( /[*/]/.test(roll.formula) ) {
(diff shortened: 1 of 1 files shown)


module/dice/simplify-roll-formula.mjs
```diff
@@ -66,14 +66,6 @@ export default function simplifyRollFormula(formula, { preserveFlavor=false, det
...
   roll.terms = _simplifyOperatorTerms(roll.terms);
 
-  // If the formula contains multiplication or division we cannot easily simplify
-  if ( /[*/]/.test(roll.formula) ) {
-    if ( roll.isDeterministic && !/d\(/.test(roll.formula) && (!/\[/.test(roll.formula) || !preserveFlavor) ) {
-      return String(roll.evaluateSync().total);
-    }
-    else return roll.constructor.getFormula(roll.terms);
-  }
-
```
krbz999, line 76: The code removed here was added to fix a different issue, however. We should rather locate a way to fix the current problem without reintroducing the previous one.

PR #3964, opened 2024-08-09
[#1964] Remove activation-based preparation from items

arbron:
This should resolve any data preparation warnings caused by data that has been moved into activities. This means that most labels are now generated on an activity's `label` object rather than on the item, unless it is something like a spell or weapon that has certain intrinsic data.

The handfull of type-specific label generation that is left on items has been moved into data models to clean up `Item5e`.

The `ability` data for `AttackActivity` and `SaveActivity` has been moved into the `attack` and `save` objects respectively to make room at the top level for an `ability` getter which is used to set `@mod` in roll data.

A number of the getters on `Item5e` have been marked as deprecated if they call into deprecated getters on the data model.

### DIFF
--- a/module/data/activity/base-activity.mjs
+++ b/module/data/activity/base-activity.mjs
@@ -1,3 +1,4 @@
+import simplifyRollFormula from "../../dice/simplify-roll-formula.mjs";
 import { formatNumber, staticID } from "../../utils.mjs";
 import FormulaField from "../fields/formula-field.mjs";
 import ActivationField from "../shared/activation-field.mjs";
@@ -116,6 +117,16 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
 
   /* -------------------------------------------- */
   /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * The primary ability for this activity that will be available as `@mod` in roll data.
+   * @type {string|null}
+   */
+  get ability() {
+    return null;
+  }
+
   /* -------------------------------------------- */
 
   /**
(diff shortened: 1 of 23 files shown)


module/data/activity/base-activity.mjs
```diff
@@ -478,6 +528,26 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
...
 
+  /**
+   * Check whether an object exists without triggering potential deprecation warnings.
+   * This can be removed in DnD5e 4.4 when the data shims are removed.
+   * @param {object} object
+   * @param {string} keyPath
+   * @returns {boolean}
+   * @internal
+   */
+  _safePropertyExists(object, keyPath) {
```
Fyorl, line 539: Could be static, but maybe better as a util.

module/data/activity/save-data.mjs
```diff
@@ -82,4 +98,34 @@ export default class SaveActivityData extends BaseActivityData {
...
+        this.save.dc.value = simplifyBonus(this.save.dc.formula, rollData);
+        break;
+      case "spellcasting":
+        ability = this.isSpell ? this.item.system.availableAbilities?.first() : null;
+        ability ??= this.actor?.system.attributes?.spellcasting;
+        break;
+      default:
+        ability = this.save.dc.calculation;
+        break;
+    }
```
Fyorl, line 122: Is there some issue with calling into `this.ability` in `prepareFinalData`? Maybe some values might still change? If not, we should be able to remove some duplication here:

```suggestion
    let ability;
    if ( this.save.dc.calculation === "custom" ) this.save.dc.value = simplifyBonus(this.save.dc.formula, rollData);
    else ability = this.ability;
```

Fyorl (changes requested):

module/data/activity/save-data.mjs
arbron: No problem, just wrote `get ability` after this and never went back to simplify it :)

Fyorl (approved):

PR #3970, opened 2024-08-12
[#1964] Add buttons to activity messages, rollFormula for Utility

arbron:
Buttons created in chat messages through activities can now take `[data-action]` parameters which will automatically call handlers defined in the activity's metadata.

The `UtilityActivity` now has a rolling button which calls the `rollFormula` method, duplicating the functionality of the old method from `Item5e` using the new rolling API.

The old `rollFormula` method has been deprecated and will call the new method provided a `UtilityActivity` is present on the item.

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -707,7 +707,7 @@ export default Base => class extends PseudoDocumentMixin(Base) {
    */
   activateChatListeners(message, html) {
     html.addEventListener("click", event => {
-      const target = event.target.closest("[data-action");
+      const target = event.target.closest("[data-action]");
       if ( target ) this.#onChatAction(event, target, message);
     });
   }
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -14,7 +14,9 @@ export default Base => class extends PseudoDocumentMixin(Base) {
    * @property {string} type                              Type name of this activity.
    * @property {string} img                               Default icon.
    * @property {string} title                             Default title.
+   * @property {typeof ActivitySheet} sheetClass          Sheet class used to configure this activity.
    * @property {object} usage
+   * @property {Record<string, Function>} usage.actions   Actions that can be triggered from the chat card.
    * @property {string} usage.chatCard                    Template used to render the chat card.
    * @property {typeof ActivityUsageDialog} usage.dialog  Default usage prompt.
    */
@@ -26,6 +28,7 @@ export default Base => class extends PseudoDocumentMixin(Base) {
   static metadata = Object.freeze({
     name: "Activity",
     usage: {
+      actions: {},
       chatCard: "systems/dnd5e/templates/chat/activity-card.hbs",
       dialog: ActivityUsageDialog
     }
@@ -202,6 +205,7 @@ export default Base => class extends PseudoDocumentMixin(Base) {
       data: {
         flags: {
           dnd5e: {
+            messageType: "usage",
             activity: { type: this.type, id: this.id, uuid: this.uuid },
             item: { type: this.item.type, id: this.item.id, uuid: this.item.uuid }
           }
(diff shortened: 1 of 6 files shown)


module/documents/activity/utility.mjs
```diff
@@ -21,7 +21,129 @@ export default class UtilityActivity extends ActivityMixin(UtilityActivityData)
...
+     * A hook event that fires before a formula is rolled for an Utility activity.
+     * @function dnd5e.preRollFormulaV2
+     * @memberof hookEvents
+     * @param {UtilityActivity} activity               The activity performing the roll.
+     * @param {BasicRollProcessConfiguration} config   Configuration information for the roll.
+     * @param {BasicRollDialogConfiguration} dialog    Configuration for the roll dialog.
+     * @param {BasicRollMessageConfiguration} message  Configuration for the roll message.
+     * @returns {boolean}                   Explicitly return `false` to prevent the roll from being performed.
+     */
+    if ( Hooks.call("dnd5e.preRollFormulaV2", this, rollConfig, dialogConfig, messageConfig) === false ) return;
```
krbz999, line 97: Are the hook names placeholders that will later be renamed without the V2 once deprecation is over, or is this permanent?

arbron: Probably permanent, since removing the V2 from the name would require another long deprecation period, but I suppose after the old format is dropped we could just call both.

module/documents/activity/mixin.mjs
```diff
@@ -673,6 +696,49 @@ export default Base => class extends PseudoDocumentMixin(Base) {
...
+  /* -------------------------------------------- */
+
+  /**
+   * Activate listeners on a chat message.
+   * @param {ChatMessage} message  Associated chat message.
+   * @param {HTMLElement} html     Element in the chat log.
+   */
+  activateChatListeners(message, html) {
+    html.addEventListener("click", event => {
+      const target = event.target.closest("[data-action");
```
Fyorl, line 710:
```suggestion
      const target = event.target.closest("[data-action]");
```

module/documents/activity/utility.mjs
```diff
@@ -21,7 +21,129 @@ export default class UtilityActivity extends ActivityMixin(UtilityActivityData)
...
+      console.warn(`No formula defined for the activity ${this.name} on ${this.item.name} (${this.uuid}).`);
+      return;
+    }
+
+    const rollConfig = foundry.utils.deepClone(config);
+    rollConfig.rolls = [{ parts: [this.roll.formula], data: this.getRollData() }].concat(config.rolls ?? []);
+
+    const dialogConfig = foundry.utils.mergeObject({
+      configure: true
+    }, dialog);
```
Fyorl, line 71:
```suggestion
    const dialogConfig = foundry.utils.mergeObject({ configure: true }, dialog);
```

module/documents/chat-message.mjs
```diff
@@ -594,6 +595,18 @@ export default class ChatMessage5e extends ChatMessage {
...
   /* -------------------------------------------- */
 
+  /**
+   * Add event listeners for chat messages created from activities.
+   * @param {HTMLElement} html  The chat message HTML.
+   */
+  _activateActivityListeners(html) {
+    if ( !this.getFlag("dnd5e", "activity") ) return;
+    const activity = this.getAssociatedActivity();
+    activity?.activateChatListeners(this, html);
```
Fyorl, line 605:
```suggestion
    this.getAssociatedActivity()?.activateChatListeners(this, html);
```

templates/chat/activity-card.hbs
```diff
@@ -18,7 +18,9 @@
     {{#if buttons}}
     <div class="card-buttons">
         {{#each buttons}}
-        <button type="button" name="{{ @key }}" {{ dnd5e-dataset dataset }}>{{{ label }}}</button>
+        <button type="button" name="{{ @key }}" {{ dnd5e-dataset dataset }}>
+            {{{ icon }}} {{{ label }}}
```
Fyorl, line 22: Might have trouble styling this if we don't have more predictable structure here. The icon is maybe ok if we assume it's always an `<i>` or `<dnd5e-icon>` but the label should probably be wrapped in a `<span>` either here or in `_usageChatButtons`.

Fyorl (changes requested):

Fyorl (approved):

Issue #1037, opened 2021-03-12
Add a field to specify the Reach of a melee weapon attack


zeel01:
Originally in GitLab by @zeel01

Currently, all weapons can be given a short/long range, and they can have a Boolean "Reach" property. However, Reach attacks are not always 10ft., Ancient dragons for instance have 20ft. tail attacks.

Typically DMs might configure this by setting the lower range increment on a melee weapon to represent the "reach" of the weapon. However this is not actually correct, and won't work for weapons which have both the *reach* and *thrown* property.

Examples: Bullywug Royal: Spear (GoS), Colossus of Akros: Spear (MOT), Zariel: Javelin (MTF). Plus, all player characters of the Bugbear Race (VGM) have:

> **Long-Limbed.** When you make a melee attack on your turn, your reach for it is 5 feet greater than normal.

Which means their reach is at least 10ft. for *all* thrown weapons.

Correct Reach values are necessary for determining the Attack of Opportunity range of a creature, so in addition to being an important value for players and DMs to know, it's also important for any automations that would need to detect AOO opportunities.

The use of the range field to represent reach could be confusing for both the users attempting to read the properties of the weapon, as well as to automation tools. And in the case of weapons with both properties, it's simply not currently possible to represent these correctly.

In some cases, a user might use a work around and have two weapons, one representing the ranged attack, the other representing the melee attack. This adds additional confusion, and makes ammunition tracking problematic. If the ranged weapons is self-consuming like a javelin, the melee weapon should no longer be available if there is no ammo (you don't have a javelin to attack with), but the melee attack should not itself consume any ammo. This setup becomes complicated and confusing very quickly.

PR #3972, opened 2024-08-12
Item Sheets V2: Activated Items

Fyorl:
image

Obviously some duplicated code between the main sheet and the activities. Hopefully I can reconcile them when I do a UI pass over the activity sheets.


templates/items/details/details-uses.hbs
```diff
@@ -0,0 +1,72 @@
...
+    </div>
+
+    {{!-- Auto-Destroy --}}
+    {{#if fields.uses.fields.autoDestroy}}
+    {{ formField fields.uses.fields.autoDestroy value=source.uses.autoDestroy localize=true
+                 input=inputs.createCheckboxInput }}
+    {{/if}}
+</fieldset>
+
+{{#if system.hasLimitedUses}}
```
arbron, line 31: This means recovery isn't visible on non-embedded items if they reference an actor property in the max formula (e.g. `@prof`), because that resolves to `0` during data preparation. We'll probably want to tweak `ActivitiesTemplate#hasLimitedUses` to return `true` for non-embedded items if they have any max uses formula in the source data.

templates/items/details/details-feat.hbs
```diff
@@ -0,0 +1,38 @@
...
+
+    {{!-- Max Enchantments --}}
+    {{ formField fields.enchantment.fields.items.fields.max value=source.enchantment.items.max localize=true
+                 label="DND5E.Enchantment.FIELDS.enchantment.items.max.label"
+                 hint="DND5E.Enchantment.FIELDS.enchantment.items.max.hint" }}
+
+    {{!-- Enchantment Replacement --}}
+    {{ formField fields.enchantment.fields.items.fields.period value=source.enchantment.items.period localize=true
+                 choices=config.enchantmentPeriods labelAttr="label" blank="DND5E.UsesPeriods.Never"
+                 hint="DND5E.Enchantment.FIELDS.enchantment.items.period.hint" }}
```
arbron, line 36: Getting an error from these because `system.enchantment` was part of `ActionTemplate` that is no longer included in `FeatData`. Since the only thing left from that are these two fields which are only used on features, we'll need to move `items.max` and `items.period` over to `FeatData` (perhaps migrating to a new path like `system.enchantment.max` and `system.enchantment.period`).

templates/items/details/details-spell.hbs
```diff
@@ -0,0 +1,232 @@
...
+        {{/if}}
+    </div>
+
+    {{!-- Duration --}}
+    <div class="form-group split-group">
+        <label>{{ localize "DND5E.Duration" }}</label>
+        <div class="form-fields">
+
+            {{!-- Amount --}}
+            {{#if system.hasScalarDuration}}
```
arbron, line 125: (there is a bug in the `.scalar` values on these fields at the moment that I fixed in my other PR)

```suggestion
            {{#if system.duration.scalar}}
```

templates/items/details/details-spell.hbs
```diff
@@ -0,0 +1,232 @@
...
+<fieldset>
+    <legend>{{ localize "DND5E.TargetPl" }}</legend>
+
+    {{!-- Target Type --}}
+    <div class="form-group split-group">
+        <label>{{ localize "DND5E.Type" }}</label>
+        <div class="form-fields">
+
+            {{!-- Amount --}}
+            {{#if scalarTarget}}
```
arbron, line 208:
```suggestion
            {{#if system.target.affects.scalar}}
```

templates/items/details/details-spell.hbs
```diff
@@ -0,0 +1,232 @@
...
+                     placeholder=(localize "DND5E.ItemActivationCondition") classes="full-width" }}
+    </div>
+
+    {{!-- Range --}}
+    <div class="form-group split-group">
+        <label>{{ localize "DND5E.Range" }}</label>
+        <div class="form-fields">
+
+            {{!-- Value --}}
+            {{#if system.hasScalarRange}}
```
arbron, line 101:
```suggestion
            {{#if system.range.scalar}}
```

templates/items/details/details-spell.hbs
```diff
@@ -0,0 +1,232 @@
...
+
+            {{!-- Amount --}}
+            {{#if (and source.activation.type (not isCostlessAction))}}
+                {{ formField fields.activation.fields.value value=source.activation.value label="DND5E.Amount" placeholder="—"
+                             localize=true hint=false classes="label-top" }}
+            {{/if}}
+
+            {{!-- Type --}}
+            {{ formField fields.activation.fields.type value=source.activation.type label="DND5E.Cost" localize=true
+                         classes="label-top" hint=false choices=config.abilityActivationTypes blank="" }}
```
arbron, line 87: We'll want to switch this over to using the new `CONFIG.DND5E.activityActivationTypes` with grouping.

templates/items/details/details-spell.hbs
```diff
@@ -0,0 +1,232 @@
...
+<fieldset>
+    <legend>{{ localize "DND5E.Casting" }}</legend>
+
+    {{!-- Activation --}}
+    <div class="form-group split-group">
+        <label>{{ localize "DND5E.SpellCastTime" }}</label>
+        <div class="form-fields">
+
+            {{!-- Amount --}}
+            {{#if (and source.activation.type (not isCostlessAction))}}
```
arbron, line 80:
```suggestion
            {{#if system.activation.scalar}}
```

templates/items/details/details-weapon.hbs
```diff
@@ -0,0 +1,78 @@
...
+    {{#unless isMountable}}
+    {{ formField fields.proficient value=source.proficient blank="DND5E.Automatic" localize=true
+                 choices=config.weaponAndArmorProficiencyLevels }}
+    {{/unless}}
+
+    {{!-- Weapon Properties --}}
+    {{ formField fields.properties options=properties.options label="DND5E.ItemWeaponProperties" localize=true
+                 input=inputs.createMultiCheckboxInput stacked=true classes="checkbox-grid" }}
+
+    {{!-- Range --}}
```
arbron, line 22: We also have the new `system.range.reach` that we should show if the reach property is set.

module/data/item/templates/mountable.mjs
```diff
@@ -30,7 +30,11 @@ export default class MountableTemplate extends SystemDataModel {
         max: new NumberField({ required: true, integer: true, min: 0, label: "DND5E.HitPointsMax" }),
         dt: new NumberField({ required: true, integer: true, min: 0, label: "DND5E.DamageThreshold" }),
         conditions: new StringField({required: true, label: "DND5E.HealthConditions"})
-      }, {label: "DND5E.HitPoints"})
+      }, {label: "DND5E.HitPoints"}),
+      speed: new SchemaField({
+        value: new NumberField({required: true, min: 0, label: "DND5E.Speed"}),
+        conditions: new StringField({required: true, label: "DND5E.SpeedConditions"})
+      }, {label: "DND5E.Speed"})
```
arbron, line 37: Since these are only exposed for equipment items, does it make sense to include them with all mountable equipment?

templates/items/details/details-uses.hbs
Fyorl: I would think checking for a non-empty string is probably good enough for embedded and non-embedded cases. Presumably if you're filling in *some* value, you're expecting that it might have recovery profiles, even if it happens to evaluate to `0` currently.

module/data/item/templates/mountable.mjs
arbron: Maybe `required: false, initial: undefined` on the `SchemaField` so it isn't included by default in the data saved to the DB?

templates/items/details/details-feat.hbs
Fyorl: Used `enchant` since there's a class with the `enchantment` shim.

module/data/item/templates/mountable.mjs
Fyorl: I've made that change, let's try it out. If it doesn't work we can just move it back to `EquipmentData`, it's not a big deal.

module/data/item/templates/activities.mjs
```diff
@@ -90,7 +90,7 @@ export default class ActivitiesTemplate extends SystemDataModel {
    * @type {boolean}
    */
   get hasLimitedUses() {
-    return this.uses.max > 0;
+    return !!this.uses.max;
```
arbron, line 93: You'll need to check source data for this, because during data preparation we resolve the formula and if no roll data is available it defaults to `0`.

templates/items/details/details-weapon.hbs
```diff
@@ -73,6 +52,56 @@
...
+            {{!-- Normal --}}
+            {{ formField fields.range.fields.value value=source.range.value label="DND5E.Normal" localize=true
+                         hint=false classes="label-top" }}
+
+            {{!-- Long --}}
+            {{ formField fields.range.fields.long value=source.range.long label="DND5E.Long" localize=true
+                         hint=false classes="label-top" }}
+
+            {{!-- Reach --}}
+            {{#if (eq system.attackType "melee")}}
```
arbron, line 73: Right now this appears for all melee weapons, do we want to limit to only melee weapons with the "Reach" property? We could add a `hasReach` getter to `WeaponData` alongside the existing `hasRange` getter to handle that.

module/applications/item/item-sheet-2.mjs
```diff
@@ -55,6 +55,97 @@ export default class ItemSheet5e2 extends ItemSheetV2Mixin(ItemSheet5e) {
...
+
+    // Range
+    context.rangeTypes = [
+      ...Object.entries(CONFIG.DND5E.rangeTypes).map(([value, label]) => ({ value, label })),
+      ...Object.entries(CONFIG.DND5E.movementUnits).map(([value, label]) => {
+        return { value, label, group: "DND5E.RangeDistance" };
+      })
+    ];
+    context.hasRange = (this.item.type === "weapon")
+      && (properties.has("thr") || ["simpleR", "martialR"].includes(type?.value));
```
arbron, line 82: We do have the `WeaponData#hasRange` getter that performs this same check.

arbron (approved):
Looking good.

templates/items/details/details-weapon.hbs
Fyorl: My main hesitation was around NPC weapons. I wasn't sure if we really wanted to require checking the 'reach' property for something like the Kraken's Tentacle, because that property has well-defined meaning in game terms, which does not apply to most NPC natural weapons.

> This weapon adds 5 feet to your reach when you attack with it, as well as when determining your reach for opportunity attacks with it

On the other hand, our weapon properties don't always map 1:1 with game terms, and we are looser with some of the definitions. So let me know what you think.

MaxPat931: related conversation from a while back: #1037 
Strixhaven has some npcs with Spell attacks with Reach, e.g.: `Force Strike. Melee or Ranged Spell Attack: +14 to hit, reach 15 ft. or range 120 ft.`

PR #3975, opened 2024-08-12
[#3910] Move rollAttack to AttackActivity

arbron:
Adds the Attack button to attack activities and the `rollAttack` method, which replicates the behavior of that method from `Item5e` but uses the new rolling API. Internally this constructs config objects according to the new API to call the `preRollAttackV2` hook, and then translates that back into the old configuration format for calling the deprecated hooks and the roll constructor.

A new `criticalThreshold` getter on `AttackActivity` now handles most of the functionality of the old getter on `ActionTemplate`. The `criticalThreshold` getter on `ItemDataModel` replaces `_typeCriticalThreshold` which is no longer required (but is still checked for backwards compatibility).

A few of the signatures of the new rolling hooks have changed. The pre hooks no longer pass the activity as the first parameter because that is included as part of `rollConfig`. The post hooks now include the activity as part of the second data object, not the first value. These changes should help in situations like enrichers where you might be rolling attacks or damage without an associated activity.

### Todo
- [ ] Handle ammunition for weapons
- [ ] Refactor configuration positioning into `RollConfigurationDialog`

### DIFF
--- a/module/dice/d20-roll.mjs
+++ b/module/dice/d20-roll.mjs
@@ -1,6 +1,35 @@
 const { Die, NumericTerm, OperatorTerm } = foundry.dice.terms;
 
 /**
+ * Configuration data for the process of rolling d20 rolls.
+ *
+ * @typedef {BasicRollProcessConfiguration} D20RollProcessConfiguration
+ * @property {boolean} elvenAccuracy         Use three dice when rolling with advantage.
+ * @property {boolean} halflingLucky         Add a re-roll once modifier to the d20 die.
+ * @property {D20RollConfiguration[]} rolls  Configuration data for individual rolls.
+ */
+
+/**
+ * D20 roll configuration data.
+ *
+ * @typedef {BasicRollConfiguration} D20RollConfiguration
+ * @property {string[]} parts          Parts used to construct the roll formula, not including the d20 die.
+ * @property {D20RollOptions} options  Options passed through to the roll.
+ */
+
+/**
+ * Options that describe a d20 roll.
+ *
+ * @typedef {BasicRollOptions} D20RollOptions
+ * @property {boolean} advantage       Is the roll granted advantage?
+ * @property {boolean} disadvantage    Is the roll granted disadvantage?
+ * @property {number} criticalSuccess  The value of the d20 die to be considered a critical success.
+ * @property {number} criticalFailure  The value of the d20 die to be considered a critical failure.
+ * @property {number} maximum          Maximum number the d20 die can roll.
+ * @property {number} minimum          Minimum number the d20 die can roll.
+ */
+
+/**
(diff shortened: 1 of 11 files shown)


module/dice/d20-roll.mjs
```diff
@@ -1,5 +1,34 @@
 const { Die, NumericTerm, OperatorTerm } = foundry.dice.terms;
 
+/**
+ * Configuration data for the process of rolling a d20 rolls.
```
krbz999, line 4:
```suggestion
 * Configuration data for the process of rolling a d20 roll.
```

module/dice/d20-roll.mjs
```diff
@@ -1,5 +1,34 @@
...
+/**
+ * Options that describe a d20 roll.
+ *
+ * @typedef {BasicRollOptions} D20RollOptions
+ * @property {boolean} advantage       Is the roll granted advantage?
+ * @property {boolean} disadvantage    Is the roll granted disadvantage?
+ * @property {number} criticalSuccess  The value of the d20 die to be considered a critical success.
+ * @property {number} criticalFailure  The value of the d20 die to be considered a critical failure.
+ * @property {number} maximum          Maximum number the d20 die can roll.
+ * @property {number} minimum          Minimum number the d20 die can roll.
```
krbz999, line 29: My kingdom for an additional `modifiers` property here.

Fyorl: Adding this would be enshrining it as an officially supported part of the API. I don't think we can guarantee that any arbitrary `modifiers` would work with all combinations of other options that also add modifiers like `advantage`, `minimum`, `halflingLucky` etc., and with any other options we want to add in the future.

module/documents/activity/attack.mjs
```diff
@@ -21,7 +22,192 @@ export default class AttackActivity extends ActivityMixin(AttackActivityData) {
...
+    }].concat(config.rolls ?? []);
+
+    const dialogConfig = foundry.utils.mergeObject({
+      configure: true,
+      options: {
+        width: 400,
+        top: config.event ? config.event.clientY - 80 : null,
+        left: window.innerWidth - 710
+      }
+    }, dialog);
```
Fyorl, line 102: Would be nice to have this factored out somewhere as well, assuming other activities are going to be doing the same thing. If it's going to be part of a refactored `D20Roll` though, then no worries.

Fyorl (approved):

module/documents/activity/attack.mjs
arbron: Yeah, we can add something to the `BasicRollConfigurationDialog` static constructor to handle this.

module/dice/d20-roll.mjs
krbz999: We have a utility function in `dnd5e.utils` that could be used to clean the Set.

Issue #3911, opened 2024-07-31
Activity Type: Saving Throw


arbron:
The saving throw activity will be able to call for a saving throw on a target and roll damage. It will contain information on setting the save DC and what type of save is required. It will also contain information on how the damage/effects are applied on a successful save (e.g. "No Damage on Save", "Half Damage on Save", "Full Damage on Save", "No Effects on Save", etc.).

maxobremer:
Hey Arbron, will different activities also be able to have different active effects tied to them? Let's say you want to make the item [Staff of Power](https://www.dndbeyond.com/magic-items/4764-staff-of-power) and have the spells on it as activities. Will it be possible to have an active effect for the hold monster and ray of enfeeblement separately?

PR #3976, opened 2024-08-12
[#3911] Add Saving Throw button for `SaveActivity`

arbron:
Adds a saving throw button to the card generated by the `SaveActivity` and transfer the logic from `_onChatCardAction`.

Modifies the logic in `ChatMessage5e` for hiding buttons to give activites more control over what buttons are visible. Right now they can use `data-visibility` to specify whether a button is available only go GMs (`gm`), to the message creator (blank value), or everyone (`all`). It then asks the activity using the `shouldHideChatButton` method to implement more complex logic if necessary.

### DIFF
--- a/module/data/item/feat.mjs
+++ b/module/data/item/feat.mjs
@@ -179,7 +179,7 @@ export default class FeatData extends ItemDataModel.mixin(
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
   static #migrateEnchantment(source) {
-    if ( foundry.utils.getType(source.enchantment) !== "Object" ) return;
+    if ( foundry.utils.getType(source.enchantment?.items) !== "Object" ) return;
     const { items } = source.enchantment;
     source.enchant ??= {};
     if ( "max" in items ) source.enchant.max = items.max;
--- a/module/data/item/feat.mjs
+++ b/module/data/item/feat.mjs
@@ -179,7 +179,7 @@ export default class FeatData extends ItemDataModel.mixin(
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
   static #migrateEnchantment(source) {
-    if ( !("enchantment" in source) || !("items" in source.enchantment) ) return;
+    if ( foundry.utils.getType(source.enchantment) !== "Object" ) return;
     const { items } = source.enchantment;
     source.enchant ??= {};
     if ( "max" in items ) source.enchant.max = items.max;
--- a/module/documents/activity/utility.mjs
+++ b/module/documents/activity/utility.mjs
@@ -42,7 +42,7 @@ export default class UtilityActivity extends ActivityMixin(UtilityActivityData) 
       icon: '<i class="fa-solid fa-dice" inert></i>',
       dataset: {
         action: "rollFormula",
-        visibility: this.roll.visible ? "all" : "creator"
+        visibility: this.roll.visible ? "all" : undefined
       }
     }];
   }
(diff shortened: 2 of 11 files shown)


module/data/item/feat.mjs
```diff
@@ -179,7 +179,7 @@ export default class FeatData extends ItemDataModel.mixin(
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
   static #migrateEnchantment(source) {
-    if ( !("enchantment" in source) || !("items" in source.enchantment) ) return;
+    if ( foundry.utils.getType(source.enchantment) !== "Object" ) return;
```
Fyorl, line 182: Doesn't this throw an Error now for partial updates to `enchantment` that do not include the `items` property?

module/documents/chat-message.mjs
```diff
@@ -166,26 +166,36 @@ export default class ChatMessage5e extends ChatMessage {
...
-        const optionallyHide = (selector, hide) => {
-          const element = chatCard[0].querySelector(selector);
-          if ( element && hide ) element.style.display = "none";
-        };
-        optionallyHide('button[data-action="summon"]', !SummonsData.canSummon);
-        optionallyHide('button[data-action="placeTemplate"]', !game.user.can("TEMPLATE_CREATE"));
-        optionallyHide('button[data-action="consumeUsage"]', this.getFlag("dnd5e", "use.consumedUsage"));
-        optionallyHide('button[data-action="consumeResource"]', this.getFlag("dnd5e", "use.consumedResource"));
-        return;
+      const isCreator = actor?.isOwner || (this.author.id === game.user.id);
```
Fyorl, line 170: If a message is visible to the creator only, it should probably be visible to the DM too.

Fyorl (changes requested):
Looks good, only two comments, and a question about the following:

> to the message creator (blank value)

In the utility roll PR I think we used `"creator"` as the value for visible to the message creator. If blank is the default, then showing to the creator by default seems reasonable, so I think I prefer blank over the `"creator"` string.

module/data/item/feat.mjs
arbron: Good point (had to change it because the old one was throwing an error when `source.enchantment` was `null`)

arbron:
> In the utility roll PR I think we used `"creator"` as the value for visible to the message creator. If blank is the default, then showing to the creator by default seems reasonable, so I think I prefer blank over the `"creator"` string.

Yeah, no value is treated as `"creator"`.

Originally in utility roll I had it as `visibility: this.roll.visible ? "all" : undefined` but that lead to `dnd5e-dataset` sticking `data-visiblity="undefined"` in the HTML, but I suppose I could just fix out handlebars helper instead :)

module/documents/activity/utility.mjs
```diff
@@ -42,7 +42,7 @@ export default class UtilityActivity extends ActivityMixin(UtilityActivityData)
       icon: '<i class="fa-solid fa-dice" inert></i>',
       dataset: {
         action: "rollFormula",
-        visibility: this.roll.visible ? "all" : "creator"
+        visibility: this.roll.visible ? "all" : undefined
```
Fyorl, line 45: Could be blank as well, no?

```suggestion
        visibility: this.roll.visible ? "all" : ""
```

Fyorl (approved):

module/documents/activity/utility.mjs
arbron: Well, a blank string will results in `data-visibility=""` being added, while this will result in nothing being added (`dnd5e-dataset` doesn't check for falsy values because we might want a blank string to be added).

PR #3978, opened 2024-08-13
[#3912] Implement damage rolls for Damage & Heal activities

arbron:
Adds the `rollDamage` method to the activity mixin which provides a shared method for all activity types that roll damage. This calls into `getDamageConfig` and `processDamagePart` which provide places for individual activity types a place to configure the data provided to damage rolls.

This PR implements rolling damage for `DamageActivity` and `HealActivity`, and leaves out some parts of the damage data preparation that are only relevant to attack damage rolls.

Note: Currently just falls back to using the first of the provided damage types.

## Todo
- [ ] Implement damage rolls for `AttackActivity` and `SaveActivity`
- [ ] Handle attack-specific damage bonuses
- [ ] Handle ammunition damage bonuses
- [ ] Add damage type selection to the damage configuration interface

### DIFF
--- a/module/data/activity/base-activity.mjs
+++ b/module/data/activity/base-activity.mjs
@@ -124,6 +124,10 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
    * @type {string|null}
    */
   get ability() {
+    if ( this.isSpell ) {
+      return this.item.system.availableAbilities?.first()
+        ?? this.actor?.system.attributes?.spellcasting ?? null;
+    }
     return null;
   }
 
@@ -132,7 +136,6 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
   /**
    * Helper property to translate this activity type into the old `actionType`.
    * @type {string}
-   * @deprecated since DnD5e 4.0, targeted for removal in DnD5e 4.4
    */
   get actionType() {
     return this.metadata.data;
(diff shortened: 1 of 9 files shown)


module/data/activity/base-activity.mjs
```diff
@@ -508,6 +511,55 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
...
+
+    return rollConfig;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Process a single damage part into a roll configuration.
+   * @param {DamageData} damage                          Damage to prepare for the roll.
+   * @param {DamageRollProcessConfiguration} rollConfig  Roll configuration being built.
```
Fyorl, line 537:
```suggestion
   * @param {Partial<DamageRollProcessConfiguration>} rollConfig  Roll configuration being built.
```

module/data/activity/base-activity.mjs
```diff
@@ -508,6 +511,55 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
...
+   * @returns {DamageRollConfiguration}
+   */
+  processDamagePart(damage, rollConfig) {
+    const scaledFormula = damage.scaledFormula(rollConfig.scaling);
+    const parts = scaledFormula ? [scaledFormula] : [];
+    const data = this.getRollData();
```
Fyorl, line 543: Not great that we have to call this on every part, but I don't know if it can be helped if this is intended to be a public method. Otherwise if it's only called as part of `getDamageConfig`, we can pass it in from there.

module/data/activity/damage-data.mjs
```diff
@@ -55,4 +55,22 @@ export default class DamageActivityData extends BaseActivityData {
...
+  /** @override */
+  getDamageConfig(config={}) {
+    const rollConfig = super.getDamageConfig(config);
+
+    rollConfig.allowCritical ??= this.damage.critical.allow;
+    if ( this.damage.critical.bonus ) {
+      rollConfig.critical ??= {};
+      rollConfig.critical.bonusDamage = this.damage.critical.bonus;
+    }
+    rollConfig.critical ??= {};
```
Fyorl, line 72: I suppose it would also be nice if we could use `rollConfig.critical.allow` instead of `rollConfig.allowCritical` but I guess `allowCritical` is what the damage roll config expects?

```suggestion
    rollConfig.critical ??= {};
    if ( this.damage.critical.bonus ) rollConfig.critical.bonusDamage = this.damage.critical.bonus;
```

module/dice/damage-roll.mjs
```diff
@@ -1,5 +1,46 @@
...
+ *
+ * @typedef {BasicRollConfiguration} DamageRollConfiguration
+ * @property {DamageRollOptions} [options] - Options passed through to the roll.
+ */
+
+/**
+ * Options that describe a damage roll.
+ *
+ * @typedef {BasicRollOptions} DamageRollOptions
+ * @property {boolean} [isCritical]                    Should critical damage by calculated for this roll?
```
krbz999, line 24:
```suggestion
 * @property {boolean} [isCritical]                    Should critical damage be calculated for this roll?
```

module/documents/item.mjs
```diff
@@ -1339,9 +1339,29 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
-    if ( !this.hasDamage ) throw new Error("You may not make a Damage Roll with this Item.");
+  async rollDamage({ spellLevel, ...options }={}) {
+    foundry.utils.logCompatibilityWarning(
+      "The Item5e#rollDamage method has been deprecated and should now be called directly on an activity.",
+      { since: "DnD5e 4.0", until: "DnD5e 4.4" }
+    );
+
+    let item = this;
+    if ( spellLevel ) {
+      if ( this.type === "spell" ) item = item.clone({ "system.item.level": spellLevel });
```
krbz999, line 1352:
```suggestion
      if ( this.type === "spell" ) item = item.clone({ "system.level": spellLevel });
```

Fyorl (approved):

module/documents/item.mjs
Fyorl: Seems we have this error also in #3975. Let's make sure to clean that up as well before we merge this.

PR #3979, opened 2024-08-13
[#3911] Add damage rolls for `SaveActivity`

arbron:
Adds the damage button for the save activity. Saves are set to not allow critical damage.

Also adds a flag to the chat message indicating how much damage is done on a save, which is displayed in the chat message.

Screenshot 2024-08-13 at 14 43 44

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -1,4 +1,3 @@
-import { SummonsData } from "../data/item/fields/summons-field.mjs";
 import aggregateDamageRolls from "../dice/aggregate-damage-rolls.mjs";
 import DamageRoll from "../dice/damage-roll.mjs";
 import simplifyRollFormula from "../dice/simplify-roll-formula.mjs";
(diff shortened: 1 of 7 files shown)


module/documents/chat-message.mjs
```diff
@@ -324,11 +324,19 @@ export default class ChatMessage5e extends ChatMessage {
...
       const isCritical = (roll.type === "damage") && this.rolls[0]?.options?.critical;
-      const subtitle = roll.type === "damage"
-        ? isCritical ? game.i18n.localize("DND5E.CriticalHit") : game.i18n.localize("DND5E.DamageRoll")
-        : roll.type === "attack"
-          ? game.i18n.localize(`DND5E.Action${item.system.actionType.toUpperCase()}`)
-          : item.system.type?.label ?? game.i18n.localize(CONFIG.Item.typeLabels[item.type]);
+      let subtitle;
+      if ( roll.type === "damage" ) {
+        subtitle = isCritical ? game.i18n.localize("DND5E.CriticalHit") : game.i18n.localize("DND5E.DamageRoll");
+        if ( roll.damageOnSave ) subtitle = `${subtitle} (${
```
Fyorl, line 330: The subtitle is not really intended for large amounts of informational text like this. This would be better as the message or roll flavour text I think.

Fyorl (changes requested):

module/documents/chat-message.mjs
arbron: Screenshot 2024-08-14 at 09 22 38

krbz999: Clearly the entire Silmarillion you shoved in there.

Fyorl (approved):

PR #3980, opened 2024-08-13
[#3910] Add damage rolls for `AttackActivity`

arbron:
Adds the damage button for the attack activity and handles all of the attack-specific bonuses.

This adds a new option to the configuration for `mode`, which will be determined by the attack roll and modifies damage based on whether it is a one-handed, two-handed, off-hand, or thrown attack.

### Todo
- [ ] Handle ammunition bonuses
- [ ] Set attack mode in attack dialog and propagate to next damage roll

### DIFF
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -286,8 +286,9 @@ export default class AttackActivityData extends BaseActivityData {
     roll.base = true;
 
     if ( this.item.type === "weapon" ) {
-      // Ensure `@mod` is present in damage unless it is a off hand attack
-      if ( (rollConfig.mode !== "offHand") && !roll.parts.some(p => p.includes("@mod")) ) roll.parts.push("@mod");
+      // Ensure `@mod` is present in damage unless it is positive and an off-hand attack
+      const includeMod = (rollConfig.mode !== "offHand") || (roll.data.mod < 0);
+      if ( includeMod && !roll.parts.some(p => p.includes("@mod")) ) roll.parts.push("@mod");
 
       // Add magical bonus
       if ( this.item.system.magicalBonus && this.item.system.magicAvailable ) {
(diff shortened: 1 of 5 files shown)


module/data/activity/attack-data.mjs
```diff
@@ -223,4 +223,84 @@ export default class AttackActivityData extends BaseActivityData {
...
+      versatile.types = damage.types;
+      damage = versatile;
+    }
+
+    const roll = super._processDamagePart(damage, rollConfig, rollData);
+    roll.base = true;
+
+    if ( this.item.type === "weapon" ) {
+      // Ensure `@mod` is present in damage unless it is a off hand attack
+      if ( (rollConfig.mode !== "offHand") && !roll.parts.some(p => p.includes("@mod")) ) roll.parts.push("@mod");
```
Fyorl, line 290: The modifier is still added for off-hand attacks if it's negative.

That aside, it's probably more ergonomic to have an `off-hand` weapon property that does this automatically, rather than having to choose the 'off-hand' mode every time you roll with the weapon. This doesn't work for versatile or thrown since in those cases you have a single weapon with different ways to attack, but with dual-wielding you have two separate weapons that can have separate properties no problem.

Fyorl (changes requested):

module/data/activity/attack-data.mjs
Fyorl: Just to clarify: not suggesting we add an off-hand property in this PR. Either we decide the `mode` solution is good, or if we prefer the `properties` solution then we remove this bit of code and defer to later.

arbron: I was thinking of storing the last used mode that you select when you attack. That will make it easy to change if you decide to use your normally off-hand weapon as your main weapon without having to open up the item sheet to change the settings, while still making it fast to activate.

Fyorl: Yeah that makes sense, we need some solution for this for the versatile case anyway, since selecting that option every time is not tenable.

Fyorl (approved):

PR #3986, opened 2024-08-15
[#1964] Implement initial activity usage dialog

arbron:
Adds the initial activity usage dialog which re-implements the item usage dialog in ApplicationV2. The new setup should be more flexible to allow for specific activity types like Enchant and Summon to add their own controls. It also now re-renders when changes are made allowing for dynamic inputs and messages.

The dialog currently doesn't implement the consumption warnings from the original dialog pending the implementation of the consumption system, nor does it re-scale the item when changes to scaling are made.

Usage Dialog V1

### Todo
- [ ] Dynamically re-scale item when changes are made to scaling
- [ ] Display more detailed information on each consumption target (target name, scaled amount consumed)
- [ ] Display proper consumption warnings (will consume last quantity and destroy item, not enough uses left, etc.)

### DIFF
--- a/module/applications/activity/activity-usage-dialog.mjs
+++ b/module/applications/activity/activity-usage-dialog.mjs
@@ -1,11 +1,411 @@
+import { simplifyBonus } from "../../utils.mjs";
 import Application5e from "../api/application.mjs";
+
+const { BooleanField, NumberField, SetField, StringField } = foundry.data.fields;
 
 /**
  * Dialog for configuring the usage of an activity.
  */
 export default class ActivityUsageDialog extends Application5e {
+  constructor(options={}) {
+    super(options);
+    this.#activityId = options.activity.id;
+    this.#item = options.activity.item;
+    this.#config = options.config;
+  }
 
-  // TODO: Implement activation dialog
+  static DEFAULT_OPTIONS = {
+    classes: ["activity-usage"],
+    tag: "form",
+    window: {
+      icon: "",
+      minimizable: false
+    },
+    actions: {
+      use: ActivityUsageDialog.#onUse
+    },
+    activity: null,
+    config: null,
+    form: {
+      handler: ActivityUsageDialog.#onSubmitForm,
+      submitOnChange: true
+    },
+    position: {
+      width: 420
+    }
+  };
+
+  static PARTS = {
+    scaling: {
+      template: "systems/dnd5e/templates/activity/activity-usage-scaling.hbs"
+    },
+    concentration: {
+      template: "systems/dnd5e/templates/activity/activity-usage-concentration.hbs"
+    },
+    consumption: {
+      template: "systems/dnd5e/templates/activity/activity-usage-consumption.hbs"
+    },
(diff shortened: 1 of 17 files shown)


module/applications/activity/activity-usage-dialog.mjs
```diff
@@ -1,11 +1,411 @@
...
+   * @param {ApplicationRenderContext} context  Context being prepared.
+   * @param {HandlebarsRenderOptions} options   Options which configure application rendering behavior.
+   * @returns {Promise<ApplicationRenderContext>}
+   * @protected
+   */
+  async _prepareScalingContext(context, options) {
+    context.hasScaling = true;
+    context.notes = [];
+
+    if ( this.actor.system.spells && this.activity.isSpell && (this.item.system.level > 0) ) {
```
Fyorl, line 299:
```suggestion
    if ( this.activity.requiresSpellSlot ) {
```

module/applications/activity/activity-usage-dialog.mjs
```diff
@@ -1,11 +1,411 @@
...
+      const minimumLevel = this.item.system.level ?? 1;
+      const maximumLevel = Object.values(this.actor.system.spells)
+        .reduce((max, d) => d.max ? Math.max(max, d.level) : max, 0);
+
+      const spellSlotOptions = Object.entries(this.actor.system.spells).map(([value, slot]) => {
```
Fyorl, line 304: We should probably use the same logic as `AbilityUseDialog._createSpellSlotOptions` (or even call into that method since it's static) in order to make sure we pick up any alternative casting options.

Fyorl (changes requested):

module/applications/activity/activity-usage-dialog.mjs
arbron: Maybe I'll have @krbz999 check this, but I believe this should continue working with alt types because all alt types should exist in `actor.system.spells`. The old code looped over spell levels and alt spellcasting types and then fetched the data from `system.spells`, while this loops over `system.spells` and then uses the data there directly.

I did make a change to `prepareAltSlots` so it places the `type` property into each slot so the system can track it back to a specific spellcasting type.

I just did a tweak to hopefully keep the label more in line with the old implementation.

Fyorl: Sounds good to me then. Let's try it out.

Fyorl (approved):

module/applications/activity/activity-usage-dialog.mjs
krbz999: > Maybe I'll have @krbz999 check this, but I believe this should continue working with alt types because all alt types should exist in `actor.system.spells`. The old code looped over spell levels and alt spellcasting types and then fetched the data from `system.spells`, while this loops over `system.spells` and then uses the data there directly.

Yeah that sounds about right. Nothing is ever deleted from the `spells` object, so I believe we checked for the existence of a `.levels` property.

Issue #3724, opened 2024-06-11
Consider always cloning the item before the dnd5e.preUseItem hook


JPMeehan:
As a module developer, I would like to be able to cleanly mutate item properties in the `preUse` hook without it affecting the original item. This would help when trying to model non-spell slot based upcasting or any other kind of "spend resources to upgrade the effect".

The dnd5e system uses `item.clone` to model upcasting, a strategy which is not available to modules because the `dnd5e.preUseItem` hook passes the item directly rather than as part of an object. Rather than reworking the hook signature and all the breaks that would entail, I think a simpler solution would be to _always_ run `item.clone` prior to the hook. Then, modules (and the system itself) can use `item.updateSource` to perform synchronous updates on the cloned item.

PR #3987, opened 2024-08-16
[#1964] Properly scale items when activated

arbron:
Changes how scaling occurs to be able to handle scaling on all items, not just spells. It does this by adding the `dnd5e.scaling` flag which contains the number of steps an item was activated ahove its base level. So for a spell cast at its base level, this would be `0` and it would go up `1` for each level it is upcast.

This value is made available in roll data in two ways. For spells, their roll data is automatically updates to `@item.level` always coresponds to the proper spell level without having to update that value when cloning the item. There is also the `@scaling` value which is the scaling value and `@scaling.increase` which is the increase.

So for a spell that targets one creature at its base level and one additional creature for each slot above base level, you would just enter `@scaling` in the targets field. For a spell that starts at 30 foot range that increase by 10 feet per level you could enter `30 + (10 * @scaling.increase)` into the distance field. This should make scaled values easier to enter without having to subtract the base level from the upcast level like currently.

Cantrips behave a bit differently than normal spells. They override the `scalingIncrease` value provided by `ItemDataModel` setting their scaling increase to a fixed value based on character level. This allows cantrips to scale their damage properly without having to set a specific cantrip scaling mode.

There is some extra logic to handle the `spellLevel` flag provided by spell scrolls and tattoos, but the spell scroll creation code will have to be revisited to ensure it creates the proper activities.

### Todo
- [ ] Rewrite `createScrollFromSpell` to properly copy over activities & other data

### DIFF
--- a/module/applications/activity/activity-usage-dialog.mjs
+++ b/module/applications/activity/activity-usage-dialog.mjs
@@ -130,7 +130,7 @@ export default class ActivityUsageDialog extends Application5e {
 
   /** @inheritDoc */
   async _prepareContext(options) {
-    // TODO: Adjust item scaling when scaling is changed
+    this.#item = this.#item.clone({ "flags.dnd5e.scaling": this.config.scaling ?? 0 });
     return {
       ...await super._prepareContext(options),
       activity: this.activity
@@ -296,7 +296,7 @@ export default class ActivityUsageDialog extends Application5e {
     context.hasScaling = true;
     context.notes = [];
 
-    if ( this.activity.requiresSpellSlot ) {
+    if ( this.activity.requiresSpellSlot && (this.config.scaling !== false) ) {
       const minimumLevel = this.item.system.level ?? 1;
       const maximumLevel = Object.values(this.actor.system.spells)
         .reduce((max, d) => d.max ? Math.max(max, d.level) : max, 0);
@@ -326,7 +326,7 @@ export default class ActivityUsageDialog extends Application5e {
       });
     }
 
-    else if ( this.activity.consumption.scaling.allowed ) {
+    else if ( this.activity.consumption.scaling.allowed && (this.config.scaling !== false) ) {
       const scale = this.activity.consumption.scaling;
       const max = scale.max ? simplifyBonus(scale.max, this.activity.getRollData({ deterministic: true })) : Infinity;
       context.scaling = {
--- a/module/data/item/spell.mjs
+++ b/module/data/item/spell.mjs
@@ -296,6 +296,31 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
   }
 
   /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  get scalingIncrease() {
+    const actor = this.parent.actor;
+    if ( (this.level !== 0) || !actor ) return null;
+    let level = actor.system.details?.level;
+    if ( !level && (actor.type === "npc") ) {
+      if ( this.preparation.mode === "innate" ) level = Math.ceil(actor.system.details.cr);
+      else level = actor.system.details.spellLevel;
+    }
+    return Math.floor(((level ?? 0) + 1) / 6);
+  }
+
+  /* -------------------------------------------- */
+  /*  Helpers                                     */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  getRollData(...options) {
+    const data = super.getRollData(...options);
+    data.item.level = data.item.level + (this.parent.getFlag("dnd5e", "scaling") ?? 0);
+    return data;
+  }
+
+  /* -------------------------------------------- */
   /*  Shims                                       */
   /* -------------------------------------------- */
 
--- a/module/data/item/spell.mjs
+++ b/module/data/item/spell.mjs
@@ -300,7 +300,7 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
   /** @inheritDoc */
   get scalingIncrease() {
     if ( this.level !== 0 ) return null;
-    return Math.floor(((this.parent.actor?.system.cantripLevel ?? 0) + 1) / 6);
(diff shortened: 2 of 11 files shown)


module/applications/activity/activity-usage-dialog.mjs
```diff
@@ -130,7 +130,7 @@ export default class ActivityUsageDialog extends Application5e {
 
   /** @inheritDoc */
   async _prepareContext(options) {
-    // TODO: Adjust item scaling when scaling is changed
+    this.#item = this.#item.clone({ "flags.dnd5e.scaling": this.config.scaling ?? 0 });
```
krbz999, line 133: Is the item always cloned regardless of change to scaling? (re: #3724)

module/data/item/spell.mjs
```diff
@@ -295,6 +295,31 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
...
 
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  get scalingIncrease() {
+    const actor = this.parent.actor;
+    if ( (this.level !== 0) || !actor ) return null;
+    let level = actor.system.details?.level;
+    if ( !level && (actor.type === "npc") ) {
+      if ( this.preparation.mode === "innate" ) level = Math.ceil(actor.system.details.cr);
```
krbz999, line 306: We should add a getter for relevant level property (`details.level` and `details.cr`) on the actor system data models, then this would play nicer with module subtypes.

module/documents/item.mjs
```diff
@@ -1250,7 +1237,9 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     );
 
     let item = this;
-    if ( spellLevel && (this.type === "spell") ) item = item.clone({ "system.item.level": spellLevel });
+    if ( spellLevel && (this.type === "spell") ) {
+      item = item.clone({ "flags.dnd5e.scaling": Math.max(0, item.system.level - spellLevel) }, { keepId: true });
+    }
```
krbz999, line 1242: Ah, answered my own question. Could we?
```suggestion
    if ( this.type === "spell" ) {
      item = item.clone({ "flags.dnd5e.scaling": Math.max(0, item.system.level - (spellLevel ?? 0)) }, { keepId: true });
    }
```

module/applications/activity/activity-usage-dialog.mjs
```diff
@@ -296,7 +296,7 @@ export default class ActivityUsageDialog extends Application5e {
     context.hasScaling = true;
     context.notes = [];
 
-    if ( this.activity.requiresSpellSlot ) {
+    if ( this.activity.requiresSpellSlot && (this.config.scaling !== false) ) {
```
Fyorl, line 299: You should still be able to cast a spell with a higher level slot, even if the spell doesn't scale, no? Or perhaps I have misunderstood this section of the code.

module/documents/activity/mixin.mjs
```diff
@@ -535,7 +545,20 @@ export default Base => class extends PseudoDocumentMixin(Base) {
...
+      usageConfig.scaling = Math.max(0, levelingFlag.value - levelingFlag.base);
+    } else if ( this.isSpell ) {
+      const level = this.actor.system.spells?.[usageConfig.spell?.slot]?.level;
+      if ( level ) usageConfig.scaling = level - item.system.level;
+    }
+
+    if ( usageConfig.scaling ) {
+      foundry.utils.setProperty(messageConfig, "data.flags.dnd5e.scaling", usageConfig.scaling);
+      item.updateSource({ "flags.dnd5e.scaling": usageConfig.scaling });
+      item.prepareData();
```
Fyorl, line 559: `prepareData` should be called automatically as part of `updateSource`.

module/documents/item.mjs
```diff
@@ -1250,7 +1237,9 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     );
 
     let item = this;
-    if ( spellLevel && (this.type === "spell") ) item = item.clone({ "system.item.level": spellLevel });
+    if ( spellLevel && (this.type === "spell") ) {
+      item = item.clone({ "flags.dnd5e.scaling": Math.max(0, item.system.level - spellLevel) }, { keepId: true });
```
Fyorl, line 1241: Has `item.system.level` already been increased here, and `spellLevel` is the base level?

Fyorl (changes requested):

module/applications/activity/activity-usage-dialog.mjs
arbron: Yes, though that occurs in the `use` method rather than here.

module/applications/activity/activity-usage-dialog.mjs
arbron: Spells are always able to scale, this just prevents scaling for specific use cases where they are activated from scrolls or tattoos.

module/documents/item.mjs
arbron: Nope, got the math backwards :O

module/data/item/spell.mjs
arbron: Hmm, that was my original design but NPC casting depends on the preparation mode of the cantrip, at least according to the old logic. I wonder if it would make sense to depart from the previous cantrip scaling behavior and just use something like `this.details.level ?? this.details.spellLevel ?? this.details.cr` (to take NPCs with classes into account also).

Fyorl: Innate Spellcasting on monsters must use their CR for scaling according to the rules in the Monster Manual. Using their configured spellcaster level would be incorrect.

module/applications/activity/activity-usage-dialog.mjs
Fyorl: Ah ok, that sounds good then

module/data/item/spell.mjs
arbron: Fixed, using a helper method now rather than a getter.

Fyorl (approved):

PR #3995, opened 2024-08-18
[#3913] Move summoning functionality into `SummonActivity`

### LINKED ISSUE #3913: Activity Type: Summon
The summoning activity will behave the same as the existing summoning behavior on items. It will require a custom activity usage dialog to handle selecting the summons.

arbron:
Shifts the summoning behavior from `SummonData` into `SummonActivity` and adds a modified dialog to select summoning profile, sizes, and types.

Adds `button` options to `ActivityUseDialog` for customizing the buttons that appear and `display` to control what sections are displayed or not.

### Todo
- [ ] Rework how item changes are applied to avoid issues with activities


module/documents/activity/summon.mjs
```diff
@@ -32,4 +42,612 @@ export default class SummonActivity extends ActivityMixin(SummonActivityData) {
...
+        const effect = (new ActiveEffect({
+          _id: staticID("dnd5eItemChanges"),
+          changes,
+          disabled: false,
+          icon: "icons/skills/melee/strike-slashes-orange.webp",
+          name: game.i18n.localize("DND5E.Summoning.ItemChanges.Label"),
+          origin: this.uuid,
+          flags: {
+            dnd5e: { type: "enchantment" }
+          }
```
Fyorl, line 522: I guess we need a different solution for damage bonus now

```suggestion
          type: "enchantment"
```

Fyorl (approved):

module/documents/activity/summon.mjs
arbron: The whole item changes section will need to be reworked to handle activities.

module/applications/activity/summon-usage-dialog.mjs
```diff
@@ -0,0 +1,113 @@
...
+          break;
+      }
+    }
+    label ??= "—";
+
+    let count = simplifyRollFormula(Roll.replaceFormulaData(profile.count ?? "1", rollData));
+    if ( Number.isNumeric(count) ) {
+      count = parseInt(count);
+      label = `${count} × ${label}`;
+    } else if ( count ) label = `${count} × ${label}`;
```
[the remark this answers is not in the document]
arbron: Looks like, since we have no method of enforcing that a formula contains an integer. I added an extra check in `getPlacement` to avoid any errors if the formula results in a non-integer value.

PR #3997, opened 2024-08-18
[#3920, #3982] Add rules version setting, implement new exhaustion

### LINKED ISSUE #3920: Implement 2024 Exhaustion
The 2024 PHB overhauls Exhaustion; it now inflicts a -2 penalty to D20 Tests and -5' to speed per level, and still has 6 levels.

For backwards compatibility purposes it would probably be a good idea to implement a setting that preserves the 2014 Exhaustion rules.

### LINKED ISSUE #3982: Add setting to toggle between 2014 & 2024 rules
Cover a few changes between 2014 and 2024 rules that should be applied at the same time, rather than one at a time.

- Jack of All Trades doesn't add to initiative
- Remarkable athlete gives advantage on initiative & athletics checks, not half proficiency
- Exhaustion: https://github.com/foundryvtt/dnd5e/issues/3920
- Regain all hit dice on long rest

arbron:
Adds a setting to change between the Legacy (2014) and Modern (2024) rules and implements a few of those rules changes:

- Exhaustion now reduces D20 tests by 2 and speed by 5 for each level of exhaustion, without the other legacy effects.
- Jack of All Trades no longer adds to initiative bonus
- Remarkable athlete no longer adds half proficiency, but does give advantage on initiative rolls
- All hit dice are recovered on a long rest

Closes #3920
Closes #3982

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2989,6 +2989,12 @@
   "THEME": {
     "Name": "Theme",
     "Hint": "Theme that will apply to the UI and all sheets by default. Automatic will be determined by your browser or operating system settings."
+  },
+  "RULESVERSION": {
+    "Name": "Rules Version",
+    "Hint": "Change handling of various rules (including exhausion, Jack of All Trades, Remarkable Athlete, and hit dice recovery) between the 2024 and 2014 rule sets.",
+    "Legacy": "Legacy Rules (2014)",
+    "Modern": "Modern Rules (2024)"
   }
 },
 
--- a/lang/en.json
+++ b/lang/en.json
@@ -2992,7 +2992,7 @@
   },
   "RULESVERSION": {
     "Name": "Rules Version",
-    "Hint": "Change handling of various rules (including exhausion, Jack of All Trades, Remarkable Athlete, and hit dice recovery) between the 2024 and 2014 rule sets.",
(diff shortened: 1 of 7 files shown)


krbz999:
While I am not advocating for a piece-meal selection of modern and legacy ruleset, I know that it would be very well received if the modern exhaustion rule was split into its own setting.

lang/en.json
```diff
@@ -2989,6 +2989,12 @@
   "THEME": {
     "Name": "Theme",
     "Hint": "Theme that will apply to the UI and all sheets by default. Automatic will be determined by your browser or operating system settings."
+  },
+  "RULESVERSION": {
+    "Name": "Rules Version",
+    "Hint": "Change handling of various rules (including exhausion, Jack of All Trades, Remarkable Athlete, and hit dice recovery) between the 2024 and 2014 rule sets.",
```
Fyorl, line 2995: I don't know if we really need to name rules here. I'm sure the list will grow (we're missing Wild Shape, for example), and then these particular choices will seem arbitrary.

Fyorl:
> While I am not advocating for a piece-meal selection of modern and legacy ruleset, I know that it would be very well received if the modern exhaustion rule was split into its own setting.

We can't really base the decision on whether to make an exception here on the unverifiable claim of 'it will be very well received'.

Fyorl (approved):

PR #3999, opened 2024-08-18
[#2212, #3998] Add primary ability to classes

### LINKED ISSUE #2212: Add ability to specify ability score required for multiclassing into a class
Classes generally have a minimum value that a certain ability score (or multiple in some classes) must meet in order to to multiclass with it. This information should be available in the class in some way (though not enforced with core dnd5e). This would also be able to be displayed on the class summary journal sheet in a multiclassing section.

### LINKED ISSUE #3998: Add ability to specify primary ability for classes
In the 2024 rules classes now specify their primary ability which is used to determine multi-classing restrictions and is a useful reference for character creation. The system should add this to class data and show it in the class journal pages if set.

Related to https://github.com/foundryvtt/dnd5e/issues/2212

arbron:
Adds primary ability to classes and displays in on the class journal page.

Closes #2212
Closes #3998

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -598,6 +598,9 @@
 "DND5E.CheckBonus": "Check Bonus",
 
 "DND5E.CLASS": {
+  "Multiclass": {
+    "Title": "Multiclassing"
+  },
   "FIELDS": {
     "hitDice": {
       "label": "Hit Point Die"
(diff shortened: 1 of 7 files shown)


Fyorl (changes requested):
It's possible for a class to have more than one primary ability, for example Paladins have Strength and Charisma as primary abilities, so this probably needs to be a `<multi-select>`. Fighters also have Strength *or* Dexterity listed as their primary ability. Let's not over-complicate this though, a single checkbox that controls whether all ability requirements must be met, or only one, should be enough.

Fyorl (approved):
Nice, thank you

Issue #3983, opened 2024-08-14
Add support for weapon masteries


arbron:
https://www.dndbeyond.com/posts/1742-your-guide-to-weapon-mastery-in-the-2024-players

- [x] Allow selecting a weapon mastery for each weapon
- [x] Add list of weapons in which a player has mastery on their sheet, must first have proficiency
- [x] Allow adding weapon masteries using `TraitAdvancement`
- [ ] Support swapping weapon masteries on attack (see Fighter's Tactical Master feature)

PR #4003, opened 2024-08-19
[#3983] Add weapon masteries to weapons & characters

arbron:
Adds configuration for weapon masteries and adds them to weapon sheets. Also adds weapon mastery tracking to character data and displays them in the weapons list using a symbol similar to resistance bypasses.

Weapon Mastery - List

To configure these masteries for a character, this implements a new version of the `TraitSelector` application built in AppV2. This application will hopefully be the basis for new versions of the config applications for actors.

Weapon Mastery - Config

Also adds a `splitSemicolons` utility method to handle splitting semi-colon-separated lists and removing empty entries.


module/applications/actor/api/base-config-sheet.mjs
```diff
@@ -0,0 +1,68 @@
...
+
+  /* -------------------------------------------- */
+  /*  Event Listeners and Handlers                */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _onRender(context, options) {
+    await super._onRender(context, options);
+    if ( !this.isEditable ) return;
+    for ( const override of this._getActorOverrides() ) {
```
Fyorl, line 37: I don't know if we really want to do this in the config sheets. If you have an AE granting you a proficiency, it should appear on your sheet in play mode. If you switch to edit mode to edit the traits, I think it should just show you the source values, never the derived ones.

templates/actors/config/weapons-config.hbs
```diff
@@ -0,0 +1,53 @@
...
+    </fieldset>
+</section>
+
+{{#*inline ".weapon-list"}}
+<ol class="unlist trait-list">
+    {{#if (and children topLevel (not otherGroup))}}
+    <li>
+        <label class="name">{{ localize "DND5E.TraitAll" category=label }}</label>
+        <div class="proficiency">
+            {{ formInput @root.checkbox name=(concat "system.traits.weaponProf.value." @key) value=chosen }}
```
Fyorl, line 31: We should use `dnd5e-checkbox`es for these, and below.

```suggestion
            {{ formInput @root.checkbox name=(concat "system.traits.weaponProf.value." @key) value=chosen input=inputs.createCheckboxInput }}
```

```js
// BaseConfigSheet#_prepareContext
context.inputs = { ...foundry.applications.fields, ...dnd5e.applications.fields };
```

Fyorl (changes requested):
Looks great, thank you Jeff. In addition to my PR comments, I had some small styling comments too:

image

We can go with the square buttons for our AppV2 sheets (you really hate the circle-clipped document link, huh?), but in that case we should change the Compendium Browser header buttons to match, since that's the only other V2-styled AppV2 app that we have I think. (not required in this PR)

image

We should make sure we are using consistent V2 form styles, so this text input, for example, should match the text inputs in the V2 item sheet. I am not actually sure why it's not already. Maybe because it's not in a `.form-group`? We maybe don't need to put the V2 form styles behind the `.form-group` selector.

arbron:
> We should make sure we are using consistent V2 form styles, so this text input, for example, should match the text inputs in the V2 item sheet. I am not actually sure why it's not already. Maybe because it's not in a `.form-group`? We maybe don't need to put the V2 form styles behind the `.form-group` selector.

I moved the form styles out of `.form-group`. Had to fix a specificity issue with `item-list-controls`, but that is the only issue I spotted.

Changed over the checkboxes but am having an issue with the checkbox heights when all of the entries are not in the unchecked state:

https://github.com/user-attachments/assets/0e54f79c-ca04-4d02-ba16-7360c91e2f89

Haven't been able to track down exactly what is causing that so if you had any insight it would be welcome.

Fyorl (approved):

PR #4004, opened 2024-08-19
[#3910, #3911, #3912] Add damage type selection to roll dialog

### LINKED ISSUE #3911: Activity Type: Saving Throw

### LINKED ISSUE #3912: Activity Type: Damage/Healing

arbron:
Adds the damage type selection to the damage roll dialog, falling back to the first damage type specified if the dialog is skipped.

Closes #3911
Closes #3912

### DIFF
--- a/templates/chat/roll-dialog.hbs
+++ b/templates/chat/roll-dialog.hbs
@@ -4,6 +4,14 @@
         <label>{{ localize "DND5E.Formula" }} {{#if type}}({{type}}){{/if}}</label>
         <input type="text" name="formula" value="{{formula}}" disabled>
     </div>
+    {{#if types}}
+    <div class="form-group">
+        <label>{{ localize "DND5E.DamageType" }}</label>
+        <select name="roll.{{ @index }}.type">
+            {{ selectOptions types }}
+        </select>
+    </div>
+    {{/if}}
     {{/each}}
     {{#if chooseModifier}}
     <div class="form-group">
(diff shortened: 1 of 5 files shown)


templates/chat/roll-dialog.hbs
```diff
@@ -4,6 +4,14 @@
         <label>{{ localize "DND5E.Formula" }} {{#if type}}({{type}}){{/if}}</label>
         <input type="text" name="formula" value="{{formula}}" disabled>
     </div>
+    {{#if types}}
+    <div class="form-group">
+        <label>{{ localize "DND5E.DamageType" }}</label>
+        <select name="roll.{{ @index }}.type">
```
Fyorl, line 10: I imagine that in most cases, there's only 1 damage type, so I don't know if we really need to show a drop-down all the time. I suppose the alternative would be to just not show the damage type at all if there's only one, or show a readonly field or something. Both of those options are more work though so maybe it's not worth it. I imagine this dialog might get some updates, so maybe we revisit it then.

Fyorl (approved):

PR #4005, opened 2024-08-20
[#3910] Add attack mode selection, relay mode to damage rolls

arbron:
Adds attack modes that can be declared by item and are selectable within the attack roll dialog. When one of these modes is selected it is stored in the attack roll message flags.

When the damage button is clicked for an `AttackActivity`, it fetches the attack mode from the most recent attack roll performed from the same chat message and passes that to `rollDamage`. This is facilitated by a new `rollMessages` entry in the registry.

### Todo
- [ ] Remember previously selected attack mode for each weapon so user doesn't have to select it each time

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1430,6 +1430,14 @@
 "DND5E.FilterGroupAction": "Group by Action",
 "DND5E.FilterNoSpells": "No spells found for this set of filters.",
 "DND5E.NoSpellLevels": "This character has no spellcaster levels, but you may add spells manually.",
+
+"DND5E.FLAGS": {
+  "EnhancedDuelWielding": {
+    "Name": "Enhanced Duel Wielding",
+    "Hint": "Allow bonus action extra attacks using any melee weapon without the Two-Handed property."
+  }
+},
+
 "DND5E.FlagsInstructions": "Configure character features and traits which fine-tune behaviors of the DnD5e system.",
 "DND5E.FlagsSave": "Update Special Traits",
 "DND5E.FlagsTitle": "Configure Special Traits",
--- a/lang/en.json
+++ b/lang/en.json
@@ -1432,8 +1432,8 @@
 "DND5E.NoSpellLevels": "This character has no spellcaster levels, but you may add spells manually.",
 
 "DND5E.FLAGS": {
-  "EnhancedDuelWielding": {
-    "Name": "Enhanced Duel Wielding",
+  "EnhancedDualWielding": {
+    "Name": "Enhanced Dual Wielding",
     "Hint": "Allow bonus action extra attacks using any melee weapon without the Two-Handed property."
   }
 },
--- a/lang/en.json
+++ b/lang/en.json
@@ -517,6 +517,13 @@
     "Unarmed": "Unarmed",
     "Weapon": "Weapon"
   },
+  "Mode": {
+    "Label": "Attack Mode",
+    "Offhand": "Offhand",
+    "OneHanded": "One-Handed",
+    "Thrown": "Thrown",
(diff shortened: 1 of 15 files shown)


module/registry.mjs
```diff
@@ -61,11 +61,72 @@ class EnchantmentRegisty {
   }
 }
 
+/* -------------------------------------------- */
+/*  Message Rolls                               */
+/* -------------------------------------------- */
+
+class RollMessageRegistry {
```
Fyorl, line 68: I have a preference for calling this just `MessageRegistry` and exporting it as `registry.messages`. It's terser, and maybe we want to repurpose it later. I know it currently only works with roll messages due to its reliance on the `roll.type` flag.

Fyorl (changes requested):

module/data/item/weapon.mjs
```diff
@@ -246,6 +246,39 @@ export default class WeaponData extends ItemDataModel.mixin(
...
+      });
+
+      // Weapons with the "Two-Handed" property or with the "Versatile" property will have Two-Handed attack
+      if ( this.isVersatile || this.properties.has("two") ) modes.push({
+        value: "twoHanded", label: game.i18n.localize("DND5E.ATTACK.Mode.TwoHanded")
+      });
+    }
+
+    // Weapons with the "Light" property will have Offhand attack
+    if ( this.properties.has("lgt") ) modes.push({
```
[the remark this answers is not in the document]
arbron: Hmm, I'd prefer not to add it to all users since most users can't use it. Maybe we can just take this opportunity to add an Enhanced Duel Wielding flag to handle that feat.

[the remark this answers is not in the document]
Fyorl: I just don't see the harm in restricting it. If it's not applicable they can just... not select it, right? We use this philosophy throughout the system.

lang/en.json
```diff
@@ -1423,6 +1430,14 @@
 "DND5E.FilterGroupAction": "Group by Action",
 "DND5E.FilterNoSpells": "No spells found for this set of filters.",
 "DND5E.NoSpellLevels": "This character has no spellcaster levels, but you may add spells manually.",
+
+"DND5E.FLAGS": {
+  "EnhancedDuelWielding": {
+    "Name": "Enhanced Duel Wielding",
```
Fyorl, line 1436:
```suggestion
    "Name": "Enhanced Dual Wielding",
```

lang/en.json
```diff
@@ -1423,6 +1430,14 @@
 "DND5E.FilterGroupAction": "Group by Action",
 "DND5E.FilterNoSpells": "No spells found for this set of filters.",
 "DND5E.NoSpellLevels": "This character has no spellcaster levels, but you may add spells manually.",
+
+"DND5E.FLAGS": {
+  "EnhancedDuelWielding": {
```
Fyorl, line 1435:
```suggestion
  "EnhancedDualWielding": {
```

module/config.mjs
```diff
@@ -3193,6 +3193,12 @@ DND5E.characterFlags = {
     section: "DND5E.Feats",
     type: Boolean
   },
+  enhancedDualWielding: {
+    name: "DND5E.FLAGS.EnhancedDuelWielding.Name",
+    hint: "DND5E.FLAGS.EnhancedDuelWielding.Hint",
```
Fyorl, line 3198:
```suggestion
    name: "DND5E.FLAGS.EnhancedDualWielding.Name",
    hint: "DND5E.FLAGS.EnhancedDualWielding.Hint",
```

Fyorl (approved):

PR #4012, opened 2024-08-20
[#1866, #4011] Update ammo types, add ammo damage & type to weapon

### LINKED ISSUE #1866: Add base ammunition types & allow configuring them on weapons
Currently, the only way to configure ammunition on a weapon is to have that item on your sheet and to select a specific type of ammo. This means that adding a crossbow to your sheet requires additional configuration in the item sheet before it can be used. Adding the ability to specify a base ammunition type that a weapon consumes would allow clearer and easier ammunition configuration.

```javascript
/**
 * The basic ammunition types in 5e.
 * @enum {string}
 */
DND5E.ammunitionIds = {
  arrow: "3c7JXOzsv55gqJS5",
  blowgunNeedle: "gBQ8xqTA5f8wP5iu",
  crossbowBolt: "SItCnYBqhzqBoaWG",
  slingBullet: "z9SbsMIBZzuhZOqT"
};
```

Weapons could be configured to specify what type of these ammunitions they consume. Then, when added to the sheet if ammunition consumption is set they will automatically attempt to associated themselves with the appropriate ammunition. The ammunition consumption dropdown could also be restricted to only ammo items with the same base type.

This might also benefit from adding an ammo dropdown to the inventory for any weapon that uses a specific ammo type, allowing to be easily selected without manually opening the item sheet. TidySheet5e implements a similar UI design:

Screen Shot 2022-10-11 at 10 37 36

### LINKED ISSUE #4011: Adjust ammunition type names to match updated rules
The new rulebooks list it as just "Needle" rather than "Blowgun Needle" and "Bolt" rather than "Crossbow Bolt" that the current system uses, and introduces "Firearm Bullet" in addition to "Sling Bullet".

arbron:
Updates the names for ammunition types and add "Firearm Bullet" type to bring everything up to date with the 2024 rules.

Adds damage data to consumable items that is only visible for ammunition. This damage has an additional option defining whether it supplements the weapon damage or replaces it.

Adds a "Ammunition Type" field to weapons defining what type of ammo they use. This data will be used in a followup PR to add a ammo selection UI to the attack roll dialog that only lists ammo that can be used with the weapon.

Also modified the damage section on weapons to add a hint about how the weapon damage is used and how `@mod` isn't necessary. Plus hide damage types for versatile damage since the system just fetches those from the base damage. Finally it displays a placeholder for the number and die fields in versatile damage to indicate that those fields can be inferred automatically.

Closes #1866
Closes #4011

### Todo
- [x] Display ammo selection in attack dialog
- [x] Handle ammo bonuses in attack & damage rolls

### DIFF
--- a/module/data/item/consumable.mjs
+++ b/module/data/item/consumable.mjs
@@ -1,5 +1,7 @@
 import { filteredKeys } from "../../utils.mjs";
 import { ItemDataModel } from "../abstract.mjs";
+import BaseActivityData from "../activity/base-activity.mjs";
+import DamageField from "../shared/damage-field.mjs";
 import UsesField from "../shared/uses-field.mjs";
 import ActivitiesTemplate from "./templates/activities.mjs";
 import EquippableItemTemplate from "./templates/equippable-item.mjs";
@@ -9,7 +11,7 @@ import ItemTypeTemplate from "./templates/item-type.mjs";
 import PhysicalItemTemplate from "./templates/physical-item.mjs";
 import ItemTypeField from "./fields/item-type-field.mjs";
 
-const { BooleanField, NumberField, SetField, StringField } = foundry.data.fields;
+const { BooleanField, NumberField, SchemaField, SetField, StringField } = foundry.data.fields;
 
 /**
  * Data definition for Consumable items.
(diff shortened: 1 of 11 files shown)


module/data/item/consumable.mjs
```diff
@@ -78,11 +96,26 @@ export default class ConsumableData extends ItemDataModel.mixin(
...
   }
 
   /* -------------------------------------------- */
 
+  /**
+   * Migrate weapon damage from old parts.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static #migrateDamage(source) {
+    if ( "value" in (source.damage ?? {}) ) return;
```
Fyorl, line 110: Should this be `base`?

```suggestion
    if ( "base" in (source.damage ?? {}) ) return;
```

Fyorl (approved):

PR #4013, opened 2024-08-20
Add cover statuses

mattexdee:
This will allow features which provide cover regardless of position to include an Active Effect that applies the relevant status. It can later be checked in AC calculations and dex saves.


arbron:
Screenshot 2024-09-11 at 14 10 44

krbz999:
Wonder if it would not be cleaner to add a single condition with 3 levels. It should not be too difficult to change the name of the effect at will, and this would also prevent users from applying several different types of cover at the same time.

lang/en.json
```diff
@@ -2891,6 +2891,8 @@
   "StatusEthereal": "Ethereal",
   "StatusExceedingCarryingCapacity": "Exceeding Carrying Capacity",
   "StatusFlying": "Flying",
+  "StatusFullCover": "Full Cover",
```
Fyorl, line 2894: We should use the game term, which is 'Total Cover'.

Fyorl (changes requested):

arbron:
> Wonder if it would not be cleaner to add a single condition with 3 levels. It should not be too difficult to change the name of the effect at will, and this would also prevent users from applying several different types of cover at the same time.

I think three statuses is good, because unlike exhaustion you don't cycle through levels of cover, you just apply the one you want directly. What I think I'll add is that when one level of cover is added it removes any current levels of cover.

Fyorl (approved):

PR #4014, opened 2024-08-20
[#1039, #3268, #3314] Add ammo selection, consumption, & bonuses

### LINKED ISSUE #1039: A flag to disable ammo consumption when calling Item5e.rollAttack()
Originally in GitLab by @CarlosFdez

So currently I'm using rollAttack() to get the item's attack formula in a way that is compatible with other modules that inject themselves into the pipeline with libwrapper. However there is no way to disable ammo consumption with rollAttack(), which is a concern when your module makes it possible to not consume ammo on certain rolls.

For now I plan to use libwrapper to override rollAttack() to strip that functionality if chatMessage is disabled, however some way in the core module to do that would be quite helpful.

(also if you make a spell or anything with a dialog box consume ammo, it consumes ammo when the card is displayed and when the attack is rolled, so there's probably merit into looking into the ammo consumption pipeline in general.

### LINKED ISSUE #3268: Show damage button in chat if ammo has damage but not weapon
### User Experience

The software does currently not allow ammunition to cause the damage when used by a weapon in 5e.

Example:
I fire a cannon. The cannon itself can be loaded with 7 different types of ammunition, each with unique range and damage properties. Currently, I need to configure each ammo for each cannon type as its own weapon and manually assign those to the ships.

Reasoning: The submanual for naval warfare contains roughly 60 different cannons. This, multiplied by about 4 types of ammo each cannon and calibre can use, means 180 items and manual swaps, making it impractical to use it.

### LINKED ISSUE #3314: Ability to chose specific ammunition when firing a ranged weapon
When a ranged weapon fires it automatically consumes the ammunition it is tied to. This can be quite the hassle when players want to switch ammunition as they need to edit the ranged weapon, click details, find resource consumption, and then change the ammunition. It would be so much easier if a player fires a ranged weapon and a pop up asks the player "what kind of ammunition do you want to use?"(similar to spells). 

TDLR: Please add pop up that asks the player what ammo they want to use.

arbron:
Adds a list of usable ammunition types to the attack dialog determined by the ammunition type on the weapon. When one of these is selected any magic bonus it has is added to the attack roll and a piece of ammunition is consumed. If the consumption destroys the ammunition then it is stored in the roll message to ensure it can be used for damage calculation.

When rolling damage the ammunition used on the last attack is sent to the `rollDamage` method. If the ammo has its own damage it either replaces the weapon's base damage or adds a new damage part. The ammo's magical bonus is added to the base damage.

Closes #1039
Closes #3268
Closes #3314

### Todo
- [ ] Remember previously-selected ammunition

### DIFF
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -267,13 +267,13 @@ export default class AttackActivityData extends BaseActivityData {
         // If mode is "replace" and base part is present, replace the base part
         if ( ammo.damage.replace & (basePartIndex !== -1) ) {
           damage.base = true;
-          rollConfig.rolls.splice(basePartIndex, 1, super._processDamagePart(damage, config, rollData));
+          rollConfig.rolls.splice(basePartIndex, 1, this._processDamagePart(damage, config, rollData));
         }
 
         // Otherwise stick the ammo damage after base part (or as first part)
         else {
           damage.ammo = true;
-          rollConfig.rolls.splice(basePartIndex + 1, 0, super._processDamagePart(damage, rollConfig, rollData));
+          rollConfig.rolls.splice(basePartIndex + 1, 0, this._processDamagePart(damage, rollConfig, rollData));
         }
       }
     }
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -233,6 +233,7 @@ export default class AttackActivityData extends BaseActivityData {
 
   /**
    * @typedef {AttackDamageRollProcessConfiguration} [config={}]
+   * @property {Item5e} ammunition                                      Ammunition used with the attack.
    * @property {"oneHanded"|"twoHanded"|"offhand"|"thrown"} attackMode  Attack mode.
    */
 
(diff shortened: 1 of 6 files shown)


module/documents/activity/attack.mjs
```diff
@@ -177,14 +184,21 @@ export default class AttackActivity extends ActivityMixin(AttackActivityData) {
     const roll = await d20Roll(oldRollConfig);
     if ( roll === null ) return;
 
+    const ammo = this.actor?.items.get(roll.options.ammunition);
+    let ammoUpdate = null;
+    if ( ammo ) {
+      ammoUpdate = { id: ammo.id, quantity: Math.max(0, ammo.system.quantity - 1) };
+      ammoUpdate.destroy = ammo.system.uses.autoDestroy && (ammoUpdate.quantity === 0);
```
krbz999, line 191: Ammo isn't currently destroyed when the last of it is spent, and that'd be pretty annoying when a ranged character has to refill all the time.

arbron: Well, it would only be deleted if you have "Destroy on Empty" set on the item, which it isn't for any of the SRD ammo.

module/data/activity/attack-data.mjs
```diff
@@ -244,24 +245,38 @@ export default class AttackActivityData extends BaseActivityData {
...
+      // Add the ammunition's damage
+      if ( ammo.damage.base.formula ) {
+        const basePartIndex = rollConfig.rolls.findIndex(i => i.base);
+        const damage = ammo.damage.base.clone();
+        const rollData = this.getRollData();
+
+        // If mode is "replace" and base part is present, replace the base part
+        if ( ammo.damage.replace & (basePartIndex !== -1) ) {
+          damage.base = true;
+          rollConfig.rolls.splice(basePartIndex, 1, super._processDamagePart(damage, config, rollData));
```
Fyorl, line 270: `this._processDamagePart`? Or are we deliberately skipping this layer of the implementation? It would mean that any subclasses that override `_processDamagePart` will not have their method called here.

```suggestion
          rollConfig.rolls.splice(basePartIndex, 1, this._processDamagePart(damage, config, rollData));
```

module/data/activity/attack-data.mjs
```diff
@@ -244,24 +245,38 @@ export default class AttackActivityData extends BaseActivityData {
...
+        }
+
+        // Otherwise stick the ammo damage after base part (or as first part)
+        else {
+          damage.ammo = true;
+          rollConfig.rolls.splice(basePartIndex + 1, 0, super._processDamagePart(damage, rollConfig, rollData));
```
Fyorl, line 276:
```suggestion
          rollConfig.rolls.splice(basePartIndex + 1, 0, this._processDamagePart(damage, rollConfig, rollData));
```

module/data/item/weapon.mjs
```diff
@@ -235,12 +235,12 @@ export default class WeaponData extends ItemDataModel.mixin(
     if ( this.parent.labels.damages?.length ) {
       const config = { ...CONFIG.DND5E.damageTypes, ...CONFIG.DND5E.healingTypes };
       context.info.push({ value: this.parent.labels.damages.reduce((str, { formula, damageType }) => {
-        const { label, icon } = config[damageType];
+        const type = config[damageType];
```
Fyorl, line 238: Probably going to conflict with a similar change on the `activities/ui` branch.

Fyorl (approved):

module/documents/activity/attack.mjs
Fyorl: Probably need a different hint for ammo consumables only, since I think the normal `autoDestroy` consumable behaviour is to reduce the quantity by 1 when charges are used up (and then delete it at 0). But for ammo we ignore charges and always reduce the quantity by 1, but only destroy it at 0 if `autoDestroy` is set, which is fairly different behaviour.

module/data/item/weapon.mjs
arbron: Yeah, saw we both fixed the same bug :)

module/documents/activity/attack.mjs
arbron: I've been thinking we need two separate options. One allows reducing the quantity when uses are used up, and the second for destroying the item when its quantity reaches zero from consumption. That would allow for situations where you want to just have the quantity hit `0` by using uses but keep the item on the sheet.

Fyorl: I think the default behaviour we have currently is sensible, we just need a different hint to explain it. We can't add configuration for every niche use case and preference, otherwise the UI gets overwhelming.

PR #4015, opened 2024-08-21
Activity Sheets: Attack

Fyorl:
A lot of this code is mostly reshuffling the already-styled fields for uses, targets, duration, activation, etc. so that they can be shared between Item and Activity sheets, but there is bespoke styling for various elements. I've also tried to consolidate our system V2 AppV2 styles so they can be used automatically by any new AppV2 apps we make.

image

image

image


less/v2/apps.less
```diff
@@ -760,27 +760,80 @@
...
       position: absolute;
-      inset-block-start: calc(var(--header-height) - 4px);
+      inset-block-start: calc(var(--header-height) - 20px);
       inset-inline: 0;
       color: var(--color-text-dark-primary);
       font-family: var(--dnd5e-font-roboto-slab);
       font-weight: bold;
       font-size: var(--font-size-24);
       text-align: center;
+      pointer-events: none;
```
arbron, line 775: This does greatly reduce the draggable area of the header. I tested using a `z-index` instead and that allowed continued interaction with the header buttons while still allowing the expanded drag area.

```suggestion
      z-index: -1;
```

templates/activity/parts/activity-effects.hbs
```diff
@@ -1,29 +1,38 @@
...
-    </button>
 </fieldset>
 
 {{#*inline ".effect"}}
 <div class="details flexrow">
     <img class="gold-icon" src="{{ effect.img }}" alt="{{ effect.name }}">
     {{{ dnd5e-linkForUuid effect.uuid }}}
     <div class="list-controls flexrow">
-        <button type="button" class="unbutton" data-action="deleteEffect"
+        <button type="button" class="unbutton control-button" data-action="deleteEffect"
```
arbron, line 32: The one issue with removing the `<multi-select>` is there is no longer a method to unassociated an effect with an activity without deleting it. That could be annoying if you accidentally associate one, the only way to fix it would be to delete either the activity or the effect.

Perhaps we could replace the standard `deleteDialog` in that case with one that gives the option of just removing from the list. Or we could have the default behavior just remove it from the list and when Shift is held down swap over to triggering effect deletion (with an icon swap like we have in the chat log).

arbron (changes requested):
Everything's really coming together!

In addition to my other comments, I noticed that field placeholders are really getting lost when they are near the bottom of the sheet:

Screenshot 2024-08-21 at 13 48 25

templates/activity/parts/activity-effects.hbs
```diff
@@ -1,29 +1,38 @@
...
     </ul>
-    <button type="button" class="unbutton" data-action="addEffect">
-        {{ localize "DND5E.EFFECT.Action.Create" }}
-    </button>
 </fieldset>
 
 {{#*inline ".effect"}}
 <div class="details flexrow">
     <img class="gold-icon" src="{{ effect.img }}" alt="{{ effect.name }}">
     {{{ dnd5e-linkForUuid effect.uuid }}}
```
arbron, line 30: This causes a bug when viewing an activity within a compendium because it uses `fromUuidSync` which cannot retrieve embedded documents in that context.

templates/activity/parts/activity-effects.hbs
Fyorl: Added a separate 'dissociate' control since we have a control list, and it's a slightly less involved solution.

arbron (approved):

PR #4016, opened 2024-08-21
[#1964] Re-implement uses recovery, adjust resting design

arbron:
Re-implements uses recovery to support uses on items and activities and multiple recovery profiles. Rest types now declare what type of periods they recover in prefered order.

This also introduces some other changes to the resting system to support better configuration of the process, allowing rest types to declare what types of spellcasting slots they recover and whether they recover hit points or hit dice. This should make it more feasible to implement custom rest types.

### DIFF
--- a/module/data/shared/uses-field.mjs
+++ b/module/data/shared/uses-field.mjs
@@ -88,8 +88,15 @@ export default class UsesField extends SchemaField {
 
     // Search the recovery profiles in order to find the first matching period,
     // and then find the first profile that uses that recovery period
-    const matchingPeriod = periods.find(p => this.uses.recovery.find(r => r.period === p));
-    const profile = this.uses.recovery.find(r => r.period === matchingPeriod);
+    let profile;
+    for ( const period of periods ) {
+      for ( const recovery of this.uses.recovery ) {
+        if ( recovery.period === period ) {
+          profile = recovery;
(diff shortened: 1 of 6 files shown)


module/data/shared/uses-field.mjs
```diff
@@ -71,4 +71,71 @@ export default class UsesField extends SchemaField {
...
+   * @param {object} rollData   Roll data to use when evaluating recover formulas.
+   * @returns {Promise<{ updates: object, rolls: BasicRoll[] }|false>}
+   */
+  static async recoverUses(periods, rollData) {
+    if ( !this.uses.recovery.length ) return false;
+
+    // Search the recovery profiles in order to find the first matching period,
+    // and then find the first profile that uses that recovery period
+    const matchingPeriod = periods.find(p => this.uses.recovery.find(r => r.period === p));
+    const profile = this.uses.recovery.find(r => r.period === matchingPeriod);
```
Fyorl, line 92: Avoids some unnecessary iteration.
```suggestion
    let profile;
    for ( const period of periods ) {
      for ( const recovery of this.uses.recovery ) {
        if ( recovery.period === period ) {
          profile = recovery;
          break;
        }
      }
    }
```

Fyorl (approved):

PR #4018, opened 2024-08-22
[#1964] Add back Place Template & Consume Resource buttons

arbron:
Add the "Place Template" and "Consume Resource" buttons back to the chat card if necessary.

Also modified the `clone` data on items to automatically call `prepareFinalAttributes` to prevent bugs caused by cloned items not being fully prepared.

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -1,4 +1,5 @@
 import ActivityUsageDialog from "../../applications/activity/activity-usage-dialog.mjs";
+import AbilityTemplate from "../../canvas/ability-template.mjs";
 import { ConsumptionError } from "../../data/activity/fields/consumption-targets-field.mjs";
 import { damageRoll } from "../../dice/dice.mjs";
 import PseudoDocumentMixin from "../mixins/pseudo-document.mjs";
@@ -172,8 +173,6 @@ export default Base => class extends PseudoDocumentMixin(Base) {
 
     // Create an item clone to work with throughout the rest of the process
     let item = this.item.clone({}, { keepId: true });
-    item.prepareData();
-    item.prepareFinalAttributes();
     let activity = item.system.activities.get(this.id);
 
     const usageConfig = activity._prepareUsageConfig(usage);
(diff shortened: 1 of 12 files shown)


module/documents/activity/mixin.mjs
```diff
@@ -645,12 +644,13 @@ export default Base => class extends PseudoDocumentMixin(Base) {
...
       supplements.push(`<strong>${game.i18n.localize("DND5E.Materials")}</strong> ${data.materials.value}`);
     }
+    const buttons = this._usageChatButtons();
     return {
       activity: this,
       actor: this.item.actor,
       item: this.item,
       token: this.item.actor?.token,
-      buttons: this._usageChatButtons(),
+      buttons: buttons.length ? this._usageChatButtons() : null,
```
Fyorl, line 653:
```suggestion
      buttons: buttons.length ? buttons : null,
```

module/documents/activity/mixin.mjs
```diff
@@ -905,7 +915,11 @@ export default Base => class extends PseudoDocumentMixin(Base) {
     target.disabled = true;
     try {
       if ( handler ) await handler.call(activity, event, target, message);
-      else await activity._onChatAction(event, target);
+      else if ( action === "consumeResource" ) await this.#consumeResource(event, target, message);
+      else if ( action === "placeTemplate" ) await this.#placeTemplate();
```
Fyorl, line 919: Shouldn't these be in `metadata.usage.actions`?

module/documents/item.mjs
```diff
@@ -533,6 +532,17 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
+  /* -------------------------------------------- */
+  /*  Data Initialization                         */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  clone(...args) {
+    const item = super.clone(...args);
+    if ( !item.parent ) item.prepareFinalAttributes();
+    return item;
+  }
```
Fyorl, line 544: This doesn't take into account the `save` option, which makes the return value a `Promise`. We probably also need to always call `prepareFinalAttributes`, regardless of whether the Item has a `parent`. Cloned Items are not placed into their parents' collections (unless `save` is `true`), so `prepareFinalAttributes` wouldn't be called by the parent.

```suggestion
  clone(data={}, options={}) {
    if ( options.save ) return super.clone(data, options);
    const item = super.clone(data, options);
    item.prepareFinalAttributes();
    return item;
  }
```

Fyorl (changes requested):

module/documents/activity/mixin.mjs
arbron: Can't because with the Mixin there isn't a concrete class to reference.

module/documents/item.mjs
arbron: If it doesn't have a parent then it calls `prepareFinalAttributes` itself at the end of `prepareData`.

Fyorl (approved):

PR #4020, opened 2024-08-22
[#1964] Migrate reach & versatile damage for non-weapons

arbron:
- Creates additional `DamageActivity` for items that aren't weapons but have versatile damage
- Migrates `range.value` to `range.reach` for melee waepons without the thrown property
- Adjust the API of `refreshCompendium` to call `migrateCompendium` first with proper error handling if the migration fails

### DIFF
--- a/module/data/activity/base-activity.mjs
+++ b/module/data/activity/base-activity.mjs
@@ -188,21 +188,22 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
 
   /**
    * Migrate data from the item to a newly created activity.
-   * @param {object} source  Item's candidate source data.
-   * @param {number} offset  Adjust the default ID using this number when creating multiple activities.
+   * @param {object} source              Item's candidate source data.
+   * @param {object} [options={}]
+   * @param {number} [options.offset=0]  Adjust the default ID using this number when creating multiple activities.
    */
-  static createInitialActivity(source, offset=0) {
+  static createInitialActivity(source, { offset=0, ...options }={}) {
     const activityData = this.transformTypeData(source, {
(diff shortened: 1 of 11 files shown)


module/data/activity/base-activity.mjs
```diff
@@ -212,9 +213,10 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
   /**
    * Fetch data from the item source and transform it into an activity's activation object.
    * @param {object} source  Item's candidate source data to transform.
+   * @param {object} options       Additional options passed to the creation process.
```
Fyorl, line 216: Check alignment on these

module/data/item/weapon.mjs
```diff
@@ -168,6 +169,23 @@ export default class WeaponData extends ItemDataModel.mixin(
...
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * Migrate the range value to the reach field for melee weapons without the thrown property.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static #migrateReach(source) {
+    if ( !source.properties || !source.range?.value || !source.type?.value ) return;
```
Fyorl, line 179:
```suggestion
    if ( !source.properties || !source.range?.value || !source.type?.value || source.range?.reach ) return;
```

Fyorl (approved):

krbz999:
> * Adjust the API of `refreshCompendium` to call `migrateCompendium` first with proper error handling if the migration fails

I take this to mean it will no longer be needed for a module author to call both, but simply loop over all their packs and calling `refreshCompendium`?

PR #4023, opened 2024-08-22
[#1964] Migrate uses, fix recovery on charges items w/o formula

arbron:
Performs a concrete migrate for `uses.value` because it relies on the resolved max uses formula. The involved some changes to how migrations are called to pass the fully prepared item through the process.

Also fixed a migration issue for items that have charges but not a recovery formula set getting a recovery period.

### DIFF
--- a/module/migration.mjs
+++ b/module/migration.mjs
@@ -816,6 +816,7 @@ function _migrateItemUses(item, itemData, updateData, flags) {
     foundry.utils.setProperty(updateData, "system.uses.spent", parseInt(max) - parseInt(value));
     flags.persistSourceMigration = true;
   }
+  updateData["flags.dnd5e.-=migratedUses"] = null;
 }
 
 /* -------------------------------------------- */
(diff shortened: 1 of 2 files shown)


module/migration.mjs
```diff
@@ -799,6 +802,24 @@ function _migrateEffectArmorClass(effect, updateData) {
...
+ * @param {object} updateData  Existing update to expand upon.
+ * @param {object} flags       Track the needs migration flag.
+ */
+function _migrateItemUses(item, itemData, updateData, flags) {
+  const value = foundry.utils.getProperty(itemData, "flags.dnd5e.migratedUses");
+  const max = foundry.utils.getProperty(item, "system.uses.max");
+  if ( (value !== undefined) && (max !== undefined) && Number.isNumeric(max) ) {
+    foundry.utils.setProperty(updateData, "system.uses.spent", parseInt(max) - parseInt(value));
+    flags.persistSourceMigration = true;
+  }
```
Fyorl, line 818: Do we need to remove `flags.dnd5e.migratedUses`?

```suggestion
  }
  updateData["flags.dnd5e.-=migratedUses"] = null;
```

Fyorl (approved):

PR #4025, opened 2024-08-22
[#4024] Add `CheckActivity` type for ability & tool checks

### LINKED ISSUE #4024: Activity Type: Check
The check activity will be able to make ability checks, optionally modified by skill or tool proficiency.

arbron:
Adds the check activity for rolling ability, skill, or tool checks. It must specify an ability and can optionally specify a skill or tool that can contribute proficiency, if the player has it.

Check Sheet V2

If added to a tool item, then it can infer the ability and the associated proficiency from the tool, along with any special bonus the tool provides.

Closes #4024


module/applications/activity/check-sheet.mjs
```diff
@@ -0,0 +1,70 @@
...
+  /** @inheritDoc */
+  async _prepareEffectContext(context) {
+    context = await super._prepareEffectContext(context);
+
+    context.abilityOptions = Object.entries(CONFIG.DND5E.abilities).map(([value, config]) => ({
+      value, label: config.label
+    }));
+    if ( this.item.type === "tool" ) {
+      const ability = CONFIG.DND5E.abilities[this.item.system.ability]?.label?.toLowerCase();
+      context.abilityOptions.unshift(
```
Fyorl, line 41: Every skill and tool has a default ability associated with it, so perhaps this functionality should not be limited to tool items only? We could move the 'Associated Skill or Tool' field above the 'Check Ability' field in the form, so it's a more natural flow from picking a skill to having the 'Default' option appear in the ability options?

module/data/activity/check-data.mjs
```diff
@@ -0,0 +1,107 @@
...
+    if ( this.item.type === "tool" ) {
+      if ( !this.check.associated ) this.check.associated = this.item.system.type.baseItem;
+      if ( !this.check.ability ) this.check.ability = this.item.system.ability
+        || this.actor?.system.tools?.[this.check.associated]?.ability
+        || Object.keys(CONFIG.DND5E.abilities)[0];
+    }
+
+    let ability;
+    if ( !this.check.dc.calculation ) this.check.dc.value = simplifyBonus(this.check.dc.formula, rollData);
+    else ability = this.ability;
```
Fyorl, line 101:
```suggestion
    if ( this.check.dc.calculation ) ability = this.ability;
    else this.check.dc.value = simplifyBonus(this.check.dc.formula, rollData);
```

module/applications/activity/save-sheet.mjs
```diff
@@ -49,7 +49,7 @@ export default class SaveSheet extends ActivitySheet {
     }));
     context.calculationOptions = [
       { value: "custom", label: game.i18n.localize("DND5E.SAVE.FIELDS.save.dc.CustomFormula") },
```
Fyorl, line 51: Could we use blank as the value for 'Custom Formula' here like we do with this new Check Activity? Having it be the string 'custom' causes some issues with newly-migrated Save Activities. We could fix the migration to appropriately migrate the `calculation` field to the `"custom"` string, but I think I'd prefer we be consistent here, and I like using blank as the de-facto 'default' option.

module/documents/activity/save.mjs
```diff
@@ -47,7 +47,7 @@ export default class SaveActivity extends ActivityMixin(SaveActivityData) {
       `,
       icon: '<i class="fa-solid fa-shield-heart" inert></i>',
       dataset: {
-        ability, dc,
+        ability: this.save.ability, dc,
```
Fyorl, line 50:
```suggestion
        dc,
        ability: this.save.ability,
```

module/enrichers.mjs
```diff
@@ -570,7 +570,7 @@ async function enrichItem(config, label, options) {
   if ( givenItem.startsWith(".") ) {
     try {
       foundItem = await fromUuid(givenItem, { relative: options.relativeTo });
-    } catch { return null; }
+    } catch(err) { return null; }
```
Fyorl, line 573: Is the linter complaining? If we don't use the `err` in the catch block then we don't need to declare it.

Fyorl (changes requested):

module/enrichers.mjs
arbron: Yeah, it complains about spacing because of how we space our catch blocks. Seems like the linter just isn't smart about it.

module/applications/activity/save-sheet.mjs
arbron: The reason I use `custom` rather than a blank value for saves is to allow for auto-detecting when a value hasn't been filled in yet and smartly filling it in based on whether the activity was created on a spell or not (spells default to "spellcasting" while other items default to "custom"). I suppose we could set some other arbitrary value to allow us to check for an unset initial value.

Fyorl: I was mainly hoping to easily fix this issue whereby the custom formula input is incorrectly disabled when you first create a save activity (on a non-spell I suppose).

https://github.com/user-attachments/assets/f4e46066-5bfb-4a6d-a752-c70df757b6a0

arbron: Huh, that was very specifically set up to prevent that issue. Looks like that is because the template is checking `source.save.dc.calculation` rather than the derived value.

module/applications/activity/check-sheet.mjs
Fyorl: I think the simplest solution is to allow for a blank option on 'check ability' and have it always start blank. A user will almost always have to change this drop-down anyway. The default of 'Strength' is only going to be correct 1/6th of the time. If they create a Check Activity and forget to change it, there's no much difference between it being incorrectly Strength, and being blank.

arbron: Okay, added blank default and set it so if no ability is found at usage time, it simply won't display a button.

Fyorl (approved):

PR #4030, opened 2024-08-23
[#1964] Move concentration into activity duration

arbron:
Adds `concentration` into the duration data on activities and removed it as a property on all items except spells. This allows concentration to be handled on a per-activity basis rather than for an item as a whole.

For spells, they get the derived `duration.concentration` value based on their properties and that will be inferred on the activity unless the duration is overridden.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1201,7 +1201,7 @@
       "label": "Duration",
       "concentration": {
         "label": "Concentration",
-        "hint": "User must maintain concentration while active."
+        "hint": "Creature must maintain concentration while active."
       },
       "override": {
         "label": "Override Duration",
--- a/lang/en.json
+++ b/lang/en.json
@@ -1199,6 +1199,10 @@
   "FIELDS": {
     "duration": {
       "label": "Duration",
+      "concentration": {
+        "label": "Concentration",
+        "hint": "User must maintain concentration while active."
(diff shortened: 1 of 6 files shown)


lang/en.json
```diff
@@ -1199,6 +1199,10 @@
   "FIELDS": {
     "duration": {
       "label": "Duration",
+      "concentration": {
+        "label": "Concentration",
+        "hint": "User must maintain concentration while active."
```
Fyorl, line 1204: I know you don't mean 'user' as in the 'user of the software' case, but more the 'user of the activity' case, but I think 'creature' is a stronger game term here.
```suggestion
        "hint": "Creature must maintain concentration while active."
```

Fyorl (approved):

module/data/activity/base-activity.mjs
```diff
@@ -342,7 +348,9 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
    */
   static transformDurationData(source, options) {
     if ( source.type === "spell" ) return {};
+    const concentration = !!source.sytem.properties?.findSplice(p => p === "concentration");
```
krbz999, line 351:
```suggestion
    const concentration = !!source.system.properties?.findSplice(p => p === "concentration");
```

PR #4033, opened 2024-08-23
[#1964] Adjust styling of usage, switch to <dialog> element

arbron:
Adjusts the styling of the activity usage dialog to match the V2 system styling more closely. This includes a redesigned range input which wasn't previously styled.

Usage Dialog V2

Adds a subtitle to the header of all V2 sheets configurable via `options.subtitle` or the `subtitle` getter. This will be hidden when the application is minimized.

Also changes the outer element of the usage application to be a `<dialog>` element and the content tag to be `<form>`.

Note: Resource consumption currently isn't styled awaiting a followup PR that makes more significant changes to those options.

### DIFF
--- a/less/v2/activities.less
+++ b/less/v2/activities.less
@@ -124,6 +124,7 @@
     .separator, .max {
       flex: 0;
       padding-inline: 3px;
+      font-weight: bold;
     }
     .separator { color: var(--dnd5e-color-gold); }
   }
--- a/less/v2/activities.less
+++ b/less/v2/activities.less
@@ -114,3 +114,17 @@
     border-radius: 4px;
   }
 }
+
+/* ----------------------------------------- */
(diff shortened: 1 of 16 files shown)


Fyorl (changes requested):
Styling looks good. Some higher-level things I noticed:

- These apps don't get the dark mode styling which should be already complete for them. I think it's because `dark/apps.less` specifies `&.sheet, &.compendium-browser`, but it can probably be `&.sheet, &.application` now.
- The button still has the default bright orange background colour on-hover. Not sure why, again, since I thought our new button styles were fairly general.
- Not opposed to the subtitles in general, but I think at least for these dialogs we could do-away with them, and just promote the current subtitle to the title.

image

Fyorl (approved):

PR #4034, opened 2024-08-23
[#1964] Communicate consumption costs in usage dialog

arbron:
Creates hints for each resource consumption target in the usage dialog displaying what will be consumed and how much, adjusting automatically based on the scaling.

Screenshot 2024-08-23 at 15 49 38

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -935,21 +935,24 @@
   "Type": {
     "ActivityUses": {
       "Label": "Activity Uses",
-      "Prompt": "Consume Activity Use?",
+      "PromptDecrease": "Consume Activity Use?",
+      "PromptIncrease": "Restore Activity Use?",
       "PromptHintDecrease": "Spend <strong>{cost}</strong> {use} of this activity.",
       "PromptHintIncrease": "Recover <strong>{cost}</strong> {use} of this activity.",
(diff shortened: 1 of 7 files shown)


less/v2/activities.less
```diff
@@ -127,4 +127,11 @@
     }
     .separator { color: var(--dnd5e-color-gold); }
   }
+  .hint code {
+    background: none;
+    border: none;
+    color: var(--dnd5e-color-crimson);
```
Fyorl, line 133: Check how this looks in dark mode.

lang/en.json
```diff
@@ -936,32 +936,56 @@
...
+      "PromptHintIncrease": "Increase quantity of {item} by <strong>{cost}</strong>.",
       "Warning": "of {name}"
     },
+    "SpellSlot": {
+      "one": "{level} slot",
+      "other": "{level} slots"
+    },
     "SpellSlots": {
       "Label": "Spell Slots",
       "Prompt": "Consume Spell Slot?",
```
Fyorl, line 981: Might be nice if the labels cost the same treatment as the hints, i.e. `"Restore Spell Slot?"`, etc.

Fyorl (approved):
Very nice

PR #4035, opened 2024-08-24
[#1964] Support activities on spells that don't consume slots

arbron:
Adds `consumption.spellSlot` that only appears when an activity is on a spell that controls whether a spell consumes a slot. If a spell wouldn't normally consume a slot then this setting does nothing. Unchecking this still allows upcasting the spell.

### DIFF
--- a/templates/activity/parts/activity-consumption.hbs
+++ b/templates/activity/parts/activity-consumption.hbs
@@ -8,6 +8,12 @@
                 <i class="fas fa-plus" inert></i>
             </button>
         </legend>
+
+        {{#if showConsumeSpellSlot}}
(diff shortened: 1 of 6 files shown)


templates/activity/parts/activity-consumption.hbs
```diff
@@ -46,6 +46,11 @@
         {{else}}
         <div class="empty">{{ localize "DND5E.None" }}</div>
         {{/each}}
+
+        {{#if showConsumeSpellSlot}}
```
Fyorl, line 50: Let's move this to above the consumption target parts. This will match what we do with things like damage parts and summon profiles.

Fyorl (approved):

PR #4041, opened 2024-08-26
[#1964] Update spell scroll creation to work with activities

arbron:
Modifies spell scroll creation to deal with activities rather than the old activation data. This involves copying over the activities, merging in data that was originally inferred from the spell, and applying save DC & attack bonus changes directly to the proper Save and Attack activities. Because the reference scroll data no longer contains the values we used to use for determining DC & attack bonus under the 2014 rules, this adds a config object to provide those values.

This also fixes a couple of bugs preventing activities from being used from destroyed items.

Note: While the API side has been updated to support custom DC and attack bonus values for the 2024 rules, the spell scroll UI hasn't been updated to provide those values yet.

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2609,7 +2609,7 @@ DND5E.spellScrollIds = {
  */
 
 /**
- * Spell scroll save DCs and attack bonus values based on level in the 2014 rules. If matching level isn't found,
+ * Spell scroll save DCs and attack bonus values based on spell level. If matching level isn't found,
  * then the nearest level lower than it will be selected.
  * @enum {SpellScrollValues}
  */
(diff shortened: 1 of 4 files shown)


module/config.mjs
```diff
@@ -2602,6 +2602,27 @@ DND5E.spellScrollIds = {
...
 /* -------------------------------------------- */
 
+/**
+ * @typedef {object} SpellScrollValues
+ * @property {number} bonus  Attack to hit bonus.
+ * @property {number} dc     Saving throw DC.
+ */
+
+/**
+ * Spell scroll save DCs and attack bonus values based on level in the 2014 rules. If matching level isn't found,
```
Fyorl, line 2612: Let's assume these values will be the same in the 2024 rules too unless we see otherwise.

```suggestion
 * Spell scroll save DCs and attack bonus values based on level. If matching level isn't found,
```

Fyorl (approved):

PR #4044, opened 2024-08-26
[#2978] Add spell preparation count to character sheet.

### LINKED ISSUE #2978: Add display of number of prepared spells back to spellbook
The old character sheet displayed the number of prepared spells in the header. This number is no longer shown in the new character sheet.
This was useful for wizards and clerics, who can know any number of spells, but only have access to a limited number of them at one time, dictated by level and ability modifier.
As a bonus feature, which would require significantly more work, it would be great if this count was separate for each class. That is, a multiclassed cleric/wizard would have two prepared skill counts. That means that each spell in the list would need to be qualified as wizard or cleric. This is easy for spells that only one class has access to (magic missile and cure wounds, for example), but would require the user to select a class for spells that both classes have access to, like detect magic, or wizard spells that certain cleric domains have access to.

Fyorl:
- Closes #2978

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -627,16 +627,30 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     // Translate the list of classes into spellcasting progression
     const progression = { slot: 0, pact: 0 };
     const types = {};
+    const preparations = {};
+    let classes = {};
 
     // Grab all classes with spellcasting
-    const classes = this.items.filter(cls => {
-      if ( cls.type !== "class" ) return false;
-      const type = cls.spellcasting.type;
-      if ( !type ) return false;
-      types[type] ??= 0;
-      types[type] += 1;
-      return true;
+    for ( const item of this.items ) {
+      if ( item.type === "class" ) {
+        const type = item.spellcasting.type;
+        if ( !type ) continue;
+        types[type] = (types[type] ?? 0) + 1;
+        classes[item.identifier] = item;
+      } else if ( item.type === "spell" ) {
+        const { preparation, sourceClass } = item.system;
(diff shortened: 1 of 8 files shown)


module/documents/actor/actor.mjs
```diff
@@ -627,16 +627,30 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
 
     // Grab all classes with spellcasting
-    const classes = this.items.filter(cls => {
-      if ( cls.type !== "class" ) return false;
-      const type = cls.spellcasting.type;
-      if ( !type ) return false;
-      types[type] ??= 0;
-      types[type] += 1;
-      return true;
+    for ( const item of this.items ) {
```
Fyorl, line 634: A little clumsier than it could be, but I'm trying hard to avoid iterating over the entire item list too many times.

arbron: We could always do that in `prepareFinalData` in `SpellData`. So it would do something like:

```javascript
const config = CONFIG.DND5E.spellPreparationModes[this.preparation.mode];
if ( this.parent.isOwned && this.sourceClass && config?.prepares && ((this.preparation.mode === "always") || prepared ) {
  const sourceClass = this.parent.actor.spellcastingClasses[this.sourceClass];
  if ( sourceClass ) sourceClass.system.spellcasting.preparation.value += 1;
}
```

We are already calling this for spells so there is no additional loops performed, and the system ensures that `prepareFinalData` is always called once.

arbron (approved):
Looks good, just an optional thought on how we calculate the value.

module/documents/actor/actor.mjs
krbz999: `actor.itemTypes` is cached now, so that could help.

PR #4045, opened 2024-08-26
[#4043] Refresh Create Scroll UI, allow custom spell values

### LINKED ISSUE #4043: Allow creating spell scrolls using creator's DC & attack bonus, not fixed values
In the 2024 rules spell scrolls crafted by players use their spell save DC and attack bonus, not the fixed values specified in spell scrolls found in the Magic Items list. It should be possible when creating a scroll from a spell to either use the fixed DC & bonus or the values from the crafter.

arbron:
Reworks the Create Scroll using ApplicationV2 in a way that allows for dynamic updates of the form. It now displays fields for spell save DC and attack bonus with placeholders updating dynamically when the spell level is changed.

These fields can also be set manually, overriding the default DC of spells. If using the modern rules version and create a scroll directly from a spell in an actor's inventory, then these values will be initial set based on that actor's spellcasting stats.

Closes #4043

### DIFF
--- a/module/applications/api/dialog.mjs
+++ b/module/applications/api/dialog.mjs
@@ -1,0 +1,93 @@
+import Application5e from "./application.mjs";
+
+/**
+ * Application for creating dnd5e dialogs.
+ */
+export default class Dialog5e extends Application5e {
+  /** @override */
+  static DEFAULT_OPTIONS = {
+    tag: "dialog",
+    window: {
+      contentTag: "form",
+      minimizable: false
+    }
+  };
+
+  /** @override */
+  static PARTS = {
+    content: {
+      template: ""
+    },
+    footer: {
+      template: "templates/generic/form-footer.hbs"
+    }
+  };
+
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Form element within the dialog.
+   * @type {HTMLFormElement|void}
+   */
+  get form() {
+    return this.options.tag === "form" ? this.element : this.element.querySelector("form");
+  }
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preparePartContext(partId, context, options) {
+    context = foundry.utils.deepClone(await super._preparePartContext(partId, context, options));
+    if ( partId === "content" ) return this._prepareContentContext(context, options);
+    if ( partId === "footer" ) return this._prepareFooterContext(context, options);
+    return context;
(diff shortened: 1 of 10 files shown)


module/applications/api/dialog.mjs
```diff
@@ -0,0 +1,93 @@
...
+    return this.options.tag === "form" ? this.element : this.element.querySelector("form");
+  }
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preparePartContext(partId, context, options) {
+    context = foundry.utils.deepClone(await super._preparePartContext(partId, context, options));
```
krbz999, line 44: Is there a particular need for `deepClone`?

arbron: Because core's `HandlebarsApplicationMixin` passes the same object into `_preparePartContext` for each call, it is possible for data from one part to pollute another part, so I've been adding this to avoid that problem.

module/data/item/fields/spellcasting-field.mjs
```diff
@@ -27,4 +28,37 @@ export default class SpellcastingField extends SchemaField {
...
+   * @param {object} rollData  Roll data used for formula replacements.
+   */
+  static prepareData(rollData) {
+    this.spellcasting.preparation.max = simplifyBonus(this.spellcasting.preparation.formula, rollData);
+
+    // Temp method for determining spellcasting type until this data is available directly using advancement
+    if ( CONFIG.DND5E.spellcastingTypes[this.spellcasting.progression] ) {
+      this.spellcasting.type = this.spellcasting.progression;
+    } else this.spellcasting.type = Object.entries(CONFIG.DND5E.spellcastingTypes).find(([type, data]) => {
+      return !!data.progression?.[this.spellcasting.progression];
```
Fyorl, line 48: `find` doesn't require a strictly boolean return. I'm surprised the linter does not complain actually, I thought I added that rule.

```suggestion
    } else this.spellcasting.type = Object.entries(CONFIG.DND5E.spellcastingTypes).find(([, { progression }]) => {
      return progression?.[this.spellcasting.progression];
```

Fyorl (approved):

module/applications/api/dialog.mjs
krbz999: Seems rather expensive to deepClone the entire config object (in `Application5e#_prepareContext`) each time but it's probably negligible.

arbron: We could probably get away with creating a shallow clone, since the top-level keys are the only place I was really getting a problem.

PR #4046, opened 2024-08-26
[#3786] Add system for consolidated spell lists

### LINKED ISSUE #3786: Automatically consolidate spell lists into master lists
Currently the system has Spell List journal page types that the system and modules can provide, but when multiple modules are enabled these spell lists stay separate. For example, the system provides a Wizard spell list with all of the SRD spells and Tasha's contains another Wizard spell list with all of the spells provided by that book. In order to make use of the spell list as a Wizard for the Compendium Browser or in the planned Spellcasting advancement these two lists will need to be combined.

Modules should be able to declare the UUIDs of spell list journal pages that will be consolidated, and then the system should iterate over all of these lists and merge them into a single master list per-class (based on the type & identifier already specified in the Spell List journal page) that can then be referenced by other systems.

arbron:
Adds a new registry entry for tracking spell lists. This contains `SpellList` objects grouped by type and identifier, with all spell list journal entry pages contributing spells to the shared list. Each `SpellList` object contains the means to access the UUIDs, indexes, and fully-loaded spell documents for all listed spells.

A new module data registration method has been added to allow for modules to declare spell lists in their package manifets using the `dnd5e.spellLists` flag.

Closes #3786

### Todo
- [ ] Add filters to `CompendiumBrowser` based on these spell lists

### DIFF
--- a/module/registry.mjs
+++ b/module/registry.mjs
@@ -247,6 +247,131 @@ class MessageRegistry {
 }
 
 /* -------------------------------------------- */
+/*  Spell Lists                                 */
+/* -------------------------------------------- */
+
+class SpellListRegistry {
+  /**
+   * Registration of spell lists grouped by type and identifier.
+   * @type {Map<string, SpellList>}
+   */
+  static #lists = new Map();
+
+  /* -------------------------------------------- */
+
+  /**
+   * Retrieve a specific spell list from the registry.
+   * @param {string} type        Type of list as defined in `CONFIG.DND5E.spellListTypes`.
+   * @param {string} identifier  Identifier of the specific spell list.
+   * @returns {SpellList|null}
+   */
+  static list(type, identifier) {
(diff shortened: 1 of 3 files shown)


module/registry.mjs
```diff
@@ -122,6 +122,131 @@ class MessageRegistry {
...
+   * @param {string} uuid  UUID of a spell list journal entry page.
+   */
+  static async register(uuid) {
+    if ( !game.ready ) {
+      Hooks.once("ready", () => this.register(uuid));
+      return;
+    }
+
+    const page = await fromUuid(uuid);
+    if ( !page ) throw new Error(`Journal entry page could not be found "${uuid}" to register as spell list.`);
```
krbz999, line 161:
```suggestion
    if ( !page ) throw new Error(`Journal entry page "${uuid}" could not be found to register as spell list.`);
```

Fyorl (approved):

PR #4047, opened 2024-08-27
[#3836, #3853] Add Has Spellcasting & Class filters to subclasses

### LINKED ISSUE #3836: Add 'Grants spellcasting' to compendium browser when searching through subclasses
This would be an option to match the 'has spellcasting' filter for classes.

For subclasses such as Eldritch Knight, Arcane Trickster, etc.

### LINKED ISSUE #3853: Allow for filtering subclasses by class or class identifier in the Compendium Browser
When searching through subclasses, there is no way to tell which subclass belongs to which class.

There is of course some consideration to make to find the appropriate class to retrieve a correct label (if there are multiple, I would just use the first entry found).

The fallback would simply be the class identifier itself.

arbron:
Adds a "Has Spellcasting" filter to subclasses in the compendium browser indicating the subclass provides spellcasting. Also adds the ability to filter by the class to which the subclass belongs.

Subclass Filters

To support the second filter, this introduces the `ItemRegistry` to the system registry that scans compendiums and loads all classes registered in different compendium packs. These are grouped by identifier and different versions of a class in different packs are consolidated with their UUIDs listed in the `sources` list. This only supports classes at the moment but could easily be expanded to incude subclasses, backgrounds, or species, which maybe useful to work on spell lists.

Closes #3836
Closes #3853

### DIFF
--- a/module/registry.mjs
+++ b/module/registry.mjs
@@ -103,7 +103,17 @@ class ItemRegistry {
    * Has initial loading been completed?
    * @type {number}
    */
-  #status = 0;
+  #status = ItemRegistry.#STATUS_STATES.NONE;
+
+  /**
+   * Possible preparation states for the item registry.
+   * @enum {number}
+   */
+  static #STATUS_STATES = Object.freeze({
+    NONE: 0,
+    LOADING: 1,
+    READY: 2
+  });
 
   /* -------------------------------------------- */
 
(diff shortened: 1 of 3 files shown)


module/registry.mjs
```diff
@@ -61,6 +62,119 @@ class EnchantmentRegisty {
...
+   */
+  #itemType;
+
+  /* -------------------------------------------- */
+
+  /**
+   * Has initial loading been completed?
+   * @type {number}
+   */
+  #status = 0;
```
Fyorl, line 106: Let's use an enum

module/registry.mjs
```diff
@@ -61,6 +62,119 @@ class EnchantmentRegisty {
...
+    if ( this.#status > 0 ) return;
+    if ( !game.ready ) {
+      Hooks.once("ready", () => this.initialize());
+      return;
+    }
+    this.#status = 1;
+
+    const indexes = await CompendiumBrowser.fetch(Item, {
+      types: new Set([this.#itemType]),
+      indexFields: new Set(["system.identifier"]),
```
Fyorl, line 161:
```suggestion
      indexFields: new Set(["system.identifier"]),
      index: true,
```

Fyorl (approved):

module/registry.mjs
arbron: Index defaults to `true`.

PR #4049, opened 2024-08-27
[#1964] Extend [[/item]] enricher to trigger activities

arbron:
Allows the item enricher to include an activity ID or name that will be triggered if found on the item:

```
[[/item Bite activity=Paralyze]]
[[/item amUUCouL69OK1GZU activity=G8ng63Tjqy5W52OP]]
```

### DIFF
--- a/module/documents/macro.mjs
+++ b/module/documents/macro.mjs
@@ -82,10 +82,11 @@ function getMacroTarget(name, documentType) {
 /**
  * Trigger an item to be used when a macro is clicked.
  * @param {string} itemName                Name of the item on the selected actor to trigger.
- * @param {string} [activityName]          Name of a specific activity on the item to trigger.
+ * @param {object} [options={}]
+ * @param {string} [options.activityName]  Name of a specific activity on the item to trigger.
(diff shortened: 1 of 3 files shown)


module/documents/macro.mjs
```diff
@@ -80,12 +80,15 @@ function getMacroTarget(name, documentType) {
...
- * Trigger an item to roll when a macro is clicked.
+ * Trigger an item to be used when a macro is clicked.
  * @param {string} itemName                Name of the item on the selected actor to trigger.
- * @returns {Promise<ChatMessage|object>}  Roll result.
+ * @param {string} [activityName]          Name of a specific activity on the item to trigger.
+ * @returns {Promise<ChatMessage|object>}  Usage result.
  */
-export function rollItem(itemName) {
-  return getMacroTarget(itemName, "Item")?.use();
+export function rollItem(itemName, activityName) {
```
Fyorl, line 88: It's conceivable that we'll add more options to the item enricher in the future so maybe we should move this parameter into an `options` object pre-emptively, since it is public API.

Fyorl (approved):

Issue #1012, opened 2021-02-25
[Feature Request] Track which classes provide access to each spell and display which character class provides each spell on the spellbook tab


aaclayton:
Originally in GitLab by @crash1115

Pain Point: A character with two spellcasting classes that use the same preparation mode (Always Prepared/Spells Known vs Prepared) has no way to easily distinguish which spells come from which class.

Current Workaround: Set all the spells from one class as Prepared, and all the spells from the other as Always Prepared, and let the color-coding some sheet modules provide do the work. This stops being a viable solution if you have domain spells which are always prepared, or have more than 2 casting classes.

Potential Solution: Add a "Source Class" field or something similar on the Spell item sheet, and display that field... Somewhere. Preference would be as a column in the Spellbook tab, but that's pretty crowded already.

Zanderaf:
An extra column could be nice, could use it to track what advancement a spell came from, such as an advancement called "Mage Hand Ledgermian" that adds the mage hand cantrip to a characters spellbook would have the name of the advancement in that extra column, or just have a field in the Grant Items advancement type that automatically sets it.

_**Main example I could think of would be something like D&D Beyond.**_

![Capture](https://github.com/foundryvtt/dnd5e/assets/65571170/eb0466d8-460f-43c0-9fef-e8d237bc82ad)

krbz999:
Spell lists are a different issue, but all that is lacking from this thread is the display of the origin class on the actor sheet, not just the spell sheet.

Fyorl:
Putting the source class in the spell's subtitle seems like a reasonable solution. It might get a bit crowded with the limited uses in there as well though.

PR #4052, opened 2024-08-27
[#1012, #1229] Display classes providing spells on spell sheet

### LINKED ISSUE #1229: Display classes providing a spell on the spell sheet
Originally in GitLab by @Shareenx

Spells are about the only place where such data does not exist. Monsters have types, there are Prereq text boxes, all sorts of implementations for this type of sorting, but the closest is weapon properties.
The simplest, imo, would be to have checkboxes (like weapon properties) to assign a spell to classes.

This way, if a shared compendium of spells (items) is shared, a Journal Entry listing which spell goes to which class does not have to come along. All the information is available in the spell description or details, and can, if needed, be parsed via the json.

Right now, for a custom spell or a customized spell or a shared spell, it is unclear which lists it is intended to be in without finding the source material or author.

arbron:
Modifies the `SpellListRegistry` to keep track of what spell lists provide a spell in addition to spell lists by type. This allows spells to retrieve a list of spell lists they appear on and turn it into a list of classes to be displayed on the spell sheet.

Class Lists

Part of this work involves fetching the name of the class from the item registry rather than using the name of the spell list itself. In order to get around the loading time involved a new API has been added to the registry that allows defering a callback until all registries have fully loaded.

Closes #1229

### DIFF
--- a/module/data/item/spell.mjs
+++ b/module/data/item/spell.mjs
@@ -186,6 +186,16 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
       return obj;
     }, { all: [], vsm: [], tags: [] });
     labels.components.vsm = game.i18n.getListFormatter({ style: "narrow" }).format(labels.components.vsm);
+
+    dnd5e.registry.whenReady(() => {
+      labels.classes = game.i18n.getListFormatter({ style: "narrow" }).format(
+        Array.from(dnd5e.registry.spellLists
+          .forSpell(this.parent._stats.compendiumSource ?? this.parent.uuid))
+          .filter(list => list.metadata.type === "class")
+          .map(list => list.name)
+          .sort((lhs, rhs) => lhs.localeCompare(rhs))
+      );
+    });
   }
 
   /* -------------------------------------------- */
@@ -240,7 +250,8 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
     context.subtitles = [
       { label: context.labels.level },
       { label: context.labels.school },
-      { label: context.itemStatus }
+      { label: context.itemStatus },
+      { label: context.labels.classes, classes: "full-width" }
     ];
     context.properties.active = this.parent.labels?.components?.tags;
     context.parts = ["dnd5e.details-spell", "dnd5e.field-uses"];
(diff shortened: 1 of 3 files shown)


module/data/item/spell.mjs
```diff
@@ -186,6 +186,16 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
       return obj;
     }, { all: [], vsm: [], tags: [] });
     labels.components.vsm = game.i18n.getListFormatter({ style: "narrow" }).format(labels.components.vsm);
+
+    dnd5e.registry.whenReady(() => {
```
Fyorl, line 190: It might become a bit hard to reason about Item state if we start scheduling callbacks inside Item preparation. Since we only need this information in the sheet, we should just do this during sheet rendering.

module/registry.mjs
```diff
@@ -311,6 +378,29 @@ export class SpellList {
 
   /* -------------------------------------------- */
 
+  /**
+   * Information on the spell list.
+   * @type {{ identifier: string, type: string }}
```
Fyorl, line 383:
```suggestion
   * @type {{ identifier: string, type: string, name: string }}
```

module/registry.mjs
```diff
@@ -421,11 +520,64 @@ class SummonRegistry {
...
+  }
+}();
+
+/* -------------------------------------------- */
+
+/**
+ * Execute a callback when all registries have finished loading.
+ * @param {Function} callback
+ */
+function whenReady(callback) {
```
Fyorl, line 569: I think providing a `Promise` API would be a bit more ergonomic. It offers a pretty similar interface, but you can also optionally `await` it, e.g.:

```js
// This:
dnd5e.registry.whenReady(() => {
  // ...
});

// Becomes:
dnd5e.registry.whenReady().then(() => {
  // ...
});

// Or:
await dnd5e.registry.whenReady();
// ...
```

```js
const RegistryStatus = new class extends Map {
  constructor() {
    super();
    this.#ready = new Promise(resolve => this.#resolve = resolve);
  }

  #ready;
  #resolve;

  get ready() { return this.#ready };

  set(key, value) {
    super.set(key, value);
    if ( Array.from(this.values()).every(s => s) ) this.#resolve();
  }
}();

export default {
  // [snip]
  whenReady: RegistryStatus.ready
}
```

Fyorl (changes requested):

module/registry.mjs
```diff
@@ -521,63 +521,64 @@ class SummonRegistry {
...
 /* -------------------------------------------- */
 
 /**
  * Track the ready status of various registries.
  * @type {Map<string, boolean>}
  */
 const RegistryStatus = new class extends Map {
+  constructor(iterable) {
+    super(iterable);
+    const { promise, resolve } = Promise.withResolvers();
```
arbron, line 534: Love `withResolvers`! (and I tested it with the app, so we should be good to use it despite it being fairly new)

Fyorl (approved):

PR #4054, opened 2024-08-27
[#1964] Re-implement effect application for activities

arbron:
Re-implements the effect application UI for activities using a custom element like damage application. This allows for applying effects to targeted or selected tokens. The top part displays the targets (with checkboxes to control whether they are applied when using the "Targeted" mode) and below that is a list of effects to apply.

Screenshot 2024-08-27 at 14 38 44

Refactor some shared code from `DamageApplicationElement` into `TargetedApplicationMixin` to handle targeting tokens.

Also makes a change to the damage application UI to display token names rather than actor names.


etiquettestartshere:
When trying to apply an effect to a target that already has a copy of that effect, this error is thrown and duration does not seem to be correctly refreshed. Not sure what is causing it as the code in question doesn't really look changed to me (apologies if this is a misunderstanding on my part, just thought i should say something)
image
https://github.com/foundryvtt/dnd5e/pull/4054/files#diff-10fdd5d211d5ec059c5724df3e0c9d35e0b9242b3b72c97500d4041509119a86R179

module/applications/components/effect-application.mjs
```diff
@@ -0,0 +1,240 @@
...
+   */
+  effectsList;
+
+  /* -------------------------------------------- */
+
+  /**
+   * Currently registered hook for monitoring for changes to selected tokens.
+   * @type {number|null}
+   */
+  selectedTokensHook = null;
```
Fyorl, line 41:
```suggestion
```

module/applications/components/targeted-application-mixin.mjs
```diff
@@ -0,0 +1,144 @@
...
+      switch ( this.targetingMode ) {
+        case "targeted":
+          targetedTokens = (this.chatMessage?.getFlag("dnd5e", "targets") ?? [])
+            .map(t => ({ uuid: t.uuid, name: t.name }));
+          break;
+        case "selected":
+          targetedTokens = canvas.tokens?.controlled?.map(t => ({ uuid: t.actor.uuid, name: t.name })) ?? [];
+          break;
+      }
+      targetedTokens = Array.from(new Set(targetedTokens));
```
Fyorl, line 109: Putting `targetedTokens` in a `Set` won't remove any duplicates, if that was the intention, because `targetedTokens` are all newly-created, plain objects.

module/applications/components/targeted-application-mixin.mjs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /* -------------------------------------------- */
+
+    /**
+     * Create a list entry for a single target.
+     * @param {object} data
+     * @param {string} data.uuid  UUID of the targeted actor.
+     * @param {string} data.name  Name of the targeted token.
+     * @returns {HTMLLIElement|void}
```
Fyorl, line 127: I think this must be implemented by any subclasses. If so, we should mark it `@abstract`.

```suggestion
     * @returns {HTMLLIElement|void}
     * @abstract
```

module/documents/chat-message.mjs
```diff
@@ -555,6 +551,26 @@ export default class ChatMessage5e extends ChatMessage {
...
+  /* -------------------------------------------- */
+
+  /**
+   * Display the effects tray with effects the user can apply.
+   * @param {HTMLLiElement} html  The chat card.
+   * @protected
+   */
+  _enrichUsageEffects(html) {
+    const item = this.getAssociatedItem();
+    const effects = this.getFlag("dnd5e", "use.effects")
```
Fyorl, line 563: We should also make sure that Items that are posted directly to chat (i.e. when they have no activities) also still show applicable effects.

Fyorl (changes requested):

module/documents/chat-message.mjs
arbron: Is that something we want to handle? Because with activities you explicitly define what effects to include, but for posting to chat it will just put all of them?

module/applications/components/targeted-application-mixin.mjs
arbron: Aw yes, left over from when `targetedTokens` just included UUIDs.

module/documents/chat-message.mjs
Fyorl: This would only be for items that don't have any activities anyway, so their effects can't be associated with anything but the item.

module/documents/chat-message.mjs
```diff
@@ -555,6 +551,30 @@ export default class ChatMessage5e extends ChatMessage {
...
+
+  /**
+   * Display the effects tray with effects the user can apply.
+   * @param {HTMLLiElement} html  The chat card.
+   * @protected
+   */
+  _enrichUsageEffects(html) {
+    const item = this.getAssociatedItem();
+    let effects;
+    if ( this.getFlag("dnd5e", "type") === "usage" ) {
```
etiquettestartshere, line 564: Should this be `"type"` or should it be `"messageType"`? `"messageType"` is used elsewhere

Fyorl (approved):

PR #4058, opened 2024-08-28
Add activity choice dialog

Fyorl:
image


module/applications/components/inventory.mjs
```diff
@@ -370,18 +370,19 @@ export default class InventoryElement extends HTMLElement {
...
-   * @param {string} action   Action being triggered.
+   * @param {PointerEvent} event  The triggering event.
+   * @param {Element} target      Button or context menu entry that triggered this action.
+   * @param {string} action       Action being triggered.
    * @returns {Promise}
    * @protected
    */
-  async _onAction(target, action) {
-    const event = new CustomEvent("inventory", {
+  async _onAction(event, target, action) {
```
arbron, line 379: This broke the context menus, which are calling `_onAction` but don't have an event to pass on. Perhaps add it into an option object at the end of the list?

arbron (changes requested):

module/applications/components/inventory.mjs
Fyorl: Oops, good catch. I suppose we *should* have a the event that was fired when the user clicked the context menu option, but it seems core does not provide it in the callback.

arbron (approved):
Looks good, though they should be sorted in the dialog to match the order on the item.

PR #4062, opened 2024-08-28
[#1964] Automatically create activities on new Weapons & Tools

arbron:
Automatically creates an `AttackActivity` on newly created weapons and a `CheckActivity` on newly created tools.

### DIFF
--- a/module/data/item/tool.mjs
+++ b/module/data/item/tool.mjs
@@ -192,7 +192,7 @@ export default class ToolData extends ItemDataModel.mixin(
 
   /** @inheritDoc */
   _preCreate(data, options, user) {
-    super._preCreate(data, options, user);
+    if ( super._preCreate(data, options, user) === false ) return false;
     const activityData = new CONFIG.DND5E.activityTypes.check.documentClass({}, { parent: this.parent }).toObject();
     this.parent.updateSource({ [`system.activities.${activityData._id}`]: activityData });
   }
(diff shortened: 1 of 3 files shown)


module/data/item/tool.mjs
```diff
@@ -185,4 +185,15 @@ export default class ToolData extends ItemDataModel.mixin(
...
     return Math.max(baseItemProf?.value ?? 0, categoryProf?.value ?? 0);
   }
+
+  /* -------------------------------------------- */
+  /*  Socket Event Handlers                       */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _preCreate(data, options, user) {
+    super._preCreate(data, options, user);
```
Fyorl, line 195: I think we need to check for a `false` return, and make sure we propagate it.

```suggestion
    if ( super._preCreate(data, options, user) === false ) return false;
```

module/data/item/weapon.mjs
```diff
@@ -457,4 +457,15 @@ export default class WeaponData extends ItemDataModel.mixin(
...
     return Number(isProficient);
   }
+
+  /* -------------------------------------------- */
+  /*  Socket Event Handlers                       */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _preCreate(data, options, user) {
+    super._preCreate(data, options, user);
```
Fyorl, line 467:
```suggestion
    if ( super._preCreate(data, options, user) === false ) return false;
```

Fyorl (approved):

PR #4065, opened 2024-08-28
[#1011] Adjust emanation template size when placing

### LINKED ISSUE #1011: Feature Request: emanation type measured template
Originally in GitLab by @marcoseither

There are quite some spells with a range "all creatures within x feet of you. The current measured templates don't really support this definition. Using a sphere is possible but it originates from the center of a square, so a "all creatures within 10 feet of you" must be translated into a sphere 12.5 to create the correct template.
A separate type "emanation" for those spells which automatically creates the correct template would be nice.
In addition, an additional flag "exclude self" in the spell details would be great to differenciate, if the spell includes or excludes the caster from it's effect.

arbron:
Dynamically updates the template size for the emanation area of effect to take the hovered token's size into account when placing the template.

https://github.com/user-attachments/assets/8ed3c78b-ed01-481c-98b0-8e48fb3a9966

Also adjust the code for area of effect dimensions to handle one without a sizes object and merged the two implementations.

Closes #1011

### DIFF
--- a/module/canvas/ability-template.mjs
+++ b/module/canvas/ability-template.mjs
@@ -242,13 +242,13 @@ export default class AbilityTemplate extends MeasuredTemplate {
     const baseDistance = this.document.flags.dnd5e?.dimensions?.size;
     if ( this.document.flags.dnd5e?.dimensions?.adjustedSize && baseDistance ) {
       const rectangle = new PIXI.Rectangle(center.x, center.y, 1, 1);
-      const hoveredToken = canvas.tokens.placeables
-        .find(t => t.visible && !t.document.isSecret && t._overlapsSelection(rectangle));
+      const hoveredToken = canvas.tokens.quadtree.getObjects(rectangle, {
+        collisionTest: ({ t }) => t.visible && !t.document.isSecret }).first();
       if ( hoveredToken && (hoveredToken !== this.#hoveredToken) ) {
         this.#hoveredToken = hoveredToken;
         this.#hoveredToken._onHoverIn(event);
         const size = Math.max(hoveredToken.document.width, hoveredToken.document.height);
-        updates.distance = baseDistance + (size * canvas.grid.distance * .5);
+        updates.distance = baseDistance + (size * canvas.grid.distance / 2);
       } else if ( !hoveredToken && this.#hoveredToken ) {
         this.#hoveredToken._onHoverOut(event);
         this.#hoveredToken = null;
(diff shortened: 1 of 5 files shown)


module/canvas/ability-template.mjs
```diff
@@ -223,7 +236,27 @@ export default class AbilityTemplate extends MeasuredTemplate {
...
     const center = event.data.getLocalPosition(this.layer);
-    this.document.updateSource(canvas.templates.getSnappedPoint(center));
+    const updates = canvas.templates.getSnappedPoint(center);
+
+    // Adjust template size to take hovered token into account if `adjustedSize` is set
+    const baseDistance = this.document.flags.dnd5e?.dimensions?.size;
+    if ( this.document.flags.dnd5e?.dimensions?.adjustedSize && baseDistance ) {
+      const rectangle = new PIXI.Rectangle(center.x, center.y, 1, 1);
+      const hoveredToken = canvas.tokens.placeables
+        .find(t => t.visible && !t.document.isSecret && t._overlapsSelection(rectangle));
```
Fyorl, line 246: We should use the quad tree rather than iterating over all tokens:

```suggestion
      const hoveredToken = canvas.tokens.quadtree.getObjects(rectangle, { collisionTest: t => t.visible && !t.document.isSecret }).first();
```

module/canvas/ability-template.mjs
```diff
@@ -223,7 +236,27 @@ export default class AbilityTemplate extends MeasuredTemplate {
...
+      if ( hoveredToken && (hoveredToken !== this.#hoveredToken) ) {
+        this.#hoveredToken = hoveredToken;
+        this.#hoveredToken._onHoverIn(event);
+        const size = Math.max(hoveredToken.document.width, hoveredToken.document.height);
+        updates.distance = baseDistance + (size * canvas.grid.distance * .5);
```
Fyorl, line 251:
```suggestion
        updates.distance = baseDistance + (size * canvas.grid.distance / 2);
```

module/data/shared/target-field.mjs
```diff
@@ -90,4 +92,27 @@ export default class TargetField extends SchemaField {
       labels.target = parts.filterJoin(" ");
     }
   }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Create the template dimensions labels for a template type.
+   * @param {string} type  Area of effect type.
+   * @returns {{ size: string, width: [string], height: [string] }}
```
Fyorl, line 101:
```suggestion
   * @returns {{ size: string, [width]: string, [height]: string }}
```

Fyorl (changes requested):

Fyorl (approved):

PR #4067, opened 2024-08-28
[#1012] Select source class for spells when added to sheet

### LINKED ISSUE #1012: [Feature Request] Track which classes provide access to each spell and display which character class provides each spell on the spellbook tab

arbron:
Using the new consolidated spell lists this attempts to pick the proper `sourceClass` for spells when they are added to a character sheet. It uses the following logic:

1. Grab spellcasting classes, if none exists then don't set source
2. If only one spellcasting class exists, set it as the source
3. If preparation mode is pact and a class exists that offers pact casting, then set that class as the source
4. Create intersection of classes on actor with classes that offer the spell, set source as first intersecting class or first class on actor if no intersections are found

Closes #1012

### DIFF
--- a/module/data/item/spell.mjs
+++ b/module/data/item/spell.mjs
@@ -342,23 +342,29 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
   /** @inheritDoc */
   _preCreate(data, options, user) {
     if ( super._preCreate(data, options, user) === false ) return false;
-    if ( !this.parent.isEmbedded || ["atwill", "innate"].includes(this.preparation.mode) ) return;
+    if ( !this.parent.isEmbedded || ["atwill", "innate"].includes(this.preparation.mode) || this.sourceClass ) return;
     const classes = new Set(Object.keys(this.parent.actor.spellcastingClasses));
     if ( !classes.size ) return;
 
-    // If only a single spellcasting class is present, use that
-    if ( classes.size === 1 ) {
-      this.parent.updateSource({ "system.sourceClass": classes.first() });
-      return;
-    }
+    // Set the source class, and ensure the preparation mode is pact if adding a prepared spell to a pact class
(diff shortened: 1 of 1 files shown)


module/data/item/spell.mjs
```diff
@@ -335,6 +335,46 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
...
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _preCreate(data, options, user) {
+    if ( super._preCreate(data, options, user) === false ) return false;
+    if ( !this.parent.isEmbedded || ["atwill", "innate"].includes(this.preparation.mode) || this.sourceClass ) return;
+    const classes = new Set(Object.keys(this.parent.actor.spellcastingClasses));
+    if ( !classes.size ) return;
+
+    // Set the source class, and ensure the preparation mode is pact if adding a prepared spell to a pact class
```
krbz999, line 349: This can be generalized to any "alternative" spell slot type.

Fyorl (approved):

module/data/item/spell.mjs
arbron: Give my new PR a look to make sure that will work with alt types: https://github.com/foundryvtt/dnd5e/pull/4068

PR #4068, opened 2024-08-29
[#1012] Handle alt-spellcasting types properly

arbron:
Handle alternate spellcasting types, not just pact casting, when selecting a class for a spell added to the sheet.

### DIFF
--- a/module/data/item/spell.mjs
+++ b/module/data/item/spell.mjs
@@ -346,18 +346,19 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
     const classes = new Set(Object.keys(this.parent.actor.spellcastingClasses));
     if ( !classes.size ) return;
 
-    // Set the source class, and ensure the preparation mode is pact if adding a prepared spell to a pact class
+    // Set the source class, and ensure the preparation mode matches if adding a prepared spell to an alt class
     const setClass = cls => {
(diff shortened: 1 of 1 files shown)


krbz999:
Seems to work. Used the Apothecary class from SebCrowe as an example, the `sourceClass` was set correctly when dropping on the Apothecary Magic section, even while multiclassed as druid/apothecary/warlock.

module/data/item/spell.mjs
```diff
@@ -349,15 +349,16 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
...
+      if ( (type !== "leveled") && (this.preparation.mode === "prepared") && (this.level > 0)
+        && (type in CONFIG.DND5E.spellPreparationModes) ) update["system.preparation.mode"] = type;
       this.parent.updateSource(update);
     };
 
     // If preparation mode is "pact" and pact class exists, set as that class
-    if ( this.preparation.mode === "pact" ) {
-      const pactClasses = classes.filter(i => this.parent.actor.classes[i].spellcasting.type === "pact");
-      if ( pactClasses.size === 1 ) setClass(pactClasses.first());
+    if ( this.preparation.mode in CONFIG.DND5E.spellcastingTypes ) {
```
krbz999, line 359: Comment here should be amended.

PR #4072, opened 2024-08-29
[#1407] Add Subclass advancement & selection interface on sheet

### LINKED ISSUE #1017: Add subclassing information to Class item
Add an item to the Class template which indicates what level the class takes a subclass (3rd level for most classes, 1st for Clerics, 2nd for Wizards) and the custom name for that subclass (Arcane Tradition, Bardic College, Ranger Archetype).

```json
{
  "subclass": {
    "name": "Ranger Archetype",
    "level": 3,
    "selected": "gloom-stalker"
  }
}
```

The name would allow for better customization of the UI per class.

The level data would allow for a theoretical future level-up interface to prompt the player to select a subclass when they reach the level that requires it.

This would replace the current `subclass` field, the value of which would be moved to `subclass.selected`.

### LINKED ISSUE #1407: Advancement Type: Subclass
This advancement allows the player to select their subclass at a certain level. This can only be added to `"class"` items and each class can only have a single advancement of this type. Much of this advancement would be the same as the Item Choice advancement (#1401) and could probably be an extension of that one.

This would resolve #1017.

## Level Up Interface

The player would be presented with a list of available subclasses as well as have a drop area for adding an unlisted subclass. They can choose a single subclass before continuing.

The advancement would ensure that any item added is of a new `subclass` type created by #1080.

## Configuration

This configuration object would contain a `pool` of item UUIDs and an option to disable the drop area.

```javascript
{
  allowDrops: true,
  pool: ["Compendium.dnd5e.subclasses.MW1ExvBLm8Hg82aA"]
}
```

## Value

The value object would contain the origin UUID of the subclass added and the ID of the object added to the actor.

```javascript
{
  "EZsonMThTNLZq35j": "Compendium.dnd5e.subclasses.MW1ExvBLm8Hg82aA"
}
```

## Future Directions

If a system for scanning compendiums for items of specific types is implemented, this could be used to automatically build a list of available subclasses that would be presented to the player.

arbron:
Introduces a new subclass advancement! This advancement doesn't do anything on its own other than indicate the level at which the class takes a subclass due to limitations in the advancement system.

Subclass Advancement - List

Subclass Advancement - Flow

When a player reaches the level of the subclass advancement, their features tab adds a "Add Subclass" button where the subclass icon will eventually appear. Clicking this opens the compendium browser with the subclass list locked to the current class's identifier.

Subclass Advancement - Add Button

Subclass Advancement - Browser

While this doesn't fully implement what was originally desired from the subclass advancement, this workflow should work pretty well to fill in a major gap in the system's level-up workflow.

Closes #1017
Closes #1407


module/documents/advancement/subclass.mjs
```diff
@@ -0,0 +1,42 @@
...
+    const subclass = this.item.subclass;
+    if ( configMode || !item ) return "";
+    return subclass.toAnchor({ classes: ["content-link"] }).outerHTML;
+  }
+
+  /* -------------------------------------------- */
+  /*  Editing Methods                             */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
```
krbz999, line 38: inheritDoc or inheritdoc?

icons/svg/subclass.svg
Fyorl, line 1: We could also reuse the subclass icon from the item defaults, if having some consistency there is useful.

module/documents/advancement/subclass.mjs
```diff
@@ -0,0 +1,42 @@
...
+
+  /* -------------------------------------------- */
+  /*  Display Methods                             */
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  summaryforLevel(level, { configMode=false }={}) {
+    const subclass = this.item.subclass;
+    if ( configMode || !item ) return "";
+    return subclass.toAnchor({ classes: ["content-link"] }).outerHTML;
```
Fyorl, line 31: That `content-link` issue should be fixed now.

```suggestion
    if ( configMode || !subclass ) return "";
    return subclass.toAnchor().outerHTML;
```

templates/actors/parts/actor-classes.hbs
```diff
@@ -9,6 +9,12 @@
             {{#if cls.subclass}}
             <img class="gold-icon item-tooltip" src="{{ cls.subclass.img }}" alt="{{ cls.subclass.name }}"
                  data-action="edit" data-item-id="{{ cls.subclass.id }}">
+            {{else if ctx.needsSubclass}}
+            <a class="gold-icon subclass-icon" data-action="findItem" data-item-type="subclass"
+               data-class-identifier="{{ cls.identifier }}"  data-tooltip="DND5E.SubclassAdd"
```
Fyorl, line 14:
```suggestion
               data-class-identifier="{{ cls.identifier }}" data-tooltip="DND5E.SubclassAdd"
```

Fyorl (approved):

PR #4073, opened 2024-08-29
Implement 2024 shape-shift options for adding temp HP, and retaining HP, HD, and creature type.

Fyorl:


module/documents/actor/actor.mjs
```diff
@@ -3048,13 +3070,23 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
       if ( keepVision ) d.system.traits.senses = o.system.traits.senses;
 
+      // Keep creature type
+      if ( keepType ) d.system.details.type = o.system.details.type;
+
+      // Keep HP & HD
+      if ( keepHP ) d.system.attributes.hp = o.system.attributes.hp;
+
+      // Add temporary hit points
+      if ( addTemp ) d.system.attributes.hp.temp = target.system.attributes.hp.max;
```
arbron, line 3080: When Wild Shaping this doesn't guarantee HP stays the same if the Con mod is different between the two forms.

arbron (changes requested):
Will adding temp HP based on Druid level also be handled?

module/documents/actor/actor.mjs
Fyorl: Hmm, I've updated it now to set the *source* HP max equal to the subject's *derived* HP max. That should override HP calculations, regardless of constitution modifier?

arbron (approved):

PR #4074, opened 2024-08-29
[#3983] Add weapon mastery support to `TraitAdvancement`

arbron:
Weapon masteries can now be granted or chosen through `TraitAdvancement` using the new "Mastery" mode.

### DIFF
--- a/module/applications/advancement/trait-config.mjs
+++ b/module/applications/advancement/trait-config.mjs
@@ -84,7 +84,9 @@ export default class TraitConfig extends AdvancementConfig {
     context.selectedIndex = this.selected;
 
     context.validTraitTypes = Object.entries(CONFIG.DND5E.traits).reduce((obj, [key, config]) => {
-      if ( (this.config.mode === "default") || config.expertise ) obj[key] = config.labels.title;
+      if ( (this.config.mode === "default") || (this.config.mode === "mastery" ? config.mastery : config.expertise) ) {
+        obj[key] = config.labels.title;
+      }
       return obj;
     }, {});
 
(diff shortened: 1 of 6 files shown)


module/applications/advancement/trait-config.mjs
```diff
@@ -222,7 +230,8 @@ export default class TraitConfig extends AdvancementConfig {
 
     // If one of the expertise modes is selected, filter out any traits that are not of a valid type
     if ( (configuration.mode ?? this.config.mode) !== "default" ) {
-      const validTraitTypes = filteredKeys(CONFIG.DND5E.traits, c => c.expertise);
+      const checkKey = (configuration.mode ?? this.config.mode) === "mastery" ? "mastery" : "experties";
```
Fyorl, line 233:
```suggestion
      const checkKey = (configuration.mode ?? this.config.mode) === "mastery" ? "mastery" : "expertise";
```

module/documents/actor/trait.mjs
```diff
@@ -62,6 +62,8 @@ export async function actorValues(actor, trait) {
     data.value.forEach(v => setValue(v, 1));
   }
 
+  if ( (trait === "weapon") ) data.mastery?.forEach(v => setValue(v, 2));
```
Fyorl, line 65:
```suggestion
  if ( trait === "weapon" ) data.mastery?.forEach(v => setValue(v, 2));
```

Fyorl (changes requested):
Works well when I tested, I just noticed one small issue when configuring the trait advancement. It's pretty easy to reproduce by just switching the mode from expertise to mastery, or vice versa. The right-hand list of proficiencies does not update appropriately.

Fyorl (approved):

PR #4075, opened 2024-08-29
[#3983] Add mastery selection & display mastery property in chat

### LINKED ISSUE #3983: Add support for weapon masteries

arbron:
Adds the `traits.weaponProf.bonsuMasteries` data to characters that contains extra masteries that can be applied to any weapon in which they have mastery, in place of the weapon's existing mastery.

When making an attack roll using a weapon with which the character has mastery, a dropdown will appear in the dialog if there is more then one mastery option, otherwise the first mastery will be selected automatically for the attack.

Mastery in Chat

The mastery used will appear in the attack roll chat card and if a reference was provided for that mastery, it will be a link so players can hover and see a description of the mastery property.

Closes #3983

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -373,10 +373,25 @@ export default class ChatMessage5e extends ChatMessage {
    */
   _enrichAttackTargets(html) {
     const attackRoll = this.rolls[0];
-    const targets = this.getFlag("dnd5e", "targets");
     const visibility = game.settings.get("dnd5e", "attackRollVisibility");
     const isVisible = game.user.isGM || (visibility !== "none");
-    if ( !isVisible || !(attackRoll instanceof dnd5e.dice.D20Roll) || !targets?.length ) return;
+    if ( !isVisible || !(attackRoll instanceof dnd5e.dice.D20Roll) ) return;
+
+    const masteryConfig = CONFIG.DND5E.weaponMasteries[attackRoll.options.mastery];
+    if ( masteryConfig ) {
+      const p = document.createElement("p");
+      p.classList.add("supplement");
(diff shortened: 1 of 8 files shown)


module/documents/chat-message.mjs
```diff
@@ -373,10 +373,25 @@ export default class ChatMessage5e extends ChatMessage {
...
+    const masteryConfig = CONFIG.DND5E.weaponMasteries[attackRoll.options.mastery];
+    if ( masteryConfig ) {
+      const p = document.createElement("p");
+      p.classList.add("supplement");
+      let mastery = masteryConfig.label;
+      if ( masteryConfig.reference ) mastery = `
+        <a class="content-link" draggable="true" data-link data-uuid="${masteryConfig.reference}"
+           data-tooltip="${mastery}">${mastery}</a>
+      `;
+      p.innerHTML = game.i18n.format("DND5E.WEAPON.Mastery.Flavor", { mastery });
```
Fyorl, line 389: I don't think this should be done via localisation, we wouldn't expect the word order to change here based on language since it's in `key: value` format. That probably goes for 'on save' as well.

```suggestion
      p.innerHTML = `<strong>${game.i18n.localize("Mastery")}</strong> ${mastery}`;
```

Fyorl (approved):
See my comments but this looks good

Issue #4076, opened 2024-08-30
Allow Removal of Perception Skill  for NPCs with ActorSheet5eNPC2


Larkinabout:
If removing the Perception skill for homebrew purposes, the ActorSheet5eNPC2 app will error due to its requirement for `context.senses.passivePerception`: https://github.com/foundryvtt/dnd5e/blob/00724328fdffe35b813a121ee8c7e9f3cc01eb0b/module/applications/actor/npc-sheet-2.mjs#L109-L111

Can this please be modified to allow for Perception to be removed and therefore Passive Perception to be excluded from the NPC sheet.

PR #4077, opened 2024-08-30
[#4076] Allow removal of Perception for ActorSheet5eNPC2

Larkinabout:
Allows the removal of the Perception skill without error by checking if the actor has the `prc` skill before adding it to senses.

### DIFF
--- a/module/applications/actor/npc-sheet-2.mjs
+++ b/module/applications/actor/npc-sheet-2.mjs
@@ -119,7 +119,8 @@ export default class ActorSheet5eNPC2 extends ActorSheetV2Mixin(ActorSheet5eNPC)
     context.skills = Object.fromEntries(Object.entries(context.skills).filter(([, v]) => v.value));
 
     // Senses
-    context.senses.passivePerception = {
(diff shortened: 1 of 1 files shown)


Larkinabout:
Closes #4076.

module/applications/actor/npc-sheet-2.mjs
```diff
@@ -119,7 +119,8 @@ export default class ActorSheet5eNPC2 extends ActorSheetV2Mixin(ActorSheet5eNPC)
     context.skills = Object.fromEntries(Object.entries(context.skills).filter(([, v]) => v.value));
 
     // Senses
-    context.senses.passivePerception = {
+    if (Object.hasOwn(this.actor.system.skills, 'prc')) {
```
krbz999, line 122: Got a syntax error here. As well as linting but it's probably enough to just check `if ( this.actor.system.skills.prc )`

Larkinabout: Serves me right me for not actually building it out. I'll sort in a sec.

Fyorl (approved):

PR #4078, opened 2024-08-30
Migrate SRD compendiums

arbron:


Fyorl (changes requested):
I double-checked several migrations and all but one looked good. I noticed that dragon breath weapons did not migrate to recharge correctly, and noticed the saving throws for breath weapons and frightful presence also did not migrate correctly.

image

arbron:
> I double-checked several migrations and all but one looked good. I noticed that dragon breath weapons did not migrate to recharge correctly, and noticed the saving throws for breath weapons and frightful presence also did not migrate correctly.

I'll check out the Recharge issue. For that, it looks like Adult Black Dragon just didn't have a DC set before the migration. All of the other dragons seem to have good DCs. I'm doing a followup PR after this to clean up some issues, so I'll add that to the list.

arbron:
@Fyorl Migrate the recharge into the activity uses, so seeing them at the moment might be dependent on your UI PR.

Fyorl:
> @Fyorl Migrate the recharge into the activity uses, so seeing them at the moment might be dependent on your UI PR.

I don't think it's dependent on that PR. Items have base recovery, which is what should be shown on the main item line in the sheet. If the recovery is migrated into the activity, it won't show on the item's line, only in the activity breakdown. Is there a problem with migrating to base recovery?

arbron:
> > @Fyorl Migrate the recharge into the activity uses, so seeing them at the moment might be dependent on your UI PR.
> 
> I don't think it's dependent on that PR. Items have base recovery, which is what should be shown on the main item line in the sheet. If the recovery is migrated into the activity, it won't show on the item's line, only in the activity breakdown. Is there a problem with migrating to base recovery?

Well I used activity uses to track the charges, rather than the whole item uses, so the recharge occurs at the activity level. I move them out to item uses, but that would conflicts with the other uses if the item also has Limited Uses set before the migration. As it is:

Limited Uses => Item Uses
Recharge => Activity Uses

Fyorl:
> Well I used activity uses to track the charges, rather than the whole item uses, so the recharge occurs at the activity level. I move them out to item uses, but that would conflicts with the other uses if the item also has Limited Uses set before the migration. As it is:
> 
> Limited Uses => Item Uses
> Recharge => Activity Uses

Right, because it was possible to have limited uses *and* a recharge before, I understand now. However, I don't think the vast majority of items ever used both, and I was fairly certain these breath weapons didn't either. Looking at our original migration code, it looks like the breath weapon should have correctly ended up with a recharge recovery, so I'm not sure why it didn't. Did they actually use both for some reason, and that caused it to fail?

We should migrate recharges into the base item if at all possible, otherwise every single recharge feature on all monsters are going to disappear off their sheets.

arbron:
> We should migrate recharges into the base item if at all possible, otherwise every single recharge feature on all monsters are going to disappear off their sheets.

Okay, migrated to use item uses for recharge unless item uses were previously defined, in which case it uses activity uses.

We should make sure that we handle the use case of an item with a single activity that contains recharge as well as item uses. That might be as simple as always showing activities in the dropdown even if an item only has one.

Fyorl (approved):
> That might be as simple as always showing activities in the dropdown even if an item only has one.

Yeah, will probably have to do this. Migrations aside you can already configure an item with bases uses and a single activity with its own uses, which I hadn't considered at the time.

PR #4080, opened 2024-08-30
[#4079] Handle limited uses on granted spells

### LINKED ISSUE #4079: Update `ItemGrantActivity` & `ItemChoiceActivity` to handle activities when adding limited uses
The current method of setting limited uses on spells doesn't quite work with the new activities system, because it doesn't set consumption. It also tries to set `ability` on spells which doesn't exist anymore.

arbron:
Handles setting the limited uses on spells granted through an advancement in a way that respects activities. To support all the desired workflows, a "Requires Slot" checkbox has been added to the spell config which controls how the limited uses are applied.

- If preparation mode doesn't require a spell slot by default or "Requires Slot" is checked, then consumption is added to all activities on the spell marked to "Consume Spell Slot"
- If preparation mode requies a slot and "Requires Slot" isn't checked, then each activity marked to "Consume Spell Slot" is duplicated, given consumption, removed spell slot consumption, and has "Free Casting" added to its name

This combination should handle all of the standard configurations needed for official content.

This PR also reintroduces `system.ability` to spells as a means of overriding the default ability from the class or actor. This field only appears for embedded spells and shows the default ability when not set.

Closes #4079

### DIFF
--- a/module/data/advancement/spell-config.mjs
+++ b/module/data/advancement/spell-config.mjs
@@ -43,8 +43,9 @@ export default class SpellConfigurationData extends foundry.abstract.DataModel {
     if ( this.preparation ) foundry.utils.setProperty(itemData, "system.preparation.mode", this.preparation);
 
     if ( this.uses.max && this.uses.per ) {
-      foundry.utils.setProperty(itemData, "system.uses.per", this.uses.per);
       foundry.utils.setProperty(itemData, "system.uses.max", this.uses.max);
+      itemData.system.uses.recovery ??= [];
+      itemData.system.uses.recovery.push({ period: this.uses.per, type: "recoverAll" });
 
       const preparationConfig = CONFIG.DND5E.spellPreparationModes[itemData.system.preparation?.mode];
       const createForwardActivity = !this.uses.requireSlot && preparationConfig?.upcast;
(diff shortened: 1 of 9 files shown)


module/data/advancement/spell-config.mjs
```diff
@@ -30,12 +31,58 @@ export default class SpellConfigurationData extends foundry.abstract.DataModel {
...
+   * @param {string} [config.ability]  Spellcasting ability selected during advancement process.
+   */
+  applySpellChanges(itemData, { ability }={}) {
+    ability = this.ability.size ? this.ability.has(ability) ? ability : this.ability.first() : null;
+    if ( ability ) foundry.utils.setProperty(itemData, "system.ability", ability);
+    if ( this.preparation ) foundry.utils.setProperty(itemData, "system.preparation.mode", this.preparation);
+
+    if ( this.uses.max && this.uses.per ) {
+      foundry.utils.setProperty(itemData, "system.uses.per", this.uses.per);
+      foundry.utils.setProperty(itemData, "system.uses.max", this.uses.max);
```
Fyorl, line 47:
```suggestion
      foundry.utils.setProperty(itemData, "system.uses.max", this.uses.max);
      itemData.system.uses.per ??= [];
      itemData.system.uses.per.push({ period: this.uses.per, type: "recoverAll" });
```

Fyorl (approved):

PR #4081, opened 2024-08-30
[#3913, #3914] Handle enchantments that target moved fields

arbron:
Adds system data model for enchantment active effects that has custom application logic to handle enchantments targeting fields that no longer exist on items. These handle `attack.bonus`, `attack.flat`, `damage.parts`, `save.dc`, `save.scaling`. These fields cover everything applied by summoning as well as the most common fields modified by enchantment spells or features.

### DIFF
--- a/module/data/active-effect/enchantment.mjs
+++ b/module/data/active-effect/enchantment.mjs
@@ -19,9 +19,8 @@ export default class EnchantmentData extends foundry.abstract.TypeDataModel {
    * @returns {boolean|void}             Return false to prevent normal application from occurring.
    */
   _applyLegacy(item, change, changes) {
-    console.log("EnchantmentData#_applyLegacy", item, change, changes);
     let key = change.key.replace("system.", "");
-    switch (change.key) {
+    switch ( change.key ) {
       case "system.attack.bonus":
       case "system.attack.flat":
         for ( const activity of item.system.activities?.getByTypes("attack") ?? [] ) {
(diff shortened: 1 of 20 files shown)


module/data/active-effect/enchantment.mjs
```diff
@@ -0,0 +1,67 @@
...
+            const value = damage.clone();
+            value.enchantment = true;
+            value.locked = true;
+            changes[`system.activities.${activity.id}.damage.parts`] = ActiveEffect.applyField(
+              activity, { ...change, key: "damage.parts", value: value }
+            );
+          }
+          return false;
+        } catch(err) {
+          console.log(err);
```
krbz999, line 47: Should we return false here as well if an error is thrown?

module/data/active-effect/enchantment.mjs
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Handle enchantment-specific changes to the item.
+   * @param {Item5e} item                The Item to whom this effect should be applied.
+   * @param {EffectChangeData} change    The change data being applied.
+   * @param {Record<string, *>} changes  The aggregate update paths and their updated values.
+   * @returns {boolean|void}             Return false to prevent normal application from occurring.
+   */
+  _applyLegacy(item, change, changes) {
+    console.log("EnchantmentData#_applyLegacy", item, change, changes);
```
krbz999, line 22: Log-a-rooni.

module/data/active-effect/enchantment.mjs
```diff
@@ -0,0 +1,67 @@
...
+    let key = change.key.replace("system.", "");
+    switch (change.key) {
```
Fyorl, line 24:
```suggestion
    switch ( change.key ) {
```

module/data/active-effect/enchantment.mjs
```diff
@@ -0,0 +1,67 @@
...
+          let damage;
+          const parsed = JSON.parse(change.value);
+          if ( foundry.utils.getType(parsed) === "Object" ) damage = new DamageData(parsed);
+          else damage = new DamageData({ custom: { enabled: true, formula: parsed[0][0] }, types: [parsed[0][1]] });
+          for ( const activity of item.system.activities?.getByTypes("attack", "damage", "save") ?? [] ) {
+            const value = damage.clone();
+            value.enchantment = true;
+            value.locked = true;
+            changes[`system.activities.${activity.id}.damage.parts`] = ActiveEffect.applyField(
+              activity, { ...change, key: "damage.parts", value: value }
```
Fyorl, line 42:
```suggestion
              activity, { ...change, key, value }
```

Fyorl (approved):
Nice

module/data/active-effect/enchantment.mjs
arbron: My thought was just to fall back to the standard behavior and see if it can make sense of it.

PR #4090, opened 2024-09-01
masteryOptions can be null, account for this in attack activity

MrPrimate:
If not weapon mastery options, weapon attacks failed. Null can be returned by masteryOptions get.


Fyorl:
This change would mean that `dialogConfig.masteryOptions` is never `undefined`, so we unnecessarily show a blank drop-down. I've pushed a fix for rolling attacks that preserves passing `undefined` to `dialogConfig.masterOptions` when appropriate.

MrPrimate:
Thanks, wasn't sure if this was the right approach!

PR #4094, opened 2024-09-01
[#4093] Update to magical axe weapon icons

### LINKED ISSUE #4093: Diversify magic weapon icons: Axes
Whilst some weapons like blowguns or slings may have very few icon options, there are a variety of icons for other weapon types. 

Some weapons may have to be stuck with all magical variants having the same icon (as is the current format), but this should be the fallback rather than the precedent, and ultimately improved icon diversity would be nice to have.

Here are my proposed icons for the first set of changes, axes:

Battleaxes:

image

Greataxes:

image

Handaxes:

image

PR incoming, followed by other weapon types. Any specific changes here should take precedent over the general changes in #4083

M0nk3yy:
Closes #4093 and takes precedent over #4083


M0nk3yy:
Checked and conflicts resolved.

PR #4098, opened 2024-09-02
Display activities in-sheet

Fyorl:
image

image

image


module/data/shared/uses-field.mjs
```diff
@@ -145,4 +151,68 @@ export default class UsesField extends SchemaField {
...
+   * Rolls a recharge test for an Item or Activity that uses the d6 recharge mechanic.
+   * @this {Item5e|Activity}
+   * @returns {Promise<Roll|void>}
+   */
+  static async rollRecharge() {
+    const uses = this.system ? this.system.uses : this.uses;
+    const recharge = uses?.recovery.find(({ period }) => period === "recharge");
+    if ( !recharge ) return;
+
+    const rollConfig = {
```
arbron, line 167: Should we swap this over to using `BasicRoll` and the new APIs? I could handle adding the deprecated hooks in a followup PR if you want.

arbron (approved):

module/data/shared/uses-field.mjs
Fyorl: We can use `BasicRoll` but I think it wouldn't affect anything since we never show a dialog. This `rollConfig` is really just for the hooks I think. Or do you mean we should show a dialog?

arbron: Just for the hooks, yeah.

Fyorl: Sure, we can do that they for some consistency

PR #4106, opened 2024-09-02
[#3801] Automatically determine source book from module

### LINKED ISSUE #3801: Source field should default to registered source if possible
When the source field on an item or NPC is filled in, the system should endeavor to populate it with a source book if a module has registered one (see https://github.com/foundryvtt/dnd5e/issues/3799). It can use the current compendium or source ID to match a document to its originating module and then check to see if that module has `flags.dnd5e.sourceBooks` in its manifest. If the source data for the item is empty and a single source book is declared in the manifest, it should then use that information to display a default source.

arbron:
Uses the new `sourceBooks` manifest flag to automatically select the proper source book to use for source labels if no source book is specified. This only works for documents in compendiums (or imported from compendiums) for modules that define a single source book.

Also re-designs the source configuration app using ApplicationV2.

Screenshot 2024-09-02 at 16 36 41

Closes #3801

### DIFF
--- a/module/applications/source-config.mjs
+++ b/module/applications/source-config.mjs
@@ -47,7 +47,7 @@ export default class SourceConfig extends DocumentSheet5e {
     const context = await super._prepareContext(options);
     context.buttons = [{ icon: "fa-regular fa-save", label: "Submit", type: "submit" }];
     context.data = foundry.utils.getProperty(this.document, this.options.keyPath);
-    context.fields = this.document.system.schema.getField(this.options.keyPath.replace("system.", "")).fields;
+    context.fields = this.document.system.schema.getField(this.options.keyPath.replace(/^system\./, "")).fields;
     context.keyPath = this.options.keyPath;
     context.source = foundry.utils.getProperty(this.document.toObject(), this.options.keyPath);
     context.sourceUuid = this.document._stats.compendiumSource;
--- a/module/applications/source-config.mjs
+++ b/module/applications/source-config.mjs
@@ -1,25 +1,41 @@
+import DocumentSheet5e from "./api/document-sheet.mjs";
+
 /**
  * Application for configuring the source data on actors and items.
  */
-export default class SourceConfig extends DocumentSheet {
-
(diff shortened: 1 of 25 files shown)


module/applications/source-config.mjs
```diff
@@ -1,50 +1,57 @@
...
-    context.appId = this.id;
-    context.CONFIG = CONFIG.DND5E;
-    context.source = foundry.utils.getProperty(this.document, this.options.keyPath);
-    context.sourceUuid = this.document._stats.compendiumSource
-      ?? foundry.utils.getProperty(this.document, "flags.core.sourceId");
+  async _prepareContext(options) {
+    const context = await super._prepareContext(options);
+    context.buttons = [{ icon: "fa-regular fa-save", label: "Submit", type: "submit" }];
+    context.data = foundry.utils.getProperty(this.document, this.options.keyPath);
+    context.fields = this.document.system.schema.getField(this.options.keyPath.replace("system.", "")).fields;
```
Fyorl, line 50:
```suggestion
    context.fields = this.document.system.schema.getField(this.options.keyPath.replace(/^system\./, "")).fields;
```

module/data/shared/source-field.mjs
```diff
@@ -21,28 +29,52 @@ export default class SourceField extends SchemaField {
...
+    let manifest;
+    const pack = foundry.utils.parseUuid(uuid)?.collection?.metadata;
+    switch ( pack?.packageType ) {
+      case "module": manifest = game.modules.get(pack.packageName); break;
+      case "system": manifest = game.system; break;
+      case "world": manifest = game.world; break;
+    }
+    if ( !manifest ) return null;
+    const sourceBooks = manifest.flags?.dnd5e?.sourceBooks;
+    const keys = Array.from(Object.keys(sourceBooks ?? {}));
```
Fyorl, line 76: `Object.keys` already returns an array.

```suggestion
    const keys = Object.keys(sourceBooks ?? {});
```

Fyorl (approved):

PR #4114, opened 2024-09-03
[#4107] Fix applying rider effects on enchantment. Fix flagging rider effects.

### LINKED ISSUE #4107: 4.0.x: Riders on Enchantments don't get transfered
Riders from Enchantments are still fetched from the flag `dnd5e.enchantment.riders.*`, they are now stored on the activity. As such they do not get applied.

Fyorl:
- Closes #4107

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -413,12 +413,31 @@ export default class ActiveEffect5e extends ActiveEffect {
    * Create additional effects that are applied separately from an enchantment.
    * @param {object} options  Options passed to the effect creation.
    */
-  async createRiderEnchantments(options) {
-    const message = game.messages.get(options?.chatMessageOrigin);
-    const item = message?.getAssociatedItem();
-    const activity = message?.getAssociatedActivity();
-    const profile = activity?.effects.find(e => e._id === message?.getFlag("dnd5e", "use.enchantmentProfile"));
-    if ( !profile ) return;
+  async createRiderEnchantments(options={}) {
+    let item;
+    let profile;
+    const { chatMessageOrigin } = options;
+    const { enchantmentProfile, activityId } = options.dnd5e ?? {};
+
+    if ( chatMessageOrigin ) {
+      const message = game.messages.get(options?.chatMessageOrigin);
+      item = message?.getAssociatedItem();
+      const activity = message?.getAssociatedActivity();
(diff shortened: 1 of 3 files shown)


module/documents/active-effect.mjs
```diff
@@ -414,25 +414,28 @@ export default class ActiveEffect5e extends ActiveEffect {
    * @param {object} options  Options passed to the effect creation.
    */
   async createRiderEnchantments(options) {
-    const origin = await fromUuid(this.origin)?? game.messages.get(options?.chatMessageOrigin)?.getAssociatedItem();
```
Fyorl, line 417: This is something we lost when we moved the enchantment data out of flags. We cannot get the rider information from the AE anymore, we have to look it up on the original enchantment Activity, but to do that we need to know the `enchantmentProfile` that was used. We can't get this information at all without `options.chatMessageOrigin`.

So we lose the ability to apply enchantments by drag-dropping the enchantment AE onto an Item. They must be applied via the chat card now. If they don't have riders then I suppose it's still possible to apply via drag-drop.

I don't know of an easy way around this limitation where the `origin` points to the Item rather than the Activity (in the case of drag-dropping the AE), since the relationship from enchantment Activities to AEs is many-to-many now.

module/documents/item.mjs
```diff
@@ -1451,6 +1451,33 @@ export default class Item5e extends SystemDocumentMixin(Item) {
 
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  _onUpdate(changed, options, userId) {
```
Fyorl, line 1455: This is restoring functionality that was lost from the old enchantment config. We need to flag AEs as riders, otherwise they affect the Actor when they're supposed to just be stored on the Item for use as part of an enchantment.

module/documents/active-effect.mjs
krbz999: Could we override drag-drop and add the id of an activity from the item?

Fyorl: There could be multiple activities that reference the same AE, so we can't know which one to use in the general case.

krbz999: Using the first found might be "good enough" in the meantime?

Fyorl: Yeah, you're right, it's better than nothing

arbron (approved):

PR #4123, opened 2024-09-04
[#4024] Rework Check Activity to support multiple checks

arbron:
Changes the `check.associated` value on `CheckActivity` to be a set of associated values rather than a single value. If multiple associated values are set then more than one button will appear in the chat card.

Check Activity - Multiple Associated

Setting an explicit ability for the activity will cause all of the created checks to use that ability, otherwise they will attempt to determine the best ability to use (skills from the default value in config, tools from the tool item or character data).

### DIFF
--- a/module/applications/activity/check-sheet.mjs
+++ b/module/applications/activity/check-sheet.mjs
@@ -39,15 +39,15 @@ export default class CheckSheet extends ActivitySheet {
       ...Object.entries(CONFIG.DND5E.abilities).map(([value, config]) => ({ value, label: config.label }))
     ];
     let ability;
-    if ( this.item.type === "tool" ) ability = CONFIG.DND5E.abilities[this.item.system.ability]?.label?.toLowerCase();
-    else if ( this.activity.checkType === "skill" ) ability = CONFIG.DND5E.abilities[
-      CONFIG.DND5E.skills[this.activity.check.associated].ability
-    ]?.label?.toLowerCase();
+    const associated = this.activity.check.associated;
+    if ( (this.item.type === "tool") && !associated.size ) {
+      ability = CONFIG.DND5E.abilities[this.item.system.ability]?.label?.toLowerCase();
+    } else if ( (associated.size === 1) && associated.first() in CONFIG.DND5E.skills ) {
+      ability = CONFIG.DND5E.abilities[CONFIG.DND5E.skills[associated.first()].ability]?.label?.toLowerCase();
+    }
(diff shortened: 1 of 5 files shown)


module/applications/activity/check-sheet.mjs
```diff
@@ -39,15 +39,15 @@ export default class CheckSheet extends ActivitySheet {
...
     ];
     let ability;
-    if ( this.item.type === "tool" ) ability = CONFIG.DND5E.abilities[this.item.system.ability]?.label?.toLowerCase();
-    else if ( this.activity.checkType === "skill" ) ability = CONFIG.DND5E.abilities[
-      CONFIG.DND5E.skills[this.activity.check.associated].ability
-    ]?.label?.toLowerCase();
+    const associated = this.activity.check.associated;
+    if ( (this.item.type === "tool") && !associated.size ) {
+      ability = CONFIG.DND5E.abilities[this.item.system.ability]?.label?.toLowerCase();
+    } else if ( (associated.size === 1) && associated.first() in CONFIG.DND5E.skills ) {
```
Fyorl, line 45:
```suggestion
    } else if ( (associated.size === 1) && (associated.first() in CONFIG.DND5E.skills) ) {
```

module/data/activity/check-data.mjs
```diff
@@ -61,6 +61,15 @@ export default class CheckActivityData extends BaseActivityData {
   /*  Data Migrations                             */
   /* -------------------------------------------- */
 
+  /** @override */
+  static migrateData(source) {
```
Fyorl, line 65: We shouldn't need to migrate anything. This data model changed during development, not between releases.

module/data/activity/check-data.mjs
Fyorl, line 1: The `checkType` getter needs to be updated (or deleted).

Fyorl (changes requested):

module/data/activity/check-data.mjs
```diff
@@ -61,6 +61,15 @@ export default class CheckActivityData extends BaseActivityData {
...
+    if ( foundry.utils.getType(source.check?.associated) === "string" ) {
+      source.check.associated = source.check.associated ? [source.check.associated] : [];
+    }
+  }
```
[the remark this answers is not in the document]
arbron: Yes, but `new MODEL({test: ""})` results in `test = Set([""])`

Fyorl (approved):

PR #4124, opened 2024-09-04
Add default abilities to tools.

Fyorl:


module/config.mjs
```diff
@@ -505,49 +505,173 @@ DND5E.toolProficiencies = {
 };
 preLocalize("toolProficiencies", { sort: true });
 
+DND5E.tools = {
```
arbron, line 508:
```suggestion
/**
 * @typedef {object} ToolConfiguration
 * @property {string} ability  Default ability used for the tool.
 * @property {string} id       UUID of reference tool or ID within pack defined by `DND5E.sourcePacks.ITEMS`.
 */

/**
 * Configuration data for tools.
 * @enum {ToolConfiguration}
 */
DND5E.tools = {
```

arbron (approved):
I've merged in the changes to `CheckActivity` so you'll probably want to tweak the code in `getAbility` to use these new abilities.

PR #4127, opened 2024-09-04
Toggle localization changes with Rules Version setting

arbron:
Adds legacy versions of certain names back and uses them if the legacy rules are set (Race/Species, Armor Proficiency/Armor Training, Exotic Languages/Rare Languages).

Also removes the old `ITEM.Types*` localization strings that haven't been used in several versions.


Fyorl (approved):
We should probably change 'emanation' back to 'radius' in the legacy rules too. If you only have the 2014 rules then you probably don't know what 'emanation' is.

PR #4133, opened 2024-09-04
Document activities system and each activity type for Wiki

arbron:


Fyorl (approved):
The `attack-chat-dialog.jpg` screenshot shows an old-style damage dialog which could do with an update, but not urgent.

PR #4136, opened 2024-09-04
Update AE Guide, Enrichers, & Roll Formulas, document Module Registration

arbron:


wiki/Enrichers.md
```diff
@@ -99,6 +99,10 @@ Item enrichers allow you to use an item from an enriched link. There are several
 
 This functions similarly to a system macro, as if you had dragged that item to the macro hotbar. When clicked, it will check for a selected token, or your assigned actor. If the token or actor has an item of a matching name, it will be used, otherwise a warning will be displayed.
 
+**By Item & Activity Name**: `[[/item Bite activity=Poison]]`
+
+Using the item name without an activity will cause the activity selection dialog to open if more than one activity is present. To trigger a specific activity on the item the activity name can be included. The activity name must be proceeded by `activity=`, and it must be wrapped in paranthesis if there is a space in the name (e.g. `[[/item Tentacles activity="Escape Tentacles"]]`).
```
Fyorl, line 104:
```suggestion
Using the item name without an activity will cause the activity selection dialog to open if more than one activity is present. To trigger a specific activity on the item the activity name can be included. The activity name must be proceeded by `activity=`, and it must be wrapped in quotes if there is a space in the name (e.g. `[[/item Tentacles activity="Escape Tentacles"]]`).
```

Fyorl (approved):

PR #4137, opened 2024-09-04
Add rules version & revision to source data

arbron:
Adds two new data fields to source data:
- `rules`: Was this item created for the legacy or modern rules
- `revision`: Internal revision number usable for tracking changes when a module updates.

Also moves the source data out of `system.details` for NPCs and vehicles so all documents have it in the same location.

### DIFF
--- a/module/applications/source-config.mjs
+++ b/module/applications/source-config.mjs
@@ -13,8 +13,7 @@ export default class SourceConfig extends DocumentSheet5e {
     },
     form: {
       closeOnSubmit: true
-    },
-    keyPath: "system.details.source"
+    }
   };
 
   /* -------------------------------------------- */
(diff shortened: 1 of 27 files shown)


module/applications/source-config.mjs
```diff
@@ -35,7 +34,7 @@ export default class SourceConfig extends DocumentSheet5e {
 
   /** @override */
   get title() {
-    return game.i18n.localize("DND5E.SourceConfig");
+    return game.i18n.localize("DND5E.SOURCE.Action.Configure");
```
krbz999, line 37: Since `window.title` is localized automatically, is this getter even needed?

arbron: This is because we are expanding `DocumentSheetV2`, which provides its own title getter that uses the document's name instead.

templates/journal/page-spell-list-unlinked-config.hbs
Fyorl, line 1: Does this template also need the extra fields we added to the source config?

Fyorl (approved):

templates/journal/page-spell-list-unlinked-config.hbs
arbron: I'm not sure any of those other fields are helpful here, but I'll exclude them from the data model.

PR #4138, opened 2024-09-05
[#3803] Add rider activities to enchantments

### LINKED ISSUE #3803: Add rider activities to enchantments
Once activities are implemented, it would be good to implement rider activities similar to rider effects so enchantments can easily add activities to items.

arbron:
Allows activities to be applied to the item alongside enchantments like effects and items are currently.

The previous method of using a flag on effects to mark them as riders wasn't practical in this case due to activites not having flags, so this stores the rider activities and effects in a new flag on the item itself. A migration is included to move the `rider` flag from the effects to the item flags.

Closes #3803

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1319,6 +1319,10 @@
         },
         "riders": {
           "label": "Attached",
+          "activity": {
+            "label": "Additional Activities",
+            "hint": "These additional activities will be added to the enchant item when this enchantment is added, and removed when the enchantment is removed."
+          },
           "effect": {
             "label": "Additional Effects",
             "hint": "These additional effects will be added to the enchanted item when this enchantment is added, and removed when the enchantment is removed."
--- a/lang/en.json
+++ b/lang/en.json
@@ -1321,11 +1321,11 @@
           "label": "Attached",
           "activity": {
             "label": "Additional Activities",
-            "hint": "These additional activities will be added to the enchant item when this enchantment is added, and removed when the enchantment is removed."
+            "hint": "These additional activities will be added to the enchanted item when this enchantment is applied, and removed when the enchantment is removed."
           },
           "effect": {
             "label": "Additional Effects",
-            "hint": "These additional effects will be added to the enchanted item when this enchantment is added, and removed when the enchantment is removed."
+            "hint": "These additional effects will be added to the enchanted item when this enchantment is applied, and removed when the enchantment is removed."
           },
           "item": {
             "label": "Additional Items",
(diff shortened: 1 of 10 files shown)


lang/en.json
```diff
@@ -1319,6 +1319,10 @@
         },
         "riders": {
           "label": "Attached",
+          "activity": {
+            "label": "Additional Activities",
+            "hint": "These additional activities will be added to the enchant item when this enchantment is added, and removed when the enchantment is removed."
```
Fyorl, line 1324:
```suggestion
            "hint": "These additional activities will be added to the enchanted item when this enchantment is applied, and removed when the enchantment is removed."
```

module/applications/activity/activity-choice-dialog.mjs
```diff
@@ -85,8 +85,10 @@ export default class ActivityChoiceDialog extends Application5e {
         `<img src="systems/dnd5e/icons/svg/mouse-left.svg" alt="${game.i18n.localize("DND5E.Controls.LeftClick")}">`
       );
     }
-    const activities = this.#item.system.activities.map(this._prepareActivityContext.bind(this));
-    activities.sort((a, b) => a.sort - b.sort);
+    const activities = this.#item.system.activities
+      .filter(a => !this.#item.flags.dnd5e?.riders?.activity?.includes(a.id))
```
Fyorl, line 89:
```suggestion
      .filter(a => !this.#item.getFlag("dnd5e", "riders.activity")?.includes(a.id))
```

module/documents/item.mjs
```diff
@@ -801,11 +801,12 @@ export default class Item5e extends SystemDocumentMixin(Item) {
       );
       event = dialog?.event;
     }
-    if ( this.system.activities.size ) {
+    let activities = this.system.activities?.filter(a => !this.flags.dnd5e?.riders?.activity?.includes(a.id));
```
Fyorl, line 804:
```suggestion
    let activities = this.system.activities?.filter(a => !this.getFlag("dnd5e", "riders.activity")?.includes(a.id));
```

module/migration.mjs
```diff
@@ -477,7 +477,10 @@ export function migrateItemData(item, itemData, migrationData, flags={}) {
 
   // Migrate embedded effects
   if ( itemData.effects ) {
-    const effects = migrateEffects(itemData, migrationData);
+    if ( itemData.flags?.dnd5e?.riders?.effect ) {
+      itemUpdateData["flags.dnd5e.riders.effect"] = itemData.flags.dnd5e.riders.effect;
+    }
+    const effects = migrateEffects(itemData, migrationData, updateData);
```
Fyorl, line 483: Need to make sure items with this rider flag don't get migrated every time, since we set `diff: false` for these updates.

```suggestion
    const riders = foundry.utils.getProperty(itemData, "flags.dnd5e.riders.effect");
    if ( riders?.length ) updateData["flags.dnd5e.riders.effect"] = riders;
    const effects = migrateEffects(itemData, migrationData, updateData);
    if ( riders?.length === updateData["flags.dnd5e.riders.effect"]?.length ) delete updateData["flags.dnd5e.riders.effect"];
```

Fyorl (approved):

PR #4141, opened 2024-09-05
[#3785] Add refund button to chat cards

### LINKED ISSUE #3785: Add control for refunding uses from chat card
There are some abilities that allow recovering uses if an attempt is not successful, and sometimes players activate something that spends uses only to realize they can't use that ability or just want to do something else. It would be good to have a way from a chat card to refund whatever consumption was consumed when an item was activated.

The system should probably store details on what was consumed in the chat card so that it knows what can be restored.

This would probably be best as a context menu option, since I doubt it would come up frequently enough to warrant its own button.

arbron:
Adds a button for refunding consumption to chat cards. To support this the system calculates deltas on the actor and item changes during the consumption process and stores those values in the message flags.

Closes #3785

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -143,10 +143,11 @@ export default Base => class extends PseudoDocumentMixin(Base) {
    * Message configuration for activity usage.
    *
    * @typedef {object} ActivityMessageConfiguration
-   * @property {boolean} [create=true]  Whether to automatically create a chat message (if true) or simply return
-   *                                    the prepared chat message data (if false).
-   * @property {object} [data={}]       Additional data used when creating the message.
-   * @property {string} [rollMode]      The roll display mode with which to display (or not) the card.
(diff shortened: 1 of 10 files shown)


module/documents/activity/mixin.mjs
```diff
@@ -965,12 +1038,29 @@ export default Base => class extends PseudoDocumentMixin(Base) {
...
+   * Handle refunding consumption from a chat card.
+   * @param {PointerEvent} event     Triggering click event.
+   * @param {HTMLElement} target     The capturing HTML element which defined a [data-action].
+   * @param {ChatMessage5e} message  Message associated with the activation.
+   */
+  async #refundResource(event, target, message) {
+    const consumed = message.getFlag("dnd5e", "use.consumed");
+    const messageConfig = {};
+    if ( !foundry.utils.isEmpty(consumed) ) {
+      await this.refund(consumed, messageConfig);
```
Fyorl, line 1057: Seems this parameter isn't used.
```suggestion
      await this.refund(consumed);
```

Fyorl (approved):

PR #4144, opened 2024-09-06
[#4143] Move Active Effect formula changes into FormulaField

### LINKED ISSUE #4143: Move custom handling for `FormulaField` into the field itself
Currently we implement custom application on formula fields in `ActiveEffect5e`, but in V12 core added a way to define this on `DataField` so we should take advantage of that.

arbron:
Moves the custom active effect handling for formulas into the `FormulaField` class using the new system offered by Foundry V12.

Also refactors a lot of the other application behavior in `ActiveEffect5e` to avoid duplicating a bunch of core's code. This includes removing some handling of overriding sets because that is properly handled by core now.

Closes #4143

### DIFF
--- a/module/data/fields/formula-field.mjs
+++ b/module/data/fields/formula-field.mjs
@@ -29,5 +29,54 @@ export default class FormulaField extends foundry.data.fields.StringField {
     }
     super._validateType(value);
   }
+
+  /* -------------------------------------------- */
+  /*  Active Effect Integration                   */
+  /* -------------------------------------------- */
+
+  /** @override */
+  _castChangeDelta(delta) {
+    return this._cast(delta).trim();
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeAdd(value, delta, model, change) {
+    if (!value) return delta;
+    const operator = delta.startsWith("-") ? "-" : "+";
+    delta = delta.replace(/^[+-]?/, "").trim();
+    return `${value} ${operator} ${delta}`;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeMultiply(value, delta, model, change) {
+    if (!value) return delta;
+    const terms = new Roll(value).terms;
+    if ( terms.length > 1 ) return `(${value}) * ${delta}`;
+    return `${value} * ${delta}`;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeUpgrade(value, delta, model, change) {
+    if (!value) return delta;
+    const terms = new Roll(value).terms;
+    if ( (terms.length === 1) && (terms[0].fn === "max") ) return current.replace(/\)$/, `, ${delta})`);
+    return `max(${value}, ${delta})`;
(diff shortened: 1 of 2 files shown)


module/data/fields/formula-field.mjs
```diff
@@ -29,5 +29,54 @@ export default class FormulaField extends foundry.data.fields.StringField {
...
+  /** @override */
+  _castChangeDelta(delta) {
+    return this._cast(delta).trim();
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeAdd(value, delta, model, change) {
+    if (!value) return delta;
```
Fyorl, line 46:
```suggestion
    if ( !value ) return delta;
```

module/data/fields/formula-field.mjs
```diff
@@ -29,5 +29,54 @@ export default class FormulaField extends foundry.data.fields.StringField {
...
+    const operator = delta.startsWith("-") ? "-" : "+";
+    delta = delta.replace(/^[+-]?/, "").trim();
```
Fyorl, line 48:
```suggestion
    delta = delta.replace(/^[+-]/, "").trim();
```

module/data/fields/formula-field.mjs
```diff
@@ -29,5 +29,54 @@ export default class FormulaField extends foundry.data.fields.StringField {
...
+    return `${value} ${operator} ${delta}`;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeMultiply(value, delta, model, change) {
+    if (!value) return delta;
```
Fyorl, line 56:
```suggestion
    if ( !value ) return delta;
```

module/data/fields/formula-field.mjs
```diff
@@ -29,5 +29,54 @@ export default class FormulaField extends foundry.data.fields.StringField {
...
+    const terms = new Roll(value).terms;
+    if ( terms.length > 1 ) return `(${value}) * ${delta}`;
+    return `${value} * ${delta}`;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeUpgrade(value, delta, model, change) {
+    if (!value) return delta;
```
Fyorl, line 66:
```suggestion
    if ( !value ) return delta;
```

module/data/fields/formula-field.mjs
```diff
@@ -29,5 +29,54 @@ export default class FormulaField extends foundry.data.fields.StringField {
...
+    const terms = new Roll(value).terms;
+    if ( (terms.length === 1) && (terms[0].fn === "max") ) return current.replace(/\)$/, `, ${delta})`);
+    return `max(${value}, ${delta})`;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeDowngrade(value, delta, model, change) {
+    if (!value) return delta;
```
Fyorl, line 76:
```suggestion
    if ( !value ) return delta;
```

Fyorl (approved):

PR #4148, opened 2024-09-06
Add sources filtering to compendium browser.

Fyorl:
Some challenges here around filtering on a derived field. I also wasn't able to figure out how to use `compendiumBrowserFilters` without either: scanning all packs at startup in order to identify custom source fields; or only using pack metadata and then losing the ability to filter on custom source fields.

I think this approach ended up not being too messy, with only a small amount of special-casing that is maybe warranted for something as ubiquitous as the source field.

### TODO
- [ ] Provide setting for excluding compendium packs from compendium browser entirely.
- [ ] Update SRD to commit source field migrations so that NPC sources are available in the index.

### DIFF
--- a/module/applications/compendium-browser.mjs
+++ b/module/applications/compendium-browser.mjs
@@ -316,6 +316,12 @@ export default class CompendiumBrowser extends foundry.applications.api.Handleba
       { inplace: false }
     );
     filters.documentClass ??= "Item";
+    if ( filters.additional?.source ) {
+      filters.additional.source = Object.entries(filters.additional.source).reduce((obj, [k, v]) => {
+        obj[k.slugify({ strict: true })] = v;
+        return obj;
+      }, {});
+    }
     return filters;
   }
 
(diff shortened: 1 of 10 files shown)


module/applications/compendium-browser.mjs
```diff
@@ -1017,6 +1070,14 @@ export default class CompendiumBrowser extends foundry.applications.api.Handleba
         // Apply module art to the new index
         .then(index => game.dnd5e.moduleArt.apply(index)))
 
+        // Derive source values
+        .map(i => {
+          if ( !index ) return i;
```
arbron, line 1075: Because we do filtering on the index before fetching the full documents, this will prevent source filter from being used when fetching full documents.

```suggestion
```

arbron (changes requested):
So I'm noticing that some older content modules that still use the `source.custom` field for their whole source data are creating a whole bunch of entries:

Screenshot 2024-09-06 at 07 58 11

I'm wondering if it would be worth it to do a quick filtering step on sources to remove everything after `pg.` to avoid this issue.

The other option would be to ignore the `custom` field entirely, so source filter is only available for content that has moved to the separate fields.

arbron (approved):

PR #4159, opened 2024-09-06
[#2138] Refactor `rollRecharge` to use `BasicRoll` & new hooks

arbron:
The roll recharge method now makes use of the new configuration structure and rolling hooks to be consistant with other rolls in the system. It also gains a `dnd5e.postRollRecharge` hook that fires after the usage updates have been performed. A similar post hook has been added to `rollHitDice` so that all rolls that perform updates now have a `post*` hook after those updates are complete.

In order to support modifying the chat message flavor based on the result of the recharge roll, a `preCreate` callback has been added to `BasicRollMessageConfiguration` that is called right before the message is creating, allowing for any last-minute changes to the message. This hook will also be very helpful when attack rolls are hit by the roll refactor.

A small tweak has been made to the config and hooks to consistantly use `subject` for the Document performing the roll, rather than using `origin` in the initial config and a document-type specific key in the other hooks.

### DIFF
--- a/module/data/shared/uses-field.mjs
+++ b/module/data/shared/uses-field.mjs
@@ -157,62 +157,111 @@ export default class UsesField extends SchemaField {
   /**
    * Rolls a recharge test for an Item or Activity that uses the d6 recharge mechanic.
    * @this {Item5e|Activity}
-   * @returns {Promise<Roll|void>}
+   * @param {BasicRollProcessConfiguration} config   Configuration information for the roll.
+   * @param {BasicRollDialogConfiguration} dialog    Configuration for the roll dialog.
+   * @param {BasicRollMessageConfiguration} message  Configuration for the roll message.
+   * @returns {Promise<BasicRoll[]|void>}            The created Roll instances, or `null` if no die was rolled.
    */
-  static async rollRecharge() {
+  static async rollRecharge(config={}, dialog={}, message={}) {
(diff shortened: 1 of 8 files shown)


module/data/shared/uses-field.mjs
```diff
@@ -157,62 +157,111 @@ export default class UsesField extends SchemaField {
...
+    rollConfig.subject = this;
+
+    const dialogConfig = foundry.utils.mergeObject({ configure: false }, dialog);
+
+    const messageConfig = foundry.utils.mergeObject(({
+      create: true,
+      data: {
+        speaker: ChatMessage.getSpeaker({ actor: this.actor, token: this.actor.token })
+      },
+      preCreate: (rolls, rollConfig, messageConfig) => {
```
Fyorl, line 188: I'm having some trouble following this design. This logic is specific to rolling recharges, so it should exist in the `rollRecharge` method. I don't see the benefit of adding an extra layer of indirection by passing a callback to `BasicRoll` here.

```js
if ( messageConfig.create ) {
  messageConfig.create = false;
  const [roll] = await CONFIG.Dice.BasicRoll.build(rollConfig, dialogConfig, messageConfig);
  if ( roll.isSuccess ) {
    // Perform updates ...
    roll.toMessage({ flavor: /* ... */, speaker: /* ... */ });
  }
}
```

Fyorl (approved):

PR #4162, opened 2024-09-06
[#1539] Add identifier field to all items

arbron:
Adds `identifier` field to `ItemDescriptionTemplate` so it is added to all items. For classes, subclasses, backgrounds, and species it can still be edited from the details tab, but for other item types it is now available in the Configure Source app.

Source with Identifier

Also makes a small change to how identifiers are created to handle slashes in the middle of words. This avoids an issue where a name like "Blindness/Deafness" would result in an identifier of only one word (e.g. `blindnessdeafness`) and instead adds a dash in place of the slash (e.g. `blindness-deafness`).

### Todo
- [ ] Create migration to automatically create identifier for any item that doesn't have one
- [ ] Add identifiers to all content in the SRD

### DIFF
--- a/module/applications/source-config.mjs
+++ b/module/applications/source-config.mjs
@@ -56,6 +56,11 @@ export default class SourceConfig extends DocumentSheet5e {
       { value: "2024", label: game.i18n.localize("SETTINGS.DND5E.RULESVERSION.Modern") },
       { value: "2014", label: game.i18n.localize("SETTINGS.DND5E.RULESVERSION.Legacy") }
     ];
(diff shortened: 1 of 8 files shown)


Fyorl (approved):
Code looks good, but I find myself second-guessing the decision to put identifiers in the 'Configure Source' dialog. The identifier doesn't feel related to an Item's *source*, rather its *identity*, no?

But putting it in the details tab for every Item is maybe too much. It might give the impression that users need to fill it out carefully for every Item, when really they shouldn't have to worry about it for the most part.

Maybe in the absence of a better place, the source config is the best for now.

PR #4177, opened 2024-09-09
Add application for configuring compendium browser source exclusions.

Fyorl:
image


arbron (approved):
I'm wondering if we want to stick the source config application into a settings folder to avoid filling up the top-level applications folder if we get more settings apps.

PR #4181, opened 2024-09-09
[#4024] Add spellcasting ability as option for Check activity

arbron:
Adds "Spellcasting Ability" as one of the options for the check required, allowing something like "Counterspell" to automatically select the correct ability to roll.


Fyorl (approved):
Factoring out `spellcastingAbility` is nice

PR #4187, opened 2024-09-10
[#1539] Use migration to set identifier on items

arbron:
Sets the identifier for items using a one-time concrete migration. Avoided an in-memory migration because an empty identifier is considered valid.

Adds a `bypassVersionCheck` flag to migrations that allows skipping the check that prevents certain migrations from being applied to items updated in `4.0.0` or later. This is a flag to allow for re-migrating items that have already been migrated to `4.0`.

### DIFF
--- a/module/migration.mjs
+++ b/module/migration.mjs
@@ -1,8 +1,11 @@
 /**
- * Perform a system migration for the entire World, applying migrations for Actors, Items, and Compendium packs
+ * Perform a system migration for the entire World, applying migrations for Actors, Items, and Compendium packs.
+ * @param {object} [options={}]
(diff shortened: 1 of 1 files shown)


Fyorl (approved):
Let's just make sure this `bypassVersionCheck` flag isn't getting persisted. I can't see it being removed anywhere.

arbron:
> Let's just make sure this `bypassVersionCheck` flag isn't getting persisted. I can't see it being removed anywhere.

Probably a bit of confusing naming when we originally added those, but the `flags` object passed between migration methods isn't actually saved to document flags, but instead its values are set based on the flags on the document.

PR #4188, opened 2024-09-10
[#1539] Add identifiers & rules versions to all items

arbron:
Uses a migration to set the identifiers and rules versions on items throughout the SRD, and then tweaks a couple of identifiers to match those that will be used in the 2024 rules for forwards compatibility.

**Note:** The second commit contains just the identifier renames for easier comparison.


packs/_source/classfeatures/ranger/ranger-features/vanish.json
```diff
@@ -98,7 +98,7 @@
     "prerequisites": {
       "level": null
     },
-    "identifier": "vanish"
+    "identifier": "natures-veil"
```
Fyorl, line 101: What criteria are we using to determine if a feature has been renamed vs. replaced?

arbron: D&D Beyond's updated terminology list :)

https://www.dndbeyond.com/posts/1810-updates-in-the-players-handbook-2024#New_Terminology

Fyorl: OK nice, thank you

Fyorl (approved):

PR #4190, opened 2024-09-10
Add context menus to in-sheet Activities.

Fyorl:
Some additional polish for activities.

- Fires the Activity's context menu if you right-click an activity row inside the character sheet, rather than firing the parent Item's context menu.
- Always show activities in the expanded view, even if the Item only has one.
- Adds context menu to Items inside favourites.


arbron (approved):
Looks good, just a comment about locked compendiums.

PR #4191, opened 2024-09-10
[#2125] Only display features in table, simplify subclass tables

### LINKED ISSUE #2125: Class Summary Page feature requests
1 Should it validate that the subclasses added to the sheet have a matching Class identifier to the Class?

2 Include Optional advancements in the generated class table, currently they are not shown on the Class table but have a separate section and custom table at the bottom of the generated item details

3 The class table lacks identifying when a subclass feature is granted, perhaps once the subclass items are added it can state "[Subclass Header] Feature" in plain text in the table
image

4 Special Handling for Item Grant advancements with only Spell Items
Currently if you use the advancement system to grant Prepared spellcasters spells, the table and item details sections balloon quite a lot, perhaps showing an inline link for the advancement item in the class table or just plain text, then in the description section a table to show the links to the spells similar to the optional display (but without the description of the individual spell items)
Current:
image
Proposed:
image
image

Thanks

arbron:
- Filter out any non-feature items when building tables and feature lists to avoid pollution from spells
- Prevent features that couldn't be found from displaying in tables
- Don't display subclass rows if scale value is the same as the previous row and everything else is equal

Closes #2125

### DIFF
--- a/module/applications/journal/class-page-sheet.mjs
+++ b/module/applications/journal/class-page-sheet.mjs
@@ -185,9 +185,15 @@ export default class JournalClassPageSheet extends JournalPageSheet {
     if ( scaleValues.length ) cols.push({class: "scale", span: scaleValues.length});
     if ( spellProgression ) cols.push(...spellProgression.cols);
 
+    const prepareFeature = uuid => {
+      const index = fromUuidSync(uuid);
+      if ( index?.type !== "feat" ) return null;
+      return linkForUuid(uuid);
+    };
+
     const rows = [];
     for ( const level of Array.fromRange((CONFIG.DND5E.maxLevel - (initialLevel - 1)), initialLevel) ) {
-      const features = [];
+      let features = [];
       for ( const advancement of item.advancement.byLevel[level] ) {
         switch ( advancement.constructor.typeName ) {
           case "AbilityScoreImprovement":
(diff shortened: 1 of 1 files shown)


module/applications/journal/class-page-sheet.mjs
```diff
@@ -209,7 +216,14 @@ export default class JournalClassPageSheet extends JournalPageSheet {
       if ( spellCells ) cells.push(...spellCells);
 
       // Skip empty rows on subclasses
-      if ( (item.type === "subclass") && !features.length && !scaleValues.length && !spellCells ) continue;
+      if ( item.type === "subclass" ) {
```
krbz999, line 219: Should probably change the comment here.

module/applications/journal/class-page-sheet.mjs
```diff
@@ -209,7 +216,14 @@ export default class JournalClassPageSheet extends JournalPageSheet {
...
+        let displayRow = features.length || spellCells;
+        if ( rows.length ) displayRow ||= !rows.at(-1).some((cell, index) =>
+          (cell.class !== "scale") && (cell.content !== cells[index].content)
+        );
+        else if ( scaleValues.length ) displayRow ||= cells.filter(c => (c.class === "scale") && c.content).length > 0;
```
Fyorl, line 224:
```suggestion
        else if ( scaleValues.length ) displayRow ||= cells.filter(c => (c.class === "scale") && c.content).length;
```

PR #4198, opened 2024-09-11
append item rollData when recovering during rest

### LINKED ISSUE #4197: Use Item RollData for resource recovery instead of Actor RollData
If I understand correctly, the only difference between an Item's rollData and an Actor's is that the item also has access to its own relevant info. It would be nice to be able to put in `@item.uses.max / 2` as a recovery formula - right now it's limited to only Actor rollData.

roth-michael:
Closes #4197

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -2718,6 +2718,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     result.rolls ??= [];
     for ( const item of this.items ) {
       if ( foundry.utils.getType(item.system.recoverUses) !== "function" ) continue;
+      rollData.item = {...item.system};
       const { updates, rolls } = await item.system.recoverUses(recovery, rollData);
       if ( !foundry.utils.isEmpty(updates) ) {
         const updateTarget = result.updateItems.find(i => i._id === item.id);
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -2718,6 +2718,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     result.rolls ??= [];
     for ( const item of this.items ) {
       if ( foundry.utils.getType(item.system.recoverUses) !== "function" ) continue;
+      rollData.item = {...item.system};
```
krbz999, line 2721: You should at the very least use `item.getRollData()` rather than 'reconstruct' the method here.

roth-michael: The reason I did it this way was to avoid having to call `Actor#getRollData()` for every single item - instead only inserting the item-specific data each time.

krbz999: I imagine it won't matter, that's how it was before. It's also a one-off method, not during, say, data preparation.

roth-michael: Makes sense, and improves the readability. Done.

arbron (approved):

PR #4199, opened 2024-09-11
Allow activities to be added directly to favourites.

Fyorl:


less/v2/dark/forms.less
```diff
@@ -10,7 +10,11 @@
   }
 
   .form-group {
-    &.card { background: var(--dnd5e-background-10); }
+    &.card {
+      background: var(--dnd5e-background-10);
+      border-color: var(--color-fieldset-border);
+    }
```
Fyorl, line 16: Some unrelated style tweaks.

module/data/actor/character.mjs
```diff
@@ -16,10 +16,10 @@ const {
 
 /**
  * @typedef {object} ActorFavorites5e
- * @property {"effect"|"item"|"skill"|"slots"|"tool"} type  The favorite type.
+ * @property {"effect"|"item"|"skill"|"slots"|"tool"|"activity"} type  The favorite type.
```
arbron, line 19: Just to keep things in alphabetical order :)

```suggestion
 * @property {"activity"|"effect"|"item"|"skill"|"slots"|"tool"} type  The favorite type.
```

arbron (approved):

PR #4203, opened 2024-09-12
[#4135] Pre-create `MappingField` data before applying effects

### LINKED ISSUE #4135: Applying active effects in `MappingField` should create whole internal object
Right now if you use an active effect to add proficiency in a tool that player doesn't currently have on their sheet the whole sheet breaks because the AE only adds the `value` and not the other parts of the data structure that are expected. Some custom application code in `MappingField` should be able to handle creating the whole necessary data structure rather then just the keys specified in the AE.

arbron:
Fix bugs caused by applying an active effect targeting a value in and entry in `system.tools` when the actor doesn't currently have that entry. The bug was caused by the whole object not being created and other parts of the code expecting it to exist.

This solves the problem by checking if an active effect is trying to apply to an `MappingField` entry that doesn't exist yet, and if so it will first create the initial entry before continuing with the application process.

Closes #4135

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -1,4 +1,5 @@
 import FormulaField from "../data/fields/formula-field.mjs";
+import MappingField from "../data/fields/mapping-field.mjs";
 import { staticID } from "../utils.mjs";
 
 const { SetField, StringField } = foundry.data.fields;
@@ -140,6 +141,23 @@ export default class ActiveEffect5e extends ActiveEffect {
         else current.add(value);
       }
       return current;
+    }
+
+    // If attempting to apply active effect to empty MappingField entry, create it
+    if ( (current === undefined) && change.key.startsWith("system.") ) {
+      let mappingField = field;
+      const [, ...parts] = change.key.split(".");
+      let last;
+      while ( !(mappingField instanceof MappingField) ) {
+        last = parts.pop();
+        mappingField = model.system.schema.getField(parts.join("."));
+        if ( !mappingField || !parts.length ) break;
+      }
+      const keyPath = ["system", ...parts, last].join(".");
+      if ( mappingField && (foundry.utils.getProperty(model, keyPath) === undefined) ) {
+        const created = mappingField.model.initialize(mappingField.model.getInitialValue(), mappingField);
+        foundry.utils.setProperty(model, keyPath, created);
+      }
     }
 
     return super.applyField(model, change, field);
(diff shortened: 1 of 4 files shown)


arbron:
@Fyorl Marked this as `4.0.0`, but it isn't super urgent so if needed we can push it back to a point update.

module/documents/active-effect.mjs
```diff
@@ -142,6 +143,23 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+      while ( !(mappingField instanceof MappingField) ) {
+        last = parts.pop();
+        mappingField = model.system.schema.getField(parts.join("."));
+        if ( !mappingField || !parts.length ) break;
+      }
+      const keyPath = ["system", ...parts, last].join(".");
+      if ( mappingField && (foundry.utils.getProperty(model, keyPath) === undefined) ) {
+        const created = mappingField.model.initialize(mappingField.model.getInitialValue(), mappingField);
+        foundry.utils.setProperty(model, keyPath, created);
+      }
```
Fyorl, line 160: Any disadvantage to using the `parent` chain here instead? I think it's probably faster, which is important if it happens during preparation.

```suggestion
      let mappingField = field;
      while ( !(mappingField instanceof MappingField) && mappingField ) mappingField = mappingField.parent;
      if ( mappingField && (foundry.utils.getProperty(model, mappingField.fieldPath) === undefined) ) {
        const created = mappingField.model.initialize(mappingField.model.getInitialValue(), mappingField);
        foundry.utils.setProperty(model, mappingField.fieldPath, created);
      }
```

Fyorl (changes requested):

module/documents/active-effect.mjs
arbron: MappingField wasn't properly setting itself as the `parent` of its containing model.

module/documents/active-effect.mjs
```diff
@@ -142,6 +143,20 @@ export default class ActiveEffect5e extends ActiveEffect {
...
 
+    // If attempting to apply active effect to empty MappingField entry, create it
+    if ( (current === undefined) && change.key.startsWith("system.") ) {
+      let keyPath = change.key;
+      let mappingField = field;
+      while ( !(mappingField instanceof MappingField) && mappingField ) {
+        if ( mappingField.name ) keyPath = keyPath.substring(0, keyPath.length - mappingField.name.length - 1);
+        mappingField = mappingField.parent;
+      }
+      if ( mappingField && (foundry.utils.getProperty(model, keyPath) === undefined) ) {
```
Fyorl, line 154: I guess you must have tested it, but `mappingField.fieldPath` is not working correctly?

arbron: That gets the path to the `MappingField` (e.g. `system.tools`), but that will already exist. We need the path to the specific entry within the field (e.g. `system.tools.thief`), which isn't available from one of the `DataFields` because it isn't static.

Fyorl (approved):

Issue #4229, opened 2024-09-14
null error in rollAttackV2


gludington:
The jsdocs for rollAttackV2 specify that AmmunicationUpdate can be null

     * @param {AmmunitionUpdate|null} data.ammoUpdate  Any updates related to ammo consumption for this attack.

However, it is used without a null check in the hook:

      const oldAmmoUpdate = [{ _id: ammoUpdate.id, "system.quantity": ammoUpdate.quantity }];
      Hooks.callAll("dnd5e.rollAttack", this.item, roll, oldAmmoUpdate);
      if ( oldAmmoUpdate[0] ) {
        ammoUpdate.id = oldAmmoUpdate[0]._id;
        ammoUpdate.quantity = foundry.utils.getProperty(oldAmmoUpdate[0], "system.quantity");
      }
    }

This will break firing of the hook for consumers.  One way to take a look at this is to enable the Automated Animations module, and cast Firebolt.  This will call this hook with a null ammoUpdate argument, and the hook will break there.  If you open your debug tools and change ammoUpdate to an empty object, the hook will fire (and the animation display).

gludington:
referenced in PR here: https://github.com/foundryvtt/dnd5e/pull/4232

PR #4232, opened 2024-09-14
add null guard around ammoUpdate

gludington:
This null guard allows the hook to proceed even when calling the deprecated old hook, however, doing so reveals what appears to be a similar issue downstream in mixin.js, as on some subseuqent casts, in

_createDeprecatedConfigs(usageConfig, dialogConfig, messageConfig)

Uncaught (in promise) TypeError: usageConfig.consume?.resources.includes is not a function


gludington:
Closes https://github.com/foundryvtt/dnd5e/issues/4229

arbron (approved):

PR #4269, opened 2024-09-16
Re-shuffle languages based on 2024 rules. Add Common Sign Language.

Fyorl:


module/config.mjs
```diff
@@ -3244,23 +3244,26 @@ DND5E.languages = {
     label: "DND5E.LanguagesStandard",
     children: {
       common: "DND5E.LanguagesCommon",
+      draconic: "DND5E.LanguagesDraconic",
```
arbron, line 3247: Any movement of languages between categories is going to cause problems with existing trait advancements, because they store the full path: `languages:exotic:draconic`. So in this case anyone using the Modern Rules wouldn't be offered Draconic when taking the legacy Dragonborn. Same with Thieves' Cant and Druidic.

Maybe we could provide some sort of language remapping that `TraitAdvancement` uses.

arbron (approved):

PR #4281, opened 2024-09-17
Update Modifying-Your-Game-with-Scripts.md

hakr14:
Make step 3 clearer
Fix error in item properties script


arbron (changes requested):
It might also be helpful to display the whole changed `world.json` after step 4 to make it clear what it should look like after everything is done. Maybe remove `esmodules` from the first example because it won't be in most people's worlds.

arbron (approved):

PR #4299, opened 2024-09-17
[#4278, #4279, #4297] Fix concentration from activities

### LINKED ISSUE #4278: Deprecation warning when using non-spell concentration item
The `createConcentrationEffectData` calls `getEfffectDurationFromItem`, which tries to fetch from `system.duration`. This triggers a deprecation warning for all items but spells. These methods should be reworked to use activity data rather than item data.

Screenshot 2024-09-16 at 16 49 22

### LINKED ISSUE #4279: Using activity with concentration fails if item doesn't also require concentration
The `createConcentrationEffectData` method checks `item.requiresConcentration` when it should be checking the individual activity if it requires concentration.

### LINKED ISSUE #4297: Activities with concentration on items without concentration don't trigger concentration

arbron:
Changes a number of concentration implementation details to fully work with activities and fix issues with concentration on non-spell items or activities that offer concentration even if the base item does not.

Modified `beginConcentrating` and `createConcentrationEffectData` to accept an activity rather than an item, with a fallback that selects the first activity and shows a deprecation warning.

Deprecated `ActiveEffect5e#getEffectDurationFromItem` and replaced it with a `getEffectData` method on `DurationField` which serves the same purpose.

Modified concentration checks in the activity usage process to use `requiresConcentration` on the activity rather than the item so that activities that require concentration on items that do not still properly trigger.

Also tweaked the flags added to the concentration effect to include
activity and scaling data.

Closes #4278
Closes #4279
Closes #4297

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1168,12 +1168,21 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /**
    * Initiate concentration on an item.
-   * @param {Item5e} item                        The item on which to being concentration.
+   * @param {Activity} activity                  The activity on which to being concentration.
    * @param {object} [effectData]                Effect data to merge into the created effect.
    * @returns {Promise<ActiveEffect5e|void>}     A promise that resolves to the created effect.
    */
-  async beginConcentrating(item, effectData={}) {
-    effectData = ActiveEffect5e.createConcentrationEffectData(item, effectData);
+  async beginConcentrating(activity, effectData={}) {
+    if ( activity instanceof Item ) {
+      foundry.utils.logCompatibilityWarning(
+        "The `beginConcentrating` method on Actor5e now takes an Activity, rather than an Item.",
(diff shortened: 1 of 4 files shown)


module/documents/actor/actor.mjs
```diff
@@ -1182,9 +1191,10 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
      * @param {Actor5e} actor         The actor initiating concentration.
      * @param {Item5e} item           The item that will be concentrated on.
      * @param {object} effectData     Data used to create the ActiveEffect.
+     * @param {Activity} activity     The activity that triggered the concentration.
      * @returns {boolean}             Explicitly return false to prevent the effect from being created.
      */
-    if ( Hooks.call("dnd5e.preBeginConcentrating", this, item, effectData) === false ) return;
+    if ( Hooks.call("dnd5e.preBeginConcentrating", this, activity.item, effectData, activity) === false ) return;
```
Fyorl, line 1197: I guess we need a corresponding wiki update.

Fyorl (approved):

PR #4300, opened 2024-09-17
Some additional chat card polish.

Fyorl:
- Adds activation conditions indiscriminately now under a 'trigger' label, rather than it being hard-coded to reactions only.
- Re-adds the upcast level of the spell to the spell subtitle, something we accidentally regressed during 4.0.
- Adds more rules-accurate attack type information to attack roll cards.

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -234,7 +234,7 @@ export default Base => class extends PseudoDocumentMixin(Base) {
     }
 
     // Handle scaling
-    activity._prepareUsageScaling(usageConfig, messageConfig, item);
+    await activity._prepareUsageScaling(usageConfig, messageConfig, item);
     activity = item.system.activities.get(this.id);
 
     // Handle consumption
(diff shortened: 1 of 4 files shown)


module/documents/activity/mixin.mjs
```diff
@@ -593,7 +593,7 @@ export default Base => class extends PseudoDocumentMixin(Base) {
    * @param {Item5e} item                                 Clone of the item that contains this activity.
    * @protected
    */
-  _prepareUsageScaling(usageConfig, messageConfig, item) {
+  async _prepareUsageScaling(usageConfig, messageConfig, item) {
```
Fyorl, line 596: Noticed that `EnchantActivity` overrides this and makes it `async`, so I updated the signature to match, and made sure it was appropriately `await`ed in `Activity#use`.

arbron (approved):
It would also be nice to display the attack mode in damage card so players can know that the correct damage was rolled.

Fyorl:
> It would also be nice to display the attack mode in damage card so players can know that the correct damage was rolled.

We don't (yet) associate a given damage roll with an attack roll, so I don't think it's possible to know for sure which attack mode was used for what damage roll.

PR #4306, opened 2024-09-17
Fix effects tray appearing on attack and damage roll cards.

Fyorl:


module/documents/chat-message.mjs
```diff
@@ -583,6 +583,8 @@ export default class ChatMessage5e extends ChatMessage {
     } else {
       effects = item?.effects.filter(e => (e.type !== "enchantment")
         && !item.getFlag("dnd5e", "riders.effect")?.includes(e.id));
+      const type = this.getFlag("dnd5e", "roll.type");
+      if ( (type === "attack") || (type === "damage") ) return;
```
arbron, line 587: Since this code path is only intended to be used for legacy module items or items without activities, we can probably just filter out anything that has the `roll.type` flag set (otherwise it will also appear on Utility rolls).

```suggestion
      if ( this.getFlag("dnd5e", "roll.type") ) return;
      effects = item?.effects.filter(e => (e.type !== "enchantment")
        && !item.getFlag("dnd5e", "riders.effect")?.includes(e.id));
```

arbron (changes requested):

arbron (approved):

Issue #4310, opened 2024-09-17
Rebuild actor configuration applications using ApplicationV2 & dnd5e2 styles


arbron:
- [x] Ability Config (https://github.com/foundryvtt/dnd5e/pull/4311)
- [x] AC Config (https://github.com/foundryvtt/dnd5e/pull/4512)
- [x] Concentration Config (https://github.com/foundryvtt/dnd5e/pull/4320)
- [x] Damage Modification Config (https://github.com/foundryvtt/dnd5e/pull/4517)
- [x] Hit Dice Config (https://github.com/foundryvtt/dnd5e/pull/4508)
- [x] Hit Points Config (https://github.com/foundryvtt/dnd5e/pull/4508)
- [x] Skills Config (https://github.com/foundryvtt/dnd5e/pull/4450)
- [x] Spell Slots Config (https://github.com/foundryvtt/dnd5e/pull/4520)
- [x] Initiative Config (https://github.com/foundryvtt/dnd5e/pull/4516)
- [x] Movement Config (https://github.com/foundryvtt/dnd5e/pull/4521)
- [x] Senses Config (https://github.com/foundryvtt/dnd5e/pull/4521)
- [ ] Special Traits Config (https://github.com/foundryvtt/dnd5e/pull/4545)
- [x] Proficiency Config (https://github.com/foundryvtt/dnd5e/pull/4328)
- [x] Trait Selector (https://github.com/foundryvtt/dnd5e/pull/4434)
- [x] Tool Selector (https://github.com/foundryvtt/dnd5e/pull/4450)
- [x] Type Config (https://github.com/foundryvtt/dnd5e/pull/4547)
- [x] Weapons Config

PR #4311, opened 2024-09-17
[#4310] Re-design ability config to use V2 app & styles

arbron:
Deprecates `ActorAbilityConfig` to replace it with an ApplicationV2 version with the new styles.

Ability Config V2

Also modifies `Application5e` and `DocumentSheet5e` to use a common mixin to ensure all features are available across both.


module/applications/actor/config/ability-config.mjs
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preparePartContext(partId, context, options) {
+    context = await super._preparePartContext(partId, context, options);
+    const source = this.document.system.toObject();
```
Fyorl, line 48: Something I've been trying to be more aware of, is that we can pass `this.document.system._source` directly (assuming we never mutate it, which I think is the case here), rather than calling `toObject`, which should make our rendering and re-rendering a bit faster.

module/applications/api/document-sheet.mjs
```diff
@@ -1,30 +1,13 @@
-const { DocumentSheetV2, HandlebarsApplicationMixin } = foundry.applications.api;
+import V2Mixin from "./v2-mixin.mjs";
+
+const { DocumentSheetV2 } = foundry.applications.api;
 
 /**
  * Base document sheet from which all document-based application should be based.
  */
-export default class DocumentSheet5e extends HandlebarsApplicationMixin(DocumentSheetV2) {
+export default class DocumentSheet5e extends V2Mixin(DocumentSheetV2) {
```
Fyorl, line 8: We unfortunately have overloaded the V2 term so it's not clear which 'V2' this refers to from just the name. Maybe we can call it something like `ApplicationV2Mixin`?

templates/actors/config/ability-config.hbs
```diff
@@ -0,0 +1,34 @@
+<section class="flexcol">
+    {{!-- Ability Score & Maxmimum --}}
+    <fieldset class="card">
+        <legend>{{ localize "DND5E.ABILITY.SECTIONS.Score" ability=ability localize=true }}</legend>
+        {{ formField fields.value name=(concat prefix "value") value=data.value localize=true }}
+        {{ formField fields.max name=(concat prefix "max") value=data.max placeholder="20" localize=true }}
```
Fyorl, line 6: Consider a split-group here with `current` and `max`.

Fyorl (approved):
Looks nice, I like it

templates/actors/config/ability-config.hbs
```diff
@@ -0,0 +1,41 @@
+<section class="flexcol">
+    {{!-- Ability Score & Maxmimum --}}
+    <fieldset class="card">
+        <legend>{{ localize "DND5E.ABILITY.SECTIONS.Score" ability=ability localize=true }}</legend>
+        <div class="form-group">
```
Fyorl, line 5: Missing `split-group` class

PR #4315, opened 2024-09-18
Add Magical status effect

dev7355608:
Adds the Magical status to mark active effects as magical. Effects without the Magical status would be considered nonmagical.

We'll need to update all effects in compendium packs and add the Magical statuses where the effect is magical before merging this in. Perhaps even add migration that adds the Magical status to existing effects that originate from a spell or magical item.


krbz999:
Might be easier to just set a flag on the effect during the enchanting/effect grant process, rather than having to edit every effect in every piece of content.

Since every item can have the `mgc` property, that can be used to check whether this flag should be added.

dev7355608:
The status effect has advantages over a flag:
- The user can configure what active effects a magical and which are not via UI.
- Whether a token is affected by magic can be tested efficiently by testing the actor's statuses, which is useful for detection modes like Detect Magic.

Also every spell effect being automatically magical is probably undesirable. A spell could have nonmagical effects: it could give you exhaustion or sever a limb for example.

krbz999:
> The user can configure what active effects a magical and which are not via UI.

Adding it to the UI would be desirable, but non-magical effects are the exception not the norm, so perhaps a simple boolean property for `nonmagical` could be added. The result would be that only a small handful of effects would need to be amended, rather than who knows how many, across the system and every content module, for Detect Magic to work properly.

Fyorl:
Do we need this anymore if we have `system.magical` now? Or is it complementary?

PR #4328, opened 2024-09-18
[#4310] Re-design skill & tool config to use V2 apps & styles

arbron:
Deprecates `ProficiencyConfig` and turns it into `SkillToolConfig` using the new styles. Re-factors out some code that can be shared between this new config and `AbilityConfig` into `BaseProficiencyConfig` to avoid repetition.

Screenshot 2024-09-18 at 16 25 10


module/applications/actor/config/skill-tool-config.mjs
```diff
@@ -0,0 +1,41 @@
...
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preparePartContext(partId, context, options) {
+    context = await super._preparePartContext(partId, context, options);
+    context.abilityOptions = Object.entries(CONFIG.DND5E.abilities).map(([value, { label }]) => ({ value, label }));
+    context.proficiencyOptions = Object.entries(CONFIG.DND5E.proficiencyLevels)
+      .map(([value, label]) => ({ value, label }));
+    context.section = `DND5E.${this.options.trait === "skills" ? "SKILL" : "TOOL"}.SECTIONS.`;
+    if ( this.options.trait === "tool" ) delete context.global;
```
Fyorl, line 38: Global ability check bonuses should still apply to tools.

templates/actors/config/skill-tool-config.hbs
```diff
@@ -0,0 +1,40 @@
...
+            <p class="hint">{{ localize (concat section "Bonuses.Hint") label=label }}</p>
+        </div>
+        {{#if fields.bonuses.fields.passive}}
+        {{ formField fields.bonuses.fields.passive name=(concat prefix "bonuses.passive") value=data.bonuses.passive
+                     localize=true }}
+        {{/if}}
+        {{ formField fields.bonuses.fields.check name=(concat prefix "bonuses.check") value=data.bonuses.check
+                     localize=true }}
+        <hr>
+        {{ formField fields.roll.fields.min name=(concat prefix "roll.min") type="number" value=data.roll.min
```
Fyorl, line 24: Also now that I see this here too, the idea of separating it into a separate fieldset with an appropriate legend (e.g. 'acrobatics roll') feels pretty good to me. And then we can make that consistent with the concentration dialog. What do you think?

templates/actors/config/skill-tool-config.hbs
```diff
@@ -0,0 +1,40 @@
...
+    </fieldset>
+
+    {{!-- Global Skill Bonuses --}}
+    {{#if global}}
+    <fieldset class="card">
+        <legend>{{ localize (concat section "Global.Label") }}</legend>
+        <div class="form-group">
+            <p class="hint">{{ localize (concat section "Global.Hint") }}</p>
+        </div>
+        {{ formField global.fields.skill value=global.data.skill localize=true }}
```
Fyorl, line 37: Global ability check bonuses should apply to skills too.

Fyorl (approved):

PR #4338, opened 2024-09-19
Allow removal of exhaustion condition

Larkinabout:
For the purposes of homebrew, allow the removal of the Exhaustion condition without breaking the character sheets.
remove-exhaustion-condition


module/applications/actor/character-sheet-2.mjs
```diff
@@ -96,19 +96,21 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
...
-      const filled = attributes.exhaustion >= n;
-      if ( filled ) classes.push("filled");
-      if ( n === max ) classes.push("death");
-      const pip = { n, label, filled, tooltip: label, classes: classes.join(" ") };
-
-      if ( n <= max / 2 ) acc.left.push(pip);
-      else acc.right.push(pip);
-      return acc;
-    }, { left: [], right: [] });
+    if (CONFIG.DND5E.conditionTypes.exhaustion) {
```
arbron, line 99:
```suggestion
    if ( CONFIG.DND5E.conditionTypes.exhaustion ) {
```

arbron (changes requested):

arbron (changes requested):
Just tested on the latest version and it seems to push the AC to the side with exhaustion disabled (maybe a result of Kim's recent CSS fixes for V13):

Screenshot 2024-10-02 at 09 50 21

Adjusting line `354` of `character.less` to `justify-content: center;` seems to fix it with no issues on the sheet when exhaustion is enabled.

Larkinabout:
Fixed the position when there are no pips, but also made some tweaks for when there are more than 6 pips. Not sure of the method, but it works.

## Previous
### 6 Pips
exhaustion-previous-6-pips

### 12 Pips
exhaustion-previous-12-pips

## New
### 6 Pips
exhaustion-new-6-pips

### 12 Pips
exhaustion-new-12-pips

### No Pips
exhaustion-new-no-pips

less/v2/character.less
```diff
@@ -414,6 +420,11 @@
               }
             }
           }
+
+          .lozenges.no-margin-top {
+            margin-top: 0px;
+          }
+
```
arbron, line 427: Would prefer to avoid adding a class just for this styling.

```suggestion
          .top:has([data-n="7"]) + .lozenges { margin-block-start: 0; }
```

arbron (changes requested):

arbron (approved):

PR #4344, opened 2024-09-19
Savage Attacks AE

### LINKED ISSUE #2913: Fix descriptive note on Savage Attacks (or add an Active Effect to the feature instead)
The SRD `Savage Attacks` feature currently says to "Enable this feature in the Special Traits section of the character sheet", but there is no longer a named "Savage Attacks" feature in there.

This should _either_ be changed to read "Add 1 to the Melee Critical Damage Dice field in the Special Traits section of the character sheet", _or_ an Active Effect should be added to this feature item that targets `flags.dnd5e.meleeCriticalDamageDice` (and adds `1`).

CleanShot 2024-02-04 at 21 23 17

krbz999:
Closes #2913.

### DIFF
--- a/packs/_source/races/orc/orc-features/savage-attacks.json
+++ b/packs/_source/races/orc/orc-features/savage-attacks.json
@@ -7,7 +7,7 @@
   "type": "feat",
   "system": {
     "description": {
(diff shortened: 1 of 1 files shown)


arbron:
@krbz999 Can you comment on the original issue so I can assign it?

arbron (approved):

Issue #4408, opened 2024-09-23
Add rollData to journal pages to allow for referencing flags for DCs


MrVauxs:
Currently the ability to make variable DCs is not possible to make in journal pages. 
Attempting to make `[[/damage ]]`, `[[/save ]]`, or `[[/check ]]` tags with, for example `dc=@flags.module.dc`, results in the variable being ignored and defaulted to `0` by [`simplifyBonus`](https://github.com/foundryvtt/dnd5e/blob/3e9da363d92825175df22c319074274870d10309/module/enrichers.mjs#L235) because the `config.rollData` provided to it by the Enricher is `undefined` when rendering on a journal page.

Adding this functionality would allow for creation of multi-level adventures, such as those from 3rd-party creators like Loot Tavern (primarily Heliana's Hunts modules). 

Currently the alternatives are to either provide the enriched rolls for each level, making the text significantly larger for no purpose, or to use code to either modify the journal data to be the chosen DC or modifying the end-result HTML of the enricher.

PR #4419, opened 2024-09-25
Enricher access to select data from document

MrVauxs:
Implements #4408

This is a prototype / proof of concept PR for adding the ability to reference data, specifically flags, that originate specifically from the document the enricher is being rendered "on", as opposed to relying entirely on `getRollData`.

The naming has been chosen purely to avoid colliding with any other properties. This would probably be best wrapped into `getRollData` function itself, but that would also require adding it to documents such as Journals as this is the primary intent of this PR.

Here is an example of what such a feature would allow for premium adventure modules to do:

https://github.com/user-attachments/assets/84eb78b3-85a9-463b-9d13-f44f1bc0ff1f

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -396,8 +396,6 @@ async function enrichDamage(config, label, options) {
     }
   });
 
-  console.log(config)
-
   config.formula = Roll.defaultImplementation.replaceFormulaData(formulaParts.join(" "), options.rollData ?? {});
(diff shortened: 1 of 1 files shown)


arbron:
I've posted another PR that handles this using a custom `JournalEntryPage` type, which allows it to work on other enricher types (e.g. `[[/r @page.flags.my-module.blah]]` will work):

https://github.com/foundryvtt/dnd5e/pull/4476

PR #4434, opened 2024-09-26
[#4310] Re-design damage, armor, language configs to use V2 apps

arbron:
Deprecates `TraitSelector` and creates new `TraitsConfig` and `DamagesConfig` apps. Also refactors `WeaponsConfig` to make use of a lot of shared code between all of these apps.

Armor, Languages,   Resistances Config


less/v2/apps.less
```diff
@@ -1113,16 +1114,27 @@ dialog.dnd5e2.application {
...
-    .proficiency { grid-column: proficiency; }
-    .mastery { grid-column: mastery; }
-    .proficiency, .mastery { text-align: center; }
+    .proficiency {
+      grid-column: proficiency;
+      text-align: center;
+    }
     li {
       align-items: center;
       &:not(:last-of-type) { border-bottom: var(--dnd5e-border-dotted); }
```
Fyorl, line 1123: Some style tweaks for the sub-category display. Works pretty well on smaller lists like Primordial here, but might not look great on longer lists, not sure:

image

```suggestion
      &:not(:last-of-type) { border-bottom: var(--dnd5e-border-dotted); }
      &.category { gap: 4px; } 
      li {
        position: relative;
        margin: 0 0 4px 0;
        &::before {
          content: "┗";
          position: absolute;
          left: -1.8em;
          font-size: var(--font-size-10);
          font-weight: bold;
          top: 3px;
          color: var(--color-text-dark-5);
        }
      }
```

Fyorl (approved):

module/applications/actor/config/traits-config.mjs
```diff
@@ -0,0 +1,145 @@
...
+    return options;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preparePartContext(partId, context, options) {
+    context = await super._preparePartContext(partId, context, options);
+    context.keyPath = Trait.actorKeyPath(this.options.trait);
+    context.data = foundry.utils.getProperty(this.document.toObject(), context.keyPath);
```
Fyorl, line 67: Missed this before. Should be safe to do, we are not expecting `_processChoices` to mutate `data`, right?

```suggestion
    context.data = foundry.utils.getProperty(this.document._source, context.keyPath);
```

PR #4439, opened 2024-09-27
Update Lay On Hands to roll healing

AndrewBurian:
Allows the paladin's Lay On Hands feature to actually roll healing, and scaling to use more than one point from the pool at a time.

Also adds a second activity called "Cure Disease and Poison" that consumes 5 points at a time from the pool, and also allows scaling do cure more than one disease in a single action.

### DIFF
--- a/packs/_source/classfeatures/paladin/paladin-features/lay-on-hands.json
+++ b/packs/_source/classfeatures/paladin/paladin-features/lay-on-hands.json
@@ -57,7 +57,7 @@
           ],
           "scaling": {
             "allowed": true,
-            "max": ""
+            "max": "@item.uses.value"
           },
           "spellSlot": true
         },
@@ -143,7 +143,7 @@
           ],
           "scaling": {
             "allowed": true,
-            "max": ""
+            "max": "floor(@item.uses.value / 5)"
           },
           "spellSlot": true
         },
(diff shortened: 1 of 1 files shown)


packs/_source/classfeatures/paladin/paladin-features/lay-on-hands.json
```diff
@@ -50,13 +50,13 @@
...
-                "mode": "",
+                "mode": "amount",
                 "formula": ""
               }
             }
           ],
           "scaling": {
-            "allowed": false,
+            "allowed": true,
             "max": ""
```
arbron, line 60: You should set a max scaling value here so it can display the slider rather than the text input.

Screenshot 2024-09-30 at 16 13 04

```suggestion
            "max": "@item.uses.value"
```

packs/_source/classfeatures/paladin/paladin-features/lay-on-hands.json
```diff
@@ -93,29 +93,100 @@
...
+              "value": "5",
+              "scaling": {
+                "mode": "amount",
+                "formula": "5"
+              }
+            }
+          ],
+          "scaling": {
+            "allowed": true,
+            "max": ""
```
arbron, line 146: Scaling max here also.

```suggestion
            "max": "floor(@item.uses.value / 5)"
```

arbron (changes requested):

arbron (approved):

PR #4460, opened 2024-09-27
[#2094] Move item socket event handlers into type data models

arbron:
Moved various `_pre` and `_on` socket event handlers into item data models rather than on the item. All socket handlers have been removed from items except setting the initial identifier during `_preCreate`, performing activity rider updates in `_preUpdate` (though the implementation has been moved), and ending concentration within `_onDelete`.

### DIFF
--- a/module/data/item/templates/equippable-item.mjs
+++ b/module/data/item/templates/equippable-item.mjs
@@ -126,7 +126,7 @@ export default class EquippableItemTemplate extends SystemDataModel {
   preCreateEquipped(data, options, user) {
     if ( ["character", "npc"].includes(this.parent.actor?.type)
       && !foundry.utils.hasProperty(data, "system.equipped") ) {
-      this.parent.updateSource({ "system.equipped": this.parent.actor.type === "npc" });
+      this.updateSource({ equipped: this.parent.actor.type === "npc" });
     }
   }
 }
--- a/module/data/item/templates/equippable-item.mjs
+++ b/module/data/item/templates/equippable-item.mjs
@@ -112,4 +112,21 @@ export default class EquippableItemTemplate extends SystemDataModel {
     const attunement = this.attuned || (this.attunement !== "required");
     return attunement && this.properties.has("mgc") && this.validProperties.has("mgc");
   }
+
+  /* -------------------------------------------- */
+  /*  Socket Event Handlers                       */
+  /* -------------------------------------------- */
+
+  /**
+   * Set as equipped for NPCs, and unequipped for PCs.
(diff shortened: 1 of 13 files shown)


module/data/item/templates/equippable-item.mjs
```diff
@@ -112,4 +112,21 @@ export default class EquippableItemTemplate extends SystemDataModel {
...
+  /**
+   * Set as equipped for NPCs, and unequipped for PCs.
+   * @param {object} data     The initial data object provided to the document creation request.
+   * @param {object} options  Additional options which modify the creation request.
+   * @param {User} user       The User requesting the document creation.
+   */
+  preCreateEquipped(data, options, user) {
+    if ( ["character", "npc"].includes(this.parent.actor?.type)
+      && !foundry.utils.hasProperty(data, "system.equipped") ) {
+      this.parent.updateSource({ "system.equipped": this.parent.actor.type === "npc" });
```
Fyorl, line 129:
```suggestion
      this.updateSource({ "equipped": this.parent.actor.type === "npc" });
```

module/data/item/templates/physical-item.mjs
```diff
@@ -212,22 +212,35 @@ export default class PhysicalItemTemplate extends SystemDataModel {
 
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _preUpdate(changed, options, user) {
+    await super._preUpdate(changed, options, user);
```
Fyorl, line 217: Should this check for `false` return?

module/data/item/templates/physical-item.mjs
```diff
@@ -212,22 +212,35 @@ export default class PhysicalItemTemplate extends SystemDataModel {
...
     this._renderContainers({ formerContainer: options.formerContainer });
   }
 
   /* -------------------------------------------- */
 
   /** @inheritDoc */
   _onDelete(options, userId) {
+    super._onDelete(options, userId);
     this._renderContainers();
   }
```
Fyorl, line 245: Some of these were `async` for sub-types. Is it different for the template mixins? We don't necessarily have to `await` the `super` call, especially if this is render stuff, but we should make sure the signature at least matches if we are doing `@inheritDoc`.

Fyorl (approved):

module/data/item/templates/physical-item.mjs
arbron: The signature for these methods isn't `async` in `TypeDataModel`, so that's what I've been going off of. I think most of the `async` `_on*` method can sync instead.

Issue #4061, opened 2024-08-28
Activity Type: Cast a Spell


arbron:
Activity type that triggers a spell for items like Staffs and Wands.

Rabulias:
Would this also apply to _instruments of the bards_?

arbron:
> Would this also apply to _instruments of the bards_?

This would apply to any item that casts spells.

Rabulias:
Staves and instruments of the bards use the wielder's spell save DC instead of a set DC on the item. I got mixed up thinking wands did the same, and thought this issue was focused on that.

Zanderaf:
Any preview on how the spells are going to be displayed? I feel like having them show up in the spellbook similarly to how modules like Magic Items or Items with Spells does it feels the best imo; just having all the spells in the "same place".

roth-michael:
Noticed in #4474 you have "Currently does nothing to change how these spells appear on the actor sheet, so they will all display in the spell list as if they were normal spells" with "Hide cached spells on actor sheet" as a Todo. Wondering if it'd make sense to make use of that behavior, and maybe have a togglable option (either a system setting or per-cast-activity) on whether to show the spell in the spellbook, perhaps under its own "Granted by item" category. Personally I'd prefer to keep spells granted by items on the item, not have them show up in my spellbook, but I know there are many who feel the opposite.

Zanderaf:
Yea, I would very much appreciate spells granted by items in their own category instead of being mixed in with other spells.

PR #4474, opened 2024-09-30
[#4061] Add spell caching system to Cast activities

arbron:
Adds a system for Cast activities to automatically import a copy with changes applied when added to the actor, keep the cached spell updated when the activities change, and remove the cached copy when the item is removed.

Uses an enchantment effect on the spell to apply the actualy changes needed for easy updating.

Currently does nothing to change how these spells appear on the actor sheet, so they will all display in the spell list as if they were normal spells.

In order to better support multiple spells on an actor sheet with the same compendium source, `Actor#sourcedItems` has been changed to store sets of items, rather than just a single item.

### Todo
- [x] Handle ignored properties
- [ ] Hide cached spells on actor sheet
- [x] Apply consumption changes to cached spells based on activity
- [x] Activate cached spell with Cast activity is used

### DIFF
--- a/module/data/item/templates/activities.mjs
+++ b/module/data/item/templates/activities.mjs
@@ -412,7 +412,7 @@ export default class ActivitiesTemplate extends SystemDataModel {
     const cachedInserts = [];
     for ( const id of Object.keys(changed.system.activities) ) {
       const activity = this.activities.get(id);
-      if ( !activity ) continue;
+      if ( !(activity instanceof CastActivity) ) continue;
       const existingSpell = activity.cachedSpell;
       if ( existingSpell ) {
         const enchantment = existingSpell.effects.get(CastActivity.ENCHANTMENT_ID);
--- a/module/data/item/templates/activities.mjs
+++ b/module/data/item/templates/activities.mjs
@@ -357,7 +357,7 @@ export default class ActivitiesTemplate extends SystemDataModel {
 
   /**
    * Prepare any item or actor changes based on activity changes.
-   * @param {object} changed  The differential data that is changed relative to the documents prior values.
+   * @param {object} changed  The differential data that is changed relative to the document's prior values.
    * @param {object} options  Additional options which modify the update request.
    * @param {User} user       The User requesting the document update.
    */
@@ -395,7 +395,7 @@ export default class ActivitiesTemplate extends SystemDataModel {
 
   /**
    * Perform any additional updates when an item with activities is updated.
-   * @param {object} changed  The differential data that is changed relative to the documents prior values.
+   * @param {object} changed  The differential data that is changed relative to the document's prior values.
    * @param {object} options  Additional options which modify the update request.
    * @param {string} userId   The id of the User requesting the document update.
    */
--- a/module/data/item/templates/activities.mjs
+++ b/module/data/item/templates/activities.mjs
@@ -1,3 +1,4 @@
+import CastActivity from "../../../documents/activity/cast.mjs";
 import SystemDataModel from "../../abstract.mjs";
 import { ActivitiesField } from "../../fields/activities-field.mjs";
 import UsesField from "../../shared/uses-field.mjs";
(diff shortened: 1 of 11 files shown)


module/data/item/templates/activities.mjs
```diff
@@ -333,6 +334,109 @@ export default class ActivitiesTemplate extends SystemDataModel {
...
+    const spells = (await Promise.all(this.activities.getByType("cast").map(a => a.getCachedSpellData())))
+      .filter(_ => _);
+    if ( spells.length ) this.parent.actor.createEmbeddedDocuments("Item", spells);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Prepare any item or actor changes based on activity changes.
+   * @param {object} changed  The differential data that is changed relative to the documents prior values.
```
Fyorl, line 360:
```suggestion
   * @param {object} changed  The differential data that is changed relative to the document's prior values.
```

module/data/item/templates/activities.mjs
```diff
@@ -333,6 +334,109 @@ export default class ActivitiesTemplate extends SystemDataModel {
...
+      return null;
+    }).filter(_ => _);
+    if ( removed.length ) foundry.utils.setProperty(options, "dnd5e.removedCachedItems", removed);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Perform any additional updates when an item with activities is updated.
+   * @param {object} changed  The differential data that is changed relative to the documents prior values.
```
Fyorl, line 398:
```suggestion
   * @param {object} changed  The differential data that is changed relative to the document's prior values.
```

module/documents/activity/cast.mjs
```diff
@@ -24,4 +32,69 @@ export default class CastActivity extends ActivityMixin(CastActivityData) {
...
+   * @returns {object}
+   */
+  getSpellChanges() {
+    const changes = [];
+    for ( const type of ["activation", "duration", "range", "target"] ) {
+      if ( !this[type].override ) continue;
+      const data = this.toObject()[type];
+      delete data.override;
+      changes.push({ key: `system.${type}`, mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE, value: JSON.stringify(data) });
+    }
```
Fyorl, line 94:
```suggestion
    const changes = [];
    const source = this.toObject();
    for ( const type of ["activation", "duration", "range", "target"] ) {
      if ( !this[type].override ) continue;
      const data = source[type];
      delete data.override;
      changes.push({ key: `system.${type}`, mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE, value: JSON.stringify(data) });
    }
```

module/documents/activity/cast.mjs
```diff
@@ -24,4 +32,69 @@ export default class CastActivity extends ActivityMixin(CastActivityData) {
...
+    for ( const property of this.spell.properties ) {
+      changes.push({ key: "system.properties", mode: CONST.ACTIVE_EFFECT_MODES.ADD, value: `-${property}` });
```
Fyorl, line 96: Probably shouldn't remove concentration. There are some exceptions, but the majority of items still require the caster to concentrate on the spell after casting it from the item. We could hard-code removing VSM only (or non-'`isTag`' components), but Spell Scrolls, for example, only remove Material components, not others.

module/documents/actor/actor.mjs
```diff
@@ -3657,3 +3657,28 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     return this.update({ "system.favorites": favorites });
   }
 }
+
+/* -------------------------------------------- */
+
+class SourcedItemsMap extends Map {
```
Fyorl, line 3663:
```suggestion
/**
 * @extends {Map<string, Set<Item5e>>}
 */
class SourcedItemsMap extends Map {
```

Fyorl (changes requested):

Fyorl (commented):
Good stuff, thank you Jeff, see my comments above, but they're all pretty minor

module/data/item/templates/activities.mjs
arbron: That's what I get for grabbing JSDocs straight from core :P

module/documents/activity/cast.mjs
arbron: These aren't hard-coded removals, only what is explicitly set in the `Cast` activity.

Fyorl (approved):

module/data/item/templates/activities.mjs
Fyorl: (:

PR #4484, opened 2024-10-01
[#2138] Re-implement `rollAbilityTest` & `rollAbilitySave`

arbron:
Re-implements the method for rolling ability checks and saves for the re-factored `D20Roll` process. Since these two methods were already very similar, this merges their two implementations into a single private method while maintaining their public interface.

To simplify some of the roll preparation, this also introduces the `buildRoll` helper method that takes all the various potential roll parts and the initial roll data, adding values to the final parts array only if the value actually exists. This helps avoid needing a bunch of `if` statements for each individual roll part.


module/dice/build-roll.mjs
```diff
@@ -0,0 +1,17 @@
+/**
+ * Construct roll parts and populate its data object.
+ * @param {object} parts   Information on the parts to be constructed.
+ * @param {object} [data]  Roll data to use and populate while constructing the parts.
+ * @returns {{ parts: string[], data: object }}
+ */
+export default function buildRoll(parts, data={}) {
```
Fyorl, line 7: Could be a static helper on `BasicRoll`. `buildRoll` feels not quite right since we're not returning a `Roll`. Since we're mutating the roll data, maybe something like `applyParts`?

module/documents/activity/check.mjs
```diff
@@ -110,7 +110,9 @@ export default class CheckActivity extends ActivityMixin(CheckActivityData) {
         }
         await token.actor.rollToolCheck(tool, checkData);
       } else {
-        await token.actor.rollAbilityTest(ability, data);
+        await token.actor.rollAbilityTest(
+          { ability, target: data.targetValue }, {}, { data: { speaker: data.speaker } }
```
Fyorl, line 114: Do we need to propagate the `event` also?

module/documents/activity/check.mjs
```diff
@@ -110,7 +110,9 @@ export default class CheckActivity extends ActivityMixin(CheckActivityData) {
...
+        );
```
Fyorl, line 115:
```suggestion
        await token.actor.rollAbilityTest({ ability, target: data.targetValue }, {}, { data: { speaker: data.speaker } });
```

module/dice/basic-roll.mjs
```diff
@@ -75,6 +77,8 @@ export default class BasicRoll extends Roll {
    */
   static fromConfig(config, process) {
     const formula = (config.parts ?? []).join(" + ");
+    config.options ??= {};
+    config.options.target ??= process.target;
```
Fyorl, line 81: Are we passing `target` as a `process` parameter somewhere? Should we not be? It seems it's not documented in `BasicRollProcessConfiguration`.

module/documents/activity/save.mjs
```diff
@@ -107,9 +107,15 @@ export default class SaveActivity extends ActivityMixin(SaveActivityData) {
...
-      });
+      await token.actor.rollAbilitySave(
+        {
+          event,
+          ability: target.dataset.ability ?? this.save.ability,
+          target: Number.isFinite(dc) ? dc : this.save.dc.value
+        },
+        {},
+        { data: { speaker } }
+      );
```
Fyorl, line 118:
```suggestion
      await token.actor.rollAbilitySave({
        event,
        ability: target.dataset.ability ?? this.save.ability,
        target: Number.isFinite(dc) ? dc : this.save.dc.value
      }, {}, { data: { speaker } });
```

module/data/shared/uses-field.mjs
```diff
@@ -178,6 +178,7 @@ export default class UsesField extends SchemaField {
         }
       }]
     }, config);
+    rollConfig.hookNames = ["recharge"].concat(config.hooksNames ?? []);
```
Fyorl, line 181:
```suggestion
    rollConfig.hookNames = ["recharge"].concat(config.hookNames ?? []);
```

module/documents/activity/utility.mjs
```diff
@@ -65,8 +65,9 @@ export default class UtilityActivity extends ActivityMixin(UtilityActivityData)
     }
 
     const rollConfig = foundry.utils.deepClone(config);
-    rollConfig.subject = this;
+    rollConfig.hookNames = ["formula"].concat(config.hooksNames ?? []);
```
Fyorl, line 68:
```suggestion
    rollConfig.hookNames = ["formula"].concat(config.hookNames ?? []);
```

module/documents/actor/actor.mjs
```diff
@@ -1550,22 +1552,25 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-   * @param {string} abilityId    The ability id (e.g. "str")
-   * @param {object} options      Options which configure how ability tests or saving throws are rolled
+   * @param {AbilityRollProcessConfiguration} config  Configuration information for the roll.
+   * @param {BasicRollDialogConfiguration} dialog     Configuration for the roll dialog.
+   * @param {BasicRollMessageConfiguration} message   Configuration for the roll message.
    */
-  rollAbility(abilityId, options={}) {
+  rollAbility(config={}, dialog={}, message={}) {
+    let abilityId = config;
+    if ( foundry.utils.getType(config) === "Object" ) abilityId = config.ability;
```
Fyorl, line 1561: Should we emit a deprecation warning for this method's signature change?

module/documents/actor/actor.mjs
```diff
@@ -1574,171 +1579,140 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-     * @memberof hookEvents
-     * @param {Actor5e} actor                Actor for which the ability test is being rolled.
-     * @param {D20RollConfiguration} config  Configuration data for the pending roll.
-     * @param {string} abilityId             ID of the ability being rolled as defined in `DND5E.abilities`.
-     * @returns {boolean}                    Explicitly return `false` to prevent ability test from being rolled.
-     */
-    if ( Hooks.call("dnd5e.preRollAbilityTest", this, rollData, abilityId) === false ) return;
-
-    const roll = await d20Roll(rollData);
+  async rollAbilityTest(config={}, dialog={}, message={}) {
```
Fyorl, line 1588: Maybe we should taking this opportunity to deprecate `rollAbilityTest` in favour of `rollAbilityCheck`, so that it better matches the game term. Similarly with `rollAbilitySave` -> `rollSavingThrow`, with the assumption that we will eventually re-work the pipeline to match the game rules, i.e.:

```
rollSkillCheck -> rollAbilityCheck -> rollD20Test
rollDeathSave/rollConcentration -> rollSavingThrow -> rollD20Test
```

Then we could rename the private `#rollAbility` to `#rollAbilityTest` in order to align with the 'D20 Test' term, without it being confused with the public `rollAbilityTest` (after the deprecation period that is).

Maybe it's too early for that though. I just wondered what your thoughts were.

module/documents/actor/actor.mjs
```diff
@@ -2143,8 +2117,9 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
     formula ??= `max(0, 1${config.denomination} + @abilities.con.mod)`;
     const rollConfig = foundry.utils.deepClone(config);
-    rollConfig.subject = this;
+    rollConfig.hookNames = ["hitDie"].concat(config.hooksNames ?? []);
```
Fyorl, line 2120:
```suggestion
    rollConfig.hookNames = ["hitDie"].concat(config.hookNames ?? []);
```

Fyorl (changes requested):
Thanks for taking care of the deprecations, I know it's a slog

module/dice/basic-roll.mjs
arbron: Should be part of the typedef for process:

```javascript
/**
 * Configuration data for the process of creating one or more basic rolls.
 *
 * @typedef {object} BasicRollProcessConfiguration
 * @property {BasicRollConfiguration[]} rolls  Configuration data for individual rolls.
 * @property {Event} [event]                   Event that triggered the rolls.
 * @property {string[]} [hookNames]            Name suffixes for configuration hooks called.
 * @property {Document} [subject]              Document that initiated this roll.
 * @property {number} [target]                 Default target value for all rolls.
 */
```

module/documents/actor/actor.mjs
arbron: Yeah, we can do that. I'll name the private method just `#rollD20Test`, since it is designed to work even without an ability specified.

Fyorl: Yeah that sounds good. Attack rolls are also D20 Tests though, so we should use that name only so long as we plan that attack rolls will *eventually* go through this method and/or fire a `rollD20Test` hook some other way.

Fyorl (approved):

module/documents/actor/actor.mjs
arbron: They should definitely call the hook, but they might not be able to use the same method due to some of the roll preparation complications. I'll see when I get there.

PR #4487, opened 2024-10-02
[#4287] Spell Scroll Icon

### LINKED ISSUE #4287: Make auto-created spell scrolls utilize a scroll icon rather than a spell icon
Currently, when spell scrolls are created by dragging a spell to the inventory, the generated item depicts the icon of the spell. I would suggest instead; getting the icon from the SRD scroll of the appropriate level and using that.

Whilst I can see that the current method does provide some visual indication of what the scroll contains, my expectation when browsing the inventory is to see equipment icons. When looking for scrolls, I expect to see scrolls, rather than magic effects etc. This is the same philosophy I have attempted to implement across other inventory items.

Current:
image

Suggested:
image

krbz999:
Closes #4287.


arbron (approved):

arbron:
Please comment on the original issue :)

PR #4493, opened 2024-10-02
[#4489] AC cover calculation

### LINKED ISSUE #4489: [Feature Request] Add Active Effects to Half and Three-Quarters Cover
Currently applying the 'Half Cover' and 'Three-Quarters Cover' only applies a status icon but still requires manual input of -2/-5 to an attack or manual adjustment of the AC.

It would be nice for those two statuses to apply the AC bonuses automatically while the status is applied to an actor, which will help avoid players and GMs forgetting to add a situational bonus of -2/-5 to the roll dialog and keeping the visuals for hits/miss in the chat cards consistent to the AC of an actor.

krbz999:
During AC preparation, adds +5 to AC if three-quarters cover, else +2 to AC if half cover, else 0.

If a target has full cover during an attack roll, shows their ac as infinity.

image

Closes #4489.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -572,6 +572,9 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       ac.equippedShield = shields[0];
     }
 
+    // Compute cover.
+    ac.cover = this.statuses.has("coverThreeQuarters") ? 5 : this.statuses.has("coverHalf") ? 2 : 0;
+
     // Compute total AC and return
     ac.min = simplifyBonus(ac.min, rollData);
     ac.bonus = simplifyBonus(ac.bonus, rollData);
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -71,6 +71,19 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
   /* -------------------------------------------- */
 
   /**
+   * Calculate the bonus from any cover the actor is affected by.
+   * @type {number}     The cover bonus to AC and dexterity saving throws.
+   */
+  get coverBonus() {
+    const { coverHalf, coverThreeQuarters } = CONFIG.DND5E.statusEffects;
+    if ( this.statuses.has("coverThreeQuarters") ) return coverThreeQuarters?.coverBonus;
+    else if ( this.statuses.has("coverHalf") ) return coverHalf?.coverBonus;
+    return 0;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
    * Get all classes which have spellcasting ability.
    * @type {Record<string, Item5e>}
    */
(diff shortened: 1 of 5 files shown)


module/documents/chat-message.mjs
```diff
@@ -415,10 +415,10 @@ export default class ChatMessage5e extends ChatMessage {
         <li data-uuid="${uuid}" class="target ${isMiss ? "miss" : "hit"}">
           <i class="fas ${isMiss ? "fa-times" : "fa-check"}"></i>
           <div class="name">${name}</div>
-          ${ac ? `
+          ${(ac !== "") ? `
           <div class="ac">
             <i class="fas fa-shield-halved"></i>
-            <span>${ac}</span>
+            <span>${(ac === null) ? "<i class='fa-solid fa-infinity'></i>" : ac}</span>
```
arbron, line 421: Using the infinity letter rather than the font awesome would probably look more correctly sized:

Screenshot 2024-10-02 at 10 03 17
Screenshot 2024-10-02 at 10 03 07


```suggestion
            <span>${(ac === null) ? "&infin;" : ac}</span>
```

module/documents/actor/actor.mjs
```diff
@@ -572,6 +572,9 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       ac.equippedShield = shields[0];
     }
 
+    // Compute cover.
+    ac.cover = this.statuses.has("coverThreeQuarters") ? 5 : this.statuses.has("coverHalf") ? 2 : 0;
```
arbron, line 576: This will complete squash any active effects targeting cover, so we might want to do an upgrade rather than just replacement.

```suggestion
    ac.cover = Math.max(ac.cover, this.statuses.has("coverThreeQuarters") ? 5 : this.statuses.has("coverHalf") ? 2 : 0);
```

arbron (changes requested):
We should probably also add to dexterity saving throws.

module/documents/actor/actor.mjs
```diff
@@ -1669,6 +1685,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       data.saveBonus = Roll.replaceFormulaData(globalBonuses.save, data);
     }
 
+    // Include cover in dexterity saving throws
+    if ( abilityId === "dex" && data.attributes?.ac?.cover ) {
```
arbron, line 1689:
```suggestion
    if ( (abilityId === "dex") && data.attributes?.ac?.cover ) {
```

module/documents/actor/actor.mjs
```diff
@@ -277,6 +277,19 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
   /* -------------------------------------------- */
 
+  /**
+   * Calculate the bonus from any cover the actor is affected by.
+   * @returns {number}      The cover bonus to AC and dexterity saving throws.
+   */
+  getCoverBonus() {
+    const { coverHalf, coverThreeQuarters } = CONFIG.DND5E.statusEffects;
+    if ( this.statuses.has("coverThreeQuarters") ) return coverThreeQuarters.coverBonus;
+    else if ( this.statuses.has("coverHalf") ) return coverHalf.coverBonus;
```
arbron, line 287: Just in case someone removes these conditions.

```suggestion
    if ( this.statuses.has("coverThreeQuarters") ) return coverThreeQuarters?.coverBonus;
    else if ( this.statuses.has("coverHalf") ) return coverHalf?.coverBonus;
```

module/documents/actor/actor.mjs
```diff
@@ -277,6 +277,19 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /* -------------------------------------------- */
 
+  /**
+   * Calculate the bonus from any cover the actor is affected by.
+   * @returns {number}      The cover bonus to AC and dexterity saving throws.
+   */
+  getCoverBonus() {
```
arbron, line 284: Since we aren't doing anything complex here, it is probably better as just a standard getter up in the properties section.

```suggestion
  get coverBonus() {
```

arbron (changes requested):

arbron (approved):

arbron:
@krbz999 Comment on the original issue if you don't mind

PR #4496, opened 2024-10-03
[#4248] Set DamageRoll's options.appearance pre-evaluation

### LINKED ISSUE #4248: Adding support for Dice So Nice damage type
Hello,

I've received multiple requests on Dice So Nice repository since dnd5e v3 and the change done to damage type from the core "flavor" feature to system-specific damageroll.options.type. People could previously see dice skins based on the type of damage dealt (fire, ice, etc).
This feature was enabled by default but could be turned off.

I'm opening this ticket for these users and providing insider info if you want to add this feature.
If not, it can be the base for a tiny module someone else could develop.

Dice So Nice will use any info under roll.options.appearance or diceterm.options.appearance to override current user dice appearance.

This means doing this somewhere in the dnd5e code:

```js
mydamageroll.options.appearance = {colorset: mydamageroll.options.type};
```

Corresponding documentation:  https://gitlab.com/riccisi/foundryvtt-dice-so-nice/-/wikis/API/Customization#changing-the-appearance-of-dice-in-a-roll-programmatically

However, please note that this feature could previously be turned off by disabling a Dice So Nice feature that looks at the dice favor value (like 1d8[fire]).
So doing this without having a corresponding user setting would prevent users from disabling the damage type appearance override.

nb: It was previously not possible to pass an appearance directly to a Roll object, only to individual DiceTerms. I've added this QOL feature in the newest version of Dice So Nice to simplify this kind of use-case.

Related DsN ticket: https://gitlab.com/riccisi/foundryvtt-dice-so-nice/-/issues/401

roth-michael:
closes #4248 

Seems straightforward enough. Only potential issues I can think of:
- Will this catch _all_ damage rolls? I'm not intimately familiar with the rolling workflow.
- Would a check be desired for whether either each roll is a `DamageRoll` or whether they have `roll.options.type`? Based on my understanding & testing, if it's undefined the dice will roll with the default appearance.
- Should this be moved down a few lines, so it's inside the `message.create !== false` check? In theory DSN will only roll the dice if they're part of a message, but it felt like I should leverage the existing for loop for this one.

(__Per Zhell's suggestion, moved into `DamageRoll#evaluate` - none of the above are concerns any longer__)


krbz999:
Why not inside the `DamageRoll` class?

roth-michael:
Checked in there originally, but couldn't find an appropriate place for it. The appearance has to be set after damage type is chosen in the dialog (if any), but before the roll is sent to chat. Since 4.0, I think that all happens in the `BasicRoll` class's `build` function.

krbz999:
> Checked in there originally, but couldn't find an appropriate place for it. The appearance has to be set after damage type is chosen in the dialog (if any), but before the roll is sent to chat. Since 4.0, I think that all happens in the `BasicRoll` class's `build` function.

Without having looked at it, `DamageRoll#evaluate` may be the place.

roth-michael:
Yup. Duh. Brain said "I don't see evaluate function so I can't override."

module/dice/damage-roll.mjs
```diff
@@ -298,4 +298,14 @@ export default class DamageRoll extends BasicRoll {
...
     );
   }
+
+  /* -------------------------------------------- */
+  /*  Evaluate Methods                            */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async evaluate(options={}) {
+    this.options.appearance = { colorset: this.options.type };
```
arbron, line 308: It would probably be good to have this behind a check and to use `setProperty` to avoid deleting any existing appearance options:

```suggestion
    if ( this.options.type && !this.options.appearance?.colorset ) {
      foundry.utils.setProperty(this.options, "appearance.colorset", this.options.type);
    }
```

arbron (changes requested):

arbron:
@roth-michael Give a comment on the original issue so I can assign it to you

roth-michael:
`evaluateSync`, if for whatever reason called, should never result in a DSN dice roll, right? There's no weird situation I’m not thinking of that would necessitate an override of that method too?

krbz999:
> `evaluateSync`, if for whatever reason called, should never result in a DSN dice roll, right? There's no weird situation I’m not thinking of that would necessitate an override of that method too?

Sync rolls cannot involve dice.

roth-michael:
That was my understanding as well, just wanted to verify. Thanks.

Sayshal:
Would be great to see this merged.

JiDW:
Hello! Is this still  on the table? I'm prep'ing for the next major version of Dice So Nice and I have some features planned around damage type but I would need to know if this will eventually makes its way into the code or not to plan around it. 

I see that you have DamageRoll.options.type set to the damage type, I'm down to standardize this into a DsN API for other systems to follow, that could be even cleaner tbh

Thanks!

JiDW:
It's me again :wave: 
I've been exploring a lot of systems these last hours to see how they handle damage types, I've come to the conclusion that the best move would be to decline this pull request and let DsN pulls from your DamageRoll.type.

If that works from you, you can decline and close the issue, now or when dsn v6 releases

roth-michael:
Closing in favor of DsN handling!

PR #4500, opened 2024-10-03
Move hit dice data, add "additional" property

### LINKED ISSUE #2197: Add support (in world scripts) for multiple hit die per level if possible
I was using a homebrew class and translating it to foundry, and wanted to use a 2d8 as my hit dice. It didn't work at all, and broke a ton of stuff. I'm not sure if it is possible, but it is a suggestion I have

krbz999:
Due to the order of things being prepared, this means the HitDice class has to resort to getters. Not an issue, in fact a small performance improvement to only calculate the max or value once when needed.

image

Closes #2197.

### DIFF
--- a/module/documents/actor/hit-dice.mjs
+++ b/module/documents/actor/hit-dice.mjs
@@ -30,7 +30,7 @@ export default class HitDice {
    * @type {number}
    */
   get value() {
-    if ( this.#value ) return this.#value;
+    if ( this.#value !== undefined ) return this.#value;
     this.#value = this.classes.reduce((acc, cls) => acc + cls.system.hd.value, 0);
     return this.#value;
   }
@@ -44,7 +44,7 @@ export default class HitDice {
    * @type {number}
    */
   get max() {
-    if ( this.#max ) return this.#max;
+    if ( this.#max !== undefined ) return this.#max;
     this.#max = this.classes.reduce((acc, cls) => acc + cls.system.hd.max, 0);
     return this.#max;
   }
--- a/module/documents/actor/hit-dice.mjs
+++ b/module/documents/actor/hit-dice.mjs
@@ -10,12 +10,8 @@ export default class HitDice {
     this.actor = actor;
 
     for ( const item of Object.values(actor.classes) ) {
-      if ( /^d\d+$/.test(item.system.hitDice) ) {
-        this.classes.add(item);
-        this.value += item.system.levels - item.system.hitDiceUsed;
-        this.max += item.system.levels;
-        this.sizes.add(parseInt(item.system.hitDice.slice(1)));
-      }
+      this.classes.add(item);
+      this.sizes.add(parseInt(item.system.hd.denomination.slice(1)));
     }
   }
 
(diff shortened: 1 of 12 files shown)


module/documents/actor/hit-dice.mjs
```diff
@@ -10,12 +10,8 @@ export default class HitDice {
     this.actor = actor;
 
     for ( const item of Object.values(actor.classes) ) {
-      if ( /^d\d+$/.test(item.system.hitDice) ) {
-        this.classes.add(item);
-        this.value += item.system.levels - item.system.hitDiceUsed;
-        this.max += item.system.levels;
-        this.sizes.add(parseInt(item.system.hitDice.slice(1)));
-      }
+      this.classes.add(item);
```
krbz999, line 13: The data validation was performed in the class item already, so it's not necessary here. The `value` and `max` properties are moved to getters instead due to data preparation order (`additional` is not evaluated yet here).

templates/items/details/details-class.hbs
```diff
@@ -15,18 +15,20 @@
...
                          localize=true classes="label-top" }}
 
             {{!-- Spent --}}
-            {{ formField fields.hitDiceUsed value=source.hitDiceUsed placeholder="0" localize=true label="DND5E.Spent"
+            {{ formField fields.hd.fields.spent value=source.hd.spent placeholder="0" localize=true label="DND5E.Spent"
                          classes="label-top" }}
         </div>
     </div>
+
+    {{ formGroup fields.hd.fields.additional value=source.hd.additional localize=true }}
```
Fyorl, line 31: I thought maybe this should be an AE-only field. There's really no precedence in the rules for classes having 'additional' hit dice that I'm aware of.

Fyorl (approved):
Good changes. This feels a lot better to me. I've left some review comments but I can also potentially take care of them if you lack availability, let me know.

templates/items/details/details-class.hbs
krbz999: I didn't look into whether that would result in complications during data prep, but I imagine it would.

krbz999: can probably get to it within this or next week but likely you can get to it before then so fine by me

krbz999: > I didn't look into whether that would result in complications during data prep, but I imagine it would.

An enchantment might be able to do it, but otherwise it would require redoing parts of the `HitDie` class - and an effect targeting `classes` would throw an error. I'm open to suggestions.

Fyorl: Right, I keep forgetting that this is on an Item. Ignore my AE suggestion then. I meant more that it's fine to stay in the data model if something needs to adjust class hit dice, but I was less sure about it being available directly in the UI by default, since the game rules do not make use of it at all. Does that sound reasonable, or do you think it really needs to be in the UI by default?

krbz999: We've so far only excluded properties from the UI if they are not yet functional, isn't that case?

I don't feel so strongly about it if having a cleaner UI with fewer options is important, those who know about it can update a class item via API if need be, then.

krbz999: Though in that case, I think we should include it in the UI if it has a truthy value.

Fyorl (approved):

Fyorl (approved):

PR #4506, opened 2024-10-04
[#4261] Add consumption warnings to activity usage dialog

### LINKED ISSUE #4261: Consumption Dialog missing information
The current/max charges and recovery information is no longer shown in the dialog, nor are the warnings for if an item will be destroyed, or if there are no remaining uses available

proposal:
image

arbron:
Adds warnings to the activity usage dialog when attempting to perform a consumption with not enough uses and if consuming uses will result in a consumable being destroyed. Also adds info on the available uses in the hints when consuming.

Consumption Waring

Consumption Notes

The `getConsumptionLabels` can now return two additional pieces of data to support this:
- `notes` is an array of notes to display beneath the consumption section
- `warn` indicates that a warning icon should be displayed next to the label if the checkbox is checked

Also removes the old `AbilityUseDialog` that hadn't been fully removed in the `4.0` release.

Closes #4261


module/applications/activity/activity-usage-dialog.mjs
```diff
@@ -250,12 +250,16 @@ export default class ActivityUsageDialog extends Dialog5e {
       context.resources = [];
       const isArray = foundry.utils.getType(this.config.consume?.resources) === "Array";
       for ( const [index, target] of this.activity.consumption.targets.entries() ) {
+        const value = (isArray && this.config.consume.resources.includes(index))
+          || (!isArray && (this.config.consume?.resources !== false) && (this.config.consume !== false));
+        const { label, hint, notes, warn } = target.getConsumptionLabels(this.config, value);
+        if ( notes?.length ) context.notes.push(...notes);
```
Fyorl, line 256: I might have missed some places, but it seemed like the only place we used the newly-added `consumed` parameter was in `consumptionLabelsItemUses`, and it was used to control whether a note was generated. Since we are calling `getConsumptionLabels`, can we assume that any notes returned are related to consumption, so we only need to display them if `consumed` (`value` here) is `true`? That way we don't have to propagate this boolean through the whole `getConsumptionLabels` pipeline. If not then, maybe we pre-emptively make it an `options` object rather than a single boolean.

```suggestion
        if ( notes?.length && value ) context.notes.push(...notes);
```

Fyorl (changes requested):

module/applications/activity/activity-usage-dialog.mjs
arbron: I stuck it into an options object, because I can foresee we may decide to display certain notes even if a certain consumption type is currently unchecked.

Fyorl (approved):

PR #4507, opened 2024-10-05
[#4471] Add unarmored/armored ac bonuses, and max dex per armor type

### LINKED ISSUE #4471: Additional Armor Class properties
Common use case for Armor Class bonuses are
- AC bonus that applies only if the actor _is not_ wearing armor.
- AC bonus that applies only if the actor *is* wearing armor.
- Modification to the max dex modifier (max of 3 instead of max of 2).

It would be nice if three properties were added to the `attributes.ac` object during data prep to apply these bonuses conditionally. Eg.:
- `attributes.ac.bonuses.unarmored`
- `attributes.ac.bonuses.armored`
- `attributes.ac.bonuses.ability` (need a better name...)

Since `attributes.ac.bonus` exists and already uses roll data, these three properties would also be able to use roll data easily.

roth-michael:
closes #4471 

Mostly complete, still a minor WIP:
- Not sold on `noMaxDex` being the property key
- In a perfect world, that property would only show for non-shield armor (rather than all equipment), but adding that condition to the template felt like too large a change for this one use case
- I'd love to clean up the placeholder conditional if possible (basically: If it's a number, display the number, otherwise display infinity)
- A one-time migration should ideally be made on any medium armor currently set with no maximum dex bonus to have the new property added. (Light & natural armor can be ignored for this, as they default to no max bonus, and heavy armor can be ignored as prior to this change it was hardcoded to have 0 max regardless of what was on the armor item)


module/data/actor/templates/attributes.mjs
```diff
@@ -114,6 +114,12 @@ export default class AttributesFields {
     ac.armor = 10;
     ac.shield = ac.cover = 0;
     ac.min = ac.bonus = "";
+    ac.ability = {
+      light: "", medium: "", heavy: "", natural: ""
+    };
```
krbz999, line 119: Could use a loop over `CONFIG.DND5E.armorMaxDex` instead to leave this configurable.

module/config.mjs
```diff
@@ -1292,6 +1292,19 @@ preLocalize("armorTypes");
 
 /* -------------------------------------------- */
 
+/**
+ * Specific equipment types that modify base AC.
+ * @enum {string}
+ */
+DND5E.armorMaxDex = {
```
krbz999, line 1299: Wondering if this could be merged into `DND5E.armorTypes`, but unsure what to do with the `shield` value.

templates/items/details/details-equipment.hbs
```diff
@@ -26,7 +26,7 @@
             {{!-- Max Dex --}}
             {{#if hasDexModifier}}
             {{ formField fields.armor.fields.dex value=source.armor.dex label="DND5E.ItemEquipmentDexModAbbr"
-                         placeholder="∞" localize=true classes="label-top" }}
+                         placeholder=(ifThen (and (gte (lookup config.armorMaxDex source.type.value) 0) (ne (lookup config.armorMaxDex source.type.value) null)) (dnd5e-numberFormat (lookup config.armorMaxDex source.type.value)) "∞") localize=true classes="label-top" }}
```
krbz999, line 29: Should probably be in the `getData` method instead given how verbose this one line is.

roth-michael:
Latest push turns `CONFIG.DND5E.armorTypes` into an object with `label` being the old value and `maxAbility` being a new value, integer or `null` for all except `shield`. This is checked for only in the newly (this PR) added line in `prepareBaseArmorClass`:
```js
for ( const [k, v] of Object.entries(CONFIG.DND5E.armorTypes) ) if ( v.maxAbility !== undefined ) ac.ability[k] = "";
```
Only bit of mystery for me is the following in `starting-equipment.mjs`:
```js
static CATEGORIES = {
  armor: {
    label: "DND5E.Armor",
    config: "armorTypes"
  },
  focus: {
    label: "DND5E.Focus.Label",
    config: "focusTypes"
  },
  tool: {
    label: "TYPES.Item.tool",
    config: "toolTypes"
  },
  weapon: {
    label: "TYPES.Item.weapon",
    config: "weaponProficiencies"
  }
};
```
Not sure how, if at all, to modify this given `armorTypes` now contains nested objects. The compiler tells me `CATEGORIES` is never used though, so maybe a non-issue.

roth-michael:
@arbron Though I'm loathe to ping you on a PR you're not already involved in, could use your input on something:
To summarize the above PR,
- `ac.bonuses.unarmored` AC bonus when not wearing armor
- `ac.bonuses.armored` AC bonus when wearing armor
- `ac.ability.[heavy|medium|light|natural]` modification to max dex bonus for each

There are some features (at least one first party - Swashbuckler Monster's Suave Defense) which provide an AC bonus if unarmored _or_ wearing light armor. This potentially raises a case for `ac.bonuses.[heavy|medium|light|natural|shield]`. It wouldn't be very difficult to implement as part of the PR, but wanted to get your thoughts on it. Thanks.

krbz999:
> There are some features (at least one first party - Swashbuckler Monster's Suave Defense) which provide an AC bonus if unarmored _or_ wearing light armor. This potentially raises a case for `ac.bonuses.[heavy|medium|light|natural|shield]`.

A different implementation would be to use the `dnd5e.Filters` methods and cause the effect to only optionally apply, thus no need for any additional properties. UI would be its own issue of course.

roth-michael:
#6658 obviates _most_ of the code in here, I think. wouldn't handle being able to modify default max dex bonus per armor type, though.

roth-michael:
Gonna close this now as the main functionality is already implemented. Can re-open a new PR to handle adjusting max dex bonus if desired.

PR #4512, opened 2024-10-06
[#4310] Re-design armor class config to use V2 app & styles

arbron:
Reworks `ActorArmorConfig` as `ArmorClassConfig` using the new design. Also added a section that displays the currently equipped armor and shield with their AC value and magical bonus.

Armor Class Config V2


Fyorl (commented):
Looks like this is also due a rebase after the HP/HD configs went in

Fyorl (approved):

PR #4514, opened 2024-10-07
Allow different base die for core d20 rolls

Larkinabout:
image

For homebrew purposes, allow modules or world scripts to modify the number of faces of the base die through hooks such as `dnd5e.preRollInitiativeDialog` (added with this PR) or `dnd5e.dnd5e.preRollAbilitySave`. Otherwise, default to the expected d20.

This change adds a `faces` property to `rollData` (or `rollOptions` in the case of initiative), which defaults to 20.

I did consider whether this meant needing to rename all the 'D20' references to something like 'AdvDisRoll', as this change is primarily to allow the base die to be changed for all rolls that use the Advantage/Disadvantage mechanic and all associated features therein. But, I think it's okay to keep the naming around the d20 and this is just a slight allowance for homebrew within that core functionality.


Larkinabout:
I've just realised I've introduced this change right when a major overhaul of the d20 roll code is happening, so if it can be baked into those changes, or I can reintroduce the change once the overhaul is complete.

Larkinabout:
Closing as outdated.

PR #4516, opened 2024-10-07
[#4310] Re-design initiative config to use V2 app & styles

arbron:
Deprecates `ActorInitiativeConfig` in favor of `InitiativeConfig` which uses ApplicationV2 and the new styles. Also displays the check bonuses for the selected ability to clarify what contributes to initiatve rolls.

Initiative Config V2


krbz999:
A good time to deprecate the Alert and advantage flags perhaps?

module/applications/actor/config/initiative-config.mjs
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preparePartContext(partId, context, options) {
+    context = await super._preparePartContext(partId, context, options);
+    const source = this.document.toObject();
```
Fyorl, line 42:
```suggestion
    const source = this.document._source;
```

Fyorl (approved):

Fyorl:
> A good time to deprecate the Alert and advantage flags perhaps?

Since the roll refactor unblocks `min` and `max` we could at least add those to this dialog. I don't think we can deprecate the alert feat flag because of how it behaves differently between the 2014 and 2024 rules.

We could deprecate the initiative advantage flag and use the `mode`, but I don't think we have all of the architecture in place yet to support `mode`, do we?

krbz999:
> We could deprecate the initiative advantage flag and use the `mode`, but I don't think we have all of the architecture in place yet to support `mode`, do we?

Currently only in Concentration saves where it sets the default. (That may have changed in the 4.1 branch, I have not checked.)

PR #4517, opened 2024-10-07
[#4310] Implement damage modification config in `DamagesConfig`

arbron:
Adjust how the `traits-list` partial works to take a customizable field and modify `DamagesConfig` to pass in a formula input rather than a checkbox input when handling damage modifications.

Damage Modification Config V2

This also involves some changes to `TraitAdvancement` so that damage modification can be registered in `CONFIG.DND5E.traits` without appearing in the advancement list.


Fyorl (commented):
I guess this needs a rebase since it seems to have all of `TraitsConfig` in here too.

Fyorl (approved):

PR #4521, opened 2024-10-08
[#4310] Merge & re-design movement & senses configs for V2

arbron:
Deprecates `ActorMovementConfig` and `ActorSensesConfig` in favor of a merged `MovementSensesConfig`.

Movement   Senses Configs V2


Fyorl (approved):
Would be nice to style the fly + hover with an inline checkbox like we do in a few other places:

image

I appreciate that it doesn't fit in with how the dialog is setup to be generic, so ultimately not a big deal if it's too much trouble.

PR #4526, opened 2024-10-09
[#4192] Support multiple save abilities to `SaveActivity`

### LINKED ISSUE #4192: Allow for multiple saving throws on a single activity
It would be nice for an activity to ask for multiple saving throws. Grappling and shoving in the new rules allows for a dex or str saving throw (the one being grappled chooses which) so it would be nice if the chat card has both of those saves there.

The module [Roll Groups](https://foundryvtt.com/packages/rollgroups) has this functionality, so maybe it can be used as an example.

arbron:
Modifies `SaveActivity` to store a set of abilities rather than a single one that can be rolled to save. When multiple are selected if will display multiple buttons in chat to roll.

Closes #4192

### DIFF
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -875,9 +875,7 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
       if ( foundry.utils.getType(save?.ability) === "Set" ) save = {
         ...save, ability: save.ability.size > 2
           ? game.i18n.localize("DND5E.AbbreviationDC")
-          : game.i18n.getListFormatter({ style: "narrow" }).format(
-            Array.from(save.ability).map(k => CONFIG.DND5E.abilities[k].abbreviation)
-          )
+          : Array.from(save.ability).map(k => CONFIG.DND5E.abilities[k].abbreviation).join(" / ")
       };
 
       const css = [];
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -872,6 +872,14 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
         save, range, reference, toggle, suppressed, level
       } = data;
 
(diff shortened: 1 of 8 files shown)


module/applications/actor/character-sheet-2.mjs
```diff
@@ -702,6 +702,14 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
         save, range, reference, toggle, suppressed, level
       } = data;
 
+      if ( foundry.utils.getType(save?.ability) === "Set" ) save = {
+        ...save, ability: save.ability.size > 2
+          ? game.i18n.localize("DND5E.AbbreviationDC")
+          : game.i18n.getListFormatter({ style: "narrow" }).format(
+            Array.from(save.ability).map(k => CONFIG.DND5E.abilities[k].abbreviation)
+          )
+      };
```
Fyorl, line 711: Need the same treatment where these DCs appear on items in sheets. That's the 'roll' column for spells on character sheets, and I think NPC sheets also display this on items on the main tab as well as the spells tab. In those places it should probably use the "DC" label if the number of abilities is >1, I don't think there's room for 2 abilities in those columns.

module/applications/actor/character-sheet-2.mjs
```diff
@@ -702,6 +702,14 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
         save, range, reference, toggle, suppressed, level
       } = data;
 
+      if ( foundry.utils.getType(save?.ability) === "Set" ) save = {
+        ...save, ability: save.ability.size > 2
+          ? game.i18n.localize("DND5E.AbbreviationDC")
+          : game.i18n.getListFormatter({ style: "narrow" }).format(
+            Array.from(save.ability).map(k => CONFIG.DND5E.abilities[k].abbreviation)
```
Fyorl, line 709: I think, e.g. `STR / DEX` might look better than `STR, DEX`.

Fyorl (changes requested):

Fyorl (changes requested):
NPC sheet features tab still needs fixes for save formatting

image

Fyorl (approved):

PR #4531, opened 2024-10-11
[#4061] Allow Cast Activity to trigger linked spells

arbron:
Overrides the `use` method on `CastActivity` to trigger the cached spell, rather than the activity itself. If the spell isn't cached, it will first create a new cached copy. Triggers the `use` method on the spell item so that the activity selection dialog can be shown if necessary.

The `CastActivity` introduces two new hooks: `dnd5e.preUseLinkedSpell` and `dnd5e.poseUseLinkedSpell`.

Adds a new option to `ActivtyUseConfiguration` called `triggering` which contains the relative UUID of the activity triggering this one as well as resource consumption options. This value is filled automatically if the item has the `cachedFor` flag unless it is explicitly set to `false`.

If `triggering` is set in the usage configuration, then during the consumption process the system will also trigger the consumption process on the triggering activity and merge its changes with that of the main activity.

Some changes have been made to `ActivityUsageDialog` to show consumption for both the triggering and triggered activities, and to modify the spell slots list to reflect the scaling restriction on the triggering activity.

## Todo
- [x] Properly handle consumption warnings on triggering activity once #4261 is resolved

### DIFF
--- a/module/documents/activity/cast.mjs
+++ b/module/documents/activity/cast.mjs
@@ -47,6 +47,50 @@ export default class CastActivity extends ActivityMixin(CastActivityData) {
   }
 
   /* -------------------------------------------- */
+  /*  Activation                                  */
+  /* -------------------------------------------- */
+
+  /** @override */
+  async use(usage={}, dialog={}, message={}) {
+    if ( !this.item.isEmbedded || this.item.pack ) return;
+    if ( !this.item.isOwner ) {
+      ui.notifications.error("DND5E.DocumentUseWarn", { localize: true });
+      return;
+    }
+
+    /**
+     * A hook event that fires before a linked spell is used by a Cast activity.
+     * @function dnd5e.preUseLinkedSpell
+     * @memberof hookEvents
+     * @param {CastActivity} activity                                Cast activity being used.
+     * @param {Partial<ActivityUseConfiguration>} usageConfig        Configuration info for the activation.
+     * @param {Partial<ActivityDialogConfiguration>} dialogConfig    Configuration info for the usage dialog.
+     * @param {Partial<ActivityMessageConfiguration>} messageConfig  Configuration info for the created chat message.
+     * @returns {boolean}  Explicitly return `false` to prevent activity from being used.
+     */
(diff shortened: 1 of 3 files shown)


module/documents/activity/cast.mjs
```diff
@@ -46,6 +46,50 @@ export default class CastActivity extends ActivityMixin(CastActivityData) {
...
+     * @param {Partial<ActivityUseConfiguration>} usageConfig        Configuration info for the activation.
+     * @param {Partial<ActivityDialogConfiguration>} dialogConfig    Configuration info for the usage dialog.
+     * @param {Partial<ActivityMessageConfiguration>} messageConfig  Configuration info for the created chat message.
+     * @returns {boolean}  Explicitly return `false` to prevent activity from being used.
+     */
+    if ( Hooks.call("dnd5e.preUseLinkedSpell", this, usage, dialog, message) === false ) return;
+
+    let spell = this.cachedSpell;
+    if ( !spell ) {
+      spell = await this.actor.createEmbeddedDocuments("Item", [await this.getCachedSpellData()])[0];
```
krbz999, line 75:
```suggestion
      spell = (await this.actor.createEmbeddedDocuments("Item", [await this.getCachedSpellData()]))[0];
```

module/documents/activity/mixin.mjs
```diff
@@ -1235,4 +1291,21 @@ export default Base => class extends PseudoDocumentMixin(Base) {
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Merge the activity updates into this activity's item updates.
+   * @param {ActivityUsageUpdates} updates
+   * @internal
+   */
+  _mergeActivityUpdates(updates) {
+    if ( !foundry.utils.isEmpty(updates.activity) ) {
```
Fyorl, line 1303: Prefer an early return if `updates.activity` is empty

Fyorl (approved):
Looks like this needs a rebase, but I think I was able separate out the parts that were from `cast-cache`.

This looks good though, see my suggestions around naming but I think this is in a good place.

PR #4532, opened 2024-10-11
[#4383] Add `dnd5e.postRollConfiguration` hooks

### LINKED ISSUE #4383: Add post-configuration hooks for rolls
While doing some work on a module, I wanted to determine something from, say, the attack mode of an Attack roll. The pre hook gives access to all the modes that are available, not the one chosen (because one has not been chosen yet), and the post hook is after the roll is fully evaluated and the chat message created, which is too late.

It would be good to have a hook for when the roll configuration dialog is closed and the roll fully configured, using the same parameters if possible.

arbron:
Introduces a new hook after roll configuration is completed but before the rolls are evaluated to give modules a chance to modify the rolls or cancel the rolling process. This allows modules to respond to new elements added to the roll configuration dialog.

Rather than adding a single hook, this introduces the `hookNames` option to the roll process configuration that allows it to call a series of hooks. So for skills checks for example it would call `dnd5e.postSkillRollConfiguration`, `dnd5e.postAbilityTestRollConfiguration`, and `dnd5e.postRollConfiguration`.

Closes #4383

### DIFF
--- a/module/data/shared/uses-field.mjs
+++ b/module/data/shared/uses-field.mjs
@@ -180,7 +180,6 @@ export default class UsesField extends SchemaField {
     }, config);
     rollConfig.hookNames = [...(config.hookNames ?? []), "recharge"];
     rollConfig.subject = this;
-    rollConfig.hookNames = ["recharge"].concat(config.hookNames ?? []);
 
     const dialogConfig = foundry.utils.mergeObject({ configure: false }, dialog);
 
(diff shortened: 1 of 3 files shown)


module/data/shared/uses-field.mjs
```diff
@@ -179,6 +179,7 @@ export default class UsesField extends SchemaField {
       }]
     }, config);
     rollConfig.subject = this;
+    rollConfig.hookNames = (config.hookNames ?? []).concat(["recharge"]);
```
Fyorl, line 182: Any reason to use `concat` rather than the spread operator? It avoids an extra array creation.

```suggestion
    rollConfig.hookNames = [...(config.hookNames ?? []), "recharge"];
```

Fyorl (approved):

PR #4542, opened 2024-10-14
[#1530, #2138] Re-implement `rollSkill` and `rollToolCheck`

### LINKED ISSUE #1530: Remarkable Athlete flag doesn't contribute to skill roll if normally non-applicable skill is rolled with different ability

arbron:
Merges the functionality ot skill & tool checks into a single method that takes advantage of the new rolling APIs. These new methods make use of a `buildConfig` callback that is called by `RollConfigurationDialog` during its `buildConfig` method. This allows for more extensive modifications of the configuration when the ability is changed.

Closes #1530

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -443,7 +443,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
     // Compute modifier
     const checkBonusAbl = simplifyBonus(abilityData?.bonuses?.check, rollData);
-    skillData.baseValue = skillData.value;
+    skillData.effectValue = skillData.value;
     skillData.bonus = baseBonus + globalCheckBonus + checkBonusAbl + globalSkillBonus;
     skillData.mod = abilityData?.mod ?? 0;
     skillData.prof = this.system.calculateAbilityCheckProficiency(skillData.value, skillData.ability);
@@ -474,7 +474,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       const ability = this.system.abilities[tool.ability];
       const baseBonus = simplifyBonus(tool.bonuses.check, bonusData);
       const checkBonusAbl = simplifyBonus(ability?.bonuses?.check, bonusData);
-      tool.baseValue = tool.value;
+      tool.effectValue = tool.value;
       tool.bonus = baseBonus + checkBonus + checkBonusAbl;
       tool.mod = ability?.mod ?? 0;
       tool.prof = this.system.calculateAbilityCheckProficiency(tool.value, tool.ability);
@@ -1508,7 +1508,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const rollData = this.getRollData();
     const abilityId = formData?.get("ability") ?? process.ability;
     const ability = this.system.abilities?.[abilityId];
-    const prof = this.system.calculateAbilityCheckProficiency(relevant.baseValue, abilityId);
+    const prof = this.system.calculateAbilityCheckProficiency(relevant.effectValue, abilityId);
 
     let { parts, data } = CONFIG.Dice.BasicRoll.constructParts({
       mod: ability?.mod,
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -443,10 +443,11 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
     // Compute modifier
     const checkBonusAbl = simplifyBonus(abilityData?.bonuses?.check, rollData);
+    skillData.baseValue = skillData.value;
     skillData.bonus = baseBonus + globalCheckBonus + checkBonusAbl + globalSkillBonus;
     skillData.mod = abilityData?.mod ?? 0;
     skillData.prof = this.system.calculateAbilityCheckProficiency(skillData.value, skillData.ability);
-    skillData.proficient = skillData.value;
+    skillData.value = skillData.proficient = skillData.prof.multiplier;
     skillData.total = skillData.mod + skillData.bonus;
     if ( Number.isNumeric(skillData.prof.term) ) skillData.total += skillData.prof.flat;
 
@@ -473,11 +474,13 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       const ability = this.system.abilities[tool.ability];
       const baseBonus = simplifyBonus(tool.bonuses.check, bonusData);
       const checkBonusAbl = simplifyBonus(ability?.bonuses?.check, bonusData);
+      tool.baseValue = tool.value;
       tool.bonus = baseBonus + checkBonus + checkBonusAbl;
       tool.mod = ability?.mod ?? 0;
       tool.prof = this.system.calculateAbilityCheckProficiency(tool.value, tool.ability);
       tool.total = tool.mod + tool.bonus;
       if ( Number.isNumeric(tool.prof.term) ) tool.total += tool.prof.flat;
+      tool.value = tool.prof.multiplier;
     }
   }
 
@@ -1503,11 +1506,9 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
   _buildSkillToolConfig(type, initialRoll, process, config, formData, index) {
     const relevant = type === "skill" ? this.system.skills?.[process.skill] : this.system.tools?.[process.tool];
     const rollData = this.getRollData();
-    const defaultAbility = relevant?.ability
-      ?? (type === "skill" ? CONFIG.DND5E.skills[process.skill].ability : CONFIG.DND5E.tools[process.tool].ability);
-    const abilityId = formData?.get("ability") ?? process.ability ?? defaultAbility;
+    const abilityId = formData?.get("ability") ?? process.ability;
     const ability = this.system.abilities?.[abilityId];
-    const prof = this.system.calculateAbilityCheckProficiency(relevant.value, abilityId);
+    const prof = this.system.calculateAbilityCheckProficiency(relevant.baseValue, abilityId);
 
     let { parts, data } = CONFIG.Dice.BasicRoll.constructParts({
       mod: ability?.mod,
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1,8 +1,8 @@
 import ShortRestDialog from "../../applications/actor/short-rest.mjs";
 import LongRestDialog from "../../applications/actor/long-rest.mjs";
+import SkillToolRollConfigurationDialog from "../../applications/dice/skill-tool-configuration-dialog.mjs";
 import PropertyAttribution from "../../applications/property-attribution.mjs";
 import { _applyDeprecatedD20Configs, _createDeprecatedD20Config } from "../../dice/d20-roll.mjs";
-import { d20Roll } from "../../dice/dice.mjs";
 import { createRollLabel } from "../../enrichers.mjs";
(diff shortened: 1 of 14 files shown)


module/documents/actor/actor.mjs
```diff
@@ -1359,191 +1359,208 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-    if ( tool?.bonuses.check || options.bonus ) {
-      parts.push("@toolBonus");
-      const bonus = [];
-      if ( tool?.bonuses.check ) bonus.push(Roll.replaceFormulaData(tool.bonuses.check, data));
-      if ( options.bonus ) bonus.push(Roll.replaceFormulaData(options.bonus, data));
-      data.toolBonus = bonus.join(" + ");
-    }
+    const hasRemarkableAthlete = this.getFlag("dnd5e", "remarkableAthlete")
+      && (game.settings.get("dnd5e", "rulesVersion") === "legacy");
+    const buildConfig = (process, config, formData, index) => {
```
Fyorl, line 1443: Rather than a callback, could we have our `SkillToolRollConfiguration` subclass override `_buildRoll` and call a well-known method on the Actor? Something like `_prepareSkillCheck`/`_prepareToolCheck` maybe?

Fyorl (changes requested):

module/documents/actor/actor.mjs
arbron: That does add some complications to parts because right now we rely on bound variables for some of the calculations (relevant skill, default ability, and roll data so we only have to prepare it once, plus initial roll data because that is passed in to the initial configuration function). I could perhaps split this out into its own method but bind these values into it before passing it to the roll config:

```javascript
const buildConfig = this._buildSkillToolConfig.bind(this, { defaultAbility, initialRoll, relevant, rollData });
```

Fyorl: Binding is an option. It's not an issue to recompute things like `rollData`, or `hasRemarkableAthlete` inside the method, since it's only called a single time as part of user interaction.

This is how far I got with just a pure method-based solution:

```js
// SkillToolRollConfigurationDialog
// Add 'type' and 'actor' to dialog options
/**
 * @typedef {BasicRollConfigurationDialogOptions} SkillToolRollConfigurationDialogOptions
 * @property {boolean} chooseAbility  Should the ability be selectable?
 * @property {"skill"|"tool"} type    The type of roll.
 * @property {Actor5e} actor          The Actor making the roll.
 */
_buildConfig(config, formData, index) {
  super._buildConfig(config, formData, index);
  this.options.actor._prepareSkillTool(this.options.type, this.config, config, formData);
}
```

```js
// Actor5e
async #rollSkillTool(type, config={}, dialog={}, message={}) {
  // [snip]
  const rollConfig = { /* ... */ };
  this._prepareSkillTool(type, rollConfig, rollConfig.rolls[0], null);
  const dialogConfig = foundry.utils.mergeObject({
    applicationClass: SkillToolRollConfigurationDialog,
    options: { type, chooseAbility: true, actor: this }
  }, dialog);
}

/**
 * @protected
 */
_prepareSkillTool(type, process, config, formData) {
  const rollData = this.getRollData();
  // Should not need to derive defaultAbility here since process.ability is already set to that value.
  const abilityId = formData?.get("ability") ?? process.ability;
  const ability = this.system.abilities?.[abilityId];
  const skill = this.system.skills?.[process.skill];
  const tool = this.system.tools?.[process.tool];
  const relevant = type === "skill" ? skill : tool;
  const hasRemarkableAthlete = this.getFlag("dnd5e", "remarkableAthlete")
    && (game.settings.get("dnd5e", "rulesVersion") === "legacy");

  let prof = process.prof ?? relevant?.prof;
  if ( prof && hasRemarkableAthlete ) {
    const validAbilities = CONFIG.DND5E.characterFlags.remarkableAthlete.abilities;
    if ( validAbilities.includes(abilityId) && (prof.multiplier <= .5) && (prof.rounding === "down") ) {
      prof = prof.clone({ multiplier: .5, roundDown: false });
    } else if ( !validAbilities.includes(abilityId) && (prof.multiplier === .5) && (prof.rounding === "up") ) {
      if ( this.getFlag("dnd5e", "jackOfAllTrades") ) prof = prof.clone({ roundDown: true });
      else prof = prof.clone({ multiplier: 0 });
    }
  }

  const { parts, data } = CONFIG.Dice.BasicRoll.constructParts({
    mod: ability?.mod,
    prof: prof?.hasProficiency ? prof.term : null,
    [`${config[type]}Bonus`]: relevant?.bonuses?.check,
    extraBonus: process.bonus,
    [`${abilityId}CheckBonus`]: ability?.bonuses?.check,
    [`${type}Bonus`]: this.system.bonuses?.abilities?.[type],
    abilityCheckBonus: this.system.bonuses?.abilities?.check,
    situational: config.data?.situational
  }, { ...rollData });
  const options = config.options ?? {};

  this.addRollExhaustion(parts, data);
  config.parts = parts;
  config.data = data;
  config.data.abilityId = abilityId;
  config.options = options;
}
```

I think passing `type` to the `SkillToolRollConfigurationDialog` is pretty reasonable. The `SkillToolRollConfigurationDialog` should know whether it's a skill roll or a tool roll. But I don't love having to pass the `Actor` really. So I think this exercise has me leaning more towards the binding solution, where we bind the `Actor` as `this` and then the roll dialog doesn't need to know.

I would like to understand a bit better what we are doing with `initialRoll` though since that's the only part missing from the above. Is it a holdover from the old API? I don't think we use this at all internally, so is this something related to module integration?

module/documents/actor/actor.mjs
```diff
@@ -1359,191 +1359,208 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+
+      let prof = process.prof ?? relevant?.prof;
+      if ( prof && hasRemarkableAthlete ) {
+        const validAbilities = CONFIG.DND5E.characterFlags.remarkableAthlete.abilities;
+        if ( validAbilities.includes(abilityId) && (prof.multiplier <= .5) && (prof.rounding === "down") ) {
+          prof = prof.clone({ multiplier: .5, roundDown: false });
+        } else if ( !validAbilities.includes(abilityId) && (prof.multiplier === .5) && (prof.rounding === "up") ) {
+          if ( this.getFlag("dnd5e", "jackOfAllTrades") ) prof = prof.clone({ roundDown: true });
+          else prof = prof.clone({ multiplier: 0 });
+        }
```
Fyorl, line 1455: I think I understand what we're doing here. Where we try to work out whether remarkable athlete or JoAT has been applied during preparation. But perhaps a bit clearer would be to derive this from the beginning by looking at `_source.skills[config.skill].value` rather than attempting to undo the existing preparation. It could potentially share code with the preparation method.

module/documents/actor/actor.mjs
arbron: `initialRoll` is there to better support calling `rollSkill` or `rollToolCheck` with default options. So if you call:

```javascript
actor.rollSkill({ roll: [{ parts: ["1d4"] }] });
```

That will add the provided `1d4` part to the roll that is automatically constructed, so you end up with `1d20 + @mod + 1d4`. Same with the data and options provided there, allowing you to override what would be automatically calculated.

For all other parts we just append a new roll, so if we didn't handle initial roll we'd end up with one roll `1d20 + @mod`, and a second roll `1d20 + @mod + 1d4` (that's still possible in this system, but you have to pass multiple options to `roll` in the first object).

Fyorl: OK I think that makes sense. What confused me was that we call `config.rolls?.pop()`, but it sounds like from what you described that we should be doing `config.rolls?.shift()`, no? Also, why do we need to do this merge every time in the callback? If these options are supposed to be merged with the generated roll, can't we do it just once when we assemble `rollConfig`? I can see that we are supplying some roll options already to the first roll when we assemble `rollConfig`.

module/documents/actor/actor.mjs
```diff
@@ -1359,191 +1359,208 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-        "flags.dnd5e.roll": {type: "tool", toolId}
+      reliableTalent: (relevant?.value >= 1) && this.getFlag("dnd5e", "reliableTalent"),
+      rolls: [{
+        options: {
+          advantage: relevant?.roll.mode === CONFIG.Dice.D20Roll.ADV_MODE.ADVANTAGE,
+          disadvantage: relevant?.roll.mode === CONFIG.Dice.D20Roll.ADV_MODE.DISADVANTAGE,
+          maximum: relevant?.roll.max,
+          minimum: relevant?.roll.min
+        }
+      }]
```
Fyorl, line 1496: This `rolls` array we define here will be clobbered by `config.rolls` if it is also defined.

module/documents/actor/actor.mjs
arbron: Okay, I was able to go that way but I had to make a tweak to ensure active effects are handled properly. So now during skill/tool preparation it no longer adjusts `skill.value` based on Jack of All Trades & Remarkable Athlete, instead it just takes those flags into account when creating the Proficiency object. This seems fine for our code since we use the proficiency object for determining proficiency, but anyone who was relying on `value` directly might have issues. Not sure that is worth worrying about.

module/documents/actor/actor.mjs
arbron: Aw yeah, shift is correct. We do this every preparation because we need to merge these in after the default roll parts, data, and options are prepared since these should take priority over those options, and those underlying bits might change every time the user makes a choice.

module/documents/actor/actor.mjs
Fyorl: Ah right, I hadn't considered Active Effects. It's possible that not updating `value` will cause some downstream disruption. There were a couple of places we were using it in our code judging by this PR, so it's reasonable to assume it's used downstream too.

If we just need a value that is pre-JoAT/RA, but post AE-application, then I would think we can just write it ourselves during preparation, e.g.:

```js
skill.preFeat = skill.value;
skill.value = this.calculateAbilityCheckProficiency(skill.value, skill.ability).multiplier;
```

Fyorl (changes requested):

module/documents/actor/actor.mjs
```diff
@@ -1359,191 +1332,208 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-    if ( globalBonuses.check ) {
-      parts.push("@checkBonus");
-      data.checkBonus = Roll.replaceFormulaData(globalBonuses.check, data);
-    }
+  _buildSkillToolConfig(type, initialRoll, process, config, formData, index) {
+    const relevant = type === "skill" ? this.system.skills?.[process.skill] : this.system.tools?.[process.tool];
+    const rollData = this.getRollData();
+    const defaultAbility = relevant?.ability
+      ?? (type === "skill" ? CONFIG.DND5E.skills[process.skill].ability : CONFIG.DND5E.tools[process.tool].ability);
+    const abilityId = formData?.get("ability") ?? process.ability ?? defaultAbility;
```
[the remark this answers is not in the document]
arbron: I think you're right, just a bit of confusion that slipped in when I split the methods.

Fyorl (approved):

PR #4544, opened 2024-10-15
[#4541] Option to auto-roll NPC HP

### LINKED ISSUE #4541: Add option to roll for NPC HP
The v12 rolling changes make it much harder to execute randomized NPC hp for modules, but the system's `_preCreate` handling is still asynchronous. Rolling for monster hit points is a core rule and now has a strong reason why it needs to be supported by the system rather than left to module functionality.

roth-michael:
Closes #4541 

Don't think I've missed any edge cases here (only one I could think of is for an NPC without an HP formula, e.g. a named NPC or something). 

The thought had occurred to me to maybe make this rolling specifically always a whispered chat message if enabled, but that'd require modifying the `rollNPCHitPoints` function to take another input (or copying the relevant code over and not using that function) so figured just adding the "Silent" option was a good compromise.


module/settings.mjs
```diff
@@ -440,6 +440,21 @@ export function registerSystemSettings() {
...
+    hint: "SETTINGS.DND5E.AUTOROLLNPCHP.Hint",
+    scope: "world",
+    config: true,
+    default: "no",
+    type: String,
+    choices: {
+      no: "SETTINGS.DND5E.AUTOROLLNPCHP.No",
+      silent: "SETTINGS.DND5E.AUTOROLLNPCHP.Silent",
+      yes: "SETTINGS.DND5E.AUTOROLLNPCHP.Yes"
+    }
```
krbz999, line 456: Could (should?) use a data field here.

roth-michael: Just based mine on the existing single-select dropdowns. Happy to change if necessary though - what's the benefit of doing it that way? (Might also make sense as a separate maintenance PR that changes _all_ relevant settings to use data fields.)

JPMeehan: DataFields have two big upsides
- Functions for initial value & `choices` (more useful when you're doing things like picking a document ID)
- Supports a labeled blank option.

module/documents/token.mjs
```diff
@@ -165,6 +165,28 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
 
+  /* -------------------------------------------- */
+  /*  Event Handlers                              */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preCreate(data, options, user) {
+    if ( (await super._preCreate(data, options, user)) === false ) return false;
+
+    if ( this.actor?.type === "npc" && !this.actorLink && foundry.utils.getProperty(this.actor, "system.attributes.hp.formula")?.trim().length ) {
```
arbron, line 176:
```suggestion
    if ( (this.actor?.type === "npc") && !this.actorLink && foundry.utils.getProperty(this.actor, "system.attributes.hp.formula")?.trim().length ) {
```

arbron (changes requested):

arbron:
I have a PR up to move some combat-related settings into their own app, so this will probably want to go into there once it is merged: https://github.com/foundryvtt/dnd5e/pull/5103

module/documents/token.mjs
```diff
@@ -165,6 +165,28 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
 
+  /* -------------------------------------------- */
+  /*  Event Handlers                              */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preCreate(data, options, user) {
+    if ( (await super._preCreate(data, options, user)) === false ) return false;
+
+    if ( (this.actor?.type === "npc") && !this.actorLink && foundry.utils.getProperty(this.actor, "system.attributes.hp.formula")?.trim().length ) {
```
Fyorl, line 176: Check line length

module/documents/token.mjs
```diff
@@ -165,6 +165,28 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+      const autoRoll = game.settings.get("dnd5e", "autoRollNPCHP");
+      if ( autoRoll !== "no" ) {
+        const roll = await this.actor.rollNPCHitPoints({ chatMessage: autoRoll === "yes" });
+        await this.delta.updateSource({
+          "system.attributes.hp": {
+            max: roll.total,
+            value: roll.total
+          }
+        });
+      }
```
Fyorl, line 186: Reduce rightward-shift

```suggestion
      if ( autoRoll === "no" ) return;
      const roll = await this.actor.rollNPCHitPoints({ chatMessage: autoRoll === "yes" });
      this.delta.updateSource({
        "system.attributes.hp": {
          max: roll.total,
          value: roll.total
        }
      });
```

module/documents/token.mjs
```diff
@@ -165,6 +165,28 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+
+  /** @inheritDoc */
+  async _preCreate(data, options, user) {
+    if ( (await super._preCreate(data, options, user)) === false ) return false;
+
+    if ( (this.actor?.type === "npc") && !this.actorLink && foundry.utils.getProperty(this.actor, "system.attributes.hp.formula")?.trim().length ) {
+      const autoRoll = game.settings.get("dnd5e", "autoRollNPCHP");
+      if ( autoRoll !== "no" ) {
+        const roll = await this.actor.rollNPCHitPoints({ chatMessage: autoRoll === "yes" });
+        await this.delta.updateSource({
```
Fyorl, line 180: `updateSource` is synchronous

```suggestion
        this.delta.updateSource({
```

Fyorl (changes requested):

Fyorl (approved):
Approved pending update to accommodate #5103

arbron:
@roth-michael The combat settings PR is merged, so if you want to tweak this one to add this setting to that application (probably inside a new "NPCs" section).

roth-michael:
Minor point, should the i18n for this (and crit settings) move into `SETTINGS.DND5E.COMBAT`?

arbron (approved):

PR #4545, opened 2024-10-16
[#2449, #2912] Re-design special traits config as actor sheet tab

### LINKED ISSUE #886: Suggestion: For bonuses and flags have the hover/title text on the special traits page  show the .data value of the field

Since there is no way to see the current value of bonuses/traits in the actor-flags app, it would be useful to show the current .dada value as title/hover text and leave the actual display as the ._data values of those fields.

### LINKED ISSUE #2449: [Feature Request] Consider moving Special Traits config to a separate character sheet tab.
I thought about asking for this mainly due to the number of times users with concantanated Global Bonuses from AEs, seem to overlook the Special Traits button on the character sheet.

Having a separate tabbed entry for Special Traits on the character sheet would raise awareness and seems like a more straightforward solution for adding all these options 🤔

### LINKED ISSUE #2912: Values displayed in the Special Traits window should reflect any modifications made by Active Effects on the actor
Active Effects on an actor that target `flags.dnd5e` and the various Special Traits in there will _function_ as intended — the values are changed, and the bonuses do apply to rolls — but the fields in the Special Traits window do not reflect these amended values.

CleanShot 2024-02-04 at 21 17 57
Here, you can see the fields being (properly) disabled by the AEs, but the values are still shown as defaults.

arbron:
Moves the actor flags & global bonuses from a separate application into an actor sheet tab. This should hopefully make those traits more discoverable. This also allows the traits to take advantage of the mode toggle to display derived values when in Play mode while leaving the source values editable in Edit mode.

Special Traits Tab

Closes #886
Closes #2449
Closes #2912


MaxPat931:
It may be a good idea to reconsider the Star icon, since that is shared by the Favorites feature as well?
Although, I can't think of another icon to represent Special Traits, perhaps the Favorites could change to `fa-bookmark`?
image

krbz999:
Not gonna lie, not sure I would feel great about having the Special Traits added to its own tab right on the sheet. It seems very out of place. These flags are all a random assortment of properties, some of which should even be deprecated.

Fyorl:
> Not gonna lie, not sure I would feel great about having the Special Traits added to its own tab right on the sheet. It seems very out of place. These flags are all a random assortment of properties, some of which should even be deprecated.

Having special traits as its own tab quite neatly solves the issues with active effect overrides. Is your main objection just that there are a lot of deprecated/sunset properties in there? Or that it's just not very cohesive?

krbz999:
> Having special traits as its own tab quite neatly solves the issues with active effect overrides. Is your main objection just that there are a lot of deprecated/sunset properties in there? Or that it's just not very cohesive?

Having one tab for inventory, one for features, one for spells, the main tab for skills and attributes, and one for an actual fillable form is not very cohesive, no.

less/v2/actors.less
```diff
@@ -262,6 +262,17 @@
     }
   }
 
+  /* ---------------------------------- */
+  /*  Biography                         */
```
Fyorl, line 266: Biography?

module/applications/actor/npc-sheet-2.mjs
```diff
@@ -191,6 +192,25 @@ export default class ActorSheet5eNPC2 extends ActorSheetV2Mixin(ActorSheet5eNPC)
...
+  _prepareFlags() {
+    const flags = super._prepareFlags();
+    const source = (this._mode === this.constructor.MODES.PLAY ? this.document : this.document._source);
+
+    flags.sections.unshift({
+      label: game.i18n.localize("DND5E.NPC.Label"),
+      fields: [{
+        field: this.document.system.schema.fields.traits.fields.important,
+        name: "system.traits.important",
+        value: source.system.traits.important
```
Fyorl, line 205: Make sure we use `createCheckboxInput`

module/applications/actor/sheet-flags.mjs
```diff
@@ -7,6 +7,15 @@ const { BooleanField } = foundry.data.fields;
  * An application class which provides advanced configuration for special character flags which modify an Actor.
  */
 export default class ActorSheetFlags extends BaseConfigSheet {
+  constructor(...args) {
+    foundry.utils.logCompatibilityWarning(
+      "The `ActorSheetFlags` application has been deprecated and replaced with a tab on the character sheet.",
+      { since: "DnD5e 4.1", until: "DnD5e 4.3" }
```
Fyorl, line 13: Needs updating

Fyorl (approved):

InnilToo:
Imma put my 2 cents in here to say, this will be even more confusing that it currently is.
Star = favorites
I always look for special traits in the details tab and then have to ask after 5 minutes where they disappeared to, to be pointed next to Short and Long Rest.
Where I look for it is Bottom right section of the character sheet where senses, immunities etc are. If I would get a vote I would put it there.
image

NMyers17:
I'm not sure what the purpose is of having a tab which is only functional whilst edit mode is active, however the tab is visible the entire time. The original star worked fine as it was tucked away, adding another tab onto the sheet adds more visual clutter to an already visually intensive character sheet. If it's best used as a tab the middle ground would be that it only appears when edit mode is toggled on, or in a place similar that @InnilToo has suggested.

This kind of feels like trying to streamline the design of the sheet but in the course of doing that actually making it less cohesive and annoying to use.

Fyorl:
> I'm not sure what the purpose is of having a tab which is only functional whilst edit mode is active

It's not only functional whilst edit mode is active. In play mode, the fields show their derived values.

PR #4547, opened 2024-10-17
[#4310] Re-design creature type config for V2 apps & styles

arbron:
Reworks `ActorTypeConfig` as `CreatureTypeConfig` to be used by actors and species using ApplicationV2 and V2 styles. Removed the functionality that changed the creature type to `custom` when focusing into the custom field because it was preventing tabbing between the creature type radio buttons and the subtype field without changing actor data.

Creature Type Config V2


Fyorl (changes requested):
Looks good. I've attached a small patch with some suggested tweaks. We wrap the radio button inside the label itself which means we don't have to use the `for` and `id` attributes. This is the same technique we use with the New Item dialog. I checked in the accessibility section of the devtools and the radio button is receiving the correct label even with the text wrapped in a `<span>`.

Then the rest are style tweaks to add a cursor and some hover feedback to make it clear that you can click anywhere on the row to select the radio button. It also pulls the labels and radio buttons together a bit more. I found that this:

image

Looked a little bit like Humanoid was selected due to their proximity and resemblance to the expertise proficiency toggle we use elsewhere.

[label-wrap.patch](https://github.com/user-attachments/files/17481619/label-wrap.patch)

Fyorl (approved):

PR #4549, opened 2024-10-17
[#4548] Save chat tray state between message renders

### LINKED ISSUE #4548: Updating chat messages resets chat tray open states
The opened state of chat message tray's isn't persisted across renders, causing them to jump open or closed if the chat message is updated.

arbron:
Stores the state of each chat tray present in a given chat message as well as their collapsed state, and then reuses these details when re-rendering the message.

Because of the way the chat log re-renders messages, the state is stored in the `updateMessage` method on a new custom `ChatLog5e` class.

Closes #4548

### DIFF
--- a/module/applications/chat-log.mjs
+++ b/module/applications/chat-log.mjs
@@ -1,0 +1,17 @@
+import ChatMessage5e from "../documents/chat-message.mjs";
+
+/**
+ * Custom implementation of the chat log to support saving tray states.
+ */
+export default class ChatLog5e extends ChatLog {
+  /** @inheritDoc */
+  async updateMessage(message, notify=false) {
+    const element = this.element[0].querySelector(`.message[data-message-id="${message.id}"]`);
+    if ( element ) message._trayStates = new Map([
+      ...Array.from(element.querySelectorAll(".card-tray"))
+        .map(t => [t.className.replace(" collapsed", ""), t.classList.contains("collapsed")]),
+      ...Array.from(element.querySelectorAll(ChatMessage5e.TRAY_TYPES.join(", "))).map(t => [t.tagName, t.open])
+    ]);
+    await super.updateMessage(message, notify);
+  }
+}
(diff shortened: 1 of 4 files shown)


module/documents/chat-message.mjs
```diff
@@ -4,6 +4,12 @@ import simplifyRollFormula from "../dice/simplify-roll-formula.mjs";
 
 export default class ChatMessage5e extends ChatMessage {
 
+  /**
+   * HTML tag names for chat trays that can open and close.
+   * @type {string[]}
+   */
+  static TRAY_TYPES = ["damage-application", "enchantment-application"];
```
etiquettestartshere, line 11: apologies if I am confused, but should this include effect-application?
```diff
- static TRAY_TYPES = ["damage-application", "enchantment-application"];
+ static TRAY_TYPES = ["damage-application", "effect-application"];
```
In my testing it seems the effects tray no longer responds to the `"autoCollapseChatTrays"` setting without the effect application element here, and currently the enchantment tray cannot be collapsed (at least from what I've seen in testing)

module/applications/chat-log.mjs
```diff
@@ -0,0 +1,17 @@
...
+ */
+export default class ChatLog5e extends ChatLog {
+  /** @inheritDoc */
+  async updateMessage(message, notify=false) {
+    const element = this.element[0].querySelector(`.message[data-message-id="${message.id}"]`);
+    if ( element ) message._trayStates = new Map([
+      ...Array.from(element.querySelectorAll(".card-tray"))
+        .map(t => [t.className.replace(" collapsed", ""), t.classList.contains("collapsed")]),
+      ...Array.from(element.querySelectorAll(ChatMessage5e.TRAY_TYPES.join(", "))).map(t => [t.tagName, t.open])
+    ]);
```
Fyorl, line 14: Just needs a slight tweak for v13 compatibility.

```suggestion
    const element = this.element instanceof HTMLElement ? this.element : this.element[0];
    const card = element.querySelector(`.message[data-message-id="${message.id}"]`);
    if ( card ) message._trayStates = new Map([
      ...Array.from(card.querySelectorAll(".card-tray"))
        .map(t => [t.className.replace(" collapsed", ""), t.classList.contains("collapsed")]),
      ...Array.from(card.querySelectorAll(ChatMessage5e.TRAY_TYPES.join(", "))).map(t => [t.tagName, t.open])
    ]);
```

Fyorl (approved):

Issue #4459, opened 2024-09-27
Chat Card Buttons add Redundant Clicks when Rolling Items from the Character Sheet


alaustin0:
# THE PROBLEM WITH THE CURRENT ROLLING WORKFLOW:
Currently, rolling items from the character sheet has a common complaint from a chunk of the community: it takes too many clicks to roll. Specifics on this opinion vary, but I can personally attest that almost all of my own players have made a comment like this at some point about the 5e system, and I would agree with it myself.

While “you have to click too many times to do something” is a simple problem on paper, explaining WHY that is the case is more complicated. I’m going to lay out some common workflows for rolling items to help illustrate where the alleged issue is occurring and why.

(The following workflows are all assuming that your character sheet is already open and on the Inventory tab, or the weapon is favorited. If not, there are even more clicks involved than listed, including opening up the Chat tab if it’s not already open. These potential extra clicks are worth noting, but I will not be counting those clicks going forward, because I want to paint the current workflow in the best possible light.)

Let’s go over some common workflows for rolling items from the sheet:

### Attack roll from a “basic” weapon
(read: a weapon that only has an Attack activity, isn’t Versatile, doesn’t consume ammo, etc)

This is one of the most common things you will do in the system.
- Click on the item on your character sheet, sending an item card to chat
- Move your cursor over to chat, and click Attack again
- Move your cursor over to the roll dialogue and click “Normal”

That’s three clicks in three different screen locations:

1. Character Sheet Click
2. Chat Card Attack Click
3. Attack Roll Dialogue Click **

(Clicks marked with ** are skippable by holding a modifier key.)

<img width="550" alt="basic-current" src="https://github.com/user-attachments/assets/b8eb3f93-f002-4a0d-a308-3a4d1b6fbef9">

While three clicks may not seem like a lot, that’s three clicks in three different locations just to make an attack roll on an item that only has one function: making attacks (and rolling damage, but that part comes from the Attack activity in this case).

We should note that the last click in this sequence can be skipped by holding “Shift” (which still requires an additional input to skip). So, at best, we are still working with two clicks + a keypress to roll an attack.

In this specific workflow, the Chat Card Attack Click is redundant, as clicking that button serves no purpose other than to actually prompt the Attack Roll Dialogue, which could have easily been prompted when we clicked on the item itself. The Chat Card in this case only serves to give you a Damage button. In terms of rolling an attack, the Chat Card is nothing but an extra click.

### Attack from an item that has multiple activities
[...]

PR #4555, opened 2024-10-18
[#4459] Add action that triggers automatically with activity use

arbron:
Adds a step after the end of the activity usage that allows specific activity types to trigger a primary action. For example, this allows an `AttackActivity` to trigger its attack roll dialog immediately after the item is activated, rather than requiring a separate click in chat.

- `AttackActivity`: Triggers attack dialog
- `DamgeActivity`: Triggers damage dialog
- `HealActivity`: Triggers healing dialog

Adjusts the `dnd5e.postUseActivity` hook to be cancelable, allowing modules to prevent these subsequent actions. The new method is intentionally not awaited allowing `use` to return immediately rather than waiting for these actions to resolve.

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -270,8 +270,9 @@ export default Base => class extends PseudoDocumentMixin(Base) {
      * @param {Activity} activity                     Activity being activated.
      * @param {ActivityUseConfiguration} usageConfig  Configuration data for the activation.
      * @param {ActivityUsageResults} results          Final details on the activation.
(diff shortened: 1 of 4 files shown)


module/documents/activity/mixin.mjs
```diff
@@ -897,6 +901,16 @@ export default Base => class extends PseudoDocumentMixin(Base) {
     results.templates = config.create?.measuredTemplate ? await this.#placeTemplate() : [];
   }
 
+  /* <><><><> <><><><> <><><><> <><><><> */
```
Fyorl, line 904: Check formatting

Fyorl (approved):

PR #4567, opened 2024-10-20
[#4566] Fix item use tracking

### LINKED ISSUE #4566: Item uses cannot be tracked on tokens
Broke again in v4 it seems

roth-michael:
Closes #4566 
`per` no longer exists - nor is it necessary to check anything except whether `max` exists (as far as I can tell).

### DIFF
--- a/module/applications/token-config.mjs
+++ b/module/applications/token-config.mjs
@@ -44,8 +44,8 @@ export default class TokenConfig5e extends TokenConfig {
   _addItemAttributes(attributes) {
     const actor = this.object?.actor;
     const items = actor?.items.reduce((arr, i) => {
-      const { per, max } = i.system.uses ?? {};
-      if ( per && max ) arr.push([i.getRelativeUUID(actor), i.name]);
+      const { max } = i.system.uses ?? {};
+      if ( max ) arr.push([i.getRelativeUUID(actor), i.name]);
       return arr;
     }, []) ?? [];
     if ( items.length ) {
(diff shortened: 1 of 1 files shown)


module/applications/token-config.mjs
```diff
@@ -44,8 +44,8 @@ export default class TokenConfig5e extends TokenConfig {
   _addItemAttributes(attributes) {
     const actor = this.object?.actor;
     const items = actor?.items.reduce((arr, i) => {
-      const { per, max } = i.system.uses ?? {};
-      if ( per && max ) arr.push([i.getRelativeUUID(actor), i.name]);
+      const { max } = i.system.uses ?? {};
+      if ( max ) arr.push([i.getRelativeUUID(actor), i.name]);
```
Fyorl, line 48: We have `Item5e#hasLimitedUses` for this now.

```suggestion
      if ( item.hasLimitedUses ) arr.push([i.getRelativeUUID(actor), i.name]);
```

Fyorl (changes requested):

module/applications/token-config.mjs
roth-michael: Updated PR to use that, thanks.

Fyorl (approved):

arbron:
@roth-michael Hey, could you comment on the original issue so I can assign it to you: https://github.com/foundryvtt/dnd5e/issues/4566

PR #4584, opened 2024-10-23
Add icons to DamagesConfig.

Fyorl:
image

### DIFF
--- a/less/v2/apps.less
+++ b/less/v2/apps.less
@@ -1315,6 +1315,30 @@ dialog.dnd5e2.application {
 }
 
 /* ---------------------------------- */
+
+.dnd5e2.damages-config.trait-selector {
+  .traits {
+    grid-template-columns: [icon] 16px [name] 2fr [proficiency] 1fr;
+    .trait-list li {
+      dnd5e-icon, i {
+        --icon-fill: var(--color-text-light-6);
+        color: var(--color-text-light-6);
+      }
+
+      > i {
+        font-family: var(--font-awesome);
+        font-weight: bold;
+        font-style: normal;
+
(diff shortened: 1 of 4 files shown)


templates/apps/parts/traits-list.hbs
```diff
@@ -9,6 +9,13 @@
     {{/if}}
     {{#each choices}}
     <li {{~#if category}} class="category"{{/if}}>
+        {{#if icon}}
+        {{#if icon.svg}}
+        <dnd5e-icon src="{{ icon.src }}"></dnd5e-icon>
+        {{else}}
+        <i class="{{ icon.src }}" inert></i>
+        {{/if}}
+        {{/if}}
```
arbron, line 18: We can save a few lines here.

```suggestion
        {{#if icon.svg}}
        <dnd5e-icon src="{{ icon.src }}"></dnd5e-icon>
        {{else if icon}}
        <i class="{{ icon.src }}" inert></i>
        {{/if}}
```

less/v2/apps.less
```diff
@@ -1314,6 +1314,30 @@ dialog.dnd5e2.application {
...
+    .trait-list li {
+      dnd5e-icon, i {
+        --icon-fill: var(--color-text-light-6);
+        color: var(--color-text-light-6);
+      }
+
+      > i {
+        font-family: var(--font-awesome);
+        font-weight: bold;
+        font-style: normal;
```
arbron, line 1331: I'm wondering if we want to center these to create a bit more balance:

Screenshot 2024-10-23 at 14 31 52
Screenshot 2024-10-23 at 14 31 41

arbron (approved):
Definitely spruces up the interface 👍

PR #4599, opened 2024-10-25
[#4596] Allow activity usage dialog to be skipped

### LINKED ISSUE #4596: Use Skip Dialog keybindings for skipping Activity Selection & Activity Usage Dialog windows
We use these keybindings for roll dialogs, but currently the Activity Selection dialog is hardcoded to use the "Shift" key and the Activity Usage Dialog doesn't support any skip button.

arbron:
Uses the various "Skip Dialog" keybindings to skip the activity usage dialog, using the activity with the default configuration. Does some modifications for the `ActivityChoiceDialog` to return an event so the usage dialog can be skipped from there also.

The same keybindings will now also skip the activity choice dialog rather than just being hardcoded to the Shift key. This does not yet attempt to change the hint displayed in the UI of that dialog.

Closes #4596

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -2,7 +2,7 @@ import ActivitySheet from "../../applications/activity/activity-sheet.mjs";
 import ActivityUsageDialog from "../../applications/activity/activity-usage-dialog.mjs";
 import AbilityTemplate from "../../canvas/ability-template.mjs";
 import { ConsumptionError } from "../../data/activity/fields/consumption-targets-field.mjs";
-import { formatNumber, getTargetDescriptors, localizeSchema } from "../../utils.mjs";
+import { areKeysPressed, formatNumber, getTargetDescriptors, localizeSchema } from "../../utils.mjs";
 import PseudoDocumentMixin from "../mixins/pseudo-document.mjs";
 
 /**
@@ -223,7 +223,6 @@ export default function ActivityMixin(Base) {
       const usageConfig = activity._prepareUsageConfig(usage);
 
       const dialogConfig = foundry.utils.mergeObject({
-        configure: true,
         applicationClass: this.metadata.usage.dialog
       }, dialog);
 
@@ -242,6 +241,8 @@ export default function ActivityMixin(Base) {
         },
         hasConsumption: usageConfig.hasConsumption
       }, message);
+
+      this._applyKeybindings(usageConfig, dialogConfig, messageConfig);
 
       /**
        * A hook event that fires before an activity usage is configured.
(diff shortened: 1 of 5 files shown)


module/documents/activity/mixin.mjs
```diff
@@ -425,6 +426,21 @@ export default Base => class extends PseudoDocumentMixin(Base) {
...
+   * Apply any keybindings that might affect usage process.
+   * @param {ActivityUseConfiguration} config       Configuration info for the activation.
+   * @param {ActivityDialogConfiguration} dialog    Configuration info for the configuration dialog.
+   * @param {ActivityMessageConfiguration} message  Configuration info for the chat message created.
+   * @protected
+   */
+  _applyKeybindings(config, dialog, message) {
+    dialog.configure ??= !areKeysPressed(config.event, "skipDialogNormal")
+      && !areKeysPressed(config.event, "skipDialogAdvantage")
+      && !areKeysPressed(config.event, "skipDialogDisadvantage");
```
Fyorl, line 439: Let's just use SHIFT for now, and keep the other modifiers in reserve, I could imagine them having some use in certain consumption contexts.

module/documents/item.mjs
```diff
@@ -792,14 +793,22 @@ export default class Item5e extends SystemDocumentMixin(Item) {
       let dialogConfig = dialog;
       let messageConfig = message;
       let activity = activities[0];
-      if ( (activities.length > 1) && !event?.shiftKey ) activity = await ActivityChoiceDialog.create(this);
+      const skipDialog = areKeysPressed(event, "skipDialogNormal")
+        || areKeysPressed(event, "skipDialogAdvantage")
+        || areKeysPressed(event, "skipDialogDisadvantage");
```
Fyorl, line 798: Same comment as above: Just use SHIFT for now, and keep the other modifiers in reserve.

Fyorl (changes requested):
Need a solution for this before we merge:

> This does not yet attempt to change the hint displayed in the UI of that dialog.

Or maybe we pull that part out of this PR for now, since the rest of it seems important for #4555

module/documents/activity/mixin.mjs
arbron: The trouble is, if we don't handle the other buttons then you won't be able to skip the usage dialog and roll the subsequent primary action with advantage/disadvantage, because that initial event is what is forwarded to the primary action.

Fyorl: I think we need to reconsider then, since we've overloaded the SHIFT modifier here. The control hint is that you can SHIFT+Click to skip the activity selection dialog, but actually, with these changes, you not only skip activity selection, but you also skip consumption, and you force the resulting attack roll from #4555 to be a normal roll, no?

Having the modifier fast-forward the entire process is *probably* what we want, but I need to think of some way to communicate that more clearly in the control hint. Let me know if you have any suggestions.

PR #4600, opened 2024-10-25
[#4031] Add Forward activity

### LINKED ISSUE #4031: Activity Type: Forward
An activity that, when activated, triggers another activity on the same item but with parts of the triggered activity overridden. This would be able to support things like abilities that you can use once for free and then cast other times using an ability. So the main activity would have `1/1 uses` that it consumes, and then the forward activity will target the first activity but bypass its consumption, instead defining its own consumption but allowing everything else on the original activity to be used.

arbron:
Adds the Forward activity type that is able to trigger another activity on the same item with different consumption configuration. This allows for spells that have 1 free use per long rest and can otherwise be used with spells slots to be easily configured.

Screenshot 2024-10-25 at 13 25 24

Closes #4031

### Todo
- [ ] Update `ItemGrant` & `ItemChoice` advancement types to use forward activity, rather than duplicating existing ones when adding a free casting to prepared spells


module/documents/activity/forward.mjs
```diff
@@ -0,0 +1,48 @@
+import ForwardSheet from "../../applications/activity/forward-sheet.mjs";
+import ForwardActivityData from "../../data/activity/forward-data.mjs";
+import ActivityMixin from "./mixin.mjs";
+
+/**
+ * Activity for casting a spell from another item.
```
Fyorl, line 6: Check comment

PR #4614, opened 2024-10-29
[#2138] Re-implement `rollAttack` using new rolling APIs

### LINKED ISSUE #2138: Refactor roll pipeline

arbron:
Reworks `rollAttack` to use the new `build` method on `D20Roll` and implement a custom configuration application to support ammunition, attack modes, and mastery.

In the process this moves the preparation of ammunition options into `WeaponData` and reorders the logic for ammunition, attack mode, and mastery options to be easier to follow.

Closes #2138

### DIFF
--- a/module/applications/dice/attack-configuration-dialog.mjs
+++ b/module/applications/dice/attack-configuration-dialog.mjs
@@ -1,0 +1,43 @@
+import D20RollConfigurationDialog from "./d20-configuration-dialog.mjs";
+
+/**
+ * @typedef {BasicRollConfigurationDialogOptions} AttackRollConfigurationDialogOptions
+ * @property {FormSelectOption[]} ammunitionOptions - Ammunition that can be used with the attack.
+ * @property {FormSelectOption[]} attackModeOptions - Different modes of attack.
+ * @property {FormSelectOption[]} masteryOptions - Available masteries for the attacking weapon.
+ */
+
+/**
+ * Extended roll configuration dialog that allows selecting attack mode, ammunition, and weapon mastery.
+ */
+export default class AttackRollConfigurationDialog extends D20RollConfigurationDialog {
+  /** @override */
+  static DEFAULT_OPTIONS = {
+    ammunitionOptions: [],
+    attackModeOptions: [],
+    masteryOptions: []
+  };
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _prepareConfigurationContext(context, options) {
+    context = await super._prepareConfigurationContext(context, options);
+    const optionsFields = [
+      { key: "attackMode", label: "DND5E.ATTACK.Mode.Label", options: this.options.attackModeOptions },
+      { key: "ammunition", label: "DND5E.CONSUMABLE.Type.Ammunition.Label", options: this.options.ammunitionOptions },
+      { key: "mastery", label: "DND5E.WEAPON.Mastery.Label", options: this.options.masteryOptions }
+    ];
+    for ( const { key, label, options } of optionsFields.reverse() ) {
+      if ( options?.length ) context.fields.unshift({
+        field: new foundry.data.fields.StringField({ label: game.i18n.localize(label) }),
+        name: key,
+        options,
+        value: this.config[key]
+      });
+    }
+    return context;
+  }
+}
--- a/module/applications/dice/attack-configuration-dialog.mjs
+++ b/module/applications/dice/attack-configuration-dialog.mjs
@@ -2,9 +2,9 @@ import D20RollConfigurationDialog from "./d20-configuration-dialog.mjs";
 
 /**
  * @typedef {BasicRollConfigurationDialogOptions} AttackRollConfigurationDialogOptions
- * @property {FormSelectOption[]} ammunitionOptions - Ammunition that can be used with the attack.
- * @property {FormSelectOption[]} attackModeOptions - Different modes of attack.
- * @property {FormSelectOption[]} masteryOptions - Available masteries for the attacking weapon.
+ * @property {FormSelectOption[]} ammunitionOptions  Ammunition that can be used with the attack.
+ * @property {FormSelectOption[]} attackModeOptions  Different modes of attack.
+ * @property {FormSelectOption[]} masteryOptions     Available masteries for the attacking weapon.
  */
 
 /**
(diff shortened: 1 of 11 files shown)


module/applications/dice/attack-configuration-dialog.mjs
```diff
@@ -0,0 +1,43 @@
+import D20RollConfigurationDialog from "./d20-configuration-dialog.mjs";
+
+/**
+ * @typedef {BasicRollConfigurationDialogOptions} AttackRollConfigurationDialogOptions
+ * @property {FormSelectOption[]} ammunitionOptions - Ammunition that can be used with the attack.
+ * @property {FormSelectOption[]} attackModeOptions - Different modes of attack.
+ * @property {FormSelectOption[]} masteryOptions - Available masteries for the attacking weapon.
```
Fyorl, line 7:
```suggestion
 * @property {FormSelectOption[]} ammunitionOptions  Ammunition that can be used with the attack.
 * @property {FormSelectOption[]} attackModeOptions  Different modes of attack.
 * @property {FormSelectOption[]} masteryOptions     Available masteries for the attacking weapon.
```

module/applications/dice/attack-configuration-dialog.mjs
```diff
@@ -0,0 +1,43 @@
...
+
+  /** @inheritDoc */
+  async _prepareConfigurationContext(context, options) {
+    context = await super._prepareConfigurationContext(context, options);
+    const optionsFields = [
+      { key: "attackMode", label: "DND5E.ATTACK.Mode.Label", options: this.options.attackModeOptions },
+      { key: "ammunition", label: "DND5E.CONSUMABLE.Type.Ammunition.Label", options: this.options.ammunitionOptions },
+      { key: "mastery", label: "DND5E.WEAPON.Mastery.Label", options: this.options.masteryOptions }
+    ];
+    for ( const { key, label, options } of optionsFields.reverse() ) {
```
Fyorl, line 33: I can't see `optionsFields` used anywhere else so we should be able to just define them in the order we need them rather than calling `reverse`, no?

module/data/activity/attack-data.mjs
```diff
@@ -228,32 +228,28 @@ export default class AttackActivityData extends BaseActivityData {
...
+    const ammo = this.actor?.items.get(config?.ammunition)?.system;
+    const { parts, data } = CONFIG.Dice.BasicRoll.constructParts({
+      mod: this.attack.ability !== "none" ? rollData.mod : null,
+      prof: weapon.prof?.term,
+      bonus: this.attack.bonus,
+      weaponMagic: weapon.magicAvailable ? weapon.magicalBonus : null,
+      ammoMagic: ammo?.magicAvailable ? ammo.magicalBonus : null,
+      actorBonus: this.actor?.system.bonuses?.[this.actionType]?.attack,
+      situational
+    }, this.getRollData());
```
Fyorl, line 249:
```suggestion
    }, rollData);
```

module/data/item/weapon.mjs
```diff
@@ -285,6 +285,23 @@ export default class WeaponData extends ItemDataModel.mixin(
...
+   */
+  get ammunitionOptions() {
+    if ( !this.parent.actor ) return [];
+    return this.parent.actor.items
+      .filter(i => (i.type === "consumable") && (i.system.type.value === "ammo")
+        && (!this.ammunition?.type || (i.system.type.subtype === this.ammunition.type)))
+      .map(i => ({
+        value: i.id, label: `${i.name} (${i.system.quantity})`, item: i, disabled: !i.system.quantity
+      }))
+      .sort((lhs, rhs) => lhs.label.localeCompare(rhs.label, game.i18n.lang));
```
Fyorl, line 300: `itemTypes` is cached now so it saves us iterating over the entire items collection here:

```suggestion
    return this.parent.actor.itemTypes.consumable
      .filter(i => (i.system.type.value === "ammo")
        && (!this.ammunition?.type || (i.system.type.subtype === this.ammunition.type)))
      .map(i => ({
        value: i.id, label: `${i.name} (${i.system.quantity})`, item: i, disabled: !i.system.quantity
      }))
      .sort((lhs, rhs) => lhs.label.localeCompare(rhs.label, game.i18n.lang));
```

module/documents/activity/attack.mjs
```diff
@@ -277,9 +250,42 @@ export default class AttackActivity extends ActivityMixin(AttackActivityData) {
...
+   * @param {Partial<D20RollConfiguration>} [initialRoll]  Initial roll passed to the rolling method.
+   * @param {D20RollProcessConfiguration} process          Configuration for the entire rolling process.
+   * @param {D20RollConfiguration} config                  Configuration for a specific roll.
+   * @param {FormDataExtended} [formData]                  Any data entered into the rolling prompt.
+   * @param {number} index                                 Index of the roll within all rolls being prepared.
+   */
+  _buildAttackConfig(initialRoll, process, config, formData, index) {
+    if ( formData?.has("ammunition") ) process.ammunition = formData.get("ammunition");
+    if ( formData?.has("attackMode") ) process.attackMode = formData.get("attackMode");
+    if ( formData?.has("mastery") ) process.mastery = formData.get("mastery");
```
Fyorl, line 272: Why do we need to update `process` here? Shouldn't it be immutable?

Fyorl (changes requested):

module/documents/activity/attack.mjs
arbron: Hmm, right now `getAttackData` fetches information on the ammunition from the attack process config (and potentially attack mode in a future version). I suppose we could pass those options directly rather than passing in the whole config object.

Fyorl (approved):

PR #4616, opened 2024-10-30
[#4061, #4613] Add Cast overrides, activity changes via enchanting

### LINKED ISSUE #4613: Add syntax for applying enchantment changes directly to activities
Currently the only way to change an activity using an enchantment is to specifically target it by ID or by using one of the hard-coded exceptions that were grandfathered in from the pre-activity days. This isn't super useful for generic enchantments. The system should implement a system for applying changes to activities based on their type:

`activity.attack.attack.bonus` | `OVERRIDE` | `4`
`activity.attack.attack.flat` | `OVERRIDE` | `true`

This active effect would apply a change to any `AttackActivity` on an item, giving it a flat to-hit bonus of +4. The first part `activity` indicates that this is a change to the item's activities, then the first `attack` is the type of activity to change. Everything that follows is the key path to the property on the activity to change (e.g. `attack.bonus`).

arbron:
Makes the DC & Attack overrides on the Cast activity functional. It does this through the Spell Changes enchantment which uses a new syntax to apply directly to all Attack & Save activities on the spell.

The new syntax allows for modifying all activities of a certain type using the change key like `activity.attack.attack.bonus`, where `activity.attack` indicates this should change all attack activities and `attack.bonus` being the key path to the field to change. At the same time this also fixes some issues with applying enchantment changes directly to individual activities by ID.

Closes #4613

### DIFF
--- a/module/documents/activity/cast.mjs
+++ b/module/documents/activity/cast.mjs
@@ -141,15 +141,31 @@ export default class CastActivity extends ActivityMixin(CastActivityData) {
   getSpellChanges() {
     const changes = [];
     const source = this.toObject();
+
+    // Override spell details
     for ( const type of ["activation", "duration", "range", "target"] ) {
       if ( !this[type].override ) continue;
       const data = source[type];
       delete data.override;
       changes.push({ key: `system.${type}`, mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE, value: JSON.stringify(data) });
     }
+
+    // Remove ignored properties
     for ( const property of this.spell.properties ) {
       changes.push({ key: "system.properties", mode: CONST.ACTIVE_EFFECT_MODES.ADD, value: `-${property}` });
     }
+
+    // Set challenge overrides
+    const challenge = this.spell.challenge;
+    if ( challenge.override && challenge.attack ) changes.push(
+      { key: "activity.attack.attack.bonus", mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE, value: challenge.attack },
+      { key: "activity.attack.attack.flat", mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE, value: true }
+    );
+    if ( challenge.override && challenge.save ) changes.push(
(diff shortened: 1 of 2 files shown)


module/documents/activity/cast.mjs
```diff
@@ -141,15 +141,31 @@ export default class CastActivity extends ActivityMixin(CastActivityData) {
...
     }
+
+    // Set challenge overrides
+    const challenge = this.spell.challenge;
+    if ( challenge.override && challenge.attack ) changes.push(
+      { key: "activity.attack.attack.bonus", mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE, value: challenge.attack },
+      { key: "activity.attack.attack.flat", mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE, value: true }
+    );
+    if ( challenge.override && challenge.save ) changes.push(
+      { key: "activity.save.save.dc.calculation", mode: CONST.ACTIVE_EFFECT_MODES.OVERRIDE, value: "" },
```
Fyorl, line 165: Since this is a 'special' key that doesn't operate as normal, I wonder if we want to indicate that in a way that's a bit more explicit. I was thinking of maybe using a sigil like `$activity` or `!activity`, but I would like to see more use-cases for these sorts of special keys before committing to something like that. Also, there are some special sigils used for serialisation in various parts of the core API that might complicate things.

For now maybe just a bit more explicit like `activitiesOfType.attack...`, `activitiesOfType.save...`, etc.? What do you think?

Fyorl (changes requested):

module/documents/activity/cast.mjs
arbron: Hmm, we could do something like `activities[attack].attack…` perhaps, to keep a clear separation between the types and the following key path.

krbz999:
Makes sense from a code perspective, but for end users that's gonna be hella confusing that some things are dot-separated strings and some, for reasons they will never need to know, have square brackets or use different symbols.

Confusion on when to use `system` and when to use `@` already comes up a lot.

Fyorl (approved):

PR #4621, opened 2024-10-31
Add bastion tab to character sheet.

Fyorl:
image

The occupant drag-drop UI could do with an indication of which slot the occupant is being dragged to because they're quite small and easy to drop into the wrong slot, and we're missing drag-drop reordering of facilities, but I need to move on for now.


templates/items/details/details-facility.hbs
```diff
@@ -9,12 +9,27 @@
...
+    {{!-- Properties --}}
+    <div class="form-group split-group">
+        <label>{{ localize "DND5E.FACILITY.Properties" }}</label>
+        <div class="form-fields">
+
+            {{!-- Level --}}
+            {{ formField fields.level value=source.level label="DND5E.Level" localize=true classes="label-top" }}
+
+            {{!-- Order --}}
+            {{ formField fields.order value=source.order classes="label-top" options=orders.available blank="" }}
```
Fyorl, line 29: Moved this back under a 'properties' split group alongside 'level'. We have the order that the facility is capable of executing, which is what this is, and it's an intrinsic property of the facility. And we also have the order that the facility is currently executing, which might be the aforementioned order, the Maintain order, or nothing. That order is now part of the 'progress' schema and I put it inside the 'orders' fieldset alongside the progress fields.

module/applications/actor/character-sheet-2.mjs
```diff
@@ -53,7 +53,8 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
     { tab: "features", label: "DND5E.Features", icon: "fas fa-list" },
     { tab: "spells", label: "TYPES.Item.spellPl", icon: "fas fa-book" },
     { tab: "effects", label: "DND5E.Effects", icon: "fas fa-bolt" },
-    { tab: "biography", label: "DND5E.Biography", icon: "fas fa-feather" }
+    { tab: "biography", label: "DND5E.Biography", icon: "fas fa-feather" },
+    { tab: "bastion", label: "Bastion", icon: "fas fa-chess-rook" }
```
arbron, line 57: Needs localization

module/applications/actor/character-sheet-2.mjs
```diff
@@ -82,14 +83,16 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
     const isUpdate = (context === "update") || (context === "updateActor");
     const hp = foundry.utils.getProperty(data ?? {}, "system.attributes.hp.value");
     if ( isUpdate && (hp === 0) ) this._toggleDeathTray(true);
+    const [bastion] = this.element.find('nav.tabs [data-tab="bastion"]');
+    if ( bastion ) bastion.toggleAttribute("hidden", this.object.system.details.level < 5);
```
arbron, line 87: Doesn't seem to actually be hiding the tab.

module/applications/actor/character-sheet-2.mjs
```diff
@@ -315,7 +330,7 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
     // Apply special context menus for items outside inventory elements
     const featuresElement = html[0].querySelector(`[data-tab="features"] ${this.options.elements.inventory}`);
     if ( featuresElement ) {
-      new ContextMenu5e(html, ".pills-lg [data-item-id], .favorites [data-item-id]", [], {
+      new ContextMenu5e(html, ".pills-lg [data-item-id], .favorites [data-item-id], .facility[data-item-id]", [], {
```
arbron, line 333: We should add facilities to the `canExpand` method at the bottom of the sheet so they don't get that context menu option.

module/applications/actor/character-sheet-2.mjs
```diff
@@ -478,6 +495,25 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
...
+   * @protected
+   */
+  _onDeleteOccupant(event) {
+    event.stopPropagation();
+    const { facilityId } = event.target.closest("[data-facility-id]")?.dataset ?? {};
+    const { prop } = event.target.closest("[data-prop]")?.dataset ?? {};
+    const { index } = event.target.closest("[data-index]")?.dataset ?? {};
+    const facility = this.actor.items.get(facilityId);
+    if ( !facility || !prop || (index === undefined) ) return;
+    const value = foundry.utils.getProperty(facility, prop).filter((_, i) => i != index);
```
arbron, line 511:
```suggestion
    const value = foundry.utils.getProperty(facility, prop).filter((_, i) => i !== index);
```

templates/actors/tabs/character-bastion.hbs
```diff
@@ -0,0 +1,158 @@
...
+            <div class="slot order-slot {{#unless executing}}empty{{/unless}}">
+                {{#if executing}}
+                <dnd5e-icon src="{{ executing }}"></dnd5e-icon>
+                {{/if}}
+            </div>
+            {{/if}}
+        </div>
+    </div>
+    {{#if hirelings.length}}
+    <div class="slots facility-occupants hirelings" data-prop="system.hirelings">
```
arbron, line 122:
```suggestion
    <div class="slots facility-occupants hirelings" data-prop="system.hirelings.value">
```

templates/actors/tabs/character-bastion.hbs
```diff
@@ -0,0 +1,158 @@
...
+        {{#each hirelings}}
+            {{> ".occupant" type="hireling" index=@key }}
+        {{/each}}
+    </div>
+    {{/if}}
+    {{#if defenders.length}}
+    <div class="slots facility-occupants defenders" data-prop="system.defenders">
```
arbron, line 129:
```suggestion
    <div class="slots facility-occupants defenders" data-prop="system.defenders.value">
```

arbron (changes requested):

module/applications/actor/character-sheet-2.mjs
Fyorl: Not able to reproduce this

module/applications/actor/character-sheet-2.mjs
Fyorl: We're comparing a string and number here so using strict equals doesn't work

module/applications/actor/character-sheet-2.mjs
arbron: Are you running V13? Because it hides properly in V13 but is visible in V12:

Screenshot 2024-11-01 at 08 36 42

module/applications/actor/character-sheet-2.mjs
arbron: Linter doesn't like it :(. Could always to `i !== Number(index)`

arbron (approved):
Other than the hiding issue on V12, looks solid.

PR #4624, opened 2024-10-31
[#4623] Add `createScrollFromCompendiumSpell` that uses Cast

### LINKED ISSUE #4623: Convert `createScrollFromSpell` to use the Cast activity
Will probably only be a good idea if creating a spell scroll from a spell in a compendium, but in that case it would be nice to use the cast activity rather than copying over activities like is done currently.

arbron:
It probably isn't safe to use the Cast activity for spells scrolls created from world or items on actors, but for ones in compendiums it should be safe. This adds a new method that is called if the provided spell is inside a pack.

Closes #4623

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1,3 +1,4 @@
+import ActivityChoiceDialog from "../applications/activity/activity-choice-dialog.mjs";
 import AdvancementManager from "../applications/advancement/advancement-manager.mjs";
 import AdvancementConfirmationDialog from "../applications/advancement/advancement-confirmation-dialog.mjs";
 import CreateScrollDialog from "../applications/item/create-scroll-dialog.mjs";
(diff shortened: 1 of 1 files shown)


module/documents/item.mjs
```diff
@@ -1680,6 +1647,149 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
+    let scrollUuid;
+    const id = CONFIG.DND5E.spellScrollIds[spell.system.level];
+    if ( foundry.data.validators.isValidId(id) ) {
+      scrollUuid = game.packs.get(CONFIG.DND5E.sourcePacks.ITEMS).index.get(id).uuid;
+    } else {
+      scrollUuid = id;
+    }
+    const scrollItem = await fromUuid(scrollUuid);
+    const scrollData = scrollItem.toObject();
+    delete scrollData._id;
```
Fyorl, line 1695:
```suggestion
    const scrollData = game.items.fromCompendium(scrollItem);
```

Alternatively, we call this at the end, i.e.

```js
return new this(game.items.fromCompendium(spellScrollData));
```

Fyorl (approved):

PR #4635, opened 2024-11-03
[#4634] Fix Wing Attack damage of Adult Green Dragon

### LINKED ISSUE #4634: Adult Green Dragon Wing Attack damage bonus is wrong
Should be `2d6 + 6` but is `2d6 + 1`

strongpauly:
Fixes #4634

### DIFF
--- a/packs/_source/monsters/dragon/adult-green-dragon.json
+++ b/packs/_source/monsters/dragon/adult-green-dragon.json
@@ -2297,7 +2297,7 @@
           "base": {
             "number": 2,
             "denomination": 6,
-            "bonus": 6,
+            "bonus": "",
             "types": [
               "bludgeoning"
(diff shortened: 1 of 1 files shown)


packs/_source/monsters/dragon/adult-green-dragon.json
```diff
@@ -2297,7 +2297,7 @@
           "base": {
             "number": 2,
             "denomination": 6,
-            "bonus": "",
+            "bonus": 6,
```
arbron, line 2300: No need to change this (in fact, this damage data isn't actually used, base weapon damage isn't added to saves, only attacks).

```suggestion
            "bonus": "",
```

arbron (changes requested):

PR #4651, opened 2024-11-06
Implement Order Activity for Facility Items

Fyorl:
Apologies for the size of this PR, since it encompasses a considerable chunk of ordering automation for facilities. General breakdown is as follows:

- Added the Order Activity type. I also added the concept of `configurable` activities. Non-configurable activities are hidden from the activities list and cannot be created via the UI. We can consider alternative ways to handle this one.
- Facilities have various Order Activities injected during data preparation.
- Added an `OrderUsageDialog` that allows for order execution to be configured by the user. By default, each order allows for days and gold costs to be configured. Certain other orders have additional configuration and associated automation.
- Added the `chooseActivity` option to `Item5e#use` that forces the activity choice dialog in cases where it would normally be skipped for only have one activity. SHIFT+Clicking will still bypass the prompt even with this option. Not sure this parameter is quite necessary anymore.
- Order usage chat cards allow the facility owner to pay for the costs associated with the order (if any) either automatically, or manually. Added `Actor5e#deductCurrency` API method to facilitate the former. It has several options based on potential user preference with how currency might be automatically deducted. Let me know if this feels like a reasonable API to you.
- Costs for crafting an item are automatically calculated when issuing the Craft order. Added `ItemDataModel#getCraftCost` API method to facilitate this. Let me know if this feels like a reasonable API to you.

The following Order types have some automation associated with them:
- Enlarge
- Craft
- Harvest
- Trade

The rest require manual resolution.

image

image

image

image

image

image

image


module/applications/activity/activity-usage-dialog.mjs
```diff
@@ -170,7 +170,6 @@ export default class ActivityUsageDialog extends Dialog5e {
       case "concentration": return this._prepareConcentrationContext(context, options);
       case "consumption": return this._prepareConsumptionContext(context, options);
       case "creation": return this._prepareCreationContext(context, options);
-      case "footer": return this._prepareFooterContext(context, options);
```
Fyorl, line 173: Seems this was already being called in `super`, so no need to call it again here.

module/applications/activity/activity-usage-dialog.mjs
```diff
@@ -461,9 +460,9 @@ export default class ActivityUsageDialog extends Dialog5e {
    * Perform any pre-processing of the form data to prepare it for updating.
    * @param {SubmitEvent} event          Triggering submit event.
    * @param {FormDataExtended} formData  Data from the submitted form.
-   * @returns {object}
+   * @returns {Promise<object>}
    */
-  _prepareSubmitData(event, formData) {
+  async _prepareSubmitData(event, formData) {
```
Fyorl, line 465: Had a need for this method to be `async`, but it seemed safe to make the change given that the two places it was currently called were also `async` contexts.

module/config.mjs
```diff
@@ -2035,6 +2035,112 @@ DND5E.currencies = {
...
 preLocalize("currencies", { keys: ["label", "abbreviation"] });
 
+/* -------------------------------------------- */
+
+/**
+ * @typedef CraftingConfiguration
+ * @property {CraftingCostsMultiplier} consumable     Discounts for crafting a magical consumable.
+ * @property {Record<string, CraftingCosts>} magic    Magic item crafting costs.
+ * @property {CraftingCostsMultiplier} mundane        Multipliers for crafting mundane items.
+ * @property {CraftingCosts} potionOfHealing          Crafting costs for a potion of healing.
```
Fyorl, line 2045: Such an awkward exception. Not sure if it's worth bothering with really.

module/data/activity/base-activity.mjs
```diff
@@ -70,7 +70,7 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
         blank: false, required: true, readOnly: true, initial: () => this.metadata.type
       }),
       name: new StringField({ initial: undefined }),
-      img: new FilePathField({ initial: undefined, categories: ["IMAGE"] }),
+      img: new FilePathField({ initial: undefined, categories: ["IMAGE"], base64: false }),
```
Fyorl, line 73: Unrelated. Since we can't do sanitisation within activities, any `FilePathField`s are required to have `base64: false`.

module/documents/item.mjs
```diff
@@ -606,6 +606,32 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
+      switch ( this.type ) {
+        case "weapon": identifier = CONFIG.DND5E.weaponIds[baseItem]; break;
+        case "tool": identifier = CONFIG.DND5E.toolIds[baseItem]; break;
+        case "equipment":
+          if ( value === "shield" ) identifier = CONFIG.DND5E.shieldIds[baseItem];
+          else identifier = CONFIG.DND5E.armorIds[baseItem];
+          break;
+      }
+      this.system.type.identifier = identifier;
+    }
```
Fyorl, line 633: Maybe these should move to their associated templates? But then we need to make sure every item type is calling these template preparation methods manually, and I wasn't sure if that was the convention we decided on.

module/config.mjs
krbz999: We could add a configuration object, mapping together cost and uuids, then shove Potion of Healing in there. More work but less awkward and future proofs it in case more of these outliers would appear in future content.

Fyorl: Yeah that's probably better. That solves my issue of which UUID is the 'canonical' Potion of Healing. SRD? PHB? What about the SRD 5.2? With a configuration object we would just add all of the UUIDs. We could also consider using the `identifier`. A potion of healing is a potion of healing, regardless of where it came from or who created it.

krbz999: Agree that using `identifier` would be better. Also means the '24 PHB (and whichever other content modules) will not require adding any data.

templates/actors/tabs/character-bastion.hbs
```diff
@@ -99,7 +102,7 @@
     </div>
     {{/if}}
     <div class="facility-header">
-        <div class="name-stacked">
+        <div class="name-stacked" data-action="useFacility">
```
arbron, line 105: The entire `facility-header` section highlights on hover, but only the name portion actually does something when clicked. Should probably adjust the hover CSS to only apply to the name.

The highlight is also applied to a non-editable actor where clicking does nothing.

templates/chat/order-activity-card.hbs
```diff
@@ -0,0 +1,21 @@
+<div class="dnd5e2 chat-card activation-card">
+
+    {{!-- Description --}}
+    <p class="order-description">{{{ description }}}</p>
```
arbron, line 4: I wonder if we could sneak the order icon onto the card somehow, maybe large and mostly transparent on the background? Not a priority.

module/config.mjs
arbron: It sounds like a perfect use of `identifier` 👍

module/config.mjs
```diff
@@ -2035,6 +2035,112 @@ DND5E.currencies = {
 };
 preLocalize("currencies", { keys: ["label", "abbreviation"] });
 
+/* -------------------------------------------- */
+
+/**
+ * @typedef CraftingConfiguration
+ * @property {CraftingCostsMultiplier} consumable     Discounts for crafting a magical consumable.
+ * @property {Record<string, CraftingCosts>} magic    Magic item crafting costs.
```
arbron, line 2043:
```suggestion
 * @property {Record<string, CraftingCosts>} magic    Magic item crafting costs by rarity.
```

module/config.mjs
```diff
@@ -2035,6 +2035,112 @@ DND5E.currencies = {
...
+ * @property {CraftingCostsMultiplier} mundane        Multipliers for crafting mundane items.
+ * @property {CraftingCosts} potionOfHealing          Crafting costs for a potion of healing.
+ * @property {Record<number, CraftingCosts>} scrolls  Crafting costs for spell scrolls.
```
arbron, line 2046:
```suggestion
 * @property {Record<number, CraftingCosts>} scrolls  Crafting costs for spell scrolls by level.
```

module/documents/activity/order.mjs
```diff
@@ -0,0 +1,278 @@
...
+ * An activity for issuing an order to a facility.
+ */
+export default class OrderActivity extends ActivityMixin(OrderActivityData) {
+  /* -------------------------------------------- */
+  /*  Model Configuration                         */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  static metadata = Object.freeze(foundry.utils.mergeObject(super.metadata, {
+    configurable: false,
```
arbron, line 35: Does this make more sense here or in `CONFIG.activityTypes`. Right now we have `hidden` in the configuration object which controls whether activities are displayed in the creation list, so perhaps this makes sense alongside it?

module/applications/activity/order-usage-dialog.mjs
```diff
@@ -0,0 +1,361 @@
...
+   * Prepare render context for the craft section.
+   * @param {ApplicationRenderContext} context  Render context.
+   * @param {HandlebarsRenderOptions} options   Render options.
+   * @protected
+   */
+  async _prepareCraftContext(context, options) {
+    context.craft = {
+      legend: game.i18n.localize(`DND5E.FACILITY.Orders.${this.activity.order}.present`),
+      item: {
+        field: new DocumentUUIDField({ type: "Item" }),
```
arbron, line 67: Getting the same `type: undefined` error here.

```suggestion
        field: new DocumentUUIDField(),
```

module/data/item/facility.mjs
```diff
@@ -86,6 +107,39 @@ export default class FacilityData extends ItemDataModel.mixin(ActivitiesTemplate
...
+
+  /** @inheritDoc */
+  prepareBaseData() {
+    super.prepareBaseData();
+
+    if ( this.type.value === "basic" ) this.enlargeable = true;
+    if ( this.size === "vast" ) this.enlargeable = false;
+
+    // Activities
+    if ( this.type.value === "special" ) this._createOrderActivity("dnd5eFacOrder", this.order);
```
arbron, line 137:
```suggestion
    if ( (this.type.value === "special") && this.order ) this._createOrderActivity("dnd5eFacOrder", this.order);
```

module/documents/activity/order.mjs
```diff
@@ -0,0 +1,278 @@
...
+    }
+  }, { inplace: false }));
+
+  /* -------------------------------------------- */
+  /*  Activation                                  */
+  /* -------------------------------------------- */
+
+  /** @override */
+  shouldHideChatButton(button, message) {
+    return !this.actor.isOwner;
```
arbron, line 54: `ChatMessage#_displayChatActionButtons` should already be hiding all buttons from non-owners/non-GMs.

module/documents/activity/order.mjs
```diff
@@ -0,0 +1,278 @@
...
+      buttons: buttons.length ? buttons : null,
+      description: game.i18n.format("DND5E.FACILITY.Use.Description", {
+        order: game.i18n.localize(`DND5E.FACILITY.Orders.${this.order}.inf`),
+        link: this.item.toAnchor().outerHTML,
+        facilityType: facilityType.toLocaleLowerCase(game.i18n.lang)
+      })
+    };
+  }
+
+  /* -------------------------------------------- */
```
arbron, line 251:
```suggestion
  /* -------------------------------------------- */
  /*  Event Listeners and Handlers                */
  /* -------------------------------------------- */
```

module/documents/actor/actor.mjs
```diff
@@ -356,6 +356,64 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+   * @param {number} amount                          The amount of currency.
+   * @param {string} denomination                    The currency's denomination.
+   * @param {object} [options]
+   * @param {boolean} [options.recursive=false]      Deduct currency from containers as well as the base Actor. TODO
+   * @param {"high"|"low"} [options.priority="low"]  Prioritize higher denominations before lower, or vice-versa.
+   * @param {boolean} [options.exact=true]           Prioritize deducting the requested denomination first.
+   * @throws {Error} If the Actor does not have sufficient currency.
+   * @returns {Promise<Actor5e>|void}
+   */
+  deductCurrency(amount, denomination, options={}) {
```
arbron, line 372: We should probably implement these on the `CurrencyManager` application which also houses `convertCurrency` and `transferCurrency` implementations. That would make it easier to add it to the currency manager UI in the future if we wanted to.

module/documents/item.mjs
arbron: That is generally the pattern I've been adopting, like we have `prepareDescriptionData` for `ItemDescriptionTemplate` already. We just have to make sure things don't immediately break if a module sub-type doesn't call them.

lang/en.json
```diff
@@ -1306,6 +1306,7 @@
 "DND5E.DeathSaveRoll": "Roll a Death Saving Throw",
 "DND5E.DeathSavingThrow": "Death Saving Throw",
 "DND5E.DeathSaveUnnecessary": "You do not need to roll death saves because you have a positive number of hit points or have already reached 3 successes or failures.",
+"DND5E.DeductCurrencyError": "Failed to deduct {amount}{denomination} from {name}. Insufficient funds.",
```
arbron, line 1309: If we move `deduceCurrency` into `CurrencyManager`, then this should probably also move to `DND5E.CurrencyManager.Warning.InsufficientFunds`.

arbron (changes requested):

templates/chat/order-activity-card.hbs
Fyorl: I like it. Will try to get to it

module/config.mjs
Fyorl: Alright I have removed the `potionOfHealing` property and added the `exceptions` property, which is an object of `CraftingCosts` keyed by Item identifier. Actually taking these exceptions into account when calculating crafting costs is still a TODO though because I am out of time.

module/documents/item.mjs
Fyorl: Moved price preparation to `PhysicalItemTemplate` and hardened code paths to not assume `price.valueInGP` exists. I moved the `type.identifier` preparation to `WeaponData`, `ToolData`, etc. since it is dependent on the actual Item type so it made sense to just do the preparation in that exact context.

module/documents/activity/order.mjs
Fyorl: I was sure I tested it, and the buttons were still visible to a user that did not own the Actor. I will leave it for now, but we can remove it if we do some testing and determine it's definitely redundant.

module/applications/item/item-sheet-2.mjs
```diff
@@ -135,7 +135,9 @@ export default class ItemSheet5e2 extends ItemSheetV2Mixin(ItemSheet5e) {
     }));
 
     // Activities
-    context.activities = (activities ?? []).map(({ _id: id, name, img, sort }) => ({
+    context.activities = (activities ?? []).filter(a => {
+      return a.metadata?.configurable;
```
arbron, line 139:
```suggestion
      return CONFIG.DND5E.activityTypes[a.type]?.configurable !== false;
```

arbron (approved):
Just a missing configuration update, otherwise looking solid.

module/applications/item/item-sheet-2.mjs
Fyorl: oops

PR #4652, opened 2024-11-06
Bastion configuration settings

Fyorl:
image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -569,6 +569,22 @@
 "DND5E.Background": "Background",
 "DND5E.BackgroundAdd": "Add Background",
 "DND5E.BackgroundName": "Background Name",
+
+"DND5E.Bastion": {
+  "Duration": {
+    "Label": "Bastion Turn Duration (days)"
+  },
+  "Enabled": {
+    "Label": "Enable Bastion Functionality",
+    "Hint": "When enabled, player characters will have access to a new Bastion tab on their character sheets when reaching level 5, and will be able to build facilities and issue orders to them."
+  },
+  "Configuration": {
+    "Name": "Bastions",
+    "Label": "Configure Bastions",
+    "Hint": "Various configuration options for the Bastion system as presented in the DMG 2024."
+  }
+},
+
 "DND5E.Biography": "Biography",
 "DND5E.BiographyPublic": "Public Biography",
 "DND5E.BiopgrahyPublicEdit": "Edit Public Biography",
--- a/lang/en.json
+++ b/lang/en.json
@@ -571,6 +571,11 @@
 "DND5E.BackgroundName": "Background Name",
 
 "DND5E.Bastion": {
+  "Configuration": {
+    "Name": "Bastions",
+    "Label": "Configure Bastions",
+    "Hint": "Various configuration options for the Bastion system as presented in the DMG 2024."
+  },
   "Duration": {
     "Label": "Bastion Turn Duration (days)"
   },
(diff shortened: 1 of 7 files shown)


lang/en.json
```diff
@@ -569,6 +569,22 @@
 "DND5E.Background": "Background",
 "DND5E.BackgroundAdd": "Add Background",
 "DND5E.BackgroundName": "Background Name",
+
+"DND5E.Bastion": {
```
arbron, line 573: We can move `DND5E.FACILITY.Bastion` over here as `DND5E.Bastion.Label`.

module/applications/_module.mjs
```diff
@@ -13,6 +13,7 @@ export * as shared from "./shared/_module.mjs";
 
 export {default as Accordion} from "./accordion.mjs";
 export {default as Award} from "./award.mjs";
+export {default as BastionConfig} from "./bastion.mjs";
```
arbron, line 16: Perhaps we should add a `settings` folder to hold these settings apps (and `CompendiumBrowserSourceConfig`.

module/applications/bastion.mjs
```diff
@@ -0,0 +1,79 @@
...
+ */
+export class BastionSetting extends foundry.abstract.DataModel {
+  /** @override */
+  static defineSchema() {
+    return {
+      duration: new NumberField({
+        required: true, positive: true, integer: true, initial: 7, label: "DND5E.Bastion.Duration.Label"
+      }),
+      enabled: new BooleanField({
+        required: true, label: "DND5E.Bastion.Enabled.Label", hint: "DND5E.Bastion.Enabled.Hint"
```
arbron, line 75: Enable by default? Or will the DMG enable it?

arbron (approved):

module/applications/actor/character-sheet-2.mjs
```diff
@@ -84,7 +84,8 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
     const hp = foundry.utils.getProperty(data ?? {}, "system.attributes.hp.value");
     if ( isUpdate && (hp === 0) ) this._toggleDeathTray(true);
     const [bastion] = this.element.find('nav.tabs [data-tab="bastion"]');
-    if ( bastion ) bastion.toggleAttribute("hidden", this.object.system.details.level < 5);
+    const { enabled } = game.settings.get("dnd5e", "bastionConfiguration");
+    if ( bastion ) bastion.toggleAttribute("hidden", (this.actor.system.details.level < 5) || !enabled);
```
krbz999, line 88: Would be good to have this (the level requirement) in the config rather than hardcoded. It doesn't need to be a setting, just a simple numeric property. To make it simple, any module or script changing the level requirement should also be required to change the localization to match.

Fyorl: We have `CONFIG.DND5E.facilities.advancement` which is closely tied to this number. Would using the min of that variable work?

krbz999: Didn't realize we had that. Seems good.

module/applications/bastion.mjs
Fyorl: Even in the DMG it's optional. I know there will be issues with discoverability, but I'm assuming more people will not use bastions than will, so opt-in seems the sensible default.

module/applications/actor/character-sheet-2.mjs
Fyorl: OK I will make that change then

module/applications/bastion.mjs
krbz999: I think all options so far are opt-out barring some of the more obscure ones (like Honor/Sanity).
I think this would be on the same level as "feats instead of ASI" in the legacy rules, which is opt-out.

module/applications/_module.mjs
Fyorl: Good idea. Will do it later though

module/applications/bastion.mjs
Fyorl: I am ok with having it opt-out, but feats and multiclassing 'optional' rules were in the PHB. I think this would be more like having Flanking enabled by default if we ever implemented it.

arbron (approved):

PR #4655, opened 2024-11-06
Add build order for basic facilities.

Fyorl:
image

image


arbron (approved):

arbron (approved):
Looks like unbuilt facilities also have the "Enlarge" order in the item sheet, which doesn't seem like it should be available until after building has completed.

PR #4658, opened 2024-11-07
Bastion turn automation

Fyorl:
image

### DIFF
--- a/module/documents/actor/bastion.mjs
+++ b/module/documents/actor/bastion.mjs
@@ -1,0 +1,401 @@
+import { formatNumber } from "../../utils.mjs";
+
+/**
+ * @typedef BastionTurnResult
+ * @property {string} [order]             The order that was completed, if any.
+ * @property {number} [gold]              Gold generated.
+ * @property {BastionTurnItem[]} [items]  Items created.
+ */
+
+/**
+ * @typedef BastionTurnItem
+ * @property {string} uuid      The UUID of the generated Item.
+ * @property {number} quantity  The quantity of items generated.
+ */
+
+/**
+ * A singleton class that manages global Bastion activity.
+ */
+export default class Bastion {
+  /**
+   * The template for the chat card summary of a bastion turn.
+   * @type {string}
+   */
+  static SUMMARY_TEMPLATE = "systems/dnd5e/templates/chat/bastion-turn-summary.hbs";
+
+  /* -------------------------------------------- */
+  /*  Public API                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Advance all bastions by a turn.
+   * @returns {Promise<void>}
+   */
+  async advanceAllBastions() {
+    // TODO: Should this advance game.time?
+    const { duration } = game.settings.get("dnd5e", "bastionConfiguration");
+    const haveBastions = game.actors.filter(a => (a.type === "character") && a.itemTypes.facility.length);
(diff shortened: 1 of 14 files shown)


module/documents/actor/bastion.mjs
```diff
@@ -0,0 +1,401 @@
...
+  /* -------------------------------------------- */
+
+  /**
+   * Handle dragging an item created as part of order completion.
+   * @param {DragEvent} event    The initiating drag event.
+   * @param {Item5e} item        The created item.
+   * @param {object} [updates]   Updates to apply to the Item.
+   */
+  #onDragItem(event, item, updates={}) {
+    // TODO: Need some way to mark the item as 'claimed' when it is dropped onto an Actor sheet.
```
arbron, line 297: We could pass the message ID with the data transfer, and then in `_onDropItemCreate` we can relay that back into the options for `createDocuments` and then have a hook in bastions watching for that item's creation. A bit of a roundabout process, but it would ensure that the message is updated properly even if the player dropping doesn't have permissions on it.

arbron (approved):
Everything seems to be working. Just left a few of my thoughts where you had `TODO`s.

module/documents/actor/bastion.mjs
```diff
@@ -0,0 +1,401 @@
...
+  /* -------------------------------------------- */
+  /*  Public API                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Advance all bastions by a turn.
+   * @returns {Promise<void>}
+   */
+  async advanceAllBastions() {
+    // TODO: Should this advance game.time?
```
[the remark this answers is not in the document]
Fyorl: Yeah a checkbox probably for the best for now until we have more rigorous game time automation. If we add that though, then should it trigger a long rest and item recharge as well? More checkboxes?

module/documents/actor/bastion.mjs
Fyorl: I suppose there has to be at least *someone* online who has permission to modify the chat message, right? Or am I misunderstanding the proposal?

arbron: Well, supposing the messages are created by the GM when they hit Advance Bastion Turn, then they will need to be the one to update it. So players would drag the item to their sheet and send out the `createDocument`, and then the GM would listen for `onCreateDocument` with the message option and update the chat message when it sees that.

module/documents/actor/bastion.mjs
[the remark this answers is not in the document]
arbron: Or we could go the other way and have "Advance Bastion Turn" as a checkbox when doing a long rest from the Group Actor, which already advances time. So GMs can use the normal "Advance Bastion Turn" button if their players long rest themselves, or the checkbox in the group rest if they want to handle it.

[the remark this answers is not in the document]
Fyorl: Yeah that is a good idea, we should definitely have an 'advance bastion turn' checkbox in the group rest dialog. Leaving that as another TODO though for now.

module/documents/actor/bastion.mjs
Fyorl: Yep, got it. I was only considering cases like the players logging in between sessions to do book-keeping and attempting to claim items. But that's an issue with various other things like concentration as well.

Not looking like I'll be able to get to it at this rate, but the TODO is there to come back to.

PR #4660, opened 2024-11-07
Add change activity for Gardens

Fyorl:


Fyorl:
> On using the "Advance Bastion Turn", I'm wondering if we want to change the message from "No orders issued to special facilities this turn." to "No special facilities have completed their orders this turn." if ongoing orders are in effect. We could even show progress for ongoing orders, but that might be a future UI improvement.

It's important to note if no orders were issued, because that might be because the characters were absent, in which case a bastion event is triggered. But I agree we could still note that progress was made on ongoing orders so that the message doesn't get confused for 'nothing was done'.

PR #4663, opened 2024-11-08
Implement bastion attack resolution.

Fyorl:
image


arbron (approved):
Looking good, but if you have a facility in your favorites you can still activate it even when it's disabled. Might also be good to have a tooltip over disabled facilities to indicate why they can't be interacted with.

Fyorl:
> Looking good, but if you have a facility in your favorites you can still activate it even when it's disabled. Might also be good to have a tooltip over disabled facilities to indicate why they can't be interacted with.

A tooltip makes sense. Will need to stop being lazy with `pointer-events: none` though, so I will add it to the list of TODOs. I don't think you can drag facilities to favourites anymore. You could in earlier iterations, is it possible you had some lingering favourites from earlier rounds of testing?

PR #4695, opened 2024-11-11
Update Starter Heroes with fresh items and add favorites

### LINKED ISSUE #4394: Refresh Starter Heroes embedded docs
Many of the Starter Heroes items are now out of sync with the updated compendium items, refreshing the embedded docs from their source would bring them up to date. Some Starter Heroes also have weapon configurations that are no longer needed (e.g. thrown and melee Javelins, Off Hand Handaxes, etc)

MaxPat931:
Closes #4394 
I hope I did this right for updating the Starter Heroes. I refreshed all of their items and put them through fresh advancements from the classes and added Favorites so everything should be up to date.

### DIFF
--- a/module/data/item/equipment.mjs
+++ b/module/data/item/equipment.mjs
@@ -168,7 +168,7 @@ export default class EquipmentData extends ItemDataModel.mixin(
   /** @inheritDoc */
   prepareBaseData() {
     super.prepareBaseData();
-    if ( this.armor.base === undefined ) this.armor.base = this.armor.value ??= 0;
+    this.armor.base = this.armor.value = (this._source.armor.value ?? 0);
   }
 
   /* -------------------------------------------- */
--- a/system.json
+++ b/system.json
@@ -2,14 +2,14 @@
   "id": "dnd5e",
   "title": "Dungeons & Dragons Fifth Edition",
   "description": "A system for playing the fifth edition of the world's most popular role-playing game in the Foundry Virtual Tabletop environment.",
-  "version": "4.2.1",
+  "version": "4.2.2",
   "compatibility": {
     "minimum": "12.331",
     "verified": "13.335"
   },
   "url": "https://github.com/foundryvtt/dnd5e/",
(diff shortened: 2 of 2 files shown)


Fyorl:
Noticed the actors have duplicated art credits in their biographies. It can happen if you import into the World then save back to the compendium. So I think the credits, and actor and token images all need to be cleared (they will be re-added automatically). Github is having some trouble with these diffs, but nothing else jumped out at me from a brief check, they look good.

MaxPat931:
Ah yep that's exactly what I did, I'll see about making those changes later today, thanks!

MaxPat931:
I was able to remove the credits in the bios, but couldnt set the image paths null in the ui

Fyorl:
> I was able to remove the credits in the bios, but couldnt set the image paths null in the ui

That's fine, I should be able to handle that. I did notice that some items lost their `compendiumSource`s. For example Quillathe seems to have lost the `compendiumSource` from their High Elf race, their Ranger class, as well as from their racial and class features. Their equipment seems to have kept the `compendiumSource`. How did you refresh these items? Deleted the old ones and dragged new ones from the SRD?

MaxPat931:
Just did a quick test, it looks like when you use the "Add Class/Race/etc" pill and choose from the Compendium Browser, it does not set the compendiumSource, but dragging in from the compendium does

Cleric from the Browser, Druid dragged on:
image

Fyorl:
> Just did a quick test, it looks like when you use the "Add Class/Race/etc" pill and choose from the Compendium Browser, it does not set the compendiumSource, but dragging in from the compendium does
> 
> Cleric from the Browser, Druid dragged on: image

Great, thank you for investigating, I wondered if it was some bug.

Fyorl:
GitHub automatically closed this, not sure why. Either way, I have fixed the bug that was discovered here. I have also cleaned the baked `img` and `prototypeToken` data, and repaired the `compendiumSource`s. So I will go ahead and get this merged. Thank you for putting this together.

PR #4699, opened 2024-11-11
[#4628] Replace roll parts rather than rebuilding config entirely

### LINKED ISSUE #4628: 4.1: d20 roll ConfigurationDialogs do not make use of any added parts added programmatically
This works in the damage roll and hit dice roll configuration dialogs:
```js
dialog.rolls[0].parts.push("1d100");
dialog.rebuild();
```

The same does not work for any of the d20 dialogs.

arbron:
To remove an issue with programmatically altering roll config for skill/tool checks & attack rolls, this modifies the `buildConfig` method to swap part data in-place, rather than completely rebuild the parts & data objects. To do this easily this introduces `BasicRoll#swapParts` which is designed to swap part data and part keys in place without losing the original ordering of the parts.

### Todo
- [ ] Also apply to attack rolls

Closes #4628

### DIFF
--- a/module/dice/basic-roll.mjs
+++ b/module/dice/basic-roll.mjs
@@ -107,6 +107,41 @@ export default class BasicRoll extends Roll {
   /* -------------------------------------------- */
 
   /**
+   * @typedef {object} RollPartSwapData
+   * @property {string} key                          Name of the roll part to replace.
+   * @property {any} value                           Value to insert in place of the current value.
+   * @property {string} [newKey]                     Change the key name when replacing the data.
+   * @property {"beginning"|"end"} [position="end"]  Where new parts should be inserted into the parts list.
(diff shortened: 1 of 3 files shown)


Fyorl (commented):
This might be the only solution but I'd like to try to keep our auto-computed and arbitrarily added parts separate so that they can always be re-computed from scratch, rather than merging parts eagerly and then having to swap them in-place. I feel like if we could incorporate the `initialRoll` concept into this framework then we'd be even better off than when we started.

I think this needs a bit more thought, so we might have to pull this PR from the milestone for now so that it doesn't hold up the rest of the other fixes.

arbron:
Closed in favor of https://github.com/foundryvtt/dnd5e/pull/5116

PR #4719, opened 2024-11-13
[#4714] Fix error when using chat action for deleted activity

### LINKED ISSUE #4714: Chat buttons after issuing Barracks enlarge order throw errors
repro:
- add Barracks facility to actor
- click Barracks facility
- issue Enlarge order
- click any resulting button in chat

image

arbron:
Fixes an issue caused by enlarging special bastion facilities to their maximum size, which caused the Enlarge activity to be removed, and thus no subsequent chat actions could be triggered.

This fix covers the immediate problem, but the chat actions will still not be usable if the player refreshes or the chat message is re-rendered, because this fix only works because the deleted activity is still held in memory as part of the event handler.

Closes #4714


arbron:
@Fyorl This is a temporary fix that solves some of the problem, but there should probably be a better fix involving how we handle the ephemeral activities on bastion facilities.

arbron:
Closed in favor of https://github.com/foundryvtt/dnd5e/pull/4750

Issue #4708, opened 2024-11-13
Redesign Advancement applications using ApplicationV2 & V2 styles


arbron:
- [x] Advancement Manager (https://github.com/foundryvtt/dnd5e/pull/4765)
- [x] Advancement Config
  - [x] AbilityScoreImprovementConfig (https://github.com/foundryvtt/dnd5e/pull/4766)
  - [x] HitPointsConfig (https://github.com/foundryvtt/dnd5e/pull/4709)
  - [x] ItemChoiceConfig (https://github.com/foundryvtt/dnd5e/pull/4736)
  - [x] ItemGrantConfig (https://github.com/foundryvtt/dnd5e/pull/4722)
  - [x] ScaleValueConfig (https://github.com/foundryvtt/dnd5e/pull/4767)
  - [x] SizeConfig (https://github.com/foundryvtt/dnd5e/pull/4768)
  - [x] TraitConfig (https://github.com/foundryvtt/dnd5e/pull/4769)
- [x] Advancement Flow
  - [x] AbilityScoreImprovementFlow (https://github.com/foundryvtt/dnd5e/pull/6256)
  - [x] HitPointsFlow (https://github.com/foundryvtt/dnd5e/pull/6239)
  - [x] ItemChoiceFlow (https://github.com/foundryvtt/dnd5e/pull/6249)
  - [x] ItemGrantFlow (https://github.com/foundryvtt/dnd5e/pull/6241)
  - [x] ScaleValueFlow (https://github.com/foundryvtt/dnd5e/pull/6055)
  - [x] SizeFlow (https://github.com/foundryvtt/dnd5e/pull/6055)
  - [x] SubclassFlow (https://github.com/foundryvtt/dnd5e/pull/6053)
  - [x] TraitFlow (https://github.com/foundryvtt/dnd5e/pull/6053)
- [x] AdvancementConfirmationDialog (https://github.com/foundryvtt/dnd5e/pull/6725)
- [x] AdvancementMigrationDialog (https://github.com/foundryvtt/dnd5e/pull/5227)
- [x] AdvancementSelection (https://github.com/foundryvtt/dnd5e/pull/5228)

JPMeehan:
Bugfix request as part of this - Handlebars and textareas are tricky, could the whitespace please be fixed for the hints? I believe `{{~> 'my-partial}}` for partials and `{{~value~}}` for value references would help resolve the tab/indentation issues.

arbron:
[...]

PR #4736, opened 2024-11-15
[#4708] Redesign `ItemChoiceConfig` for V2, implement multi-column

arbron:
Converts `ItemChoiceConfig` to use `ApplicationV2` and the V2 styles. Adds support for multi-column layout within advancement config apps.

Config V2 Choose Items


lang/en.json
```diff
@@ -256,6 +256,65 @@
...
     "Starting": "Hit Points at 1st Level"
   },
+  "ItemChoice": {
+    "Choose": "choose {count}",
+    "Chosen": "Chosen: {current} of {max}",
+    "Details": "Choice Details",
+    "DropHint": "Drop Items here to add them to the list that can be chosen with this advancement.",
+    "FIELDS": {
+      "allowDrops": {
+        "hint": "Should players be able to drop their own choices into this advancement?",
```
krbz999, line 266: Gonna infuriate Unsoluble.

module/applications/api/application-v2-mixin.mjs
```diff
@@ -57,6 +57,28 @@ export default function ApplicationV2Mixin(Base) {
...
+        if ( !config.containerClasses ) continue;
+        const element = this.element.querySelector(`[data-application-part="${part}"]`);
+        if ( !element ) continue;
+        const className = config.containerClasses.join(" ");
+        if ( !containers[className] ) {
+          const div = document.createElement("div");
+          div.classList.add(...config.containerClasses);
+          containers[className] = div;
+          element.replaceWith(div);
+          div.append(element);
```
Fyorl, line 74: Already done two lines below

```suggestion
```

module/applications/api/application-v2-mixin.mjs
```diff
@@ -57,6 +57,28 @@ export default function ApplicationV2Mixin(Base) {
...
 
+    /** @inheritDoc */
+    _onFirstRender(context, options) {
+      super._onFirstRender(context, options);
+      const containers = {};
+      for ( const [part, config] of Object.entries(this.constructor.PARTS) ) {
+        if ( !config.containerClasses ) continue;
+        const element = this.element.querySelector(`[data-application-part="${part}"]`);
+        if ( !element ) continue;
+        const className = config.containerClasses.join(" ");
```
Fyorl, line 68: Using the classes for this seems fragile and also a conflation of concerns. If we want to do something like this I think it would be better to have a separate `container` identifier and `containerClasses`.

templates/advancement/item-choice-config-levels.hbs
```diff
@@ -0,0 +1,31 @@
+<fieldset class="card levels">
+    <legend>{{ localize "DND5E.ADVANCEMENT.ItemChoice.FIELDS.choices.label" }}</legend>
+    <div class="header">
+        <span class="level-header">
+            {{ localize "DND5E.AbbreviationLevel" }}
+        </span>
+        <span id="{{ partId }}-choices" class="choices-header">
```
Fyorl, line 7: Seems weird to give `<span>`s an `id`. If we're just doing this for the sake of labelling the `<input>`s, can we put an `aria-label` on them directly? Or is there some other benefit to using `aria-labelledby` here?

Fyorl (changes requested):

Fyorl (approved):

Issue #3958, opened 2024-08-08
Add proper plural localizations for time periods and other units


arbron:
Currently the duration for a spell's label will say something like "Duration: 1 hours". We should add proper plural forms to these localization values so it will be able to say "Duration: 1 hour" instead.

PR #4749, opened 2024-11-18
[#3958] Improve localization of rest messages

arbron:
Updates the rest message to select the proper plural form of Hit Points and Hit Dice to match the number recovered.

Also adjusts the flavor to automatically display the rest time in the message flavor, rather than relying on hard-coded labels for each rest variant. This ensures the flavor will properly display the rest time even if it doesn't exactly match the default values.

In order to support this as well as improve other time-based localization in the future, this PR adds two new utlity methods:
- `convertTime` converts a time from one set of units to another. It will also attempt to pick the best unit automatically if the `to` units are not defined, selecting the largest unit that can fully represent the value (with some wiggle room so it will display 90 minutes, rather than 1.5 hours).
- `formatTime` converts a time value and units into a string for display, relying on the system-provided plural localization if available and if not using Javascript's standard unit formatting for numbers.

### DIFF
--- a/module/utils.mjs
+++ b/module/utils.mjs
@@ -123,6 +123,34 @@ export function formatText(value) {
 /* -------------------------------------------- */
 
 /**
+ * A helper function that formats a time in a human-readable format.
+ * @param {number} value         Time to display.
+ * @param {string} unit          Units as defined in `CONFIG.DND5E.timeUnits`.
+ * @param {object} [options={}]  Formatting options passed to `formatNumber`.
+ * @returns {string}
+ */
+export function formatTime(value, unit, options={}) {
+  options.maximumFractionDigits ??= 0;
+  options.unitDisplay ??= "long";
+  const config = CONFIG.DND5E.timeUnits[unit];
+  if ( config?.counted ) {
+    if ( (options.unitDisplay === "narrow") && game.i18n.has(`${config.counted}.narrow`) ) {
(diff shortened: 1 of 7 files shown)


module/utils.mjs
```diff
@@ -426,10 +454,44 @@ export function getSceneTargets() {
 /*  Conversions                                 */
 /* -------------------------------------------- */
 
+/**
+ * Convert the provided time value to another unit. If not final unit is provided, then will convert it to the largest
```
Fyorl, line 458:
```suggestion
 * Convert the provided time value to another unit. If ot final unit is provided, then will convert it to the largest
```

module/utils.mjs
```diff
@@ -426,10 +454,44 @@ export function getSceneTargets() {
...
+ * unit that can still represent the value as a whole number.
+ * @param {number} value                    The time being converted.
+ * @param {string} from                     The initial unit as defined in `CONFIG.DND5E.timeUnits`.
+ * @param {object} [options={}]
+ * @param {boolean} [options.combat=false]  Use combat units when auto-selecting units, rather than normal units.
+ * @param {string} [options.to]             The final units, is explicitly provided.
```
Fyorl, line 464:
```suggestion
 * @param {string} [options.to]             The final units, if explicitly provided.
```

Fyorl (approved):

module/utils.mjs
arbron: I think you deleted the wrong letter there :)

PR #4750, opened 2024-11-18
[#4713, #4714] Disallow using same order twice, fix enlarge bug

### LINKED ISSUE #4713: Disallow issuing an order that is already issued
Currently you can issue an Enlarge order on a Facility that is already progressing on said order.

### LINKED ISSUE #4714: Chat buttons after issuing Barracks enlarge order throw errors

arbron:
Adds a new `canUse` method to activities. The `OrderActivity` will not allow usage of an activity if the facility is already executing the same order or if it is the enlarge order and the facility is already as maximum size.

Closes #4713
Closes #4714

### DIFF
--- a/module/documents/activity/order.mjs
+++ b/module/documents/activity/order.mjs
@@ -56,9 +56,23 @@ export default class OrderActivity extends ActivityMixin(OrderActivityData) {
   get canUse() {
     return super.canUse
       // Don't allow usage if facility is already executing the same order
-      && (this.parent.progress.order !== this.order)
+      && !this.inProgress
       // Enlarge order cannot be executed if facility is already maximum size
       && ((this.order !== "enlarge") || (this.parent.size !== "vast"));
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Is this order currently in the process of being executed by its facility?
+   * @type {boolean}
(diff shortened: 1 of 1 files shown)


module/documents/activity/order.mjs
```diff
@@ -48,6 +48,19 @@ export default class OrderActivity extends ActivityMixin(OrderActivityData) {
...
 
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  get canUse() {
+    return super.canUse
+      // Don't allow usage if facility is already executing the same order
+      && (this.parent.progress.order !== this.order)
```
Fyorl, line 59: #4713 is not really a bug. It's intentional that you can re-issue the same order in case you made a mistake and want to change some parameter of it. We can probably disallow it if `progress.value > 0`, and anyone who wants to change their order at that point has to manually edit the facility to reset progress.

Fyorl (changes requested):

Fyorl (approved):

PR #4765, opened 2024-11-23
[#4708] Convert `AdvancementManager` to V2 app & styles

arbron:
Converts `AdvancementManager` to use `ApplicationV2` and V2 system styles. Because old V1 advancement flow applications are embedded within this application, there are some V1 classes added to ensure that styling of old flows aren't completely broken.

Flow V2 Manager

This also takes a number of soft-private properties and methods on the manager and makes them hard-private.

### DIFF
--- a/module/applications/advancement/advancement-manager.mjs
+++ b/module/applications/advancement/advancement-manager.mjs
@@ -29,70 +29,95 @@ import Application5e from "../api/application.mjs";
  * @param {Actor5e} actor        Actor on which this advancement is being performed.
  * @param {object} [options={}]  Additional application options.
  */
-export default class AdvancementManager extends Application {
+export default class AdvancementManager extends Application5e {
   constructor(actor, options={}) {
     super(options);
-
-    /**
-     * The original actor to which changes will be applied when the process is complete.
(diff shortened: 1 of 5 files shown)


module/applications/advancement/advancement-manager.mjs
```diff
@@ -327,7 +370,7 @@ export default class AdvancementManager extends Application {
    * @param {string} classItem      Class being changed.
    * @param {number} levelDelta     Levels by which to increase or decrease the class.
    * @returns {AdvancementManager}  Manager with new steps.
-   * @private
+   * @protected
```
Fyorl, line 373: Surely a mistake this was annotated `@private` in the first place, since it doesn't have an underscore? Looks like public API to me.

Fyorl (approved):
Worth re-evaluating the public and protected API for this class. For example, I imagine it's useful for a subclass to be able to read the `#stepIndex` or `#advancing` state. I'm not exactly sure what sort of use cases we have in mind for `AdvancementManager` subclasses though. Is it more superficial things like 're-skinning', or should a subclass be able to do things like modify how a user moves through the process entirely?

PR #4766, opened 2024-11-23
[#4708] Redesign `AbilityScoreImprovementConfig` for V2

arbron:
Converts `AbilityScoreImprovementConfig` to `ApplicationV2` and the V2 system styles. Adds `PseudoDocumentSheet#submit` to be able to programmatically submit pseudo document forms.

Config V2 ASI


arbron:
@Fyorl I was also considering a horizontal arrangement, not sure which I prefer (this does give more header space):

Screenshot 2025-01-10 at 08 21 46

Fyorl (approved):
Definitely prefer the vertical design. I wasn't sure about how the raised, legend-less fieldset looked with all the content floating in the middle so I did a style pass and also swept up a few issues that I would have otherwise left comments on.

image

If you really hate it let me know. I had some other ideas too. Otherwise feel free to merge.

PR #4767, opened 2024-11-23
[#4708] Redesign `ScaleValueConfig` for V2

arbron:
Converts `ScaleValueConfig` to `ApplicationV2` and the V2 system styles. In order to take advantage of new form rendering this now automatically creates all the necessary fields in the scale value list using the schema data, rather than manually handling each scale value type.

Config V2 Scale Value


module/applications/advancement/scale-value-config.mjs
```diff
@@ -47,8 +71,11 @@ export default class ScaleValueConfig extends AdvancementConfig {
     let levels = Array.fromRange(CONFIG.DND5E.maxLevel + 1);
     if ( ["class", "subclass"].includes(this.advancement.item.type) ) levels = levels.slice(1);
     return levels.reduce((obj, level) => {
-      obj[level] = { placeholder: this._formatPlaceholder(lastValue), value: null };
-      const value = this.advancement.configuration.scale[level];
+      const value = this.advancement.configuration.scale[level]?.clone();
```
Fyorl, line 74: We didn't have to clone before?

templates/advancement/scale-value-config-levels.hbs
```diff
@@ -0,0 +1,27 @@
+<fieldset class="card levels" data-type="{{ configuration.data.type }}">
+    <legend>{{ localize "DND5E.ADVANCEMENT.ScaleValue.FIELDS.scale.label" }}</legend>
+    <div class="header">
+        <span class="level-level">
+            {{ localize "DND5E.AbbreviationLevel" }}
+        </span>
+        {{#each type.fields}}
+        {{#unless options.hidden}}
+        <span id="{{ @root.partId }}-{{ @key }}" class="{{ @key }}-header">
```
Fyorl, line 9: This `id` seemingly not used for anything.

Fyorl (approved):

module/applications/advancement/scale-value-config.mjs
arbron: I believe we previously had a small issue where `_mergeScaleValues` was making changes to the data on the advancement rather than just preparing changes for preview.

PR #4769, opened 2024-11-23
[#4708] Redesign `TraitConfig` for V2 app and styles

arbron:
Converts `TraitConfig` to `ApplicationV2` and the V2 system styles.

Config V2 Traits

Modifies `dnd5e-checkbox` to not throw `input` or `change` events when programatically setting the `checked` value, because that causes the form's `onChange` handler to trigger which causes the form to re-render, leading to an infinite loop. This brings `dnd5e-checkbox` in line with standard HTML checkboxes which only trigger these events on user interaction.


templates/advancement/trait-config-choices.hbs
```diff
@@ -0,0 +1,31 @@
+<fieldset class="selected-trait">
+    <legend>
+        {{ localize "DND5E.ADVANCEMENT.Trait.FIELDS.choices.label" }}
+        <a data-action="addChoice" data-tooltip="DND5E.ADVANCEMENT.Trait.Action.AddChoice"
+           aria-label="{{ localize 'DND5E.ADVANCEMENT.Trait.Action.AddChoice' }}">
+            <i class="fas fa-plus" inert></i>
+        </a>
```
Fyorl, line 7: Should probably be a `button.control-button`. As far as I understand WCAG guidelines anyway, anchor elements that behave like buttons (i.e. don't actually link to anything and instead perform some action on click) should just be buttons.

templates/advancement/trait-config-choices.hbs
```diff
@@ -0,0 +1,31 @@
...
+            <label {{~#if selected}} class="selected"{{/if}}>
+                <input type="radio" name="selectedIndex" value="{{ @index }}" {{ checked selected }}>
+                {{#if selected}}
+                <i class="fa-solid fa-arrow-right-long"></i>
+                {{else}}
+                <i class="fa-solid fa-gear"></i>
+                {{/if}}
+                <span>{{ label }}</span>
+            </label>
+            <a data-action="removeChoice" data-tooltip="DND5E.ADVANCEMENT.Trait.Action.RemoveChoice"
```
Fyorl, line 24: Same comment as above.

module/applications/advancement/trait-config.mjs
```diff
@@ -103,69 +131,84 @@ export default class TraitConfig extends AdvancementConfig {
...
+    };
 
     // Disable selecting categories in mastery mode
     if ( this.advancement.configuration.mode === "mastery" ) {
-      context.choiceOptions.forEach((key, value) => value.disabled = !!value.children);
+      context.trait.options.forEach((key, value) => value.disabled = !!value.children);
     }
 
+    context.mode = {
+      hint: CONFIG.DND5E.traitModes[this.advancement.configuration.mode].hint,
```
Fyorl, line 153: Throws an error and prevents the sheet from opening again if you set the mode to blank and close the sheet.

Fyorl (changes requested):

module/applications/advancement/trait-config.mjs
arbron: Blank isn't a valid mode, how are you setting that?

Screenshot 2025-01-17 at 09 55 31

Fyorl (approved):

PR #4784, opened 2024-11-26
[#4718] Automatically consume legendary actions

### LINKED ISSUE #4718: Activities with "Legendary Action" activation type should consume automatically
Currently in order to properly set up legendary actions requires setting the activation type to Legendary Action, and adding attribute consumption to consume the legendary actions attribute on NPCs. It would be simpler if setting the activation type caused the system to handle that consumption on its own, without the extra step.

As part of the migration, the system should check whether legendary actions are already set in consumption to ensure that they are only consumed once.

arbron:
Updates activities to consume legendary actions automatically if their activation type is set to Legendary Action, they are being used on an NPC actor, and they don't have existing legendary action consumption configured using the old resource setup.

Closes #4718

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -730,15 +730,15 @@ export default Base => class extends PseudoDocumentMixin(Base) {
         let message;
         if ( legendary.value === 0 ) message = "DND5E.ACTIVATION.Warning.NoActions";
         else if ( count > legendary.value ) message = "DND5E.ACTIVATION.Warning.NotEnoughActions";
-        if ( !message ) {
-          updates.actor["system.resources.legact.value"] = legendary.value - count;
-        } else {
+        if ( message ) {
           const err = new ConsumptionError(game.i18n.format(message, {
             type: game.i18n.localize("DND5E.LegAct"),
             required: formatNumber(count),
             available: formatNumber(legendary.value)
           }));
           errors.push(err);
+        } else {
+          updates.actor["system.resources.legact.value"] = legendary.value - count;
         }
       }
     }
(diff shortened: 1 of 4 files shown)


krbz999:
Is this consumption without having to specify actual consumption in the activity, instead derived entirely from the activation type?

module/documents/activity/mixin.mjs
```diff
@@ -715,6 +720,29 @@ export default Base => class extends PseudoDocumentMixin(Base) {
...
+    // Handle action economy
+    if ( ((config.consume === true) || config.consume.action) && (this.activation.type === "legendary") ) {
+      const containsLegendaryConsumption = this.consumption.targets
+        .find(t => (t.type === "attribute") && (t.target === "resources.legact.value"));
+      const count = this.activation.value ?? 1;
+      const legendary = this.actor.system.resources?.legact;
+      if ( legendary && !containsLegendaryConsumption ) {
+        let message;
+        if ( legendary.value === 0 ) message = "DND5E.ACTIVATION.Warning.NoActions";
+        else if ( count > legendary.value ) message = "DND5E.ACTIVATION.Warning.NotEnoughActions";
```
Fyorl, line 732: Not sure there's much value in distinguishing between these two cases.

module/documents/activity/mixin.mjs
```diff
@@ -715,6 +720,29 @@ export default Base => class extends PseudoDocumentMixin(Base) {
...
+        if ( !message ) {
+          updates.actor["system.resources.legact.value"] = legendary.value - count;
+        } else {
+          const err = new ConsumptionError(game.i18n.format(message, {
+            type: game.i18n.localize("DND5E.LegAct"),
+            required: formatNumber(count),
+            available: formatNumber(legendary.value)
+          }));
+          errors.push(err);
+        }
```
Fyorl, line 742: Prefer if these branches are swapped so that the `else` branch doesn't read like a double-negative.

Fyorl (approved):

module/documents/activity/mixin.mjs
arbron: Just used the same warning pattern as the other consumption bits, which distinguish. More useful for 2014 content which have things that spend more than 1 legendary action.

Issue #3214, opened 2024-03-08
Expand Limited Uses field options


revilowaldow:
Expand the limited uses field on features, items etc. to support values of "turn" and "round" that automatically reset the limited uses as the combat tracker is advanced.

krbz999:
Related: https://github.com/foundryvtt/dnd5e/issues/4738

PR #4785, opened 2024-11-26
[#2359] Automatically reset legendary actions during combat

### LINKED ISSUE #2359: Feature Request - Legendary Actions should reset at the start of a monster's turn
Any monster that has Legendary Actions, gets them back at the start of their turn.  It would be helpful if this was automated, so that as soon as a monster's turn starts, the number of available Legendary Actions is reset back to its maximum.

arbron:
Automatically resets NPCs legendary actions count at the start of a new combat and at the end of the actor's turn. The reset is at the end of the turn to avoid issues with using legendary actions between the end of the previous combatant's turn and this one.

This works using a new `_recoverUses` method on `Combat5e` that is called by `startCombat` and `nextTurn`. It is passed a series of recovery periods with either `true` (indicating that this should be recovered for all combatants) or a specific combatant (indicating that this should only be recovered for that specific combatant). At the start of combat the `encounter` recovery is called for all combatants, and on `nextTurn` the `turnEnd` is called for the previous combatant and `turnStart` is called for the incoming combatant.

This then calls `recoverCombatUses` on the individual combatants, so long as they have at least one relevant period. This method handles calling `recoverCombatUses` on the actor's system data and sending out the necessary hooks. If additional combat recovery types requested in #3214 are implemented this is where they will also be handled.

Closes #2359

### DIFF
--- a/module/documents/combatant.mjs
+++ b/module/documents/combatant.mjs
@@ -11,13 +11,69 @@ export default class Combatant5e extends Combatant {
   /* -------------------------------------------- */
 
   /**
+   * Reset combat-related uses.
+   * @param {Set<string>} periods  Which recovery periods should be considered.
+   */
+  async recoverCombatUses(periods) {
+    /**
+     * A hook event that fires before combat-related recovery changes.
+     * @function dnd5e.preCombatRecovery
+     * @memberof hookEvents
+     * @param {Combatant5e} combatant     Combatant that is being recovered.
+     * @param {object} data
+     * @param {Set<string>} data.periods  Periods to be recovered.
+     * @returns {boolean}                 Explicitly return `false` to prevent recovery from being performed.
+     */
+    if ( Hooks.call("dnd5e.preCombatRecovery", this, { periods }) === false ) return;
+
+    const updates = { actor: {}, item: [] };
+    await this.actor?.system.recoverCombatUses?.(periods, updates);
+
+    // TODO: If https://github.com/foundryvtt/dnd5e/issues/3214 is implemented, this is where
+    // item/activity uses with combat related recovery can be recovered
+
+    /**
+     * A hook event that fires after combat-related recovery changes have been prepared, but before they have been
+     * applied to the actor.
+     * @function dnd5e.combatRecovery
+     * @memberof hookEvents
+     * @param {Combatant5e} combatant              Combatant that is being recovered.
+     * @param {object} data
+     * @param {Set<string>} data.periods           Periods that were recovered.
+     * @param {{ actor: object, item: object[] }}  Update that will be applied to the actor and their items.
+     * @returns {boolean}                          Explicitly return `false` to prevent updates from being performed.
+     */
+    if ( Hooks.call("dnd5e.combatRecovery", this, { periods, updates }) === false ) return;
+
+    if ( !foundry.utils.isEmpty(updates.actor) ) await this.actor.update(updates.actor);
+    if ( updates.item.length ) await this.actor.updateEmbeddedDocuments("Item", updates.item);
+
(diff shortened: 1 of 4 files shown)


module/documents/combatant.mjs
```diff
@@ -10,14 +10,70 @@ export default class Combatant5e extends Combatant {
...
+    /**
+     * A hook event that fires before combat-related recovery changes.
+     * @function dnd5e.preCombatRecovery
+     * @memberof hookEvents
+     * @param {Combatant5e} combatant     Combatant that is being recovered.
+     * @param {object} data
+     * @param {Set<string>} data.periods  Periods to be recovered.
+     * @returns {boolean}                 Explicitly return `false` to prevent recovery from being performed.
+     */
+    if ( Hooks.call("dnd5e.preCombatRecovery", this, { periods }) === false ) return;
```
Fyorl, line 27: Not sure if this is maybe a consistent hook style we're trying to align on, but if not then I'm not sure of the benefit of an extra layer of nesting here.

```suggestion
    if ( Hooks.call("dnd5e.preCombatRecovery", this, periods) === false ) return;
```

module/documents/combatant.mjs
```diff
@@ -10,14 +10,70 @@ export default class Combatant5e extends Combatant {
...
+
+    /**
+     * A hook event that fires after combat-related recovery changes have been prepared, but before they have been
+     * applied to the actor.
+     * @function dnd5e.combatRecovery
+     * @memberof hookEvents
+     * @param {Combatant5e} combatant              Combatant that is being recovered.
+     * @param {object} data
+     * @param {Set<string>} data.periods           Periods that were recovered.
+     * @param {{ actor: object, item: object[] }}  Update that will be applied to the actor and their items.
```
Fyorl, line 43: Missing param name

module/documents/combatant.mjs
```diff
@@ -10,14 +10,70 @@ export default class Combatant5e extends Combatant {
...
+     * @returns {boolean}                          Explicitly return `false` to prevent updates from being performed.
+     */
+    if ( Hooks.call("dnd5e.combatRecovery", this, { periods, updates }) === false ) return;
```
Fyorl, line 46: Same as above comment on extra nesting.

```suggestion
    if ( Hooks.call("dnd5e.combatRecovery", this, periods, updates) === false ) return;
```

module/documents/combatant.mjs
```diff
@@ -10,14 +10,70 @@ export default class Combatant5e extends Combatant {
...
+
+    /**
+     * A hook event that fires after combat-related recovery changes have been applied.
+     * @function dnd5e.postCombatRecovery
+     * @memberof hookEvents
+     * @param {Combatant5e} combatant     Combatant that is being recovered.
+     * @param {object} data
+     * @param {Set<string>} data.periods  Periods that were recovered.
+     */
+    Hooks.callAll("dnd5e.postCombatRecovery", this, { periods });
```
Fyorl, line 62:
```suggestion
    Hooks.callAll("dnd5e.postCombatRecovery", this, periods);
```

Fyorl (approved):

module/documents/combatant.mjs
krbz999: If more data ends up being passed through in the future, having an object to add the data to would make it a nonbreaking change. (Probably not in this case since it can just be appended to the tail of the arguments list but still.)

Fyorl: Following that logic ad absurdum, all hook and function signatures should just take a single object argument. The combatant in this hook is not part of the data object, presumably because it's considered required information. It seems like the periods information is similarly important, and not really optional, but maybe not.

PR #4793, opened 2024-11-27
[#3355] Add attack enricher

### LINKED ISSUE #3355: Add attack enricher
In addition to the current damage enricher, it would be good to have an attack enricher in the style of `[[/attack +5]]` which would compose a generic attack roll, taking advantage of the roll dialog and the new targeting features without having to be tied to a specific item.

This should also be usable in the form `[[/attack]]`, which would attempt to automatically determine the attack modifier from the item it is embedded within, and roll using that item's attack rather than a generic attack. When paired with the change suggested in #2568, this would allow for writing generic NPC weapon descriptions that adapt to the NPC's stats:
`Melee Weapon Attack: [[/attack]], reach 5 ft., one target. Hit. [[/damage average]] damage.`
would be turned into:
`Melee Weapon Attack: [+3] to hit, reach 5 ft., one target. Hit 3 ([1d4 + 1]) slashing damage.`

arbron:
Adds a new attack enricher that can be used to roll an attack using a fixed formula:

```
[[/attack +5]]
[[/attack formula=5]]
```

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -1,5 +1,6 @@
 import { formatNumber, getSceneTargets, getTargetDescriptors, simplifyBonus } from "./utils.mjs";
 import Award from "./applications/award.mjs";
+import AttackRollConfigurationDialog from "./applications/dice/attack-configuration-dialog.mjs";
 import * as Trait from "./documents/actor/trait.mjs";
 import { rollItem } from "./documents/macro.mjs";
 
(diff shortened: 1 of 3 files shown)


module/enrichers.mjs
```diff
@@ -78,6 +82,43 @@ function parseConfig(match) {
...
+  for ( const value of config.values ) formulaParts.push(value);
+  config.formula = Roll.defaultImplementation.replaceFormulaData(formulaParts.join(" "), options.rollData ?? {});
+  if ( !config.formula.startsWith("+") && !config.formula.startsWith("-") ) config.formula = `+${config.formula}`;
+  config.type = "attack";
+
+  if ( label ) return createRollLink(label, config);
+
+  const span = document.createElement("span");
+  span.innerHTML = game.i18n.format(`EDITOR.DND5E.Inline.Attack${config.format === "long" ? "Long" : "Short"}`, {
+    formula: createRollLink(config.formula, config).outerHTML
```
Fyorl, line 117: We should run the label through `simplifyBonus`. Without it, a simple formula like `[[/attack @prof + @abilities.str.mod]]` displays as "+2 + 2 to hit".

```suggestion
    formula: createRollLink(simplifyBonus(config.formula), config).outerHTML
```

Fyorl (approved):
Seems to be missing this functionality from #3355:
> This should also be usable in the form `[[/attack]]`, which would attempt to automatically determine the attack modifier from the item it is embedded within, and roll using that item's attack rather than a generic attack.

So I don't think we can consider this as closing #3355 but it's a good first step.

module/enrichers.mjs
arbron: Probably `simplifyRollFormula`, cause we don't want to squash dice values, but yeah.

PR #4796, opened 2024-11-27
[#3357] Support variable types & multiple damage parts in enricher

### LINKED ISSUE #3357: Add support for multiple damage types in single damage enricher
The current damage enricher supports rolling a single set of damage with a single type. There are a number of times where content includes rolling two different types as part of a single damaging operation, and with current changes to damage application and concentration having those as a single overarching damage roll is important. There are two significant design issues that need to be resolved:

### Enricher Design
First question is how do we represent separate damage parts in a single damage enricher. Right now the enrichers are written in such a that the order of entries isn't important. You can write `[[/damage 1d6 + 5 piercing]]` or `[[/damage piercing 1d6 + 5]]` or even `[[/damage 1d6 piercing + 5]]` if you are feeling really wild, but they will all result in the final formula of `1d6 + 5` with the type `piercing `.

One option is to add a separating symbol in enrichers that splits their configs into separate parts. Something like `[[/damage 1d6 + 5 piercing & 2d6 fire average]]` would allow the parser to split the enricher into two different configs, the first containing `{ values: ["1d6", "+", "5", "piercing] }` and the second `{ values: ["2d6", "fire", "average"] }`.

Another option would be to require a more specific format with duplicate keys. Write an enricher like `[[/damage formula="1d6 + 5" type="piercing" formula="2d6" type="fire" average]]` would result in the final configuration data: `{ formula: ["1d6 + 5", "2d6"], type: ["piercing", "fire"] values: ["average"] }`.

Either of these formats would produce results that are parsable by the enricher system.

### Output Design
The second question is how these enrichers will be displayed in text and what parts will be linked. Currently we include the minimum amount of text as possible by default (square brackets indicating the rolling link): `3 ([1d4 + 1]) slashing` if average is included or `[1d4 + 1] slashing` without average. We specifically don't include the word "damage" in the text, but only the parts that absolutely need to be defined by the enricher.

In the SRD, there are generally two ways multi-part damage is written. For NPC stat blocks, it generally says something like `3 (1d4 + 1) slashing damage plus 2 (1d4) fire damage` while spells generally say `4d6 fire damage and 4d6 radiant damage`. The main difference is between using `plus` versus `and` in connecting the two types. If had to choose one or the other then `and` would be easier to implement because we could use javascript's list formatter to quickly localize it and it would work regardless of how many damage parts are present. Using `plus` breaks down after more than two damage types (don't know if there are any official NPCs that do more than two damage types in a single attack so I can't say the official wording of that). One potential option is to use `plus` if the `average` option is set and there are exactly two options, otherwise fall back to `and`. There could also be an additional formatting option that specifies which to use.

We would also need to decide whether the include the word `damage` in the middle of the enricher. Since we don't include the training damage that would result in the enriched text being `4d6 fire damage and 4d6 radiant`, which is a bit odd. It might be clearer to drop the included damage and simplify it to `4d6 fire and 4d6 radiant` or `3 (1d4 + 1) slashing plus 2 (1d4) fire`, which isn't quite what is used in official text but is closer to previous decisions regarding single-type enrichers.

The final issue with the output is where to put the link. Current single-type enrichers are very simple and put the link around just the roll formula regardless of whether average is used or not. If we continued this pattern with the new enrichers it would result in two separate links `[4d6] fire and [4d6] radiant`, both of which roll both damage types. One way to make it clearer might be to have the whole enriched text be a single link, but mark it so only the formulas themselves have the link stying. So hovering over either formula with give any hover indication on both parts, but there will only be one actual link.

arbron:
Adds support for variable damage types when rolling damage through a damage enricher as well as multiple damage parts.

Variable Damage Types   Multiple Damages

To indicate variable damage types, the explicit `type` key must be used when writing the enricher and the different types must be separated by a pipe (`|`) character:

```
[[/damage 2d6 type=fire|cold]]
```

To specify multiple damage parts the enricher config is split by an amperand (`&`), with each side supporting its own formula and damage types:

```
[[/damage 1d6 fire & 1d6 cold]]
[[/damage 1d6 fire & 1d6 cold average]]
```

Any configuration options that apply to the whole enricher can be placed anywhere in the configuration.

If the average option is used and there are exactly two damage parts, then they will be written as "A plus B", otherwise they will be written "A and B". This is to match the format of NPC statblocks.

Closes #3357

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -359,17 +359,17 @@ async function enrichSave(config, label, options) {
  * ```[[/damage 2d6 type=bludgeoning]]``
  * becomes
  * ```html
- * <a class="roll-action" data-type="damage" data-formula="2d6" data-damage-type="bludgeoning">
- *   <i class="fa-solid fa-dice-d20"></i> 2d6
- * </a> bludgeoning
+ * <a class="roll-link-group" data-type="damage" data-formulas="2d6" data-damage-types="bludgeoning">
+ *   <span class="roll-link"><i class="fa-solid fa-dice-d20"></i> 2d6</span> bludgeoning
+ * </a>
  * ````
  *
  * @example Display the average:
  * ```[[/damage 2d6 type=bludgeoning average=true]]``
  * becomes
  * ```html
- * 7 (<a class="roll-action" data-type="damage" data-formula="2d6" data-damage-type="bludgeoning">
- *   <i class="fa-solid fa-dice-d20"></i> 2d6
+ * 7 (<a class="roll-link-group" data-type="damage" data-formulas="2d6" data-damage-types="bludgeoning">
+ *   <span class="roll-link"><i class="fa-solid fa-dice-d20"></i> 2d6</span>
  * </a>) bludgeoning
  * ````
  *
@@ -377,8 +377,8 @@ async function enrichSave(config, label, options) {
  * ```[[/damage 8d4dl force average=666]]``
  * becomes
  * ```html
- * 666 (<a class="roll-action" data-type="damage" data-formula="8d4dl" data-damage-type="force">
- *   <i class="fa-solid fa-dice-d20"></i> 8d4dl
+ * 666 (<a class="roll-link-group" data-type="damage" data-formulas="8d4dl" data-damage-types="force">
+ *   <span class="roll-link"><i class="fa-solid fa-dice-d20"></i> 8d4dl</span>
  * </a> force
  * ````
  *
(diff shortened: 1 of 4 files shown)


module/enrichers.mjs
```diff
@@ -749,6 +781,25 @@ function createRollLink(label, dataset) {
...
+ * @param {object} [dataset={}]                    Data that will be added to the link for the rolling method.
+ * @param {object} [options={}]
+ * @param {boolean} [options.classes="roll-link"]  Class to add to the link.
+ * @param {string} [options.tag="a"]               Tag to use for the main link.
+ * @returns {HTMLElement}
+ */
+function createRollLink(label, dataset={}, { classes="roll-link", tag="a" }={}) {
+  const link = document.createElement(tag);
+  link.className = classes;
+  link.innerHTML = `<i class="fa-solid fa-dice-d20" inert></i>${label}`;
```
Fyorl, line 798:
```suggestion
  link.insertAdjacentHTML("afterbegin", "<i class="fa-solid fa-dice-d20" inert></i>");
  link.append(label);
```

module/enrichers.mjs
```diff
@@ -870,13 +921,21 @@ async function rollAction(event) {
...
+  let { formulas, damageTypes, rollType } = target.dataset;
+  formulas = formulas.split("&");
+  damageTypes = damageTypes.split("&");
 
   const rollConfig = {
     event,
     hookNames: ["damage"],
-    rolls: [{ parts: [formula], options: { type: damageType } }]
+    rolls: formulas.map((formula, idx) => {
+      const types = damageTypes[idx]?.split("|") ?? [];
```
Fyorl, line 933:
```suggestion
      const types = damageTypes[idx]?.split(/[|/]/g) ?? [];
```

Fyorl (changes requested):

module/enrichers.mjs
arbron: I decided to normalize to `|` in the config data, so you can use either `/` or `|` but the system doesn't have to worry about it later.

module/enrichers.mjs
```diff
@@ -718,37 +775,51 @@ export function createRollLabel(config) {
...
   _addDataset(span, dataset);
-
-  // Add main link
-  const link = document.createElement("a");
-  link.dataset.action = "roll";
-  link.innerHTML = '<i class="fa-solid fa-dice-d20"></i>';
-  link.append(label);
-  span.insertAdjacentElement("afterbegin", link);
+  if (label instanceof HTMLElement) span.insertAdjacentElement("afterbegin", label);
+  else span.insertAdjacentHTML("afterbegin", label);
```
Fyorl, line 788: Need to be careful to not parse arbitrary user-entered strings as HTML. Using `append` here to append a string ensures that it is rendered as a text node, and not HTML.

```suggestion
  if ( label instanceof HTMLElement ) span.insertAdjacentElement("afterbegin", label);
  else span.append(label);
```

module/enrichers.mjs
```diff
@@ -718,37 +775,51 @@ export function createRollLabel(config) {
...
+ * @param {object} [dataset={}]                    Data that will be added to the link for the rolling method.
+ * @param {object} [options={}]
+ * @param {boolean} [options.classes="roll-link"]  Class to add to the link.
+ * @param {string} [options.tag="a"]               Tag to use for the main link.
+ * @returns {HTMLElement}
+ */
+function createRollLink(label, dataset={}, { classes="roll-link", tag="a" }={}) {
+  const link = document.createElement(tag);
+  link.className = classes;
+  link.insertAdjacentHTML("afterbegin", '<i class="fa-solid fa-dice-d20" inert></i>');
```
Fyorl, line 818: Missed adding the label:

```suggestion
  link.insertAdjacentHTML("afterbegin", '<i class="fa-solid fa-dice-d20" inert></i>');
  link.append(label);
```

Fyorl (approved):

PR #4805, opened 2024-11-29
[#4804] Add help icon and tooltip to AE Attribute Key header

### LINKED ISSUE #4804: Add help icon & tooltip to Attribute Key section of Active Effect window's Effects tab
Related: #4745 
Summarizing related: The Active Effects Guide page of the wiki is currently the only way of knowing which attribute keys are valid, and how to use them. An in-app solution would be ideal to allow new users to quickly find the keys they're looking for.
As a stopgap solution (or perhaps a permanent solution, if the related issue is determined not to make sense in the system itself), adding a little "Help" icon beside the Attribute Key header, with a tooltip linking users to the AE Guide wiki page, would provide guidance for users unfamiliar with how to configure AEs. Example mockup:
image

roth-michael:
Closes #4804 

image

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -609,7 +609,7 @@ Hooks.on("renderDocumentSheetConfig", (app, html) => {
   }
 });
 
-Hooks.on("renderActiveEffectConfig", (app, html, data) => documents.ActiveEffect5e.onRenderActiveEffectConfig(html));
+Hooks.on("renderActiveEffectConfig", (app, html, data) => app.document.onRenderActiveEffectConfig(html));
 
 Hooks.on("targetToken", canvas.Token5e.onTargetToken);
 
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -609,6 +609,8 @@ Hooks.on("renderDocumentSheetConfig", (app, html) => {
   }
 });
 
+Hooks.on("renderActiveEffectConfig", (app, html, data) => documents.ActiveEffect5e.onRenderActiveEffectConfig(html));
+
 Hooks.on("targetToken", canvas.Token5e.onTargetToken);
 
 Hooks.on("renderCombatTracker", (app, html, data) => app.renderGroups(html));
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -609,8 +609,6 @@ Hooks.on("renderDocumentSheetConfig", (app, html) => {
   }
 });
 
-Hooks.on("renderActiveEffectConfig", (app, html, data) => app.document.onRenderActiveEffectConfig(html));
-
 Hooks.on("targetToken", canvas.Token5e.onTargetToken);
 
 Hooks.on("renderCombatTracker", (app, html, data) => app.renderGroups(html));
--- a/lang/en.json
+++ b/lang/en.json
@@ -303,6 +303,7 @@
   }
 },
 
+"DND5E.ActiveEffectAttributeKeyTooltip": "For a list of common keys, see <a href=https://github.com/foundryvtt/dnd5e/wiki/Active-Effect-Guide>the wiki</a>.",
 "DND5E.ActiveEffectOverrideWarning": "This value is being modified by an Active Effect and cannot be edited. Disable the effect to edit it.",
 "DND5E.Add": "Add",
 "DND5E.AdditionalControls": "Additional Controls",
--- a/lang/en.json
+++ b/lang/en.json
@@ -255,6 +255,10 @@
   }
 },
 
+"DND5E.ACTIVEEFFECT": {
+  "AttributeKeyTooltip": "For a list of common keys, see <a href=\"{url}\">the wiki</a>."
+},
+
 "DND5E.ACTIVITY": {
   "Action": {
     "Create": "Create Activity",
@@ -303,7 +307,6 @@
   }
 },
 
-"DND5E.ActiveEffectAttributeKeyTooltip": "For a list of common keys, see <a href=https://github.com/foundryvtt/dnd5e/wiki/Active-Effect-Guide>the wiki</a>.",
 "DND5E.ActiveEffectOverrideWarning": "This value is being modified by an Active Effect and cannot be edited. Disable the effect to edit it.",
 "DND5E.Add": "Add",
 "DND5E.AdditionalControls": "Additional Controls",
--- a/lang/en.json
+++ b/lang/en.json
@@ -247,16 +247,13 @@
 },
 
 "DND5E.ACTIVEEFFECT": {
+  "AttributeKeyTooltip": "For a list of common keys, see <a href=\"{url}\">the wiki</a>.",
   "FIELDS": {
     "statuses": {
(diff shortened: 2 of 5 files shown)


roth-michael:
V13 screenshot:
image

Don't love where I put the v2 style rule, but couldn't find a more appropriate place.

module/documents/active-effect.mjs
```diff
@@ -615,6 +615,24 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+   */
+  static onRenderActiveEffectConfig(html) {
+    html = html instanceof HTMLElement ? html : html[0];
+    const helpIconElement = document.createElement("i");
+    helpIconElement.classList.add("fa-solid", "fa-circle-question");
+    const tooltipText = game.i18n.localize("DND5E.ActiveEffectAttributeKeyTooltip");
+    helpIconElement.setAttribute("data-tooltip", tooltipText);
+    helpIconElement.setAttribute("data-tooltip-direction", "RIGHT");
+    helpIconElement.setAttribute("data-locked", true);
+    const targetElement = html.querySelector("section[data-tab='effects'],section[data-tab='changes'] .key");
```
arbron, line 630: This sticks it in the wrong spot in V12, you can add `.key` back to the first part or use `:is` to grab either:

```suggestion
    const targetElement = html.querySelector("section:is([data-tab='effects'],[data-tab='changes']) .key");
```

or

```suggestion
    const targetElement = html.querySelector("section[data-tab='effects'] .key,section[data-tab='changes'] .key");
```

less/v1/apps.less
```diff
@@ -532,6 +532,10 @@ h5 {
   }
 }
 
+.active-effect-sheet header.effects-header i{
+  margin-left: 0.25rem
+}
```
arbron, line 537: A help cursor would be nice here to suggest that hovering will do something.

```suggestion
.active-effect-sheet header.effects-header i {
  margin-left: 0.25rem;
  cursor: help;
}
```

less/v2/apps.less
```diff
@@ -1560,6 +1560,11 @@ dialog.dnd5e2.application {
   }
 }
 
+
+.active-effect-config section.changes header div.key i{
+  margin-left: 0.25rem
+}
```
arbron, line 1566:
```suggestion
.active-effect-config section.changes header div.key i {
  margin-left: 0.25rem;
  cursor: help;
}
```

lang/en.json
```diff
@@ -203,6 +203,7 @@
 
 "DND5E.ActorWarningInvalidItem": "{itemType} items cannot be added to a {actorType}.",
 "DND5E.ActorWarningSingleton": "Only a single {itemType} can be added to a {actorType}.",
+"DND5E.ActiveEffectAttributeKeyTooltip": "For a list of common keys, see <a href=https://github.com/foundryvtt/dnd5e/wiki/Active-Effect-Guide>the wiki</a>.",
```
arbron, line 206: Let's stick this in a nested object. Plus we should be able to handle alternate URLs for enchantments because they have a different set of keys.

```suggestion
"DND5E.ACTIVEEFFECT": {
  "AttributeKeyTooltip": "For a list of common keys, see <a href=\"{url}\">the wiki</a>."
},
```

The ordering here also isn't quite alphabetical, so this should move above the "DND5E.ACTIVITY" line (and leave an extra space surrounding it).

dnd5e.mjs
```diff
@@ -578,6 +578,8 @@ Hooks.on("getItemDirectoryEntryContext", documents.Item5e.addDirectoryContextOpt
 
 Hooks.on("renderJournalPageSheet", applications.journal.JournalSheet5e.onRenderJournalPageSheet);
 
+Hooks.on("renderActiveEffectConfig", (app, html, data) => documents.ActiveEffect5e.onRenderActiveEffectConfig(html));
```
arbron, line 581: To detect whether this is for an enchantment or not, let's make this an instance method:

```suggestion
Hooks.on("renderActiveEffectConfig", (app, html, data) => app.document.onRenderActiveEffectConfig(html));
```

module/documents/active-effect.mjs
```diff
@@ -615,6 +615,24 @@ export default class ActiveEffect5e extends ActiveEffect {
 
   /* -------------------------------------------- */
 
+  /**
+   * Add help icon & tooltip to Attribute Key header
+   * @param {jQuery|HTMLElement} html  The Active Effect Config HTML.
+   */
+  static onRenderActiveEffectConfig(html) {
```
arbron, line 622: Non-static so we can grab info on the active effect being displayed. This probably also shouldn't exist under the "Exhaustion and Concentration Handling" heading, so move it down under "Helpers" just beneath `addOverriddenChoices`.

```suggestion
 onRenderActiveEffectConfig(html) {
```

module/documents/active-effect.mjs
```diff
@@ -615,6 +615,24 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+    html = html instanceof HTMLElement ? html : html[0];
+    const helpIconElement = document.createElement("i");
+    helpIconElement.classList.add("fa-solid", "fa-circle-question");
+    const tooltipText = game.i18n.localize("DND5E.ActiveEffectAttributeKeyTooltip");
```
arbron, line 626:
```suggestion
    const tooltipText = game.i18n.format("DND5E.ACTIVEEFFECT.AttributeKeyTooltip", {
      url: this.type === "enchantment"
        ? "https://github.com/foundryvtt/dnd5e/wiki/Enchantment"
        : "https://github.com/foundryvtt/dnd5e/wiki/Active-Effect-Guide"
    });
```

arbron (changes requested):

module/documents/active-effect.mjs
```diff
@@ -808,6 +790,28 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+    helpIconElement.classList.add("fa-solid", "fa-circle-question");
+    const tooltipText = game.i18n.format("DND5E.ACTIVEEFFECT.AttributeKeyTooltip", {
+      url: this.type === "enchantment"
+        ? "https://github.com/foundryvtt/dnd5e/wiki/Enchantment"
+        : "https://github.com/foundryvtt/dnd5e/wiki/Active-Effect-Guide"
+    });
+    helpIconElement.setAttribute("data-tooltip", tooltipText);
+    helpIconElement.setAttribute("data-tooltip-direction", "RIGHT");
+    helpIconElement.setAttribute("data-locked", true);
+    const targetElement = html.querySelector("section:is([data-tab='effects'],section[data-tab='changes']) .key");
```
arbron, line 809:
```suggestion
    const targetElement = html.querySelector("section:is([data-tab='effects'],[data-tab='changes']) .key");
```

arbron (changes requested):

module/documents/active-effect.mjs
roth-michael: Well, good thing at least one of us is paying attention.

Fyorl (approved):

PR #4814, opened 2024-12-02
[#4751] Split `BasicRoll#build` into three parts

### LINKED ISSUE #4751: item/activity rollRecharge ignores message.create set to false in dnd5e.preRollRechargeV2 but honors it when set via dnd5e.preRollRecharge
Due to the roll refactor rollRecharge ignores changing message.create to false preRollRechargeV2.
In the code createMessage (which determines display the chat card or not) is recorded before calling BasicRoll.build, but after the preRollRecharge hook fires, so that changes made in preRollRechargeV2 are ignored, but changes made in preRollRecharge are honored.

I could not see any simple way to resolve the issue but wanted to understand if this is a wont fix and will work around it.,

```js
    if ( "dnd5e.preRollRecharge" in Hooks.events ) {
        "The `dnd5e.preRollRecharge` hook has been deprecated and replaced with `dnd5e.preRollRechargeV2`.",
        { since: "DnD5e 4.0", until: "DnD5e 4.4" }
      );
      const hookData = {
        formula: rollConfig.rolls[0].parts[0], data: rollConfig.rolls[0].data,
        target: rollConfig.rolls[0].options.target, chatMessage: messageConfig.create
      };
      if ( Hooks.call("dnd5e.preRollRecharge", this, hookData) === false ) return;
      rollConfig.rolls[0].parts[0] = hookData.formula;
      rollConfig.rolls[0].data = hookData.data;
      rollConfig.rolls[0].options.target = hookData.target;
      messageConfig.create = hookData.chatMessage;
    }

    const createMessage = messageConfig.create !== false;
    messageConfig.create = false;
    const rolls = await CONFIG.Dice.BasicRoll.build(rollConfig, dialogConfig, messageConfig);
    ```js

### PR DESCRIPTION
**Pull request by arbron**
Splits the `build` method on `BasicRoll` into three parts:
1. `buildConfigure`: Applies keybindings, calls the `preRoll` hook, displays the configuration dialog, and calls the `postRollConfiguration` hook.
2. `buildEvaluate`: Evalutes the rolls if `evaluate` is `true`.
3. `buildPost`: Sets the `originatingMessage` flag and posts a message to chat.

The `build` method now calls these three methods in turn.

This process now also adjust `config.hookNames` to incude the empty value rather than using a local variable so the final hook names can be used elsewhere, and saves the created chat message into the message configuration so it can be referenced after the fact.

Closes #4751

### DIFF
--- a/module/dice/basic-roll.mjs
+++ b/module/dice/basic-roll.mjs
@@ -187,7 +187,7 @@ export default class BasicRoll extends Roll {
    * @param {BasicRollMessageConfiguration} [message={}]  Configuration for message creation.
    */
   static async buildEvaluate(rolls, config={}, message={}) {
-    if ( (config.evaluate !== false) ) {
+    if ( config.evaluate !== false ) {
       for ( const roll of rolls ) await roll.evaluate();
     }
   }
@@ -207,10 +207,10 @@ export default class BasicRoll extends Roll {
     if ( messageId ) foundry.utils.setProperty(message.data, "flags.dnd5e.originatingMessage", messageId);
 
     if ( rolls?.length && (config.evaluate !== false) && (message.create !== false) ) {
-      message.created = await this.toMessage(rolls, message.data, { rollMode: message.rollMode });
+      message.document = await this.toMessage(rolls, message.data, { rollMode: message.rollMode });
     }
 
-    return message.created;
+    return message.document;
   }
 
   /* -------------------------------------------- */
--- a/module/dice/basic-roll.mjs
+++ b/module/dice/basic-roll.mjs
@@ -49,9 +49,10 @@ const { DiceTerm, NumericTerm } = foundry.dice.terms;
  * Configuration data for creating a roll message.
  *
  * @typedef {object} BasicRollMessageConfiguration
- * @property {boolean} [create=true]  Create a message when the rolling is complete.
- * @property {string} [rollMode]      The roll mode to apply to this message from `CONFIG.Dice.rollModes`.
- * @property {object} [data={}]       Additional data used when creating the message.
+ * @property {boolean} [create=true]     Create a message when the rolling is complete.
+ * @property {ChatMessage5e} [document]  Final created chat message document once process is completed.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: module/dice/basic-roll.mjs [hunk 1]
```diff
@@ -153,23 +170,47 @@ export default class BasicRoll extends Roll {
...
+   */
+  static async buildPost(rolls, config, message) {
     message.data = foundry.utils.expandObject(message.data ?? {});
     const messageId = config.event?.target.closest("[data-message-id]")?.dataset.messageId;
     if ( messageId ) foundry.utils.setProperty(message.data, "flags.dnd5e.originatingMessage", messageId);
 
-    if ( rolls?.length && (message.create !== false) ) {
-      await this.toMessage(rolls, message.data, { rollMode: message.rollMode });
+    if ( rolls?.length && (config.evaluate !== false) && (message.create !== false) ) {
+      message.created = await this.toMessage(rolls, message.data, { rollMode: message.rollMode });
```
Fyorl, line 210: Elsewhere and in the docs this is `document`. Not sure if this should be `message.document =` then, or if the other places should change to use `created` instead.

module/documents/activity/attack.mjs
```diff
@@ -179,16 +179,15 @@ export default class AttackActivity extends ActivityMixin(AttackActivityData) {
       _applyDeprecatedD20Configs(rollConfig, dialogConfig, messageConfig, oldConfig);
     }
 
-    const createMessage = messageConfig.create !== false;
-    messageConfig.create = false;
     const rolls = await CONFIG.Dice.D20Roll.build(rollConfig, dialogConfig, messageConfig);
     if ( !rolls.length ) return null;
-    if ( createMessage ) {
+    if ( messageConfig.document ) {
```
Fyorl, line 184: We could use the same technique as `rollRecharge` and perform `buildConfigure` and `buildEvaluate` separately, then add the flags before calling `buildPost`. That saves a round-trip to the server and a re-render.

Fyorl (approved):
This worked out very neatly, which is encouraging.

PR #4815, opened 2024-12-02
[#4800] Add visibility to check & save activities

### LINKED ISSUE #4800: Add visibility options to check & save activities
Right now all checks and saves from these activities are visible to all users, but there are some features that call for checks and saves from the creature activating them, rather than requesting from others. A checkbox for specific activities that controls this button visibility would be welcome, similar to the one that exists already on the Utility activity.

arbron:
Adds a "Visible to All" option to `CheckActivity` & `SaveActivity` to match the option currently on `UtilityActivity`. This allows checks and save that are only visible to the person activing the item, rather than to all players.

This also modifies how the behavior fields on the identity tab of activity sheets are rendered to avoid having to replace the whole template just to add a single field.

Closes #4800

### DIFF
--- a/module/data/activity/_types.mjs
+++ b/module/data/activity/_types.mjs
@@ -83,6 +83,7 @@
  * @property {object} check.dc
  * @property {string} check.dc.calculation   Method or ability used to calculate the difficulty class of the check.
  * @property {string} check.dc.formula       Custom DC formula or flat value.
+ * @property {boolean} check.visible         Should this check be displayed to all players?
  */
 
 /**
@@ -143,6 +144,7 @@
  * @property {object} save.dc
  * @property {string} save.dc.calculation           Method or ability used to calculate the difficulty class.
  * @property {string} save.dc.formula               Custom DC formula or flat value.
+ * @property {boolean} save.visible                 Should this check be displayed to all players?
  */
 
 /**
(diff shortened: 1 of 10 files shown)


module/data/activity/_types.mjs
```diff
@@ -143,6 +144,7 @@
  * @property {object} save.dc
  * @property {string} save.dc.calculation           Method or ability used to calculate the difficulty class.
  * @property {string} save.dc.formula               Custom DC formula or flat value.
+ * @property {boolean} save.visible                 Should this check be displayed to all players?
```
```suggestion
 * @property {boolean} save.visible                 Should this save be displayed to all players?
```

Fyorl (changes requested):
I guess we added the 'visibility' section some time after this PR was originally made, but should the 'visible to all' checkbox be under 'visibility' rather than 'behaviour'? I guess with the Utility activity it's still under 'behaviour'.

arbron:
> I guess we added the 'visibility' section some time after this PR was originally made, but should the 'visible to all' checkbox be under 'visibility' rather than 'behaviour'? I guess with the Utility activity it's still under 'behaviour'.

So far I've only used the "Visibility" section to represent visibility of the activity as a whole, rather than rolls or anything else produced by it. We've used behaviors to handle everything that happens post-usage.

Fyorl (approved):

PR #4817, opened 2024-12-02
[#4798] Add option to conceal item descriptions from players

### LINKED ISSUE #4798: Add setting to hide item descriptions from players in chat when activated by GM
Currently there are two ways to item description from players in chat:
1. Write something else in the chat description (non-spoilery description or just an empty space)
2. Use secret blocks in the normal description to hide the portions that shouldn't be visible to players

Both of these options involve modifying each individual item by either GMs or content creators, and once done there is not easy way for the GM to reveal the chat description without going back and re-editing the items. It would be useful to have a setting that hid the item descriptions from players by default if they were activated by GMs.

If an item already has a custom chat description, then this setting would be ignored.

This could possibly be accompanied by an option in the context menu to reveal the description to players so GMs could show on a one-by-one basis.

arbron:
Adds a new setting, "Conceal Item Descriptions", which hides item description in chat messages for players so long as they are (1) posted by the GM and (2) don't have a chat description.

To avoid overloading the settings dialog any more, this adds a new setting configuration app that now contains all of the player visibility-releated settings, such as attack result visibility, challenge visibility, and bloodied effect visibility. Some reshuffling of the existing settings applications was done so they are all now organized into the `applications/settings` folder.

Visibility Settings

A new `BaseSettingsConfig` application was introduced to make it easy to add new settings applications that follow this pattern, and the bastions config application was refactored to use this new base application.

Closes #4798


lang/en.json
```diff
@@ -3830,6 +3816,29 @@
     "Hint": "Change handling of various rules between the 2024 and 2014 rule sets.",
     "Legacy": "Legacy Rules (2014)",
     "Modern": "Modern Rules (2024)"
+  },
+  "VISIBILITY": {
+    "Hint": "Various configuration options that affect player visibility for certain information that the DM may wish to keep secret.",
```
Fyorl, line 3821:
```suggestion
    "Hint": "Various configuration options that affect player visibility of certain information that the DM may wish to keep secret.",
```

Fyorl (approved):
Nice

Issue #4380, opened 2024-09-20
Add config object for attack modes


krbz999:
```js
const attackModes = {
  offhand: "DND5E.ATTACK.Mode.Offhand",
  oneHanded: "DND5E.ATTACK.Mode.OneHanded",
  thrown: "DND5E.ATTACK.Mode.Thrown",
  "thrown-offhand": "DND5E.ATTACK.Mode.ThrownOffhand",
  twoHanded: "DND5E.ATTACK.Mode.TwoHanded"
};
```

While creating a script to add a bonus in a damage roll hook, there was nowhere to find these values for comparison, so ended up creating it myself. Would be good if this was in `CONFIG.DND5E` or elsewhere publicly available.

PR #4823, opened 2024-12-04
[#2568, #3355] Add activity linking for attack & damage enrichers

### LINKED ISSUE #2568: Add 'blank' damage enricher
Basically extending the damage enricher so you can leave it blank, such as just `[[/damage]]`, which would then retrieve from the item's damage formulas rather than any written input.

Would also be helpful if the item itself is passed to the roll hook.

### LINKED ISSUE #3355: Add attack enricher

arbron:
Adds the ability for attack and damgae enrichers to be linked to a specific activity, allowing them to fetch their formulas and any other information directly from that activity and to roll that activity when clicked, rather than using the generic rolling.

This can be done in two ways, either by explicitly stating the activity ID: `[[/attack activity=RLQlsLo5InKHZadn]]`, or by not specifiying a formula and letting the system find the first matching activity: `[[/attack]]` or `[[/damage average]]`.

This also adds support for specifying the attack mode in the enricher. For attack enrichers this just controls what attack mode is selected in the dialog by default (or used if the dialog is bypassed), and for damage enrichers this adjusts the damage formula if it changes based on the attack mode (so versatile weapons can contain enrichers for one-handed and two-handed damage.

Closes #2568
Closes #3355

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2720,6 +2720,31 @@ preLocalize("attackClassifications", { key: "label" });
 /* -------------------------------------------- */
 
 /**
+ * Attack modes available for weapons.
+ * @enum {string}
+ */
+DND5E.attackModes = Object.seal({
+  oneHanded: {
+    label: "DND5E.ATTACK.Mode.OneHanded"
+  },
+  twoHanded: {
+    label: "DND5E.ATTACK.Mode.TwoHanded"
+  },
+  offhand: {
+    label: "DND5E.ATTACK.Mode.Offhand"
+  },
+  thrown: {
+    label: "DND5E.ATTACK.Mode.Thrown"
+  },
+  "thrown-offhand": {
+    label: "DND5E.ATTACK.Mode.ThrownOffhand"
+  }
+});
+preLocalize("attackModes", { key: "label" });
+
+/* -------------------------------------------- */
+
+/**
  * Types of attacks based on range.
  * @enum {{ label: string }}
  */
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -15,7 +15,7 @@ export function registerCustomEnrichers() {
     "attack", "award", "check", "concentration", "damage", "healing", "item", "save", "skill", "tool"
   ];
   CONFIG.TextEditor.enrichers.push({
-    pattern: new RegExp(`\\[\\[/(?<type>${stringNames.join("|")}) (?<config>[^\\]]+)]](?:{(?<label>[^}]+)})?`, "gi"),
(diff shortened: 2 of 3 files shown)


krbz999:
Closes https://github.com/foundryvtt/dnd5e/issues/4380

module/enrichers.mjs
```diff
@@ -106,12 +106,56 @@ function parseConfig(match, { multiple=false }={}) {
...
+  if ( config.activity && config.formula ) {
+    console.warn(`Activity ID and formula found while enriching ${config._input}, only one is supported.`);
+    return null;
+  }
+
   const formulaParts = [];
   if ( config.formula ) formulaParts.push(config.formula);
-  for ( const value of config.values ) formulaParts.push(value);
+  for ( const value of config.values ) {
+    if ( value in CONFIG.DND5E.attackModes ) config.attackMode = value;
```
Fyorl, line 137: Maybe just `mode`? This is the `attack` enricher after all so writing `[[/attack attackMode=foo]]` feels a bit redundant. I don't feel strongly though, given that the `attackMode` key can be entirely omitted, which is probably a more idiomatic.

Fyorl (approved):

module/config.mjs
```diff
@@ -2719,6 +2719,31 @@ preLocalize("attackClassifications", { key: "label" });
...
+  twoHanded: {
+    label: "DND5E.ATTACK.Mode.TwoHanded"
+  },
+  offhand: {
+    label: "DND5E.ATTACK.Mode.Offhand"
+  },
+  thrown: {
+    label: "DND5E.ATTACK.Mode.Thrown"
+  },
+  "thrown-offhand": {
```
krbz999, line 2739: Any strong feelings regarding renaming this `thrownOffhand`?

arbron: It complicates some code elsewhere, because as it is we can simply check `mode.endsWith("offhand")` to match both `offhand` and `thrown-offhand` modes.

module/enrichers.mjs
arbron: I'd say yes if it wasn't also used in the damage enricher, where `mode` is a bit more ambiguous.

PR #4824, opened 2024-12-04
[#4631, #4569] Add compendium source where applicable.

### LINKED ISSUE #4569: Items added via a container drop (eg `Item.implementation.createWithContents`) do not have `_stats` and other properties set
The method uses `toObject` rather than `game.items.fromCompendium`.

### LINKED ISSUE #4631: Item._stats.compendiumSource Not Added with Advancements
**Bug:**
When adding a class, subclass, race, or background and going through the advancements, no compendiumSource value is added to the items _stats field. compendiumSource is also not added to the class, subclass, race, or background when clicking on "Add ______" in the character sheet and choosing the respective item. 

**Versions Details:**
Foundry Version: 12.331
DND5E Version: 4.0.4
Hosting: Local

**Steps to Reproduce:**
1. Create a new world with the DND5E game system.
2. Create a new Player Actor named TestActor.
3. Add one of the SRD class, subclass, race, or background to the Actor and click through the advancements.
4. Open the javascript console and type ```game.actors.getName("TestActor").items```.
5. Hit enter and explore the items of the character.
6. Opening the _stats field of any item will show that it's compendiumSource field is null.

**Other Notes:**
This bug was found when trying to create a macro that updates items on a world actor based on the items in their original compendiums. I noticed that the sorcedItems field does contain the items and their full UUIDs, however upon using those and updating the items via .updateEmbeddedDocuments(), the sorcedItems map no longer contains the updated items. Is this intended?

krbz999:
Closes #4631.
Closes #4569.

Was surprised to learn that `fromCompendium` adds `_stats.compendiumSource` even if the item is not from a compendium so this *could* be simplified by just using the `uuid` of the given item but I decided not go that route since it seems misaligned with core behavior, at the very least.

### DIFF
--- a/module/documents/advancement/advancement.mjs
+++ b/module/documents/advancement/advancement.mjs
@@ -355,8 +355,10 @@ export default class Advancement extends BaseAdvancement {
   async createItemData(uuid, id) {
     const source = await fromUuid(uuid);
     if ( !source ) return null;
+    const { _stats } = game.items.fromCompendium(source);
     return source.clone({
       _id: id ?? foundry.utils.randomID(),
+      _stats: _stats,
       "flags.dnd5e.sourceId": uuid,
       "flags.dnd5e.advancementOrigin": `${this.item.id}.${this.id}`
     }, {keepId: true}).toObject();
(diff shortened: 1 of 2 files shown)


module/documents/advancement/advancement.mjs
```diff
@@ -355,8 +355,10 @@ export default class Advancement extends BaseAdvancement {
   async createItemData(uuid, id) {
     const source = await fromUuid(uuid);
     if ( !source ) return null;
+    const { _stats } = game.items.fromCompendium(source);
     return source.clone({
       _id: id ?? foundry.utils.randomID(),
+      _stats: _stats,
```
Fyorl, line 361:
```suggestion
    return source.clone({
      _stats,
      _id: id ?? foundry.utils.randomID(),
```

Fyorl (approved):

Issue #4828, opened 2024-12-04
Add `extended` format for `[[/attack]]` and `[[/damage]]` enrichers


arbron:
Most NPC stat blocks in 5e follow a similar pattern in how they express attack and damage entries:

```
// Attack 2024 Style:
Melee or Ranged Attack Roll: +14, reach 10 ft. or range 60 ft.
Melee Attack Roll: +5, reach 5 ft.
Ranged Attack Roll: +5, range 30/90 ft.

// Attack 2014 Style:
Melee or Ranged Weapon Attack: +5 to hit, reach 5 ft. or range 20/60 ft., one target.
Melee Weapon Attack: +5 to hit, reach 5 ft., one target.

// Damage 2024 Style:
Hit: 6 (1d6 + 3) Slashing damage.

// Damage 2014 Style:
Hit: 6 (1d6 + 3) slashing damage.
```

Adding an additional formatting option to the attack and damage enrichers to display this extended format could simplify the creation of NPCs and allow more reusable monster features.

The variations between the two styles can be handled using the rules data already present in the NPC/item source data, falling back to the global rules setting if no rules data is present on the actor.

PR #4829, opened 2024-12-04
[#4828] Add extended formatting option to attack & damage enrichers

arbron:
Adds an new formatting option to the attack and damage enrichers that displays them in a fashion similar to full NPC stat blocks:

```html
<!-- Normal enrichers -->
<em>Melee Weapon Attack:</em> [[/attack]], reach 15 ft.<strong>,</strong> one target. <em>Hit:</em> [[/damage average]] damage.

<!-- Extended enrichers -->
[[/attack format=extended]]. [[/damage format=extended]].
```

The formatting of the generated content will adjust slightly based on what rules version is being used. This will use the rules version defined in the containing actor if present, then in the containing item, and if neither is found then it will use the global rules setting.

### Todo
- [ ] Implement target block
- [ ] Swap to properly localized units

### DIFF
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -371,7 +371,7 @@ export default class AttackActivityData extends BaseActivityData {
       if ( this.range.override ) range = `${this.range.value} ${this.range.units ?? ""}`;
       else {
         const { value, long, units } = this.item.system.range;
-        if ( value !== long ) range = `${value}/${long} ${units}`;
+        if ( long && (value !== long) ) range = `${value}/${long} ${units}`;
         else range = `${value} ${units}`;
       }
       parts.push(game.i18n.format("DND5E.RANGE.Formatted.Range", { range }));
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -131,6 +131,25 @@ export default class AttackActivityData extends BaseActivityData {
   }
 
   /* -------------------------------------------- */
+
+  /**
+   * Potential attack types when attacking with this activity.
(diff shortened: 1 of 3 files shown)


module/data/activity/attack-data.mjs
```diff
@@ -329,6 +348,40 @@ export default class AttackActivityData extends BaseActivityData {
...
+      parts.push(game.i18n.format("DND5E.RANGE.Formatted.Reach", { reach: `${reach ?? 5} ${units}` }));
+    }
+
+    // Add range for ranged or thrown weapons, unless the activity is explicitly specified as melee
+    if ( this.validAttackTypes.has("ranged") ) {
+      let range;
+      if ( this.range.override ) range = `${this.range.value} ${this.range.units ?? ""}`;
+      else {
+        const { value, long, units } = this.item.system.range;
+        if ( value !== long ) range = `${value}/${long} ${units}`;
```
Fyorl, line 374: We should check there is actually a value for `long`. I know it's required to have both increments for a ranged or thrown weapon, but the form doesn't enforce it.

```suggestion
        if ( long && (value !== long) ) range = `${value}/${long} ${units}`;
```

module/enrichers.mjs
```diff
@@ -489,6 +553,7 @@ async function enrichDamage(configs, label, options) {
     }
   }
   config.damageTypes = config.damageTypes.map(t => t.replace("/", "|"));
```
Fyorl, line 555: Noticed a bug where this line throws an error if you don't specify damage type, e.g. `[[/damage 1d6]]`.

Fyorl (approved):

PR #4833, opened 2024-12-05
Improve consistency of Item Sheet display

### LINKED ISSUE #4364: items Sheets not displaying updated properties after being enchanted
dnd version : 4.0.2
foundry version : 12.331
0 modules enabled

items Sheets display the properties of the initail item and not the overriden properties given by the enchantment.
Enchantment has : 

image

system.properties and overrides  are correctly set up after enchantement is applied but do not reflect in the weapons properties of the item sheet

image

note that this work from tsystem perspective :
=> the "Attunement" section appears, meaning that the item is considered magical (and the magical icon shows on the inventory tab of the PC)
=> the Ammunition section is not shown when launching an attack with the enchanted weapon and the "Use Ammunition' section disapparead from the item sheet.

### LINKED ISSUE #4381: Compendium-locked item properties are difficult to parse
While there is a color difference here between the checked and unchecked properties, it is not very obvious. One option might be to bold the text of any checked property?

image

### LINKED ISSUE #4809: "Base Weapon" select on item sheets show the options for the current value of "Weapon Type" which can be from an Enchantment, rather than what the item is by default
eg

weapon (simple ranged, shortbow)

enchantment sets `system.type.value` to `martialR`.

dropdown on sheet now shows only Martial Ranged options for base weapon.

Fyorl:
- Closes #4381 
- Closes #4809 
- Closes #4364 

This brings the Item sheet in-line with the V2 Actor sheet in terms of consistency around editability in Edit vs. Play mode. Form fields are now disabled in Play mode, and will now show any enchanted values, where before it was not possible to see them. Switching the sheet to edit mode will make the fields editable, with their source values displayed, avoiding the issue of potentially persisting enchantment-modified values, or making changes that do not visually update due to being enchantment-modified.

The sheets will now default to opening in Edit mode also, where appropriate, as most contexts where the Item sheet is opened imply editability (e.g. the right-click context menu on Actor sheets is labelled 'edit', and opening an Item from the sidebar is assumed for editing purposes). Contexts where the Item is not editable, such as a locked compendium, or when viewing an Item on an Actor with Observer permissions, the sheet will display in Play mode as Edit mode is not available in those contexts.

I've also improved a few areas of styling that were not displaying correctly in a disabled context.

image

image

### DIFF
--- a/less/v2/forms.less
+++ b/less/v2/forms.less
@@ -168,6 +168,7 @@
     align-items: center;
     gap: 4px;
 
+    &:has(input[type=checkbox]:disabled, dnd5e-checkbox[disabled]) { cursor: default; }
     > span { line-height: calc(var(--form-field-height) + 1px); }
   }
 
@@ -408,7 +409,7 @@
 
   select {
     height: var(--form-field-height);
-    cursor: pointer;
+    &:not(:disabled) { cursor: pointer; }
   }
 
   :is(document-tags, multi-select) .tags.input-element-tags .tag {
@@ -418,6 +419,12 @@
     background: var(--dnd5e-background-card);
     border-radius: 3px;
     &:hover a { text-shadow: 0 0 6px var(--color-shadow-primary); }
+  }
+
+  :is(document-tags, multi-select):has(select:disabled) .tags.input-element-tags .tag {
+    cursor: default;
+    a { cursor: default; }
+    &:hover a { text-shadow: none; }
   }
 
   multi-select.hidden-tags .tags.input-element-tags { display: none; }
@@ -474,7 +481,7 @@
   dnd5e-checkbox {
     &[disabled]:not([checked]) {
       --checkbox-border-color: var(--color-text-light-6);
-      --checkbox-empty-color: var(--color-text-light-6);
+      --checkbox-empty-color: transparent;
     }
   }
 
--- a/module/applications/item/item-directory.mjs
+++ b/module/applications/item/item-directory.mjs
@@ -1,4 +1,5 @@
 import Item5e from "../../documents/item.mjs";
+import ItemSheet5e2 from "./item-sheet-2.mjs";
 
 /**
  * Items sidebar with added support for item containers.
@@ -23,5 +24,14 @@ export default class ItemDirectory5e extends ItemDirectory {
 
     // Let parent method perform sorting
     super._handleDroppedEntry(target, item.toDragData());
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  async _onClickEntryName(event) {
+    const { entryId } = event.target.closest("[data-entry-id]")?.dataset ?? {};
+    const item = this.collection.get(entryId);
+    item?.sheet.render(true, { mode: ItemSheet5e2.MODES.EDIT });
   }
 }
--- a/module/applications/actor/sheet-v2-mixin.mjs
+++ b/module/applications/actor/sheet-v2-mixin.mjs
@@ -2,6 +2,7 @@ import * as Trait from "../../documents/actor/trait.mjs";
 import { formatNumber, simplifyBonus, splitSemicolons, staticID } from "../../utils.mjs";
 import Tabs5e from "../tabs.mjs";
 import DocumentSheetV2Mixin from "../mixins/sheet-v2-mixin.mjs";
+import ItemSheet5e2 from "../item/item-sheet-2.mjs";
 
 /**
  * Adds common V2 Actor sheet functionality.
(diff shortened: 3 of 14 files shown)


krbz999:
Probably closes #4809. :)

Fyorl:
> Probably closes #4809. :)

Well it would if I had known about that issue, but it at least provides a clear way to address it. I've pushed a fix for that now, and I imagine there will be a few other places where the source vs. enchantment-modified data is being mixed together that we will have to squash as we find them.

krbz999:
> mixed together that we will have to squash as we find them.

I took the liberty of doing a quick browse, and most fields are only contextually hidden, which should be fine. Only contender is the Area fieldset of a spell item, but I think it's gucci.

less/v2/forms.less
```diff
@@ -168,6 +168,7 @@
     align-items: center;
     gap: 4px;
 
+    &:has(input[type=checkbox]:disabled, dnd5e-checkbox[disabled]) { cursor: default; }
```
arbron, line 171: The select elements also have a pointer when disabled.

less/v2/forms.less
```diff
@@ -420,6 +421,11 @@
     &:hover a { text-shadow: 0 0 6px var(--color-shadow-primary); }
   }
 
+  :is(document-tags, multi-select):has(select:disabled) .tags.input-element-tags .tag{
```
arbron, line 424: In testing this I noticed that you can still delete damage types on weapons when the sheet is disabled (you can even click to delete them on items in locked compendiums, but the change isn't saved of course). I think we fixed that issue for activities inside our custom `_disableFields` implementation which specifically disables `document-tags` and `multi-select` elements.

arbron (changes requested):
The sheets look good.

I have a concern about opening all sheets in edit mode particularly for classes/background/species items. Because they only have one description that means that people opening those sheets sees the item's raw description, rather than the enriched description. Could be fine I suppose.

Fyorl:
> I have a concern about opening all sheets in edit mode particularly for classes/background/species items. Because they only have one description that means that people opening those sheets sees the item's raw description, rather than the enriched description. Could be fine I suppose.

You are right, this probably needs some more thought. At the very least it should open in Edit mode if you are deliberately accessing it via the 'edit' context menu option. But there are a few other ways to view item sheets, in particular the ones you mentioned which can be opened with left-click directly from the actor sheet. I'm less certain of what the default should be when opening an item from the sidebar tab (or unlocked compendium).

krbz999:
> I'm less certain of what the default should be when opening an item from the sidebar tab (or unlocked compendium).

Having messed with similar in a different system, I more often than not ended up wanting Edit Mode primarily on these.

Fyorl:
Item sheets will now open in Edit mode when explicitly choosing 'edit' in a context menu. Things like left-clicking the species item or left-clicking a container to view its contents will open in Play mode. Items in the sidebar and unlocked compendia will also open in Edit mode when opened.

module/applications/item/item-directory.mjs
```diff
@@ -24,4 +25,13 @@ export default class ItemDirectory5e extends ItemDirectory {
...
     super._handleDroppedEntry(target, item.toDragData());
   }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  async _onClickEntryName(event) {
+    const { entryId } = event.target.closest("[data-entry-id]")?.dataset ?? {};
+    const item = this.collection.get(entryId);
+    item?.sheet.render(true, { mode: ItemSheet5e2.MODES.EDIT });
```
Fyorl, line 35: For sidebar I'm still a bit unsure what the preferred mode would be. Maybe it should only force a mode if the sheet hasn't been opened yet?

arbron: I think we should only ever apply the mode on first render. If you've previously had the sheet open from anywhere it would be nice if it retained the previous mode on re-open especially since the View button isn't displayed in the actor inventory unless you don't have edit permission, so if you want to view an item in play mode you have to change the mode every time you open it (or we could make the View button always visible so players always have control over what mode it opens with).

arbron (approved):

module/applications/item/item-directory.mjs
Fyorl: That does seem like it would offer the least friction. I will make the change.

Fyorl: I have made this change now, though I think adding a separate view context menu option could also work. These behaviours are mutually exclusive so I think it might come down to user preference.

I think there are still issues with this technique. Presumably if an item is enchanted to be magical, for example, then you should be able to toggle attunement in Play mode, but that field will be disabled unless in Edit mode.

I think a more complete solution will involve identifying areas, like attunement, that are the target of various enchantments, and moving those out of this form-based editing, and into the sheet 'proper'. Attunement could be a header button, like equipped and identified, for example.

If enchantments get particularly sophisticated, then perhaps we need an entirely bespoke 'details' tab that is not a form at all when in Play mode, and only show the form in Edit mode.

For now, though, this is probably good enough to unblock some issues while we gather feedback.

arbron (approved):

PR #4838, opened 2024-12-06
Track advantage modes

krbz999:
Adds a new `AdvantageModeField` which tracks the number of effects that apply advantage and disadvantage. The final value is then either -1 if only disadvantages exist (including the initial value), 1 if only advantages exist (including initial value), otherwise 0 if there are both advantages and disadvantages (including initial value).

TODO: Add +5 (-5) to the skill's passive if the actor has advantage (disadvantage).

### DIFF
--- a/module/data/fields/advantage-mode-field.mjs
+++ b/module/data/fields/advantage-mode-field.mjs
@@ -1,0 +1,59 @@
+/**
+ * Subclass of NumberField that tracks the number of changes made to a roll mode.
+ */
+export default class AdvantageModeField extends foundry.data.fields.NumberField {
+  /** @inheritDoc */
+  static get _defaults() {
+    return foundry.utils.mergeObject(super._defaults, {
+      choices: [-1, 0, 1],
+      initial: 0,
+      label: "DND5E.AdvantageMode"
+    });
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Number of advantage modifications.
+   * @type {number}
+   */
+  #advantage;
+
+  /* -------------------------------------------- */
+
+  /**
+   * Number of disadvantage modifications.
+   * @type {number}
+   */
+  #disadvantage;
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  initialize(value, model, options={}) {
+    this.#advantage = Number(value === 1);
+    this.#disadvantage = Number(value === -1);
+    return value;
+  }
+
+  /* -------------------------------------------- */
+  /*  Active Effect Integration                   */
+  /* -------------------------------------------- */
+
+  /** @override */
+  applyChange(value, model, change) {
+    const delta = this._castChangeDelta(change.value);
+    if ( change.mode === CONST.ACTIVE_EFFECT_MODES.CUSTOM ) {
+      return this._applyChangeCustom(value, delta, model, change);
+    }
+    switch (delta) {
+      case 1:
+        this.#advantage++;
+        break;
+      case -1:
+        this.#disadvantage++;
+        break;
+    }
+    return Math.sign(this.#advantage) - Math.sign(this.#disadvantage);
+  }
+}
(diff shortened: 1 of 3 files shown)


arbron (approved):

module/data/fields/advantage-mode-field.mjs
```diff
@@ -0,0 +1,59 @@
...
+    this.#disadvantage = Number(value === -1);
+    return value;
+  }
+
+  /* -------------------------------------------- */
+  /*  Active Effect Integration                   */
+  /* -------------------------------------------- */
+
+  /** @override */
+  applyChange(value, model, change) {
```
Fyorl, line 44: Does this mean that we essentially ignore the mode (if it's not CUSTOM) for these fields? I guess it's not really obvious what the mode should be, but I think it might be useful to reserve some modes in case we want to implement some special semantics here later. If people are creating AEs with any mode and having it work, then if we try to change the semantics of a mode later, it will break their AEs and we are on the hook for a migration.

I think OVERRIDE is probably the one we should support, and the rest should explicitly do nothing (i.e. not make a change), so that it's clear that they do not work, and people don't start creating AEs with the other modes.

module/data/fields/advantage-mode-field.mjs
```diff
@@ -0,0 +1,59 @@
...
+    const delta = this._castChangeDelta(change.value);
+    if ( change.mode === CONST.ACTIVE_EFFECT_MODES.CUSTOM ) {
+      return this._applyChangeCustom(value, delta, model, change);
+    }
+    switch (delta) {
```
Fyorl, line 49:
```suggestion
    switch ( delta ) {
```

module/data/fields/advantage-mode-field.mjs
krbz999: Since it is numeric values, we have so far been using UPGRADE and DOWNGRADE, so I think those two should be supported. I think ADD makes sense as well (e.g. you add a new instance of advantage or disadvantage), and OVERRIDE is arguably similar.

CUSTOM is of course only for modules. I don't disagree that MULTIPLY should perhaps be ignored, but I did not consider it at the time.

Anyway that was my reasoning.

Fyorl: It's just that, at the moment, ADD, OVERRIDE, UPGRADE, and DOWNGRADE all the do the same thing, and I wondered if maybe it would be useful to reserve some of these modes for special behaviour around advantage and disadvantage that we might want to implement later. The idea was to pre-empt people creating AEs with these modes and having their behaviour change out from underneath them later, but it sounds like people are already creating AEs despite these modes not actually being supported yet?

Anyway, UPGRADE and DOWNGRADE happen to work because it's numerical data, but advantage and disadvantage aren't numerical concepts. You can't say "roll this with advantage if not already rolling 'higher' than advantage", like you would with the Gauntlets of Ogre Power, for example, which say to 'set your Strength score to 19 if not already higher'.

krbz999: Sounds fine to me. If an actor has both OVERRIDE 1 and OVERRIDE -1 or both DOWNGRADE -1 and UPGRADE 1, do we simply want to let AEs handle them the standard way or should we treat this as 0?

Fyorl: I think these modes should probably do nothing, i.e. not modify the source value at all, to make it clear that they are unsupported at this time. People hopefully won't create a bunch of AEs using OVERRIDE, thinking that it works one way, and then find out later that we implemented different behaviour.

So, in summary: OVERRIDE, UPGRADE, DOWNGRADE, and MULTIPLY do not make any changes at all, and ADD applies advantage/disadvantage per the rules on multiple sources of advantage and disadvantage.

krbz999: Ah I was too fast for you, I've made a second PR here https://github.com/foundryvtt/dnd5e/pull/4890
But I can adjust the remaining 3 methods if we feel strongly about it.

Fyorl: No that's fine. I only proposed they do nothing in the absence of actual supported functionality, but if you've added the support for them already then that's great. Sorry for the churn. I see there are two commits so I can just review the first one and if that's good we can go with that, thank you.

PR #4839, opened 2024-12-06
[#4384, #4828] Adds plural targets, add target to attack enricher

### LINKED ISSUE #4384: Pluralize `Target` display of features that target multiple entities
The `Target` of a feature targeting 2 creatures shows as `2 Creature`.

image

image

It would be nice if this could check if>1 and convert to `2 Creatures`.

This would also affect `Ally`, `Enemy`, `Object`, `Space`, `Creature or Object`, and `Willing Creature`. Most of these would append an `s` but   `2 Ally` would become `2 Allies` and `2 Enemy` would become `2 Enemies`, etc.

### LINKED ISSUE #4828: Add `extended` format for `[[/attack]]` and `[[/damage]]` enrichers

arbron:
Adds the targeting information to the extended attack enricher when using the 2014 rules (e.g. "one target"). Two additional localization features were added to support this.

First, `formatNumber` now takes the `spellOut` option which will attempt to write the number out if the current language provides a written form (e.g. `one` instead of `1`). The does not fall back to the default language, so a language where this doesn't make sense can simply not provide the appropriate localization keys and it will just use the numeric form instead. For English this is set up to spell out the numbers 1–9 only.

Second, the targeting localization was modified to include counted plural localizations so now the system will be able to display "2 creatures" rather than "2 creature" when it displays target labels. This change required changing the `individualTargetTypes` config entry to contain objects.

Closes #4384
Closes #4828

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3268,7 +3268,7 @@
     "CreatureOrObject": {
       "Label": "Creature or Object",
       "Counted": {
-        "any": "any creature or objects",
+        "any": "any creatures or objects",
         "every": "every creature or object",
         "one": "{number} creature or object",
         "other": "{number} creatures or objects"
--- a/lang/en.json
+++ b/lang/en.json
@@ -2445,6 +2445,18 @@
   }
 },
 
+"DND5E.NUMBER": {
+  "1": "one",
+  "2": "two",
+  "3": "three",
+  "4": "four",
+  "5": "five",
+  "6": "six",
+  "7": "seven",
+  "8": "eight",
+  "9": "nine"
+},
+
 "DND5E.Number": "Number",
 "DND5E.OtherFormula": "Other Formula",
 "DND5E.PactMagic": "Pact Magic",
@@ -3146,6 +3158,13 @@
 "DND5E.Suppressed": "Suppressed",
 
 "DND5E.TARGET": {
+  "Action": {
+    "PlaceTemplate": "Place Measured Template"
+  },
+  "Count": {
+    "Any": "Any",
+    "Every": "Every"
+  },
   "FIELDS": {
     "target": {
       "label": "Targeting",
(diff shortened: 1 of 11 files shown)


module/config.mjs
```diff
@@ -2513,22 +2513,57 @@ preLocalize("encumbrance.effects", { key: "name" });
...
+    label: "DND5E.TARGET.Type.Space.Label",
+    counted: "DND5E.TARGET.Type.Space.Counted"
+  },
+  creatureOrObject: {
+    label: "DND5E.TARGET.Type.CreatureOrObject.Label",
+    counted: "DND5E.TARGET.Type.CreatureOrObject.Counted"
+  },
+  any: {
+    label: "DND5E.TARGET.Type.Any.Label",
+    scalar: false
```
Fyorl, line 2558: If 'any' is 'creature, object, or location', per the rules on making attacks, then it should be scalar, to allow for targeting, say '3 creatures, objects, or locations' (i.e. '3 targets').

module/data/shared/target-field.mjs
```diff
@@ -76,25 +80,39 @@ export default class TargetField extends SchemaField {
...
+    const affectsConfig = CONFIG.DND5E.individualTargetTypes[this.target.affects.type];
+    this.target.affects.labels = {
+      sheet: affectsConfig?.counted ? game.i18n.format(
+        `${affectsConfig.counted}.${this.target.affects.count ? pr.select(this.target.affects.count) : "other"}`, {
+          number: this.target.affects.count ? formatNumber(this.target.affects.count)
+            : game.i18n.localize(`DND5E.TARGET.Count.${this.target.template.type ? "Every" : "Any"}`)
+        }
+      ).trim().capitalize() : (affectsConfig?.label ?? ""),
+      statblock: game.i18n.format(
+        `${affectsConfig?.counted ?? "DND5E.TARGET.Type.Target.Counted"}.${pr.select(this.target.affects.count || 1)}`,
```
Fyorl, line 110: Do we need to handle the case of `self` here, which doesn't have a `counted`? Maybe not for NPC stat blocks?

module/utils.mjs
```diff
@@ -33,10 +33,12 @@ export function formatModifier(mod) {
  * A helper for using Intl.NumberFormat within handlebars.
  * @param {number} value    The value to format.
  * @param {object} options  Options forwarded to {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat}
+ * @param {boolean} [options.spellOut]  Spell out number as full word, if possible.
```
Fyorl, line 36: Maybe `wordForm` or `words`? I think 'spell out' is correct it just doesn't feel very... technical to me. I'm not sure.

Fyorl (changes requested):

lang/en.json
```diff
@@ -2440,6 +2440,18 @@
   }
 },
 
+"DND5E.NUMBER": {
```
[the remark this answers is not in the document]
arbron: Sadly not :(

module/utils.mjs
arbron: `words` matches well with `numerals`, so I'll go with that.

module/data/shared/target-field.mjs
arbron: Haven't seen any attacks that target self yet, maybe just a wait-and-see.

Fyorl: Yeah, fair enough

Fyorl (approved):

PR #4847, opened 2024-12-08
Add parameter for preventing subsequent actions

krbz999:
I had a use case for just getting the return of the usage without the follow-up actions.

Note: I considered `followup` as the name for the property but did not want to clash with a possible future feature. Open to suggestions, however.

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -157,7 +157,7 @@ export default function ActivityMixin(Base) {
      *                                                 scaling is not allowed.
      * @property {object} spell
      * @property {number} spell.slot                   The spell slot to consume.
-     * @property {boolean} [triggerSubsequentActions]  If explicitly false, subsequent actions will not be triggered.
+     * @property {boolean} [subsequentActions=true]    Trigger subsequent actions defined by this activity.
      * @property {object} [cause]
      * @property {string} [cause.activity]             Relative UUID to the activity that caused this one to be used.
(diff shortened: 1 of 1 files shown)


module/documents/activity/mixin.mjs
```diff
@@ -157,6 +157,7 @@ export default function ActivityMixin(Base) {
      *                                                 scaling is not allowed.
      * @property {object} spell
      * @property {number} spell.slot                   The spell slot to consume.
+     * @property {boolean} [triggerSubsequentActions]  If explicitly false, subsequent actions will not be triggered.
```
arbron, line 160: Maybe just `subsequentActions`, then we can indicate the inferred default value without having to re-indent the other config options :)

```suggestion
     * @property {boolean} [subsequentActions=true]    Trigger subsequent actions defined by this activity.
```

module/documents/activity/mixin.mjs
```diff
@@ -321,7 +322,9 @@ export default function ActivityMixin(Base) {
       }
 
       // Trigger any primary action provided by this activity
-      activity._triggerSubsequentActions(usageConfig, results);
+      if ( usageConfig.triggerSubsequentActions !== false ) {
```
arbron, line 325:
```suggestion
      if ( usageConfig.subsequentActions !== false ) {
```

arbron (changes requested):

arbron (approved):

PR #4849, opened 2024-12-08
Socket framework proof of concept

krbz999:
Added a new class, `Sockets5e`, available in `game.dnd5e.sockets`. It registers socket configurations, each with `event`, `initiate`, and `finalize`.

We can then call `game.dnd5e.sockets.initiate(eventName, ...args)`, which initiates a specific function, and then - if needed - emits to other clients.

The `initiate` method in the socket config is used to set up the data given a few parameters.

The `finalize` method in the socket config is calling the relevant function under the assumption that the user calling it is meant to do so.

This PR includes a sample setup with a "grantEffect" configuration, which allows a user to create an effect on an actor they may or may not be owner of.

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -25,6 +25,7 @@ import {default as registry} from "./module/registry.mjs";
 import * as utils from "./module/utils.mjs";
 import {ModuleArt} from "./module/module-art.mjs";
 import registerModuleData from "./module/module-registration.mjs";
+import Sockets5e from "./module/sockets.mjs";
 import Tooltips5e from "./module/tooltips.mjs";
 
 /* -------------------------------------------- */
@@ -88,6 +89,9 @@ Hooks.once("init", function() {
 
   // Configure bastions
   game.dnd5e.bastion = new documents.Bastion();
+
+  // Configure sockets
+  game.dnd5e.sockets = new Sockets5e();
 
   // Configure tooltips
   game.dnd5e.tooltips = new Tooltips5e();
--- a/module/sockets.mjs
+++ b/module/sockets.mjs
@@ -1,0 +1,67 @@
+import sockets from "./sockets/_module.mjs";
+
+/**
+ * @typedef {object} SocketEventConfig
+ * @property {string} event           The unique socket event name.
+ * @property {Function} initiate      A function used to initiate the socket event.
+ * @property {Function} finalize      A function that finalizes the socket event as the valid user(s).
+ */
+
+export default class Sockets5e {
+  constructor() {
+    game.socket.on("system.dnd5e", this.#handleSocket.bind(this));
+    for (const config of sockets) this.#register(config);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Store event configurations.
+   * @type {Map<string, SocketEventConfig>}
+   */
+  #events = new Map();
+
+  /* -------------------------------------------- */
+
+  /**
+   * Register a sovket event config.
+   * @param {SocketEventConfig} config
+   */
+  #register(config) {
+    this.#events.set(config.event, config);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Handle the socket event for the correct user(s).
+   * @param {object} data              Socket data.
+   * @param {boolean} [data._emit]     Internally used parameter for whether this was initiated or emitted.
+   * @returns {*}
+   */
+  #handleSocket({ _emit, ...data }) {
+    if ( !data.userIds.includes(game.user.id) ) {
+      if (!_emit) game.socket.emit("system.dnd5e", { ...data, _emit: true });
+      else return;
+    }
+
+    const config = this.#events.get(data.event);
+    if ( config ) return config.finalize(data);
+
+    throw new Error(`'${data.event}' is not a valid socket event action!`);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Initiate a given event, which may or may not be emitted.
+   * @param {string} event      The event name.
+   * @param {any[]} args        Function parameters for the handler.
+   */
+  initiate(event, ...args) {
+    const config = this.#events.get(event);
+    if ( !config ) return;
+    const data = config.initiate(...args);
+    this.#handleSocket(data);
+  }
+}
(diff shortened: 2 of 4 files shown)


module/sockets.mjs
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Handle the socket event for the correct user(s).
+   * @param {object} data              Socket data.
+   * @param {boolean} [data._emit]     Internally used parameter for whether this was initiated or emitted.
+   * @returns {*}
+   */
+  #handleSocket({ _emit, ...data }) {
+    if ( !data.userIds.includes(game.user.id) ) {
+      if (!_emit) game.socket.emit("system.dnd5e", { ...data, _emit: true });
+      else return;
```
arbron, line 45: Seems like this shouldn't be in an `else`, otherwise it causes emits the socket event and then also attempts to trigger locally.

module/sockets/grant-effect.mjs
```diff
@@ -0,0 +1,68 @@
+/**
```
arbron, line 1: Seems like we could stick this into a class to formalize the interface a bit. Perhaps an abstract `SocketEvent` class that defines placeholder `name` property and abstract `initiative` and `finalize` methods.

module/sockets/grant-effect.mjs
```diff
@@ -0,0 +1,68 @@
...
+ * @typedef {object} GrantEffectConfig
+ * @property {string} event           The event name.
+ * @property {string} actor           The uuid of the actor to receive the effect.
+ * @property {object} effectData      Data used to create the effect.
+ * @property {string[]} userIds       The ids of the users that should perform the operation.
```
arbron, line 6: It looks like `userIds` is required to be part of every socket event data definition, so that should be split off into a separate typedef that this one extends.

module/sockets/grant-effect.mjs
```diff
@@ -0,0 +1,68 @@
...
+    throw new Error("You must supply both an actor and effect instance for this operation.");
+  }
+
+  // Get a valid user to perform the operation.
+  // TODO: improve in v13
+  const userId = actor.isOwner ? game.user.id : game.users.find(user => {
+    return user.active && actor.testUserPermission(user, "OWNER");
+  })?.id;
+
+  if (!userId) {
```
arbron, line 46:
```suggestion
  if ( !userId ) {
```

dnd5e.mjs
```diff
@@ -25,6 +25,7 @@ import {default as registry} from "./module/registry.mjs";
 import * as utils from "./module/utils.mjs";
 import {ModuleArt} from "./module/module-art.mjs";
 import registerModuleData from "./module/module-registration.mjs";
+import Sockets5e from "./module/sockets.mjs";
```
arbron, line 28: It would be nice to have the socket orchestration class inside the `sockets` folder, rather than having two entries named `sockets` inside the module folder.

arbron (changes requested):
Sockets!

module/sockets.mjs
krbz999: Literally went through the logic several times.

Gosh darn.

module/sockets/socket-event.mjs
```diff
@@ -0,0 +1,33 @@
...
+ * @property {string} event     The unique event data name.
+ * @property {string[]} userIds     The ids of users that should perform the operation.
+ */
+
+export default class SocketEvent {
+  /**
+   * The socket event name, which must be unique and is used to call the event.
+   * @type {string}
+   */
+  static eventName;
```
krbz999, line 12: Might be cleaner to have these be non-static methods. Lemme know what you think.

module/sockets/socket-event.mjs
```diff
@@ -0,0 +1,33 @@
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * As a user allowed to do so, perform the operation.
+   * @param {EventData} data      The event data.
```
arbron, line 18:
```suggestion
   * @param {EventData} data      The event data.
   * @abstract
```

module/sockets/socket-event.mjs
```diff
@@ -0,0 +1,33 @@
...
+   */
+  static async finalize(data) {
+    throw new Error("The 'finalize' method of a socket event must be subclassed.");
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Initiate the socket event. Subclasses can and should change the signature of this method.
+   * @returns {EventData}
```
arbron, line 28:
```suggestion
   * @returns {EventData}
   * @abstract
```

module/sockets/grant-effect.mjs
```diff
@@ -0,0 +1,57 @@
...
+    if ( actor?.documentName !== "Actor" ) return;
+    getDocumentClass("ActiveEffect").create(data.effectData, { parent: actor });
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Initiate the socket event.
+   * @param {Actor5e} actor                 The actor to receive the effect.
+   * @param {ActiveEffecet5e} effect        The effect to be duplicated onto the actor.
```
arbron, line 30:
```suggestion
   * @param {Actor5e} actor               The actor to receive the effect.
   * @param {ActiveEffecet5e} effect      The effect to be duplicated onto the actor.
```

module/sockets/socket-event.mjs
```diff
@@ -0,0 +1,33 @@
+/**
+ * @typedef {object} EventData
+ * @property {string} event     The unique event data name.
```
arbron, line 3:
```suggestion
 * @property {string} event         The unique event data name.
```

module/sockets/socket-event.mjs
arbron: I think this works fine, the only real thing that would get us would be losing the `static` keywords a few places, but we'd have to create an instance during registration in that case, so it would be a trade-off.

arbron (changes requested):
We'll see what Kim thinks during the dnd meeting :)

module/sockets/grant-effect.mjs
krbz999: We'll pretend we both saw the typo.

krbz999:
Probably better handled using `CONFIG.queries`.

PR #4850, opened 2024-12-09
[#4827] Add hook for when NPC stat blocks are rendered

### LINKED ISSUE #4827: Add a hook for npc stat block embeds
This would be useful for adding an additional note next to the size/type/alignment text for example. The *Flee, Mortals* book adds a "Role" property which it would be very good to display there.

I imagine the arguments supplied don't need to be complex; only the `Actor` and the final html content would be needed, but the configuration of the embed enricher itself would also be good.

arbron:
Adds a `dnd5e.renderNPCStatBlock` hook that provides the actor, collection of HTML elements, and embedding configuration for modification before the stat block is fullying inserted.

Closes #4827

### DIFF
--- a/module/data/actor/npc.mjs
+++ b/module/data/actor/npc.mjs
@@ -410,6 +410,18 @@ export default class NPCData extends CreatureTemplate {
     const context = await this._prepareEmbedContext();
     const section = document.createElement("section");
     section.innerHTML = await renderTemplate("systems/dnd5e/templates/actors/embeds/npc-embed.hbs", context);
+
+    /**
+     * A hook event that fires after an embedded NPC stat block rendered.
+     * @function dnd5e.renderNPCStatBlock
+     * @memberof hookEvents
+     * @param {Actor5e} actor                   NPC being embedded.
+     * @param {HTMLCollection} contents         HTML elements to be inserted.
+     * @param {DocumentHTMLEmbedConfig} config  Configuration for embedding behavior.
+     * @param {EnrichmentOptions} options       Original enrichment options.
+     */
+    Hooks.call("dnd5e.renderNPCStatBlock", this.parent, section.children, config, options);
+
     return section.children;
   }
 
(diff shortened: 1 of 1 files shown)


module/data/actor/npc.mjs
```diff
@@ -410,6 +410,18 @@ export default class NPCData extends CreatureTemplate {
...
+    /**
+     * A hook event that fires after an embedded NPC stat block rendered.
+     * @function dnd5e.renderNPCStatBlock
+     * @memberof hookEvents
+     * @param {Actor5e} actor                   NPC being embedded.
+     * @param {HTMLCollection} contents         HTML elements to be inserted.
+     * @param {DocumentHTMLEmbedConfig} config  Configuration for embedding behavior.
+     * @param {EnrichmentOptions} options       Original enrichment options.
+     */
+    Hooks.call("dnd5e.renderNPCStatBlock", this.parent, section.children, config, options);
```
Fyorl, line 423: Is passing `section.children` useful? You can manipulate the elements and delete them, but you can't insert any I don't think. Most render hooks give you the root element, so it's probably fine to do that here too.

arbron: My one concern is that the root element isn't actually inserted as part of the embed, so it might cause people to try and add classes whatnot to that element which wouldn't do anything.

Fyorl: Should we create a `template` element instead maybe? That should hopefully not imply that setting classes on it is meaningful. Alternatively we could convert it to an array before sending it to the hook, allow the hook to mutate it, and then call `section.replaceChildren`, but that's extra performance cost when there's no guarantee a hook will even be called.

I just thought that being able to insert elements is something an API consumer may want to do, but I don't feel strongly so I've marked it as approved and you can go with whatever you feel is best.

Fyorl (approved):

module/data/actor/npc.mjs
arbron: Yeah, it looks like the enrichment system is fine accepting a `DocumentFragment`, so I can just use a template and return `template.contents`.

PR #4853, opened 2024-12-10
[#3958] Add formatters for distance and weight units

arbron:
Adds two new utility methods, `formatDistance` and `formatWeight` that perform proper localized formatting of those two unit types. By default these use the standard javascript unit localization, but the config object can be used to override this or provide custom localization for units that javascript doesn't support (such as tons).

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2346,16 +2346,16 @@ preLocalize("movementTypes", { sort: true });
 
 /**
  * @typedef {object} UnitConfiguration
- * @property {string} label             Localized label for the unit.
- * @property {string} abbreviation      Localized abbreviation for the unit.
- * @property {number} conversion        Multiplier used to convert between various units.
- * @property {string} [counted]         Localization path for counted plural forms in various unit display modes.
- *                                      Only necessary if non-supported unit or using a non-standard name for a
- *                                      supported unit.
- * @property {string} [formattingUnit]  Unit formatting value as supported by javascript's internationalization system:
- *                                      https://tc39.es/ecma402/#table-sanctioned-single-unit-identifiers. Only
- *                                      required if the formatting name doesn't match the unit key.
- * @property {string} type              Whether this is an "imperial" or "metric" unit.
+ * @property {string} label              Localized label for the unit.
+ * @property {string} abbreviation       Localized abbreviation for the unit.
+ * @property {number} conversion         Multiplier used to convert between various units.
+ * @property {string} [counted]          Localization path for counted plural forms in various unit display modes.
+ *                                       Only necessary if non-supported unit or using a non-standard name for a
+ *                                       supported unit.
+ * @property {string} [formattingUnit]   Unit formatting value as supported by javascript's internationalization system:
+ *                                       https://tc39.es/ecma402/#table-sanctioned-single-unit-identifiers. Only
+ *                                       required if the formatting name doesn't match the unit key.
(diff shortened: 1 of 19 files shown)


module/config.mjs
```diff
@@ -2344,18 +2344,57 @@ preLocalize("movementTypes", { sort: true });
...
+ * @property {string} label             Localized label for the unit.
+ * @property {string} abbreviation      Localized abbreviation for the unit.
+ * @property {number} conversion        Multiplier used to convert between various units.
+ * @property {string} [counted]         Localization path for counted plural forms in various unit display modes.
+ *                                      Only necessary if non-supported unit or using a non-standard name for a
+ *                                      supported unit.
+ * @property {string} [formattingUnit]  Unit formatting value as supported by javascript's internationalization system:
+ *                                      https://tc39.es/ecma402/#table-sanctioned-single-unit-identifiers. Only
+ *                                      required if the formatting name doesn't match the unit key.
+ * @property {string} type              Whether this is an "imperial" or "metric" unit.
```
Fyorl, line 2358:
```suggestion
 * @property {"imperial"|"metric"} type              Whether this is an "imperial" or "metric" unit.
```

module/config.mjs
```diff
@@ -2344,18 +2344,57 @@ preLocalize("movementTypes", { sort: true });
...
+    label: "DND5E.UNITS.DISTANCE.Mile.Label",
+    abbreviation: "DND5E.UNITS.DISTANCE.Mile.Abbreviation",
+    conversion: 5_280,
+    formattingUnit: "mile",
+    type: "imperial"
+  },
+  m: {
+    label: "DND5E.UNITS.DISTANCE.Meter.Label",
+    abbreviation: "DND5E.UNITS.DISTANCE.Meter.Abbreviation",
+    conversion: 3.281,
```
Fyorl, line 2384: Reasonably sure that D&D localisations use a simplified imperial to metric conversion for feet at least, where they convert 5 ft to 1.5m. That would put this conversion at 3⅓. From some quick REPL tests, it looks like we get round numbers when converting 5ft increments, even with a recurring decimal like this, so maybe we can use it.

```suggestion
    conversion: 3 + (1/3), // D&D uses a simplified 5ft -> 1.5m conversion.
```

module/config.mjs
```diff
@@ -2344,18 +2344,57 @@ preLocalize("movementTypes", { sort: true });
...
+    formattingUnit: "meter",
+    type: "metric"
+  },
+  km: {
+    label: "DND5E.UNITS.DISTANCE.Kilometer.Label",
+    abbreviation: "DND5E.UNITS.DISTANCE.Kilometer.Abbreviation",
+    conversion: 3_281,
```
Fyorl, line 2391: There might also be a simplified conversion for this but I don't know it off the top of my head. I will ask on discord.

Fyorl (changes requested):

module/config.mjs
arbron: Well, if we do a 3,333 conversion, that will maintain the 1 km -> 1,000 m conversion and be the same relation of ft -> m and mi -> km.

module/config.mjs
krbz999: This is correct. 3ft -> 1.5m

module/config.mjs
Fyorl: That sounds right, yeah

Fyorl (approved):

PR #4855, opened 2024-12-10
[#4790] Death save bonus and config

### LINKED ISSUE #4790: Feature Request: Allow Death Saving Throws to have added bonuses like other saving throws.
Would be appreciated if death saving throws could get the same treatment as "system.abilities.[abbreviation].bonuses.save" via active effects and other means of granting bonuses to it. It technically is a saving throw and having a way to apply bonuses / penalties to it would be nice.

krbz999:
Closes #4790.

Character sheet in edit mode:

image

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -44,6 +44,8 @@ const {
  * @property {string} attributes.hp.bonuses.level         Bonus formula applied for each class level.
  * @property {string} attributes.hp.bonuses.overall       Bonus formula applied to total HP.
  * @property {object} attributes.death
+ * @property {object} attributes.death.bonuses
+ * @property {string} attributes.death.bonuses.save       Numeric or dice bonus to death saving throws.
  * @property {number} attributes.death.success            Number of successful death saves.
  * @property {number} attributes.death.failure            Number of failed death saves.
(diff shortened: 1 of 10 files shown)


module/data/actor/character.mjs
```diff
@@ -118,6 +118,9 @@ export default class CharacterData extends CreatureTemplate {
           }),
           failure: new NumberField({
             required: true, nullable: false, integer: true, min: 0, initial: 0, label: "DND5E.DeathSaveFailures"
+          }),
+          bonuses: new SchemaField({
```
arbron, line 122: Do we foresee other bonuses being added? If not we can probably just go with `bonus` (similar to `attributes.init.bonus`).

Also make sure to add this to the JSDoc above the class.

arbron (changes requested):

module/data/actor/character.mjs
krbz999: I was of the opinion that `init` was the outlier and that the general schema should be `bonuses` with `.check`, `.save` or both. The same is done in `attributes.concentration`.

arbron (approved):

arbron (approved):

Fyorl (approved):

PR #4857, opened 2024-12-11
[#3214, #4738] Add combat-related usage recovery periods

### LINKED ISSUE #3214: Expand Limited Uses field options

### LINKED ISSUE #4738: Add "initiative" and "combat start/end" recovery types
In addition to rests and dawn/day/dusk, this would be for items that recover uses when

- an actor rolls initiative (a lot of items have notes such as "you regain x uses when you roll initiative if you have no uses left"). I was sure there was an issue for this already but "initiative" didn't turn up anything.
- items or activities that can be used once per encounter (e.g. "Villain Actions" in *Flee, Mortals* / *Where Evil Lives*).

arbron:
Introduces five new uses recover periods relating to combat:
- `encounter`: Recovered at the start of combat
- `turnStart` & `turnEnd`: Recovered once per round at the start or end of a specific combatant's turn
- `turn`: Recovered at the start of all combatant's turns
- `initiative`: Recovered when an actor rolls initiative

First first four (but not `initiative`) are special "combat" recovery types, which changes their behavior outside combat in two ways:

1. They are automatically recovered at the end of combat
2. They will not be consumed by default if combat isn't occuring

The second point means that if an activity is consuming item or activity uses and the only recovery periods are combat periods, then the consumption button will be unchecked by default.

Closes #3214
Closes #4738

### DIFF
--- a/module/data/activity/fields/consumption-targets-field.mjs
+++ b/module/data/activity/fields/consumption-targets-field.mjs
@@ -69,7 +69,7 @@ export class ConsumptionTargetData extends foundry.abstract.DataModel {
    */
   get combatOnly() {
     let recovery;
-    switch (this.type) {
+    switch ( this.type ) {
       case "activityUses":
         recovery = this.activity.uses.recovery;
         break;
--- a/module/data/activity/fields/consumption-targets-field.mjs
+++ b/module/data/activity/fields/consumption-targets-field.mjs
@@ -63,6 +63,27 @@ export class ConsumptionTargetData extends foundry.abstract.DataModel {
   /* -------------------------------------------- */
 
   /**
+   * Should this consumption only be performed during initiative? This will return `true` if consuming activity or item
+   * uses and those uses only recover on "combat" periods.
+   * @type {boolean}
+   */
+  get combatOnly() {
+    let recovery;
+    switch (this.type) {
+      case "activityUses":
+        recovery = this.activity.uses.recovery;
+        break;
+      case "itemUses":
+        recovery = (this.target ? this.actor?.items.get(this.target) : this.item)?.system.uses.recovery;
+        break;
+      default: return false;
+    }
+    return recovery?.every(r => CONFIG.DND5E.limitedUsePeriods[r.period]?.type === "combat") ?? false;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
    * Item to which this consumption target's activity belongs.
    * @type {Item5e}
    */
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -850,7 +850,7 @@ export default function ActivityMixin(Base) {
      */
     _requiresConfigurationDialog(config) {
       const checkObject = obj => (foundry.utils.getType(obj) === "Object")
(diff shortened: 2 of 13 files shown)


module/data/activity/fields/consumption-targets-field.mjs
```diff
@@ -62,6 +62,27 @@ export class ConsumptionTargetData extends foundry.abstract.DataModel {
...
   /* -------------------------------------------- */
 
+  /**
+   * Should this consumption only be performed during initiative? This will return `true` if consuming activity or item
+   * uses and those uses only recover on "combat" periods.
+   * @type {boolean}
+   */
+  get combatOnly() {
+    let recovery;
+    switch (this.type) {
```
Fyorl, line 72:
```suggestion
    switch ( this.type ) {
```

module/documents/activity/mixin.mjs
```diff
@@ -847,7 +849,8 @@ export default function ActivityMixin(Base) {
      * @protected
      */
     _requiresConfigurationDialog(config) {
-      const checkObject = obj => (foundry.utils.getType(obj) === "Object") && Object.values(obj).some(v => v);
+      const checkObject = obj => (foundry.utils.getType(obj) === "Object")
+        && Object.values(obj).some(v => v === true || v?.length > 0);
```
Fyorl, line 853:
```suggestion
        && Object.values(obj).some(v => (v === true) || v?.length);
```

Fyorl (changes requested):
Code looks good but I'd like to better understand the difference between `encounter` and `initiative` recovery periods. I understand that one is a 'combat' type and the other isn't, so the former wouldn't be consumed if used outside of initiative, but I can't think of any such features where that is the case.

As far as I can tell, at least in official content, anything that is intended to recover 'per encounter' simply recovers whenever initiative is rolled. They seem to be functionally identical.

Fyorl:
> The on initiative type was to similar PC features that recharge when you roll initiative, while the encounter type was designed more for NPC features (like Zhell's examples in "Flee, Mortals" and elsewhere) that are marked as once per encounter. They are fairly similar, so I suppose we could drop the encounter type, but I felt the difference in consumption behavior was sufficient to warrant two types.

Do you have an example of such a feature though? Are villain actions intended to be used infinitely by a creature outside of combat? I'm trying to understand the practical different between something that recovers 'per encounter' vs. 'when initiative is rolled'. Official content prefers the latter, so I think we should use that unless there's some meaningful mechanical difference.

arbron:
> Do you have an example of such a feature though? Are villain actions intended to be used infinitely by a creature outside of combat? I'm trying to understand the practical different between something that recovers 'per encounter' vs. 'when initiative is rolled'. Official content prefers the latter, so I think we should use that unless there's some meaningful mechanical difference.

I'll just pull it out for now and we can always add it back later if further argument shows it is useful.

Fyorl (approved):

krbz999:
> I'll just pull it out for now and we can always add it back later if further argument shows it is useful.

Unless a user somehow bypasses initiative rolls, it will likely end up the same, yeah.

PR #4867, opened 2024-12-12
[#4856] Show temporary spell slots

### LINKED ISSUE #4856: Display "temporary" slots when number of available slots exceeds maximum
Sorcerers are able to turn sorcery points into spell slots, allowing them to create more spells slots of a certain level than they would normally have. This currently works in the system and is displayed properly in the activity usage dialog when casting a spell, but the pips in the actor's spells tab don't display more than their maximum, giving no indication during normal usage that the actor possesses these extra slots. It would be nice to display these on that tab, perhaps using blue markers with a similar color to temp HP to make it clear that they are temporary spell slots.

roth-michael:
Closes #4856 
Adds temporary (value exceeding max) spell slots to the right of "normal' spell slots, colored in to match the temphp color.
Screenshot 2024-12-12 115140


less/v2/inventory.less
```diff
@@ -183,6 +183,11 @@
         box-shadow: none;
         filter: drop-shadow(0 0 1px var(--dnd5e-color-black));
 
+        &.tmp.filled { 
+          --icon-fill: var(--dnd5e-color-blue);
+          --icon-stroke: var(--dnd5e-color-blue)
+        }
+        &.tmp:hover { filter: drop-shadow(0 0 2px var(--dnd5e-color-blue)); }
```
Fyorl, line 190: Slight preference for using purple here. That's loosely what the design uses as the 'magic' colour, like in the Spellcasting header. So either `--dnd5e-color-sc-1` or `--dnd5e-color-sc-2` (probably `--dnd5e-color-sc-2`).

Fyorl (changes requested):

less/v2/inventory.less
roth-michael: image
Yep, looks good

Fyorl (approved):

PR #4868, opened 2024-12-12
[#4866] Display nicer text for a dc-less passive skill enricher

### LINKED ISSUE #4866: Allow for the DC of a passive skill enricher to be optional
Use case being the ability to make an enricher for text such as "the party member with the highest passive Perception". When using an enricher without a DC (`[[/skill perception passive]]`), it displays as "DC undefined passive Wisdom (Perception)".

It would be nice to be able to use an enricher to simply display a passive skill without a need for a DC.

roth-michael:
Closes #4866

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -956,7 +956,7 @@ export function createRollLabel(config) {
         label = ability;
       }
       if ( config.passive ) {
-        label = game.i18n.format(`EDITOR.DND5E.Inline.DCPassive${longSuffix}`, { dc: config.dc, check: label });
+        label = game.i18n.format(`EDITOR.DND5E.Inline.${config.dc ? "DC" : ""}Passive${longSuffix}`, { dc: config.dc, check: label });
       } else {
         if ( showDC ) label = game.i18n.format("EDITOR.DND5E.Inline.DC", { dc: config.dc, check: label });
         label = game.i18n.format(`EDITOR.DND5E.Inline.Check${longSuffix}`, { check: label });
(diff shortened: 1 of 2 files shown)


arbron:
Throw up a comment on the original issue so I can assign it to you

krbz999:
If we added a `highest` property, we could support the use case in the issue explicitly, highlighting only those with the highest passive skill as a success.

roth-michael:
Would be fairly trivial - I've got some local code for it now (for passive checks, at least). Would we want to change the enricher text for that so that "highest" was included in it, or just leave it?

krbz999:
> (Also: What should happen if both a `dc` and `highest` are provided - my gut says `dc` takes precedence and `highest` is ignored)

I suppose it would be the highest that also passes the DC but imma let Arby sort out if this is a wanted feature.

arbron (approved):

arbron:
You can add `highest` as a different PR if you still want.

module/enrichers.mjs
```diff
@@ -956,7 +956,7 @@ export function createRollLabel(config) {
         label = ability;
       }
       if ( config.passive ) {
-        label = game.i18n.format(`EDITOR.DND5E.Inline.DCPassive${longSuffix}`, { dc: config.dc, check: label });
+        label = game.i18n.format(`EDITOR.DND5E.Inline.${config.dc ? "DC" : ""}Passive${longSuffix}`, { dc: config.dc, check: label });
```
arbron, line 959: You should use `showDC` here, since it takes the `hideDC` option into account:

```suggestion
        label = game.i18n.format(
          `EDITOR.DND5E.Inline.${showDC ? "DC" : ""}Passive${longSuffix}`, { dc: config.dc, check: label }
        );
```

arbron (changes requested):

arbron (approved):

Issue #4799, opened 2024-11-28
Add GM-only chat message on monster's turns with button to roll recharges for monster abilities


arbron:
Currently it is easy to forget to roll to recharge a monster's ability on their turn, so it would be nice to provide a reminder to GMs at the start of a monster's turn what abilities they have that can be potentially recharged.

PR #4871, opened 2024-12-13
[#2359, #4799] Add new turn chat message

arbron:
Introduces a new chat message type for communicating combat recovery information and offering combat-related controls. At this point the message just displays a summary of recovery changes that occured at the start of combat, the start of a turn, or the end of a turn.

Turn Change Chat Message

This introduces a new `turn` chat message type which stores information on the relevant combatant, what kind of combat change event triggered the message, and deltas for any changes made to the actor or their items.

### DIFF
--- a/module/data/abstract/chat-message-data-model.mjs
+++ b/module/data/abstract/chat-message-data-model.mjs
@@ -1,0 +1,122 @@
+/**
+ * @typedef {object} ChatMessageDataModelMetadata
+ * @property {Record<string, ApplicationClickAction>} actions  Default click actions for buttons on the message.
+ * @property {string} template                                 Template to use when rendering this message.
+ */
+
+/**
+ * Abstract base class to add some shared functionality to all of the system's custom chat message types.
+ * @abstract
+ */
+export default class ChatMessageDataModel extends foundry.abstract.DataModel {
+
+  /**
+   * Metadata for this chat message type.
+   * @type {ChatMessageDataModelMetadata}
+   */
+  static metadata = Object.freeze({
+    actions: {},
+    template: ""
+  });
+
+  get metadata() {
+    return this.constructor.metadata;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Template to use when rendering this message.
+   * @type {string}
+   */
+  get template() {
+    return this.metadata.template;
+  }
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /**
+   * Perform any changes to the chat message's element before displaying in the list.
+   * @param {HTMLElement} element  Element representing the entire chat message.
+   */
+  async getHTML(element) {
+    const rendered = await this.render();
+    if ( rendered ) element.querySelector(".message-content").innerHTML = rendered;
+    this.parent._enrichChatCard(element);
+
+    const click = this.#onClick.bind(this);
+    element.addEventListener("click", click);
+    element.addEventListener("contextmenu", click);
+    this._onRender(element);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Render the contents of this chat message.
+   * @returns {string}
+   */
+  async render() {
+    if ( !this.template ) return "";
+    return renderTemplate(this.template, await this._prepareContext());
+  }
+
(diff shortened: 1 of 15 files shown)


module/documents/combatant.mjs
```diff
@@ -1,7 +1,58 @@
...
+     * @function dnd5e.preCreateCombatMessage
+     * @memberof hookEvents
+     * @param {Combatant5e} combatant         Combatant for which the message will be created.
+     * @param {object} messageConfig
+     * @param {boolean} messageConfig.create  Should the chat message be posted?
+     * @param {object} messageConfig.data     Data for the created chat message.
+     */
+    Hooks.callAll("dnd5e.preCreateCombatMessage", this, messageConfig);
+
+    if ( messageConfig.create ) return ChatMessage.implementation.create(messageConfig.data);
```
Fyorl, line 51: This looks a little unconventional with the use of `callAll` and `messageConfig.create` vs. `call` with a return of `false` cancelling the `preCreate`.

templates/chat/turn-card.hbs
```diff
@@ -0,0 +1,26 @@
+<div class="dnd5e2 chat-card turn-card">
+    {{#if actor}}
+
+    {{!-- Deltas --}}
+    {{#if deltas.length}}
+    <section class="deltas">
+        <h5 class="roboto-condensed-upper">{{ localize "DND5E.CHATMESSAGE.TURN.Recovery" }}</h5>
```
Fyorl, line 7: Prefer a `<strong>` rather than a heading element since it's not really clear to me where such a heading exists in the page's hierarchy.

Fyorl (approved):

module/data/chat-message/_module.mjs
```diff
@@ -0,0 +1,10 @@
+import TurnMessageData from "./turn-message-data.mjs";
+
+export {
+  TurnMessageData
+};
+export {default as ActorDeltasField, IndividualDeltaField} from "./fields/deltas-field.mjs";
```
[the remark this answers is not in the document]
arbron: It does break from the convention we've used in `dnd5e.dataModels.activity` and `dnd5e.dataModels.item` for a while now. If we did that I'd probably go with `dnd5e.dataModels.chatMessage.fields.ActorDeltasField`.

module/data/abstract/chat-message-data-model.mjs
```diff
@@ -0,0 +1,122 @@
...
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /**
+   * Perform any changes to the chat message's element before displaying in the list.
+   * @param {HTMLElement} element  Element representing the entire chat message.
+   */
+  async getHTML(element) {
```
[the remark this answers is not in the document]
arbron: Maybe I'll adjust the naming a bit to step away from the old `getHTML` method and name it more like `renderMessageContent`. Then in V13 `renderHTML` can be focused on rendering the entire message and this method can be focused on handling just the content.

module/documents/combatant.mjs
arbron: The reason I took this approach is because there are situations where the chat message might normally not be created (when there isn't any delta data), but a hook might want to create it anyway. So if there isn't delta data the hook will be passed `{ create: false }` and then can choose to set it to `true`. I could use `Hooks.call` and provide that as another method of canceling.

templates/chat/turn-card.hbs
arbron: I was working off the message header which is an `<h4>`.

module/data/abstract/chat-message-data-model.mjs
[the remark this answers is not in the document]
arbron: I'll leave it as `getHTML` for now.

module/documents/combatant.mjs
Fyorl: No objections from me; just wanted to note that it looked a little strange in case it wasn't intentional.

templates/chat/turn-card.hbs
Fyorl: I think I ended up not changing that in v13 due to backwards compatibility concerns, but the message header should also probably be a `<strong>`.

PR #4875, opened 2024-12-13
[#1982] Add support for ranged communication like telepathy

### LINKED ISSUE #1982: Add telepathy range to language data
Many stat blocks include telepathy out to a certain range. It would be useful if telepathy could be a proper data field in languages rather than requiring the use of the custom field.

```
"traits": {
  "languages": {
    "value": [],
    "telepathy": {
      "value": 30,
      "units": "ft"
    }
    "custom": ""
  }
}
```

arbron:
Adds `communication` data to the languages data model that tracks distances and units for features like telepathy.

Ranged Communication

Also includes a large refactor of how trait data models are constructed to match the custom field pattern the system has adopted elsewhere.

Closes #1982


krbz999:
Should it not display as "Telepathy | 120" rather than "Telepathy 120 ft" to match senses?

module/data/actor/character.mjs
```diff
@@ -14,6 +15,10 @@ const {
   ArrayField, BooleanField, HTMLField, IntegerSortField, NumberField, SchemaField, SetField, StringField
 } = foundry.data.fields;
 
+/**
+ * @typedef {import("./fields/simple-trait.mjs").SimpleTraitData} SimpleTraitData
```
Fyorl, line 19: At least with typedoc, we can use a more ergonomic invocation. Might depend on your IDE whether it works for intellisense or not.

```suggestion
 * @import { SimpleTraitData } from "./fields/simple-trait.mjs";
```

module/data/actor/npc.mjs
```diff
@@ -501,7 +499,13 @@ export default class NPCData extends CreatureTemplate {
...
 
         // Languages (e.g. `Common, Draconic`)
-        languages: prepareTrait(this.traits.languages, "languages") || game.i18n.localize("None"),
+        languages: [
+          prepareTrait(this.traits.languages, "languages"),
+          ...Object.entries(CONFIG.DND5E.communicationTypes).map(([key, { label }]) => {
+            const data = this.traits.languages.communication[key];
+            return data?.value ? `${label} ${formatDistance(data.value, data.units)}`.toLowerCase() : null;
+          }).filter(_ => _)
+        ].join("; ") || game.i18n.localize("None"),
```
Fyorl, line 508:
```suggestion
          })
        ].filterJoin("; ") || game.i18n.localize("None"),
```

module/data/actor/templates/traits.mjs
```diff
@@ -1,36 +1,39 @@
...
+import DamageTraitField from "../fields/damage-trait-field.mjs";
+import SimpleTraitField from "../fields/simple-trait-field.mjs";
 
-const { SchemaField, SetField, StringField } = foundry.data.fields;
+const { NumberField, SchemaField, SetField, StringField } = foundry.data.fields;
 
 /**
- * Shared contents of the traits schema between various actor types.
+ * @typedef {import("../fields/damage-trait.mjs").DamageTraitData} DamageTraitData
+ * @typedef {import("../fields/simple-trait.mjs").SimpleTraitData} SimpleTraitData
```
Fyorl, line 11:
```suggestion
 * @import { DamageTraitData } from "../fields/damage-trait.mjs";
 * @import { SimpleTraitData } "../fields/simple-trait.mjs";
```

module/applications/actor/sheet-v2-mixin.mjs
```diff
@@ -260,6 +260,14 @@ export default function ActorSheetV2Mixin(Base) {
         if ( values.length ) traits.dm = values;
       }
 
+      // Display ranged communication
+      for ( const [key, { label }] of Object.entries(CONFIG.DND5E.communicationTypes) ) {
+        const data = this.actor.system.traits?.languages?.communication?.[key];
+        if ( !data?.value ) return;
```
Fyorl, line 266:
```suggestion
        if ( !data?.value ) continue;
```

Fyorl (changes requested):

module/data/actor/character.mjs
arbron: Our linter doesn't like `@import`, but I can add an exception for that. I was following core's styling for these, but I see you switched things up in V13 :).

Fyorl: I think it was a relatively recent addition to typedoc

templates/actors/npc-sheet-2.hbs
```diff
@@ -384,32 +384,8 @@
             <div class="sidebar">
 
                 {{!-- Speed --}}
-                <div class="pills-group {{#unless speed}}empty{{/unless}}">
```
Fyorl, line 387: Good simplification. We can probably apply it to the actor sheet as well.

Fyorl (approved):

PR #4879, opened 2024-12-14
[#4876] Add support for modifier keys during drag operations

### LINKED ISSUE #4876: Add keyboard shortcuts that control whether a drag/drop operation is move or copy
Currently the drag & drop behavior in the system varies depending on where you are dragging to and from:

Actor Inventory -> Another Actor: Copy
Actor Inventory -> Container on Actor: Move
Sidebar -> Actor Inventory: Copy
Sidebar -> Container in Sidebar: Move
Compendium -> Anywhere Else: Copy
Compendium -> Container in Compendium: Move

While these default behaviors should be retained, it would be very nice to be able to toggle to the opposite mode when dragging. This would involve holding down a certain keyboard modifier much like you would in a file browser on your computer. This should also involve changing the drag cursor to make it clear whether a drag will result in a copy or a move operation.

arbron:
Adds the ability to use the OS-defined modifier key (usually Alt on Windows and Option on Mac) to toggle between the default drop behavior and the opposite behavior. So if dragging within the same actor this changes from the default move behaior to copy behavior and the opposite when dragging between different actors or to the sidebar.

Enabling this required access to the current drag payload during the `ondragover` event, so this extends the `DragDrop` class provided by core to store that information during the `ondragstart` event and adds a new handler for the `ondragend` event to clear the stored payload.

Currently this only covers dragging items, with some minor improvements to dragging favorites on the character sheet. This framework could be expanded in the future to support dragging actors into the bastion tab as well as dragging active effects, advancements, or activities.

Closes #4876

### DIFF
--- a/module/applications/mixins/drag-drop-mixin.mjs
+++ b/module/applications/mixins/drag-drop-mixin.mjs
@@ -1,3 +1,5 @@
+import { areKeysPressed } from "../../utils.mjs";
+
 /**
  * @typedef {import("../../drag-drop.mjs").DropEffectValue} DropEffectValue
  */
@@ -32,13 +34,9 @@ export default function DragDropApplicationMixin(Base) {
       let behavior = event.dataTransfer.dropEffect;
 
       if ( event.type === "dragover" ) {
-        behavior = this._defaultDropBehavior(event, data);
-
-        // An initial `dropEffect` of `copy` indicates that a modifier key is held down
-        if ( (event.type === "dragover") && (event.dataTransfer.dropEffect === "copy") ) {
-          if ( (behavior === "copy") && allowed.has("move") ) return "move";
-          else if ( (behavior === "move") && allowed.has("copy") ) return "copy";
-        }
+        if ( areKeysPressed(event, "dragMove") ) behavior = "move";
+        else if ( areKeysPressed(event, "dragCopy") ) behavior = "copy";
+        else behavior = this._defaultDropBehavior(event, data);
       }
 
       if ( (behavior !== "none") && !allowed.has(behavior) ) return allowed.first() ?? "none";
--- a/module/applications/mixins/drag-drop-mixin.mjs
+++ b/module/applications/mixins/drag-drop-mixin.mjs
@@ -1,7 +1,7 @@
 import { areKeysPressed } from "../../utils.mjs";
 
 /**
- * @typedef {import("../../drag-drop.mjs").DropEffectValue} DropEffectValue
+ * @import { DropEffectValue } from "../../drag-drop.mjs"
  */
 
 /**
@@ -28,7 +28,7 @@ export default function DragDropApplicationMixin(Base) {
      */
     _dropBehavior(event, data) {
       const allowed = this._allowedDropBehaviors(event, data);
-      let behavior = DragDrop.dropEffect ?? event.dataTransfer.dropEffect;
+      let behavior = DragDrop.dropEffect ?? event.dataTransfer?.dropEffect;
 
       if ( event.type === "dragover" ) {
         if ( areKeysPressed(event, "dragMove") ) behavior = "move";
--- a/module/applications/mixins/drag-drop-mixin.mjs
+++ b/module/applications/mixins/drag-drop-mixin.mjs
@@ -14,11 +14,8 @@ export default function DragDropApplicationMixin(Base) {
     /** @override */
     _onDragOver(event) {
       const data = DragDrop.getPayload(event);
-      if ( foundry.utils.getType(data) === "Object" ) {
-        event.dataTransfer.dropEffect = this._dropBehavior(event, data);
-      } else {
-        event.dataTransfer.dropEffect = "copy";
-      }
+      DragDrop.dropEffect = event.dataTransfer.dropEffect = (foundry.utils.getType(data) === "Object")
+        ? this._dropBehavior(event, data) : "copy";
     }
 
     /* -------------------------------------------- */
@@ -31,7 +28,7 @@ export default function DragDropApplicationMixin(Base) {
      */
     _dropBehavior(event, data) {
       const allowed = this._allowedDropBehaviors(event, data);
-      let behavior = event.dataTransfer.dropEffect;
+      let behavior = DragDrop.dropEffect ?? event.dataTransfer.dropEffect;
 
       if ( event.type === "dragover" ) {
         if ( areKeysPressed(event, "dragMove") ) behavior = "move";
--- a/module/applications/mixins/drag-drop-mixin.mjs
+++ b/module/applications/mixins/drag-drop-mixin.mjs
@@ -1,0 +1,74 @@
+/**
+ * @typedef {import("../../drag-drop.mjs").DropEffectValue} DropEffectValue
+ */
+
+/**
+ * Adds drop behavior functionality to all sheets.
+ * @param {typeof Application} Base  The base class being mixed.
+ * @returns {typeof DragDropApplication}
+ */
+export default function DragDropApplicationMixin(Base) {
+  return class DragDropApplication extends Base {
+    /** @override */
+    _onDragOver(event) {
+      const data = DragDrop.getPayload(event);
+      if ( foundry.utils.getType(data) === "Object" ) {
+        event.dataTransfer.dropEffect = this._dropBehavior(event, data);
+      } else {
+        event.dataTransfer.dropEffect = "copy";
+      }
+    }
+
+    /* -------------------------------------------- */
+
+    /**
+     * The behavior for the dropped data.
+     * @param {DragEvent} event  The drag event.
+     * @param {object} data      The drag payload.
+     * @returns {DropEffectValue}
+     */
+    _dropBehavior(event, data) {
+      const allowed = this._allowedDropBehaviors(event, data);
+      let behavior = event.dataTransfer.dropEffect;
+
+      if ( event.type === "dragover" ) {
+        behavior = this._defaultDropBehavior(event, data);
+
+        // An initial `dropEffect` of `copy` indicates that a modifier key is held down
+        if ( (event.type === "dragover") && (event.dataTransfer.dropEffect === "copy") ) {
+          if ( (behavior === "copy") && allowed.has("move") ) return "move";
+          else if ( (behavior === "move") && allowed.has("copy") ) return "copy";
+        }
+      }
+
+      if ( (behavior !== "none") && !allowed.has(behavior) ) return allowed.first() ?? "none";
+      return behavior || "copy";
+    }
+
+    /* -------------------------------------------- */
+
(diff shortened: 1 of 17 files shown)


Fyorl (commented):
Just some high-level comments before I do a full review: I think it would be preferable to have fixed modifier keys that map to a specific action, rather than a key that toggles to the opposite of the default action. For example, holding Ctrl should always copy, and holding Shift should always move (doesn't have to be those exact modifiers), regardless of what the default action is.

Fyorl:
> Hmm, the thing I like about this approach is it takes advantage of the keys people already use with their operating system, so nothing new to learn (it also keeps the cursor updated so it is always clear what action will be performed).

Not sure what you mean. Ctrl is copy and Shift is move in Windows. Alt is create link. So holding Alt to switch between move or copy depending on what the default action is would be new behaviour.

unsoluble:
Option-drag is always "duplicate" on macOS; there are no other OS-wide drag modifiers here. For what it's worth :)

arbron:
Okay, I've done a bit more research and here is what is provided by OSes:

#### Windows
- Ctrl — Copy file
- Shift — Move file
- Alt — Create shortcut

#### Mac
- Option — Copy file
- Command — Move file
- Option + Command — Create alias

So we could setting on something that works for both:
- Ctrl or Option (aka Alt) - Copy
- Shift or Command (aka Meta) - Move

unsoluble:
Ehhh I'm in full agreement on Option/Alt means copy, but Command doesn't actually have any relation to "move" on macOS, and is already overloaded as the Control analogue in Foundry.

Also this is all confused by the fact that drag moves are _already_ Copy events by default, so I'm not really sure what the imagined workflow is here.

arbron:
> Ehhh I'm in full agreement on Option/Alt means copy, but Command doesn't actually have any relation to "move" on macOS, and is already overloaded as the Control analogue in Foundry.

Command changes a copy-as-default drag to a move drag when moving between **different drives** on Mac OS. This is similar to Windows, where any drag-drop that is considered "free" (aka doesn't move between drives or computers) is move-by-default, while any drag-drop that is considered expensive is copy-by-default.

> Also this is all confused by the fact that drag moves are _already_ Copy events by default, so I'm not really sure what the imagined workflow is here.

Drags are copy by default if moving from one actor to another, between the sidebar and an actor, and to or from a compendium. Drags are move by default if moving within an actor or moving into or out of a container on the same actor/sidebar/compendium. This would offer a way to change the behavior in all of those locations using modifier keys.

unsoluble:
> Command changes a copy-as-default drag to a move drag when moving between different drives on Mac OS.

Sure, but that's just a _Finder_ shortcut; isn't used anywhere else, or expected behaviour in Mac apps.

I do get where you're going with this, that we need a way to "do the other thing". But the challenge here is the tension between Kim's (reasonable) desire to have each modifier to a specific action, versus the fact that there are currently different behaviours in different contexts.

My two cents on this would be: The Alt/Option key is generally considered a "do an alternate behaviour" modifier, so would make intuitive sense to have it reverse whatever the default behaviour of a given drag is.

Fyorl:
> But the challenge here is the tension between Kim's (reasonable) desire to have each modifier do a specific action, versus the fact that there are currently different behaviours in different contexts.

There are different behaviours in different contexts, and the proposal is that you may hold a modifier to *force* a specific behaviour that you intend that overrides the default behaviour.

If you want to copy, you hold Ctrl, regardless of whether the default behaviour was a move or a copy. If you want to move, you hold Shift. You don't need to know what the default behaviour is beforehand, or take care to watch your cursor icon to see what the result of your modifier + context combo is. You only need to learn that you hold Ctrl if you intend to copy and hold Shift if you intend to move.

unsoluble:
Yeah I can get behind that :)

Worth making sure this isn't going to collide with potential core modifiers — I know it's been suggested that Control-drag might be used to make token drags into teleports, for example, but perhaps in the future it might be nice to be able to modifier-drag tokens to create duplicates, so using the same key globally for that concept would be best.

(also isn't this really a core-level concept regardless?)

module/applications/mixins/drag-drop-mixin.mjs
```diff
@@ -0,0 +1,72 @@
...
+export default function DragDropApplicationMixin(Base) {
+  return class DragDropApplication extends Base {
+    /** @override */
+    _onDragOver(event) {
+      const data = DragDrop.getPayload(event);
+      if ( foundry.utils.getType(data) === "Object" ) {
+        event.dataTransfer.dropEffect = this._dropBehavior(event, data);
+      } else {
+        event.dataTransfer.dropEffect = "copy";
+      }
```
Fyorl, line 21: Per main comment:

```suggestion
      }
      
      DragDrop.dropEffect = event.dataTransfer.dropEffect;
```

module/applications/mixins/drag-drop-mixin.mjs
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * The behavior for the dropped data.
+     * @param {DragEvent} event  The drag event.
+     * @param {object} data      The drag payload.
+     * @returns {DropEffectValue}
+     */
+    _dropBehavior(event, data) {
+      const allowed = this._allowedDropBehaviors(event, data);
+      let behavior = event.dataTransfer.dropEffect;
```
Fyorl, line 34: Per main comment:

```suggestion
      let behavior = DragDrop.dropEffect;
```

module/applications/mixins/sheet-v2-mixin.mjs
```diff
@@ -277,13 +279,51 @@ export default function DocumentSheetV2Mixin(Base) {
         const content = await renderTemplate("systems/dnd5e/templates/items/parts/item-summary.hbs", context);
         summary.querySelectorAll(".item-summary").forEach(el => el.remove());
         summary.insertAdjacentHTML("beforeend", content);
-        await new Promise(resolve => requestAnimationFrame(resolve));
+        await new Promise(resolve => { requestAnimationFrame(resolve); });
```
Fyorl, line 282: Let's just remove this linting rule. I don't think it's ever helped flag potentially buggy behaviour for me.

Fyorl (commented):
I started testing this, and nothing was really working. It seemed like `dropEffect` was always `none` which meant that most workflows just bailed completely. I dug a bit deeper, and of course the issue is that browsers are seemingly just completely ignoring the spec or can't agree on how it should work.

I found [this snippet on stack overflow](https://stackoverflow.com/questions/20471273/html5-drag-and-drop-effectallowed-and-dropeffect) which seems to match my findings fairly well:

> I thought that `dropEffect` would be useful to read ondrop to see what to do, but it isn't available on Chrome, the `dropEffect` does not appear on the drop handler, e.g. trying to read the `dataTransfer.dropEffect` will say that the `dropEffect` is `none` even though you set it on dragover. Setting the `dropEffect` as noted above does influence the way the cursor is displayed.

> On Firefox, the `dropEffect` does come through on the dropzone after being set on dragover, but it does not influence the display of the mouse cursor.

Since we already have our custom `DragDrop5e`, we should use it to store the user's intended `dropEffect` as well as the pending `payload`, and then we don't have to deal with cross-browser inconsistencies at all here. I've noted some places in the comments to that effect, but I wanted to give some more in-depth context to them.

Apparently `dropEffect` should be available in `dragend` rather than `drop`, but I feel like that would overly complicate the implementation compared to the above proposal.

Fyorl:
> (also isn't this really a core-level concept regardless?)

The line of responsibility is slightly blurry. For example, the concept of containers and how they behave in the sidebar, compendia, and on Actor sheets is entirely a system concept.

But in general, yes, core could provide some baseline conventions here. We are trialling it in a system for now.

module/applications/mixins/drag-drop-mixin.mjs
```diff
@@ -0,0 +1,69 @@
+import { areKeysPressed } from "../../utils.mjs";
+
+/**
+ * @typedef {import("../../drag-drop.mjs").DropEffectValue} DropEffectValue
```
Fyorl, line 4: Flagging for new-style imports

Fyorl (changes requested):
Code looks good thanks. I did some testing with the various permutations and it seems like containers are perhaps not working correctly.

These are what I tested:

1. Sidebar item -> sidebar container
   - Always moves, never copies.

2. Sidebar container -> sidebar
   - Works as expected.

3. Sidebar container -> sidebar container
   - Always moves, never copies.

4. Actor inventory -> same Actor container
   - Always moves, never copies.

5. Actor container -> same Actor inventory
   - Works as expected.

5. Actor container -> same Actor container
   - Always moves, never copies.

6. Actor -> other Actor
   - Works as expected.

7. Actor -> other Actor container
   - Always copies, never moves.

8. Actor container -> other Actor inventory
   - Works as expected.

Fyorl (changes requested):
I pushed a rebase onto the v13 compatibility changes.

Looks like containers are working well, thank you. Some interactions with the sidebar I think were not quite working correctly. It seems we also flipped the default behaviour when transferring from an Actor's container to their own inventory. By default it used to move, and now it copies. Maybe not a big deal now that we have modifiers to explicitly indicate user intent, but it also might catch some people out, so let me know what you think.

I've included my test results below:

1. Sidebar item -> sidebar container
   - Works as expected. ✅

2. Sidebar container -> sidebar
   - Always moves, never copies. ⚠️
   - Cursor does not update when hovering over sidebar with modifier keys held. ⚠️

3. Sidebar container -> sidebar container
   - Works as expected. ✅

4. Sidebar -> Actor inventory
   - Works as expected. ✅

5. Actor inventory -> Sidebar
   - Always copies, never moves. (maybe intentional?) ⚠️
   - Cursor does not update. ⚠️

6. Actor inventory -> same Actor container
   - Works as expected. ✅

7. Actor container -> same Actor inventory
   - Works as expected. ✅
   - Default operation seems to be changed from move to copy. Maybe not intentional. ⚠️

8. Actor container -> same Actor container
   - Works as expected. ✅

9. Actor -> other Actor
   - Works as expected. ✅

10. Actor -> other Actor container
   - Works as expected. ✅

11. Actor container -> other Actor inventory
   - Works as expected. ✅

arbron:
> Looks like containers are working well, thank you. Some interactions with the sidebar I think were not quite working correctly. It seems we also flipped the default behaviour when transferring from an Actor's container to their own inventory. By default it used to move, and now it copies. Maybe not a big deal now that we have modifiers to explicitly indicate user intent, but it also might catch some people out, so let me know what you think.

Fixed the accidentally modified behavior for dragging in actors, but it looks like the sidebar issues in V13 is because the new sidebar application no longer registers the `_onDragOver` callback when setting up its `DragDrop` class, so that will probably need a core fix.

Fyorl (approved):
Works great now

Issue #4478, opened 2024-10-01
[Feature Request] Have the Active Effects' Status Conditions dropdown apply the status conditions separately rather than "embed"


Krigsmaskine:
The current situation "embeds"/makes the system consider the activeEffect as the chosen status condition(s) in the dropdown. I argue that is counter-intuitive and goes against the (assumed) design of the status conditions/effects. I will most likely use status conditions/effects interchangeably, bear with me. 

I find the whole point of status effects that they're universal, visual and recognizable from the canvas. With the current situation, the only way to find out a token with the activeEffect is "also" affected by the selected status conditions is to open that token's character sheet, if you don't recognize the activeEffect's icon as being one of those activeEffects with x, y or z added as a status condition. 

- This is not feasible for players the vast majority of the time, as they don't own the tokens' they're attacking.
- Even if it's an actor the player owns, such as their player character, everything they traditionally use in combat are on different tabs than the active effect tab/status condition tab. 
- Sometimes activeEffects from items can apply their own active effect and then also apply a status condition, but they're not necessarily interlinked. Having the status condition applied separately means it'd be easy to dismiss it from the token hud (right-click remove). 
- It's not ideal for neither GM or players to get a quick overview of what's currently happening in the scene for playing their turns. You want it to be quickly evident who is Prone, who is Paralyzed and other similar status conditions that change how the dice are rolled for people who do not own the token, too.

JPMeehan:
Worth noting that this is related to core handling — that statuses field is defined by the BaseActiveEffect schema.

Kapuzenjoe:
+1 here. 
The current solution of status conditions via the dropdown menu in active effects is very confusing if the conditions are not displayed in sync with the sheet/token conditions. You quickly lose track if you can only get this information via the active effect (where the PCs only have access to their own).
The mod DAE had a solution for this (Dependet Condition), but the mod is very buggy in DnD4.1 and I think that a uniform procedure should be implemented for setting the conditions. Or that the dropdown menu of the active effects is synchronized with the sheet/token conditions.

PR #4881, opened 2024-12-14
Add deletion dialog when multiple sources of the same status exists on the actor

krbz999:
- If more than one effect exists on the actor granting a `status`, when the user clicks the condition a prompt will appear similar to concentration.
- If the actor has an active ActiveEffect with a status, the given condition is highlighted on the token HUD regardless of it being the OG condition.

This PR also fixes errors being thrown on v13 re: exhaustion on the token HUD.

TODO:
- [ ] Rework concentration deletion dialog to use this same dialog.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2584,6 +2584,10 @@
     "DeleteEffect": "Delete Effect",
     "DissociateEffect": "Dissociate Effect",
     "EditEffect": "Edit Effect",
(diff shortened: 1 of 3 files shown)


Fyorl:
Related #5296

Fyorl:
Related #4478

Fyorl (approved):

PR #4883, opened 2024-12-15
Allow for "spellcasting" as the chosen ability in enrichers

### LINKED ISSUE #5092: Add spellcasting modifier as a /check enricher option
As the title, this would help adventures dealing with stuff such as Dispel Magic saying something is dispellable using a "DC 15 spellcasting ability check", which are not uncommon given magic locks and whatnot.

krbz999:
We had a use case for a lot of "make a saving throw with your spellcasting ability" in adventures.

image

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -4462,6 +4462,10 @@ Object.defineProperty(DND5E, "enrichmentLookup", {
       addFullKeys("abilities");
       addFullKeys("skills");
       addFullKeys("spellSchools");
+
+      _enrichmentLookup.abilities.spellcasting = {
+        label: game.i18n.localize("DND5E.Spellcasting")
+      };
     }
     return _enrichmentLookup;
   },
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -937,8 +937,8 @@ function createPassiveTag(label, dataset) {
  * @returns {string}
  */
 export function createRollLabel(config) {
-  const { label: ability, abbreviation } = CONFIG.DND5E.abilities[config.ability] ?? {};
-  const skill = CONFIG.DND5E.skills[config.skill]?.label;
+  const { label: ability, abbreviation } = CONFIG.DND5E.enrichmentLookup.abilities[config.ability] ?? {};
+  const skill = CONFIG.DND5E.enrichmentLookup.skills[config.skill]?.label;
   const toolUUID = CONFIG.DND5E.enrichmentLookup.tools[config.tool];
   const tool = toolUUID ? Trait.getBaseItem(toolUUID, { indexOnly: true })?.name : null;
   const longSuffix = config.format === "long" ? "Long" : "Short";
@@ -1114,6 +1114,7 @@ async function rollAction(event) {
       }
 
       for ( const actor of actors ) {
+        if ( ability === "spellcasting" ) options.ability = actor.system.attributes?.spellcasting;
         switch ( type ) {
           case "check":
             await actor.rollAbilityCheck(options);
--- a/module/tooltips.mjs
+++ b/module/tooltips.mjs
@@ -126,12 +126,13 @@ export default class Tooltips5e {
     const abilityConfig = CONFIG.DND5E.abilities[ability ?? skillConfig.ability];
 
     let label;
+    const abilityLabel = ability === "spellcasting" ? game.i18n.localize("DND5E.Spellcasting") : abilityConfig.label;
     if ( skillConfig ) {
-      label = game.i18n.format("DND5E.SkillPassiveSpecificHint", { skill: skillConfig.label, ability: abilityConfig.label });
+      label = game.i18n.format("DND5E.SkillPassiveSpecificHint", { skill: skillConfig.label, ability: abilityLabel });
     } else {
       // If no skill was provided, we're doing a passive ability check.
       // This isn't technically a thing in the rules, but we can support it anyway if people want to use it.
-      label = game.i18n.format("DND5E.SkillPassiveHint", { skill: abilityConfig.label });
+      label = game.i18n.format("DND5E.SkillPassiveHint", { skill: abilityLabel });
     }
 
     const party = game.settings.get("dnd5e", "primaryParty")?.actor;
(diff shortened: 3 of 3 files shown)


module/documents/actor/actor.mjs
```diff
@@ -1415,6 +1415,11 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const relevant = type === "skill" ? this.system.skills?.[config.skill] : this.system.tools?.[config.tool];
     const buildConfig = this._buildSkillToolConfig.bind(this, type, config.rolls?.shift());
 
+    if ( config.ability === "spellcasting" ) {
```
arbron, line 1418: Unless there is a particular reason to be able to use this in the rolling methods, I'd probably prefer to handle converting "spellcasting" to a specific ability in the `rollAction` method. Would allow us to write the logic only once. Not going to fight about it though.

arbron (changes requested):
The `_onHoverPassive` method in `tooltips.mjs` will also need adjustment in case someone decides to do a `passive Spellcasting (Arcana)` enricher.

arbron (approved):

Fyorl (changes requested):
Prefer `spell` over `spellcasting` (should probably support both).

> We had a use case for a lot of "make a saving throw with your spellcasting ability" in adventures.

Do you have more context on this? I ask because we're using the creature's default spellcasting ability modifier here, which will not generally be correct for multi-classed characters and/or characters that receive spells via species traits. So I'm interested in understanding how this content disambiguates. Is it a saving throw that is only made during the casting of another spell, or to remove a spell or something?

krbz999:
Scouring through the material, there are no specifics wrt multiclassed characters or anything else.

One example is a saving throw forced by a monster on targets and they "make a saving throw using their spellcasting ability," but no further distinction. Other examples are "[...] cast this spell and succeed on a DC 19 ability check using their spellcasting ability."

Fyorl:
> "[...] cast this spell and succeed on a DC 19 ability check using their spellcasting ability."

This one makes sense. Any spell you cast has an associated spellcasting ability.

> One example is a saving throw forced by a monster on targets and they "make a saving throw using their spellcasting ability," but no further distinction.

This one I don't understand. Multi-classed characters aside, what if the target is a Monk? They don't have a spellcasting ability.

In the system we happen to have a default spellcasting ability which we *can* use for this enricher, but the concept of a default spellcasting ability is more of a legacy implementation holdover. In game terms, it's mostly only applicable to NPCs. PCs don't have default spellcasting abilities. The spellcasting ability for any given spell depends entirely on the source of the spell, which may be a class Spellcasting feature, a feat, a species feature, etc.

The DMG offers some guidance wrt. spells cast from magic items:

> A magic item may require the user to use their own spellcasting ability when casting a spell from the item. If the user has more than one spellcasting ability, the user chooses which one to use with the item. If the user doesn’t have a spellcasting ability, their spellcasting ability modifier is +0 for the item, and the user’s Proficiency Bonus applies.

So we could potentially derive a default spellcasting ability for creatures that don't have one, using these rules. But these are clearly intended for magic items only, so this concept would be our own invention.

I guess I'm mostly unsure if this implementation of the enricher actually covers the intended use case here. Am I perhaps misinterpreting these examples? It's hard to say with these small snippets.

krbz999:
> So we could potentially derive a default spellcasting ability for creatures that don't have one, using these rules. But these are clearly intended for magic items only, so this concept would be our own invention.
> 
> I guess I'm mostly unsure if this implementation of the enricher actually covers the intended use case here. Am I perhaps misinterpreting these examples? It's hard to say with these small snippets.

I think we agree on the examples. Though I'll say there are, far as I remember, some magic items that use Intelligence in the absence of a dedicated spellcasting ability. Maybe that is a sensible fallback.

I've asked for clarification on the examples, but it may be a ways out. I would think, however, that it would be whichever is the highest modifier in case there are multiple.

But we do need some way to have spellcasting abilities tracked easily from other sources than classes (at least I don't think there is any property for this on a race or background item?).

Fyorl:
> I think we agree on the examples. Though I'll say there are, far as I remember, some magic items that use Intelligence in the absence of a dedicated spellcasting ability. Maybe that is a sensible fallback.

Do you have any examples? Again though, anything we do here would just be our own invention and not in-line with the game's rules.

> But we do need some way to have spellcasting abilities tracked easily from other sources than classes (at least I don't think there is any property for this on a race or background item?).

For non-spells we can track sources via advancement only at the moment. More modern feats and species typically allow you to choose the spellcasting ability when you gain the spell, in which case they would be baked into the spell's activity configuration if it features an attack or save. So we probably need a more reliable way to set and retrieve that choice.

The Dispel Magic example from #5092 is a good example, because if you use this enricher in the item's description to retrieve `attributes.spellcasting`, it will be wrong in cases where the character got the spell from a feat/species, or:

1. The character has multi-classed two spellcasting classes with different spellcasting abilities,
2. They have Dispel Magic from the class that is not set as their primary spellcasting ability class.

Maybe that's ok because (1) is pretty rare, and so both (1) & (2) maybe almost never happens. But we do know each spell's source so it is possible to get the correct result here.

We need to make sure this is the final form of the enricher and that it doesn't need additional parameters/syntax that can't be easily added later, since we can't migrate enricher syntax very easily once it has been widely embedded in text content.

krbz999:
Well maybe it boils down to two cases:
- Non-contextual enricher, where the spellcasting ability is just the highest of all the actor's spellcasting abilities.
- Contextual enricher, such as a check in the item description of Dispel Magic, where the spellcasting ability should be equal to the spellcasting ability of the spell.

Adjustments to how spellcasting is retrieved, though, I think is beyond this PR. Maybe this should be marked as blocked until then?

arbron (changes requested):
We've merged in `actor.spellcastingAbility` so this should be adjusted to take advantage of that and should hopefully be unblocked :)

krbz999:
> We've merged in `actor.spellcastingAbility` so this should be adjusted to take advantage of that and should hopefully be unblocked :)

I do not have the bandwidth at present.

arbron:
Closing this because I've adapted the changes into my other PR: https://github.com/foundryvtt/dnd5e/pull/7036

PR #4884, opened 2024-12-15
[#4787] More informative roll dialog titles

### LINKED ISSUE #4787: Roll dialogs no longer have contextual window titles
Titles are now all "Configure Roll" instead of the contextual information they used to provide
| 3.2.1 | 4.1.2 | Proposed |
| ----- | ---- | ---- |
|image | image | image |

roth-michael:
Closes #4787 
Also added a check when putting the subtitle for an activity use in a chat card whether the activity was a heal-type, if so display "Healing Roll" instead of "Damage Roll."
Attack Roll:
image
Damage Roll:
image
Skill Check:
image
Ability Check: 
image
Saving Throw:
image
Tool Check:
image
Fixed subtitle for a heal activity:
image

Haven't tested much with super long item/actor names. Might make sense to put them as subtitles instead of titles.

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -1039,7 +1039,6 @@ export default function ActivityMixin(Base) {
       rollConfig.hookNames = [...(config.hookNames ?? []), "damage"];
       rollConfig.subject = this;
 
-      const titleFlavor = game.i18n.localize(`DND5E.${this.type === "heal" ? "Healing" : "Damage"}Roll`);
       const dialogConfig = foundry.utils.mergeObject({
         options: {
           position: {
@@ -1048,8 +1047,8 @@ export default function ActivityMixin(Base) {
             left: window.innerWidth - 710
           },
           window: {
-            title: `${this.item.name} - ${titleFlavor}`,
-            subtitle: "DND5E.RollConfiguration.Title",
+            title: game.i18n.localize(`DND5E.${this.type === "heal" ? "Healing" : "Damage"}Roll`),
+            subtitle: this.item.name,
             icon: this.item.img
           }
         }
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -1047,7 +1047,7 @@ export default function ActivityMixin(Base) {
             left: window.innerWidth - 710
           },
           window: {
-            title: game.i18n.localize(`DND5E.${this.type === "heal" ? "Healing" : "Damage"}Roll`),
+            title: this.damageFlavor,
             subtitle: this.item.name,
             icon: this.item.img
           }
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -1039,12 +1039,18 @@ export default function ActivityMixin(Base) {
       rollConfig.hookNames = [...(config.hookNames ?? []), "damage"];
       rollConfig.subject = this;
 
+      const titleFlavor = game.i18n.localize(`DND5E.${this.type === "heal" ? "Healing" : "Damage"}Roll`);
       const dialogConfig = foundry.utils.mergeObject({
         options: {
           position: {
             width: 400,
             top: config.event ? config.event.clientY - 80 : null,
             left: window.innerWidth - 710
+          },
+          window: {
+            title: `${this.item.name} - ${titleFlavor}`,
+            subtitle: "DND5E.RollConfiguration.Title",
+            icon: this.item.img
           }
         }
       }, dialog);
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -1348,10 +1348,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
    */
   async rollSkill(config={}, dialog={}, message={}) {
     const skillLabel = CONFIG.DND5E.skills[config.skill]?.label ?? "";
+    const ability = this.system.skills[config.skill]?.ability ?? CONFIG.DND5E.skills[config.skill]?.ability ?? "";
+    const abilityLabel = CONFIG.DND5E.abilities[ability]?.label ?? "";
     const dialogConfig = foundry.utils.mergeObject({
       options: {
         window: {
-          title: game.i18n.format("DND5E.SkillPromptTitle", { skill: skillLabel }),
+          title: game.i18n.format("DND5E.SkillPromptTitle", { skill: skillLabel, ability: abilityLabel }),
           subtitle: this.name
         }
       }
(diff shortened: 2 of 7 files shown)


roth-michael:
Might instead make sense to put the roll type in the main title and the actor/item name in the subtitle (if it's okay to lose "Configure Roll" - which I think it would be, since it's implicit).

roth-michael:
New commit changes to match my last comment.
Skill check:
Screenshot 2024-12-15 190904
Tool check:
Screenshot 2024-12-15 191015
Damage & Healing rolls:
Screenshot 2024-12-15 190952
Screenshot 2024-12-15 191004

arbron:
@roth-michael Give a comment on the original issue so I can assign it to you

module/documents/activity/mixin.mjs
```diff
@@ -1045,6 +1045,11 @@ export default function ActivityMixin(Base) {
             width: 400,
             top: config.event ? config.event.clientY - 80 : null,
             left: window.innerWidth - 710
+          },
+          window: {
+            title: game.i18n.localize(`DND5E.${this.type === "heal" ? "Healing" : "Damage"}Roll`),
```
arbron, line 1050: You should just `this.damageFlavor` for this. You can change it for the heal activity to use `DND5E.HealingRoll` rather than just `DND5E.Healing`.

module/documents/chat-message.mjs
```diff
@@ -307,7 +307,11 @@ export default class ChatMessage5e extends ChatMessage {
     if ( this.isContentVisible && item && roll ) {
       const isCritical = (roll.type === "damage") && this.rolls[0]?.isCritical;
       const subtitle = roll.type === "damage"
-        ? isCritical ? game.i18n.localize("DND5E.CriticalHit") : game.i18n.localize("DND5E.DamageRoll")
+        ? isCritical
+          ? game.i18n.localize("DND5E.CriticalHit") 
+          : activity.type === "heal"
+            ? game.i18n.localize("DND5E.HealingRoll")
+            : game.i18n.localize("DND5E.DamageRoll")
```
arbron, line 314: `activity` isn't guaranteed to exist here. We can also use `damageFlavor` here after adjusting the flavor for `HealActivity`:
```suggestion
          ? game.i18n.localize("DND5E.CriticalHit")
          : activity?.damageFlavor ?? game.i18n.localize("DND5E.DamageRoll")
```

arbron (changes requested):
Looking good. Might also be nice to show "Concentration Save" when rolling for concentration.

roth-michael:
> Might also be nice to show "Concentration Save" when rolling for concentration

Went with the same label but with `Concentration` instead of `Strength` for instance (i.e. "Concentration Saving Throw"). Was thinking about whether instead "Concentration Save (Con)" would make sense, but as-is it looks fine to me.
image

module/documents/actor/actor.mjs
```diff
@@ -1639,7 +1639,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
    * @returns {Promise<D20Roll[]|null>}                        A Promise which resolves to the created Roll instances.
    */
   async rollSavingThrow(config={}, dialog={}, message={}) {
-    const abilityLabel = CONFIG.DND5E.abilities[config.ability]?.label ?? "";
+    const abilityLabel = config.isConcentration ? game.i18n.localize("DND5E.Concentration") : CONFIG.DND5E.abilities[config.ability]?.label ?? "";
```
arbron, line 1642: Move this into the `rollConcentration` method, this method really shouldn't know about the existence of concentration saves as a concept.

arbron (changes requested):
Just the small change about where we set the concentration label, but otherwise looking good.

arbron (approved):

Fyorl (approved):
Good changes.

Would like to change skill titles to more idiomatic language, i.e. 'Acrobatics Skill Check' -> 'Dexterity (Acrobatics) Check' (and the resulting chat flavour text). But that would involve having the window title update based on what's chosen in the ability drop-down. No idea how involved that would be without trying it out first.

roth-michael:
I suspect it wouldn't be too bad - except for the case where `chooseAbility` is true - though I wonder whether it'd feel a bit cramped. Will give it a go tomorrow and see.

roth-michael:
image
Actually doesn't look cramped. If it looks good to you I'll push it. Realized `chooseAbility` is always gonna be on currently, so I think no problem with just leaving the title by whatever the default on the rolling actor's sheet should be.

roth-michael:
This would be the, uh, "clever" way I came up with:

```js
/** @inheritDoc */
  super._onChangeForm(formConfig, event);
  if ( this.config.skill ) {
    const skillLabel = CONFIG.DND5E.skills[this.config.skill]?.label ?? "";
    const ability = this.form.querySelector('select[name="ability"]')?.value ?? this.config.ability;
    const abilityLabel = CONFIG.DND5E.abilities[ability]?.label ?? "";
    this.window.title.innerText = game.i18n.format("DND5E.SkillPromptTitle", { skill: skillLabel, ability: abilityLabel });
  }
}
```

inside `SkillToolRollConfigurationDialog`

Fyorl:
> This would be the, uh, "clever" way I came up with:
> 
> ```js
> /** @inheritDoc */
> _onChangeForm(formConfig, event) {
>   super._onChangeForm(formConfig, event);
>   if ( this.config.skill && event.target?.name === "ability" ) {
>     const skillLabel = CONFIG.DND5E.skills[this.config.skill]?.label ?? "";
>     const ability = event.target.value ?? this.config.ability;
>     const abilityLabel = CONFIG.DND5E.abilities[ability]?.label ?? "";
>     this.window.title.innerText = game.i18n.format("DND5E.SkillPromptTitle", { skill: skillLabel, ability: abilityLabel });
>   }
> }
> ```
> 
> inside `SkillToolRollConfigurationDialog`
> 
> ~Ah. One more issue with including the selected ability in the title: the title (based on the `config`) is also used for the roll flavor in the chat card. So would need to account for that as well somehow so the proper flavor is used.~ Easily fixed by setting `this.message.data.flavor` to the same thing as the title in the above snippet.

Yeah that looks good. It's doing the same thing, but you could use `ApplicationV2#_updateFrame` which is maybe slightly more future-proofed.

```js
/** @inheritDoc */
  super._onChangeForm(formConfig, event);
  if ( this.config.skill && (event.target?.name === "ability") ) {
    const skillLabel = CONFIG.DND5E.skills[this.config.skill]?.label ?? "";
    const ability = event.target.value ?? this.config.ability;
    const abilityLabel = CONFIG.DND5E.abilities[ability]?.label ?? "";
    this._updateFrame({
      window: { title: game.i18n.format("DND5E.SkillPromptTitle", { skill: skillLabel, ability: abilityLabel }) }
    });
  }
}
```

Fyorl (approved):

PR #4886, opened 2024-12-15
Extend rider conditions to generic effects

krbz999:
When an effect granted to a target by some monster feature means that you also suffer from a given condition (eg "while slimed, you are also considered Poisoned"), that effect should simply have the given status. This use case is covered well enough and improved slightly by #4881.

The other case is being affected by a generic condition *in addition* - eg "you are slimed, and also knocked prone," where the two effects/conditions are independent; they are applied at the same time but can be removed at different times. This PR helps with that specific use case.

image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -201,6 +201,7 @@
 "DND5E.ACTIVEEFFECT": {
   "FIELDS": {
     "statuses": {
+      "label": "Status Conditions",
       "hint": "While affected by this Active Effect, the target will be treated as having these additional status conditions."
     }
   }
--- a/lang/en.json
+++ b/lang/en.json
@@ -1229,6 +1229,12 @@
 "DND5E.ConCharmed": "Charmed",
 "DND5E.ConDeafened": "Deafened",
 "DND5E.ConDiseased": "Diseased",
+"DND5E.CONDITIONS": {
+  "RiderConditions": {
+    "label": "Additional Effects",
+    "hint": "Configure additional conditions that will be applied when this effect is created on an actor."
+  }
+},
 "DND5E.ConExhaustion": "Exhaustion",
 "DND5E.ConFrightened": "Frightened",
 "DND5E.ConGrappled": "Grappled",
--- a/lang/en.json
+++ b/lang/en.json
@@ -198,6 +198,14 @@
   }
 },
 
+"DND5E.ACTIVEEFFECT": {
+  "FIELDS": {
+    "statuses": {
+      "hint": "While affected by this Active Effect, the target will be treated as having these additional status conditions."
+    }
+  }
+},
+
 "DND5E.ACTIVITY": {
   "Action": {
     "Create": "Create Activity",
@@ -1231,8 +1239,8 @@
 "DND5E.ConDiseased": "Diseased",
 "DND5E.CONDITIONS": {
   "RiderConditions": {
-    "label": "Additional Effects",
-    "hint": "Configure additional conditions that will be applied when this effect is created on an actor."
+    "label": "Separate Status Conditions",
+    "hint": "When this Active Effect is applied, these additional status conditions will be applied separately."
   }
 },
 "DND5E.ConExhaustion": "Exhaustion",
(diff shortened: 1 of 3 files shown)


Fyorl (changes requested):
Code looks good but we should try to improve the wording if we can. At the moment, the user will just see two multi-selects with the same options. The second multi-select says they are 'additional conditions that will be applied when this effect is created on actor', but from the point of view of the user, the statuses in the first multi-select are also 'applied' when the AE is created. As far as they can tell, the effect is created, and the actor gets all the statuses in both multi-selects.

So we need some way to make it clearer that the first multi-select is 'connected' to the effect, while the second multi-select isn't. There's probably a better word to use than 'connected', not sure. If it helps, we should add a hint to the first multi-select also, and/or change the label test.

krbz999:
> So we need some way to make it clearer that the first multi-select is 'connected' to the effect, while the second multi-select isn't. There's probably a better word to use than 'connected', not sure. If it helps, we should add a hint to the first multi-select also, and/or change the label text.

Should we just get rid of it? Does the `statuses` input field have a good and well-defined purpose in this system?

Fyorl:
> Should we just get rid of it? Does the `statuses` input field have a good and well-defined purpose in this system?

You explained it yourself in the first paragraph of this PR's description, no?

> When an effect granted to a target by some monster feature means that you also suffer from a given condition (eg "while slimed, you are also considered Poisoned"), that effect should simply have the given status.

Am I misunderstanding?

Fyorl:
> Maybe we should just have a custom AE sheet at this point tho.

Yeah, that's certainly on the table. I'd expect the system to end up with a custom AE sheet eventually anyway.

krbz999:
image

lang/en.json
```diff
@@ -198,6 +198,15 @@
   }
 },
 
+"DND5E.ACTIVEEFFECT": {
+  "FIELDS": {
+    "statuses": {
+      "label": "Status Conditions",
```
Fyorl, line 204: In v13 the label disappeared unless I duplicated it here. Not sure why.

module/documents/active-effect.mjs
```diff
@@ -618,6 +631,34 @@ export default class ActiveEffect5e extends ActiveEffect {
...
+      hint: game.i18n.localize("DND5E.CONDITIONS.RiderConditions.hint")
+    }, {
+      name: "flags.dnd5e.riders.statuses",
+      value: app.document.getFlag("dnd5e", "riders.statuses") ?? [],
+      options: CONFIG.statusEffects.map(se => ({ value: se.id, label: se.name }))
+    }).outerHTML;
+    html.querySelector("[data-tab=details] > .form-group:last-child").insertAdjacentHTML("afterend", element);
+
+    if ( game.release.generation < 13 ) {
+      html.querySelector(".form-fields:has([name=statuses])").insertAdjacentHTML("afterend", `
```
Fyorl, line 653: In v13 this was being appended twice because the localisation hint was already being applied by the core application. I guess that's not working in v12? Worth checking if appending this is even necessary.

Fyorl (approved):
Had to make some adjustments for v13 but otherwise it looks good. Let me know if it's still good in v12 with these adjustments, and then we can merge.

krbz999:
Still looks good in v12.

module/documents/active-effect.mjs
krbz999: There is no `hint` in v12 if that is what you mean?

lang/en.json
krbz999: Probably because it does not merge.

https://github.com/foundryvtt/foundryvtt/issues/11489

I should have remembered that.

module/documents/active-effect.mjs
Fyorl: I mean the localisation prefix wasn't applying the hint in v12. Maybe it's a v13-only feature. I've been on v13 so long I don't remember.

krbz999: In v12, it doesn't use `{{formGroup}}`.
```html
        <div class="form-group">
            <label>Status Conditions</label>
            <div class="form-fields">
                <multi-select name="statuses">
                    {{#each statuses as |status|}}
                    <option value="{{status.id}}" {{status.selected}}>{{status.label}}</option>
                    {{/each}}
                </multi-select>
            </div>
        </div>
```

PR #4887, opened 2024-12-16
Update build script with more automation & flexibility

arbron:
- Allows for the same build script to be used for both systems and modules, premium or not
- Automatically determines the contents of the ZIP archive using the package manifest and the optional `includes` option in the configuration file

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,9 @@
 name: Release Creation
 
+env:
+  NODE_VERSION: 22
+
+
 on:
   push:
     tags:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,5 +1,9 @@
 name: Release Creation
 
+env:
+  NODE_VERSION: 18
```
Fyorl, line 4: Already EOL in May this year

Fyorl (approved):
Nice

.github/workflows/release.yml
arbron: I'll bump it up to `22`.

Issue #809, opened 2020-11-12
Agree upon a schema for tracking effects which provide advantage or disadvantage to a certain type of roll.


aaclayton:
I don’t think that the core dnd5e system will ever automate the choice of advantage/disadvantage to a dice roll. However, we can have a discussion about how to store/record that data so that modules can use it to perform that type of automation.

**Tentative Proposal**

Step 1 - agree on a taxonomy of roll types which are tracked in an object like:

```js
advantages: {
  "check:skill:ath": 1, // 1 = advantage
  "check:skill:ste": -1, // -1 = disadvantage
  "attack:mwak": 1,
  "check:save:str": 1,
  "check:test:str": 1
};
```

Step 2 - Which could then be configured on an active effect as:

![image](https://gitlab.com/foundrynet/dnd5e/uploads/3aaf2d26b47996676ca5d28496fac23e/image.png)

Ikaguia:
Was this ever agreed upon or was is just abandoned?

revilowaldow:
Would love to see this support labels on the new character sheet in a similar style to D&D beyond

Fyorl:
The schema is now in place with the implementation of the `RollConfigField`.

PR #4890, opened 2024-12-16
More advantage mode coverage

krbz999:
A continuation of #4838.

Modes:
- Add: Adds one source of advantage (1) or disadvantage (-1).
- Upgrade: Makes it impossible to have disadvantage (0, 1).
- Downgrade: Makes it impossible to have advantage (-1, 0).
- Multiply: Does nothing.
- Override: Forces one advantage mode regardless of other sources, downgrade, or upgrades.
- Custom: is custom.

### DIFF
--- a/module/data/fields/advantage-mode-field.mjs
+++ b/module/data/fields/advantage-mode-field.mjs
@@ -1,4 +1,17 @@
 /**
+ * @typedef AdvantageModeData
+ * @property {number|null} override               Whether the mode has been entirely overridden.
+ * @property {AdvantageModeCounts} advantages     The advantage counts.
+ * @property {AdvantageModeCounts} disadvantages  The disadvantage counts.
+ */
+
+/**
+ * @typedef AdvantageModeCounts
+ * @property {number} count          The number of applications of this mode.
+ * @property {boolean} [suppressed]  Whether this mode is suppressed.
+ */
+
+/**
  * Subclass of NumberField that tracks the number of changes made to a roll mode.
  */
 export default class AdvantageModeField extends foundry.data.fields.NumberField {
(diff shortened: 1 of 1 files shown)


module/data/fields/advantage-mode-field.mjs
```diff
@@ -14,46 +14,115 @@ export default class AdvantageModeField extends foundry.data.fields.NumberField
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Add one to the count of advantages or disadvantages.
+   * @param {Actor5e} actor           The actor gaining a source of advantage or disadvantage.
+   * @param {string} path             The path to the property affected.
+   * @param {boolean} [isAdvantage]   Is this an instance of advantage or disadvantage?
+   */
+  static addSource(actor, path, isAdvantage=true) {
```
Fyorl, line 46: The static model schemas do make accomplishing this a bit more difficult. We do have an instance available to us though, if we don't want to do everything in static scope with maps like this. Active Effects are already a part of data preparation, so I don't think there's any conceptual issue with us applying some additional prepared data that isn't the exact property that was targeted.

For example, where we wanted an instance-level `advantage` property, we could store it on the model instance itself, as a sibling to the targeted property.

```js
const roll = foundry.utils.getProperty(model, change.key.substring(0, change.key.lastIndexOf(".")));
roll.advantage++;
```

The property will be removed/reset when the Actor is re-prepared, like all derived properties. This does rely somewhat on a predictable data structure though.

We could also put these maps on the model instead, and track them via `fieldPath` in a similar way to this PR. Something like `actor.system.advantages`, similar to the original proposal in #809.

I think I prefer keeping the data closer to the targeted field though, which allows the PR to function mostly as it did before.

I did take a look at your original commit in this PR, and the logic looks good to me. So we would want to integrate whichever of the above proposals we settle on with your original logic. Sorry that I was not clearer in my comments on your other PR.

Also, apologies if this is feeling like too much whiplash. I think I have enough that I could make any final changes myself now if you'd prefer. Any thoughts you have on the above are welcome though.

krbz999: I'll force a commit and we'll see what comes of it.

module/data/fields/advantage-mode-field.mjs
```diff
@@ -14,46 +27,106 @@ export default class AdvantageModeField extends foundry.data.fields.NumberField
...
+   *                                      disadvantage (-1), or neither (0).
+   */
+  static resolveMode(model, change, counts) {
+    const { override, advantages, disadvantages } = counts ?? this.getCounts(model, change);
+    if ( override !== null ) return override;
+    const src = foundry.utils.getProperty(model._source, change.key) ?? 0;
+    let advantageCount = advantages.count + Number(src === 1);
+    let disadvantageCount = disadvantages.count + Number(src === -1);
+    if ( advantages.suppressed ) advantageCount = 0;
+    if ( disadvantages.suppressed ) disadvantageCount = 0;
```
krbz999, line 129:
```suggestion
    const advantageCount = advanages.suppressed ? 0 : advantages.count + Number(src === 1);
    const disadvantageCount = disadvantages.suppressed ? 0 : disadvantages.count + Number(src === -1);
```

Fyorl (approved):

Issue #4812, opened 2024-12-01
Add options for flat initiative on NPCs


arbron:
Add an option for NPCs rolled by the GM to always use a flat initiative value, rather than rolling. Would be paired well with grouping actors in the initiative tracker (see https://github.com/foundryvtt/dnd5e/issues/3667).

PR #4897, opened 2024-12-17
[#3667] Add grouping to initiative tracker

### LINKED ISSUE #3667: Add native system support for group initiative
From the beginning of the initiative section in the player's handbook:

> Initiative determines the order of turns during combat. When combat starts, every participant makes a Dexterity check to determine their place in the initiative order. The DM makes one roll for an entire group of identical creatures, so each member of the group acts at the same time.

The dnd5e system currently does not support rolling one initiative value for a group of identical creatures (sharing the same actors). There are modules that attempt to add support for this, but a native implementation would be better and something other modules could build upon.

arbron:
Adds automatic grouping to the initiative tracker. This will group any combatants with the same initiative and the same base actor into a group.

Combat Groups - Collapsed
Combat Groups - Expanded

Groups are displayed as collapsible sections in the combat tracker that use the name from the base actor's prototype token. It lists the number of combatants in the group, and when one of the group members is active it displays how far through the group that actor is.

Currently only offers the one grouping mode, but could be expanded in the future to support different grouping such as splitting it into "friendly" and "hostile" groups.

Also works in V13:

Screenshot 2024-12-20 at 12 26 37

Closes #3667

### DIFF
--- a/module/applications/combat/combat-tracker.mjs
+++ b/module/applications/combat/combat-tracker.mjs
@@ -58,7 +58,7 @@ export default class CombatTracker5e extends CombatTracker {
       if ( !children.length ) continue;
       const groupContainer = document.createElement("li");
       groupContainer.classList.add("combatant", "combatant-group", "collapsible", "dnd5e2-collapsible");
-      if ( !V13 ) groupContainer.classlist.add("directory-item");
+      if ( !V13 ) groupContainer.classList.add("directory-item");
       if ( !expanded ) groupContainer.classList.add("collapsed");
 
       // Determine the count
@@ -82,9 +82,7 @@ export default class CombatTracker5e extends CombatTracker {
         <div class="group-header flexrow">
           <img class="token-image" alt="${img.alt}" src="${img.src || img.dataset.src}">
           <div class="token-name flexcol">
-            <${V13 ? "strong" : "h4"} class="name">
-              ${game.i18n.format("DND5E.COMBAT.Group.Title", { name })}
-            </${V13 ? "strong" : "h4"}>
+            <${V13 ? "strong" : "h4"} class="name"></${V13 ? "strong" : "h4"}>
             <div class="group-numbers">${count}</div>
           </div>
           <div class="token-initiative">
@@ -98,6 +96,7 @@ export default class CombatTracker5e extends CombatTracker {
         </div>
       `;
       groupContainer.dataset.groupKey = key;
+      groupContainer.querySelector(".name").innerText = game.i18n.format("DND5E.COMBAT.Group.Title", { name });
       children[0].before(groupContainer);
       groupContainer.querySelector(".group-children").replaceChildren(...children);
       groupContainer.addEventListener("click", event => {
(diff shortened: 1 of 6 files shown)


krbz999:
> This will group any combatants with the same initiative and the same base actor into a group.

What about:
- `TokenDocument#disposition`
- `PrototypeToken#actorLink`
- `TokenDocument#name`

It's quite common to use a generic monster for a one-off NPC. Steve the Quest Giver Goblin the party rescues from a cage surely shouldn't be grouped with the rest of them.

arbron:
> > This will group any combatants with the same initiative and the same base actor into a group.
> 
> What about:
> 
> * `TokenDocument#disposition`
> * `PrototypeToken#actorLink`
> * `TokenDocument#name`
> 
> It's quite common to use a generic monster for a one-off NPC. Steve the Quest Giver Goblin the party rescues from a cage surely shouldn't be grouped with the rest of them.

We could take disposition & actor link into account (probably never group if actor link is true), but we wouldn't want to use the token name otherwise adding numbers or adjectives like the goblins in the screenshots above would prevent them from ever being grouped.

MaxPat931:
Is this intended to work best with flat initiative per #4812 or are there plans for streamlining the process of setting init values for a group, re:
> The DM makes one roll for an entire group of identical creatures

module/applications/combat/combat-tracker.mjs
```diff
@@ -23,4 +31,81 @@ export default class CombatTracker5e extends CombatTracker {
...
+
+      // Determine the count
+      let activeEntry;
+      for ( const [index, element] of children.entries() ) {
+        if ( element.classList.contains("active") ) activeEntry = index;
+      }
+      let count = game.i18n.format(`DND5E.COMBATANT.Counted.${getPluralRules().select(children.length)}`, {
+        number: formatNumber(children.length)
+      });
+      if ( activeEntry !== undefined ) {
```
Fyorl, line 72:
```suggestion
      if ( activeEntry ) {
```

module/applications/combat/combat-tracker.mjs
```diff
@@ -23,4 +31,81 @@ export default class CombatTracker5e extends CombatTracker {
...
+      }
+
+      const name = combatants[0].token.baseActor.prototypeToken.name;
+      const img = children[0].querySelector("img");
+      groupContainer.innerHTML = `
+        <div class="group-header flexrow">
+          <img class="token-image" alt="${img.alt}" src="${img.src || img.dataset.src}">
+          <div class="token-name flexcol">
+            <${V13 ? "strong" : "h4"} class="name">
+              ${game.i18n.format("DND5E.COMBAT.Group.Title", { name })}
```
Fyorl, line 86: This name needs to be appended safely in case it contains HTML.

Fyorl (changes requested):
Looks good to me, just spotted one thing that needs adjustment.

It might also be nice to extend the core hover functionality somewhat, assuming core makes it possible to do so. If you hover a token on the scene, the corresponding combatant is highlighted. If that combatant is in a collapsed group, it would be nice to highlight the group they belong to. Not required in this PR though.

module/applications/combat/combat-tracker.mjs
arbron: Active entry is the index, so `0` is a valid value.

Fyorl (approved):

PR #4901, opened 2024-12-18
Allow natural weapons to make ranged attacks without "thrown"

arbron:
Reworks how activities fetch their list of available attack types so they can display "Default (melee or ranged)" in the attack activity dialog if relevant (such as thrown melee weapons).

Adjust how that list of valid attack types is generated to that natural weapons can have ranged attacks so long as they have a range value. To ensure modifiers are properly supported this adds a new attack mode of "Ranged" for natural weapons that don't have the thrown property.

Fixes a bug with extra melee critical damage dice being applied even when the weapon is thrown.

Also fixes a bug causing the "reach" migration to be applied when switching weapon modes.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -547,6 +547,7 @@
     "Label": "Attack Mode",
     "Offhand": "Offhand",
     "OneHanded": "One-Handed",
+    "Ranged": "Ranged",
     "Thrown": "Thrown",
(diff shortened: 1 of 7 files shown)


krbz999:
Certainly gonna be very nice to not have your own hands decrease in quantity just because you got long arms.

Fyorl (approved):

PR #4907, opened 2024-12-19
Replace formula data when applying effects or enchantments

### LINKED ISSUE #3809: Allow enchantments to reference data from enchanter actors like summoning does.
Not exactly within the SRD or WOTC- items, but it would be great if enchantments could data belonging to an actor that enchanted an item, such as ability modifiers and other things much like summoning spells.

A main example is an enchantment done by a creature that adds one of the creature's ability score modifier to an ability check. attack roll, or saving throw made by the actor utilizing the enchantment.

- The example given above can be done as a static number enchantment, but its a bit of a hassle to constantly update.

This doesn't come up in maybe "First-Party" sources, but its been occurring semi-regularly in homebrew and third-party choices and I feel like it would be a nice addition to enchantments in general.

### LINKED ISSUE #4437: Allow various properties of active effects to be derived from the originating item when applied
If an item applies a condition or enchantment to a target creature or object, and it has a scaling duration for example, then there is currently no way to reflect this on the duration of the active effect, since these are always static.

Using flags or system data, an effect should make use of the item's data upon application.

### LINKED ISSUE #4465: Special notation for roll data to be evaluated during Effect transfer
Add a notation for Active Effect values to allow for certain roll data to be evaluated immediately when an effect is being transferred to an actor by an Activity.
In the following, the notation will be to replace `@` with `#`.
The feature should evaluate directly from the activity's roll data, so that things such as `#item.uses.value` and `#scaling` could be taken advantage of.

**Example:**
A character has a feature that allows them to grant a bonus to an ally's Survival checks equal to the character's Wisdom modifier + the ally's proficiency bonus. The feature's Activity could apply an effect with the change: `system.skills.sur.bonuses.check | Add | #abilities.wis.mod + @prof`. The `#` roll data will be evaluated before transfer, such that if the character has a Wisdom modifier of 3, the effect applied to the target would be `system.skills.sur.bonuses.check | Add | 3 + @prof`.

krbz999:
Adds a string field `replacement` to change data in the base active effects.

- If `"target"`, the target's (item or actor) roll data is used to replace roll data properties in the `change.value` field.
- If `"origin"`, the origin item / activity / actor's roll data is used.
- If blank, no change is made.

TODO:
- [x] Add these options in the effect config.
- [x] Consider whether the option should be in the data model instead; perhaps a data model for the `base` type, which the `enchantment` type could then extend.
- [ ] Add option to not replace roll data at all, preventing the core replacement taking place. (see #7256).
- [ ] Add an option to evaluate dice formulas.

Some notes:

While I did consider placing this in `_preCreate`, it isn't possible to determine if the creation of the effect was due to just wanting to copy from one item/actor to another during data entry or to grant the item/actor the effect.

Closes #4437.
Closes #3809.
Closes #4465.

### DIFF
--- a/module/data/abstract/active-effect-data-model.mjs
+++ b/module/data/abstract/active-effect-data-model.mjs
@@ -17,6 +17,7 @@ export default class ActiveEffectDataModel extends foundry.data.ActiveEffectType
     const schema = super.defineSchema();
     schema.changes.element.extendFields({
       _id: new DocumentIdField({ initial: () => foundry.utils.randomID() }),
+      conditions: new FiltersField(),
       replacement: new StringField({
         required: true,
         blank: true,
@@ -24,8 +25,7 @@ export default class ActiveEffectDataModel extends foundry.data.ActiveEffectType
           origin: "DND5E.EFFECT.Replacement.Origin",
           target: "DND5E.EFFECT.Replacement.Target"
         }
-      }),
-      conditions: new FiltersField()
+      })
     });
     return schema;
   }
--- a/module/data/abstract/active-effect-data-model.mjs
+++ b/module/data/abstract/active-effect-data-model.mjs
@@ -1,7 +1,7 @@
 import { getHumanReadableAttributeLabel } from "../../utils.mjs";
 import FiltersField from "../fields/filters-field.mjs";
 
-const { DocumentIdField } = foundry.data.fields;
+const { DocumentIdField, StringField } = foundry.data.fields;
 
 /**
  * Abstract base class to add some shared functionality to all of the system's custom active effect types.
@@ -17,6 +17,14 @@ export default class ActiveEffectDataModel extends foundry.data.ActiveEffectType
     const schema = super.defineSchema();
     schema.changes.element.extendFields({
       _id: new DocumentIdField({ initial: () => foundry.utils.randomID() }),
+      replacement: new StringField({
+        required: true,
+        blank: true,
+        choices: {
+          origin: "DND5E.EFFECT.Replacement.Origin",
+          target: "DND5E.EFFECT.Replacement.Target"
+        }
+      }),
       conditions: new FiltersField()
     });
     return schema;
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -529,16 +529,16 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
 
   /**
    * Convert roll data properties in change data.
-   * @param {object[]} changes        Effect change data in json-serializable format.
-   * @param {*} origin                The origin of the effect to be applied.
-   * @param {Actor5e|Item5e} target   The target of the effect to be applied.
-   * @returns {Promise<object[]>}     A promise that resolves to the modified change data.
+   * @param {object[]} changes         Effect change data in json-serializable format.
+   * @param {*} [origin]               The origin of the effect to be applied.
+   * @param {Actor5e|Item5e} [target]  The target of the effect to be applied.
+   * @returns {Promise<object[]>}      A promise that resolves to the modified change data.
    */
   static async forApplication(changes, origin, target) {
     const oData = origin?.getRollData() ?? {};
     const tData = target?.getRollData() ?? {};
 
-    changes = foundry.utils.deepClone(changes);
+    changes = foundry.utils.deepClone(changes ?? []);
     for ( const change of changes ) {
       // TODO: Evaluate properties in objects, as this may be valid json.
       if ( typeof change.value !== "string" ) continue;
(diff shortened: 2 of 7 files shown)


module/documents/active-effect.mjs
```diff
@@ -94,6 +94,26 @@ export default class ActiveEffect5e extends ActiveEffect {
 
   /* -------------------------------------------- */
 
+  /**
+   * Synchronously retrieve the source Actor or Item, or null if it could not be determined.
+   * @returns {Actor5e|Item5e|null}
+   */
+  getSourceSync() {
+    if ( !this.origin || this.origin.startsWith("Compendiun.") ) return null;
```
roth-michael, line 102: `Compendiun` -> `Compendium`

tposney:
Very nice - I can rip out most of the guts of dae if this goes in.

krbz999:
Most of this can probably be ripped out with v14 since "replace" doesnt need to exist as an option.

arbron:
I was thinking about this the other day and how we might be able to use a custom `changes` data model in V14 to do this without the flag. We could have an option for each change allowing creators to specify whether that specific change is evaluated before application using either the granter's roll data or the grantee's.

With V14 we don't have to worry about using formulas to apply to number fields, but even for formula fields which don't pre-evaluate in V14 we still might want to resolve rolled values on application in case the field is deterministic.

For your note in limitations above about not being able to share a value between fields, you could always start with a change to a flag that resolves the value and then the other changes can just reference that flag.

module/data/abstract/active-effect-data-model.mjs
```diff
@@ -17,6 +17,14 @@ export default class ActiveEffectDataModel extends foundry.data.ActiveEffectType
...
       _id: new DocumentIdField({ initial: () => foundry.utils.randomID() }),
+      replacement: new StringField({
+        required: true,
+        blank: true,
+        choices: {
+          origin: "DND5E.EFFECT.Replacement.Origin",
+          target: "DND5E.EFFECT.Replacement.Target"
+        }
+      }),
       conditions: new FiltersField()
```
arbron:
```suggestion
      conditions: new FiltersField(),
      replacement: new StringField({
        required: true,
        blank: true,
        choices: {
          origin: "DND5E.EFFECT.Replacement.Origin",
          target: "DND5E.EFFECT.Replacement.Target"
        }
      })
```

module/documents/activity/enchant.mjs
```diff
@@ -176,6 +176,22 @@ export default class EnchantActivity extends ActivityMixin(BaseEnchantActivityDa
     if ( concentration ) flags.dependentOn = concentration.uuid;
     const enchantmentData = effect.clone({ origin: this.uuid, "flags.dnd5e": flags }).toObject();
 
+    const originData = this.getRollData();
```
arbron: Since we need to do this in multiple places, why don't we move the actual replacement process into `ActiveEffect5e`. We could either add a new method to create an effect ready to be applied (`effect.forApplication(target, { origin })`), or by adding new options to the `clone` method (`effect.clone({}, { application: { target, origin } })`).

arbron (changes requested):

module/documents/activity/enchant.mjs
krbz999: Like mentioned in the desc it would be good (perhaps in a followup) to evaluate dice formulas, so it should be async imo. I think I would opt for a static `ActiveEffect5e.forApplication` method that creates just the `changes`.

module/documents/active-effect.mjs
```diff
@@ -483,6 +483,38 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * Convert roll data properties in change data.
+   * @param {object[]} changes        Effect change data in json-serializable format.
+   * @param {*} origin                The origin of the effect to be applied.
+   * @param {Actor5e|Item5e} target   The target of the effect to be applied.
+   * @returns {Promise<object[]>}     A promise that resolves to the modified change data.
```
arbron: To clarify that neither of these are required.

```suggestion
   * @param {object[]} changes         Effect change data in json-serializable format.
   * @param {*} [origin]               The origin of the effect to be applied.
   * @param {Actor5e|Item5e} [target]  The target of the effect to be applied.
   * @returns {Promise<object[]>}      A promise that resolves to the modified change data.
```

arbron (changes requested):
One other place we'll also want to handle this is in the `ApplyActiveEffect5eRegionBehaviorType`. If the region it belongs to was created by an activity, then the `dnd5e.activity` (recently renamed) flag on the region will get the creating activity, though we'll have to make sure it handles manually created regions that won't have origin data.

Fyorl (approved):

Issue #4644, opened 2024-11-05
Add a general key of "all" for various traits and proficiencies


krbz999:
Common use cases:
- [x] A creature that knows all languages (current workaround is to add "standard" and "exotic"). (https://github.com/foundryvtt/dnd5e/pull/4908)
- [x] Proficiency/Expertise with all tools (see: artificer nonsense). (https://github.com/foundryvtt/dnd5e/issues/6135)
- [x] Resistance to all damage. (https://github.com/foundryvtt/dnd5e/pull/6325))
- [x] Temporary immunity to all damage. (https://github.com/foundryvtt/dnd5e/pull/6325)
- [ ] Proficiency with all weapons and armor (a Fighter is technically proficient with simple, martial, light, medium, heavy, and shields, which just *happens* to be all, though a third party class might differ in this, and not be equally applicable if new categories are added, such as firearms).

System itself might be able to make use of this directly in `TraitsField.prepareResistImmune`.

Issue #3798, opened 2024-07-04
Embeddable actor stat blocks


Kikimor-rec:
At the moment, the embed extension displays the actor's biography in the journal. 
Maybe it makes sense to output a short card about an actor instead?

I made a not very good but working script for my journals to show how it could look like.

![image](https://github.com/foundryvtt/dnd5e/assets/115701754/18910b3e-b36d-4709-bebb-3a26ab29537d)

```
Hooks.once('ready', () => {
    initEnhancedActorEmbed();
});

function initEnhancedActorEmbed() {
    const originalEnrichHTML = TextEditor.enrichHTML;


    const sizeTranslations = {
        "tiny": "Tiny",
        "sm": "Small",
        "med": "Medium",
        "lg": "Large",
        "huge": "Huge",
        "grg": "Gargantuan"
    };

    function translateCR(cr) {
        const crMap = {
            "0.125": "1/8",
            "0.25": "1/4",
            "0.5": "1/2"
        };
        return crMap[cr] || cr;
    }

    TextEditor.enrichHTML = async function (content, options = {}) {

        if (content === null || content === undefined) {
            return content;
        }

        // Regex for dearch @embed[Actor...] or @embed[Compendium.actor...]
        const embedRegex = /@embed\[([^\]]+)\](?:{([^}]*)})?/g;

        const embedPromises = [];
        content = content.replace(embedRegex, (match, actorRef, label) => {
            const placeholder = `__ACTOREMBED_${embedPromises.length}__`;
            embedPromises.push(processActorEmbed(actorRef, label || actorRef));
            return placeholder;
        });

        const enriched = await originalEnrichHTML.call(this, content, options);

        const embedResults = await Promise.all(embedPromises);
        return embedResults.reduce((acc, embed, index) => {
            return acc.replace(`__ACTOREMBED_${index}__`, embed);
        }, enriched);
    };

    async function processActorEmbed(actorRef, label) {
        let actor;

        try {
            actor = await fromUuid(actorRef);
        } catch (error) {
            console.error("Error fetching actor:", error);
        }

        if (!actor) {
            console.warn(`Actor not found: ${actorRef}`);
            return `[Actor not found: ${actorRef}]`;
        }

        return createEnhancedActorEmbed(actor, label);
    }

    function createEnhancedActorEmbed(actor, label) {

        //  Get the creature's type and size, translate them.
        const type = actor.system.details.type.value || 'Unknown type';

[...]

PR #4908, opened 2024-12-19
[#3798, #4644] Add "All Languages", group primordial languages

arbron:
Adds an "All Languages" option. If selected, then only that option will show an all other language proficiencies will be hidden (custom languages and ranged communication will still be displayed as normal). This includes some broader changes that should work as the basis for adding this to other trait types (such as resistance to all damage).

Also modifies how the labels for languages are created to match modern stat blocks that display grouped languages next to the group name (e.g. "Primordial (Terran)"). This also handles some checks to prevent redundant information, such as if you have Primordial proficiency as well as Terran proficiency, it will only show the top-level group and not its children.

### DIFF
--- a/module/applications/actor/sheet-v2-mixin.mjs
+++ b/module/applications/actor/sheet-v2-mixin.mjs
@@ -263,7 +263,7 @@ export default function ActorSheetV2Mixin(Base) {
 
       // Handle languages
       const languages = this.actor.system.traits?.languages?.labels;
-      if ( languages?.dialects?.length ) traits.languages = languages.dialects.map(label => ({ label }));
+      if ( languages?.languages?.length ) traits.languages = languages.languages.map(label => ({ label }));
       for ( const [key, { label }] of Object.entries(CONFIG.DND5E.communicationTypes) ) {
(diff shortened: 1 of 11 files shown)


module/applications/actor/sheet-v2-mixin.mjs
```diff
@@ -260,13 +260,12 @@ export default function ActorSheetV2Mixin(Base) {
...
-        const data = this.actor.system.traits?.languages?.communication?.[key];
-        if ( !data?.value ) return;
-        traits.languages ??= [];
-        traits.languages.push({ label: `${label} ${formatDistance(data.value, data.units)}` });
-      }
+      // Handle languages
+      const languages = this.actor.system.traits?.languages?.labels;
+      if ( languages?.dialects?.length || languages?.ranged?.length ) traits.languages = [
+        ...languages.dialects.map(label => ({ label })),
+        ...languages.ranged.map(label => ({ label }))
```
Fyorl, line 267: Probably still want to use the original loop here so telepathy can be formatted in the same way as senses and movement.

Fyorl (changes requested):

Fyorl (approved):

PR #4909, opened 2024-12-19
[#2094] Move actor data preparation into data models

arbron:
Moves skill, tool, armor class, initiative, spellcasting ability, condition immunity, and scale value preparation into data models. This leaves only spellcasting preparation in the main actor class.


Fyorl (approved):
Nice

PR #4911, opened 2024-12-20
[#4861] Add combat activation types that display in turn message

### LINKED ISSUE #4861: Add "Start of Combat" and "Start of Turn" activation types
When combat begins or when a turn begins, this would place a message in chat with a button to allow for easily triggering them. This would be useful for an ability like Regeneration which could automatically prompt DMs at the start of an NPCs turn and then they can decide whether to actually trigger the regeneration.

arbron:
Adds three new activation types in a new "Combat" section: "Start of Encounter", "Start of Turn", and "End of Turn". These activation types will display in the combat chat message with a "Use" button to be easily activated. They also have a tooltip to display the full item description on hover.

The turn chat message stores relative UUIDs of the activiations in its data model to avoid having to check all of the actor's items every time the chat message is rendered.

Closes #4861

### DIFF
--- a/module/data/chat-message/turn-message-data.mjs
+++ b/module/data/chat-message/turn-message-data.mjs
@@ -11,6 +11,7 @@ const { DocumentIdField, SchemaField, SetField, StringField } = foundry.data.fie
 /**
  * Data stored in a combat turn chat message.
  *
+ * @property {Set<string>} activations  Activities that can be used with these periods, stored as relative UUIDs.
  * @property {ActorDeltasData} deltas   Actor/item recovery from this turn change.
  * @property {object} origin
  * @property {string} origin.combat     ID of the triggering combat.
@@ -26,6 +27,7 @@ export default class TurnMessageData extends ChatMessageDataModel {
   /** @override */
   static defineSchema() {
     return {
+      activations: new SetField(new StringField()),
       deltas: new ActorDeltasField(),
       origin: new SchemaField({
         combat: new DocumentIdField({ nullable: false, required: true }),
@@ -39,6 +41,9 @@ export default class TurnMessageData extends ChatMessageDataModel {
 
   /** @inheritDoc */
   static metadata = Object.freeze(foundry.utils.mergeObject(super.metadata, {
+    actions: {
+      use: TurnMessageData.#useActivity
+    },
     template: "systems/dnd5e/templates/chat/turn-card.hbs"
   }, { inplace: false }));
 
@@ -51,7 +56,7 @@ export default class TurnMessageData extends ChatMessageDataModel {
    * @type {Actor5e}
    */
   get actor() {
-    return this.combatant?.actor;
+    return this.combatant?.actor ?? this.parent.getAssociatedActor();
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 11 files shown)


module/documents/combatant.mjs
```diff
@@ -31,11 +33,11 @@ export default class Combatant5e extends Combatant {
         whisper: game.users.filter(u => this.actor.testUserPermission(game.user, "OWNER"))
       }
     };
+    console.log(this.actor, periods, messageConfig.data.system.activations);
```
krbz999, line 36: Gezundheit.

templates/chat/turn-card.hbs
```diff
@@ -18,9 +18,27 @@
...
+            {{#each activities}}
+            <li class="activity flexrow item-tooltip" data-activity-uuid="{{ uuid }}" data-item-uuid="{{ item.uuid }}">
+                <img class="gold-icon" src="{{ item.img }}" alt="{{ item.name }}">
+                <span class="name-stacked">
+                    <span class="title">{{ item.name }}</span>
+                    <span class="subtitle">{{ name }}</span>
+                </span>
+                <button type="button" data-action="use">
+                    {{ localize "DND5E.Use" }}
+                </button>
```
Fyorl, line 34: This button looks a bit chunky in-line compared to our other in-line button style in the case of applying effects from the effects tray. I was thinking of using a single icon and making it smaller to match the effects tray icon, but an icon to represent 'use' might be a bit abstract.

Could we instead remove this button and give the item header the `rollable` class instead? Clicking an item header is pretty well established in the system by now as a way to 'use' it I think. That also frees up the right-hand side in case we want to put something else there (maybe charges or something).

Fyorl (changes requested):
In addition to my comments I noticed a couple of issues while testing:

- It seems that `turnStart` is not fired for the first combatant of the combat when the combat is started. When I tested this I had only one combatant in the combat and they had rolled initiative. This might be a core v12 bug though.
- Giving an activity one of the new combat activations puts them in the 'actions' section on the NPC features tab if 'group by origin' is enabled. Notably this is distinct from the 'action' section where all other actions are. I think there's still a todo for pluralising those section titles, but I think this issue is unrelated, and that combat activations should probably be sorted into the 'features' section (that we should also rename to 'traits' at some point).

module/data/chat-message/turn-message-data.mjs
```diff
@@ -106,4 +116,55 @@ export default class TurnMessageData extends ChatMessageDataModel {
...
+  /**
+   * Find any activity relative UUIDs on this actor that can be used during a set of combat periods.
+   * @param {Actor5e} actor
+   * @param {string[]} periods
+   * @returns {Set<string>}
+   */
+  static getActivations(actor, periods) {
+    return actor.items
+      .map(i => i.system.activities?.filter(a => periods.includes(a.activation?.type)).map(a => a.relativeUUID) ?? [])
+      .flat();
```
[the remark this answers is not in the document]
arbron: `Collection` doesn't have `flatMap`.

arbron:
> * It seems that `turnStart` is not fired for the first combatant of the combat when the combat is started. When I tested this I had only one combatant in the combat and they had rolled initiative. This might be a core v12 bug though.

Had to move it from `nextTurn` to `_onStartTurn` & `_onEndTurn`, which also fixes a potential issue with players pressing the "Next Turn" button and creating chat messages for combatants they don't own.

I've realized we might have a bug with updating the dynamic ring during combat, because it looks like it only updates for the user who presses the "Next Turn" button. Might need to move that code into the `Combat5e._onUpdate` to ensure it is called for all clients.

Fyorl (approved):

PR #4912, opened 2024-12-20
[#3629, #4142] Add units to container capacity, add volume capacity

### LINKED ISSUE #3629: Add units to "container"-type item "Capacity" info when using "Weight" capacity type
With the introduction of units to item weights, it would be great if item containers could define their unit of capacity:

image

### LINKED ISSUE #4142: Add "volume" container capacity
Some containers specify a maximum internal volume they can contain. While this isn't something we could possibly track, it would still be nice to be able to specify it alongside the other capacity information on container sheets.

arbron:
Reworks container capacity to allow for defining more than one capacity type (so a container can define a maximum number of items and a maximum weight). Weight capacity now accepts actual units so the weight can be measured in kilograms or tonnes if desired.

Container Units

Adds a volume capacity type to containers for measuring in cubic feet or liters. This includes the addition of `volumeUnits` to the configuration and a "Use Metric Volume Units" setting to match the other units settings.

Closes #3629
Closes #4142

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2358,6 +2358,10 @@ DND5E.defaultUnits = {
     imperial: "ft",
     metric: "m"
   },
+  volume: {
+    imperial: "cubicFoot",
+    metric: "liter"
+  },
   weight: {
     imperial: "lb",
     metric: "kg"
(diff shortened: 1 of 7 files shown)


module/config.mjs
```diff
@@ -2447,6 +2451,29 @@ preLocalize("distanceUnits");
...
+    label: "DND5E.UNITS.VOLUME.CubicFoot.Label",
+    abbreviation: "DND5E.UNITS.Volume.CubicFoot.Abbreviation",
+    counted: "DND5E.UNITS.Volume.CubicFoot.Counted",
+    conversion: 1,
+    type: "imperial"
+  },
+  liter: {
+    label: "DND5E.UNITS.VOLUME.Liter.Label",
+    abbreviation: "DND5E.UNITS.Volume.Liter.Abbreviation",
+    conversion: 1 / 28.317, // Should we do an approximate conversion here? Does it matter?
```
krbz999, line 2469: I'm sure localized 5e books have some simplified conversion of cubic feet to liters. Got an example?

Fyorl: Bag of Holding

arbron: According to Zhell on the German SRD uses the real world conversion between cubit feet & liters.

krbz999: https://discord.com/channels/170995199584108546/795664425255174195/1319796563710709910

No knowledge on German legacy or modern PHB, only legacy German SRD.

Fyorl (approved):
See my comment about multiple container capacities, but we can move ahead with this for now.

PR #4913, opened 2024-12-20
[#3617] Add enchantment effect for modifying damage types

### LINKED ISSUE #3617: Allow Enchantments to change only damage types
At the moment it doesnt seem possible to change just the damage type of a weapon using an enchantment which is not ideal for features like elemental cleaver that can infuse a weapon with an element and change the damage type of the weapon (side note it can do so for magic weapons as well like a frost brand weapon, meaning making an enchantment for every weapon type isn't ideal)

arbron:
Adds the ability for enchantments to replace or add to the damage types for all parts on an activity. This is done using an active effect targeting `system.damage.types`. Using the `OVERRIDE` mode will replace all existing damage types with the specified type, while `ADD` will introduce a new damage type to the existing types.

Closes #3617

### DIFF
--- a/module/applications/activity/activity-sheet.mjs
+++ b/module/applications/activity/activity-sheet.mjs
@@ -243,15 +243,15 @@ export default class ActivitySheet extends Application5e {
       value,
       label: CONFIG.DND5E.activityConsumptionTypes[value].label
     }));
-    context.consumptionTargets = context.activity.consumption.targets.map(data => {
+    context.consumptionTargets = context.source.consumption.targets.map((data, index) => {
       const typeConfig = CONFIG.DND5E.activityConsumptionTypes[data.type] ?? {};
       const showTextTarget = typeConfig.targetRequiresEmbedded && !this.item.isEmbedded;
       const target = new ConsumptionTargetData(data, { parent: this.activity });
       return {
         data,
         fields: this.activity.schema.fields.consumption.fields.targets.element.fields,
-        prefix: `consumption.targets.${data._index}.`,
-        source: context.source.consumption.targets[data._index] ?? data,
+        prefix: `consumption.targets.${index}.`,
+        source: context.source.consumption.targets[index] ?? data,
         typeOptions: consumptionTypeOptions,
         scalingModes: canScale ? [
           { value: "", label: game.i18n.localize("DND5E.CONSUMPTION.Scaling.None") },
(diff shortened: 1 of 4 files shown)


arbron:
> There are some changes here which seem to allow the activity sheet to show various parts of consumption/recovery/damage that have been added via enchantment, but I'm not sure how sustainable this approach is.
> 
> What if an enchantment modifies something that's not in a parts collection? We have the same problem as with AEs, no? The field can't show both the source and modified versions simultaneously.
> 
> Do we add an edit lock to activity sheets as well? Or we could instead consider the activity sheet an edit-only sheet, and enchantment modifications show up only in prepared views of the activity, such as Actor sheets? Or do we restrict enchantment modifications only to various parts collections of an activity?

A big issue I was trying to resolve here was what to do with damage parts, because I want to show the original version of any part that is changed by enchantment, but I also need to show the base weapon damage part which isn't in the source data.

Perhaps we can treat them as edit-only sheets for now and just have some special handling for damage parts to display enchantment-added parts as locked.

module/applications/activity/activity-sheet.mjs
```diff
@@ -369,23 +369,20 @@ export default class ActivitySheet extends Application5e {
...
-            value, label: config.label, selected: data.types.has(value)
-          }))
-        };
-        return this._prepareDamagePartContext(context, part);
+      const typeOptions = Object.entries(CONFIG.DND5E.damageTypes).map(([value, { label }]) => ({ value, label }));
+      const makePart = (data, index) => this._prepareDamagePartContext(context, {
+        data, index, scalingOptions, typeOptions,
+        canScale: this.activity.canScaleDamage,
+        fields: this.activity.schema.fields.damage.fields.parts.element.fields,
+        prefix: `damage.parts.${index}.`,
```
Fyorl, line 377: Seems to result in the base damage part having a name of `damage.parts.undefined.`. Seemed to work ok when I tested, but perhaps not intentional.

Fyorl (changes requested):

module/applications/activity/activity-sheet.mjs
arbron: Before it was `damage.parts.-1` for base damage, so long as it isn't a valid array index it should be fine.

Fyorl (approved):

PR #4922, opened 2024-12-25
Add advantage modes to `abilities` objects

krbz999:
(Built on top of #4890.)

Adds `system.abilities.<abilityId>.check.roll` and `.save.roll` objects.

Moves the derived `abilities.<abilityId>.save` value into `.check/.save.value` with a `toString` method for the deprecation period.

(Also removes the `ability` property from the `attributes.death` object.)

TODO:
- [ ] Add the roll mode and min/max stuff to ability configs.
- [ ] Hook up actually using the data in rolls.

### DIFF
--- a/module/data/actor/templates/common.mjs
+++ b/module/data/actor/templates/common.mjs
@@ -163,9 +163,8 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
 
       // Deprecations.
       abl.save.toString = function() {
-        foundry.utils.logCompatibilityWarning("The 'abilities.<ability>.save' property is now stored in 'abilities.<ability>.save.value'.", {
-          since: "4.2", until: "4.5"
-        });
+        foundry.utils.logCompatibilityWarning("The 'abilities.<ability>.save' property is now stored in "
+          + "'abilities.<ability>.save.value'.", { since: "4.3", until: "4.5" });
         return abl.save.value;
       };
     }
--- a/module/data/actor/templates/common.mjs
+++ b/module/data/actor/templates/common.mjs
@@ -16,8 +16,8 @@ const { NumberField, SchemaField } = foundry.data.fields;
  * @property {object} bonuses        Bonuses that modify ability checks and saves.
  * @property {string} bonuses.check  Numeric or dice bonus to ability checks.
  * @property {string} bonuses.save   Numeric or dice bonus to ability saving throws.
- * @property {RollConfigFieldData} check    Properties related to ability checks.
- * @property {RollConfigFieldData} save     Properties related to ability saving throws.
+ * @property {RollConfigData} check    Properties related to ability checks.
+ * @property {RollConfigData} save     Properties related to ability saving throws.
  */
 
 /**
--- a/module/data/actor/templates/common.mjs
+++ b/module/data/actor/templates/common.mjs
@@ -16,16 +16,8 @@ const { NumberField, SchemaField } = foundry.data.fields;
  * @property {object} bonuses        Bonuses that modify ability checks and saves.
  * @property {string} bonuses.check  Numeric or dice bonus to ability checks.
  * @property {string} bonuses.save   Numeric or dice bonus to ability saving throws.
- * @property {object} check          Properties related to ability checks.
- * @property {object} check.roll
- * @property {number} check.roll.mode   The advantage mode of ability checks.
- * @property {number} check.roll.min    The minimum that can be rolled on the d20.
- * @property {number} check.roll.max    The maximum that can be rolled on the d20.
- * @property {object} save              Properties related to ability saving throws.
- * @property {object} save.roll
- * @property {number} save.roll.mode    The advantage mode of ability saving throws.
- * @property {number} save.roll.min     The minimum that can be rolled on the d20.
- * @property {number} save.roll.max     The maximum that can be rolled on the d20.
+ * @property {RollConfigFieldData} check    Properties related to ability checks.
+ * @property {RollConfigFieldData} save     Properties related to ability saving throws.
  */
 
 /**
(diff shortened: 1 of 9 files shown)


JPMeehan:
Wasn't there a discussion that for AEs to work properly you probably should split dis & adv so that way the cancelling can be parsed correctly?

krbz999:
> Wasn't there a discussion that for AEs to work properly you probably should split dis & adv so that way the cancelling can be parsed correctly?

See the AdvantageModeField.

JPMeehan:
Those are properties on the shared model schema – does this not have collisions across multiple actors of the same type?

krbz999:
> Those are properties on the shared model schema – does this not have collisions across multiple actors of the same type?

I don't see why - the field is reset when initialized, and actors don't share the same instance of a DataField.

JPMeehan:
image

Fields are properties of the static class - `myModel.schema` accesses `myModel.constructor.schema`. That's why all of the field methods like _cast take in the data they're modifying.

module/data/actor/templates/common.mjs
```diff
@@ -15,6 +16,16 @@ const { NumberField, SchemaField } = foundry.data.fields;
...
+ * @property {object} check          Properties related to ability checks.
+ * @property {object} check.roll
+ * @property {number} check.roll.mode   The advantage mode of ability checks.
+ * @property {number} check.roll.min    The minimum that can be rolled on the d20.
+ * @property {number} check.roll.max    The maximum that can be rolled on the d20.
+ * @property {object} save              Properties related to ability saving throws.
+ * @property {object} save.roll
+ * @property {number} save.roll.mode    The advantage mode of ability saving throws.
+ * @property {number} save.roll.min     The minimum that can be rolled on the d20.
+ * @property {number} save.roll.max     The maximum that can be rolled on the d20.
```
arbron, line 28:
```suggestion
 * @property {object} check          Properties related to ability checks.
 * @property {RollConfigData} check.roll
 * @property {object} save           Properties related to ability saving throws.
 * @property {RollConfigData} save.roll
```

module/data/actor/templates/common.mjs
```diff
@@ -41,7 +52,10 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
         bonuses: new SchemaField({
           check: new FormulaField({ required: true, label: "DND5E.AbilityCheckBonus" }),
           save: new FormulaField({ required: true, label: "DND5E.SaveBonus" })
-        }, { label: "DND5E.AbilityBonuses" })
+        }, { label: "DND5E.AbilityBonuses" }),
+        save: new RollConfigField({
+          ability: false
+        })
```
arbron, line 58:
```suggestion
        check: new RollConfigField({ ability: false }),
        save: new RollConfigField({ ability: false })
```

arbron (commented):
Seems like a good start 👍

module/data/shared/roll-config-field.mjs
```diff
@@ -3,8 +3,8 @@ import AdvantageModeField from "../fields/advantage-mode-field.mjs";
 const { StringField, NumberField, SchemaField } = foundry.data.fields;
 
 /**
- * @typedef {object} RollConfigData
- * @property {string} ability   Default ability associated with this roll.
+ * @typedef {object} RollConfigFieldData
```
arbron, line 6: Make sure you update other usages of this if you are renaming it, it is referenced in `actor/templates/creature.mjs`.

module/data/shared/roll-config-field.mjs
```diff
@@ -27,6 +26,13 @@ export default class RollConfigField extends foundry.data.fields.SchemaField {
       }),
       ...fields
     };
+    if ( ability !== false ) {
+      fields.ability = new StringField({
+        required: true,
+        initial: ability,
+        label: "DND5E.AbilityModifier"
+      });
+    }
```
arbron, line 35: You can follow this pattern which we have used in `MovementField` and elsewhere to allow for deleting fields (add the field back above of course). I find it a bit more readable to define all of the fields in the same way:

```suggestion
   Object.entries(fields).forEach(([k, v]) => !v ? delete fields[k] : null);
```

arbron (changes requested):

module/data/shared/roll-config-field.mjs
krbz999: Had misread those same ones, should not have been renamed.

module/data/shared/roll-config-field.mjs
```diff
@@ -18,6 +18,11 @@ export default class RollConfigField extends foundry.data.fields.SchemaField {
   constructor({roll={}, ability="", ...fields}={}, options={}) {
     const opts = { initial: null, nullable: true, min: 1, max: 20, integer: true };
     fields = {
+      ability: ( ability === false ) ? null : new StringField({
```
arbron, line 21: You won't need the check here, because the `...fields` at the end will replace this with `false` for you.

krbz999: How so? We destructure the property because it should be `""` if omitted.

arbron: Aw, yeah, I forgot you are also using `ability` to set the default value as well as to remove the field. Ignore me.

arbron (approved):

YenBenGrey:
Apologies if this is completely the wrong place to put this, however there is a tenuous link. (not sure if you can make an issue for a PR)

If this gets implemented, would it be possible for the advantage, disadvantage buttons to be clearer/brighter/shinier rather then a thin red line around the button like it shows for a normal hit and none crit damage.

krbz999:
> If this gets implemented, would it be possible for the advantage, disadvantage buttons to be clearer/brighter/shinier rather then a thin red line around the button like it shows for a normal hit and none crit damage.

Would be a separate issue.

Fyorl (approved):

Issue #4924, opened 2024-12-25
If `system.bonuses.mwak.damage === "0"`, then all AEs with the same attribute are ignored


krbz999:
Repro:

- In special traits, set `system.bonuses.mwak.damage` to `0`.
- Add an AE with `system.bonuses.mwak.damage | Add | 1d123`.
- Roll damage with a melee weapon.
- The 1d123 is not added.
- Disable the AE, remove the `0` in Special Traits, re-enable the AE.
- Roll damage with a melee weapon.
- The 1d123 is added as expected.

PR #4925, opened 2024-12-25
[#4924] Change `parseInt` to `Number`

krbz999:
Currently the system uses `parseInt`, which means a bonus such as `0 + 123` gets ignored because `parseInt("0 + 123")` is `0`.

iirc there was a reason we had this small validation step so I changed it for `Number` instead, but it would otherwise seem to me that we could just change to `bonus !== "0"`.

### DIFF
--- a/module/data/activity/base-activity.mjs
+++ b/module/data/activity/base-activity.mjs
@@ -686,7 +686,7 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
       let actionType = this.actionType;
       if ( (actionType === "mwak") && rollConfig.attackMode?.startsWith("thrown") ) actionType = "rwak";
       const bonus = foundry.utils.getProperty(this.actor ?? {}, `system.bonuses.${actionType}.damage`);
-      if ( bonus && (parseInt(bonus) !== 0) ) parts.push(bonus);
+      if ( bonus && (Number(bonus) !== 0) ) parts.push(bonus);
     }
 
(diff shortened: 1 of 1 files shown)


module/data/activity/base-activity.mjs
```diff
@@ -686,7 +686,7 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
       let actionType = this.actionType;
       if ( (actionType === "mwak") && rollConfig.attackMode?.startsWith("thrown") ) actionType = "rwak";
       const bonus = foundry.utils.getProperty(this.actor ?? {}, `system.bonuses.${actionType}.damage`);
-      if ( bonus && (parseInt(bonus) !== 0) ) parts.push(bonus);
+      if ( bonus && (Number(bonus) !== 0) ) parts.push(bonus);
```
arbron, line 689: Number behaves differently than `parseInt`, so `Number("2 + 2")` results in `NaN`, not `4`.

The intention here is to detect `"0"` or a blank string, so we could just do the `bonus !== "0"`.

arbron (changes requested):

module/data/activity/base-activity.mjs
krbz999: Well NaN is fine, then. The important thing is that it is not `"0"` (or `"000"` repeating).

arbron (approved):

PR #4926, opened 2024-12-26
Show ability-mod-using bonuses on sheet

roth-michael:
By ~~changing `simplifyBonus` to always return the total of the deterministic terms, (and~~ moving `mod` calculation prior to any bonus determination ~~)~~, something like `system.bonuses.abilities.save | Add | @abilities.cha.mod` will actually display the appropriate bonus on the sheet. ~~All non-deterministic terms are ignored still for sheet display.~~

Update: Changing `simplifyBonus` is actually not necessary for the _main_ functionality of this change, which is to have ability mod bonuses show up on the sheet. It may still be a worthwhile change, but not as part of this PR.

### DIFF
--- a/module/data/actor/templates/common.mjs
+++ b/module/data/actor/templates/common.mjs
@@ -130,13 +130,14 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
   prepareAbilities({ rollData={}, originalSaves }={}) {
     const flags = this.parent.flags.dnd5e ?? {};
     const { prof = 0, ac } = this.attributes ?? {};
+    for ( const abl of Object.values(this.abilities) ) {
+      if ( flags.diamondSoul ) abl.proficient = 1;  // Diamond Soul is proficient in all saves
+      abl.mod = Math.floor((abl.value - 10) / 2);
+    }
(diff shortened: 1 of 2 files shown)


module/data/actor/templates/common.mjs
```diff
@@ -130,13 +130,14 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
   prepareAbilities({ rollData={}, originalSaves }={}) {
     const flags = this.parent.flags.dnd5e ?? {};
     const { prof = 0, ac } = this.attributes ?? {};
+    for ( const abl of Object.values(this.abilities) ) {
+      if ( flags.diamondSoul ) abl.proficient = 1;  // Diamond Soul is proficient in all saves
+      abl.mod = Math.floor((abl.value - 10) / 2);
+    }
```
arbron, line 136: Since `diamondSoul` isn't actually necessary for this, we can just leave it down in the main loop and simplify this to a single line:

```suggestion
    Object.values(this.abilities).forEach(a => a.mod = Math.floor((a.value - 10) / 2));
```

arbron (changes requested):

arbron (approved):

PR #4933, opened 2024-12-27
Avoid crash when getting scene targets when canvas is disabled

TheGiddyLimit:
Repro:

- Disable canvas in Foundry settings
- Use a "save" activity from an actor's sheet
- Click the "DC \<X\> \<Y\> Saving Throw" button in the resulting chat card

### DIFF
--- a/module/utils.mjs
+++ b/module/utils.mjs
@@ -492,7 +492,7 @@ export function getTargetDescriptors() {
  * @returns {Token5e[]}
  */
 export function getSceneTargets() {
-  let targets = canvas.tokens.controlled.filter(t => t.actor);
+  let targets = canvas.tokens?.controlled.filter(t => t.actor) ?? [];
   if ( !targets.length && game.user.character ) targets = game.user.character.getActiveTokens();
   return targets;
 }


module/utils.mjs
```diff
@@ -492,8 +492,8 @@ export function getTargetDescriptors() {
  * @returns {Token5e[]}
  */
 export function getSceneTargets() {
-  let targets = canvas.tokens.controlled.filter(t => t.actor);
-  if ( !targets.length && game.user.character ) targets = game.user.character.getActiveTokens();
+  let targets = canvas.tokens?.controlled?.filter(t => t.actor);
+  if ( !targets?.length && game.user.character ) targets = game.user.character.getActiveTokens();
```
arbron, line 496: This might result in this function not returning an array if the current user doesn't have an assigned token, so we should tweak this to ensure an array is always returned:

```suggestion
  let targets = canvas.tokens?.controlled.filter(t => t.actor) ?? [];
  if ( !targets.length && game.user.character ) targets = game.user.character.getActiveTokens();
```

Was there a situation where `canvas.tokens` was defined but `canvas.tokens.controlled` was `undefined`? If not, then we can ignore the second question mark.

arbron (changes requested):

module/utils.mjs
TheGiddyLimit: Nah, just sloppy on my part. Suggestions pushed!

arbron (approved):

PR #4937, opened 2024-12-28
[#3695] Allow feats to define the abilities they can improve

### LINKED ISSUE #3695: [Feature Requests] Support for Half Feats in Advancements/Level Up
Currently, when a character levels up and reaches a level where they gain an ASI, we can only choose the ASI or the Feat. There are a lot of feats that offer a +1 to an ability score along with whatever the feat offers. Examples include Athlete, Fey Touched, Shadow Touched, Resilient. 

Maybe there could be a check box option for ' Is Half Feat' that can be checked that allows both the ASI to be chosen and a feat to be dropped in.

The benefit of having something like this is so we don't have to manually update an ability score on the character sheet or have to rely on active effects to simulate this.

arbron:
Adds the ability for feats to define a set of abilities that can be improved as part of taking the feat, as well as a maximum for that improvement regardless of the maximum set on the actor.

When a feat is selected as part of ability score improvement and an ability is specified, the player will now have the option of selecting an ability to improve according to what the feat specifies.

Feat ASI Support

This integrates nicely with the existing advancement workflow, but it has the limitation that these ability score increases through feats are only available through the normal advancement. If a feat is granted to a player outside of the normal leveling process (such as an epic boon feat at level 20) then they will have to apply the ability score increase manually.

Closes #3695

### DIFF
--- a/module/applications/advancement/ability-score-improvement-flow.mjs
+++ b/module/applications/advancement/ability-score-improvement-flow.mjs
@@ -43,27 +43,33 @@ export default class AbilityScoreImprovementFlow extends AdvancementFlow {
 
   /** @inheritDoc */
   async getData() {
+    const featASI = this.feat?.system.asi;
+    const featAbilityCount = featASI?.abilities.size ?? 0;
+    const canImprove = ability => this.advancement.canImprove(ability)
+      && this.feat ? featASI.abilities.has(ability) : !this.advancement.configuration.locked.has(ability);
     const points = {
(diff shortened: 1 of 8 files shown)


module/applications/advancement/ability-score-improvement-flow.mjs
```diff
@@ -178,6 +182,7 @@ export default class AbilityScoreImprovementFlow extends AdvancementFlow {
    */
   async _onItemDelete(event) {
     event.preventDefault();
+    this.assignemnts = {};
```
krbz999, line 185:
```suggestion
    this.assignments = {};
```

PR #4941, opened 2024-12-29
Minor wiki updates

### LINKED ISSUE #4811: Wiki legend and system.attributes.hp.bonuses.level entry says it can be populated with any dice formula, it cannot
legend says 

> [formula] - When formula is mentioned in this document it means this value can be populated with any dice formula. For example, Bless adds several effects with the Effect Value of 1d4. These fields always allow for the use of roll data.

but this entry does not accept dice formula

https://github.com/foundryvtt/dnd5e/wiki/Active-Effect-Guide#bonus-hp-for-each-character-level

krbz999:
Some minor fixes. Closes #4811.

### DIFF
--- a/wiki/Active-Effect-Guide.md
+++ b/wiki/Active-Effect-Guide.md
@@ -4,7 +4,7 @@ This document only covers Active Effects available to the Core dnd5e System.
 
 # Legend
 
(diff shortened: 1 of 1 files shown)


arbron (approved):

arbron:
Poke that original issue

Issue #4604, opened 2024-10-26
Simplify spellcasting CONFIG data and separate spell preparation modes from spell preparation states


krbz999:
Currently there exist a lot of spellcasting progression related data in the configuration object.
- `DND5E.SPELL_SLOT_TABLE` enumerates the progression of leveled spell slots.
- `DND5E.pactCastingProgression` enumerates the progression of pact slots.
- `DND5E.spellPreparationModes` enumerates "ways a spell can be prepared", yet this includes "always" which causes an issue for non-leveled slots; it is not currently possible to create, say, a spellcasting type that uses pact magic-esque slots but is able to both toggle prepared (possibly via a property in this object) and "always prepared"; this will shove a a spell into a leveled section.
- `DND5E.spellcastingTypes` enumerates the two standard types of slots (leveled and pact), but also the "subtypes" for leveled (different rates of progression), but then also whether the slots are returned on a short rest. *This* property is found in `DND5E.restTypes`, too.
- Then again there is `DND5E.spellProgression`, which again lists all the progressions in `DND5E.spellcastingTypes` but is just an enumeration of strings.
- And then finally we have `DND5E.spellScalingModes` which are the ways a spell can scale (via spell level or as a cantrip via character level).

These properties have built up over time and could do with a simplification. It seems reasonable that it _should_ be possible to completely do away with `DND5E.spellProgression` as its labels all exist within `DND5E.spellcastingTypes`; *this* configuration could have an additional property, `table`, which is a key in a new config object `DND5E.spellProgressionTables` which would have the two current tables nested within (eg: `spellProgressionTables = { leveled: { ... }, pact: { ... } }`).

(Whether this spell slot progression then has one "level" of spell slots that changes as the character levels up, like Pact Slots, or has several levels all at once, like regular spell slots, might be able to be inferred from the table.)

The `DND5E.spellPreparationModes` object can also likely be merged into `spellcastingTypes`; it can perhaps be assumed that if `table` is falsy, then the mode is a non-spell-slot progression such as innate / atwill / ritual. This seems to already be the case with a different property name (`upcast: Boolean`).

Secondly, it would be good if spell data was adjusted. Currently, as mentioned, it is always assumed that `"always"` is a type of leveled spell preparation mode. The current model is:
```
preparation: {
  mode: String, // leveled, pact, atwill, innate, ritual, any custom progressions, and "always"
  prepared: Boolean
}
```
An ideal model would be:
```
preparation: {
[...]

PR #4947, opened 2024-12-31
[#4604] Rework spellcasting configurations

krbz999:
- Moves relevant spellcasting configuration into `DND5E.spellcasting` and `DND5E.spellcastingTables`.
- The `prepared` mode is now just `spell` (this simplified a lot of code throughout to just have it be identical to the name of the object within `Actor#system#spells`).
- The `always` preparation mode is no more, and is instead part of `SpellData#preparation.prepared`. This property is no longer a boolean but 0, 1, or 2 (unprepared, prepared, always prepared). Migrations are added for spell items and item grant/choice advancements.
- Whether slots are recovered on SR/LR is now only in `DND5E.restTypes` instead of in two locations.

The `DND5E.spellcasting` object entirely controls the preparation and setup of a spellcasting type. Whether it grants spell slots at all (`static`, for ritual/innate/atwill), whether the slots are spread out across multiple levels (`separate`, for leveled vs pact slots), whether the spells of this mode can be prepared/always prepared (`prepares`), and whether it grants cantrips (`cantrips`).
- A spellcasting is static if `static: true` or it has no spell progression table to be found.

Each of these types also have `progression`, identical to how leveled spell progressions are set up. The `pact` progression is considered a "Full" pact caster. This will make it trivial to add for example a type of pact magic caster that gains pact slots at a higher or lower rate. (This has no repercussions for existing warlocks.)
- Caveat: Each of these nested keys within `progression` must be unique across all the spellcastings.

The `DND5E.spellcastingTables` object different from how we usually evaluate the current amount of slots. Rather than describing the amount of slots at each level (and level of said slots), it describes the increase of slots (or level of slots) at each level. A method, `ProgressionSpellcasting#calculateSlots`, returns the total configuration of slots at a given level. This makes it easier to make small adjustments such as for example adding a single 1st-level spell slot at 9th level; you would only need to add `1: 1` at level 9 rather than editing each level from 9 and onwards.

The new `DND5E.spellPreparationStates` contains the labels and values of the preparation states (0: unprepared; 1: prepared; 2: always prepared). This is localized and used to create the select options. In item grant/choice advancements, "prepared" is omitted, allowing only for setting a granted spell as Unprepared or Always Prepared.

A handful of the static methods regarding spell slot configuration in the `Actor5e` class have been shuffled around a bit to generalize them for the purpose of this configuration. The hook(s) of course still exist and are technically breaking but only minorly so. There is migration in place during the `i18nInit` hook to mutate existing spellcastings added by modules. This has been tested thoroughly with the Apothecary spellcasting added by the "Sebastian Crowe's Guide to Drakkenheim" premium module.

Screenshots of various applications with changes:

image

image

image

image

TODO:
- [x] Add shims? We have not usually added shims for CONFIG objects.
- [x] Add deprecation warnings for the static `Actor5e` methods that were renamed.
- [x] Consider changing `static` and `separate` to a single `type` (string)  property to possible help with adding a "spell points" configuration at a later point.
- [x] Initialize the different objects in `DND5E.spellcasting` (and the object itself) as custom data models? There are a lot of various checks throughout the code, it might reduce all this further to encapsulate it in a single `class`.
- [x] Fix everything sheet-related since rebase borked that.
- [ ] Fix `_prepareSpellbook` after rebase.
- [ ] Fix sorting on sheet.


module/data/item/spell.mjs
```diff
@@ -212,6 +242,8 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
     }, { all: [], vsm: [], tags: [] });
     labels.components.vsm = game.i18n.getListFormatter({ style: "narrow" }).format(labels.components.vsm);
 
+    this.preparation.static = (CONFIG.DND5E.spellcasting[this.preparation.mode] ?? { static: true }).static === true;
```
krbz999, line 245: A spell belongs to a "static" preparation mode if said prep mode does not exist, or is `static: true`.

module/documents/item.mjs
```diff
@@ -443,26 +443,11 @@ export default class Item5e extends SystemDocumentMixin(Item) {
 
   /* -------------------------------------------- */
 
-  /**
-   * Does this item scale with any kind of consumption?
-   * @type {string|null}
-   */
-  get usageScaling() {
```
krbz999, line 450: This wasn't used anywhere.

module/documents/actor/actor.mjs
```diff
@@ -799,38 +788,41 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
    * @param {object} progression   Spellcasting progression data.
+   * @param {string} key           The internal key for these spell slots on the actor.
    */
-  static preparePactSlots(spells, actor, progression) {
-    this.prepareAltSlots(spells, actor, progression, "pact", CONFIG.DND5E.pactCastingProgression);
-    spells.pact.label = game.i18n.localize("DND5E.PactMagic");
+  static prepareLeveledSlots(spells, actor, progression, key) {
+    const slots = CONFIG.DND5E.spellcasting.calculateSlots(key, progression[key]);
+    for ( const level of Array.fromRange(Object.keys(CONFIG.DND5E.spellLevels).length - 1, 1) ) {
+      const slot = spells[`${key}${level}`] ??= { value: 0 };
```
krbz999, line 820: During data prep, if the spellcasting is `separate`, then objects of with a key of `<name><level>` will be created (such as "spell1" through "spell9"), otherwise a single object is created (eg "pact").

This is entirely dynamic, and can be tested by setting `pact.table = "spell"` and `pact.separate = true`, which results in `pact1` through `pact9`.

image

module/documents/actor/actor.mjs
```diff
@@ -799,38 +788,41 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+      slot.label = CONFIG.DND5E.spellLevels[level];
```
krbz999, line 821: This needs a better label. It is currently used only in Favorites, I believe.

dnd5e.mjs
```diff
@@ -460,8 +467,73 @@ Hooks.once("i18nInit", () => {
...
+
+  // Fix progression missing table.
+  if ( (config.type !== "static") && !config.table ) {
+    foundry.utils.logCompatibilityWarning(
+      "A registered spellcasting with a progression must define the 'table' property.",
+      { from: "5.0", until: "5.4", once: true }
+    );
+
+    const actor = new CONFIG.Actor.documentClass({ name: "Deprecation", _id: "dnd5eDeprecation", type: "character" });
+    Hooks.call(`dnd5e.prepare${key.capitalize()}Slots`, actor.system.spells, actor, { [key]: 0 });
```
krbz999, line 515: Happy to provide a code example in use in a premium module this was tested with. The instantiated actor is only to grab the table that is passed through - it likely does not need to be an actor, though passing in the "correct" parameters is safer.

dnd5e.mjs
```diff
@@ -460,8 +467,73 @@ Hooks.once("i18nInit", () => {
...
+    if ( actor._table ) {
+      CONFIG.DND5E.spellcastingTables[key] = actor._table;
+      config.table = key;
+      config._deprecated = true;
```
krbz999, line 519: This key is used only to know later whether the table is using the new "additive" method or the old "override" method.

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -542,78 +568,47 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
+      if ( !(mode in CONFIG.DND5E.spellcasting) ) mode = "atwill";
+
+      const config = CONFIG.DND5E.spellcasting[mode];
+      const spellLevel = spell.system.level || 0;
+      let sectionKey = config.spellSlotKey(spellLevel);
 
       // Spells from items
       if ( spell.getFlag("dnd5e", "cachedFor") ) {
-        s = "item";
+        sectionKey = "item";
```
krbz999, line 597: Spells from Cast activities go in the `"item"` section [...]

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -542,78 +568,47 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
+        if ( !spellbook[sectionKey] ) {
+          registerSection(sectionKey);
+          spellbook[sectionKey].order = 10000;
         }
       }
 
       // Sections for higher-level spells which the caster "should not" have, but spell items exist for
-      else if ( !spellbook[s] ) {
-        registerSection(sl, s, CONFIG.DND5E.spellLevels[s], {levels: levels[sl]});
+      else if ( !spellbook[sectionKey] ) {
```
krbz999, line 606: [... ] otherwise they're tossed into a section corresponding to a key in `actor.system.spells` or one of the "static" spellcastings. Special handling for `"spell0"` which is where all cantrips go, except for static ones (as is current).

module/applications/actor/config/spell-slots-config.mjs
```diff
@@ -39,21 +39,31 @@ export default class SpellSlotsConfig extends BaseConfigSheet {
...
-    for ( const k of Object.keys(CONFIG.DND5E.spellcastingTypes) ) {
-      const hasSpell = this.document.items.some(i => i.type === "spell" && i.system.preparation.mode === k);
-      if ( parseInt(spells[k]?.level) || hasSpell ) context.overrides.push({
-        label: CONFIG.DND5E.spellPreparationModes[k].label,
-        value: source[k]?.override,
-        name: `system.spells.${k}.override`,
-        placeholder: spells[k]?.max ?? 0
-      });
+    const maxSpellLevel = Object.keys(CONFIG.DND5E.spellLevels).length - 1;
+    context.overrides = [];
```
krbz999, line 43: Possibly controversial change here is separating all the "slot types" by a rule.

krbz999:
Outside force pushing has made this PR unsalvageable. I tried. 🤷

PR #4956, opened 2025-01-02
[#4955] Add option to apply advancement steps automatically

### LINKED ISSUE #4955: Add option to apply advancement steps that don't require user intervention automatically
Add an option to `AdvancementManager` that would skip the UI for steps that don't require user intervention to apply. This requires opt-in from each advancement type in order to indicate whether automatic application is possible and to provide the data necessary for the `apply` method.

arbron:
Adds a `automaticApplication` option to the `AdvancementManager` that tells it to no display the UI for advancement steps if they can be applied without user intervention.

To support automatic application the `Advancement` API has a new method: `automaticApplicationValue`. This method takes the level to be applied and returns the data object to pass to the `apply` method if automatic application is possible, otherwise it returns `false`. Since this method returns `false` by default, only advancement types that opt-in to it will get the new behavior.

Closes #4955

### DIFF
--- a/module/documents/advancement/trait.mjs
+++ b/module/documents/advancement/trait.mjs
@@ -120,6 +120,16 @@ export default class TraitAdvancement extends Advancement {
 
   /* -------------------------------------------- */
 
+  /** @override */
+  automaticApplicationValue(level) {
+    // TODO: Ideally this would be able to detect situations where choices are automatically fulfilled because
+    // they only have one valid option, but that is an async process and cannot be called from within `render`
+    if ( this.configuration.choices.length || this.configuration.allowReplacements ) return false;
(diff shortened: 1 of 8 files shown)


module/documents/advancement/trait.mjs
```diff
@@ -120,6 +120,16 @@ export default class TraitAdvancement extends Advancement {
 
   /* -------------------------------------------- */
 
+  /** @override */
+  automaticApplicationValue(level) {
+    // TODO: Ideally this would be able to detect situations where choices are automatically fulfilled because
+    // they only have one valid option, but that is an async process and cannot be called from within `render`
```
Fyorl, line 126: I guess we can support this when we move to App V2 then,

Fyorl (approved):

PR #4959, opened 2025-01-03
[#4958] Allow features to be repeatable in item choice

### LINKED ISSUE #4958: Add repeatable data to features, allow selecting repeatable items multiple times in `ItemChoice` advancement
Some Feats (Ability Score Improvement, Magic Initiate, Skilled, etc.) and Eldritch Invocations (Agonizing Blast, Eldritch Spear, etc.) can be taken more than once, but currently there is no way to indicate this on an item and the Choose Items advancement doesn't support this.

arbron:
Adds "Repeatable" data to the prerequisite data on `feat` items. Repeatable features can be selected more than once inside an item choice advancement.

Closes #4958

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1909,6 +1909,21 @@
 "DND5E.FavoriteRemove": "Remove Favorite",
 "DND5E.Favorites": "Favorites",
 "DND5E.FeatureActionRecharge": "Action Recharge",
+
+"DND5E.FEATURE": {
+  "FIELDS": {
+    "properties": {
+      "label": "Feature Properties"
+    },
+    "requirements": {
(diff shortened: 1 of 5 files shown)


Fyorl (approved):
I did some testing with a repeatable Agonizing Blast and I couldn't break it. The only thing I noticed was that you can't, for example, select Agonizing Blast twice at the same level, which is an understandable limitation at this time.

arbron:
> I did some testing with a repeatable Agonizing Blast and I couldn't break it. The only thing I noticed was that you can't, for example, select Agonizing Blast twice at the same level, which is an understandable limitation at this time.

I'll add a todo to look into that one in the future.

PR #4969, opened 2025-01-06
[#3695] Add ASI Advancement support to feature items

### LINKED ISSUE #3695: [Feature Requests] Support for Half Feats in Advancements/Level Up

arbron:
Adds advancement data to `feature` items and allows them to use the `ASIAdvancement`. Holding off on more advancement types until more testing has been performed to ensure granting items will work properly.

Advancement on Feat

Closes #3695

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -4150,7 +4150,7 @@ const _ALL_ITEM_TYPES = ["background", "class", "race", "subclass"];
 DND5E.advancementTypes = {
   AbilityScoreImprovement: {
     documentClass: advancement.AbilityScoreImprovementAdvancement,
-    validItemTypes: new Set(["background", "class", "race"])
+    validItemTypes: new Set(["background", "class", "race", "feat"])
   },
   HitPoints: {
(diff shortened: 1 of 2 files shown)


Fyorl (approved):

sneat:
> Holding off on more advancement types until more testing has been performed to ensure granting items will work properly.

@arbron is there anywhere that I can see the planned timeline for this? One of the third-party modules that I help with has the concept of "sub-races", which we are needing to implement as "Feature" items as it's the most appropriate of the options. The sub-races give additional features at different levels, so support the "grant items" portion would solve everything for us.

PR #4970, opened 2025-01-06
[#3695] Adjust ASI advancement in modern rules to be feat-first

arbron:
In the 2024 rules ability score improvement is presented as: "Take the Ability Score Improvmeent Feat or another feat". This adjusts the interface of the ASI advancement when using the modern rules to hide the ASI interface unless the "Ability Score Improvement Feat" is checked. This change will improve the experience of half-feats because they won't be presented with two ASI interfaces in a row.

ASI 2024

Also ignores the "Allow Feats" setting when using the 2024 rules because they are no longer an optional rule.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -248,6 +248,7 @@
     },
     "Details": "Improvement Details",
     "Feat": {
+      "ASIName": "Ability Score Improvement Feat",
       "Hint": "Drop a feat here to choose one instead of an ability score improvement."
     },
     "FIELDS": {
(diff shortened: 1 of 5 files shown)


Fyorl (approved):
Very slight edge case where you can choose the ASI checkbox at level-up, and add some points to an ability. Then if you later modify choices and uncheck the ASI checkbox but *don't* drag any feat, and continue instead, then the ASI does not get un-applied.

That's a very degenerative case though and absolutely not something to hold up the PR over.

PR #4975, opened 2025-01-07
[#3788] Add enhanced rest chat message with recovery info

### LINKED ISSUE #3788: Display more information in chat when completing a rest
Currently the rest chat messages only display the number of hit points and hit dice recovered/spent during the rest. It would be very nice to have more information on things like number of spell slots and item uses recovered. Ideally the card would show the state of the actor before the rest and what changed.

arbron:
Adds a new `rest` chat message type that matches the `turn` type and displays more information on the resources recovered during the rest including item uses and spell slots.

Rest Chat Card

Expands `getHumanReadableAttributeLabel` with support for fetching labels for activity & item attributes.

Closes #3788

### DIFF
--- a/module/utils.mjs
+++ b/module/utils.mjs
@@ -1075,7 +1075,8 @@ export function getHumanReadableAttributeLabel(attr, { actor, item }={}) {
   let name;
   let type = "actor";
 
-  const getSchemaLabel = (attr, type) => {
+  const getSchemaLabel = (attr, type, doc) => {
+    if ( doc ) return doc.system.schema.getField(attr)?.label;
     for ( const model of Object.values(CONFIG[type].dataModels) ) {
       const field = model.schema.getField(attr);
       if ( field ) return field.label;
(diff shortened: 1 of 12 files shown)


module/utils.mjs
```diff
@@ -1061,18 +1070,38 @@ export function getHumanReadableAttributeLabel(attr, { actor }={}) {
...
-  let label = _attributeLabelCache.get(attr);
+  let label = item ? null : _attributeLabelCache.actor.get(attr);
   if ( label ) return label;
+  let name;
+  let type = "actor";
 
-  // Derived fields.
-  if ( attr === "attributes.init.total" ) label = "DND5E.InitiativeBonus";
-  else if ( (attr === "attributes.ac.value") || (attr === "attributes.ac.flat") ) label = "DND5E.ArmorClass";
-  else if ( attr === "attributes.spelldc" ) label = "DND5E.SpellDC";
```
Fyorl, line 1070: Still need these

module/utils.mjs
```diff
@@ -1061,18 +1070,38 @@ export function getHumanReadableAttributeLabel(attr, { actor }={}) {
...
+    for ( const model of Object.values(CONFIG[type].dataModels) ) {
+      const field = model.schema.getField(attr);
+      if ( field ) return field.label;
+    }
+  };
 
-  // Abilities.
-  else if ( attr.startsWith("abilities.") ) {
-    const [, key] = attr.split(".");
-    label = game.i18n.format("DND5E.AbilityScoreL", { ability: CONFIG.DND5E.abilities[key].label });
```
Fyorl, line 1075: I tested this change, but it's not correct:

```js
getHumanReadableAttributeLabel("abilities.dex.value", { actor: _token.actor }); // "Ability Score"
```

module/utils.mjs
```diff
@@ -1061,18 +1070,38 @@ export function getHumanReadableAttributeLabel(attr, { actor }={}) {
...
+  let label = item ? null : _attributeLabelCache.actor.get(attr);
   if ( label ) return label;
+  let name;
+  let type = "actor";
 
-  // Derived fields.
-  if ( attr === "attributes.init.total" ) label = "DND5E.InitiativeBonus";
-  else if ( (attr === "attributes.ac.value") || (attr === "attributes.ac.flat") ) label = "DND5E.ArmorClass";
-  else if ( attr === "attributes.spelldc" ) label = "DND5E.SpellDC";
+  const getSchemaLabel = (attr, type) => {
```
Fyorl, line 1078: This wasn't part of the original code, but since we sometimes have a full Actor or Item instance, we can actually look up the correct schema, rather than iterating over all of the,.

Fyorl (changes requested):

module/utils.mjs
arbron: Yeah, I think some lines might have been lost during a rebase or something while working on this, because these lines shouldn't have been removed.

Fyorl (approved):

PR #4976, opened 2025-01-08
Use config for the SensesField

### LINKED ISSUE #4667: `SensesField` should pull the properties from `CONFIG.DND5E.senses`
While attempting to add a new sense, simply adding it to the config was enough to make it appear in the Senses config, but it cannot render on the actor sheets and cannot save a value for this new sense.

It would be good if all locations (actor sheet, actor data, senses config) drew from the config object.

(This is on the 4.0 branch, don't believe there has been a change in 4.1 tho.)

Larkinabout:
Intended to allow custom senses by editing `CONFIG.DND5E.senses`.

image


module/data/shared/senses-field.mjs
```diff
@@ -6,11 +6,12 @@ const { NumberField, StringField } = foundry.data.fields;
 export default class SensesField extends foundry.data.fields.SchemaField {
   constructor(fields={}, options={}) {
     const numberConfig = { required: true, nullable: true, integer: true, min: 0, initial: null };
+    const senses = Object.entries(CONFIG.DND5E.senses).reduce((acc, [k, label]) => {
```
krbz999, line 9: This should probably be a `MappingField` if it should be configurable. Otherwise data is lost as soon as a module is disabled, or if a module removes any of the default senses.

Larkinabout: I did do some tests and it seemed to retain the values set against an actor for the core senses, and even custom senses, when removed and re-added. But, I'll admit I don't have much comprehension of it beyond where it's defining the senses initially. I'll see if I can figure out what a `MappingField` will do.

Larkinabout: Struggling a bit to understand how `MappingField` is intended to work in this context, with `units` and `special` being handled alongside the config properties. Creating a `MappingField` within fields for the config properties makes sense, as those are actual senses compared to units and special. However, it looks like it'll complicate handling elsewhere (which might be necessary).

That said, I may be misunderstanding the suggested application.

Is this what you were envisioning? Or was your suggestion something else?

image

arbron: The way I handle it in Black Flag is have something like:

```javascript
{
  senses: {
    types: MappingField,
    special: StringField,
    units: StringField
  }
}
```

This will of course require a migration and changes anywhere in the system where it accesses the sense values (part of the reason I haven't gotten around to it yet, big project).

Larkinabout: At least I wasn't totally off the mark! I'll see if I can slowly chip away at it.

Larkinabout: With this part: https://github.com/foundryvtt/dnd5e/blob/99000e24daf04d0594f9a65c4c2a73f3f398dcae/module/applications/shared/movement-senses-config.mjs#L82

How would I build it out using the `MappingField`' so that the `NumberField`s are provided to the template with `name` and `label`? Is there an existing function to do it? Otherwise, what's the correct way to provide a mapping field for the template?

Additionally, I'm guessing it'd be better to convert movement in the same way at the same time as senses?

Larkinabout:
I believe the only thing left to do is resolve `MovementSensesConfig._preparePartContext`, though I'm still not clear on how to do it. I'll keep trying to figure it out, but would be grateful for anyone's insight to speed me along.

Larkinabout:
Note to remind that we may wish to also move `group` actor movement types into a `types` object for consistency, but for the moment I've handled it in `movement-senses-config.mjs`.

Also, would be useful to allow `group` movement to also be configurable. `CONFIG.DND5E.vehicleTypes` could be viable, but it includes `space`, which would need to be excluded by default. This would cause an issue if a user actually wants to add space as a movement type.

arbron:
@Larkinabout It would be good to take the compendium changes out of this PR, we can handle them later and we'll want to make sure that the migrations in place are enough to keep compendium content working even if it doesn't have the new data layout (plus it makes it easier to review)

krbz999:
This probably also needs migrations for ActiveEffects. All AEs that modify `system.attributes.senses.*` should point at something different, or we just support the same deprecated property but point it elsewhere in the AE document class.

module/migration.mjs
```diff
@@ -843,6 +848,30 @@ function _migrateEffectArmorClass(effect, updateData) {
...
+/**
+ * Change active effects that target movement or senses.
+ * @param {object} effect      Effect data to migrate.
+ * @param {object} updateData  Existing update to expand upon.
+ * @returns {object}           The updateData to apply.
+ */
+function _migrateEffectMovementSenses(effect, updateData) {
+  let containsUpdates = false;
+  const changes = (effect.changes || []).map(c => {
+    if ((c.key.startsWith("system.attributes.movement.") || c.key.startsWith("system.attributes.senses."))
```
arbron, line 860: This looks like it will move any changes targeting `attributes.movement.units`, `attributes.movement.hover`, or `attributes.senses.special` into the types object.

module/applications/actor/base-sheet.mjs
```diff
@@ -249,15 +249,12 @@ export default class ActorSheet5e extends ActorSheetMixin(ActorSheet) {
...
-    let speeds = [
-      [movement.burrow, `${game.i18n.localize("DND5E.MovementBurrow")} ${movement.burrow}`],
-      [movement.climb, `${game.i18n.localize("DND5E.MovementClimb")} ${movement.climb}`],
-      [movement.fly, `${game.i18n.localize("DND5E.MovementFly")} ${movement.fly}${movement.hover ? ` (${game.i18n.localize("DND5E.MovementHover")})` : ""}`],
-      [movement.swim, `${game.i18n.localize("DND5E.MovementSwim")} ${movement.swim}`]
-    ];
-    if ( largestPrimary ) {
-      speeds.push([movement.walk, `${game.i18n.localize("DND5E.MovementWalk")} ${movement.walk}`]);
-    }
+    let speeds = Object.entries(CONFIG.DND5E.movementTypes).filter(([k]) => !(!largestPrimary && k === "walk"))
```
arbron, line 252:
```suggestion
    let speeds = Object.entries(CONFIG.DND5E.movementTypes)
      .filter(([k]) => largestPrimary || (k !== "walk"))
```

module/applications/shared/movement-senses-config.mjs
```diff
@@ -74,15 +78,20 @@ export default class MovementSensesConfig extends BaseConfigSheet {
...
-    }));
+    context.types = this.types.map(([k, label]) => {
+      return {
+        field: isGroup ? context.fields[k] : context.fields.types.model,
+        key: `system.${this.keyPath}.${isGroup ? "" : "types."}${k}`,
+        label,
+        placeholder: (isGroup ? placeholderData?.[k] : placeholderData?.types?.[k]) ?? "",
+        value: isGroup ? context.data[k] : context.data.types?.[k]
+      };
+    });
```
arbron, line 94:
```suggestion
    context.types = this.types.map(([k, label]) => ({
      field: isGroup ? context.fields[k] : context.fields.types.model,
      key: `system.${this.keyPath}.${isGroup ? "" : "types."}${k}`,
      label,
      placeholder: (isGroup ? placeholderData?.[k] : placeholderData?.types?.[k]) ?? "",
      value: isGroup ? context.data[k] : context.data.types?.[k]
    }));
```

arbron (changes requested):

module/data/actor/templates/attributes.mjs
```diff
@@ -121,6 +131,29 @@ export default class AttributesFields {
...
+   * @internal
+   */
+  static _migrateMovementSenses(source) {
+    const movementSenses = [
+      { key: "movement", config: CONFIG.DND5E.movementTypes },
+      { key: "senses", config: CONFIG.DND5E.senses }
+    ];
+
+    for (const { key, config } of movementSenses) {
+      const attr = source[key];
```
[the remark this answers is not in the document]
arbron: Because it is called `AttributesField._migrate(source.attributes)`, and if you call `actor.update({ "details.cr": 5 });` then `source` will just equal `{ details: { cr: 5 } }`, so migrate will get `undefined`.

Returning early would probably be okay (in which case you could adjust the JSDocs back because `source` would always exist).

arbron:
@Larkinabout Please give a comment over on https://github.com/foundryvtt/dnd5e/issues/4667 so I can assign it to you

Fyorl:
This work should be done with the vision/detection mode work in mind and in concert with it.

arbron:
Hey @Larkinabout, we're thinking of adding this to the 6.0 release. Do you have any interest in attempting to rebase this PR onto the current dev branch and continuing this work or should I take it over? Since all of the senses part of the PR has been merged this would just be handling the movement side of things.

Larkinabout:
> Hey @Larkinabout, we're thinking of adding this to the 6.0 release. Do you have any interest in attempting to rebase this PR onto the current dev branch and continuing this work or should I take it over? Since all of the senses part of the PR has been merged this would just be handling the movement side of things.

I'm happy to leave it in your more capable hands!

arbron:
Closed in favor of https://github.com/foundryvtt/dnd5e/pull/7201

PR #4980, opened 2025-01-09
Update enricher documentation to match changes in 4.2

arbron:
- Add the attack enricher
- Multiple proficiencies in check enricher
- Multiple damage types & formulas in damage enricher
- Mulitple abilities in save enricher
- Activity-linked check, damage, and save enrichers
- Activity lookup

### DIFF
--- a/wiki/Enrichers.md
+++ b/wiki/Enrichers.md
@@ -1,8 +1,8 @@
-![Up to date as of 3.2.0](https://img.shields.io/static/v1?label=dnd5e&message=3.2.0&color=informational)
+![Up to date as of 4.2.0](https://img.shields.io/static/v1?label=dnd5e&message=4.2.0&color=informational)
 
 The dnd5e system adds a number of useful enrichers that can be used within journals or in item or actor descriptions. These enrichers will generate text based on the standard formatting used throughout 5e releases and provide rolls and related behavior that properly hooks into the system.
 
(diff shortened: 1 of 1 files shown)


wiki/Enrichers.md
```diff
@@ -38,6 +87,8 @@ It is very common for a feature description or journal entry to include a call f
 
 The format also supports full-length ability names (e.g. `[[/check dexterity]]`) and prefixed terms (e.g. `[[/check ability=dex]]`) for greater clarity.
 
+Multiple skills or tools can be provided to the enricher by separating them by slashes when used with the explicit key (e.g. `[[/check skill=acr/ath]]`) or by specifying each skill separately when used without the key (e.g. `[[/check acr ath]]`). If not ability is provided, then each skill or tool will use its default ability. If an ability is provided, then all skills and tools will use that ability. *Providing multiple abilities is not supported.*
```
krbz999, line 90:
```suggestion
Multiple skills or tools can be provided to the enricher by separating them by slashes when used with the explicit key (e.g. `[[/check skill=acr/ath]]`) or by specifying each skill separately when used without the key (e.g. `[[/check acr ath]]`). If no ability is provided, then each skill or tool will use its default ability. If an ability is provided, then all skills and tools will use that ability. *Providing multiple abilities is not supported.*
```

Fyorl (approved):

Issue #4935, opened 2024-12-28
Add easy way to copy scale value from non-editable class/subclass/etc. sheet


JPMeehan:
As a content creator, when I am building a subclass for an existing class I occasionally need to reference a scale value on the base class - the most common case being Monk die and Bardic Inspiration, but others sometimes pop up as well (e.g. Barbarian Rage damage). Currently, copying the scale value's roll data string requires opening the advancement, which is not possible if the sheet is not editable (e.g. it's in the locked SRD classes compendium). 

There's already a three dot menu but clicking doesn't do anything currently if the document isn't editable; I think that would be the best place to add some more copy-to-clipboard buttons.

arbron:
I would also make sense to be able to open the advancement configuration app even when the parent item is locked, which can be done right now with activities.

PR #4983, opened 2025-01-10
[#4935] Show advancement context menu when locked

arbron:
Moves the advancement context menu configuration from the item sheet into the `Advancement` class to mimic the activity setup. In the process adds a "View Advancement" button for when the item is not editable so the advancement sheet can still be viewed.


module/documents/mixins/pseudo-document.mjs
```diff
@@ -132,14 +132,17 @@ export default function PseudoDocumentMixin(Base) {
...
+     * Lazily obtain a Application instance used to configure this PseudoDocument, or null if no sheet is available.
+     * @type {Application|ApplicationV2|null}
      */
     get sheet() {
       const cls = this.constructor.metadata.sheetClass ?? this.constructor.metadata.apps?.config;
       if ( !cls ) return null;
       if ( !this.constructor._sheets.has(this.uuid) ) {
-        this.constructor._sheets.set(this.uuid, new cls({ document: this }));
+        let sheet;
+        if ( cls.prototype instanceof Application ) sheet = new cls(this);
```
Fyorl, line 143:
```suggestion
        if ( Application.isPrototypeOf(cls) ) sheet = new cls(this);
```

Fyorl (approved):

PR #4985, opened 2025-01-11
[#2678] Fix "Reveal" button in secret blocks in multi-editor

### LINKED ISSUE #2678: Secret block "Reveal" button is non-functional in MultiEditor
Clicking Reveal silently fails in Item Description with multieditor

[RevealYourself.webm](https://github.com/foundryvtt/dnd5e/assets/86370342/87d2dc40-9b29-4668-8731-c0aae8d3aaf4)

Also, should the Reveal button display when an Item is expanded as well as in the Item Sheet?

Some tables do give Observer access to all their party's actors, so there may be an instance where a player wants to hide information about an Item or Feature from the other players, but still be able to easily reveal it later on

arbron:
Fixes the issues with reveal buttons for secret blocks for items and actors that use multiple description editors. This required overriding `_getSecretContent` and `_updateSecret` to look for `data-target` in addition to `data-edit`. The locking code was also modified to avoid disabling the secret button when the sheets are in play mode.

There is a still a bug with the PC sheet and the item sheets with only a single description (like classes), which don't retain the secret state if the play/edit toggle is changed immediately after changing the secret state. This appears to be because when it submits `FormDataExtended` is fetching old data from the editor instances on the sheet.

Closes #2678


templates/items/description.hbs
```diff
@@ -1,5 +1,7 @@
-<section class="description tab {{ tab.cssClass }} {{#if singleDescription}}single-description{{/if}} {{#if (or editingDescription (and singleDescription editable))}}editing{{/if}}"
-         data-tab="{{ tab.id }}" data-group="{{ tab.group }}">
+<section class="description tab {{ tab.cssClass }} {{#if singleDescription}}single-description{{/if}}
+         {{~#if (or editingDescription (and singleDescription editable))}}editing{{/if}}"
```
Fyorl: Won't this strip whitespace such that you end up with `class="description tab single-descriptionediting"`?

PR #4996, opened 2025-01-13
[#4929, #4931] Fix mountable data models, add cover to feat sheet

### LINKED ISSUE #4929: Cannot toggle "Crewed" on vehicle action
For example, the actions on the "Apparatus of Kwalish" vehicle actor from the '24 DMG:

[crew.webm](https://github.com/user-attachments/assets/bcebc961-9de7-455d-a377-77862f16480f)

---

Env:

- dnd5e 4.1.2
- Foundry 12.331
- modules active: `dnd-dungeon-masters-guide`

### LINKED ISSUE #4931: Cannot set "cover" value for vehicle actions
A "Cover" column is displayed under vehicle Features -> Actions, but there is no UI to set this value.

image

The `MountableTemplate` data model, which includes the `cover` field, is only mixed into weapon/equipment-type items, but neither of these allow the field to be set in the UI. Additionally, each of these have their own sheet header, neither of which have a "cover" column.

---

Exporting a working actor from dnd 3.3.1:

image

but when imported into dnd 4.1.2:

image

blanks the "Cover" display, and when exported from 4.1.2, the "cover" field is missing from the resulting JSON.

JSON exported from 3.3.1:

[fvtt-Actor-test-vehicle-PaPE3J59L6Wcq8NN.json](https://github.com/user-attachments/files/18260415/fvtt-Actor-test-vehicle-PaPE3J59L6Wcq8NN.json)

JSON exported from 4.1.2:

[fvtt-Actor-test-vehicle-T0Nk9Ntj3G34tDCY.json](https://github.com/user-attachments/files/18260416/fvtt-Actor-test-vehicle-T0Nk9Ntj3G34tDCY.json)

---

Env:

- dnd5e 4.1.2
- Foundry 12.331
- no modules active

arbron:
- Move `cover` from `MountableTemplate` to `FeatData` since feats are the only item type that display this data on vehicles
- Add "Vehicle Feature" type for features and display a cover field if this type is selected
- Remove `armor` from `MountableTemplate` and define directly on `WeaponData`, since `EquipmentData` already has armor
- Move to using localization prefixes for mountable labels
- Change cover labels shown on vehicle sheet to remove distinction between `0` and `null` values

Closes #4929
Closes #4931

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3954,6 +3954,10 @@
 "DND5E.VEHICLE": {
   "MOUNTABLE": {
     "FIELDS": {
+      "cover": {
+        "label": "Cover",
+        "hint": "Cover provided to crew operating this equipment."
+      },
       "crewed": {
         "label": "Crewed"
       },
--- a/lang/en.json
+++ b/lang/en.json
@@ -2021,6 +2021,10 @@
 
 "DND5E.FEATURE": {
   "FIELDS": {
+    "cover": {
+      "label": "Cover",
+      "hint": "Cover provided to crew by this vehicle feature."
+    },
     "properties": {
       "label": "Feature Properties"
     },
@@ -2069,6 +2073,9 @@
     "Blessing": "Blessing",
     "Charm": "Charm",
     "EpicBoon": "Epic Boon"
+  },
+  "Vehicle": {
+    "Label": "Vehicle Feature"
   }
 },
 "DND5E.Focus": {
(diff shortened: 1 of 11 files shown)


Fyorl (commented):
> * Move `cover` from `MountableTemplate` to `FeatData` since feats are the only item type that display this data on vehicles

I think this might have been more of a concession or limitation of the original sheet rather than an actual deliberate decision. Something like the Harpoon Flinger provides cover and very much looks like a weapon.

image

We'll likely improve on this in the V2 sheet so perhaps it should stay on the mountable template for now?

Vehicle Feature is generally a useful category to have regardless though.

Fyorl (approved):

Issue #4842, opened 2024-12-06
Make changes to lair behavior when using modern rules on an actor


arbron:
Modify NPCs when using the modern rules to **REDACTED**. These changes should be applied on an actor-by-actor basis depending on the rules version in the actor's source config, rather than using the global rules version setting.

PR #4998, opened 2025-01-14
[#3798, #4842] Add legendary action description to stat blocks

arbron:
Adds the legendary action description to embedded NPC stat blocks. This will automatically use the description of an item with the `legendary-actions` identifier if present, otherwise it use an auto-generated description based on the rules version.

The auto-generated description is also available from the roll data in `resources.legact.label`. A modification has been made to the lookup enricher so that if it finds a function at its target location, it will call that function with the enricher config as its paremter. This allows for substituting the actor's name: `[[lookup @resources.legact.label name=dragon]]`.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2720,7 +2720,7 @@
 "DND5E.Languages": "Languages",
 
 "DND5E.LegendaryAction": {
-  "Description": "Legendary Action Uses: {uses}. Immediately after another creature’s turn, the {name} can expend a use to take one of the following actions. The {name} regains all expended uses at the start of each of its turns.",
+  "Description": "Legendary Action Uses: {uses}.",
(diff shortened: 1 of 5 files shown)


Fyorl (approved):

guldelox:
Does this also fix the issue where any features in the Special category aren't displayed on embedded stat blocks? Legendary Actions are listed in the Special category in the actor sheet, so I wonder if this issue is now fixed.

arbron:
> Does this also fix the issue where any features in the Special category aren't displayed on embedded stat blocks? Legendary Actions are listed in the Special category in the actor sheet, so I wonder if this issue is now fixed.

Does nothing about that, but giving features that appear in the special section the "Passive Trait" property will move them to the features section and display under the "Traits" header on the stat block.

PR #5002, opened 2025-01-15
Allow embedded roll table result column to be renamed

arbron:
Adds the `resultsHeader` config option that is supported by roll tables and replaces the default "Results" header with a custom value:

```
@Embed[Compendium.dnd5e.tables.RollTable.4ryVFV5LPHPzRKxl resultsHeader="Bean Effects" rollable]
```

Screenshot 2025-01-15 at 13 02 16


krbz999:
If the system is modifying the core enricher, we should add information to the enricher wiki page as well.

Fyorl:
Might be better as a core feature. Otherwise we're stuck maintaining an alternate syntax if core goes a different direction.

arbron:
Closing because this is now handled by core

PR #5004, opened 2025-01-15
[#5003] Add ability attack, spell attack, & spell ability labels

### LINKED ISSUE #5003: Add pre-calculated spell attack & ability labels for NPCs usable in lookups
To make the spellcasting block of NPCs easier to set up for each actor, the system could use a pre-calculated spell attack bonus (prof + mod) and the full name for the spellcasting ability (currently only the ability key is easily available in roll data).

arbron:
Adds `attack` to each ability which is the default attack to hit using that ability (e.g. prof + mod).

Also adds the spell attack and ability label to actor attributes inside a new `attributes.spells` object. This moves the old `spelldc` and `spellmod` into that object with deprecation warnings.

Closes #5003

### DIFF
--- a/module/applications/actor/base-sheet.mjs
+++ b/module/applications/actor/base-sheet.mjs
@@ -1116,7 +1116,7 @@ export default class ActorSheet5e extends ActorSheetMixin(ActorSheet) {
       if ( modes[preparationMode]?.cantrips ) {
         prep.mode = "prepared";
       } else if ( !preparationMode ) {
-        const isCaster = this.document.system.details.spellLevel || progs.size;
+        const isCaster = this.document.system.attributes.spell?.level || progs.size;
         prep.mode = isCaster ? "prepared" : "innate";
       } else {
         prep.mode = preparationMode;
@@ -1127,7 +1127,7 @@ export default class ActorSheet5e extends ActorSheetMixin(ActorSheet) {
     // Case 2: Drop a leveled spell in a section without a mode.
     else if ( (level === "0") || !preparationMode ) {
       if ( this.document.type === "npc" ) {
-        prep.mode = this.document.system.details.spellLevel ? "prepared" : "innate";
+        prep.mode = this.document.system.attributes.spell.level ? "prepared" : "innate";
       } else {
         const m = progs.has("leveled") ? "prepared" : (progs.first() ?? "innate");
         prep.mode = progs.has(prep.mode) ? prep.mode : m;
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -3524,8 +3524,8 @@ preLocalize("cover");
  */
 DND5E.trackableAttributes = [
   "attributes.ac.value", "attributes.init.bonus", "attributes.movement", "attributes.senses",
-  "attributes.spell.attack", "attributes.spell.dc", "attributes.spellLevel", "details.cr",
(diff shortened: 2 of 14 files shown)


module/data/actor/templates/common.mjs
```diff
@@ -149,6 +149,7 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
 
       abl.save = abl.mod + abl.saveBonus;
       if ( Number.isNumeric(abl.saveProf.term) ) abl.save += abl.saveProf.flat;
+      abl.attack = abl.mod + prof;
```
krbz999, line 152: If this is meant only for display purposes, should it not rather be a string? A numeric value won't support proficiency dice.

module/config.mjs
```diff
@@ -3470,9 +3470,9 @@ preLocalize("cover");
  * @deprecated since v10
  */
 DND5E.trackableAttributes = [
-  "attributes.ac.value", "attributes.init.bonus", "attributes.movement", "attributes.senses", "attributes.spelldc",
-  "attributes.spellLevel", "details.cr", "details.spellLevel", "details.xp.value", "skills.*.passive",
-  "abilities.*.value"
+  "attributes.ac.value", "attributes.init.bonus", "attributes.movement", "attributes.senses",
+  "attributes.spell.attack", "attributes.spell.dc", "attributes.spellLevel", "details.cr",
+  "details.spellLevel", "details.xp.value", "skills.*.passive", "abilities.*.value"
```
Fyorl, line 3475: Seems a bit weird that `spellLevel` is under `details` for NPCs (and `attributes.spellLevel` doesn't seem to exist?). Maybe we should move it under `attributes.spell.level` for consistency.

module/data/actor/templates/common.mjs
Fyorl: It's a number on `ClassData#spellcasting` also, so we should be consistent. And speaking of consistency, we have `ClassData#spellcasting#save` rather than `dc`. Not sure which one we should align on, but I think I'd lean towards `dc` personally.

Fyorl (changes requested):
I think we also need to update `_configureTrackableAttributes` in `dnd5e.mjs`.

module/applications/actor/base-sheet.mjs
```diff
@@ -1112,7 +1112,7 @@ export default class ActorSheet5e extends ActorSheetMixin(ActorSheet) {
     // Case 2: Drop a leveled spell in a section without a mode.
     else if ( (level === "0") || !preparationMode ) {
       if ( this.document.type === "npc" ) {
-        prep.mode = this.document.system.details.spellLevel ? "prepared" : "innate";
+        prep.mode = this.document.system.attributes.spell.level ? "prepared" : "innate";
```
Fyorl, line 1115: We have optional chaining above. Maybe not necessary here.

```suggestion
        prep.mode = this.document.system.attributes.spell?.level ? "prepared" : "innate";
```

Fyorl (approved):

module/applications/actor/base-sheet.mjs
arbron: Since we've already checked that this is a NPC actor, we know that `attributes.spell` object will exist so we can skip the chaining.

PR #5005, opened 2025-01-15
[#3751] Add ability to automatically manually sort compendiums

### LINKED ISSUE #3751: [Compendium Content] Folder Colors in Spells Compendium
FVTT 11.315 and DND5E 3.2.1
It looks like the folder colors in the Spells Compendium were made with Cantrips being the first folder in the list, but the out of the box default sorting sorts them alphabetically, putting Cantrips at the end.

Spell Compendium Folders
In the screenshots above, on the left is the default. On the right is what I think it should be.

I would suggest changing the folder name (as I did above), but changing the folder colors also works for me. Just a minor nitpick, but someone went to the trouble to set the colors on the folders. The default looks jarring to me.

arbron:
Adds a new compendium flag `dnd5e.sorting` that sets the default sorting for a compendium if the user hasn't already adjusted the sorting. This allows for setting a compendium to be manually sorted.

To implement this a new `setupModulePacks` method has been added to the module registration system that is called during setup and sets compendium applications and the sorting mode.

Also adds a new `log` utility method that adds the "D&D 5e | " prefix automatically with a nice red color to make system-specific logs distinct from Foundry core logs.

Closes #3751

### DIFF
--- a/module/module-registration.mjs
+++ b/module/module-registration.mjs
@@ -65,9 +65,10 @@ export function setupModulePacks() {
       const complete = setupMethods.map(m => m(pack)).filter(r => r);
       if ( complete.length ) log(`Finished setting up ${pack.metadata.label}: ${complete.join(", ")}`);
     } catch(err) {
-      log(`Error setting up ${pack.metadata.label}\n`, { extras: [err.message], level: "error" });
+      log(`Error setting up ${pack.title}\n`, { extras: [err.message], level: "error" });
     }
   }
+  if ( sortingChanged ) game.settings.set("core", "collectionSortingModes", collectionSortingModes);
   console.groupEnd();
 }
 
@@ -89,10 +90,8 @@ function setupPackDisplay(pack) {
 
 /* -------------------------------------------- */
 
-let sortingChanges;
-const debouncedUpdateSorting = foundry.utils.debounce(() =>
-  game.settings.set("core", "collectionSortingModes", sortingChanges)
-, 250);
+let collectionSortingModes;
+let sortingChanged = false;
 
 /**
  * Set default sorting order based on `flags.dnd5e.sorting`.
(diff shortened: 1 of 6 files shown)


module/module-registration.mjs
```diff
@@ -43,3 +49,60 @@ function registerSpellLists(manifest) {
...
+  for ( const pack of game.packs ) {
+    if ( pack.metadata.type === "Item" ) pack.applicationClass = ItemCompendium5e;
+    try {
+      const complete = setupMethods.map(m => m(pack)).filter(r => r);
+      if ( complete.length ) log(`Finished setting up ${pack.metadata.label}: ${complete.join(", ")}`);
+    } catch(err) {
+      log(`Error setting up ${pack.metadata.label}\n`, { extras: [err.message], level: "error" });
+    }
+  }
+  console.groupEnd();
```
krbz999, line 71: Feel pretty strongly about not using log groups. Too often they hide valuable errors during package development.

arbron: Well, everything in here is synchronous system code, so there isn't really any chance of it hiding module errors.

module/module-registration.mjs
```diff
@@ -43,3 +49,60 @@ function registerSpellLists(manifest) {
...
+/**
+ * Set default sorting order based on `flags.dnd5e.sorting`.
+ * @param {Compendium} pack  Pack to set up.
+ * @returns {string|void}    Description of the step.
+ */
+function setupPackSorting(pack) {
+  sortingChanges ??= game.settings.get("core", "collectionSortingModes") ?? {};
+  if ( !pack.metadata.flags.dnd5e?.sorting || sortingChanges[pack.metadata.id] ) return;
+  sortingChanges[pack.metadata.id] = pack.metadata.flags.dnd5e.sorting;
+  debouncedUpdateSorting();
```
Fyorl, line 106: This seems a bit awkward. Can we move it out of the generalised `setupMethods` list and instead loop over the packs at the end of `setupModulePacks` again to build up this settings object?

Fyorl (changes requested):

Fyorl (approved):

PR #5007, opened 2025-01-16
[#4830] Only defer to item abilities if attack activity matches attack type

### LINKED ISSUE #4830: Weapons making spell attacks use the incorrect ability when set to "Default"
repro:

- create a weapon
- create an Attack activity.
- set activity to be Melee (default) and Attack Classification: Spell
- in the Attack Details, notice that the default is "Spellcasting"
- when on an actor, notice that the modifier does not correspond to its spellcasting ability
- change to Spellcasting on the activity explicitly (modifier is now correct)

roth-michael:
Would close #4830 
When the activity's `attack.ability` was left to default, it would pull from its item's `availableActivities`, meaning that a spell attack on a weapon would use str or dex, and a weapon attack on a spell would use whatever was set on the spell as the spellcasting ability. By only using the parent item's `availableActivities` if the attack type matches (i.e. weapon attack on a weapon, spell attack on a spell), this problem is avoided without any side effects (to my knowledge).

### DIFF
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -98,11 +98,15 @@ export default class AttackActivityData extends BaseActivityData {
    * @type {Set<string>}
    */
   get availableAbilities() {
-    // Defer to item if available
-    if ( this.item.system.availableAbilities ) return this.item.system.availableAbilities;
+    const attackClassification = this.attack.type.classification;
+
+    // Defer to item if available and matching attack classification
+    if ( this.item.system.availableAbilities && this.item.type === attackClassification ) {
+      return this.item.system.availableAbilities;
+    }
 
     // Spell attack not associated with a single class, use highest spellcasting ability on actor
-    if ( this.attack.type.classification === "spell" ) return new Set(
+    if ( attackClassification === "spell" ) return new Set(
       this.actor?.system.attributes?.spellcasting
         ? [this.actor.system.attributes.spellcasting]
         : Object.values(this.actor?.spellcastingClasses ?? {}).map(c => c.spellcasting.ability)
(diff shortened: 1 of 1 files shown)


module/data/activity/attack-data.mjs
```diff
@@ -98,11 +98,15 @@ export default class AttackActivityData extends BaseActivityData {
    * @type {Set<string>}
    */
   get availableAbilities() {
-    // Defer to item if available
-    if ( this.item.system.availableAbilities ) return this.item.system.availableAbilities;
+    const attackClassification = this.attack.type.classification;
+
+    // Defer to item if available and matching attack classification
+    if ( this.item.system.availableAbilities && this.item.type === attackClassification ) {
```
arbron, line 104:
```suggestion
    if ( this.item.system.availableAbilities && (this.item.type === this.attack.type.classification) ) {
```

module/data/activity/attack-data.mjs
```diff
@@ -98,11 +98,15 @@ export default class AttackActivityData extends BaseActivityData {
...
+      return this.item.system.availableAbilities;
+    }
 
     // Spell attack not associated with a single class, use highest spellcasting ability on actor
-    if ( this.attack.type.classification === "spell" ) return new Set(
+    if ( attackClassification === "spell" ) return new Set(
```
arbron, line 109:
```suggestion
    if ( this.attack.type.classification === "spell" ) return new Set(
```

module/data/activity/attack-data.mjs
```diff
@@ -98,11 +98,15 @@ export default class AttackActivityData extends BaseActivityData {
    * @type {Set<string>}
    */
   get availableAbilities() {
-    // Defer to item if available
-    if ( this.item.system.availableAbilities ) return this.item.system.availableAbilities;
+    const attackClassification = this.attack.type.classification;
+
```
arbron, line 102: Not really running into any line-length issues, so I don't think it is necessary to give this its own local variable.

```suggestion
```

arbron (changes requested):

PR #5010, opened 2025-01-16
Merge subclasses & class features into classes compendium

arbron:
Moves the contents of the Class Features and Subclasses compendiums into the Classes compendium and organize them using folders. Added UUID redirects to keep existing links from breaking.


krbz999:
Won't this break linking when using uuid for consumption? Looking like that just does string comparison atm.

arbron:
> Won't this break linking when using uuid for consumption? Looking like that just does string comparison atm.

Almost certainly!

arbron:
Blocked by https://github.com/foundryvtt/dnd5e/pull/5084

PR #5012, opened 2025-01-17
[#4966] Add post-rest activation types & display in rest messages

### LINKED ISSUE #4966: Add "End of Short Rest" and "End of Long Rest" activation types
Activation types that are triggered after a character completes a short or long rest, displayed in the rest chat message (see https://github.com/foundryvtt/dnd5e/issues/3788) similar to the new combat turn chat message.

arbron:
Adds "After a Long Rest" and "After a Short or Long Rest" activation types. These types will be automatically displayed in rest chat messages in the same manner as turn activation types.

Closes #4966


lang/en.json
```diff
@@ -178,6 +179,12 @@
         "other": "{number} legendary actions"
       }
     },
+    "LongRest": {
+      "Label": "End of a Long Rest"
+    },
+    "ShortRest": {
+      "Label": "End of a Short or Long Rest"
```
Fyorl, line 186: This makes sense for recovery, but I'm not sure about activations. It wouldn't be very useful to prompt to activate Arcane Recovery at the end of a long rest, for example. Should activations be limited to exactly the rest period they specify? The user could fairly easily add both activation periods if that's what they intend.

templates/chat/parts/card-activities.hbs
```diff
@@ -0,0 +1,16 @@
+<section class="activities">
+    <strong class="roboto-condensed-upper">{{ localize "DND5E.CHATMESSAGE.Activities" }}</strong>
+    <ul class="unlist">
+        {{#each activities}}
+        <li class="activity flexrow item-tooltip" data-activity-uuid="{{ uuid }}" data-item-uuid="{{ item.uuid }}">
```
Fyorl, line 5: Wasn't seeing these tooltips activating when I tested. Do we need additional code in `ChatMessage5e` to activate them?

Fyorl (changes requested):
Good consolidation

lang/en.json
arbron: Makes sense.

Fyorl (approved):

PR #5018, opened 2025-01-19
Allow for arbitrary bonus to save DC on activities.

krbz999:
Currently, if wanting to increase the DC of a save activity via enchantments, the activity cannot derive its DC from `save.dc.calculation`, only custom formulas work.

This adds a `save.dc.bonus` property during data prep that can be targeted by enchantments and accepts roll data. It works with both custom formula and ability-derived DCs.


Fyorl (approved):
Seems fine to me. Will leave final approval to @arbron in case we have a different strategy in mind for affecting activity save DCs via enchantments.

arbron (approved):
No objections from me.

PR #5021, opened 2025-01-20
[#4802] Add "display chat card" button to inventory item context menu

### LINKED ISSUE #4802: Add a "display" option to item context menu
With the (welcome) addition of automatically triggering actions upon activity use, there's no workflow for _just_ displaying an item as a chat card. Adding a "display in chat" type option to the item context menu which calls `Item#displayCard` would add that functionality with minimal code impact.

roth-michael:
Closes #4802 
Adds an option to the context menu for inventory items to display in chat. In order not to show the effect application tray for a display-only use of `displayCard`, used a new `dnd5e.displayOnly` flag on the message which, if `true`, blocks the enriching of usage effects.
image
Not sure whether any conditions (like owning the item, for instance) should be necessary, or whether it's in the appropriate section of the context menu.

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -652,17 +652,11 @@ export default class ChatMessage5e extends ChatMessage {
    * @protected
    */
   _enrichUsageEffects(html) {
-    if ( this.getFlag("dnd5e", "displayOnly") ) return;
+    if ( this.getFlag("dnd5e", "messageType") !== "usage" ) return;
     const item = this.getAssociatedItem();
-    let effects;
-    if ( this.getFlag("dnd5e", "messageType") === "usage" ) {
-      effects = this.getFlag("dnd5e", "use.effects")?.map(id => item?.effects.get(id));
-    } else {
-      if ( this.getFlag("dnd5e", "roll.type") ) return;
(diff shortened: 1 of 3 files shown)


module/documents/chat-message.mjs
```diff
@@ -652,6 +652,7 @@ export default class ChatMessage5e extends ChatMessage {
    * @protected
    */
   _enrichUsageEffects(html) {
+    if ( this.getFlag("dnd5e", "displayOnly") ) return;
```
arbron, line 655: I think we're actually planning on taking out the display of effects for cards without activities, so rather than set the flag who don't you just tweak this function to remove the `else` block below (which should allow simplifying to move the effects getter out of the if statement.

```javascript
    if ( this.getFlag("dnd5e", "messageType") !== "usage" ) return;
    const item = this.getAssociatedItem();
    const effects = this.getFlag("dnd5e", "use.effects")
      ?.map(id => item?.effects.get(id))
      .filter(e => e && (game.user.isGM || (e.transfer && (this.author.id === game.user.id))));
```

arbron (approved):

Fyorl (approved):

Issue #4513, opened 2024-10-07
Activity Type: Sequence


hakr14:
Activity that defines a sequence of other activities.
Using the activity would immediately trigger the first activity of the sequence, and add a button to the resultant chat card to trigger the next activity of the sequence. That activity would have a button for the next activity, and so on. The final activity of a sequence requires no additional handling.
To prevent loops no Sequence activities may be included in a sequence.

JPMeehan:
What's the difference from #4031 ?

hakr14:
> What's the difference from #4031 ?

Forward "modifies" the use of the forwarded activity, and only points to a single other activity. Sequence does not modify the activities and instead allows for multiple activities to be used in sequence directly from the chat cards, without needing to select them yourself from the item.
Mentioned this on the Discord and honeybadger said it wouldn't be much work. Figured if that's the case I may as well ticket it. https://discord.com/channels/170995199584108546/670336046164213761/1292487976173178961

DudeAndre:
This seems like a very good idea. 
I hope with the MM, where a lot of creature have attacks that trigger saves and so forth, this (or something to the same effect) will be integrated into the system.

Zanderaf:
That would be ideal, there are a fair few attacks for instance in official content that is an attack + a following saving throw.

roth-michael:
Thinking more about this:
If this does get implemented, it might be nice to have a "hide activity" checkbox which makes an activity not show up in the dropdown or when the item is clicked. The reason for this would be in certain cases, a "Sequence" of "attack -> save" would be the only valid use of an item. Rather than showing the attack, save, and sequence activities, the attack & save activities could be left to be only modifiable, not directly runnable. I _don't_ think this should be automatic if an activity is part of a sequence, but I _do_ think it would be a useful option.

PR #5022, opened 2025-01-20
Add "Followup Activity" to activities to allow one activity to "flow" into another

roth-michael:
Somewhat fulfills the ask of #4513 - perhaps obsoletes it.
Implements the "Follow-up Activities" section of #1964 

Done so far:
- Adds a "Followup Activity" dropdown to every activity sheet (except for Forward and Cast, which don't have their "own" chat cards to display). 
- The added dropdown displays a list of the other activities on the item (filtered so as not to allow any cycles, taking forwarded activities into account). 
- When an activity with a followup is used, an additional button appears in the chat card which, when clicked, will use the followup activity.
- Scaling & spell slot info, if they exists in the original message's flags, are passed into the new use, so that the "cast at level" box will be pre-filled with the level used to cast the original Activity, if it was a spell.

Parts I'm not sure about:
- Currently, I've put it in the `parts/activity-identity.hbs` template, and check if the activity type is `forward` or `cast` and, if so, don't show it. I don't know whether it belongs there, but I do think it should exist in the "default" activity template, perhaps gated by a `canHaveFollowup` static property on activity types, so that future activity types (or those added by modules) can make use of followup Activities without having to modify their templates at all.
- There's a case to be made for activities which are selected as followup activities to not be options when using an item directly. I suspect there's also a case to be made that they _should_ be accessible without having to first use whatever activity has them as a followup, so I'm not sure the best way to handle that, if at all. The middle ground would be a "Don't prompt this activity" checkbox, but that's almost entirely unrelated to the scope of this PR, and would fit in better with a PR dedicated to conditional activity visibility.

Followup activity selection:
image

Auto-filled scaling:
image

Chat after proceeding through the above dialog:
image

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -193,7 +193,8 @@
 "DND5E.ACTIVITY": {
   "Action": {
     "Create": "Create Activity",
-    "Delete": "Delete Activity"
+    "Delete": "Delete Activity",
+    "Followup": "Follow-up: {activityName}"
   },
   "Title": {
     "one": "Activity",
@@ -209,6 +210,10 @@
     },
     "effects": {
       "label": "Applied Effects"
+    },
+    "followupActivity": {
+      "label": "Followup Activity",
+      "hint": "Which activity, if any, should be usable from this activity's chat card."
     },
     "img": {
       "label": "Icon"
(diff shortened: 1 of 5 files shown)


arbron:
I will say before you get any further with this, that I think we might do something more complex for followup activities because there are situations where there might be more than one result from an activity, plus there are conditional followups.

roth-michael:
Sounds good - that is one element that occurred to me when I was looking at how best to implement it (specifically the "might have multiple" part - conditional to me felt like it might be a "know whether you should press the button or not" situation).

PR #5071, opened 2025-01-28
Update Starter Heroes with fresh items and add favorites

Fyorl:


Fyorl:
@MaxPat931 If you could comment on this PR so that I can assign it to you, to make sure the patch notes are accurate.

MaxPat931:
Awesome, thanks for the updates!

Issue #5074, opened 2025-01-29
Re-work remaining other applications using AppV2 and V2 styles


arbron:
- [ ] `Award` (already uses V2 styles) (https://github.com/foundryvtt/dnd5e/pull/5127)
- [x] `CompendiumBrowser` (already uses both, but can take advantage of the new 5e APIs to reduce size) (https://github.com/foundryvtt/dnd5e/pull/5111)
- [ ] `CurrencyManager` (already uses V2 styles) (https://github.com/foundryvtt/dnd5e/pull/5127)
- [x] `StartingEquipmentConfig` (already uses V2 styles) (https://github.com/foundryvtt/dnd5e/pull/5075)
- [x] `JournalEditor` (https://github.com/foundryvtt/dnd5e/pull/5079)
- [x] `SpellsUnlinkedConfig` (https://github.com/foundryvtt/dnd5e/pull/5079)
- [x] `PropertyAttribution` (https://github.com/foundryvtt/dnd5e/pull/5110)

PR #5075, opened 2025-01-29
[#5074] Rebuild starting equipment config using AppV2

arbron:
Builds `StartingEquipmentConfig` using `ApplicationV2` but keep the existing styles.

Starting Equipment V2

### DIFF
--- a/module/applications/item/starting-equipment-config.mjs
+++ b/module/applications/item/starting-equipment-config.mjs
@@ -1,33 +1,38 @@
 import PhysicalItemTemplate from "../../data/item/templates/physical-item.mjs";
 import { EquipmentEntryData } from "../../data/item/templates/starting-equipment.mjs";
+import DocumentSheet5e from "../api/document-sheet.mjs";
 
 /**
  * Configuration application for Starting Equipment.
  */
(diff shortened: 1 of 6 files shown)


module/applications/item/starting-equipment-config.mjs
```diff
@@ -1,42 +1,47 @@
...
   }
 
   /* -------------------------------------------- */
   /*  Rendering                                   */
   /* -------------------------------------------- */
 
   /** @inheritDoc */
-  async getData(options={}) {
-    const context = super.getData(options);
+  async _prepareContext(options={}) {
```
krbz999, line 43: Don't think options is ever optional in AppV2, would be news to me at least.
```suggestion
  async _prepareContext(options) {
```

less/v1/advancement.less
```diff
@@ -411,6 +410,8 @@
     }
   }
 
+  form[data-type="Subclass"] .drop-target > button { width: 100%; }
```
Fyorl, line 413: Unrelated; some tweaks so that v13 styles match v12 styles.

Fyorl (approved):

PR #5077, opened 2025-01-30
Add duration dialog when shift-clicking status effect enricher applyAction

Belodri:


Belodri:
Mistakenly created PR here instead of on my fork.

PR #5084, opened 2025-01-31
Adjust sourced items on actors to use remapped UUIDs

arbron:
Moves the `parseUuid` changes from `JournalSpellListPageSheet` into a shared implementation and applied it to `SourcedItemsMap` during get and set operations so it always handles remapped UUIDs. Because the initial setup of sourced items occurs before the UUID redirects are loaded, the system will re-adjust the map during the `ready` hook.


module/documents/actor/actor.mjs
```diff
@@ -3515,7 +3516,8 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
  */
 class SourcedItemsMap extends Map {
   /** @inheritDoc */
-  get(key, { legacy=true }={}) {
+  get(key, { remap=true, legacy=true }={}) {
+    if ( remap ) ({ key: uuid } = parseUuid(key));
```
Fyorl, line 3520:
```suggestion
    if ( remap ) ({ uuid: key } = parseUuid(key));
```

Fyorl (changes requested):

Fyorl (approved):

PR #5094, opened 2025-02-02
[#5093] Adjust scaling label for cantrips

### LINKED ISSUE #5093: Adjust labels for whole/half scaling when the item the activity's on is a Cantrip
A fair number of folks start out confused by scaling reading "Every Level" when on a cantrip. Situationally changing the label for cantrips would eliminate this issue.

roth-michael:
Closes #5093 
Adds `HalfCantrip` and `WholeCantrip` localizations, `labelCantrip` properties to `CONFIG.DND5E.damageScalingModes`'s fields, and uses the them instead of `label` when the item in question is a "spell" type item with `level` of 0.
On a cantrip:
Screenshot 2025-02-02 122208
On a non-cantrip:
Screenshot 2025-02-02 122225

### DIFF
--- a/module/applications/activity/activity-sheet.mjs
+++ b/module/applications/activity/activity-sheet.mjs
@@ -302,9 +302,10 @@ export default class ActivitySheet extends PseudoDocumentSheet {
       ...CONFIG.DND5E.dieSteps.map(value => ({ value, label: `d${value}` }))
     ];
     if ( context.activity.damage?.parts ) {
+      const scaleKey = (this.item.type === "spell" && this.item.system.level === 0) ? "labelCantrip" : "label";
       const scalingOptions = [
         { value: "", label: game.i18n.localize("DND5E.DAMAGE.Scaling.None") },
-        ...Object.entries(CONFIG.DND5E.damageScalingModes).map(([value, config]) => ({ value, label: config.label }))
+        ...Object.entries(CONFIG.DND5E.damageScalingModes).map(([value, { [scaleKey]: label }]) => ({ value, label }))
       ];
       const typeOptions = Object.entries(CONFIG.DND5E.damageTypes).map(([value, { label }]) => ({ value, label }));
       const makePart = (data, index) => this._prepareDamagePartContext(context, {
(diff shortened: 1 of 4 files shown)


module/applications/activity/activity-sheet.mjs
```diff
@@ -302,9 +302,10 @@ export default class ActivitySheet extends PseudoDocumentSheet {
       ...CONFIG.DND5E.dieSteps.map(value => ({ value, label: `d${value}` }))
     ];
     if ( context.activity.damage?.parts ) {
+      const scaleKey = (this.item.type === "spell" && this.item.system.level === 0) ? "labelCantrip" : "label";
```
Fyorl, line 305:
```suggestion
      const scaleKey = (this.item.type === "spell") && (this.item.system.level === 0) ? "labelCantrip" : "label";
```

Fyorl (approved):
Is 'half cantrip scaling' a thing? Maybe we could add some extra refinement by removing that option entirely for cantrips?

I will leave final approval to @arbron but I don't see anything objectionable here.

krbz999:
> Is 'half cantrip scaling' a thing? Maybe we could add some extra refinement by removing that option entirely for cantrips?

Probably simpler to just keep it. I also seem to remember a spell in some third party premium content (Valda's, perhaps?) that had two activities, each scaling every other "cantrip level".

module/config.mjs
```diff
@@ -2317,13 +2317,15 @@ DND5E.dieSteps = [4, 6, 8, 10, 12, 20, 100];
  */
 DND5E.damageScalingModes = {
   whole: {
-    label: "DND5E.DAMAGE.Scaling.Whole"
+    label: "DND5E.DAMAGE.Scaling.Whole",
+    labelCantrip: "DND5E.DAMAGE.Scaling.WholeCantrip"
```
arbron, line 2321: Just give a little update to the documentation above to indicate this new property exists (`@enumerated {{ label: string, labelCantrip: string }}`).

arbron (changes requested):

arbron (approved):

PR #5105, opened 2025-02-03
[#5099] Preserve elevation & rotation when transforming

### LINKED ISSUE #5099: Wild Shape While at Higher Elevation
If a token transforms via wild shape, polymorph, etc its elevation is always reset to 0. This is an issue with multi-story maps or maps that incorporate elevation changes.

arbron:
Copies elevation and rotation data from the originating token when transforming and reverting.

Closes #5099


krbz999:
I would make an argument for `hidden` as well.

Fyorl (approved):

PR #5110, opened 2025-02-04
[#5074] Convert `PropertyAttribution` to `ApplicationV2`

arbron:
No change in design due to property attribution being styled like core's tooltips, but now uses `ApplicationV2` rather than the old API.

### DIFF
--- a/module/applications/property-attribution.mjs
+++ b/module/applications/property-attribution.mjs
@@ -32,8 +32,9 @@ export default class PropertyAttribution extends Application5e {
   /** @override */
   static DEFAULT_OPTIONS = {
     classes: ["property-attribution"],
-    position: {
-      width: 320
+    window: {
+      frame: false,
+      positioned: false
     }
   };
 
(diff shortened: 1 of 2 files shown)


module/applications/property-attribution.mjs
```diff
@@ -27,32 +29,45 @@ export default class PropertyAttribution extends Application {
...
   async renderTooltip() {
-    const data = this.getData(this.options);
-    return (await this._renderInner(data))[0].outerHTML;
+    const options = {};
+    this._configureRenderOptions(options);
+    const context = await this._prepareContext(options);
+    const rendered = await this._renderHTML(context, options);
+    const div = document.createElement("div");
+    div.append(...Object.values(rendered));
+    return div.innerHTML;
```
Fyorl, line 64: I think we can simplify this by first setting `window.frame: false` in `DEFAULT_OPTIONS`, and adding this `_insertElement` override:

```js
/** @override */
_insertElement(element) {} // do nothing
```

Then this method becomes something like:

```js
async renderTooltip() {
  await this.render();
  return this.element.innerHTML; // Maybe outerHTML, not sure
}
```

Fyorl (changes requested):

Fyorl (approved):

PR #5112, opened 2025-02-04
[#5102] Add variant rules settings application

### LINKED ISSUE #5102: Organize more settings into sub-applications
### Combat Settings
- Initiative Dex Tiebreaker
- Initiative Score
- Critical Damage Multiply Modifiers
- Critical Damage Max Dice

### Variant Rules
- Rules Version (but this should probably stay at top-level)
- Rest Variant
- Proficiency Variant
- Allow Feats
- Honor Ability Score
- Sanity Ability Score

arbron:
Adds a settings application for variant rules that contains the rest variant, proficiency variant, leveling mode, encumbrance settings, and additional ability scores.

Variant Rules (modern)

Also includes the "Allow Feats" setting, but this only appears when the legacy rules version is selected.

Variant Rules (legacy)

Also removes the unused diagonal movement setting and some unused setting localization strings.

Closes #5102

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -4726,7 +4726,7 @@
       "Variant": "Variant (encumbered & heavily encumbered)"
     },
     "HonorScore": {
-      "Hint": "Enable the use of the optional Honor ability score. Requires the world to be reloaded.",
+      "Hint": "Enable the use of the optional Honor ability score.",
       "Name": "Honor Ability Score"
     },
     "LevelingMode": {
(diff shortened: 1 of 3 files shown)


lang/en.json
```diff
@@ -4747,6 +4706,54 @@
...
+    },
+    "Encumbrance": {
+      "Hint": "Enable automatic tracking of encumbrance and the application of status effects for characters carrying too much.",
+      "Name": "Encumbrance Tracking",
+      "None": "None",
+      "Normal": "Normal (max carrying capacity)",
+      "Variant": "Variant (encumbered & heavily encumbered)"
+    },
+    "HonorScore": {
+      "Hint": "Enable the use of the optional Honor ability score. Requires the world to be reloaded.",
```
Fyorl, line 4729: Do we need these 'required world to be reloaded' hints anymore? I can't remember if we added them before we had `requiresReload`, or whether we added them in addition to that dialog.

Fyorl (approved):

lang/en.json
arbron: These hints were here from the start when honor & sanity were added when we were using our own debounced reload.

PR #5113, opened 2025-02-04
Suppress some V13 warnings and add alternate paths to others

arbron:
- Suppress namespace warnings
- Suppress ApplicationV1 warnings
- Suppress `Set#isSubset` warning
- Suppress `ChatMessage#getHTML` warning
- Suppress `renderchatMessage` warning
- Adjust creation of `ContextMenu` objects to pass in `HTMLElement` in V13 and pass the `jQuery` option
- Modified access to `Document#compendium` when running in V13 to use either `Document#collection` or `Document#inCompendium`

### DIFF
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -346,11 +346,12 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
 
     // Apply special context menus for items outside inventory elements
     const featuresElement = html[0].querySelector(`[data-tab="features"] ${this.options.elements.inventory}`);
-    if ( featuresElement ) {
-      new ContextMenu5e(html, ".pills-lg [data-item-id], .favorites [data-item-id], .facility[data-item-id]", [], {
-        onOpen: (...args) => featuresElement._onOpenContextMenu(...args)
-      });
-    }
+    if ( featuresElement ) new ContextMenu5e(
+      game.release.generation < 13 ? html : html[0],
+      ".pills-lg [data-item-id], .favorites [data-item-id], .facility[data-item-id]",
+      [],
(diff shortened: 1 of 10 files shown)


module/applications/actor/character-sheet-2.mjs
```diff
@@ -346,11 +346,12 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
 
     // Apply special context menus for items outside inventory elements
     const featuresElement = html[0].querySelector(`[data-tab="features"] ${this.options.elements.inventory}`);
-    if ( featuresElement ) {
-      new ContextMenu5e(html, ".pills-lg [data-item-id], .favorites [data-item-id], .facility[data-item-id]", [], {
-        onOpen: (...args) => featuresElement._onOpenContextMenu(...args)
-      });
-    }
+    if ( featuresElement ) new ContextMenu5e(
+      game.release.generation < 13 ? html : html[0],
```
Fyorl, line 350: `ContextMenu` has actually supported passing an HTMLElement as the first argument for a while, even in v12.

```suggestion
      html[0],
```

module/applications/actor/character-sheet-2.mjs
```diff
@@ -346,11 +346,12 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
...
+      ".pills-lg [data-item-id], .favorites [data-item-id], .facility[data-item-id]",
+      [],
+      { onOpen: (...args) => featuresElement._onOpenContextMenu(...args), jQuery: false }
```
Fyorl, line 353: Should this be `true`? I think the inventory element context menu callbacks still expect jQuery.

```suggestion
      { onOpen: (...args) => featuresElement._onOpenContextMenu(...args), jQuery: true }
```

Fyorl (changes requested):

Fyorl (approved):

PR #5116, opened 2025-02-04
[#4628] Preserve initial roll configuration for rebuilding.

### LINKED ISSUE #4628: 4.1: d20 roll ConfigurationDialogs do not make use of any added parts added programmatically

Fyorl:
- Closes #4628 

Just a draft to run this potential solution by you. I think the root issue was we were running this line too early in `#rollSkillTool`.

```js
rollConfig.rolls.forEach((r, index) => buildConfig(rollConfig, r, null, index));
```

This is mutating our initial roll configs before they are stored by the roll dialog. That means that when we try to rebuild the rolls, we have to ignore the roll configs entirely, because they are contaminated with 'derived' data from the pre-dialog call to `buildConfig`.

It seems we are only performing this early `buildConfig` call in case the roll dialog is not instantiated at all, and never gets a chance to call `buildConfig` itself. So I think that by moving this logic into `BasicRoll.buildConfigure` and calling it only when `dialog.configure === false`, then we avoid having to contaminate the original roll configs.

Since we can trust the roll configs now, we can appropriately merge them rather than ignoring them in `_buildSkillToolConfig`.

I tested with a similar snippet to the one posted in the issue when rolling a skill check, and it seemed to correctly add the part.

```js
dialog.config.rolls[0].parts = ['1d4'];
dialog.rebuild();
```

Let me know if I'm missing something and we can't actually go with this solution. If not, I'll make the appropriate changes to the various other roll methods, and also see what I can do about the `initialRoll` logic also.

### DIFF
--- a/module/dice/basic-roll.mjs
+++ b/module/dice/basic-roll.mjs
@@ -128,7 +128,7 @@ export default class BasicRoll extends Roll {
    * @param {BasicRollProcessConfiguration} [config={}]   Configuration for the rolls.
    * @param {BasicRollDialogConfiguration} [dialog={}]    Configuration for roll prompt.
    * @param {BasicRollMessageConfiguration} [message={}]  Configuration for message creation.
-   * @returns {BasicRoll[]}
+   * @returns {Promise<BasicRoll[]>}
    */
   static async buildConfigure(config={}, dialog={}, message={}) {
     config.hookNames = [...(config.hookNames ?? []), ""];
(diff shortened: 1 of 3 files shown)


arbron:
Everything seems to work, but one potential issue is the `preRoll` hook now receives potentially incomplete roll configuration when it it is expecting the whole thing. Not sure it is worth holding up the solution over, we could perhaps just specify it as `Partial<BasicRollProcessConfiguration>` to inform uses to be careful.

arbron (approved):

PR #5128, opened 2025-02-06
[#3776, #4422, #5109] Improve handling of initiative advantage

### LINKED ISSUE #3776: [Feature Request] Disadvantage on Initiative Flag
At the moment there is a flag for advantage on initiative rolls, but there isn't one for disadvantage. It would be nice if there is one.

Crawford also mentions disadvantage on initiative in a recent [vid](https://youtu.be/MJS8z86UAOg?si=KmSEcabP8flmapgr&t=265) regarding the reworked surprised condition, so implementing it could save some work now, as it will be part of dnd5e raw later.

### LINKED ISSUE #4422: [Feature Request] Roll NPCs initiative button with disadvantage in combat tab
The new '24 Surprised condition means you roll initiative with disadvantage, as a GM, it'd be nice to be able to use the roll NPCs initiative button to roll with disadvantage when relevant.

### LINKED ISSUE #5109: If an actor has advantage on initiative, their initiative score should have +5
As can be seen for the Invisible Stalker from the new Monster Manual. They have +7 and advantage to Initiative because of the Invisible condition, resulting in a 22 Initiative Score.

Accordingly, disadvantage should give -5 to the score.

Image

arbron:
Adjust the initiative score by +5 or -5 relative to the provided advantage mode. To fully support this adds roll mode to the initiative config and deprecated the old initiative advantage flag. If an actor has the deprecated flag set it will appear in the config app, but as soon as it is unchecked the option will disappear. The same is true for the special traits tab.

Adds a new helper to `AdvantageModeField` to help setting the mode from code. If the actor has the initative advantage flag, remarkable athlete flag, or the surpised status  then the initiative roll mode will be adjusted accordingly.

The initiative rolling now reflects the roll mode rather than using the flags directly.

Closes #3776
Closes #4422
Closes #5109

### DIFF
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -1,6 +1,7 @@
 import ActiveEffect5e from "../../../documents/active-effect.mjs";
 import Proficiency from "../../../documents/actor/proficiency.mjs";
 import { convertLength, convertWeight, replaceFormulaData, simplifyBonus } from "../../../utils.mjs";
+import AdvantageModeField from "../../fields/advantage-mode-field.mjs";
 import FormulaField from "../../fields/formula-field.mjs";
 import MovementField from "../../shared/movement-field.mjs";
 import RollConfigField from "../../shared/roll-config-field.mjs";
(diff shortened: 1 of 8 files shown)


module/data/actor/templates/attributes.mjs
```diff
@@ -380,13 +381,21 @@ export default class AttributesFields {
...
+    }
+
     // Total initiative includes all numeric terms
     const initBonus = simplifyBonus(init.bonus, rollData);
     const abilityBonus = simplifyBonus(ability.bonuses?.check, rollData);
     init.total = init.mod + initBonus + abilityBonus + globalCheckBonus
       + (flags.initiativeAlert && isLegacy ? 5 : 0)
       + (Number.isNumeric(init.prof.term) ? init.prof.flat : 0);
-    init.score = 10 + init.total;
+    init.score = 10 + init.total + (init.roll.mode * 5);
```
krbz999, line 398: Is there an issue for also granting +5/-5 to passive skills/tools/checks?

arbron: Don't think there is an issue for those yet, but it is definitely something to handle when we put the finishing touches on advantage for those.

thatlonelybugbear:
In similar fashion, Incapacitated '24 should provide initiativeDisadvantage, whereas Invisible '24, initiativeAdvantage.

Fyorl (commented):
I figured we would just migrate `flags.initiativeAdv` to `attributes.init.roll.mode`. Then we don't need any special logic in the special traits tab or config to hide the deprecated fields. Is there some complication involved in doing it that way?

arbron:
> I figured we would just migrate `flags.initiativeAdv` to `attributes.init.roll.mode`. Then we don't need any special logic in the special traits tab or config to hide the deprecated fields. Is there some complication involved in doing it that way?

Well, we'd have to handle any active effects that target that that may or may not have been migrated.

Fyorl:
> Well, we'd have to handle any active effects that target that that may or may not have been migrated.

Don't think that's typically been a blocker when we've shuffled model fields around before, but maybe this particular field is more commonly an AE target.

We could migrate it in AE application like we did with the enchantment fields.

It's a little bit awkward having both the advantage mode drop-down and the 'advantage on initiative' field in the initiative config because it's two ways to do the same thing. A user might reasonably be confused at the redundancy and think the fields must mean something else. An alternative way to solve that would be by hiding the 'advantage on initiative' field if the advantage mode is anything other than 'normal' though.

Not sure which is preferable. I can look into doing the migration and seeing if it's actually just worse.

krbz999:
Are we not able to "just" adjust any entries in `ActiveEffect#changes` that have `key: "flags.dnd5e.initiativeAdv"`?

Fyorl (approved):

PR #5138, opened 2025-02-07
[#5132] Display compendium browser link in item choice advancement

### LINKED ISSUE #5132: Add Compendium Browser link for selecting items on Choose Items advancement if "Allow Drops" is set
Similar to the compendium browser button in the Subclass & ASI advancement flows, this would allow more easily selecting items. It should take the number of items that can be selected into account using the `select` method rather than `selectOne` used by the other advancement types.

arbron:
When "Allow Drops" is set for an choose items advancement then a compendium browser link will be displayed. Clicking on it will open the browser with filters applied based on the restrictions set on the advancement.

Closes #5132

### DIFF
--- a/module/applications/advancement/item-choice-flow.mjs
+++ b/module/applications/advancement/item-choice-flow.mjs
@@ -1,4 +1,5 @@
 import Actor5e from "../../documents/actor/actor.mjs";
+import CompendiumBrowser from "../compendium-browser.mjs";
 import ItemGrantFlow from "./item-grant-flow.mjs";
 
 /**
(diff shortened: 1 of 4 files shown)


module/applications/advancement/item-choice-flow.mjs
```diff
@@ -140,7 +154,65 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
...
+    // Determine how many items can be selected
+    const max = this.advancement.configuration.choices[this.level]?.count ?? 0;
+    const current = this.selected.size;
+    if ( current >= max ) {
+      ui.notifications.warn("DND5E.ADVANCEMENT.ItemChoice.Warning.MaxSelected", { localize: true });
+      return;
+    }
+
+    const config = this.advancement.configuration;
+    const filters = { locked: { additional: { }, documentClass: "Item" } };
```
```suggestion
    const filters = { locked: { additional: {}, documentClass: "Item" } };
```

PR #5142, opened 2025-02-08
[#5137] Add spell list restriction to Choose Items advancement

### LINKED ISSUE #5137: Allow restricting item choice advancement by spell list
Certain spell choices such as a Warlock's Mystic Arcanum can only pull from a specific spell list, so that should be an option in the Choose Item advancement.

arbron:
Adds a new restriction when the spell type is selected in the item choice advancement that only allows spells from a specific spell list.

Also adds a subclass registry to match the class one to improve the naming of subclass spell lists.

Closes #5137

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -409,6 +409,10 @@
           "hint": "Only allow choices from spells of this level.",
           "label": "Spell Level"
         },
+        "list": {
+          "hint": "Only allow spells from this list.",
+          "label": "Spell List"
+        },
         "subtype": {
           "label": "Subtype"
         },
@@ -434,7 +438,8 @@
       "NoOriginal": "Previously selected choice no longer available for replacement.",
       "PreviouslyChosen": "This item has already been chosen at a previous level.",
       "SpellLevelAvailable": "Only {level} or lower spells can be chosen for this advancement.",
-      "SpellLevelSpecific": "Only {level} spells can be chosen for this advancement."
+      "SpellLevelSpecific": "Only {level} spells can be chosen for this advancement.",
+      "SpellList": "Only spells available on the {list} spell list can be chosen for this advancement."
     }
   },
   "ItemGrant": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -410,8 +410,8 @@
           "label": "Spell Level"
         },
         "list": {
-          "hint": "Only allow spells from this list.",
-          "label": "Spell List"
+          "hint": "Only allow spells from these lists.",
+          "label": "Spell Lists"
         },
         "subtype": {
           "label": "Subtype"
--- a/lang/en.json
+++ b/lang/en.json
@@ -439,7 +439,7 @@
       "PreviouslyChosen": "This item has already been chosen at a previous level.",
       "SpellLevelAvailable": "Only {level} or lower spells can be chosen for this advancement.",
       "SpellLevelSpecific": "Only {level} spells can be chosen for this advancement.",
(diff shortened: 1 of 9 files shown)


Zanderaf:
Is only one spell list able to be selected or are multiple possible?

module/registry.mjs
```diff
@@ -285,7 +285,7 @@ class SpellListRegistry {
       const lists = this.#byType.get(type);
       if ( !lists ) return [];
       return Array.from(lists.entries())
-        .map(([value, list]) => ({ value, label: list.name, group, type }))
+        .map(([value, list]) => ({ value: `${type}.${value}`, label: list.name, group, type }))
```
Fyorl, line 288: Seems like a breaking change.

arbron:
> I wonder if we need a different approach. Currently we don't validate spell list at all, but I would expect that most class features restrict granted spells to that class' spell list most of the time. In fact, I'd expect that is the default, in which case we might be better served by a single boolean for 'do _not_ restrict spells to this class' list'.
> 
> That would work for the Warlock's Mystic Arcanum, and 2014 Bard's Mystical Secrets. Are there some features it doesn't work for? Arcana Cleric's Arcane Mastery I guess?

The issue with this is there are places where we want to select from a specific spell list, like the Artificer Initiate feat which allows for selecting spells from the Artificer spell list specifically or the Magic Initiate feats which allow choosing from Cleric, Druid, or Wizard lists.

Zanderaf:
Nature Domain Clerics also choose from the druid spell list when selecting their cantrip for their 1st level feature, at least 2014-wise

Fyorl:
> The issue with this is there are places where we want to select from a specific spell list, like the Artificer Initiate feat which allows for selecting spells from the Artificer spell list specifically or the Magic Initiate feats which allow choosing from Cleric, Druid, or Wizard lists.

Sounds like we do need to handle selecting multiple lists then. Creating a spell advancement on a class and having to select the same class list each time feels redundant though, so maybe we can have some sensible inference there.

Additionally, it seems like the item grant advancement should have this feature for restricting spell drops too.

module/registry.mjs
arbron: I think this is probably better to make as breaking, rather than trying to build in some deprecation path. Its relatively new anyway and the old behavior stops being useful when you want to do anything beyond a single type of spell list.

arbron:
> Sounds like we do need to handle selecting multiple lists then. Creating a spell advancement on a class and having to select the same class list each time feels redundant though, so maybe we can have some sensible inference there.

Added support for multiple lists.

As for inferring values, most times you will only have to set the spell list once since Item Choice support multiple levels (this is the case for Bard's Magical Secrets, not for Warlock's Mystical Arcanum since a different spell level needs to be set for each level you get it).

> Additionally, it seems like the item grant advancement should have this feature for restricting spell drops too.

Item grants don't support drops from players, and thus don't have any spell restrictions.

lang/en.json
```diff
@@ -409,6 +409,10 @@
           "hint": "Only allow choices from spells of this level.",
           "label": "Spell Level"
         },
+        "list": {
+          "hint": "Only allow spells from this list.",
+          "label": "Spell List"
```
```suggestion
          "label": "Spell Lists"
```

lang/en.json
```diff
@@ -409,6 +409,10 @@
           "hint": "Only allow choices from spells of this level.",
           "label": "Spell Level"
         },
+        "list": {
+          "hint": "Only allow spells from this list.",
```
```suggestion
          "hint": "Only allow spells from these lists.",
```

Fyorl (approved):

PR #5153, opened 2025-02-10
[#5152] Create new `usage` message and migrate existing messages

### LINKED ISSUE #5152: Convert usage chat messages into custom message type
Currently the usage chat messages are created as base chat messages with `flags.dnd5e.messageType` set to `usage` and other details also stored in flags. The `ChatMessage5e` class also contains a lot of logic that is specific to just usage messages. Creating a new chat message type for usage would allow the data to move from flags to the system data and would allow some cleaning up of the `ChatMessage5e` class. It would also open up the possibility of sharing some code with rest & turn messages.

arbron:
Adds a new `usage` message type that stores the usage cause, consumption deltas, and appliable effects in its system data. A migration has been added to convert all existing messages with the `flags.dnd5e.messageType` set to `usage` to this new message type and move any other flag data into the system data.

Closes #5152


module/documents/activity/mixin.mjs
```diff
@@ -974,14 +946,15 @@ export default function ActivityMixin(Base) {
...
-        rollMode: game.settings.get("core", "rollMode"),
         data: {
           content: await renderTemplate(this.metadata.usage.chatCard, context),
-          speaker: ChatMessage.getSpeaker({ actor: this.item.actor }),
           flags: {
             core: { canPopout: true }
-          }
-        }
+          },
+          speaker: ChatMessage.getSpeaker({ actor: this.item.actor }),
```
krbz999, line 954:
```suggestion
          speaker: ChatMessage.implementation.getSpeaker({ actor: this.item.actor }),
```

krbz999: (Chat message subtypes - so hot right now.)

module/data/chat-message/usage-message-data.mjs
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Control visibility of chat card action buttons based on viewing user.
+   * @param {HTMLElement} element  Rendered contents of the message.
+   * @protected
+   */
+  _displayChatActionButtons(element) {
+    if ( this.parent.shouldDisplayChallenge ) element.dataset.displayChallenge = "";
+
+    const isCreator = game.user.isGM || this.actor?.isOwner || (this.parent.author.id === game.user.id);
```
Fyorl:
```suggestion
    const isCreator = game.user.isGM || this.actor?.isOwner || this.parent.isAuthor;
```

module/migration.mjs
```diff
@@ -105,6 +105,20 @@ export async function migrateWorld({ bypassVersionCheck=false }={}) {
     incrementProgress();
   }
 
+  // Migrate World Messages
+  for ( const m of game.messages ) {
+    try {
+      const updateData = migrateMessageData(m.toObject(), migrationData);
+      if ( !foundry.utils.isEmpty(updateData) ) {
+        log(`Migrating Message document ${m.id}`);
+        await m.update(updateData, { enforceTypes: false, render: false });
```
Fyorl, line 114: Maybe this is v14-only, but you cannot change a Document's `type` without force-replacing its entire `system`. Updating with `recursive: false` is the easiest way to do this.


```suggestion
        await m.update(updateData, { enforceTypes: false, render: false, diff: false, recursive: false });
```

module/migration.mjs
```diff
@@ -662,12 +676,34 @@ export const migrateEffectData = function(effect, migrationData, { parent }={})
...
+  const updateData = {};
+  if ( (messageData.flags?.dnd5e?.messageType === "usage") && (messageData.type !== "usage") ) {
+    updateData.type = "usage";
+    updateData["flags.dnd5e.-=messageType"] = null;
+    updateData["system.cause"] = messageData.flags?.dnd5e?.use?.cause;
+    updateData["flags.dnd5e.use.-=cause"] = null;
+    updateData["system.effects"] = messageData.flags?.dnd5e?.use?.effects;
+    updateData["flags.dnd5e.use.-=effects"] = null;
+    updateData["system.deltas"] = messageData.flags?.dnd5e?.use?.consumed;
+    updateData["flags.dnd5e.use.-=consumed"] = null;
```
Fyorl: In order to facilitate the `recursive: false` above, we shouldn't use deletion keys since they will be persisted as-is. We can provide a full flags object that will be replaced at the root:


```suggestion
  const { flags } = messageData;
  if ( (flags?.dnd5e?.messageType === "usage") && (messageData.type !== "usage") ) {
    const use = flags.dnd5e.use;
    updateData.type = "usage";
    updateData["system.cause"] = use?.cause;
    updateData["system.effects"] = use?.effects;
    updateData["system.deltas"] = use?.consumed;
    delete flags.dnd5e.messageType;
    if ( use ) {
      delete use.cause;
      delete use.effects;
      delete use.consumed;
    }
    updateData.flags = flags;
```

arbron: Since we don't have any existing chat message system data to worry about, perhaps it might be a bit safer to use:

```javascript
    updateData["==system"] = {
      cause: use?.cause,
      deltas: use?.consumed,
      effects: use?.effects
    };
```

Unless there is something that may be an issue with.

Fyorl: That should work fine too

arbron:
> Just wanted to re-iterate my comment about what the plan is for the other `use` flags we still have like `spellLevel`, `concentrationId`, `enchantmentProfile`, etc., in case you missed it.

Moved `dnd5e.scaling`, `dnd5e.use.spellLevel`, and `dnd5e.use.concentrationId` into the system data. I've left `enchantmentProfile` out because that is only used by one specific advancement type, so it feels more like flag data.

PR #5157, opened 2025-02-10
[#5156] Add dialog for splitting items in inventory into two stacks

### LINKED ISSUE #5156: Add context option for splitting stack of items in two
With containers or when transferring items to another player it is often useful to take a stack of items and split them into two stacks. A context menu option in the inventory that opens a dialog with a control for how many end up in each stack would be nice.

arbron:
Adds a "Split Stack" context menu option for an item that has a quantity of at least 2. Contains a modified version of core's `<range-picker>` element that has a number input on the left side. ~~This modified element won't be required if
https://github.com/foundryvtt/foundryvtt/issues/12147 is resolved.~~

Closes #5156

### DIFF
--- a/module/applications/item/split-stack-dialog.mjs
+++ b/module/applications/item/split-stack-dialog.mjs
@@ -1,0 +1,69 @@
+import Dialog5e from "../api/dialog.mjs";
+
+/**
+ * Small dialog for splitting a stack of items into two.
+ */
+export default class SplitStackDialog extends Dialog5e {
+  /** @override */
+  static DEFAULT_OPTIONS = {
+    buttons: [{
+      id: "split",
+      label: "DND5E.SplitStack.Action",
+      icon: "fa-solid fa-arrows-split-up-and-left"
+    }],
+    classes: ["split-stack"],
+    document: null,
+    form: {
+      handler: SplitStackDialog.#handleFormSubmission
+    },
+    position: {
+      width: 400
+    },
+    window: {
+      title: "DND5E.SplitStack.Title"
+    }
+  };
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  static PARTS = {
+    ...super.PARTS,
+    content: {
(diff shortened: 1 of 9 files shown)


module/applications/item/split-stack-dialog.mjs
```diff
@@ -0,0 +1,69 @@
...
+   * @this {SplitStackDialog}
+   * @param {Event|SubmitEvent} event    The form submission event.
+   * @param {HTMLFormElement} form       The submitted form.
+   * @param {FormDataExtended} formData  Data from the dialog.
+   */
+  static async #handleFormSubmission(event, form, formData) {
+    const right = formData.object.right ?? 0;
+    const left = (this.options.document.system.quantity ?? 1) - right;
+    if ( left === this.options.document.system.quantity ) return;
+    await this.options.document.update({ "system.quantity": left }, { render: false });
```
krbz999, line 65: Is `render: false` to prevent the actor sheet rendering twice? This would prevent the item sheet rendering too, though. So maybe just `Promise.all` like is done with advancements?

module/applications/item/split-stack-dialog.mjs
```diff
@@ -0,0 +1,75 @@
...
+  /* -------------------------------------------- */
+
+  /**
+   * Handle submission of the dialog.
+   * @this {SplitStackDialog}
+   * @param {Event|SubmitEvent} event    The form submission event.
+   * @param {HTMLFormElement} form       The submitted form.
+   * @param {FormDataExtended} formData  Data from the dialog.
+   */
+  static async #handleFormSubmission(event, form, formData) {
```
Fyorl, line 61: Could we factor out the internals to a method on `Item5e`? And then we call it here with: `doc.split(right)`. That would allow us a shortcut if quantity is 2. In the context menu split handler we can check `if ( quantity === 2 ) return item.split(1)`, otherwise we invoke this dialog. Not super necessary but some nice polish I suppose.

Fyorl (approved):

PR #5158, opened 2025-02-11
[#5133] Add HP bar to short rest dialog

### LINKED ISSUE #5133: Add HP bar to short rest dialog
Display the character's HP bar in the short rest dialog so players don't have to have their main character sheet open to tracking healing during rests. It would also be cool to have a little preview based on the selected hit die showing how much rolling that hit die might heal.

roth-michael:
Closes #5133 
I anticipate some changes to style (colors, primarily - not sure what the ideal dark mode appearance should be wrt the box shadow) and potentially the "preview" part of the hp bar, depending on what the "right" way to show potential regained HP is.

Also happened across a minor bug (`context.denomination` was being set, when it should've been `context.hitDice.denomination`), though the behavior with that fixed and the `_onChangeForm` addition is that when you try to select a "0 remaining" hit die, it doesn't let you. Which doesn't necessarily feel like the _wrong_ behavior, but worth bringing up.

### DIFF
--- a/less/v2/apps.less
+++ b/less/v2/apps.less
@@ -742,54 +742,6 @@
     + .meter-group { margin-block-start: .5rem; }
   }
 
-  .meter {
-    border: 1px solid var(--dnd5e-color-gold);
-    border-radius: 4px;
-    background: var(--meter-background);
-    font-family: var(--dnd5e-font-roboto);
-    font-weight: bold;
-    box-shadow: inset 0 0 16px rgb(0 0 0 / 25%);
-    position: relative;
-    display: flex;
-    align-items: center;
-    overflow: hidden;
-
-    &.progress, .progress {
-      --bar-percentage: 0%;
-      --border-width: 3px;
-      overflow: hidden;
-
-      &::before {
-        content: "";
-        position: absolute;
-        block-size: 100%;
-        inline-size: var(--bar-percentage);
-        box-shadow: 0 0 6px var(--dnd5e-shadow-45);
-        clip-path: polygon(0 0, calc(100% + 6px) 0, calc(100% + 6px) 100%, 0 100%);
-      }
-    }
-
-    .label {
-      inline-size: 100%;
-      display: flex;
-      align-items: center;
-      gap: .25rem;
-      position: relative;
-      padding-inline-start: 8px;
-      padding-inline-end: 12px;
-      color: var(--color-text-light-0);
-
-      &[hidden] { display: none; }
-      .value, .max { text-shadow: 0 0 4px var(--dnd5e-color-gold); }
-      .separator { color: var(--dnd5e-color-gold); }
-      .bonus {
-        margin-inline-start: auto;
-        font-size: var(--font-size-14);
-        opacity: .8;
-      }
-    }
-  }
-
   /* ---------------------------------- */
   /*  Slots                             */
   /* ---------------------------------- */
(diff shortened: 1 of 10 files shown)


module/applications/actor/rest/short-rest-dialog.mjs
```diff
@@ -99,6 +112,13 @@ export default class ShortRestDialog extends BaseRestDialog {
     this.render();
   }
 
+  _onChangeForm(formConfig, event) {
```
arbron, line 115:
```suggestion

  /* -------------------------------------------- */
  /*  Event Listeners and Handlers                */
  /* -------------------------------------------- */

  _onChangeForm(formConfig, event) {
```

templates/actors/rest/short-rest.hbs
```diff
@@ -25,6 +25,11 @@
                 <i class="fa-solid fa-dice-d20" inert></i>
             </button>
         </div>
+        <div class="healthbar">
+            <div class="progress-potential" style="width:{{progressBarPotential}}%"></div>
+            <div class="progress" style="width:{{progressBar}}%"></div>
+            <div class="overlay">{{healthBarText}}</div>
+        </div>
```
arbron, line 32: I'd say we want this styled more like the HP bar on the character sheet:

```html
        <div class="meter hit-points">
            <div class="progress hit-points" role="meter" <!-- aria stuff here -->
                 style="--bar-percentage: {{ hitPoints.pct }}">
                <div class="label">
                    <span class="value">{{ hitPoints.value }}</span>
                    <span class="seperator">/</span>
                    <span class="max">{{ hitPoints.effectiveMax }}</span>
                </div>
            </div>
        </div>
```

module/applications/actor/rest/short-rest-dialog.mjs
```diff
@@ -79,6 +80,18 @@ export default class ShortRestDialog extends BaseRestDialog {
       value: context.config.autoHD
     });
 
+    const numericalDenom = Number(context.hitDice.denomination?.slice(1));
+    if ( numericalDenom ) {
+      const { value: currHP, max: maxHP } = this.actor.system.attributes.hp;
+      context.healthBarText = `${game.i18n.localize("DND5E.HP")}: ${currHP} / ${maxHP}`;
+      context.progressBar = 100 * currHP / maxHP;
+      let potentialRegain = numericalDenom;
```
arbron, line 88: Should make sure that we are taking the characters con modifier into account here. Perhaps we can display the bar in such a way so that the minimum regained hit points (con mod + 1, min of 1) is shown in red, and the potential range based on the die roll is shown in a more transparent color.

less/v2/apps.less
```diff
@@ -1575,6 +1575,39 @@ dialog.dnd5e2.application {
...
+    }
+
+    .progress-potential {
+      z-index: 1;
+      background-color: #bb7777;
+    }
+
+    .progress {
+      z-index: 2;
+      background-color: #d93131;
```
arbron, line 1596: Probably want to use a green gradient to match the character sheet's HP bar.

arbron (changes requested):

roth-michael:
Pushed some improvement to the styles, example pic below:
image

Better (more on-theme) colors/transparencies can be found I think for the min/max potential gains, and when HP is low enough the label can get kinda funky (pic below). My gut says it's unclear what the bars represent, but part of that is because of how jarringly red the minimum one is, so maybe better colors will also make it clearer what the intent is.

image

krbz999:
Probably a darker shade of green for the "minimum".

roth-michael:
https://github.com/user-attachments/assets/5a515784-604f-4506-b6bb-32352f714d86

The animation feels effective. Currently after the hp bar (gradient identical to normal hp bar) there's the min recovery in the darkest green of that bar, and then a gradient from that darkest green to a different darkish green. The difference between where minimum ends and maximum starts is a bit unclear. Might consider the below as well:


https://github.com/user-attachments/assets/16440a32-8191-4752-a077-2a587af16095

Fyorl (approved):

PR #5159, opened 2025-02-11
Fix damage shortcut key bug introduced by #5015

roth-michael:
In #5015 I set `config.isCritical` to `false` or `true`, will only ever be `undefined` if there was no "last attack." Since `applyKeybindings` relied on `config.isCritical` being `undefined` in order to care about holding the critical key down, that functionality broke.

New functionality while clicking damage:
- Holding shift (by default) skips the dialog, doing whatever the default would be
- Holding ctrl (by default) skips the dialog, rolling _normal_ damage regardless of last attack
- Holding alt (by default) skips the dialog, rolling _critical_ damage regardless of last attack


arbron (changes requested):
This code looks okay, but you should change the keybinding name for `skipDialogDisadvantage` to make it clear that it will never roll a critical.

arbron (approved):

PR #5161, opened 2025-02-12
Avoid crash when drag-dropping facility to low-level actor

TheGiddyLimit:
This change removes a junk error from the log when attempting to drag-drop a facility to a low-level actor. Note that this makes no functional difference, as a level <5 actor will not trigger the "set built" conditional.

### DIFF
--- a/module/data/item/facility.mjs
+++ b/module/data/item/facility.mjs
@@ -243,7 +243,8 @@ export default class FacilityData extends ItemDataModel.mixin(ActivitiesTemplate
     const actor = this.parent?.parent;
     if ( !actor ) return;
     const { basic } = CONFIG.DND5E.facilities.advancement;
-    const [, available = 0] = Object.entries(basic).reverse().find(([level]) => level <= actor.system.details.level) || [];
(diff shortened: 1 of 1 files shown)


module/data/item/facility.mjs
```diff
@@ -243,7 +243,7 @@ export default class FacilityData extends ItemDataModel.mixin(ActivitiesTemplate
     const actor = this.parent?.parent;
     if ( !actor ) return;
     const { basic } = CONFIG.DND5E.facilities.advancement;
-    const [, available] = Object.entries(basic).reverse().find(([level]) => level <= actor.system.details.level);
+    const [, available = 0] = Object.entries(basic).reverse().find(([level]) => level <= actor.system.details.level) || [];
```
arbron, line 246: Line is a bit too long.

```suggestion
    const [, available = 0] = Object.entries(basic ?? {}).reverse()
      .find(([level]) => level <= actor.system.details.level) ?? [];
```

arbron (changes requested):

arbron (approved):

PR #5187, opened 2025-02-15
Allow Custom Die for d20-based Rolls

Larkinabout:
Adds handling for a custom die via an `options.customDie` property to allow modules/scripts to easily swap out the d20 with a different die face or number of dice.

```js
Hooks.on("dnd5e.preRollV2", (config, dialog, message) => {
  config.rolls[0].options.customDie = "2d10";
  config.rolls[0].options.criticalSuccess = 20;
  config.rolls[0].options.criticalFailure = 2;
});
```

### DIFF
--- a/module/dice/d20-die.mjs
+++ b/module/dice/d20-die.mjs
@@ -1,3 +1,5 @@
+import { getDieParts } from "../utils.mjs";
+
 const { Die } = foundry.dice.terms;
 
 /**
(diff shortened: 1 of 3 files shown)


arbron:
The general intent of the dice system is that module wishing to replace the D20 will replace the entire `D20Die` class with their own implementation where they decide what is considered a critical and how advantage and other roll details are applied.

Larkinabout:
Could something like the above be considered as a middle ground akin to the benefit to users and part-time developers of hooks and configs to customise the system?

If so, I'm happy to make alterations if what I've done doesn't quite fit.

Fyorl:
I think this change is too invasive. When maintaining this class and its consumers, we will constantly have to reason about whether the die term is actually a d20 or not. I don't have a suggestion for how to make it less invasive. The correct solution is to swap out the `D20Die` class globally, as Jeff suggested.

```js
CONFIG.Dice.D20Die = MyDie;
```

Issue #4553, opened 2024-10-18
SRD Review


davidR1974:
Stumbled up on odd things while reviewing SRD monster data & features

i sum up first the use cases will add creatures in the comments after..

1.  Many "Poisonous" creatures have the poison damage of their "Save" activity repeating the base damage and with the mod coming from the ability of the save. 
_Example : Wyvern Stinger > Save activity is 2d6+CON mod (because DC is based on CON) instead of 7d6_

2. Saves activities that apply conditions have the damage repeated (this may be intentional).
_Example : Boar > Charge  > Save has 1d6 as damage when the rule for that save is just "knock prone"_

3. Swarm creatures :
bites for swarm creatures includes a mod when the rules says they don't.
Could include a 1/2 damage activity on the swarm to take into account their 1/2 HP status

4. Attacks that deals 1 damage sometimes includes a mod
_Example : Weasel > Attack deals 5 damage instead of 1_

Others : 
Ankheg Bite > Miss Save Activity against grappled
Spirit Naga > Correct Poison damage correctly in the Attack (but without Save) and not in the Save
Winter Wolf > Cold Breath > Has Attack and Save (Save should be enough)
Specter > Life Drain is odd

davidR1974:
Save with damage (when they only should apply conditions)

Aboleth => Tentacle > Save 
Otyugh => Bite > Save  
Boar => Charge > Save 
Dire Wolf => Bite > Save
Elk => Charge > Save
Giant Boar => Charge > Save
Giant Crocodile => Tail > Save
Giant Elk => Charge > Save
Giant Goat => Charge > Save
Giant Rat (Diseased) => Bite > Save
Giant Sea Horse => Charge => Save
Mastiff => Bite / Save
Rhinoceros => Charge / Save
Wolf => Bite > Save
Couatl > Bite > Save
Solar => Slaying Longbow > Save
Clay Golem => Slam > Save
Air Elemental => Whirlwind > Save
Balor => Whip > Save
Bearded Devil => Beard / save
Bearded Devil => Glaive / save
Bone Devil => Sting / Save
Erynies => Longbow / Save
Horned Devil => Tail / Save (+ Utility should be Damage ?)
Ice Devil => Wall of Ice => Save inflict 10d6 damage (+ utility should be damage ?)
Quasit => Claw / Save

davidR1974:
1. Wrong damage for "poison" activity 

[...]

PR #5190, opened 2025-02-15
Fix damage type in Sling Bullets of Slaying, fix activities in Manacles, many adjustments to Otyugh.

### LINKED ISSUE #4336: Manacles SRD item has incorrect activities
One is a save, and the other is a check that isnt set up right

krbz999:
Fixes wrong damage type in Sling Bullets of Slaying.

Fixes stuff with Manacles item - closes #4336.

Fixes all kinds of stuff with Otyugh - related to #4553.


arbron:
I know how skeptical you are of YAML source files, but since the system has switched over this will need to be reworked on those rather than the JSON :).

Issue #5104, opened 2025-02-03
Update wiki for 4.3 release


arbron:
- [ ] Activity
  - [ ] Visibility options on check & save activities (if https://github.com/foundryvtt/dnd5e/pull/4815 is merged)
  - [ ] Bonus field on check & save activities (if https://github.com/foundryvtt/dnd5e/pull/4786 is merged)
- [x] Enchantments
  - [x] Enchantment format for replacing damage types (https://github.com/foundryvtt/dnd5e/pull/4913)
  - [x] Bonuses to Save DCs (see https://github.com/foundryvtt/dnd5e/pull/5018)
  - [x] Damage bonus (see https://github.com/foundryvtt/dnd5e/pull/5080)
- [ ] Enricher
  - [ ] Spellcasting as ability in check & save enrichers (if https://github.com/foundryvtt/dnd5e/pull/4883 is merged)
- [x] Module Registration
  - [x] Pack sorting option (https://github.com/foundryvtt/dnd5e/pull/5005)

krbz999:
Save DCs: https://github.com/foundryvtt/dnd5e/pull/5018

PR #5206, opened 2025-02-16
Update Roll Formulas for 4.3

roth-michael:
Adding on to  #5104 
Don't think I missed any of the changes, only added `attributes.spell.attack`.


wiki/Roll-Formulas.md
```diff
@@ -168,7 +170,7 @@
 
 `@classes.*.hitDice` - The hit die denomination of the class, e.g., 'd8'.
```
krbz999, line 171: hd.denomination

arbron (approved):

PR #5207, opened 2025-02-16
[#5196] Adds handling to `replaceFormulaData` to parse shims

### LINKED ISSUE #5196: @abilities.*.save not working
As the title says the @abilities.*.save doesn't work at all, when tried to roll with that bonus this error pops up: ${"roll":{"min":null,"max":null","mode":0},"value":x}$

arbron:
To get around core limitation that prevents a `toString` shim on plain objects when used in roll formulas, this allows `toJSON` to be used as the shim. Since that results in the final value being surrounded by `$`, this looks for the pattern `$!!$` and replaces it with just the value between the marks (the bang has been added to ensure this is a deliberate shim from the system, not a unintentional usage that should show an error).

Closes #5196

### DIFF
--- a/module/data/actor/templates/common.mjs
+++ b/module/data/actor/templates/common.mjs
@@ -166,7 +166,12 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
       abl.save.toString = function() {
         foundry.utils.logCompatibilityWarning("The 'abilities.<ability>.save' property is now stored in "
           + "'abilities.<ability>.save.value'.", { since: "4.3", until: "4.5" });
-        return abl.save.value;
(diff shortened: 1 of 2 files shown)


arbron:
An alternate solution to this is to modify the `initialize` method `RollConfigField` to return a custom class. The advantage of this approach is it gives us path to shim future methods in the future where it might be harder to use a custom class.

Fyorl (commented):
Shouldn't really need to 'get around' a 'core limitation' here. The JSON serialisation was added for exactly this purpose: To allow systems some recourse for including complex objects and even arrays in their roll data.

If we want to settle on using `$!!$` as a way to deliver roll data deprecations, that seems pretty reasonable to me.

PR #5209, opened 2025-02-17
Store scale value data in a sparse format

### LINKED ISSUE #2187: Add `uses` scale value type
A number of scale values track item uses (channel divinity for cleric, action surge & indomitable for fighter, etc.). It would be useful if the rest value could be displayed along with the scale value:

**Current:**
Screenshot 2023-02-11 at 15 16 46

**Suggested:**
Screenshot 2023-02-11 at 15 18 35

arbron:
Changes how scale value data is stored from storing all keys for each level to just storing the keys that were changed at a given level.

```javascript
// Original
{
  1: { number: 1, faces: 6 },
  5: { number: 1, faces: 8 },
  11: { number: 1, faces: 10 }
}

// New
{
  1: { number: 1, faces: 6 },
  5: { faces: 8 },
  11: { faces: 10 }
}
```

Scale Value Sparse

This has the advantage of making it easier to change part of a value. For example, if you had a scale value like Sneak Attack that changes every other level and you wanted to modify it from one die to two then it currently would require changing that one in 10 places. With this change the number of dice would only need to be updated at level 1.

### DIFF
--- a/module/applications/advancement/scale-value-config.mjs
+++ b/module/applications/advancement/scale-value-config.mjs
@@ -102,6 +102,10 @@ export default class ScaleValueConfig extends AdvancementConfig {
    * @param {*} lastValue  The previous value.
    */
   _mergeScaleValues(value, lastValue) {
+    foundry.utils.logCompatibilityWarning(
+      "`ScaleValueConfig#_mergeScaleValues` has been deprecated without replacement.",
+      { since: "DnD5e 6.0", until: "DnD5e 6.2" }
+    );
     for ( const k of Object.keys(lastValue ?? {}) ) {
       if ( value[k] == null ) value[k] = lastValue[k];
     }
--- a/module/applications/advancement/scale-value-config.mjs
+++ b/module/applications/advancement/scale-value-config.mjs
@@ -33,6 +33,19 @@ export default class ScaleValueConfig extends AdvancementConfig {
   };
 
   /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Range of levels that can be used based on what item type this advancement is within.
+   * @type {number[]}
+   */
+  get levelRange() {
+    let levels = Array.fromRange(CONFIG.DND5E.maxLevel + 1);
+    return ["class", "subclass"].includes(this.advancement.item.type) ? levels.slice(1) : levels;
+  }
+
+  /* -------------------------------------------- */
   /*  Rendering                                   */
   /* -------------------------------------------- */
 
@@ -69,21 +82,14 @@ export default class ScaleValueConfig extends AdvancementConfig {
    * @protected
    */
   _prepareLevelData() {
-    let lastValue = null;
-    let levels = Array.fromRange(CONFIG.DND5E.maxLevel + 1);
-    if ( ["class", "subclass"].includes(this.advancement.item.type) ) levels = levels.slice(1);
-    return levels.reduce((obj, level) => {
-      const value = this.advancement.configuration.scale[level]?.clone();
+    const TypeClass = TYPES[this.advancement.configuration.type];
+    return this.levelRange.reduce((obj, level) => {
+      const value = new TypeClass(this.advancement.configuration.scale[level] ?? {});
+      const lastValue = this.advancement.valueForLevel(level - 1);
       obj[level] = {
-        fields: TYPES[this.advancement.configuration.type].getFields(level, value, lastValue),
-        value: null
+        fields: TypeClass.getFields(level, value, lastValue),
+        value: this.advancement.configuration.scale[level] ?? {}
       };
-      if ( value ) {
-        this._mergeScaleValues(value, lastValue);
-        obj[level].className = "new-scale-value";
-        obj[level].value = value;
-        lastValue = value;
-      }
       return obj;
     }, {});
   }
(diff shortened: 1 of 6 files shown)


module/data/advancement/_types.mjs
```diff
@@ -84,11 +84,11 @@
...
  * @typedef ScaleValueAdvancementConfigurationData
- * @property {string} identifier        Identifier used to select this scale value in roll formulas.
- * @property {string} type              Type of data represented by this scale value.
+ * @property {string} identifier             Identifier used to select this scale value in roll formulas.
+ * @property {string} type                   Type of data represented by this scale value.
  * @property {object} [distance]
- * @property {string} [distance.units]  If distance type is selected, the units each value uses.
- * @property {Object<string, *>} scale  Scale values for each level. Value format is determined by type.
+ * @property {string} [distance.units]       If distance type is selected, the units each value uses.
+ * @property {Object<string, object>} scale  Sparse scale values for each level.
```
Fyorl:
```suggestion
 * @property {Record<string, object>} scale  Sparse scale values for each level.
```

module/data/advancement/scale-value.mjs
```diff
@@ -97,6 +96,11 @@ export class ScaleValueEntryField extends foundry.data.fields.ObjectField {
  * @mixes ScaleValueStringTypeData
  */
 export class ScaleValueType extends foundry.abstract.DataModel {
+  constructor(data = {}, options = {}) {
```
Fyorl:
```suggestion
  constructor(data={}, options={}) {
```

module/applications/advancement/scale-value-config.mjs
```diff
@@ -128,61 +134,57 @@ export default class ScaleValueConfig extends AdvancementConfig {
...
+
+    let lastValue = {};
+    const scale = {};
+    for ( const level of this.levelRange ) {
+      const value = configuration.scale[level] ?? {};
+      scale[level] = {};
+
+      // Fetch data for valid keys, skipping if value is empty or the same as a previous level
+      for ( const key of validKeys ) {
+        if ( !value[key] || (value[key] === lastValue[key]) ) scale[level][key] = _del;
```
Fyorl: Is `0` a valid value, or we treat it as blank? If blank, maybe we need to change the `??=` to `||=` in `valueForLevel`. If not blank, we need something like this and might also need to revisit the use of `filteresKeys` in the `ScaleValueType` constructor.

```suggestion
        if ( (value[key] == null) || (value[key] === lastValue[key]) ) scale[level][key] = _del;
```

module/data/advancement/scale-value-data.mjs
```diff
@@ -34,7 +34,7 @@ export class ScaleValueConfigurationData extends foundry.abstract.DataModel {
       identifier: new IdentifierField({ required: true }),
       type: new StringField({ required: true, initial: "string", choices: TYPES }),
       distance: new SchemaField({ units: new StringField({ required: true }) }),
-      scale: new MappingField(new ScaleValueEntryField(), { required: true })
+      scale: new MappingField(new ObjectField(), { required: true })
```
Fyorl, line 37: Seems `ScaleValueEntryField` is now unused. I guess we're still exporting it to avoid breakage?

module/applications/advancement/scale-value-config.mjs
```diff
@@ -69,21 +82,14 @@ export default class ScaleValueConfig extends AdvancementConfig {
...
+      const value = new TypeClass(this.advancement.configuration.scale[level] ?? {});
+      const lastValue = this.advancement.valueForLevel(level - 1);
       obj[level] = {
-        fields: TYPES[this.advancement.configuration.type].getFields(level, value, lastValue),
-        value: null
+        fields: TypeClass.getFields(level, value, lastValue),
+        value: this.advancement.configuration.scale[level] ?? {}
       };
-      if ( value ) {
-        this._mergeScaleValues(value, lastValue);
```
Fyorl, line 82: `_mergeScaledValues` also unused now. It's not annotated as `@protected` or `@internal`. I guess we are keeping it to avoid breakage?

Fyorl (approved):
Looks good, just noting some places we have dead code now, which includes `ScaleValueAdvancement#testEquality`. Not sure if they were not considered, or they were considered and we are keeping them to avoid breakage.

Fyorl (changes requested):
Need to retract that approval actually. I tried testing with SRD 5.2 Rogue's Sneak Attack scale value advancement, and I'm not able to edit any of the values. They return to their previous value on save. Also, maybe this is just an issue with my test environment but after importing the Rogue class into the world, its scale value at 7th was missing. Went directly from 3d6 at 5th to 5d6 at 9th.

module/applications/advancement/scale-value-config.mjs
arbron: `0` is valid, but that change is more recent than the latest rebase of this PR :)

arbron:
> Need to retract that approval actually. I tried testing with SRD 5.2 Rogue's Sneak Attack scale value advancement, and I'm not able to edit any of the values. They return to their previous value on save. Also, maybe this is just an issue with my test environment but after importing the Rogue class into the world, its scale value at 7th was missing. Went directly from 3d6 at 5th to 5d6 at 9th.

I tested after rebasing and fixing some issues and Sneak Attack is looking good to me, let me know if you notice any other issues.

Fyorl (approved):
> I tested after rebasing and fixing some issues and Sneak Attack is looking good to me, let me know if you notice any other issues.

Yeah I think it was just missing a rebase.

PR #5213, opened 2025-02-17
[#5014] Modify usage chat cards to include activity name & icon

### LINKED ISSUE #5014: Feature Request: An option for Activity names and Icons to appear in Item Cards instead of the base item's
I would like to see an option that allows for an Item Card in chat created by an Activity to use the name/icon of that Activity, rather than the item the Activity is attached to.

For example, let's say a player has a magic "Longsword +1" that can create a Dazzling Flare twice per day, blinding creatures within 10 feet on a failed save. That ability is made as an activity of the sword with limited uses per day. When rolling that activity, displaying "Longsword +1" and the longsword icon in the chat card is not immediately clear what is occurring or why there is a saving throw button--longswords don't usually prompt saving throws. Instead, displaying the activity name "Dazzling Flare" and its associated icon would be much more visually evident at a glance, especially when scrolling back through the chat some time after it occurred.

(Apologies if this has been requested before. I did a search and found nothing, but I might have messed up my parameters).

roth-michael:
Closes #5014 
When an activity is used (or a chat action from an activity is used):
1. Add activity name to the chat card title
2. Replace the icon, if set on the activity

Example below with a longsword. The Attack activity has no icon set, but the Save activity does.
Screenshot 2025-02-17 111259

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1069,6 +1069,9 @@
   },
   "TURN": {
     "NoCombatant": "Combatant no longer exists!"
+  },
+  "USAGE": {
+    "Title": "{item}: {activity}"
   }
 },
 
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -887,6 +887,8 @@ export default function ActivityMixin(Base) {
         const { spellLevels, spellSchools } = CONFIG.DND5E;
         data.subtitle = [spellLevels[spellLevel], spellSchools[this.item.system.school]?.label].filterJoin(" &bull; ");
       }
+      const img = (this.img === this.metadata.img) ? this.item.img : this.img;
+      const title = game.i18n.format("DND5E.CHATMESSAGE.USAGE.Title", { item: this.item.name, activity: this.name});
 
       return {
         activity: this,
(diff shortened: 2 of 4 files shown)


roth-michael:
A note/question on this, actually:
Might make sense if this logic only happens for items which have multiple activities, or even more specifically if the activity used isn't the first/default activity. That way simple items which just have one activity aren't needlessly specific in their chat cards.

krbz999:
Best of both worlds could be to have it be a subtitle or tag below the regular name.

arbron:
@roth-michael Wanna give a comment on https://github.com/foundryvtt/dnd5e/issues/5014 so I can assign it to you?

roth-michael:
Potential handling for activity name:
- As in screenshots above, "Item Name: Activity Name"
  - Raises the question above of whether this should happen on a one-activity, default-activity-name item
- Tag for activity name above, for instance, "Martial Melee" in above
- That, but swap item name and activity name
- Full on bit of text, same size as item name, beneath item name

Icon question: Use item icon if default activity icon? Or always sub in activity icon.

arbron:
This is being handled during the whole chat card redesign that Kim is doing, so I'm going to close this PR.

PR #5228, opened 2025-02-19
[#4708] Deprecate `AdvancementSelection` in favor of `createDialog`

arbron:
Deprecates the V1 `AdvancementSelection` dialog in favor of using `PseudoDocument#createDialog` same as activities. To fully support using advancements with this method a few changes were made:
- Added support for proper coloring of SVG icons
- Added optional `hint` that is added as tooltip
- Added ability to disable options without removing them
- Split some logic out from the `createDialog` method into `_createDialogTypes` and `_createDialogData` methods so each type can customize the dialog behavior

Advancement Creation Dialog


module/documents/mixins/pseudo-document.mjs
```diff
@@ -318,6 +314,36 @@ export default function PseudoDocumentMixin(Base) {
...
       });
     }
+
+    /* -------------------------------------------- */
+
+    /**
+     * Prepare the data needed for the creation dialog.
+     * @param {string} type  Specific type of the PseudoDocument to prepare.
+     * @param {Item5e} parent  Parent document within which this PseudoDocument will be created.
+     * @returns {{ type: string, label: string, icon: string, hint: [string], disabled: [boolean] }}
```
Fyorl, line 324:
```suggestion
     * @returns {{ type: string, label: string, icon: string, [hint]: string, [disabled]: boolean }}
```

Fyorl (approved):

PR #5260, opened 2025-02-24
[#5259] Add temp HP formula to transformation, replacing `addTemp`

### LINKED ISSUE #5259: Add temp HP formula to transformation settings
Currently the temp HP for the 2024 Wild Shape is handled using a system-provided `dnd5e.transformActor` hook. If we added a place for a temp HP formula rather than the current "Add Temp" checkbox it would allow for defining this in the transformation dialog/preset so we can remove that hook. The formula field would also work for the polymorph preset provided both the target and source actor data is available in roll data.

arbron:
Adds a new formula field to define how much temp HP the transformed actor gets. The roll data in this field provides both the target actor's roll data as well as the source actor. This allows the polymorph preset formula to be `@source.attributes.hp.max`.

Polymorph Temp Formula

This PR also adds information on subclasses to the roll data, using a `Proxy` setup that allows for determining whether a specific subclass exists in roll data (so `@subclasses.moon` will resolve to `1` if the player has the Moon Druid subclass, or `0` if they do not). This allows setting the 2024 wild shape temp HP formula to be `max(@classes.druid.levels, @subclasses.moon.levels * 3)`.

Closes #5259

### DIFF
--- a/module/data/actor/templates/creature.mjs
+++ b/module/data/actor/templates/creature.mjs
@@ -274,40 +274,16 @@ export default class CreatureTemplate extends CommonTemplate {
   /** @inheritDoc */
   getRollData({ deterministic=false }={}) {
     const data = super.getRollData({ deterministic });
-
-    const ItemData = class {
-      constructor(data) {
-        this.#exists = !!data;
-        Object.assign(this, data);
+    data.classes = {};
+    data.subclasses = {};
+    for ( const [identifier, cls] of Object.entries(this.parent.classes) ) {
+      data.classes[identifier] = {...cls.system};
+      data.classes[identifier].hitDice = cls.system.hd.denomination; // Backwards compatibility
+      if ( cls.subclass ) {
+        data.classes[identifier].subclass = cls.subclass.system;
+        data.subclasses[cls.subclass.identifier] = { levels: cls.system.levels };
       }
-
-      #exists = false;
-
-      toString() { return this.#exists ? "1" : "0"; }
-    };
-
-    data.classes = new Proxy(this.parent, {
-      get(target, prop, receiver) {
-        const cls = target.classes[prop];
-        if ( !cls ) return new ItemData();
-        const data = { ...cls.system, hitDice: cls.system.hd.denomination };
-        if ( cls.subclass ) data.subclass = cls.subclass.system;
-        return new ItemData(data);
-      },
-      has(target, prop) {
-        return true;
-      }
-    });
-
-    data.subclasses = new Proxy(this.parent, {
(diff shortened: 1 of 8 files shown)


module/documents/actor/actor.mjs
```diff
@@ -2875,6 +2878,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       mergeSkills: settings.merge.has("skills")
     };
     const source = target.toObject();
+    const rollData = { ...this.getRollData(), source: target.getRollData() };
```
Fyorl, line 2881: The whole source/target nomenclature is very confusing in the context of transformation. I think in this method the `target` is the transformation target, i.e. the thing we are trying to transform into, because the method is called `transformInto`, but it could also easily mean the *target of the transformation effect* as well. We also call the thing being transformed the 'origin', which is synonymous with 'source', but then we call the `target`'s data `source`.

So I understand why we call this `source` in the roll data, but I'd like to straighten out the nomenclature so we can pick the best name for this roll data in case we have additional uses for it as part of the transform activity for example.

I think for the thing being transformed, `host` is reasonably unambiguous. Do we want to stick with `source` then for the thing being transformed into? I don't think we should use `target` because it's too ambigious. Do you have any better suggestions for `source` or `host`?

Fyorl (changes requested):

module/documents/actor/actor.mjs
arbron: Nothing better I can think of in the moment.

Fyorl: Alright, let's stick with `source` then.

module/data/actor/templates/creature.mjs
```diff
@@ -274,12 +274,40 @@ export default class CreatureTemplate extends CommonTemplate {
...
   getRollData({ deterministic=false }={}) {
     const data = super.getRollData({ deterministic });
-    data.classes = {};
-    for ( const [identifier, cls] of Object.entries(this.parent.classes) ) {
-      data.classes[identifier] = {...cls.system};
-      data.classes[identifier].hitDice = cls.system.hd.denomination; // Backwards compatibility
-      if ( cls.subclass ) data.classes[identifier].subclass = cls.subclass.system;
-    }
+
+    const ItemData = class {
```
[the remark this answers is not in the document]
Fyorl: Even then, you can use `sign(@subclasses.moon.levels)`, so I'm not sure about adding all of this just to make such formulae a little nicer.

Fyorl (approved):

PR #5265, opened 2025-02-24
Implement UI for pending livestock trades.

Fyorl:
Clears out some remaining TODOs from the Bastion system. Pending livestock trades will now show with a special icon, and deleting them will invalidate the trade order entirely. Sold livestock are removed immediately and marked as pending, while bought livestock are marked as pending until the bastion order is complete, at which point they appear in the Stable.

image

### DIFF
--- a/less/v2/apps.less
+++ b/less/v2/apps.less
@@ -648,17 +648,22 @@
       overflow: visible;
       cursor: pointer;
 
-      &.empty {
+      &.empty, &.pending {
         display: inline-grid;
         place-content: center;
 
         &::before {
           font-family: var(--font-awesome);
           font-weight: bold;
-          opacity: .15;
         }
       }
 
+      &.pending::before {
+        content: "\f51e";
+        color: var(--dnd5e-color-gold);
+      }
+
+      &.empty::before { opacity: .15; }
       &.hireling.empty::before { content: "\f007"; }
       &.defender.empty::before { content: "\f132"; }
       &.creature.empty::before { content: "\f7ab"; }
@@ -828,16 +833,21 @@
 
   &#bastion-turn {
     pointer-events: all;
-    margin: 0 5px 10px 15px;
+    margin: 0 0 -8px;
     color: var(--color-text-light-highlight);
-    width: var(--players-width);
+    width: var(--players-width, 200px);
     background: rgb(0 0 0 / 80%);
     border: 1px solid var(--color-border-dark);
+    cursor: var(--cursor-pointer), pointer;
+    height: 40px;
 
     &:hover {
       border-color: var(--color-border-highlight-alt);
       box-shadow: 0 0 10px var(--color-shadow-highlight);
     }
+
+    /* TODO: Remove when v12 is dropped */
+    .v12 { margin: 0 5px 10px 15px; }
   }
 
   /* ---------------------------------- */
--- a/module/applications/actor/character-sheet-2.mjs
+++ b/module/applications/actor/character-sheet-2.mjs
@@ -514,6 +514,7 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
     const { index } = event.target.closest("[data-index]")?.dataset ?? {};
     const facility = this.actor.items.get(facilityId);
     if ( !facility || !prop || (index === undefined) ) return;
+    if ( event.target.closest(".occupant-slot.pending") ) return this._onRemovePendingTrade(facility);
     let { value } = foundry.utils.getProperty(facility, prop);
     value = value.filter((_, i) => i !== Number(index));
     return facility.update({ [`${prop}.value`]: value });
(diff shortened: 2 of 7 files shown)


module/documents/activity/order.mjs
```diff
@@ -164,10 +164,13 @@ export default class OrderActivity extends ActivityMixin(OrderActivityData) {
...
-        : system.trade.creatures.value.concat(creatures);
+
+      // Sold livestock are removed immediately. Bought livestock remain pending until the order is complete.
+      if ( trade.sell ) {
+        updates["system.trade.creatures.value"] = system.trade.creatures.value.filter((_, i) => {
+          return !trade.creatures.sell[i];
+        });
+      }
     }
   }
```
arbron, line 175: This change looks good, but I noticed in testing that line `273` throws a deprecation warning in V12 due to passing `fromUuid` directly into `map`.

```javascript
else creatures.push(...await Promise.all(trade.creatures.buy.filter(_ => _).map(u => fromUuid(u))));
```

less/v2/apps.less
```diff
@@ -828,16 +833,21 @@
...
+    cursor: var(--cursor-pointer), pointer;
+    height: 40px;
 
     &:hover {
       border-color: var(--color-border-highlight-alt);
       box-shadow: 0 0 10px var(--color-shadow-highlight);
     }
+
+    /* TODO: Remove when v12 is dropped */
+    .v12 { margin: 0 5px 10px 15px; }
```
arbron, line 850:
```suggestion
    &.v12 { margin: 0 5px 10px 15px; }
```

module/data/item/facility.mjs
```diff
@@ -208,6 +208,9 @@ export default class FacilityData extends ItemDataModel.mixin(ActivitiesTemplate
 
     if ( this.disabled ) this.progress.order = "repair";
 
+    // Trade
+    if ( Number.isFinite(this.trade.creatures.max) ) this.trade.creatures.max -= this.trade.pending.creatures.length;
```
arbron, line 212: I think this is hitting into a double-preparation bug in V12, ending up with the max creatures being negative whenever a trade is in progress and preventing any creatures from actually being added when it is complete.

```suggestion
    if ( Number.isFinite(this.trade.creatures.max) ) {
      this.trade.creatures.max = this._source.trade.creatures.max -= this.trade.pending.creatures.length;
    }
```

module/documents/actor/bastion.mjs
```diff
@@ -284,11 +284,16 @@ export default class Bastion {
       if ( (trade.pending.value === null) && trade.pending.stocked ) updates["system.trade.stock.stocked"] = true;
 
       // Bought goods
-      else if ( trade.pending.value !== null && !trade.pending.creatures.length ) {
-        updates["system.trade.stock.value"] = Math.min(trade.stock.value + trade.pending.value, trade.stock.max);
+      else if ( trade.pending.value !== null ) {
+        if ( trade.pending.creatures.length ) {
+          const creatures = trade.creatures.value.concat(trade.pending.creatures);
+          const diff = trade.creatures.max - creatures.length;
+          if ( diff < 0 ) creatures.splice(diff);
```
arbron, line 291: I think this logic is a bit off, because if you are buying 2 creatures the max is `1` (`3` normal - `2` pending) and so this would be negative, resulting in only one creature being purchased.

arbron (changes requested):
It will be good to drop V12 support :)

arbron:
We might also want to hit the "Advance Bastion Turn" button in V13 with some of that fading magic, because as-is it stands out from the otherwise faded UI:

Screenshot 2025-02-24 at 15 55 47

module/applications/actor/character-sheet-2.mjs
```diff
@@ -542,6 +543,35 @@ export default class ActorSheet5eCharacter2 extends ActorSheetV2Mixin(ActorSheet
...
   /* -------------------------------------------- */
 
+  /**
+   * Handle removing a pending trade from a facility.
+   * @param {Item5e} facility  The target facility.
+   * @returns {Promise<void|Item5e>}
+   * @protected
+   */
+  async _onRemovePendingTrade(facility) {
+    const result = await foundry.applications.api.DialogV2.confirm({
```
krbz999, line 553: Total aside, feels like there should be a system-implementation of `DialogV2`. Probably not doing more than just adding some of the styling.

arbron: Yep, unfortunately I already used the name `Dialog5e` for a slightly different V2 dialog, so we'll have to find a new name :(

krbz999: This is why we can't have nice things.

module/documents/actor/bastion.mjs
Fyorl: Oops, thank you. Forgot to edit this after making the data prep changes.

Fyorl: Just removed this check. Since the `max` is a function of `pending`, and `value` can't be greater than `max`, any arithmetic we do here is just redundant.

arbron (approved):

PR #5269, opened 2025-02-25
Pass number of concentration effects to statuses.concentrating

### LINKED ISSUE #5264: [Enhancement] - Concentration on more than one items indicated as the relevant number in actor.getRollData().statuses.concentrating
`@statuses.concentrating`  when concentrating on 2 spells for example, would be returning 2

thatlonelybugbear:
Closes #5264

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -320,7 +320,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     data.name = this.name;
     data.statuses = {};
     for ( const status of this.statuses ) {
-      data.statuses[status] = status === "exhaustion" ? this.system.attributes?.exhaustion ?? 1 : 1;
+      data.statuses[status] = 
+        status === "exhaustion" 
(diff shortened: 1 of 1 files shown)


arbron (approved):

arbron:
@thatlonelybugbear Could use a rebase onto the latest `4.4` to get rid of that extra commit

arbron (changes requested):
Could you give this a rebase on the latest `4.4.x` branch?

Also, give your spacing a quick look over, it looks like there might be an extra space on the first line.

arbron (approved):

Fyorl:
`getRollData` is called during Actor preparation, are we sure there aren't any order-of-operation issues lurking here? Also `Actor5e#concentration` can be quite expensive, especially because it can perform full Item initialisation in some cases, and its results are not cached at all. I'd not expect it to be an issue since the vast majority of Actors concentrate on 0 or 1 thing (0 being the most common), but I wanted to bring it up regardless in case it hadn't been considered.

PR #5271, opened 2025-02-25
[#5171, #5173, #5174] Add advantage to abilities, skills & tools

### LINKED ISSUE #5171: Add method for combining advantage fields properly (so when you roll an athletics check, we can combine `abilities.str.roll.mode` and `skills.ath.roll.mode` and respect all advantages/disadvantages for each field)

### LINKED ISSUE #5173: Display advantage config for abilities checks, ability saves, skills, & tools

arbron:
Adds the ability to configure roll mode, minimum, and maximum to ability checks & saves, and roll mode for skills & tools.

Adds a new method `AdvantageModeField#combineFields` that takes the key paths for two or more fields on a model and combines their advantage counts to find a final advantage mode.

This new method is then used when rolling skill checks, tool checks, or concentration saves to find the final advantage mode.

Closes #5171
Closes #5173

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -4,6 +4,7 @@ import SkillToolRollConfigurationDialog from "../../applications/dice/skill-tool
 import PropertyAttribution from "../../applications/property-attribution.mjs";
 import ActivationsField from "../../data/chat-message/fields/activations-field.mjs";
 import { ActorDeltasField } from "../../data/chat-message/fields/deltas-field.mjs";
+import AdvantageModeField from "../../data/fields/advantage-mode-field.mjs";
 import { _applyDeprecatedD20Configs, _createDeprecatedD20Config } from "../../dice/d20-roll.mjs";
 import { createRollLabel } from "../../enrichers.mjs";
 import parseUuid from "../../parse-uuid.mjs";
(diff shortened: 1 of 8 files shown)


module/documents/actor/actor.mjs
```diff
@@ -1168,22 +1169,28 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     }
 
     const relevant = type === "skill" ? this.system.skills?.[config.skill] : this.system.tools?.[config.tool];
+    const abilityId = relevant?.ability ?? (type === "skill" ? skillConfig.ability : toolConfig.ability);
+    const ability = this.system.abilities?.[abilityId];
     const hostActor = this.isPolymorphed && this.flags?.dnd5e?.transformOptions?.mergeSkills && (type === "skill")
       ? game.actors.get(this.flags.dnd5e?.originalActor) : null;
     const buildConfig = this._buildSkillToolConfig.bind(this, type, hostActor);
 
+    const { advantage, disadvantage } = AdvantageModeField.combineFields(this.system, [
```
Fyorl, line 1177: Do we need to combine modes here? We will ultimately invoke `#rollD20Test` which will check the advantage mode on the ability. Is it enough to get `advantage`, `disadvantage`, `minimum`, and `maximum` from the skill/tool here, then forward it to `#rollD20Test` which will check against the ability and combine it with the `advantage`, `disadvantage`, `minimum`, and `maximum` that it gets passed?

Same goes for concentration.

Fyorl (changes requested):

module/documents/actor/actor.mjs
arbron: Currently `#rollSkillTool` doesn't call into `#rollD20Test` at all, so it wouldn't work there.

For concentration, we could probably do that but it will require some changes to `BasicRoll.#mergeConfigs` to add some special handling for the advantage and range values because the simple `mergeObject` we are using for options now won't handle that properly.

Fyorl: Looks like we're due another refactor then since we failed to accomplish one of the core objectives, which was to have our roll pipeline resemble that of the rules. It doesn't make sense for us to be gathering all of the ability check terms in `#rollSkillTool` and `_buildSkillToolConfig`. Those terms should be handled in `rollAbilityCheck`, since a skill check is an ability check, which is a D20 Test. The structure of the roll pipeline is made obvious here, to me at least.

Anyway, I suppose that's tangential to the feature we're implementing in this PR.

arbron: I've improved how we merge roll options, allowing us to remove the special handling for advantage and roll ranges in `rollConcentration` at least.

Fyorl: Alright, thanks. I noticed in testing that initiative rolls do not interact with dexterity check advantage, presumably because it isn't routed through `rollAbilityCheck`. That's at least somewhat excusable given how initiative rolls work, so let's add it to `getInitiativeRollConfig` for now I suppose.

arbron: Handled

Fyorl (approved):

Issue #4916, opened 2024-12-21
`BaseActivity#getDamageConfig` ignores `scaling` property.


krbz999:
There is no change in any data returned by this method when supplying `scaling: 0` or `scaling: 9` or any other number. The roll data `scaling` property always has `increase: 0` and the damage rolls' parts are identical.

PR #5273, opened 2025-02-25
HealActivity scaling fix

### LINKED ISSUE #5272: [v4.3.4] - HealActivity doesn't scale.
Tested with Cure Wounds from PHB. Using higher spell slots does not scale healing rolls,

_same issue_ <https://github.com/foundryvtt/dnd5e/issues/4916>

thatlonelybugbear:
reinstate scaling for HealActivity, based on the 4.3.4 fix for damage <https://github.com/foundryvtt/dnd5e/issues/4916>

Closes #5272


module/data/activity/heal-data.mjs
```diff
@@ -45,7 +45,7 @@ export default class HealActivityData extends BaseActivityData {
   getDamageConfig(config={}) {
     if ( !this.healing.formula ) return foundry.utils.mergeObject({ rolls: [] }, config);
 
-    const rollConfig = foundry.utils.mergeObject({ critical: { allow: false }, scaling: 0 }, config);
+    const rollConfig = foundry.utils.deepClone(config);
```
arbron, line 48: Well we shouldn't be pulling out the critical part of that just to fix the scaling bug.

```suggestion
    const rollConfig = foundry.utils.mergeObject({ critical: { allow: false } }, config);
```

arbron (changes requested):

arbron:
Also you got an extra commit in there from the `4.4.x` branch that will prevent this from being merged into `4.3.x`.

PR #5289, opened 2025-02-28
[#5164] Support other advancement types on features

### LINKED ISSUE #5164: [Feature Request] Add Traits, Choose Items, Grant Items, & Scale Values to Feats
Much like how you can now grant ability score improvements to Feats, it would be nice to allow us to also add Traits, Item Choices, Granting Items, and Scale Values, much like you can with Backgrounds. This would make automating feats like **Skilled** a lot easier.

arbron:
Adds enables `ItemChoice`, `ItemGrant`, `ScaleValue`, and `Trait` advancement types to be added to `feat` items.

To better support grouping on the actor sheet, this also adds a `ultimateOrigin` flag to items added via advancement that points to the origin item in the advancement chain. So if a class adds a feat which adds its own feature, then the `advancementOrigin` on the feature will point to the feat, while the `ultimateOrigin` will point to the class. A small exception to this logic has been added for subclasses added directly by classes so that subclass features point back to the subclass, not the class.

A point of future work is adding some control over the levels on feat advancement. Currently the levels for those advancement types refer to the character level, even if it is on a feature added by a class. Potentially an option could be added to advancements on feats that controls how their level information is determined.

Closes #5164

### DIFF
--- a/module/documents/advancement/advancement.mjs
+++ b/module/documents/advancement/advancement.mjs
@@ -307,11 +307,13 @@ export default class Advancement extends PseudoDocumentMixin(BaseAdvancement) {
     const source = await fromUuid(uuid);
     if ( !source ) return null;
     const { _stats } = game.items.fromCompendium(source);
+    const advancementOrigin = `${this.item.id}.${this.id}`;
     return source.clone({
       _stats,
       _id: id ?? foundry.utils.randomID(),
       "flags.dnd5e.sourceId": uuid,
-      "flags.dnd5e.advancementOrigin": `${this.item.id}.${this.id}`
+      "flags.dnd5e.advancementOrigin": advancementOrigin,
+      "flags.dnd5e.ultimateOrigin": this.item.getFlag("dnd5e", "ultimateOrigin") ?? advancementOrigin
     }, { keepId: true }).toObject();
   }
 
--- a/module/applications/actor/character-sheet.mjs
+++ b/module/applications/actor/character-sheet.mjs
@@ -91,7 +91,7 @@ export default class ActorSheet5eCharacter extends ActorSheet5e {
 
       // Item grouping
       ctx.ungroup = "passive";
-      const [originId] = item.getFlag("dnd5e", "advancementOrigin")?.split(".") ?? [];
+      const [originId] = (item.flags.dnd5e?.ultimateOrigin ?? item.flags.dnd5e?.advancementOrigin)?.split(".") ?? [];
       const group = this.actor.items.get(originId);
       switch ( group?.type ) {
         case "race": ctx.group = "race"; break;
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -4390,7 +4390,7 @@ DND5E.activityTypes = {
  * @property {boolean} [hidden]                  Should this advancement type be hidden in the selection dialog?
  */
 
-const _ALL_ITEM_TYPES = ["background", "class", "race", "subclass"];
+const _ALL_ITEM_TYPES = ["background", "class", "feat", "race", "subclass"];
 
 /**
  * Advancement types that can be added to items.
--- a/module/documents/advancement/subclass.mjs
+++ b/module/documents/advancement/subclass.mjs
@@ -56,12 +56,14 @@ export default class SubclassAdvancement extends Advancement {
   /*  Application Methods                         */
   /* -------------------------------------------- */
 
-  /** @inheritdoc */
+  /** @inheritDoc */
   async apply(level, data, retainedData) {
     const useRetained = data.uuid === foundry.utils.getProperty(retainedData, "flags.dnd5e.sourceId");
     let itemData = useRetained ? retainedData : null;
     if ( !itemData ) {
       itemData = await this.createItemData(data.uuid);
+      delete itemData.flags?.dnd5e?.advancementOrigin;
+      delete itemData.flags?.dnd5e?.ultimateOrigin;
       foundry.utils.setProperty(itemData, "system.classIdentifier", this.item.identifier);
     }
     if ( itemData ) {


Zanderaf:
Any shot this makes it into 4.4? 🙏

module/documents/advancement/advancement.mjs
```diff
@@ -307,11 +307,13 @@ export default class Advancement extends PseudoDocumentMixin(BaseAdvancement) {
...
     if ( !source ) return null;
     const { _stats } = game.items.fromCompendium(source);
+    const advancementOrigin = `${this.item.id}.${this.id}`;
     return source.clone({
       _stats,
       _id: id ?? foundry.utils.randomID(),
       "flags.dnd5e.sourceId": uuid,
-      "flags.dnd5e.advancementOrigin": `${this.item.id}.${this.id}`
+      "flags.dnd5e.advancementOrigin": advancementOrigin,
+      "flags.dnd5e.ultimateOrigin": this.item.getFlag("dnd5e", "ultimateOrigin") ?? advancementOrigin
```
Fyorl: Prefer `advancementRoot`.

Fyorl (changes requested):
> Currently the levels for those advancement types refer to the character level, even if it is on a feature added by a class.

Let's make this change now. It should be safe to assume that for any feature granted by a class or subclass, its levels refer to class level. Are there any features that operate differently?

arbron:
> > Currently the levels for those advancement types refer to the character level, even if it is on a feature added by a class.
> 
> Let's make this change now. It should be safe to assume that for any feature granted by a class or subclass, its levels refer to class level. Are there any features that operate differently?

None that I can think of at the moment. We can set it to the granting class level for now and always add more customization later if a use case for selecting otherwise becomes apparent.

Zanderaf:
> > Currently the levels for those advancement types refer to the character level, even if it is on a feature added by a class.
> 
> Let's make this change now. It should be safe to assume that for any feature granted by a class or subclass, its levels refer to class level. Are there any features that operate differently?

I assume feats granted by Ability Score Improvements won't follow this trend? Some 1st & 3rd party feats I've seen / worked on sometimes mention character level in general when discussing their features. My hope is that if a feat is granted through an ASI it just takes the character level instead of the class level.

Fyorl:
> I assume feats granted by Ability Score Improvements won't follow this trend?

Good point that feats would have a class item `advancementRoot` too.

Zanderaf:
> > I assume feats granted by Ability Score Improvements won't follow this trend?
> 
> Good point that feats would have a class item `advancementRoot` too.

I guess just so I'm not confused too- would Feats be following the Class level of the class they're granted by? Or would they go off total character level? Feels like it makes more sense to have them go off of total character level instead of the class they're gained from.

krbz999:
We do have feats in some published premium content that don't explicitly reference class level. It's rather vague in the cases we've found whether it is actually intended to depend on character level instead.

However, since feats - beyond being an alternative to an ASI during level-up - can be granted by the game master at any time to a player, they can't actually always depend on class level.

arbron:
Feats now opt-out of being linked to class levels (and so can any other type if it so chooses).

module/applications/advancement/advancement-manager.mjs
```diff
@@ -395,28 +394,36 @@ export default class AdvancementManager extends Application5e {
   createLevelChangeSteps(classItem, levelDelta) {
     const raceItem = this.clone.system?.details?.race instanceof Item ? this.clone.system.details.race : null;
     const pushSteps = (flows, data) => this.steps.push(...flows.map(flow => ({ flow, ...data })));
-    const getItemFlows = characterLevel => this.clone.items.contents.flatMap(i => {
+    const getItemFlows = (characterLevel, classLevel) => this.clone.items.contents.flatMap(i => {
       if ( ["class", "subclass", "race"].includes(i.type) ) return [];
+      if ( ["class", "subclass"].includes(i.system.advancementRootItem?.type) && i.system.advancementClassLinked ) {
+        if ( i.system.advancementRootItem !== classItem ) return [];
+        else return this.constructor.flowsForLevel(i, classLevel);
```
Fyorl:
```suggestion
        return this.constructor.flowsForLevel(i, classLevel);
```

module/applications/advancement/advancement-manager.mjs
```diff
@@ -465,7 +473,7 @@ export default class AdvancementManager extends Application5e {
    * @returns {number}       Working level.
    */
   static currentLevel(item, actor) {
-    return item.system.levels ?? item.class?.system.levels ?? actor.system.details.level ?? 0;
+    return item.system.advancementLevel ?? actor.system.details.level ?? 0;
```
Fyorl, line 476: `advancementLevel` is guaranteed to return a number, so using the nullish coalescing operator will have no effect. If the `item`'s `actor` here differs from the `actor` passed-in, we will also get the wrong result. I assume the intention is that we always read from the passed-in `actor` and ignore the `item's `actor`?

Fyorl (changes requested):
Code looks good, I only spotted one issue. I also need to do some testing in Ember to make sure we don't break character creation, so I'll leave the PR open for now.

module/applications/advancement/advancement-manager.mjs
arbron: This was more protection against cases where a module-provided item that doesn't inherit from `ItemDataModel` is used, in which case `advancementLevel` won't exist.

Fyorl: I wonder why we pass-in the `actor` param then if it can always be read from the `item`?

module/applications/advancement/advancement-manager.mjs
```diff
@@ -357,7 +356,7 @@ export default class AdvancementManager extends Application5e {
     }
 
     // All other items, just create some flows down from current character level
-    Array.fromRange((manager.clone.system.details.level ?? 0) + 1)
+    Array.fromRange(manager.currentLevel(clonedItem, manager.clone) + 1)
```
Fyorl:
```suggestion
    Array.fromRange(manager.constructor.currentLevel(clonedItem, manager.clone) + 1)
```

module/applications/advancement/advancement-manager.mjs
```diff
@@ -265,9 +266,7 @@ export default class AdvancementManager extends Application5e {
     }
 
     // All other items, just create some flows up to current character level (or class level for subclasses)
-    let targetLevel = manager.clone.system.details.level ?? 0;
-    if ( clonedItem.type === "subclass" ) targetLevel = clonedItem.class?.system.levels ?? 0;
-    Array.fromRange(targetLevel + 1)
+    Array.fromRange(manager.currentLevel(clonedItem, manager.clone) + 1)
```
Fyorl:
```suggestion
    Array.fromRange(manager.constructor.currentLevel(clonedItem, manager.clone) + 1)
```

Fyorl (changes requested):

PR #5296, opened 2025-02-28
[#5180] Improve presentation of conditions applied by effects

### LINKED ISSUE #5180: Effect Status Conditions are not applied when enabled on the character sheet.
When an active effect with rider status conditions is applied by creating and/or enabling on the character sheet effects tab no rider status conditions are applied.

To reproduce: 
- Create an active effect with rider conditions directly on a character sheet active effect tab
  or
- Apply an active effect from a chat message then disable and re-enable it from the active effect tab on the character sheet

arbron:
In order to improve communication over statuses imposed using the "Status Conditions" option on effects, this applies two UI improvements:

First, the condition will be highlighted as applied on the actor sheet (though the toggle won't be in the enabled position). This should show that that effect is taking place on the actor and will take condition immunities into account.

Highlight Effects

Secondly, when hovering over one of the Conditions on the effects tab, it will highlight the active effects that have imposed that condition, making it easier to track down what is causing your character to be paralyzed.

Closes #5180

### DIFF
--- a/less/v2/inventory.less
+++ b/less/v2/inventory.less
@@ -99,7 +99,13 @@
   .items-section {
 
     .items-header, .item {
+      outline: 2px dotted transparent;
+      outline-offset: 4px;
+      border-radius: 2px;
+      transition: outline-color 150ms ease;
+
       &[hidden] { display: none; }
+      &.highlighted { outline-color: red; }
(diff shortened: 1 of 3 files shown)


JPMeehan:
How does this affect the click behavior? And does this alter the responsibility of including relevant `changes`, e.g. Paralyzed = 0 speed.

PR #5301, opened 2025-03-01
fix missing recharge formula field for items

### LINKED ISSUE #5285: [BUG 4.3.5] - The recharge formula field is missing for items
Good Day,

the recharge formula field is missing for item objects. 
In addition, the previously set values ​​(e.g. 5-6) are set to 6 everywhere for the item objects

Image

but it is still there for activity objects:

Image

Tested with all modules deactivated.

Kapuzenjoe:
Fixes #5285

I have adopted the formula calculation from activities and now the recharge works again for items.


arbron:
Looks like this was based on the `4.4` branch, could you rebase this on the `4.3` branch and get rid of the extra commits?

PR #5315, opened 2025-03-03
[#4059] Add `TransformActivity`

### LINKED ISSUE #4059: Activity Type: Transform
An activity type for features like Wild Shape, Polymorph, etc. that exposes the features of the actor transformation system.

arbron:
Adds a new activity type that allows transforming one or more actors into a another actor. The activity has two configuration tabs similar to summons, one for setting up profiles that control what actors can be selected to transform into, and a second that allows selecting a transformation preset and customizing it if desired. Currently only compendium browser-based actor selection is supported, but more modes can be added in the future such as direct linkage by UUID.

Transform Activitiy Profiles

Transform Activity Settings

When the transform activity is used the usage dialog will appear if more than one valid profile is present. After the usage is complete the compendium browser will open allowing for the player to select which creature should be used as the transformation source. Both the profile and source actor UUID are saved in the chat message.

Transform Activity Message

The chat message then has a "Transform" button that will transform selected tokens using the source actor and settings.

Closes #4059


module/documents/activity/transform.mjs
```diff
@@ -0,0 +1,188 @@
...
+   * @property {string} profile  ID of the transformation profile to use.
+   * @property {string} [uuid]   UUID of the creature to transform into.
+   */
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _prepareUsageConfig(config) {
+    config = super._prepareUsageConfig(config);
+    config.transform ??= {};
```
Fyorl, line 77: Missing `create.transform`.

Fyorl (changes requested):

module/applications/activity/transform-sheet.mjs
```diff
@@ -0,0 +1,167 @@
...
+    }
+    return submitData;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _processSubmitData(event, submitData) {
+    // If customize is set but no settings set, save defaults
+    if ( submitData.transform?.customize && !this.activity.settings ) {
```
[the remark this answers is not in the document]
arbron: Yeah, that should be checking `!this.activity._source.settings`, just didn't update it after changing `prepareFinalData`.

The customize option will allow making changes on top of the existing preset. So when first checked (and not bugged) it will display all of the same options as are set in the preset but they can be changed. The idea is to not store any settings data if you aren't customizing the preset to avoid storing extra data and to ensure the preset can be updated automatically if the `CONFIG` is changed.

module/documents/activity/transform.mjs
arbron: `create.transform` isn't part of the configuration because transformation buttons are always added to the message. This is similar to how the Enchant activity is set up (the usage dialog still uses `create.transform` when checking `_sourceDisplay`, which matches Enchant which also has a profile but not checkbox).

module/documents/activity/transform.mjs
```diff
@@ -0,0 +1,188 @@
...
+      }
+    }, { inplace: false })
+  );
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  static localize() {
+    super.localize();
+    this._localizeSchema(this.schema.fields.profiles.element, ["DND5E.TRANSFORM.FIELDS.profiles"]);
```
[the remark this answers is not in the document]
arbron: Looks like we'll need to do a bit of duplication in localization to get this to work with both V12 and V13, because in V12 we need the keys in `DND5E.TRANSFORM.FIELDS.profiles.FIELDS` and in V13 they are in `DND5E.TRANSFORM.FIELDS.profiles`.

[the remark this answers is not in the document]
Fyorl: That is unfortunate

module/documents/activity/transform.mjs
Fyorl: How does this part work then?

> When the transform activity is used the usage dialog will appear if more than one valid profile is present.

I guess it instead needs a `_requiresConfigurationDialog` override, like `EnchantActivity`?

module/documents/activity/summon.mjs
```diff
@@ -40,7 +40,9 @@ export default class SummonActivity extends ActivityMixin(SummonActivityData) {
   /** @inheritDoc */
   static localize() {
     super.localize();
-    this._localizeSchema(this.schema.fields.profiles.element, ["DND5E.SUMMON.FIELDS.profiles"]);
+    if ( game.release.generation < 13 ) {
```
Fyorl, line 43: Need to backport these to 4.3.7

module/applications/activity/transform-usage-dialog.mjs
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Determine the label for a profile in the ability use dialog.
+   * @param {SummonsProfile} profile  Profile for which to generate the label.
+   * @param {object} rollData         Roll data used to prepare the count.
+   * @returns {string}
+   */
+  getProfileLabel(profile, rollData) {
+    if ( profile.name ) return profile.name;
+    const cr = simplifyBonus(profile.cr, rollData);
```
Fyorl, line 56: Missing import for this

module/documents/activity/transform.mjs
arbron: That would indeed be the case 👍

module/applications/activity/transform-sheet.mjs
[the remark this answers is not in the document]
Fyorl: There's some weirdness around switching between presets while 'customize' is checked, but it works pretty intuitively if you stick to the one preset and attempt to customise it I think.

I would imagine it's probably simpler if we just have a 'custom' option in the drop-down and don't attempt to handle this combination of preset + customize. But the code's already written, so I don't really mind. I'll leave it up to you.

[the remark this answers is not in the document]
arbron: The reason I wanted to keep it that way is because there might be the use case where you want to keep the `wildshape` preset but change it. That would allow any modules looking for the `wildshape` preset in the hook a chance to implement their own behavior while still giving some customization control.

module/documents/activity/summon.mjs
arbron: https://github.com/foundryvtt/dnd5e/pull/5420

module/applications/activity/transform-sheet.mjs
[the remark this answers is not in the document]
Fyorl: Sounds pretty reasonable to me

Fyorl (approved):
Working great after the most recent changed. Feel free to merge away. I will take care of those backports if you don't.

PR #5336, opened 2025-03-05
Fix onTokenHUDRender and onRenderActorDirectory for V13

### LINKED ISSUE #5335: Token HUD throws errors in V13 when token has exhaustion
Code's just still assuming jQuery being passed in, so errors out when calling `find` on an `HTMLElement`

roth-michael:
Closes #5335
Also covers the primary party not being marked as such in the actor directory.

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -663,8 +663,8 @@ export default class ActiveEffect5e extends ActiveEffect {
 
   /**
    * Adjust exhaustion icon display to match current level.
-   * @param {Application} app  The TokenHUD application.
-   * @param {jQuery | HTMLElement} html      The TokenHUD HTML.
+   * @param {Application} app            The TokenHUD application.
+   * @param {jQuery | HTMLElement} html  The TokenHUD HTML.
    */
   static onTokenHUDRender(app, html) {
     html = html instanceof HTMLElement ? html : html[0];
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -664,17 +664,19 @@ export default class ActiveEffect5e extends ActiveEffect {
   /**
    * Adjust exhaustion icon display to match current level.
    * @param {Application} app  The TokenHUD application.
-   * @param {jQuery} html      The TokenHUD HTML.
+   * @param {jQuery | HTMLElement} html      The TokenHUD HTML.
    */
   static onTokenHUDRender(app, html) {
+    html = html instanceof HTMLElement ? html : html[0];
     const actor = app.object.actor;
     const level = foundry.utils.getProperty(actor, "system.attributes.exhaustion");
     if ( Number.isFinite(level) && (level > 0) ) {
       const img = ActiveEffect5e._getExhaustionImage(level);
-      html.find('[data-status-id="exhaustion"]').css({
-        objectPosition: "-100px",
-        background: `url('${img}') no-repeat center / contain`
-      });
+      const elem = html.querySelector('[data-status-id="exhaustion"]');
+      if ( elem ) {
+        elem.style.objectPosition = "-100px";
+        elem.style.background = `url('${img}') no-repeat center / contain`;
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


roth-michael:
Actually, similar issue: Adding context menu options isn't working in V13. I could tag that onto this PR.

roth-michael:
Added said fixes to the PR. Touches a little bit more so if you'd prefer it move into its own, will do. Summary:
`getActorDirectoryEntryContext`, `getItemDirectoryEntryContext`, `getCompendiumEntryContext` are no longer called in V13. As such, added a new hook for `getEntryContextDocumentDirectory`, exiting early if it's a Compendium, otherwise calling `addDirectoryContextOptions` in either `Actor5e` or `Item5e`, according to the `documentName`. Item compendium context options are handled in `ItemCompendium5eV13` directly, avoiding having to put logic for calling the old `addCompendiumContextOptions` (which can be removed once v12 support is dropped). Technically the same could be done for `ItemCompendium5eV12`, but why bother changing something which currently works and will be deleted in 5.0.

The changes to the `addXContextOptions` functions themselves are just dual-compatible "use the `HTMLElement` instead of jQuery if necessary" changes.

module/documents/active-effect.mjs
```diff
@@ -664,17 +664,19 @@ export default class ActiveEffect5e extends ActiveEffect {
   /**
    * Adjust exhaustion icon display to match current level.
    * @param {Application} app  The TokenHUD application.
-   * @param {jQuery} html      The TokenHUD HTML.
+   * @param {jQuery | HTMLElement} html      The TokenHUD HTML.
```
arbron, line 667:
```suggestion
   * @param {jQuery | HTMLElement} html  The TokenHUD HTML.
```

module/documents/active-effect.mjs
```diff
@@ -664,17 +664,19 @@ export default class ActiveEffect5e extends ActiveEffect {
   /**
    * Adjust exhaustion icon display to match current level.
    * @param {Application} app  The TokenHUD application.
```
arbron, line 666:
```suggestion
   * @param {Application} app            The TokenHUD application.
```

module/documents/actor/actor.mjs
```diff
@@ -3221,32 +3221,36 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /**
    * Add additional system-specific sidebar directory context menu options for Actor documents
-   * @param {jQuery} html         The sidebar HTML
+   * @param {jQuery | HTMLElement} html         The sidebar HTML
    * @param {Array} entryOptions  The default array of context menu options
```
arbron, line 3225:
```suggestion
   * @param {jQuery | HTMLElement} html  The sidebar HTML
   * @param {Array} entryOptions         The default array of context menu options
```

module/documents/actor/actor.mjs
```diff
@@ -3271,12 +3276,13 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /**
    * Add class to actor entry representing the primary group.
-   * @param {jQuery} jQuery
+   * @param {jQuery | HTMLElement } html
```
arbron, line 3279:
```suggestion
   * @param {jQuery | HTMLElement} html
```

module/documents/item.mjs
```diff
@@ -1446,20 +1447,22 @@ export default class Item5e extends SystemDocumentMixin(Item) {
 
   /**
    * Add additional system-specific sidebar directory context menu options for Item documents.
-   * @param {jQuery} html            The sidebar HTML.
+   * @param {jQuery | HTMLElement} html            The sidebar HTML.
    * @param {object[]} entryOptions  The default array of context menu options.
```
arbron, line 1451:
```suggestion
   * @param {jQuery | HTMLElement} html  The sidebar HTML.
   * @param {object[]} entryOptions      The default array of context menu options.
```

arbron (changes requested):
Just a few styling issues in the comments, but otherwise it looks solid.

arbron (approved):

PR #5345, opened 2025-03-06
Incorporate "movement actions" for V13

### LINKED ISSUE #4379: Request: Add dropdown to Configure Movement Speed to set displayed speed
The displayed movement speed is currently configured to display the highest speed. Request that a dropdown be added to Configure Movement Speed that will allow for selection of the desired displayed speed, something like shown.
displayed speed

In our case, a player has a ring of swimming, so the swim speed is always shown.

roth-michael:
While this is technically functional & mergeable in its current state, more work needs to be done on it.
As a byproduct of the changes necessary, this closes #4379.
Current changes:
- Add all `CONFIG.DND5E.movementTypes` to `CONFIG.Token.movement.actions` (in V13) to allow configuration of cost multipliers for each in the new region behavior type.
- Add `current` to `MovementField`, representing the "current" movement type/action, and render as a dropdown in the movement config
- The above field is respected on the Actor sheet, if set, showing the selected movement type & speed rather than just the highest (currently, "default" behavior is the same as it was before)
- Add the relevant overrides to `Token5e` so that any movement is considered to be of the "current" type (though it currently defaults to "walk" rather than whatever the highest speed is)

Potential further changes:
- Change `CONFIG.DND5E.movementTypes` to a nested object with `label` and `icon`, to match the `CONFIG.Token.movement.actions` format
- Align default behaviors so that what's shown on the sheet always represents the "current" movement type (not sure what the best strategy is - I think maybe just always defaulting to "walk", if indeed the displayed movement speed should be coupled to the "current" movement action for v13's sake)
- Add a better way of switching "current" movement action ad-hoc, probably via the Token HUD somehow
- Consider modifying drag ruler behavior to show maximum movable distance with "current" movement action during combat, perhaps both with text and via some sort of (maybe-optional) color highlighting to represent "can do this", "could do this if dashing", and "can't do this" (this is maybe better left as a 5.0 issue)

Potential snags:
- #4976 changes `MovementField`, so would have to account for that once merged
- The "Default (whatever)" option in the movement config is based on the speeds pre-condition-changes & pre-active-effects. So, for instance, a walk speed of 30 and swim speed of 40 would result in "Default (swim)", but if the actor was then made Prone, swim speed becomes 0, and therefore walk speed would be shown on the sheet. Similarly, an AE that gives a fly speed of 80 wouldn't be registered in the movement config, and therefore the "Default (swim)" would still show up, but in actuality Fly would be displayed on the sheet. Not sure the ideal hangling for this, as I mentioned in the second "Potential further changes" bullet

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -76,6 +76,10 @@ Hooks.once("init", function() {
   CONFIG.JournalEntryPage.documentClass = documents.JournalEntryPage5e;
   CONFIG.Token.documentClass = documents.TokenDocument5e;
   CONFIG.Token.objectClass = canvas.Token5e;
+  if ( game.release.generation > 12 ) {
+    CONFIG.Token.movement.actions =
+      Object.fromEntries(Object.entries(CONFIG.DND5E.movementTypes).map(([k, v]) => ([k, {label: v}])));
+  }
   CONFIG.User.documentClass = documents.User5e;
   CONFIG.time.roundTime = 6;
   Roll.TOOLTIP_TEMPLATE = "systems/dnd5e/templates/chat/roll-breakdown.hbs";
--- a/lang/en.json
+++ b/lang/en.json
@@ -874,6 +874,11 @@
     "Gold": "{value} ({claimed})",
     "NoOrders": "No orders issued to special facilities this turn.",
     "Order": "{link} completed the <strong>{order}</strong> order."
+  },
+  "Trade": {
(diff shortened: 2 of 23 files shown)


roth-michael:
With 13.340, this'll probably make more sense to be almost entirely rewritten. New question, assuming choosing a "default to show" movement is still a part of the PR, would be how (if at all) that should interact with the currently-selected movement type of a token.

Issue #5340, opened 2025-03-06
Convert item sheets to ApplicationV2


arbron:
Implement a ApplicationV2 version of the item sheets and consolidate the various inherited sheets and mixins. 

- [x] Create ApplicationV2 version of item sheet
- [x] Remove `ItemSheet5e`, `ItemSheet5e2`, and `ItemSheetV2Mixin`
- [x] Create new version of container sheet based on new item sheet
- [x] Remove `ContainerSheet` and `ContainerSheet2`
- [x] Remove legacy sheet templates
- [x] Convert `StartingEquipmentConfig` to AppV2

PR #5353, opened 2025-03-07
[#5340] Add `PrimarySheetMixin`, move section expansion

arbron:
Introduces `PrimarySheetMixin` which is a V2-compatible version of `DocumentSheetV2Mixin` that handles tabs, edit/play toggle, and some other behavior that will be shared between the `ApplicationV2` versions of the actor and item sheets.

Also moves the section expansion code from `ActivitySheet` into `ApplicationV2Mixin` so it can be used to handle the description sections on the item sheets.

Adds the `triggerEvent` static method to `ContextMenu5e` to consolidate some duplicate code for the "Additional Options" dots buttons.

### DIFF
--- a/module/applications/api/application-v2-mixin.mjs
+++ b/module/applications/api/application-v2-mixin.mjs
@@ -18,6 +18,9 @@ export default function ApplicationV2Mixin(Base) {
   class BaseApplication5e extends HandlebarsApplicationMixin(Base) {
     /** @override */
     static DEFAULT_OPTIONS = {
+      actions: {
+        toggleCollapsed: BaseApplication5e.#toggleCollapsed
+      },
       classes: ["dnd5e2"],
       window: {
         subtitle: ""
@@ -215,21 +218,13 @@ export default function ApplicationV2Mixin(Base) {
     /*  Event Listeners and Handlers                */
     /* -------------------------------------------- */
 
-    /** @inheritDoc */
-    _onClickAction(event, target) {
-      if ( target.dataset.action === "toggleCollapsed" ) this.#toggleCollapsed(event, target);
-      else super._onClickAction(event, target);
-    }
-
-    /* -------------------------------------------- */
-
     /**
      * Handle toggling the collapsed state of collapsible sections.
      * @this {BaseApplication5e}
      * @param {Event} event         Triggering click event.
      * @param {HTMLElement} target  Button that was clicked.
      */
-    #toggleCollapsed(event, target) {
+    static #toggleCollapsed(event, target) {
       if ( event.target.closest(".collapsible-content") ) return;
       target.classList.toggle("collapsed");
       this.#expandedSections.set(
--- a/module/applications/api/application-v2-mixin.mjs
+++ b/module/applications/api/application-v2-mixin.mjs
@@ -36,6 +36,18 @@ export default function ApplicationV2Mixin(Base) {
     /* -------------------------------------------- */
 
     /**
+     * Expanded states for collapsible sections to persist between renders.
+     * @type {Map<string, boolean>}
+     */
+    #expandedSections = new Map();
+
+    get expandedSections() {
+      return this.#expandedSections;
+    }
+
+    /* -------------------------------------------- */
+
+    /**
      * A reference to the window subtitle.
      * @type {string}
      */
@@ -133,6 +145,19 @@ export default function ApplicationV2Mixin(Base) {
     /* -------------------------------------------- */
 
     /** @inheritDoc */
+    _replaceHTML(result, content, options) {
+      for ( const part of Object.values(result) ) {
+        for ( const element of part.querySelectorAll("[data-expand-id]") ) {
+          element.querySelector(".collapsible")?.classList
+            .toggle("collapsed", !this.#expandedSections.get(element.dataset.expandId));
+        }
+      }
+      super._replaceHTML(result, content, options);
+    }
+
+    /* -------------------------------------------- */
+
+    /** @inheritDoc */
     _updateFrame(options) {
       super._updateFrame(options);
       if ( options.window && ("subtitle" in options.window) ) {
@@ -145,6 +170,10 @@ export default function ApplicationV2Mixin(Base) {
     /** @inheritDoc */
     _onRender(context, options) {
       super._onRender(context, options);
+
+      this.element.querySelectorAll("[data-context-menu]").forEach(control =>
+        control.addEventListener("click", dnd5e.applications.ContextMenu5e.triggerEvent)
+      );
 
       // Allow multi-select tags to be removed when the whole tag is clicked.
       this.element.querySelectorAll("multi-select").forEach(select => {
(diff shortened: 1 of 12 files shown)


module/applications/api/primary-sheet-mixin.mjs
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    /* -------------------------------------------- */
+
+    /**
+     * Handle toggling an Item's description within an inventory list.
+     * @param {PointerEvent} event  The triggering event.
+     * @protected
+     */
+    async _onToggleDescription(event) {
```
Fyorl, line 335: I thought the PR description suggested this was going to be merged with the collapse/expand code from the activity sheet, but it seems that's not happening here. I was thinking of maybe moving this responsibility to the `InventoryElement`. I've moved in that direction on my prototype branch so if that sounds reasonable, we can pull this code out here and wait for the `InventoryElement` changes.

Fyorl (changes requested):
Looks good, I think I could already rebase my prototype App V2 group sheet onto this. Some notes before we proceed though:

- I'm building the App V2 architecture for inventory rendering on my prototype branch. That includes render context preparation, and I also have a cleaned-up V13-only version of the `InventoryElement` and the `ItemListControlsElement` that I am working on. Just a heads up so we don't duplicate any effort on those fronts.
- Would it make sense to wait for 5.0 for this work? There's not a lot of v12 compatibility code required I can see, but particularly the `partDescriptors` backport doesn't feel great. Do we need this for anything in 4.4 or shall we move it to 5.0 and excise any v12 compatibility code?

module/applications/api/application-v2-mixin.mjs
```diff
@@ -132,6 +144,19 @@ export default function ApplicationV2Mixin(Base) {
 
     /* -------------------------------------------- */
 
+    /** @inheritDoc */
+    _replaceHTML(result, content, options) {
+      for ( const part of Object.values(result) ) {
```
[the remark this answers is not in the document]
arbron: Yeah, originally they were put into `_onRender` when we moved activities to AppV2 but that was causing it to animate every time the sheet was updated so I moved them here.

module/applications/api/primary-sheet-mixin.mjs
arbron: That mention was for the main descriptions on the item sheets, not the item summaries in the inventory area. Because this needs to perform the extra fetching and rendering it couldn't be merged with the other code. Inventory element would probably be the best place for this, yeah.

arbron:
> * I'm building the App V2 architecture for inventory rendering on my prototype branch. That includes render context preparation, and I also have a cleaned-up V13-only version of the `InventoryElement` and the `ItemListControlsElement` that I am working on. Just a heads up so we don't duplicate any effort on those fronts.

Haven't need to touch those other than the bug fixes that are already merged.

> * Would it make sense to wait for 5.0 for this work? There's not a lot of v12 compatibility code required I can see, but particularly the `partDescriptors` backport doesn't feel great. Do we need this for anything in 4.4 or shall we move it to 5.0 and excise any v12 compatibility code?

Just if we wanted to swap over item sheets to AppV2 before 5.0. Other than the `partDescriptors` work, the rest of the item sheet also needed some code to initialize the ProseMirror editors because that isn't available in AppV2 and we can't use `<prose-mirror>` until V13. If we want to hold off on converting the item sheet until 5.0, then I can pull out the cross-compatibility stuff and just focus on V13.

module/applications/api/primary-sheet-mixin.mjs
Fyorl: > That mention was for the main descriptions on the item sheets, not the item summaries in the inventory area.

Ah ok, that makes sense.

> When this ends up in `InventoryElement`, it would be good to take advantage of `expandedSections` since it will be available in all of our AppV2 sheets rather than tracking those separately.

Will keep this in mind. Certainly remind me again if it doesn't end up in the PR.

Fyorl:
> Just if we wanted to swap over item sheets to AppV2 before 5.0. Other than the `partDescriptors` work, the rest of the item sheet also needed some code to initialize the ProseMirror editors because that isn't available in AppV2 and we can't use `<prose-mirror>` until V13. If we want to hold off on converting the item sheet until 5.0, then I can pull out the cross-compatibility stuff and just focus on V13.

Don't think there's a pressing need to convert those to App V2 before 5.0. It will save us effort to only convert once we've dropped v12 support, so let's re-target this PR on 5.0 unless there are other considerations I've missed.

module/applications/api/primary-sheet-mixin.mjs
```diff
@@ -0,0 +1,361 @@
...
+      const html = await super._renderFrame(options);
+      const header = html.querySelector(".window-header");
+
+      if ( !game.user.isGM && this.document.limited ) {
+        html.classList.add("limited");
+        return html;
+      }
+
+      // Add edit <-> play slide toggle.
+      if ( this.isEditable ) {
```
Fyorl, line 110: Consider moving this out to `_renderModeToggle` like we do with `_renderSource`. That would allow us to add/remove the toggle if the ownership of the Document changes.

module/applications/api/primary-sheet-mixin.mjs
```diff
@@ -0,0 +1,361 @@
...
+/**
+ * @import { FilterState5e } from "../components/item-list-controls.mjs";
+ */
+
+/**
+ * Adds V2 sheet functionality shared between primary document sheets (Actors & Items).
+ * @param {typeof DocumentSheet5e} Base  The base class being mixed.
+ * @returns {typeof PrimarySheet5e}
+ */
+export default function PrimarySheetMixin(Base) {
```
Fyorl, line 12: I wonder if this actually needs to be a mixin. Could it just be the base class for all our document sheets?

```js
export default class PrimaryDocumentSheet extends DragDropApplicationMixin(DocumentSheet5e) {}
```

Fyorl (commented):
Some quick comments before I can do a full re-review.

module/applications/api/primary-sheet-mixin.mjs
arbron: For actors we'll want to mix this into core's `ActorSheetV2` so we get all of the token header actions, rather than the base `DocumentSheetV2`.

module/applications/api/primary-sheet-mixin.mjs
```diff
@@ -0,0 +1,358 @@
...
+        && (d[`${base}Type`] === t[`${base}Type`]) ? "move" : "copy";
+    }
+
+    /* -------------------------------------------- */
+
+    /** @inheritDoc */
+    async _onDragStart(event) {
+      await super._onDragStart(event);
+      if ( !this.document.isOwner
+        || this.document[game.release.generation < 13 ? "compendium" : "collection"]?.locked ) {
```
Fyorl, line 353:
```suggestion
      if ( !this.document.isOwner || this.document.collection?.locked ) {
```

Fyorl (approved):

PR #5354, opened 2025-03-07
[#5275, #5319] Fix flat damage label, display scaled cantrips

### LINKED ISSUE #5275: [Enhancement] Reflect cantrip scaling in Spell list formula column
Cantrips using the "Cantrip Scaling" feature continue to show their base number of dice on the Character Sheet's spell list.

I.e. as a Level 5 warlock, Toll the Dead shows a 1d8 and 1d12 for the two damage activities
Image

While the activity is configured to use Cantrip scaling, and the [[/damage]] enrichers (pointing at the activities) correctly reflect the scaled damage
Image

### LINKED ISSUE #5319: Sheet damage formula shows mod for flat damage rolls
While the damage enricher shows the correct value, and the roll damage is correct in not including the modifier, the sheet still adds the modifier into the damage shown.

Image

arbron:
Switches to using `getDamageConfig` rather than using the raw damage parts when generating the damage labels. This allows the de-duplication in handling for `@mod` and other bonuses.

This also means the scaled cantrip damage is displayed, rather than just showing the base damage.

Closes #5275
Closes #5319

### DIFF
--- a/module/data/activity/base-activity.mjs
+++ b/module/data/activity/base-activity.mjs
@@ -613,20 +613,15 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
 
   /**
    * Prepare the label for a compiled and simplified damage formula.
-   * @param {DamageData[]} parts  Damage parts to create labels for.
-   * @param {object} rollData     Deterministic roll data from the item.
+   * @param {object} rollData  Deterministic roll data from the item.
    */
-  prepareDamageLabel(parts, rollData) {
-    this.labels.damage = parts.map((part, index) => {
+  prepareDamageLabel(rollData) {
+    const config = this.getDamageConfig();
+    this.labels.damage = (config.rolls ?? []).map((part, index) => {
       let formula;
       try {
-        formula = part.formula;
-        if ( part.base ) {
-          if ( this.item.system.magicAvailable ) formula += ` + ${this.item.system.magicalBonus ?? 0}`;
-          if ( (this.item.type === "weapon") && !/@mod\b/.test(formula) ) formula += " + @mod";
-        }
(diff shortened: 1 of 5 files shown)


module/data/activity/base-activity.mjs
```diff
@@ -613,20 +613,15 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
 
   /**
    * Prepare the label for a compiled and simplified damage formula.
-   * @param {DamageData[]} parts  Damage parts to create labels for.
-   * @param {object} rollData     Deterministic roll data from the item.
+   * @param {object} rollData  Deterministic roll data from the item.
    */
-  prepareDamageLabel(parts, rollData) {
-    this.labels.damage = parts.map((part, index) => {
+  prepareDamageLabel(rollData) {
```
Fyorl, line 618: Reasonably sure there are 3rd-party Activity implementations out there, so we probably need to provide a deprecation path for this signature change. Thankfully, I don't think it's possible that `rollData` would ever be an array, so checking if the first argument is an array should be a fairly straightforward check for a legacy invocation.

module/data/activity/base-activity.mjs
```diff
@@ -636,15 +631,18 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
       }
 
       let label = formula;
-      if ( part.types.size ) {
+      const types = part.options?.types ?? part.options?.type ? [part.options.type] : [];
```
Fyorl, line 634:
```suggestion
      const types = part.options?.types ?? (part.options?.type ? [part.options.type] : []);
```

Fyorl (changes requested):

Fyorl (approved):

PR #5358, opened 2025-03-08
Add DamageTraitField and SimpleTraitField exports to module

kaldigo:
With makeSimpleTrait and makeDamageTrait being deprecated I couldn't find the SimpleTraitField and DamageTraitField classes available anywhere so I added them to the export for actor data.

### DIFF
--- a/module/data/actor/_module.mjs
+++ b/module/data/actor/_module.mjs
@@ -10,6 +10,8 @@ export {
   VehicleData
 };
 export {default as GroupSystemFlags} from "./group-system-flags.mjs";
+export {default as DamageTraitField} from "./fields/damage-trait-field.mjs";
+export {default as SimpleTraitField} from "./fields/simple-trait-field.mjs";
 export {default as AttributesFields} from "./templates/attributes.mjs";
 export {default as CommonTemplate} from "./templates/common.mjs";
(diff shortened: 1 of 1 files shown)


module/data/actor/_module.mjs
```diff
@@ -15,6 +15,8 @@ export {default as CommonTemplate} from "./templates/common.mjs";
 export {default as CreatureTemplate} from "./templates/creature.mjs";
 export {default as DetailsFields} from "./templates/details.mjs";
 export {default as TraitsFields} from "./templates/traits.mjs";
+export {default as DamageTraitField} from "./fields/damage-trait-field.mjs";
+export {default as SimpleTraitField} from "./fields/simple-trait-field.mjs";
```
arbron, line 19: These should be shifted up above the `AttributesFields` export to maintain alphabetical order of source files.

arbron (changes requested):

module/data/actor/_module.mjs
kaldigo: This has now been amended.

arbron (approved):

arbron:
Looks like this will need to be rebased on the current `4.4` branch and force pushed up.

arbron:
Sorry about this, would you be able to rebase onto `4.4.x` once again :)

PR #5367, opened 2025-03-11
[#5129] Add rest request chat card that can be invoked from party

### LINKED ISSUE #5129: Add option for DM to call for a Short/Long Rest
Currently there are two rest workflows:
1. Each player rests their characters manually
2. DM uses the rest button in the group actor to rest all characters automatically

The first method allows for each player to manage their own rest including rolling their own hit dice in the short rest dialog, but it doesn't have the ability to auto-advance game time or trigger bastion turns like the group rest does. It also requires each player to set properties like the "New Day" checkbox themselves. The group rest buttons on the other hand always rest all actors in the group, which might not always be desirable if one character isn't able to get the rest.

I propose modifying the dialogs shown through the group rest buttons to display an option to call a rest, rather than just resting all actors automatically. This would post a chat message for tracking the rest progress and would automatically open the appropriate rest dialog for each player. Once each player has finished resting their character the original chat message will be updated with the rest progress. This gives players more control over the resting process for their characters while still allowing the DM to set the "New Day" option, advance the game time, and trigger bastion turns.

arbron:
Adds a new `restRequest` chat message type that contains a list of actors to be rested, details on the rest being performed, and links to the individual actor rest result messages. Players will see rest buttons allowing them to rest directly from the chat message which will be updated to show that actor has completed their rest.

The group rest dialogs have been expanded to include a button to specify whether the rest should be performed automatically or not, and options to control which group members are rested. For non-party groups all actors are automatically rested.

The rest configuration now takes an optional `request` option pointing to the rest request chat message. This is also stored in rest result messages to easily associate a rest result with its request.

### Possible Future Work
- [ ] Use V13's query system to show the dialog to players
- [ ] Improve handling of NPCs in party
- [ ] Add config option to disable resting from actor sheets
- [ ] Associate rests from sheet with recent request message

Closes #5129

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -554,6 +554,7 @@ Hooks.on("renderChatLog", (app, html, data) => {
 Hooks.on("renderChatPopout", (app, html, data) => documents.Item5e.chatListeners(html));
 
 Hooks.on("chatMessage", (app, message, data) => applications.Award.chatMessage(message));
+Hooks.on("createChatMessage", dataModels.chatMessage.RestRequestMessageData.onCreateRestMessage);
 
 Hooks.on("renderActorDirectory", (app, html, data) => documents.Actor5e.onRenderActorDirectory(html));
 
--- a/lang/en.json
+++ b/lang/en.json
@@ -21,6 +21,7 @@
 
 "TYPES.ChatMessage": {
   "rest": "Rest Message",
+  "restRequest": "Rest Request Message",
   "turn": "Combat Turn Message"
 },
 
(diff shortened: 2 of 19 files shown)


Fyorl (changes requested):
I rebased this already for testing and there were some conflicts with the temp & tempMax options added in 5.1. They probably need the same treatment as the 'New Day' option, allowing the DM to lock-in those rest settings, but maybe you have a different idea. I can push up my rebase if that's helpful.

Fyorl (approved):

PR #5368, opened 2025-03-11
[#933, #5238] Add custom movement types

### LINKED ISSUE #933: [Suggestion] Provide an option for a "speical" movement field which can contain any arbitrary string.
Originally in GitLab by @matthew.e.boyce

In order to submit an effective bug report, please include the following information along with your issue description.

### Environment Details

* Foundry VTT Version: 0.7.9
* Operating System: Linux (Ubuntu)
* How Are You Using Foundry: Chrome
* Which Game System: dnd5e 1.2.0
* Modules Enabled?: yes (not relevant)

### Issue Description

The transition from representing speed as a string to movement as an struct of integers has caused a regression for some monster types. Monsters which have have special movement options--such as the Galeb Duhr[¹](https://forgottenrealms.fandom.com/wiki/Galeb_duhr)[²](https://www.dndbeyond.com/monsters/galeb-duhr)--aren't fully representable. While requesting the addition of an open ended string field seems like an FR, this was representable prior to the more elaborate representation, so this presents as a regression. `di`/`dr` fields are represented with similar structs but have a `special` field which allows for arbitrary extension, which may be a reasonable solution.

### LINKED ISSUE #5238: Add special field to movement that can show arbitrary data
Similar to the special senses data, this will allow displaying additional movement information that can't be represented by the data model.

arbron:
Adds custom movement which allows adding new movement types with distances in addition to the standard types. These custom types are displayed on the sheet like the others and in parentheses in the NPC stat block.

Movement Custom

### Possible Future Work
- [ ] Add custom types to senses to match movement
- [ ] Support formulas in custom movement types once added to normal types

Closes #933
Closes #5238

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3421,6 +3421,8 @@
 
 "DND5E.MOVEMENT": {
   "Action": {
+    "AddCustom": "Add Custom Movement",
+    "DeleteCustom": "Delete Custom Movement",
     "Configure": "Configure Movement Speed"
   },
   "FIELDS": {
--- a/less/v2/apps.less
+++ b/less/v2/apps.less
@@ -450,6 +450,7 @@
     border-radius: 4px;
     background: var(--dnd5e-color-black);
     color: var(--dnd5e-color-gold);
+    flex: 0 0 30px;
     width: 30px;
     height: 30px;
     margin: 0;
@@ -1519,6 +1520,10 @@ dialog.dnd5e2.application {
     padding-block: 3px 2px;
     padding-inline: 10px;
     border-radius: 0;
+  }
+  fieldset > .gold-button {
+    align-self: flex-end;
+    margin-block-start: 4px;
   }
 }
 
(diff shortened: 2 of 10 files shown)


Fyorl (changes requested):
Seems like too much for how niche this is. The normal semi-colon-delimited special field should be enough I would think.

arbron:
@Fyorl I've pulled the special movement field out into its own PR: https://github.com/foundryvtt/dnd5e/pull/5804

For these custom types, I was originally thinking they would be helpful for movement tracking using the new movement measurement system since they have actual values, but now that V13 is out it doesn't seem feasible to add actor specific movement actions, which would be necessary for properly handling that.

Fyorl:
> @Fyorl I've pulled the special movement field out into its own PR: #5804
> 
> For these custom types, I was originally thinking they would be helpful for movement tracking using the new movement measurement system since they have actual values, but now that V13 is out it doesn't seem feasible to add actor specific movement actions, which would be necessary for properly handling that.

I see, yeah, makes sense. The special field is an easy add so I'll move forward with that, thanks for splitting it out. Will have to check with @dev7355608 how feasible per-Actor movement actions are.

arbron:
Marked as blocked until we can resolve the question of custom movement actions for a specific actor

PR #5379, opened 2025-03-15
[#4576] Add description to activities

### LINKED ISSUE #4576: Add a Chat Description field to Activities
## Feature Request

I think we should add a Chat Description field to Activities. This would function similar to how the Chat Description works on the items, but it would be considered first before the two descriptions on the items. In other words, when generating the chat message, it would use the descriptions in this order:

1. Chat Description on the Activity
2. Chat Description on the Item
3. Description on the Item

I thought about proposing that we move the Chat Description from the Item to the Activity but I don't think that's necessary. It might clean things up but I'm still on the fence if it'd be better so I guess I'd leave it be, for now.

### Starry Form
I was getting some practice with the new Activities feature, so I went looking for some more complex items to create. I started with the Circle of Stars Druid subclass from Tasha's and specifically the Starry Form feature. It's a very involved feature since you choose from 1 of 3 forms to transform into and consume a Wild Shape use when doing so. Then 2 of the 3 forms give you special actions which would make for good activities.

I made 3 Activities:

1. Transform - A Utility Activity that handles the transformation, consumes a Wild Shape use and lets you apply 1 of 3 active effects just to remind you which form you chose.
2. Archer Attack - An Attack Activity for the bonus action attack you get when in the Archer form
3. Chalice Healing - A Healing Activity for the bonus healing you get when in the Chalice form

I would really like to have separate chat descriptions for these three activities. That way I get a short description for each of them, since they're so different.

### Crook of Rao
There was a magic item, Crook of Rao from Tasha's too, that I had the same issue with. In addition to casting spells, which I was going to use separate items for, it has two special abilities that I'd make separate activities for. There too, the item has a very long description and even trimming that down for a chat description would mean I'd have to include information about all the special abilities rather than just the one I'm actually using.

arbron:
Adds a new HTML description field to activities that is properly sanitized. When an activity is used this description is used in the chat if provided before the system fall back to the chat description and then the full description. Using the "Display in Chat" context option allows the normal item description to be shown rather than an activity description.

### Possible Future Work
- [ ] Add activity tooltips with description to actor inventory

Closes #4576

### DIFF
--- a/module/applications/activity/activity-sheet.mjs
+++ b/module/applications/activity/activity-sheet.mjs
@@ -2,6 +2,8 @@ import { ConsumptionTargetData } from "../../data/activity/fields/consumption-ta
 import UsesField from "../../data/shared/uses-field.mjs";
 import PseudoDocumentSheet from "../api/pseudo-document-sheet.mjs";
 
+const TextEditor = foundry.applications.ux.TextEditor.implementation;
+
 /**
  * Default sheet for activities.
  */
@@ -341,6 +343,9 @@ export default class ActivitySheet extends PseudoDocumentSheet {
       field: context.fields.target.fields.prompt,
       value: context.source.target.prompt,
       input: context.inputs.createCheckboxInput
+    });
+    context.enriched = await TextEditor.enrichHTML(this.activity.description.value, {
+      relativeTo: this.activity, rollData: this.activity.getRollData(), secrets: this.activity.item.isOwner
     });
     context.placeholder = {
       name: game.i18n.localize(this.activity.metadata.title),
--- a/module/applications/activity/activity-sheet.mjs
+++ b/module/applications/activity/activity-sheet.mjs
@@ -345,7 +345,7 @@ export default class ActivitySheet extends PseudoDocumentSheet {
       input: context.inputs.createCheckboxInput
     });
     context.enriched = await TextEditor.enrichHTML(this.activity.description.value, {
-      relativeTo: this.activity, rollData: this.activity.getRollData(), secrets: this.activity.item.isOwner
+      relativeTo: this.activity.item, rollData: this.activity.getRollData(), secrets: this.activity.item.isOwner
     });
     context.placeholder = {
       name: game.i18n.localize(this.activity.metadata.title),
--- a/module/data/abstract/item-data-model.mjs
+++ b/module/data/abstract/item-data-model.mjs
@@ -200,7 +200,7 @@ export default class ItemDataModel extends SystemDataModel {
     uses = this.hasLimitedUses && (game.user.isGM || identified) ? uses : null;
     price = game.user.isGM || identified ? price : null;
 
-    enrichmentOptions = { rollData, ...enrichmentOptions };
+    enrichmentOptions = { rollData, relative: this.parent, ...enrichmentOptions };
     const context = {
       name, type, img, price, weight, uses, school, materials,
       config: CONFIG.DND5E,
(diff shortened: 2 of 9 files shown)


Zanderaf:
Any thoughts about having those descriptions or at the very least tooltips shown for activities on items and maybe when you are prompted to choose which activity you would like to use on an item?
image

module/applications/activity/activity-sheet.mjs
```diff
@@ -342,8 +344,11 @@ export default class ActivitySheet extends PseudoDocumentSheet {
       value: context.source.target.prompt,
       input: context.inputs.createCheckboxInput
     });
+    context.enriched = await TextEditor.enrichHTML(this.activity.description.value, {
+      relativeTo: this.activity, rollData: this.activity.getRollData(), secrets: this.activity.item.isOwner
```
Fyorl: Activities are not Documents. `relativeTo` requires a `ClientDocument`.

```suggestion
      relativeTo: this.activity.item, rollData: this.activity.getRollData(), secrets: this.activity.item.isOwner
```

module/data/abstract/item-data-model.mjs
```diff
@@ -210,12 +209,15 @@ export default class ItemDataModel extends SystemDataModel {
...
-          rollData, relativeTo: this.parent, ...enrichmentOptions
-        }),
+        value: await TextEditor.enrichHTML(
+          (game.user.isGM || isIdentified ? description.value : unidentified?.description) ?? "",
+          { relativeTo: this.parent, ...enrichmentOptions }
+        ),
+        chat: await TextEditor.enrichHTML(
+          activity?.description?.value || (isIdentified ? description.chat || (activity ? "" : description.value)
+            : unidentified?.description) || "",
+          { relativeTo: activity?.description?.value ? activity : this.parent, ...enrichmentOptions }
```
Fyorl:
```suggestion
          { relativeTo: this.parent, ...enrichmentOptions }
```

module/data/abstract/item-data-model.mjs
```diff
@@ -210,12 +209,15 @@ export default class ItemDataModel extends SystemDataModel {
...
-        }),
-        chat: await TextEditor.enrichHTML(chat ?? "", {
-          rollData, relativeTo: this.parent, ...enrichmentOptions
-        }),
+        value: await TextEditor.enrichHTML(
+          (game.user.isGM || isIdentified ? description.value : unidentified?.description) ?? "",
+          { relativeTo: this.parent, ...enrichmentOptions }
+        ),
+        chat: await TextEditor.enrichHTML(
+          activity?.description?.value || (isIdentified ? description.chat || (activity ? "" : description.value)
```
Fyorl: Doesn't seem to match the priority mentioned in the PR description of activity description -> item chat description -> item full description.

```suggestion
          activity?.description?.value || (isIdentified ? description.chat || description.value
```

Fyorl (changes requested):

module/data/abstract/item-data-model.mjs
```diff
@@ -197,11 +197,10 @@ export default class ItemDataModel extends SystemDataModel {
     } = this;
     const rollData = (activity ?? this.parent).getRollData();
     const isIdentified = identified !== false;
-    const chat = isIdentified ? description.chat || description.value : unidentified?.description;
-    const desc = game.user.isGM || isIdentified ? description.value : unidentified?.description;
     uses = this.hasLimitedUses && (game.user.isGM || identified) ? uses : null;
     price = game.user.isGM || identified ? price : null;
 
+    enrichmentOptions = { rollData, relative: this.parent, ...enrichmentOptions };
```
Fyorl:
```suggestion
    enrichmentOptions = { rollData, relativeTo: this.parent, ...enrichmentOptions };
```

Fyorl (approved):

Issue #5372, opened 2025-03-13
Convert actor sheets to ApplicationV2


arbron:
Implement a ApplicationV2 version of the actor sheets and consolidate the various inherited sheets and mixins. 

- [x] Create ApplicationV2 version of base actor sheet
- [x] Consolidate `ActorSheet5eNPC` and `ActorSheet5eNPC2` into single new sheet
- [x] Consolidate `ActorSheet5eCharacter` and `ActorSheet5eCharacter2` into single new sheet
- [x] Remove `ActorSheetMixin` and `ActorSheetV2Mixin`
- [x] Remove legacy sheet templates

PR #5389, opened 2025-03-18
[#5372] Convert NPC sheet to ApplicationV2

arbron:
Merges `ActorSheet5eNPC` and `ActorSheet5eNPC` into `NPCActorSheet` and merges `ActorSheetMixin` and `ActorSheetV2Mixin` into the new `BaseActorSheet` that are all built on `ApplicationV2`.

Old actor sheets are fully removed but deprecated names are left in place pointing to the new sheets to avoid immediately breaking any modules that extend them. The old mixins have been left in place with deprecation warnings.


less/v2/npc.less
```diff
@@ -5,8 +5,11 @@
 .dnd5e2.sheet.actor.npc {
   --dnd5e-sheet-header-height: 68px;
   --dnd5e-sheet-sidebar-width: 180px;
-  // min-width: 700px;
-  // min-height: 700px;
+  min-width: 700px;
+  min-height: 700px;
```
arbron, line 9: I had these commented out because they break minimization. I attempted an approach of only having them when the sheet wasn't minimized, but that caused the minimization animation to break. Is there a proper way to handle minimum sheet sizes in AppV2 (we didn't set a minimum size for Compendium Browser too for this reason)?

Fyorl: There should be a way to do it: I remember going through some effort to allow App V2 journal sheet to have a min-width but also be minimisable. I'll have a look later.

Fyorl (changes requested):
Good stuff, thank you Jeff. I've done a pass myself with the changelog here so you don't have to hunt it down:

- Fixed wrong background texture.
- Fixed inactive tabs rendering too far left, and tab icon alignment.
- Fixed various vertical alignment issues due to line-height.
- Fixed re-renders briefly showing the warnings dialog and making the screen flash.
- Centred attunement input.
- Fixed dark mode header texture rendering above header input elements in edit mode.
- Adjusted spacing between sheet header and body.
- Removed excess padding around sheet body.
- Fixed missing active tab underline on items & activity sheets.
- Fixed rendering of encumbrance bar & child creation button.
- Fixed missing item-list-controls filters.

less/v2/npc.less
krbz999: `&:not(.minimized, .minimizing, .maximizing)` ?

templates/actors/tabs/creature-spells.hbs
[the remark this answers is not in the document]
arbron: Nope

templates/actors/tabs/creature-features.hbs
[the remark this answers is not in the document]
arbron: Nope

module/applications/actor/npc-sheet.mjs
```diff
@@ -1,204 +1,587 @@
...
-    for ( let item of other ) {
-      if ( item.type === "weapon" ) features.weapons.items.push(item);
-      else if ( ["background", "class", "feat", "race", "subclass"].includes(item.type) ) {
-        if ( item.system.activities?.size ) features.actions.items.push(item);
-        else features.passive.items.push(item);
-      }
-      else features.equipment.items.push(item);
+    for ( const item of context.itemCategories.features ?? [] ) {
+      if ( item.type === "weapon" ) sections.weapons.items.push(item);
+      else if ( item.system.activities?.size ) sections.actions.items.push(item);
```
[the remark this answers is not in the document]
arbron: Hmm, looking more closely at this I don't think this organization is actually required at this point, because the grouping system re-groups all of the items during render anyways, so we can just stick all of the items in any old section and everything else will be handled.

less/v2/npc.less
krbz999: Probably worth noting that AppV2 journal sheet does *not* have the above and it does not animate minimizing/maximizing.

https://github.com/foundryvtt/foundryvtt/issues/12316

arbron: Zhell's suggestion works

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -0,0 +1,1780 @@
...
+    },
+    form: {
+      submitOnChange: true
+    },
+    window: {
+      controls: [
+        {
+          action: "restoreTransformation",
+          icon: "fa-solid fa-backward",
+          label: "ND5E.TRANSFORM.Action.Restore",
```
krbz999, line 83: Gezundheit

less/variables.less
```diff
@@ -25,6 +25,9 @@
   --color-text-dark-inactive: #7a7971;
   --color-text-light-0: white;
   --color-text-dark-5: #666;
+  --color-text-light-0: #fff;
```
Fyorl, line 28: Already defined two lines up

Fyorl (approved):

PR #5390, opened 2025-03-18
[#5372] Convert Character sheet to ApplicationV2

### LINKED ISSUE #5372: Convert actor sheets to ApplicationV2

arbron:
Merges `ActorSheet5eCharacter` and `ActorSheet5eCharacter2` into `CharacterActorSheet` and removes the old sheets, leaving behind a small deprecation that extends the new sheets to avoid immediate breakage.

Closes #5372


module/applications/actor/character-sheet.mjs
```diff
@@ -1,317 +1,1227 @@
...
+
+  /**
+   * Handle toggling the death saves tray.
+   * @this {CharacterActorSheet}
+   * @param {Event} event         Triggering click event.
+   * @param {HTMLElement} target  Button that was clicked.
+   */
+  static #toggleDeathTray(event, target) {
+    this._toggleDeathTray();
+  }
```
Fyorl, line 1030: Missing separator.

Fyorl (changes requested):
Code looks good; only minor comments.

Two issues I spotted during testing:
1. Active effect values are being persisted to the special traits tab. Easiest way to reproduce this is probably just by adding a Ring of Protection, then toggle between edit & play mode while on the special traits tab and observe the global saving throw bonus field.
2. Unrelated to this sheet but wanted to record it: Clicking the 'create advancement' button on an item sheet does nothing.

Fyorl (approved):

PR #5397, opened 2025-03-19
[#5393] Add `<identifier-field>` used by `IdentifierInput`

### LINKED ISSUE #5393: Implement custom `<identifier-input>` that is used by `IdentifierField`
Introduce a custom element for inputting identifiers into the system that will be used automatically when creating a form with an identifier field. By default this would appear as just a normal text input with some preset validation to handle identifier restrictions.

`IdentifierField` would be able to specify what document types it supports (e.g.`new IdentifierField({ types: ["class"]] })`), which would allow the new `<identifier-input>` to tie into the system's item registry to provide identifier suggestions. In this case the identifier input would appear as a select dropdown with those options available so users don't have to type in one directly. At the bottom of this list would be a "Custom Identifier" option that would swap the field back to the text input.

This will make mistakes less likely with identifiers by providing early validation for manual inputs and a direct list of classes/subclasses/etc. when a specific input is required.

arbron:
Introduces a new `<identifier-field>` custom element that is used to enter identifiers. This supports the `types` attributes, which is a comma-separated list of item types. If one or more types are provided, then the identifier field will display as a select of the registered items of those types rather than a freeform input. A custom identifier can still be entered in that case by selecting the "Custom Identifier" option in the select, which toggles back to a freeform input.

This is used by `IdentifierInput`, so any forms that use Foundry's auto-generated form fields will get this new input element for free.

This adds backgrounds, species, and subclasses to the registry so they can be used in the new input.

Closes #5393


module/applications/components/identifier-input.mjs
```diff
@@ -0,0 +1,186 @@
...
+  /**
+   * Handle changes to one of the internal fields.
+   * @type {Event} event  Triggering change event.
+   */
+  _onChangeInput(event) {
+    event.stopPropagation();
+
+    // If "Custom Identifier" is selected, show input
+    if ( event.target.value === IdentifierInputElement.CUSTOM ) {
+      this.#internal.input.value = "";
```
Fyorl, line 177:
```suggestion
      this.#internal.input.value = this.value = "";
```

Fyorl (approved):

module/applications/components/identifier-input.mjs
arbron: That causes the form to submit and the input to change back to the `<select>` in any case where the current value isn't blank. We want to avoid changing the form value until users have entered something in the text box.

PR #5399, opened 2025-03-20
Handle `5.0` deprecations including those for activities & rolling

arbron:


module/data/shared/uses-field.mjs
```diff
@@ -283,25 +257,18 @@ export default class UsesField extends SchemaField {
...
-     * @function dnd5e.rollRechargeV2
+     * @function dnd5e.rollRecharge
      * @memberof hookEvents
      * @param {BasicRoll[]} rolls             The resulting rolls.
      * @param {object} data
      * @param {Item5e|Activity} data.subject  Item or Activity for which the roll was performed.
      * @param {object} data.updates           Updates to be applied to the subject.
      * @returns {boolean}                     Explicitly return `false` to prevent updates from being performed.
      */
+    if ( Hooks.call("dnd5e.rollRecharge", rolls, { subject: this, updates }) === false ) return rolls;
```
krbz999, line 268: Will the `V2` variants be deprecated?

arbron: The plan is to just leave them in while also calling the non-V2 hooks.

Fyorl (approved):

PR #5403, opened 2025-03-21
[#5402] Allow Enchant activity to target spells & features

### LINKED ISSUE #5402: Allow enchantment activity to apply to features & spells
Right now the Enchant activity has a restriction preventing it from working for non-physical item types, but enchantments themselves can be dragged over to those items freely. This restriction on the Enchant activity should be lifted so that spells and features can be enchanted using it. This would be useful for something like a "Agonizing Blast" enchantment that needs to modify a cantrip.

arbron:
Marks `spells` and `feat` as enchantable item types so they can be targeted with the Enchant activity. Also ignores the the "Allow Magical" option if the item the enchantment is being applied to is not physical so that it doesn't need to be checked to apply to spells, which always have the magical property.

Closes #5402

### DIFF
--- a/module/applications/activity/enchant-sheet.mjs
+++ b/module/applications/activity/enchant-sheet.mjs
@@ -68,6 +68,8 @@ export default class EnchantSheet extends ActivitySheet {
         .filter(t => enchantableTypes.has(t))
         .map(value => ({ value, label: game.i18n.localize(CONFIG.Item.typeLabels[value]) }))
     ];
+    context.isTypePhysical = !context.source.restrictions.type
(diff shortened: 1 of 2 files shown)


lang/en.json
```diff
@@ -1917,7 +1917,7 @@
       "hint": "Restrictions on the type of item to which this enchantment can be applied.",
       "allowMagical": {
         "label": "Allow Magical",
-        "hint": "Allow items that are already magical to be enchanted."
+        "hint": "Allow physical items that are already magical to be enchanted."
```
Fyorl, line 1920: This field should be hidden in the UI if you add a restriction on a non-physical item type.

Fyorl (approved):

Issue #963, opened 2021-01-19
Merge Skill Proficiencies during a Transform Actor workflow causes a break in the behaviour of skill-affecting Active Effects toggles


aaclayton:
Originally in GitLab by @unsoluble1

There appears to be an unexpected interaction between the `Merge Skill Proficiencies` process in the Transform Actor workflow (either when checked as part of a Custom transformation, or as included by default in a Wild Shape) and active effects that are pre-existing on skills.

1. Add a passive effect to an item that modifies something in `data.skills`.
2. Add item to actor.
3. Make sure effect is toggled on.
4. Wild Shape the actor into something.
5. Play with the effect's toggle, note that while the UI suggests it's turning on and off, the effect won't ever actually turn off.

If you do this but make sure the effect is toggled _off_ before doing the transform, then while transformed the toggle now works as expected.

(Tested and repro'd in 1.2.3)

PR #5416, opened 2025-03-24
[#1824] Adjust how merged skills/saves are handled in transform

### LINKED ISSUE #1824: Handling of merging saving throws vs. skill proficiencies is inconsistent when transforming actors
Currently, when merging skill proficiencies, the formula we use is: `max(targetSkillProf, originalSkillProf) + skillAbilityMod`.

Merging saving throws differs in two ways: Firstly, we only attempt the merge if the target (the currently transformed actor) has that saving throw proficiency. If they do, the formula we use is `max(targetSaveBonus, originalSaveBonus)`.

The various rulings around Wild Shape, Shapechange, Polymorph, etc. are very hard to pin down and interpretations differ quite a bit. I don't want to re-open this can of worms, but we should at least pick one reasonable interpretation and be consistent between both skills and saves.

arbron:
Modifies how skills and saves are handled when they are merged during transformation to ensure the source actor's modifier is always used if it is higher than the target's modifier. This is done by calculating the difference between the two final modifiers and adding that to the save/skill bonus.

Will not take into account dice-based bonuses such as those from the Dragonmarked races in Eberron.

Closes #1824

### DIFF
--- a/module/data/actor/templates/common.mjs
+++ b/module/data/actor/templates/common.mjs
@@ -159,8 +159,13 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
 
       if ( !Number.isFinite(abl.max) ) abl.max = CONFIG.DND5E.maxAbilityScore;
 
-      // If we merged saves when transforming, take the highest bonus here.
-      if ( originalSaves && abl.proficient ) abl.save.value = Math.max(abl.save, originalSaves[id].save.value);
+      // If we merged saves when transforming, take the highest bonus
+      const difference = (originalSaves?.[id].save.value ?? 0) - abl.save.value;
+      if ( originalSaves && (difference > 0) ) {
+        abl.bonuses.save = `${abl.bonuses.save ?? ""} + ${difference}`;
+        abl.saveBonus += difference;
+        abl.save.value += difference;
+      }
 
       // Deprecations.
       abl.save.toString = function() {
--- a/module/data/actor/templates/creature.mjs
+++ b/module/data/actor/templates/creature.mjs
@@ -223,9 +223,6 @@ export default class CreatureTemplate extends CommonTemplate {
     skillData.ability = ability;
     const baseBonus = simplifyBonus(skillData.bonuses?.check, rollData);
 
-    // Polymorph Skill Proficiencies
-    if ( originalSkills ) skillData.value = Math.max(skillData.value, originalSkills[skillId].value);
-
     // Compute modifier
     const checkBonusAbl = simplifyBonus(abilityData?.bonuses?.check, rollData);
     skillData.effectValue = skillData.value;
@@ -235,6 +232,14 @@ export default class CreatureTemplate extends CommonTemplate {
     skillData.value = skillData.proficient = skillData.prof.multiplier;
     skillData.total = skillData.mod + skillData.bonus;
     if ( Number.isNumeric(skillData.prof.term) ) skillData.total += skillData.prof.flat;
+
+    // If we merged skills when transforming, take the highest bonus
+    const difference = (originalSkills?.[skillId].total ?? 0) - skillData.total;
+    if ( originalSkills && (difference > 0) ) {
+      skillData.bonuses.check = `${skillData.bonuses.check ?? ""} + ${difference}`;
+      skillData.bonus += difference;
+      skillData.total += difference;
+    }
 
     // Compute passive bonus
     const passive = flags.observantFeat && CONFIG.DND5E.characterFlags.observantFeat.skills.includes(skillId) ? 5 : 0;


arbron:
Been thinking about how to resolve https://github.com/foundryvtt/dnd5e/issues/963 is relation to this, and it seems like the best way to do it is to re-calculate the original actor's save/skill total without taking active effects into account. That will involve a bit of a larger re-working of the ability and skill preparation methods to move part of the calculation into a separate method so is probably best handled on its own.

arbron:
> The thing I'm not yet understanding is why we need to calculate the difference and apply a bonus instead of using `Math.max` like before?
> 
> ```js
> // Same as before, but remove check for abl.proficient
> abl.save.value = Math.max(abl.save.value, originalSaves[id].save.value);
> 
> // This I would think is the only other change we need
> skillData.total = Math.max(skillData.total, originalSkills[skillId].total);
> ```

The issue that we have there is that skill rolls are not constructed using the total, they are build using the individual components. If we use total, then the sheet shows the correct value but it is wrong when you try to roll it.

Upon re-reading the new Wild Shape text, I think we still might need to change this:
- Use highest proficiency multiplier between host and source
- If proficiency comes from host, use host's proficiency bonus, otherwise use source's
- If **total modifier** is higher on source than host, use source's modifier (but not necessarily the other way around)

Fyorl:
> The issue that we have there is that skill rolls are not constructed using the total, they are build using the individual components. If we use total, then the sheet shows the correct value but it is wrong when you try to roll it.

That makes sense. I guess it can be a bit difficult to track down where this bonus is coming from when rolling. Not sure if there's anything we can do about that right now. We'd need to build out a better system for breakdowns first.

> Upon re-reading the new Wild Shape text, I think we still might need to change this:
> 
> > You also retain your skill and saving throw proficiencies and use your Proficiency Bonus for them, in addition to gaining the proficiencies of the creature.
> 
> * Use highest proficiency multiplier between host and source
> * If proficiency comes from host, use host's proficiency bonus, otherwise use source's

That looks correct to me, yeah.

> > If a skill or saving throw modifier in the Beast’s stat block is higher than yours, use the one in the stat block.
> 
> * If **total modifier** is higher on source than host, use source's modifier (but not necessarily the other way around)

Can you elaborate on this? If we use the host's modifier by default, and the source's only if it's higher, then we are effectively just doing `max(hostModifier, sourceModifier)` no?

module/data/actor/templates/common.mjs
```diff
@@ -141,14 +141,21 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
     const dcBonus = simplifyBonus(this.bonuses?.spell?.dc, rollData);
     for ( const [id, abl] of Object.entries(this.abilities) ) {
       if ( flags.diamondSoul ) abl.proficient = 1;  // Diamond Soul is proficient in all saves
+      const originalAbility = originalSaves?.[id];
+      if ( originalAbility?.proficient > abl.proficient ) {
```
Fyorl: `proficient` is merged at transform time

```js
else if ( settings.merge.has("saves") && oa ) abilities[k].proficient = Math.max(prof, oa.proficient);
```

So this line can never evaluate to true. Moreover, according to my reading, we only ever need to check if the host was originally proficient. If so, we must use their proficiency bonus.

> You also retain your skill and saving throw proficiencies and use your Proficiency Bonus for them

```suggestion
      if ( originalAbility?.proficient ) {
```

module/data/actor/templates/common.mjs
```diff
@@ -141,14 +141,21 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
...
+        abl.merged = true;
+        abl.proficient = originalAbility?.proficient;
+      }
 
-      abl.checkProf = this.calculateAbilityCheckProficiency(0, id);
+      const calculatedProf = this.calculateAbilityCheckProficiency(0, id, originalAbility);
+      abl.checkProf = originalAbility?.checkProf?.multiplier > calculatedProf.mulitplier
+        ? originalAbility.checkProf.clone() : calculatedProf;
```
Fyorl, line 152: `checkProf` might not exist. There's no guarantee the original Actor was prepared before this one.

module/data/actor/templates/creature.mjs
```diff
@@ -222,20 +222,32 @@ export default class CreatureTemplate extends CommonTemplate {
     const abilityData = this.abilities[ability];
     skillData.ability = ability;
     const baseBonus = simplifyBonus(skillData.bonuses?.check, rollData);
-
-    // Polymorph Skill Proficiencies
-    if ( originalSkills ) skillData.value = Math.max(skillData.value, originalSkills[skillId].value);
+    const originalSkill = originalSkills?.[skillId];
+    if ( originalSkill?.value > skillData.value ) {
```
Fyorl:
```suggestion
    if ( originalSkill?.value >= 1 ) {
```

module/documents/actor/actor.mjs
```diff
@@ -1238,17 +1240,21 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+  _buildSkillToolConfig(type, hostActor, process, config, formData, index) {
     const relevant = type === "skill" ? this.system.skills?.[process.skill] : this.system.tools?.[process.tool];
     const rollData = this.getRollData();
     const abilityId = formData?.get("ability") ?? process.ability;
     const ability = this.system.abilities?.[abilityId];
-    const prof = this.system.calculateAbilityCheckProficiency(relevant?.effectValue ?? 0, abilityId);
+    let prof = this.system.calculateAbilityCheckProficiency(relevant?.effectValue ?? 0, abilityId);
+    const originalProf = hostActor?.system.calculateAbilityCheckProficiency?.(
+      hostActor?.system.skills[process.skill]?.value, abilityId);
+    if ( originalProf?.multiplier > prof.multiplier ) prof = originalProf;
```
Fyorl, line 1257: Didn't we already account for this during preparation? Is there also an issue whereby we swap to using the host's proficiency here and that causes the total modifier to drop below the wild shape source's modifier?

module/data/actor/templates/common.mjs
arbron: Its existence should be verified by the previous line when we validated the `checkProf?.multiplier` was greater than the calculated prop.

module/documents/actor/actor.mjs
arbron: When switching the ability associated with a skill, you may run into a situation where you gain/lose proficiency given by remarkable athlete, so we need to re-check the proficiency used to account for that.

Issue #5394, opened 2025-03-19
Bug: No concentration prompt when you have Temp. HP remaining


nschoenwald:
When you take damage while you have Temp. HP and the damage does not "touch" your real HP, you don't get a concentration save prompt in chat.

thatlonelybugbear:
Is that with no other modules active?

tposney:
I tested with no modules enabled (can reproduce) and had a look at the code.
The reason there is no concentration check when taking temphp damage is the test for current.value < current.effectiveMax. EffectiveMax makes no reference to temphp, just hp.max + hp.tempMax, which means the test will fail if no "real" hp damage has been taken.

I don't really understand in which cases a concentration check should be made and if this is working as intended or an oversight.

```js
    if ( !game.settings.get("dnd5e", "disableConcentration") && (userId === game.userId) && (changes.total < 0)
          && (options.dnd5e?.concentrationCheck !== false) && (curr.value < curr.effectiveMax) ) {
          this.challengeConcentration({ dc: this.getConcentrationDC(-changes.total) });
```

Fyorl:
It's a bug. A character makes a Concentration save when they take damage. If they lose temp HP as a result of taking damage, they should make a save. If they lose temp HP as a result of the temp HP expiring or being replaced, they shouldn't make a save.

tposney:
Just for my understanding I assume the temp hp expiring is covered by options.dnd5e.concentrationCheck !== false?
As a feature request would it be possible to pass the options parameter to the dnd5e.damageActor hook call, being able to access that same flag would be useful for some processing I'd like to trigger off that call (currently I have to hook updateActor to get access to the options).

Fyorl:
> As a feature request would it be possible to pass the options parameter to the dnd5e.damageActor hook call, being able to access that same flag would be useful for some processing I'd like to trigger off that call (currently I have to hook updateActor to get access to the options).

I'd have to check the code to be sure but I wouldn't expect it to be a problem to make that change.

Division4Studios:
---

### **Updated Description**

#### **Problem**
The current implementation of concentration checks in the `_onUpdate` method uses the condition `(curr.value < curr.effectiveMax)` to determine if a concentration check should be triggered. However, this condition is never satisfied when temporary HP (`tempHP`) is reduced, as `effectiveMax` does not account for temporary HP. This results in no concentration check being triggered when temporary HP is lost, even though the rules state that a concentration check should occur whenever damage is taken.

[...]

Issue #3955, opened 2024-08-08
Concentration check triggered when reverting hp.tempmax and the actor's hp.value is higher than hp.max


thatlonelybugbear:
Split from #3363 

- Creature gets a boon on tempmax and afterwards is healed above the normal hp.max.
- The tempmax boon reverts. The "hp loss" will trigger a concentration check, which is not considered actual damage and against RAW Concentration.

Maybe a change in `actor.mjs:3307` would make sense?
```diff
async _onUpdate(data, options, userId) {
    super._onUpdate(data, options, userId);
    if ( userId === game.userId ) {
      await this.updateEncumbrance(options);
      this._onUpdateExhaustion(data, options);
    }

     const hp = options.dnd5e?.hp;
     if ( hp && !options.isRest && !options.isAdvancement ) {
        const curr = this.system.attributes.hp;
        const changes = {
        hp: curr.value - hp.value,
        temp: curr.temp - hp.temp,
      };
      changes.total = changes.hp + changes.temp;

      if ( Number.isInteger(changes.total) && (changes.total != 0) ) {
        this._displayTokenEffect(changes);
-     if ( !game.settings.get("dnd5e", "disableConcentration") && (userId === game.userId) && (changes.total < 0) ) {
+     if ( !game.settings.get("dnd5e", "disableConcentration") && (userId === game.userId) && (changes.total < 0) && (curr.value != curr.effectiveMax) ) {

          this.challengeConcentration({ dc: this.getConcentrationDC(-changes.total) });
        }
        //...
 ```

PR #4210, opened 2024-09-13
[#3955] Concentration check triggered when reverting hp.tempmax and the actor's hp.value is higher than hp.max

thatlonelybugbear:
Closes #3955, by not triggering concentration checks when any changes to temporary Max HP are being reverted, making use of that fact if the effectiveMax is being lowered it will lower the hp.value to the same number.

PR #5424, opened 2025-03-27
[#5394] Fix Concentration Check Logic to Address Multiple Issues

Division4Studios:
#### **Problem**
The concentration check logic in `_onUpdate` has undergone several iterations to address different issues, but the current implementation still has edge cases that need to be resolved. Specifically:
1. **Original Issue**: The condition `(curr.value < curr.effectiveMax)` was added in a previous PR (#3955) to prevent concentration checks from triggering when temporary HP (`tempHP`) was reduced but no "real" HP damage occurred. However, this condition caused concentration checks to fail in valid scenarios where temporary HP was reduced, as `effectiveMax` does not account for temporary HP.
2. **Manual HP Adjustments**: Moving the concentration check logic entirely to `applyDamage` was considered but raised concerns about manual HP adjustments (e.g., fall damage) not triggering concentration checks.
3. **Edge Cases**: The logic needed to balance handling temporary HP reductions, manual HP adjustments, and ensuring concentration checks align with the rules.

#### **Solution**
To address these issues, the concentration check logic has been updated as follows:
1. **Restored Logic to `_onUpdate`**:
   - The concentration check logic remains in `_onUpdate` to ensure that all HP changes, including manual adjustments, are evaluated for concentration checks.
   - This approach aligns with the original design intent and ensures consistency.

2. **Removed Faulty Condition**:
   - The condition `(curr.value < curr.effectiveMax)` was removed because it fails to account for temporary HP and prevents valid concentration checks when temporary HP is reduced.

3. **Simplified and Improved Logic**:
   - Concentration checks are now triggered based on whether the total HP change (`changes.total`) is negative, ensuring that any reduction in HP (including temporary HP) is evaluated.
   - The condition `(changes.temp < 0 || curr.value < curr.effectiveMax)` was retained to ensure that temporary HP reductions are handled correctly while still respecting the original intent of PR #3955.

4. **Tested for Edge Cases**:
   - The updated logic has been tested to ensure it works correctly for scenarios involving temporary HP reductions, manual HP adjustments, and other edge cases.

#### **Final Implementation**
Here is the updated concentration check logic:
```javascript
// Determine if concentration check should be made
const isConcentrationEnabled = !game.settings.get("dnd5e", "disableConcentration");
const isUserInitiated = userId === game.userId;
const isDamageTaken = changes.total < 0;
const isConcentrationCheckAllowed = options.dnd5e?.concentrationCheck !== false;
const isHpReduced = changes.temp < 0 || curr.value < curr.effectiveMax;

if (isConcentrationEnabled && isUserInitiated && isDamageTaken && isConcentrationCheckAllowed && isHpReduced) {
  // Make the check using the total as input for the concentration DC.
  this.challengeConcentration({ dc: this.getConcentrationDC(-changes.total) });
}
```

#### **Why This Approach?**
1. **Consistency**:
   - Keeping the logic in `_onUpdate` ensures that all HP changes, including manual adjustments, are evaluated for concentration checks.
   - This aligns with the original design intent and avoids breaking existing functionality.

2. **Handling Temporary HP**:
   - Removing `(curr.value < curr.effectiveMax)` resolves the issue where temporary HP reductions were not triggering concentration checks.
   - The condition `(changes.temp < 0 || curr.value < curr.effectiveMax)` ensures that temporary HP reductions are handled correctly.

3. **Flexibility**:
   - This approach balances the need to handle temporary HP reductions, manual HP adjustments, and other edge cases without introducing unnecessary complexity.

4. **Community Feedback**:
   - The updated logic incorporates feedback from previous discussions and PRs, ensuring that it addresses the concerns raised by contributors.

#### **Impact**
- Concentration checks now trigger correctly for all valid scenarios, including temporary HP reductions and manual HP adjustments.
- The logic is simpler, more maintainable, and aligns with the official rules for concentration checks.
- Edge cases from previous implementations are resolved, ensuring consistent behavior.

---

Let me know if you need further refinements!


krbz999:
The current implementation was intentional.

Moving it out would also prevent players from triggering concentration.

Division4Studios:
@thatlonelybugbear 

That is actually a very valid point.  I updated it to this implementation because found it annoying during testing to see the messages pop up when I was just resetting temp hp back to 0 over and over during testing.  But you raise a good point.  Convenience/Utility over Correctness.

My first implementation was actually fixing it simply to also register when lowering temp/normal hp.  But then found the concentration popup annoying.  However thats just anecdotal and from my experience doing some repetitive testing. 

I appreciate the feedback from each of you. As well as the quick response.

@thatlonelybugbear and @krbz999 

I will update the pull request and the description.

thatlonelybugbear:
I need to add that the `curr.value < curr.effectiveMax` check had been added so that it fixed another issue.

See <https://github.com/foundryvtt/dnd5e/issues/3955> and [PR](<https://github.com/foundryvtt/dnd5e/pull/4210>)
Which had been closed via [commit](<https://github.com/foundryvtt/dnd5e/commit/800f02b2e1116285184286676cd5bc3b49f97f4d>) adding the `<` check.

I think that it needs maybe some check for
```js
if (!game.settings.get("dnd5e", "disableConcentration") 
    && (userId === game.userId) 
    && (changes.total < 0) 
    && (options.dnd5e?.concentrationCheck !== false)
    && (changes.temp < 0 || curr.value < curr.effectiveMax) {
  this.challengeConcentration({ dc: this.getConcentrationDC(-changes.total) });
}
```

Division4Studios:
@thatlonelybugbear That worked perfectly as far as my testing.  Cleaned up the implementation for easier to understand variables so if future work needs to be done its abit easier to understand.  Updated the PR and Description.

Division4Studios:
Also thank you @thatlonelybugbear and @krbz999 for being patient with me on my first open source PR here.  Hopefully i'll become more familiar with the overall project and history and be able to be a regular contributor.

module/documents/actor/actor.mjs
```diff
@@ -3343,8 +3343,20 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
         this._displayTokenEffect(changes);
-        if ( !game.settings.get("dnd5e", "disableConcentration") && (userId === game.userId) && (changes.total < 0)
-          && (options.dnd5e?.concentrationCheck !== false) && (curr.value < curr.effectiveMax) ) {
+
+        // Determine if concentration check should be made
+        const isConcentrationEnabled = !game.settings.get("dnd5e", "disableConcentration");
+        const isUserInitiated = userId === game.userId;
+        const isDamageTaken = changes.total < 0;
+        const isConcentrationCheckAllowed = options.dnd5e?.concentrationCheck !== false;
+        const isHpReduced = changes.temp < 0 || curr.value < curr.effectiveMax;
```
krbz999, line 3352:
```suggestion
        const isHpReduced = (changes.temp < 0) || (curr.value < curr.effectiveMax);
```

PR #5432, opened 2025-03-27
Swap to using cursor variables

arbron:
Swaps out everything except for uses of `cursor: help` and `cursor: not-allowed` which aren't supported as custom types.


Fyorl (approved):
Lovely, thank you

Issue #5445, opened 2025-03-30
Wiki: Explain scaling better in activities


nschoenwald:
https://github.com/foundryvtt/dnd5e/wiki/Activities

The page talks about scaling a lot but never mentions how to actually use the scaling value. That information can be found here, but I had to go to the discord to find it.

https://github.com/foundryvtt/dnd5e/wiki/Roll-Formulas#item-properties

Division4Studios:
@nschoenwald 

I have addressed your concerns here.  I also found some more areas regarding scaling that might be helpful.  So I put them in as well.

Here is the PR so you can see.

#5448

PR #5448, opened 2025-03-31
[#5445] Add Scaling References to Activities Configuration Section

Division4Studios:
#### **What Prompted the Issue**
An issue #5445 was raised regarding the lack of clarity in the Activities.md wiki page about how scaling can be used in activity configurations. While the page mentions scaling in certain contexts, it does not provide sufficient guidance or references for users to understand how to implement scaling effectively. This led to confusion, requiring users to search through other documentation or external resources (e.g., Discord) to find the necessary information.

#### **How We Addressed It**
To resolve this, we updated the **Configuring Activities** section of the Activities.md file to include references to relevant documentation that explains scaling in detail. Specifically:
- Added a note that some configurations, such as those in the **Consumption** or **Effect** tabs, allow for the use of scaling.
- Provided direct links to the following resources for more information and examples:
  - Roll Formulas - Scaling
  - Roll Formulas - Item Properties
  - Scale Value Advancements

This approach avoids duplicating content while ensuring users can easily find the information they need to understand and use scaling effectively.

### DIFF
--- a/wiki/Activities.md
+++ b/wiki/Activities.md
@@ -30,7 +30,12 @@ There are a variety of activity types offered by the DnD5e system, and modules c
 
 ## Configuring Activities
 
-While parts of the activity sheet change based on the selected activity type, many parts are shared.
+While parts of the activity sheet change based on the selected activity type, many parts are shared. Some configurations, such as those in the **Consumption** or **Effect** tabs, allow for the use of scaling to dynamically adjust an activity's behavior based on contextual factors like spell level or item properties.
+
+For more information and examples of how scaling can be used, see:
+- [Roll Formulas - Scaling](Roll-Formulas.md#scale)
+- [Roll Formulas - Item Properties](Roll-Formulas.md#item-properties)
+- [Scale Value Advancements](Advancement-Type-Scale-Value.md)
 
 ### Identity
 


roth-michael:
I actually think that the "Scale" section of Roll Formulas and the Scale Value Advancements pages may not be relevant in this case, as those don't so much have to do with "Scaling" as "Scale Values" specifically, which are different concepts in the system (scaling is about variable consumption, scale values are for referencing data that changes as a character gains class levels). I also wonder (and on this I'm less certain) whether it would make sense to instead link to the Item Properties section further down instead, by
> The "Consumption Scaling" section indicates whether scaling is allowed and how much scaling can occur. This section will not appear for spells, because the spell’s level determines how it scales.

Division4Studios:
Thank you for the feedback! Based on your suggestions, I’ve made the following updates:

1. **Clarified Scaling Context**:
   - I’ve ensured that the references focus on scaling as it pertains to activities, specifically in the **Consumption** and **Effect** sections.
   - The references now highlight how scaling dynamically adjusts behavior based on contextual factors like spell level or charges consumed.

2. **Updated References**:
   - Removed references to the "Scale" section and Scale Value Advancements, as these are more relevant to "Scale Values" tied to character progression rather than activity scaling.
   - Added a note in the **Consumption** section to explain how scaling applies to variable consumption.
   - Retained the note in the **Effect** section to explain how scaling can adjust applied effects dynamically.

3. **Direct Links to Relevant Sections**:
   - Both sections now link to the **Roll Formulas - Item Properties** page, which provides examples and explanations directly relevant to activity scaling.

@roth-michael let me know if there’s anything else you’d like to see clarified or adjusted!

PR #5451, opened 2025-03-31
[#5450] Fix for defaulting to no mastery under some circumstances

### LINKED ISSUE #5450: [Bug] Weapon mastery on an attack defaults to none if no `last` flag exists and there are "bonus" masteries on the actor
Problem stems from the `{ rule: true }` entry not having an associated `value`, leading to 
```js
const masteryOptions = this.item.system.masteryOptions;
if ( !masteryOptions?.find(m => m.value === rollConfig.mastery) ) { // <-- this line
  rollConfig.mastery = masteryOptions?.[0]?.value;
}
```
returning `true` when `rollConfig.mastery` is `undefined`

roth-michael:
Closes #5450 
Add `rule: true` to the existing first mastery, rather than creating it as a standalone object in the list.

### DIFF
--- a/module/data/item/weapon.mjs
+++ b/module/data/item/weapon.mjs
@@ -495,11 +495,12 @@ export default class WeaponData extends ItemDataModel.mixin(
     const extras = [];
     for ( const mastery of this.parent.actor.system.traits.weaponProf.mastery.bonus ?? [] ) {
       if ( mastery === this.mastery ) continue;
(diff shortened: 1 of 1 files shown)


arbron:
Could you give this a rebase onto the latest `4.4.x` branch?

roth-michael:
There we go - done.

arbron (approved):

PR #5477, opened 2025-04-11
Refactor inventory rendering

Fyorl:
Refactors inventory rendering. Removes various hard-codings, and makes things a little more modular for anticipated use when converting the remaining two Actor sheets.

Sorry about the size of the diff, but you'll see that a lot of it is boilerplate/config.

Various key improvements:
- Consolidates rendering a list of items with discrete sections into a single template.
- Removes any v12 baggage from `InventoryElement` and `ItemListControlsElement`.
- Removes various hard-coding around sorting and grouping modes from `ItemListControlsElement`. The element can now support any number of sorting or grouping modes.
- Column configuration is now modular. Each section can support its own set of columns, with their own ordering. Column priority is managed without the use of hard-coded `@container` queries. Columns can be hidden based on available width and priority ordering.
- Hoisted Actor-type specific elements out of inventory rendering and into their own sheets. The base inventory template supports only currency, search, and encumbrance by default, with the various sheets inject their bespoke elements back in.

Minor improvements:
- Converted the filter drop-down to use `ContextMenu`, resolving some issues where it would get truncated if the filter element was placed too close to the edge of the sheet.
- Container inventories now use the NPC sheet's style of rendering nested containers, reducing wasted space at the top of the sheet.
- Utilised modular column config to show attack and damage rolls in the weapon section of the character sheet inventory.


templates/inventory/columns/range.hbs
Fyorl, line 1: Some of these column templates could do with another pass, but I ran out of time. We should consolidate context preparation for items and activities so that they are using the same render context or labels properties.

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -1028,6 +1127,12 @@ export default class BaseActorSheet extends PrimarySheetMixin(
       this.element.classList.toggle("sidebar-collapsed", sidebarCollapsed);
     }
 
+    // Attunement
+    this._renderAttunement(context, options);
+
+    // Spellbook
+    this._renderSpellbook(context, options);
```
arbron, line 1134: Do we want to be rendering either of these for other actor types such as groups or vehicles? Might be better to call these methods in the character and NPC sheets, rather than in the base sheet.

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -196,6 +211,7 @@ export default class BaseActorSheet extends PrimarySheetMixin(
     // Cache concentration data and prepare items
     this._concentration = this.actor.concentration;
     await this._prepareItems(context);
+    context.spellbook = this._prepareSpellbook(context);
```
arbron, line 214: Should probably leave this to the subclassed sheets to call so we don't end up preparing the spellbook for actor types that don't have it like vehicles and groups..

module/data/abstract.mjs
```diff
@@ -440,17 +440,15 @@ export class ItemDataModel extends SystemDataModel {
...
    * @property {boolean} singleton      Should only a single item of this type be allowed on an actor?
+   * @property {InventorySectionDescriptor} [inventory]  Configuration for displaying this item type in its own section
+   *                                                     in creature inventories.
    */
 
   /** @type {ItemDataModelMetadata} */
   static metadata = Object.freeze(foundry.utils.mergeObject(super.metadata, {
     enchantable: false,
     hasEffects: false,
-    inventoryItem: false,
```
arbron, line 452: Would probably be good to have some sort of deprecation period for this change. Just in my quick testing with Tasha's this caused the tattoo items to appear in the wrong tab and be hard to find in the Compendium Browser.

module/applications/components/inventory.mjs
```diff
@@ -47,268 +269,227 @@ export default class InventoryElement extends HTMLElement {
...
+  _getContextOptions(item, element) {
+    // Standard options.
+    const options = [{
+      name: "DND5E.ItemView",
+      icon: '<i class="fa-solid fa-eye fa-fw"></i>',
+      callback: li => this._onAction(li, "view")
+    }, {
+      name: "DND5E.ContextMenuActionEdit",
+      icon: '<i class="fa-solid fa-edit fa-fw"></i>',
+      condition: () => item.isOwner && !item.collection.locked,
```
arbron, line 333: `item.collection.locked` doesn't evaluate to `true` in a locked compendium, this will probably need to check `item.collection.locked || item.actor.collection.locked` to ensure it handles locked compendiums and actors with only observer permissions.

module/applications/components/inventory.mjs
```diff
@@ -365,155 +546,313 @@ export default class InventoryElement extends HTMLElement {
...
-        return this.actor.system.removeFavorite(item.getRelativeUUID(this.actor));
-      case "use":
-        return item.use({ legacy: false, event });
+  /**
+   * Manage columns when the inventory element's inline size changes.
+   * @param {ResizeObserverEntry[]} entries
+   * @protected
+   */
+  _onResize([entry]) {
+    // TODO: Should accommodate uiScale here, but probably don't want to call game.settings.get every frame.
```
arbron, line 636: We could grab it from the styles I suppose: `document.defaultView.getComputedStyle(this).getPropertyValue("--ui-scale")`.

module/applications/components/inventory.mjs
```diff
@@ -365,155 +546,313 @@ export default class InventoryElement extends HTMLElement {
...
+  /**
+   * Handle activating an item.
+   * @param {Item5e} item                   The item.
+   * @param {object} [options]
+   * @param {PointerEvent} [options.event]  The triggering event.
+   * @returns {Promise<ActivityUsageResults|ChatMessage|object|void>}
+   * @protected
+   */
+  _onUseItem(item, { event }={}) {
+    return item.use({ legacy: false, event });
```
arbron, line 773: Legacy is no more!

```suggestion
    return item.use({ event });
```

lang/en.json
```diff
@@ -170,18 +170,76 @@
     "Vehicle": "Vehicle"
   },
   "Type": {
+    "Action": {
+      "Label": {
+        "one": "Action",
+        "other": "Actions"
+      }
```
arbron, line 177: I wonder whether we should do these slightly differently, since it is my understanding that this `PluralRules` style is only supposed to be used when the count is present in the localization (e.g. `2 actions`), and isn't necessarily accurate without the count.

```suggestion
      "Header": "Actions",
      "Label": "Action"
```

arbron (changes requested):
Looking pretty good 👌

Noticed one other issue during testing outside of your PR, looks like you get an error in `changeTab` for an actor in a locked compendium because the create child button isn't present. Should be a quick fix by adding optional chaining to line `1241` in `base-actor-sheet.mjs`.

templates/inventory/columns/quantity.hbs
```diff
@@ -0,0 +1,17 @@
+<div class="item-detail item-quantity {{#unless isItem}}empty{{/unless}}" data-column-id="quantity">
+    {{#if isItem}}
+    {{#if @root.owner}}
+    <a class="adjustment-button always-interactive" data-action="decrease" data-property="system.quantity">
```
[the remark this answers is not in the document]
Fyorl: > Is this a known V13 issue?

Nope, good catch.

module/applications/actor/api/base-actor-sheet.mjs
Fyorl: Maybe we will need spellbook on groups for group rituals. I'll move it out for now.

module/data/abstract.mjs
Fyorl: Factored this out to `ItemDataModel.inventorySection` getter and added a deprecation path for `ItemDataModel.metadata.inventoryItem`.

arbron (approved):

PR #5484, opened 2025-04-14
[#5483] Add stat blocks formatted for 2014 rules

### LINKED ISSUE #5483: Add legacy version of embedded stat blocks
For NPCs with the old rules set, display stat blocks formatted in the old style.

arbron:
Modifies the embed template to move the definition lists into the context rather than the template to support different lists on the two different styles of stat blocks. Then apply different styling depending on what rules version is selected.

Stat Block 2014 vs 2024

The styling is near the 2014 versions, but the background and borders are missing some of the texture that is present in the official designs.

Also adds `attributes.ac.label` which is used when displaying the stat block.

Closes #5483

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1132,6 +1132,7 @@
     "FastForwardHint": "<kbd>Shift</kbd> + <left-click> the item to skip this dialog"
   }
 },
+"DND5E.Challenge": "Challenge",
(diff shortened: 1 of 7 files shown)


arbron:
Added dark mode styles:

Stat Block 2014 vs 2024 Dark

PR #5497, opened 2025-04-17
[#5468] Fix errors when migrating items with changed types in V13

### LINKED ISSUE #5468: System migrations that change the document's type fail in V13
- Item: `backpack` -> `container`
- ActiveEffect: flag -> `enchantment`
- ... (?)

For testing install https://raw.githubusercontent.com/Carter7777/brancalonia-bigat/master/world.json.

```
Error: Failed dnd5e system migration for Actor Sœur Fantagaro: The type of a Document can be changed only if the system field is force-replaced (==) or updated with {recursive: false}
```

arbron:
Use `recursive: false` during migration updates to fix error thrown for items and active effects with updated types.

Closes #5468


module/migration.mjs
```diff
@@ -29,7 +29,7 @@ export async function migrateWorld({ bypassVersionCheck=false }={}) {
           updateData = foundry.utils.mergeObject(source, updateData, {inplace: false});
         }
         await actor.update(updateData, {
-          enforceTypes: false, diff: valid && !flags.persistSourceMigration, render: false
+          enforceTypes: false, diff: valid && !flags.persistSourceMigration, recursive: false, render: false
```
Fyorl, line 32: If `updateData` is not the full source, then updating with `recursive: false` might do things like truncate the entire items collection.

```suggestion
          enforceTypes: false, diff: valid && !flags.persistSourceMigration, recursive: !flags.persistSourceMigration, render: false
```

Fyorl (changes requested):

Fyorl (commented):
Code looks good, but these non-recursive updates can be quite dangerous. I'd like to see if we can solicit some pre-migrated worlds with real data so that we can check to make sure everything is working correctly before we merge.

arbron:
> Code looks good, but these non-recursive updates can be quite dangerous. I'd like to see if we can solicit some pre-migrated worlds with real data so that we can check to make sure everything is working correctly before we merge.

Ran it on one of my old worlds with the old enchantments using flags and they converted properly over to enchantments with the specific type.

PR #5502, opened 2025-04-18
Use Token Name instead of Actor ProtoToken Name for Combat Groups

MaxPat931:
image

### DIFF
--- a/module/applications/combat/combat-tracker.mjs
+++ b/module/applications/combat/combat-tracker.mjs
@@ -105,7 +105,7 @@ export default class CombatTracker5e extends (foundry.applications?.sidebar?.tab
         });
       }
 
-      const name = combatants[0].token?.baseActor.prototypeToken.name ?? combatants[0].name;
+      const name = combatants[0].token?.name ?? combatants[0].name;
       const img = children[0].querySelector("img");
       groupContainer.innerHTML = `
         <div class="group-header flexrow">


arbron:
I'd want to go with a more sophisticated heuristic here:
1) If all token names are the same, use that name
2) If any token contains the name of the actor, use that (e.g. tokens are "Goblin (1)", "Goblin (2)", etc. and the actor is just "Goblin", use that for the group name)
3) Use the name of the first member of the group

MaxPat931:
So something like this?
```js
function getGroupName(combatants) {
  const tokenNames = combatants.map(c => c.token?.name ?? c.name);
  const actorName = combatants[0].token?.baseActor.prototypeToken.name ?? combatants[0].name;

  // 1. If all token names are the same, use that name
  const allSame = tokenNames.every(name => name === tokenNames[0]);
  if (allSame) return tokenNames[0];

  // 2. If all token names include the actor name, use actor name
  const allContainActor = tokenNames.every(name => name.includes(actorName));
  if (allContainActor) return actorName;

  // 3. Fallback to first token name
  return tokenNames[0];
}

const name = getGroupName(combatants);
```
image

arbron:
> So something like this?

Yeah, I think that will work well to get the best name in the most cases. Just add some spaces inside those `if` conditions and I'll give it a review :).

module/applications/combat/combat-tracker.mjs
```diff
@@ -105,7 +105,17 @@ export default class CombatTracker5e extends (foundry.applications?.sidebar?.tab
...
-      const name = combatants[0].token?.name ?? combatants[0].name;
+      function getGroupName(combatants) {
+            const tokenNames = combatants.map(c => c.token?.name ?? c.name);
+            const actorName = combatants[0].token?.baseActor.prototypeToken.name ?? combatants[0].name;
+            const allSame = tokenNames.every(name => name === tokenNames[0]);
+            if ( allSame ) return tokenNames[0];
+            const allContainActor = tokenNames.every(name => name.includes(actorName));
+            if ( allContainActor ) return actorName;
+            return tokenNames[0];
+      }
```
arbron, line 116:
```suggestion
      const getGroupName = combatants => {
        const tokenNames = combatants.map(c => c.token?.name ?? c.name);
        const actorName = combatants[0].token?.baseActor.prototypeToken.name ?? combatants[0].name;
        if ( tokenNames.every(name => name === tokenNames[0]) ) return tokenNames[0];
        if ( tokenNames.every(name => name.includes(actorName)) ) return actorName;
        return tokenNames[0];
      };
```

arbron (changes requested):

arbron (approved):

Fyorl (approved):

PR #5504, opened 2025-04-18
[#5503] Add `dnd5e.postBuildRollConfig` hook for skipped rolls

### LINKED ISSUE #550: MARTIAL ARCHETYPES (Battle Master)
Originally in GitLab by @ThoGri

Where are no Combat Superiority in the classfeature.

Do you have a plan when this is implemented?

### LINKED ISSUE #5503: `dnd5e.buildRollConfig` hook not called for fast-forwarded rolls
The `dnd5e.buildRollConfig` hook is called by the roll dialog when configuring a roll, but not when the config is built for fast-forwarded rolls. This prevents certain configuration from being performed for rolls like skill & tool checks where the full roll config isn't ready when the `dnd5e.preRoll` hooks are called.

The only issue is that the `buildRollConfig` hook requires the roll config application, which isn't present when fast-forwarding, so perhaps a new hook will be required.

arbron:
Makes three changes to the `dnd5e.buildRollConfig` hook to make it more useful:
- Moves it after `options.buildConfig` is called so it has the fully-built data to work on
- Calls multiple hooks using `hookNames` to dial down on specific roll types
- Calls it in `buildConfigure` if the dialog is skipped to give a chance to modify the config

This can be potentially a breaking change in two ways:
- Any callers relying on the hook being called before `buildConfig` is called
- Any callers expecting the full application to be passed to the hook might get issues with fast-forwarded rolls. A stub version of the application is passed that has all of the key public properties available, but if the caller is using the hook to modify the rendered application itself this may cause an issue.

Closes #5503

### DIFF
--- a/module/applications/dice/roll-configuration-dialog.mjs
+++ b/module/applications/dice/roll-configuration-dialog.mjs
@@ -304,17 +304,6 @@ export default class RollConfigurationDialog extends Dialog5e {
   _buildConfig(config, formData, index) {
     config = foundry.utils.mergeObject({ parts: [], data: {}, options: {} }, config);
 
-    /**
-     * A hook event that fires when a roll config is built using the roll prompt.
-     * @function dnd5e.buildRollConfig
-     * @memberof hookEvents
-     * @param {RollConfigurationDialog} app    Roll configuration dialog.
-     * @param {BasicRollConfiguration} config  Roll configuration data.
-     * @param {FormDataExtended} [formData]    Any data entered into the rolling prompt.
-     * @param {number} index                   Index of the roll within all rolls being prepared.
-     */
-    Hooks.callAll("dnd5e.buildRollConfig", this, config, formData, index);
-
     const situational = formData?.get(`roll.${index}.situational`);
     if ( situational && (config.situational !== false) ) {
       config.parts.push("@situational");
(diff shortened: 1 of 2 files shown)


module/applications/dice/roll-configuration-dialog.mjs
```diff
@@ -304,17 +304,6 @@ export default class RollConfigurationDialog extends Dialog5e {
...
-    /**
-     * A hook event that fires when a roll config is built using the roll prompt.
-     * @function dnd5e.buildRollConfig
-     * @memberof hookEvents
-     * @param {RollConfigurationDialog} app    Roll configuration dialog.
-     * @param {BasicRollConfiguration} config  Roll configuration data.
-     * @param {FormDataExtended} [formData]    Any data entered into the rolling prompt.
-     * @param {number} index                   Index of the roll within all rolls being prepared.
-     */
-    Hooks.callAll("dnd5e.buildRollConfig", this, config, formData, index);
```
Fyorl, line 316: Let's make the change to add `hookName.capitalize()` here.

module/applications/dice/roll-configuration-dialog.mjs
```diff
@@ -325,6 +314,21 @@ export default class RollConfigurationDialog extends Dialog5e {
...
+     * `dnd5e.buildRollConfig`).
+     * @function dnd5e.buildRollConfig
+     * @memberof hookEvents
+     * @param {RollConfigurationDialog|object} app  Roll configuration dialog or object that simulates it.
+     * @param {BasicRollConfiguration} config       Roll configuration data.
+     * @param {FormDataExtended} [formData]         Any data entered into the rolling prompt.
+     * @param {number} index                        Index of the roll within all rolls being prepared.
+     */
+    for ( const hookName of this.#config.hookNames ?? [""] ) {
+      Hooks.callAll(`dnd5e.build${hookName.capitalize()}RollConfig`, this, config, formData, index);
```
Fyorl, line 329: Let's add a new hook here instead called something like `postRollConfig`, with a signature of `config, index, options`, where `options` has `app` and `formData`.

module/dice/basic-roll.mjs
```diff
@@ -152,8 +152,12 @@ export default class BasicRoll extends Roll {
 
     let rolls;
     if ( dialog.configure === false ) {
+      const appStub = { config, message, rolls: config.rolls, rollType: this.constructor };
       rolls = config.rolls?.map((r, index) => {
         dialog.options?.buildConfig?.(config, r, null, index);
+        for ( const hookName of config.hookNames ?? [""] ) {
+          Hooks.callAll(`dnd5e.build${hookName.capitalize()}RollConfig`, appStub, r, null, index);
```
Fyorl, line 159: Let's call `postRollConfig` here also with blank options.

Fyorl (changes requested):

Fyorl (approved):

Issue #5382, opened 2025-03-17
Long Rest does not automatically remove a level of Exhaustion


Norc:
I'm not sure if this is a bug or a missing feature, but when you have at least one level of exhaustion and take a long rest, you have the same amount of exhaustion afterwards. Per the Exhaustion rules, taking a Long Rest should remove one level of exhaustion.

_Steps:_
1. In a fresh world with the latest D&D 4.3.x code, create a new PC
2. Add two levels of exhaustion by clicking the dots below its portrait in the character sheet
3. Take a Long Rest by clicking the "tent" icon in the character sheet

_Current results:_
The character still has two levels of exhaustion

_Expected:_
The character has one level of exhaustion

arbron:
Will need to be gated behind the rules version, because in the 2014 rules you only lose a level of exhaustion if you long rest with sufficient food and water, so potentially it wouldn't happen.

Norc:
Interesting point! I had forgotten entirely about that in the 2014 version because this has... literally never happened in one of my D&D 5E games.

krbz999:
The actual rule is "Exhaustion caused by lack of food or water can't be removed until the character eats and drinks the full required amount," which seems to indicate we need some sort of "Well Fed" flag - or use the Malnutrition status (probably not, since that's 2024 only?)

PR #5514, opened 2025-04-20
[#5382] Reduce exhaustion when resting

krbz999:
- ~~Migrates `attributes.exhaustion` to an object with `value` and `delta` (null).~~
- Adds `exhaustionDelta` to `DND5E.restTypes`.
- Reduces levels of exhaustion on a long rest by an amount equal to ~~`delta` or~~ `exhaustionDelta` ~~(in that order).~~
- ~~`delta` isn't in the schema or shown anywhere, can only be modified by AEs.~~
- ~~Migrates `ActiveEffect#changes` that have `system.attributes.exhaustion` to append `.value` (unsure why anyone would do this).~~
- The Malnutrition status causes the `malnourished` condition effect, which prevents exhaustion reduction on rests entirely (which should be in accordance with both '14 and '24 rules).

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2977,6 +2977,7 @@ DND5E.hitDieTypes = ["d4", "d6", "d8", "d10", "d12"];
  * @property {string} icon                          Icon representing this rest type. Can be either a set of FontAwesome
  *                                                  classes or an image path.
  * @property {string[]} [activationPeriods]         Activation types that should be displayed in the chat card.
+ * @property {number} [exhaustionDelta]             The number of levels of exhaustion to deduct during this rest.
  * @property {boolean} [recoverHitDice]             Should hit dice be recovered during this rest?
  * @property {boolean} [recoverHitPoints]           Should hit points be recovered during this rest?
  * @property {string[]} [recoverPeriods]            What recovery periods should be applied when this rest is taken. The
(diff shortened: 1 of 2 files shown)


module/config.mjs
```diff
@@ -3892,11 +3894,12 @@ preLocalize("conditionTypes", { key: "label", sort: true });
...
-  noMovement: new Set(["exhaustion-5", "grappled", "paralyzed", "petrified", "restrained", "unconscious"]),
-  halfMovement: new Set(["exhaustion-2"]),
   crawl: new Set(["prone", "exceedingCarryingCapacity"]),
-  petrification: new Set(["petrified"]),
-  halfHealth: new Set(["exhaustion-4"])
+  halfHealth: new Set(["exhaustion-4"]),
+  halfMovement: new Set(["exhaustion-2"]),
+  malnourished: new Set(["malnutrition"]),
+  noMovement: new Set(["exhaustion-5", "grappled", "paralyzed", "petrified", "restrained", "unconscious"]),
+  petrification: new Set(["petrified"])
```
Fyorl: We've clearly made a mistake here by calling the status effect `malnutrition` instead of `malnourished` to match the other effects. And now for these `conditionEffects` we have have `malnourished` -> `malnutrition` and `petrification` -> `petrified`.

I think ideally we would have `malnourished` -> `malnourished` and `petrified` -> `petrified`, but it hardly seems worth all the associated churn to make that happen now.

Anyway, you can do what you want with the above (including ignore it), no mandatory change required here.

Fyorl (changes requested):
Is the migration for `attributes.exhaustion` required to address the issue of reducing it by 1 on a long rest? I would expect not. What is the benefit of the `delta` property? Can it be achieved with some rest configuration options instead?

Fyorl (approved):

Issue #4803, opened 2024-11-28
Destructuring in call of dnd5e.rollDeathSaveV2 hook no longer allows changing death save behaviours


malde:
The recent changes that happened with the introduction of the new hooks in a647df1 broke the ability to change some death save behaviours from outside or via modules. I'm trying to (and successfully did up to now) hide or disable the "_character is now stable with 3 successful saves_" and "_character dies because of 3 death save failures_" messages, as well as keeping the final state of the death saves (e.g. '3 success / 2 failure'). As far as I can see, this is no longer possible with the new hooks.
I'll try to explain with some small examples what my understanding of it is. I'm not sure if what I was doing was ever intended or not, but at least it worked.

While the old hook, which passes the actual object worked (and still does):

```javascript
if ( Hooks.call("dnd5e.rollDeathSave", this, roll, details) === false ) return roll;
```

This allowed to disable some chat messages and change update behaviours:

```javascript
Hooks.on("dnd5e.rollDeathSave", (actor, roll, details) => {
   if (details.chatString === "DND5E.DeathSaveSuccess") {
     details.chatString = undefined;
   }
});
```

The V2 variant hook no longer allows doing so, because the call does pass the original values as a destructured object:

```javascript
if ( Hooks.call("dnd5e.rollDeathSaveV2", rolls, { ...details, subject: this }) === false ) return returnValue;
```

So changing the values here does nothing and it is no longer possible to change the update behaviour and stop the chat messages from being sent:

```javascript
Hooks.on("dnd5e.rollDeathSaveV2", (rolls, data) => {
   if (data.chatString === "DND5E.DeathSaveSuccess") {
     data.chatString = undefined; 
   }
});
```

Is there an obvious solution to it I'm not seeing, or can this be considered an unwanted change or even bug?

malde:
Since the original change came from you @arbron, I'll try tagging you to get some attention (sorry) or at least acknowledgement. Feel free to excuse yourself from the ticket.
This will however break my module once the deprecation is removed and thus I'd at least like to stay ontop of it.

As a fallback I could fish for strings in the more generic chatmessage hooks, but I'd rather do it on the (in my opinion) correct place.
The documentation of the hook still shows

```
* @param {string} data.chatString  Localizable string displayed in the create chat message. If not set, then
*                                  no chat message will be displayed.
```

but setting it to undefined has no effect, as it's not written back to the original object.

[...]

PR #5519, opened 2025-04-22
Change rollDeathSaveV2 hook signature to allow for mutablility

malde:
The destructuring happening in the signature of the rollDeathSaveV2 hook makes the passed data/details object immutable. Thus, the following lines are not accessible for customization of either the data.chatString or data.updates params that are described in the inline documentation for the hook. This passes the actual objects als mutable parameters so the hook can actually influence the following lines.

This attempts to fix issue #4803


arbron:
You shouldn't move the subject out of details because that would be a breaking change, instead you should just be able to stick the subject into the `details` object before passing it to the hook as-is.

malde:
Yeah, thank you, that makes way more sense. Didn't see that.
Changed it.

arbron (approved):

malde:
Anything I can do to help get this on the 5.0.x branch for it to land in 5.0.1?
I've already updated the branch, but the recent force pushed to 5.0.x kind of bloat this PR up. Should I recreate the fix on a clean branch or something, after all its very little code?

malde:
@arbron can we get this assigned to the 5.0.2 milestone or merged in? Otherwise please advise on my previous comment or what I should or can do to help with this.

arbron:
> Anything I can do to help get this on the 5.0.x branch for it to land in 5.0.1? I've already updated the branch, but the recent forced pushes to 5.0.x kind of bloat this PR up. Should I recreate the fix on a clean branch or something, after all its very little code?

You should be able to `rebase` onto the current `5.0.x` branch and skip over all but your own commits to clean it up.

malde:
Ah great, thats much cleaner.

PR #5525, opened 2025-04-22
[#5522] Added a hook to BaseActorSheet#_filterItem and ContainerSheet#_filterItem

### LINKED ISSUE #5522: Add hook invocation in _filterItem so that 3rd parties can augment filtering logic

JPMeehan:
Small & simple hook; I double checked `Hooks.call` and it will default to returning `true`, so there's no good way to have a "return true if this filter is an always show" kind of deal. But this at least generalizes the implementation for stuff like Power Specializations or any other novel additions.

Closes #5522

### DIFF
--- a/module/applications/actor/api/base-actor-sheet.mjs
+++ b/module/applications/actor/api/base-actor-sheet.mjs
@@ -1911,7 +1911,18 @@ export default class BaseActorSheet extends PrimarySheetMixin(
    * @returns {boolean|void}
    * @protected
    */
-  _filterItem(item, filters) {}
+  _filterItem(item, filters) {
+    /**
+     * A hook event that fires when an Actor sheet filters an item
+     * @function dnd5e.filterItem
+     * @memberof hookEvents
+     * @param {BaseActorSheet} actorSheet The actor sheet.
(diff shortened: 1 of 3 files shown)


JPMeehan:
Should I try to also replicate this in the Container sheet?

Fyorl:
> Should I try to also replicate this in the Container sheet?

Yes please

JPMeehan:
> > Should I try to also replicate this in the Container sheet?
> 
> Yes please

Should I keep the hook as `dnd5e.filterItem` or should I have two separate hook names?

Fyorl:
> Should I keep the hook as `dnd5e.filterItem` or should I have two separate hook names?

One hook name for now, unless it becomes obvious that one might need different filtering logic based on where the inventory is embedded. Since the first param is the application instance, it should be pretty easy to discern that context anyway though.

arbron:
@JPMeehan Want to throw a comment on the original issue so I can assign it to you?

JPMeehan:
> Now that base `_filterItem` does something, we need a `super` call in `CharacterActorSheet`, otherwise this hook won't fire on character sheets.

Done

Fyorl (approved):

PR #5529, opened 2025-04-22
[#5508] Fix transforming unlinked actors in V13

### LINKED ISSUE #5508: [5.0.x] Unlinked tokens cannot be restored after transformation
foundry: 13.340
dnd5e: 5.0.0 pre-release

Unlinked tokens do not have a 'restore transformation' button in the character sheet dropdowns after they've been transformed.

To reproduce:
- Drag an unlinked NPC onto a scene to create a token and open its character sheet from the token
- Drag another token onto that character sheet and use any transform option (I've tested with all the presets)
- Note there is no way to restore the transformation
If you then switch the source actor to linked and repeat this with a new token, the 'Restore Transformation' button appears in the sheet menu dialog as expected.

arbron:
Applies the unlinked actor changes as an updated directly to the synthetic actor rather than to the `ActorDelta` to avoid an issue with unlinked actors not immediately showing delta changes in V13.

See https://github.com/foundryvtt/foundryvtt/issues/12611

Closes #5508

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -3049,13 +3049,13 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     if ( this.isToken ) {
       const tokenData = d.prototypeToken;
       delete d.prototypeToken;
-      tokenData.delta = d;
       tokenData.elevation = this.token.elevation;
       tokenData.rotation = this.token.rotation;
       const previousActorData = this.token.delta.toObject();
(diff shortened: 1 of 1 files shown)


Fyorl (approved):
https://github.com/foundryvtt/foundryvtt/issues/12611 should be fixed in the next release, so this change is perhaps not necessary, however updating the synthetic actor directly is generally recommended, so we probably want to keep the change regardless.

I added some additional code that should resolve #5530, but in v13 only. Due to various ActorDelta bugs still lingering in v12, I couldn't see a way to resolve this issue in v12.

PR #5536, opened 2025-04-24
SRD 5.1: Poisons

krbz999:
- Fixed 'Basic Poison' being a potion and added enrichers and subtype.
- Added missing subtype for Potion of Poison, and added enrichers, replacing the old Secret block that no longer made sense.
- Added convenience "Poisoned" enricher and save enricher to 'Truth Serum' and a subtype.

image


packs/_source/items/poison/potion-of-poison.yml
```diff
@@ -117,19 +117,7 @@ system:
         override: false
       damage:
         onSave: half
-        parts:
```
Fyorl, line 120: Agreed this damage shouldn't be connected to the save, but it should still have a damage activity.

Fyorl (changes requested):

PR #5540, opened 2025-04-24
Fix bonus field labels in special traits tab in v13.

Fyorl:
Let me know if you can think of any issues this might cause.


arbron (approved):
Don't think there will be any issues.

PR #5560, opened 2025-04-29
Resolve some deprecation warnings for moved implementations

arbron:
- Resolved `renderTemplate` warnings
- Resolved `TextEditor` warnings
- Resolved `TooltipManager` warnings
- Suppressed `AppV1` warnings for remaining legacy applications
- Renamed `_onClickEntryName` to `_onClickEntry` in item directory

### DIFF
--- a/module/applications/activity/order-usage-dialog.mjs
+++ b/module/applications/activity/order-usage-dialog.mjs
@@ -292,7 +292,7 @@ export default class OrderUsageDialog extends ActivityUsageDialog {
    * @protected
    */
   _onDrop(event) {
-    const data = TextEditor.getDragEventData(event);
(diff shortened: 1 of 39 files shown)


krbz999:
May have overlooked it but am not seeing a fix for deprecation of `globalThis.DocumentSheetConfig` - we also shouldn't need to unregister any default sheets anymore.

arbron:
> May have overlooked it but am not seeing a fix for deprecation of `globalThis.DocumentSheetConfig` - we also shouldn't need to unregister any default sheets anymore.

We fixed `DocumentSheetConfig` earlier (though it hasn't been fixed in the premium modules, so maybe that is what you are seeing).

Issue #5589, opened 2025-05-01
[5.0.x] getChatLogEntryContext hook replaced


vszalai:
In foundry v13 the "getChatLogEntryContext" has been replaced with "getChatMessageContextOptions", thus the addChatMessageContextOptions never gets added to damage roll chat messages when right clicked.
I am going to submit a fix PR soon for this.

PR #5590, opened 2025-05-01
[#5589] Changed to getChatMessageContextOptions hook and changed addChatMessageContextOptions function accordingly

vszalai:
When right clicking a damage type dice roll message it now shows the proper context menu.

Fixes issue #5589

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -696,8 +696,8 @@ export default class ChatMessage5e extends ChatMessage {
    * @returns {object[]}          The extended options Array including new context choices
    */
   static addChatMessageContextOptions(html, options) {
-    const canApply = (li) => game.messages.get(li.dataset.messageId)?.canApplyDamage;
-    const canTarget = (li) => game.messages.get(li.dataset.messageId)?.canSelectTargets;
+    const canApply = li => game.messages.get(li.dataset.messageId)?.canApplyDamage;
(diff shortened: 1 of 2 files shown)


module/documents/chat-message.mjs
```diff
@@ -696,42 +696,42 @@ export default class ChatMessage5e extends ChatMessage {
    * @returns {object[]}          The extended options Array including new context choices
    */
   static addChatMessageContextOptions(html, options) {
-    const canApply = ([li]) => game.messages.get(li.dataset.messageId)?.canApplyDamage;
-    const canTarget = ([li]) => game.messages.get(li.dataset.messageId)?.canSelectTargets;
+    const canApply = (li) => game.messages.get(li.dataset.messageId)?.canApplyDamage;
+    const canTarget = (li) => game.messages.get(li.dataset.messageId)?.canSelectTargets;
```
arbron, line 700:
```suggestion
    const canApply = li => game.messages.get(li.dataset.messageId)?.canApplyDamage;
    const canTarget = li => game.messages.get(li.dataset.messageId)?.canSelectTargets;
```

arbron (changes requested):
You'll also need to fix the "Select Hit Targets" and "Select Missed Targets" options below these because they are still using the jQuery style.

arbron (approved):

PR #5596, opened 2025-05-02
[#5309, #5582] Fix damage applied via enchantment

### LINKED ISSUE #5309: Enchantment to modify damage only for some activities doesn't currently work
Ex `activities[attack].damage.parts | Add | { "number": 2, "denomination": 6, "types": ["necrotic"] }` or similar. 
This definitely worked at _some point_ in 4.x, but not currently. Currently `system.damage.parts` is a solid workaround, but unfortunately modifies _all_ activities which deal damage (including heal activities, actually).

### LINKED ISSUE #5582: Enchantments adding damage parts bugging out weapons
Foundry version 13.341
dnd5e version 5.0.1
0 modules enabled
fresh world

Error message included.

Steps
- create a character
- create a weapon on said character
- give the weapon some base damage
- create a feature on said character
- give the feature an `enchant` activity
- enchantment does: `system.damage.parts | add | [["1d4","fire"]]`
- apply enchantment to the weapon
- enjoy

As a side note: This breaks much of the inventory. Items with limited uses don't display properly, weapons don't display their damage, activities on completely unrealated items can't be displayed anymore. Applying the enchantment to other types of items (I tested a feature) breaks that item, but not the entire inventory.

Changing the enchantment to `system.base.damage.parts` doesn't break, but also doesn't add the damage.

(Tested in V12/4.4.1: working as intended)

Error message when applying the enchantment:
Image

Error message when using the enchanted weapon:
Image

arbron:
Applies a patch to `_castChangeDelta` in `EmbeddedDataField` to properly initialize the data model when being applied through an advancement.

Also fixes the activity sheet to ensure it properly locks all damage parts that were applied by advancement.

Closes #5309
Closes #5582

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -131,7 +131,6 @@ Hooks.once("init", function() {
   CONFIG.ChatMessage.dataModels = dataModels.chatMessage.config;
   CONFIG.Item.dataModels = dataModels.item.config;
   CONFIG.JournalEntryPage.dataModels = dataModels.journal.config;
-  documents.ActiveEffect5e.patchEmbeddedDataField();
 
   // Add fonts
   _configureFonts();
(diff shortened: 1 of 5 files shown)


Fyorl (changes requested):
Do we need to globally patch all `EmbeddedDataField`s? Would adding a `_castChangeDelta` override to `DamageField` not be enough?

arbron:
We could do that, we'd have to fix each place we use `EmbeddedDataField` that might be targeted by AEs (probably just the consumption targets also).

Fyorl (approved):

PR #5615, opened 2025-05-05
[#5601] Change item to entry in _onRollRecharge

### LINKED ISSUE #5601: Item is not defined
In DnD v5.0.1, use an item that has a recharge (e.g., "Fire Breath" of an Adult Red Dragon from SRD 5.2). After the item consumes its charge, try to recharge it from the Actor page to see if it recharges. It won’t work, and an error will appear in the console.

Image

gludington:
closes #5601

### DIFF
--- a/module/applications/components/inventory.mjs
+++ b/module/applications/components/inventory.mjs
@@ -626,7 +626,7 @@ export default class InventoryElement extends HTMLElement {
    * @protected
    */
   _onRollRecharge(entry, { event }={}) {
-    if ( entry instanceof Item5e ) return item.system.uses?.rollRecharge({ apply: true, event });
+    if ( entry instanceof Item5e ) return entry.system?.uses?.rollRecharge({ apply: true, event });
     return entry.uses?.rollRecharge({ apply: true, event });
   }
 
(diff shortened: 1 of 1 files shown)


module/applications/components/inventory.mjs
```diff
@@ -626,7 +626,7 @@ export default class InventoryElement extends HTMLElement {
    * @protected
    */
   _onRollRecharge(entry, { event }={}) {
-    if ( entry instanceof Item5e ) return item.system.uses?.rollRecharge({ apply: true, event });
+    if ( entry instanceof Item5e ) return entry.system?.uses?.rollRecharge({ apply: true, event });
```
krbz999, line 629: If `entry` is an instance of `Item5e`, we can be certain it has a `system` property.

```suggestion
    if ( entry instanceof Item5e ) return entry.system.uses?.rollRecharge({ apply: true, event });
```

arbron:
@gludington Can you give a comment on the original issue so I can assign it to you?

arbron (approved):

gludington:
Apologies…day job prohibited me from seeing this until now.  Thank you for addressing.

Issue #3230, opened 2024-03-10
Feature Request: prepareData hooks


NeilWhite:
A number of D&D Feats and Class Features involve tweaking proficiencies. Some are handled with built in flags (ie: jackOfAllTrades, diamondSoul).

adding a pair of hooks `dnd5e.prePrepare{DocumentType}Data` and `dnd5e.prepare{DocumentType}Data` would allow for modules to easily implement other feats/class features.

Proposal is all document types, but actor may be sufficient in practice.

Examples
- toolExpertise (artificier lvl6, and rune-knight fire-rune passive): You have expertise in any tool check you have proficiency for.

Fyorl:
Document `prepareData` is too performance-sensitive to fire hooks.

PR #5617, opened 2025-05-05
Tool Expertise

Dewi-Jones:
Introduces a flag `toolExpertise` to automate the tool expertise from artificer and fire rune, replicating the behavior of the existing `jackofalltrades` flag. Add's a new function `calculateToolProficiency` and points `prepareTools` to this function.

closes one part of https://github.com/foundryvtt/dnd5e/issues/4644 and https://github.com/foundryvtt/dnd5e/issues/3230

### DIFF
--- a/module/data/actor/templates/common.mjs
+++ b/module/data/actor/templates/common.mjs
@@ -205,6 +205,13 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
   calculateToolProficiency(multiplier, ability) {
     let roundDown = true;
     if ( multiplier == 1 && this.parent.flags.dnd5e?.toolExpertise) multiplier = 2;
+	if ( multiplier < 1 ) {
+      if ( this.parent._isRemarkableAthlete(ability) ) {
+        multiplier = .5;
+        roundDown = false;
+      }
+      else if ( this.parent.flags.dnd5e?.jackOfAllTrades ) multiplier = .5;
+    }
     return new Proficiency(this.attributes.prof, multiplier, roundDown);
   }
 }
--- a/module/data/actor/templates/common.mjs
+++ b/module/data/actor/templates/common.mjs
@@ -198,4 +198,10 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
     }
     return new Proficiency(this.attributes.prof, multiplier, roundDown);
   }
+  
+  calculateToolProficiency(multiplier, ability) {
+    let roundDown = true;
+    if ( multiplier == 1 && this.parent.flags.dnd5e?.toolExpertise) multiplier = 2;
+    return new Proficiency(this.attributes.prof, multiplier, roundDown);
+  }
 }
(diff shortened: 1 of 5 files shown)


module/data/actor/templates/common.mjs
```diff
@@ -198,4 +198,10 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
     }
     return new Proficiency(this.attributes.prof, multiplier, roundDown);
   }
+  
+  calculateToolProficiency(multiplier, ability) {
```
Fyorl, line 205:
```suggestion

  /* -------------------------------------------- */

  calculateToolProficiency(multiplier, ability) {
```

module/data/actor/templates/common.mjs
```diff
@@ -198,4 +198,10 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
...
+    let roundDown = true;
+    if ( multiplier == 1 && this.parent.flags.dnd5e?.toolExpertise) multiplier = 2;
+    return new Proficiency(this.attributes.prof, multiplier, roundDown);
```
Fyorl, line 215: As I understand it, a character would still benefit from the 2014 Jack of All Trades and Remarkable Athlete features if they are not proficient in the ability used with the tool check, but this logic is missing here.

Fyorl (changes requested):
`calculateAbilityCheckProficiency` is still called for tools as part of `Actor5e#_buildSkillToolConfig` so I would expect the roll result will be wrong if you roll a tool check with this flag set.

module/data/actor/templates/common.mjs
Dewi-Jones: I mistakenly thought they are not supposed to apply. After reading the rules carefully again I changed my mind on this. Will add the logic.

Dewi-Jones: I took a look again at how these two traits are handled right now with regards to tools and don't think any logic needs to be added here.

These bonuses are added at the time of rolling, not here. Tools are not even part of `rollData` if they don't have at least a proficiency value of `0.5`. I see two cases here:

Tool isn't part of `rollData` --> this function isn't called on it
Tool is part of `rollData` and therefore has a proficiency value of at least `0.5` --> Jack of All Trades and 2014 Remarkable Athlete don't do anything

The only exception I can find, is if an actor has an effect, downgrading the tool's proficiency to 0.
Am I missing something?

Dewi-Jones: Apologies, I missed the fact, that `_buildSkillToolConfig` also points to this and therefore Jack of All Trades and Remarkable Athlete need to be taken into account here.

Fyorl:
Manually merged here https://github.com/foundryvtt/dnd5e/commit/0889435449a86ec99eb0d4395437fdecd8ba32c9

You can avoid requiring a manual merge in future by making sure you raise the PR from a new branch on your fork, rather than an existing branch of the repo.

PR #5632, opened 2025-05-07
Bring ActiveEffect5e#isSuppressed in line with core expectations

roth-michael:
`ActiveEffect#isSuppressed` has existed in core as a getter for a little while now. With V13, systems which extend `ActiveEffect` should reference its `isSuppressed`, as now `ActiveEffect#isSuppressed` checks any effect subtypes' `system.isSuppressed` implementation. This PR gets rid of the now-unnecessary `ActiveEffect5e#determineSuppression` in favor of making `ActiveEffect5e#isSuppressed` a getter, and ensures the getter respects the super implementation.

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -77,7 +77,12 @@ export default class ActiveEffect5e extends ActiveEffect {
    * Is this active effect currently suppressed?
    * @type {boolean}
    */
-  isSuppressed = false;
+  get isSuppressed() {
+    if ( super.isSuppressed ) return true;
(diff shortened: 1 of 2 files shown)


module/documents/active-effect.mjs
```diff
@@ -77,7 +77,12 @@ export default class ActiveEffect5e extends ActiveEffect {
    * Is this active effect currently suppressed?
    * @type {boolean}
    */
-  isSuppressed = false;
+  get isSuppressed() {
```
krbz999, line 80: can probably replace the annotation with inheritDoc

Issue #5633, opened 2025-05-07
Add interface for displaying date and time based on current world calendar


arbron:
With core's addition of a calendar system it would be great to have a way to display the world date and time in the interface. This would also give a location for controls for advancing time, which currently only occurs by default during combat and while resting with a group actor.

PR #5634, opened 2025-05-07
[#5633] Add standard calendar HUD

arbron:
Introduces a new calendar HUD UI element that displays at the top of the screen and is styled like the rest of the core UI. This element includes buttons for advancing and reversing time, opening the player's assigned character sheet, and opening the party group sheet. Between that it displays the current date and time along with a day visualizer to indicate the progress of the day.

Calendar HUD GM Dark

Calendar HUD Player Light

The calendar interface is set by a `CONFIG.DND5E.calendar` object that defines the application used to render the calendar, the current HUD instance if rendered, a list of formatters that can be used to display the date and time, and a pair of methods for tracking the progress through the day & night. This makes it easy to completely disable the native HUD (setting `application` to `null`) or to customize it to whatever calendar is being used.

Calendar Settings GM

The settings offer two sets of calendar options, one visible only to GM users and the other available to all users. The GM settings allow for enabling the calendar for all users and for customizing the time change provided by each of the advance/reverse buttons.

Calendar Settings Player

The preferences have a control to toggle the display of the HUD for each user and options to customize what formatter is used to display the date or time on each side of the widget.

Calendar HUD In Context

## Todo
- [ ] Show/hide sheet buttons based on sheet changes
- [ ] Improve animation when advancing over large periods of time
- [ ] Give GMs more control over what formatters players can use
- [ ] Add UI for changing to specific date/time

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -457,8 +457,14 @@ Hooks.once("i18nInit", () => {
     });
   }
   utils.performPreLocalization(CONFIG.DND5E);
+  CONFIG.DND5E.calendar.formatters.forEach(f => {
+    f.label = game.i18n.localize(f.label);
+    f.group = game.i18n.localize(f.group);
+  });
   Object.values(CONFIG.DND5E.activityTypes).forEach(c => c.documentClass.localize());
   Object.values(CONFIG.DND5E.advancementTypes).forEach(c => c.documentClass.localize());
+  foundry.helpers.Localization.localizeDataModel(dataModels.settings.CalendarConfigSetting);
+  foundry.helpers.Localization.localizeDataModel(dataModels.settings.CalendarPreferencesSetting);
   foundry.helpers.Localization.localizeDataModel(dataModels.settings.TransformationSetting);
 });
 
@@ -489,6 +495,13 @@ Hooks.once("ready", function() {
 
   // Bastion initialization
   game.dnd5e.bastion.initializeUI();
+
+  // Display the calendar HUD
+  if ( CONFIG.DND5E.calendar.application ) {
+    CONFIG.DND5E.calendar.instance = new CONFIG.DND5E.calendar.application();
+    CONFIG.DND5E.calendar.instance.render({ force: true });
+    Hooks.on("updateWorldTime", CONFIG.DND5E.calendar.application.onUpdateWorldTime);
+  }
 
   // Determine whether a system migration is required and feasible
   if ( !game.user.isGM ) return;
--- a/lang/en.json
+++ b/lang/en.json
@@ -1000,6 +1000,77 @@
   }
 },
 
+"DND5E.CALENDAR": {
+  "Action": {
+    "AdvanceTime": "Advance by {amount}",
+    "OpenCharacterSheet": "Open Character Sheet",
+    "OpenPartySheet": "Open Party Sheet",
+    "ReverseTime": "Reverse by {amount}"
+  },
+  "Configuration": {
+    "DisabledMessage": "The DM must enable the calendar interface before it can be used by players.",
+    "Hint": "Control whether the calendar HUD is displayed and how it functions.",
+    "Label": "Configure Calendar",
+    "Name": "Calendar",
+    "Preferences": "Calendar Preferences"
+  },
+  "FIELDS": {
+    "buttons": {
+      "advanceFar": {
+        "label": "Advance Far Button"
+      },
+      "advanceShort": {
+        "label": "Advance Short Button"
+      },
+      "hint": "Amount of time changed when using the time control buttons in the HUD.",
+      "label": "Time Control Buttons",
+      "reverseFar": {
+        "label": "Reverse Far Button"
+      },
+      "reverseShort": {
+        "label": "Reverse Short Button"
+      },
+      "units": "Units",
+      "value": "Value"
+    },
+    "enabled": {
+      "hint": "Allow the calendar UI to be used by all users.",
+      "label": "Enabled"
+    },
+    "formatters": {
+      "date": {
+        "label": "Date Format"
+      },
+      "time": {
+        "label": "Time Format"
+      }
+    },
+    "visible": {
+      "label": "Display Calendar HUD"
+    }
+  },
+  "Formatters": {
+    "Date": "Date Formatters",
+    "MonthDay": {
+      "Format": "{B} {D}",
+      "Label": "Month & Day"
+    },
+    "MonthDayYear": {
+      "Format": "{B} {D}, {Y}",
+      "Label": "Month, Day, & Year"
+    },
+    "HoursMinutes": {
+      "Format": "{H}:{M}",
+      "Label": "Hours & Minutes"
+    },
+    "HoursMinutesSeconds": {
+      "Format": "{H}:{M}:{S}",
+      "Label": "Hours, Minutes, & Seconds"
+    },
+    "Time": "Time Formatters"
+  }
+},
+
 "DND5E.CAST": {
   "Title": "Cast",
   "FIELDS": {
@@ -3941,23 +4012,6 @@
   "NoneTargeted": "No Tokens Targeted",
   "Selected": "Selected",
   "Targeted": "Targeted"
-},
-"DND5E.TokenRings": {
-  "BackgroundColor": "Background Color",
-  "Effects": {
-    "Label": "Effects",
-    "BackgroundWave": "Background Wave",
-    "RingGradient": "Ring Gradient",
-    "RingPulse": "Ring Pulse"
-  },
-  "Enabled": "Use Dynamic Ring",
-  "RingColor": "Ring Color",
-  "ScaleCorrection": "Scale Correction",
-  "Subject": {
-    "Label": "Subject Path",
-    "Hint": "Explicitly specify a path for the artwork placed over the dynamic token ring. If not provided, the subject image will be set to the normal token artwork."
-  },
-  "Title": "Dynamic Ring"
 },
 "DND5E.ToggleDescription": "Toggle Description",
 
--- a/less/v2/calendar-hud.less
+++ b/less/v2/calendar-hud.less
@@ -1,0 +1,139 @@
+@property --calendar-day-progress {
+  syntax: "<number>";
+  inherits: true;
+  initial-value: 0;
+}
+@property --calendar-night-progress {
+  syntax: "<number>";
+  inherits: true;
+  initial-value: 0;
+}
+
+#calendar-hud {
+  --calendar-animation-time: 250ms;
+  --calendar-animation: var(--calendar-animation-time) ease-in-out;
+  --calendar-gap: .5rem;
+  --calendar-sun-size: 6px;
+
+  display: flex;
+  gap: var(--calendar-gap);
+  block-size: var(--control-size);
+  margin-block-start: 16px;
+  margin-inline: 16px;
+
+  align-content: center;
+  color: var(--control-icon-color);
+  font-size: var(--font-size-12);
+  pointer-events: all;
+
+  .calendar-core, .calendar-button {
+    background: var(--control-bg-color);
+    border: 1px solid var(--control-border-color);
+    border-radius: 4px;
+  }
+
+  .calendar-buttons {
+    display: flex;
+    gap: var(--calendar-gap);
+    align-items: center;
+  }
+  .calendar-button {
+    width: 24px;
+    height: 24px;
+
+    button {
+      block-size: 100%;
+      inline-size: 100%;
+      background: none;
+      border: none;
+      color: inherit;
+    }
+  }
+
+  .calendar-core {
+    display: grid;
+    grid-template-columns: 1fr 80px 1fr;
+    align-items: center;
+    gap: 6px;
+  }
+  .calendar-date, .calendar-time {
+    padding: 6px;
+    text-align: center;
+  }
+  .calendar-widget {
+    --sky-top-color: rgb(0 115 188);
+    --sky-bottom-color: rgb(161 176 209);
+
+    position: relative;
+    margin-block: 2px;
+    align-self: stretch;
+    border: 1px solid var(--control-icon-color);
+    border-radius: 2px;
+    overflow: hidden;
+    isolation: isolate;
+
+    &::after {
+      content: "";
+      position: absolute;
+      inset: 0;
+      background: linear-gradient(to bottom, var(--sky-top-color), var(--sky-bottom-color));
+      z-index: -1;
+      opacity: ~"calc(.75 + sin((var(--calendar-day-progress) * pi)))";
+      transition: opacity var(--calendar-animation);
+    }
+  }
+  .calendar-stars, .calendar-sun {
+    position: absolute;
+    block-size: 100px;
+    inline-size: 100px;
+    top: 6px;
+    left: -10px;
+    border-radius: 100%;
+    transition: rotate var(--calendar-animation);
+  }
+  .calendar-stars {
+    rotate: calc(var(--calendar-night-progress) * -.5turn);
+    z-index: -2;
+
+    &::before {
+      content: "";
+      position: absolute;
+      inset: -20px;
+      border-radius: 100%;
+      background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 256 256'%3E%3Cfilter id='filter'%3E%3CfeTurbulence baseFrequency='0.3'/%3E%3CfeColorMatrix values='0 0 0 9 -4 0 0 0 9 -4 0 0 0 9 -4 0 0 0 0 1'/%3E%3C/filter%3E%3Crect width='100%25' height='100%25' filter='url(%23filter)'/%3E%3C/svg%3E%0A");
+    }
+  }
+  .calendar-sun {
+    rotate: calc(var(--calendar-day-progress) * -.28turn);
+    opacity: ~"calc(1 + sin((var(--calendar-day-progress) * pi)))";
+    transition: opacity var(--calendar-animation), rotate var(--calendar-animation);
+
+    &::before, &::after {
+      content: "";
+      position: absolute;
+    }
+    &::before {
+      block-size: 120px;
+      inline-size: 120px;
+      inset-inline-end: -50px;
+      inset-block-start: -43px;
+      background:
+        radial-gradient(20px 10px, rgb(252 249 214 / .7), transparent),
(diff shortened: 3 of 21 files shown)


module/applications/calendar/base-calendar-hud.mjs
```diff
@@ -0,0 +1,206 @@
...
+  /* -------------------------------------------- */
+
+  /**
+   * Place the HUD application in the middle-top UI section.
+   * @protected
+   */
+  _placeHUD() {
+    const location = document.querySelector("#ui-middle #ui-top");
+    location.append(this.element);
+  }
```
krbz999, line 69: Isn't this the job of `ApplicationV2#_insertElement`?

dnd5e.mjs
```diff
@@ -490,6 +496,13 @@ Hooks.once("ready", function() {
   // Bastion initialization
   game.dnd5e.bastion.initializeUI();
 
+  // Display the calendar HUD
+  if ( CONFIG.DND5E.calendar.application ) {
+    CONFIG.DND5E.calendar.instance = new CONFIG.DND5E.calendar.application();
+    CONFIG.DND5E.calendar.instance.render({ force: true });
```
krbz999, line 502: Specifically not registering in `CONFIG.ui`?

module/applications/calendar/base-calendar-hud.mjs
```diff
@@ -0,0 +1,102 @@
...
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async render(options) {
+    if ( this.rendered || this.shouldDisplay ) await super.render(options);
+  }
```
krbz999, line 47: Add a `return this;` here just to maintain the same return value?

MotoMoto1234:
Can you add the ability to switch the config in the hud like simple calendar offers, cause I personally switch between minutes and hours numerous times per live session and not being able to otherwise would lead to having to set it to minutes and then clicking a thousand times for a long rest, or downtimes.  Plus, general workflow of a live session, hexcrawls advance by hours, while dungeon crawls typically advance in minutes.  Downtimes advance in days.  
image

arbron:
Added dropdown time controls:

Calendar HUD Controls Dropdown

Kapuzenjoe:
> Added dropdown time controls:
> 
> 
> 
> Calendar HUD Controls Dropdown
> 
> 

That looks nice, but would it be possible to make the letters for minutes and hours correspond to the standard SI units? So h and m in lowercase

arbron:
Well, they technically are, but with default dnd5e styles all button text is uppercase.

Kharmans:
Hey, this will be a REALLY good addition to the system, and it looks really good too! Great work for everybody.
It will be possible to configure moon phases and seasons with this calendar? And, of course, build a custom calendar, with my own date and time?

JPMeehan:
Usability comment - the location of the buttons changes based on the length of the date, so when you move between months with short and long names you can lose your position on the time adjustment buttons.

image


image

JPMeehan:
This does not appear to account for the yearZero property

image


Testing with the https://foundryvtt.com/packages/eberron-calendar module

Sayshal:
Will the placement of the calendar be configurable? Either by making it a floating element that can be dragged/locked or a config to distinguish location.

module/applications/calendar/base-calendar-hud.mjs
```diff
@@ -0,0 +1,106 @@
...
+    });
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Respond to changes in the calendar settings.
+   */
+  onUpdateSettings() {
+    if ( this.shouldDisplay ) this.render({ force: true });
```
krbz999, line 87: Feels like this (and the `shouldDisplay` property) can be simplified using `_canRender`.

module/applications/calendar/calendar-hud.mjs
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * Handle opening the primary party's sheet.
+   * @this {CalendarHUD}
+   * @param {Event} event         Triggering click event.
+   * @param {HTMLElement} target  Button that was clicked.
+   */
+  static #openPartySheet(event, target) {
+    game.settings.get("dnd5e", "primaryParty")?.actor?.sheet.render({ force: true });
+  }
```
krbz999, line 272:
```suggestion
  static #openPartySheet(event, target) {
    game.actors.party?.sheet.render({ force: true });
  }
```

module/data/settings/calendar-setting.mjs
```diff
@@ -0,0 +1,41 @@
...
+
+  /** @override */
+  static LOCALIZATION_PREFIXES = ["DND5E.CALENDAR"];
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    return {
+      enabled: new BooleanField({ required: true })
```
krbz999: (Redundant)
```suggestion
      enabled: new BooleanField()
```

lang/en.json
```diff
@@ -1090,6 +1090,59 @@
...
+  "Configuration": {
+    "DisabledMessage": "The DM must enable the calendar interface before it can be used by players.",
+    "Hint": "Control whether the calendar HUD is displayed and how it functions.",
+    "Label": "Configure Calendar",
+    "Name": "Calendar",
+    "Preferences": "Calendar Preferences"
+  },
+  "FIELDS": {
+    "enabled": {
+      "hint": "Allow the calendar UI to be used by all users.",
```
Fyorl: 'used' perhaps might imply that players could use the other functions of the calendar, such as advancing time. I would suggest something like this, but maybe there is a better suggestion.


```suggestion
      "hint": "Allow the calendar UI to be viewed & configured by all users.",
```

less/v2/calendar-hud.less
```diff
@@ -0,0 +1,180 @@
...
+    &:not(:hover, :focus-visible) .dropdown {
+      opacity: 0;
+      visibility: hidden;
+    }
+  }
+  .calendar-button {
+    block-size: 24px;
+    inline-size: 24px;
+
+    & .calendar-button {
```
Fyorl:
```suggestion
    .calendar-button {
```

less/v2/calendar-hud.less
```diff
@@ -0,0 +1,180 @@
...
+      background: none;
+      border: none;
+      color: inherit;
+      text-transform: none;
+    }
+  }
+
+  .calendar-core {
+    display: grid;
+    grid-template-columns: 1fr 80px 1fr;
```
Fyorl: Best give these left and right columns fixed widths as well, otherwise the whole widget changes size whenever the date or time changes. 280px seems like a good size for the whole thing.

```suggestion
    grid-template-columns: 100px 80px 100px;
```

Also, I had some styling ideas, but the calendar widget seems hard-coded to this 80x26px size. Is it possible to at least make it variable on width? If it's huge effort then don't worry.

module/applications/calendar/base-calendar-hud.mjs
```diff
@@ -0,0 +1,106 @@
...
+      && (game.settings.get("dnd5e", "calendarPreferences")?.visible || false);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Should the calendar HUD be displayed for the current user?
+   * @type {boolean}
+   */
+  get shouldDisplay() {
```
Fyorl: Use `_canRender` for this

module/applications/calendar/base-calendar-hud.mjs
```diff
@@ -0,0 +1,106 @@
...
+    return this.constructor.shouldDisplay;
+  }
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async render(options) {
+    if ( this.rendered || this.shouldDisplay ) await super.render(options);
```
Fyorl: Use `_canRender` for this.

module/applications/calendar/calendar-hud.mjs
```diff
@@ -0,0 +1,280 @@
...
+  /**
+   * Default time periods to display for controlling time.
+   * @type {{ value: number, unit: string }}
+   */
+  static TIME_CONTROL_VALUES = [
+    { value: 7, unit: "day" },
+    { value: 1, unit: "day" },
+    { value: 8, unit: "hour" },
+    { value: 30, unit: "minute" },
+    { value: 1, unit: "minute" }
```
Fyorl, line 52: Maybe standard spell duration increments: 1 minute, 10 minutes, 1 hour, 8 hours. That would mean we maybe remove 30 mins? Or is 30 mins useful? I know Monks in 2014 needed 30 mins to recover Ki, but they don't anymore. Can't think of anything else that has a 30 min duration.

Also, I know that the advance button itself advanced by 1 hour, but maybe we should add 1 hour to this list regardless.

module/applications/calendar/calendar-hud.mjs
```diff
@@ -0,0 +1,280 @@
...
+   */
+  _getCalendarButtons() {
+    const config = game.settings.get("dnd5e", "calendarConfig");
+    return [
+      {
+        action: "reverse",
+        dataset: { value: 1, unit: "hour" },
+        icon: "fa-solid fa-angles-left",
+        position: "start",
+        tooltip: game.i18n.format("DND5E.CALENDAR.Action.ReverseTime", { amount: formatTime(1, "hour").titleCase() }),
```
Fyorl: Consider allowing this amount to be configurable via `TIME_CONTROL_VALUES`.

less/v2/calendar-hud.less
```diff
@@ -0,0 +1,180 @@
...
+      button { font-size: var(--font-size-12); }
+    }
+
+    button {
+      block-size: 100%;
+      inline-size: 100%;
+      background: none;
+      border: none;
+      color: inherit;
+      text-transform: none;
```
Fyorl, line 74: Maybe slightly bigger hitboxes


```suggestion
      text-transform: none;
      padding: 6px;
```

module/applications/calendar/calendar-hud.mjs
```diff
@@ -0,0 +1,280 @@
...
+          })
+        }))
+      },
+      {
+        action: "openPartySheet",
+        icon: "fa-solid fa-users",
+        position: "end",
+        tooltip: game.i18n.localize("DND5E.CALENDAR.Action.OpenPartySheet"),
+        visible: game.settings.get("dnd5e", "primaryParty")?.actor
+          ?.testUserPermission(game.user, CONST.DOCUMENT_OWNERSHIP_LEVELS.LIMITED)
```
Fyorl:
```suggestion
        visible: game.actors.party?.testUserPermission(game.user, CONST.DOCUMENT_OWNERSHIP_LEVELS.LIMITED)
```

module/applications/calendar/calendar-hud.mjs
```diff
@@ -0,0 +1,280 @@
...
+        context.buttons = context.buttons.filter(b => b.position === "start").reverse();
+        break;
+    }
+    return context;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Adjust the date, time, and progress in the core part without a full re-render to allow animation.
```
Fyorl, line 198: Are we preventing `core` from being rendered via the normal pipeline somewhere? Otherwise I'd have thought you'd need something like this:

```js
_configureRenderOptions(options) {
  super._configureRenderOptions(options);
  if ( this.rendered ) options.parts = options.parts.filter(p => p !== "core");
}
```

module/applications/calendar/calendar-hud.mjs
```diff
@@ -0,0 +1,280 @@
...
+        button.addEventListener("click", event => {
+          event.preventDefault();
+          config.onClick(event);
+        });
+      }
+    };
+
+    for ( const button of this.element.querySelectorAll(".calendar-buttons > .calendar-button > button") ) {
+      const config = context.buttons[parseInt(button.dataset.index)];
+      addButtonEvent(button, config);
```
Fyorl: I suppose you have to do this in `_onRender` because you need `context.buttons`. The ways we handled something like this elsewhere (like journal sheets, for example), is to cache the prepared context so that it can be accessed outside of the normal render pipeline.

For example, in `_prepareButtonsContext`, you could have something like:
```js
this.#buttons = context.buttons = /* ... */
```
That would allow you to handle these button clicks in `_onClickAction` without having to bind your own listeners.

module/data/calendar/calendar-data.mjs
```diff
@@ -0,0 +1,253 @@
...
+  /* -------------------------------------------- */
+
+  /**
+   * Progress between sunrise and sunset assuming it is daylight half the day duration.
+   * @param {number|Components} [time]  The time to use, by default the current world time.
+   * @returns {number}                  Progress through day period, with 0 representing sunrise and 1 sunset.
+   */
+  progressDay(time=game.time.components) {
+    const components = typeof time === "number" ? this.timeToComponents(time) : time;
+    return (CalendarData5e.hoursOfDay(components) - this.sunrise(time)) / this.daylightHours(time);
```
Fyorl: Shouldn't be necessary to repeat this check since `hoursOfDay` also accepts `number|Components`.


```suggestion
    return (CalendarData5e.hoursOfDay(time) - this.sunrise(time)) / this.daylightHours(time);
```

module/data/calendar/calendar-data.mjs
```diff
@@ -0,0 +1,253 @@
...
+
+  /**
+   * Progress between sunset and sunrise assuming it is night half the day duration.
+   * @param {number|Components} [time]  The time to use, by default the current world time.
+   * @returns {number}                  Progress through night period, with 0 representing sunset and 1 sunrise.
+   */
+  progressNight(time=game.time.components) {
+    const components = typeof time === "number" ? this.timeToComponents(time) : time;
+    const daylightHours = this.daylightHours(time);
+    let hour = CalendarData5e.hoursOfDay(components);
```
Fyorl:
```suggestion
    const daylightHours = this.daylightHours(time);
    let hour = CalendarData5e.hoursOfDay(time);
```

module/applications/calendar/calendar-hud.mjs
```diff
@@ -0,0 +1,280 @@
...
+        icon: "fa-solid fa-user",
+        position: "start",
+        tooltip: game.i18n.localize("DND5E.CALENDAR.Action.OpenCharacterSheet"),
+        visible: !!game.user.character
+      },
+      {
+        action: "advance",
+        icon: "fa-solid fa-angles-right",
+        position: "end",
+        tooltip: game.i18n.format("DND5E.CALENDAR.Action.AdvanceTime", { amount: formatTime(1, "hour").titleCase() }),
```
Fyorl: Missing `dataset`

module/config.mjs
```diff
@@ -4414,6 +4420,46 @@ DND5E.defaultArtwork = {
...
+/*  Calendar                                    */
+/* -------------------------------------------- */
+
+/**
+ * Configuration information for the calendar UI.
+ * @type {CalendarHUDConfiguration}
+ */
+DND5E.calendar = {
+  application: CalenderHUD,
+  instance: null,
```
Fyorl: Do we intend that the actual instance itself can be swapped out? You can already configure the application class that the system instantiates. I would think that's enough. In which case, the instantiated calendar instance could go somewhere like `dnd5e.calendar.instance`.

Fyorl (changes requested):
Good integration

less/v2/calendar-hud.less
arbron: This is a bit too narrow when using a formatter with the year. I've updated it with `minmax(110px 1fr)` which allows it to expand for very long labels but keeps it reasonable otherwise.

Allowing the width to be customized would require some work to figure out some spacing stuff to get the sun to rise and set in the correct location. Maybe leave it as a TODO for now.

module/applications/calendar/calendar-hud.mjs
arbron: Literally just took the list from Ember (and added 8 hours), but we could adjust. I will add 1 hour to the dropdown.

Maybe something like 15 or 20 minutes would be useful for a "walking around a dungeon" amount of time.

module/applications/calendar/calendar-hud.mjs
arbron: Well this is just for when the time is updated, in which case the hook calls `renderCore` directly. The only other time we call render directly is first render and settings updates, both of which shouldn't care about keeping the animations intact. That said, I don't think this would hurt anything.

module/config.mjs
arbron: Swapped it to `dnd5e.ui.calendar` to match core a bit more.

JPMeehan:
Is it intended that both setting dropdowns let you pick date & time?

image

less/v2/calendar-hud.less
JPMeehan: This works a lot better with the Eberron calendar, it's not jumping around a ton any more when I go from a short month like Eyre to a longer one like Sypheros, even with max length date times with years & seconds

module/applications/calendar/calendar-hud.mjs
Fyorl: If we're unsure then I think spell durations are a sensible default. Otherwise go with whatever you think is best. We do need something between 1 min and 1 hour though.

less/v2/calendar-hud.less
```diff
@@ -0,0 +1,183 @@
...
+    .calendar-button {
+      block-size: unset;
+      inline-size: unset;
+      button { font-size: var(--font-size-12); }
+    }
+
+    button {
+      block-size: 100%;
+      inline-size: 100%;
+      padding: 4px;
```
Fyorl:
```suggestion
      padding: 6px;
```

less/v2/calendar-hud.less
```diff
@@ -0,0 +1,183 @@
...
+  }
+  .calendar-widget {
+    --sky-top-color: rgb(0 115 188);
+    --sky-bottom-color: rgb(161 176 209);
+
+    position: relative;
+    margin-block: 2px;
+    align-self: stretch;
+    border: 1px solid var(--control-icon-color);
+    border-radius: 2px;
```
Fyorl:
```suggestion
    border: 1px solid var(--dnd5e-color-gold);
    border-radius: 3px;
```

Fyorl (approved):
I think if we're leaving the calendar widget widening as a TODO then I don't have much to add, it looks good as-is. Any remaining small tweaks I have just posted as comments.

PR #5639, opened 2025-05-08
[#5612] Add ability to specify source books on a per-pack basis

### LINKED ISSUE #5612: The "source" of items have changed if they were left blank and the module that contains them registers one or more source book titles
For example, if you have a personal module, before if the source of an item was left blank it would default to the name of the module.

Now, if that module has one or more source books in `Module#flags.dnd5e`, all these blank-source items use the first book in that flag object.

This is fine for most modules that contain the contents of a single book, but not eg a personal module that contains both personal homebrew and manually entered third party content.

arbron:
Adds support for the `dnd5e.sourceBook` flag in compendium packs in the manifest. If specified, this value will be used by default when inferring the source book before falling back to a single source book defined in the `dnd5e.sourceBooks` flag on the package manifest.

This reverts the change made earlier that changed the process to infer even if more than one source book is defined by the package at the root level.

Closes #5612

### DIFF
--- a/module/data/shared/source-field.mjs
+++ b/module/data/shared/source-field.mjs
@@ -38,8 +38,9 @@ export default class SourceField extends SchemaField {
    * @param {string} uuid  Compendium source or document UUID.
    */
   static prepareData(uuid) {
-    const { pkg, flags } = SourceField.getPackage(uuid) ?? {};
-    this.bookPlaceholder = flags?.dnd5e?.sourceBook ?? SourceField.getModuleBook(pkg) ?? "";
+    const collection = foundry.utils.parseUuid(uuid)?.collection;
+    const pkg = SourceField.getPackage(collection);
+    this.bookPlaceholder = collection?.metadata?.flags?.dnd5e?.sourceBook ?? SourceField.getModuleBook(pkg) ?? "";
     if ( !this.book ) this.book = this.bookPlaceholder;
 
     if ( this.custom ) this.label = this.custom;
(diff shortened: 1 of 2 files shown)


module/data/shared/source-field.mjs
```diff
@@ -38,8 +38,8 @@ export default class SourceField extends SchemaField {
    * @param {string} uuid  Compendium source or document UUID.
    */
   static prepareData(uuid) {
-    const pkg = SourceField.getPackage(uuid);
-    this.bookPlaceholder = SourceField.getModuleBook(pkg) ?? "";
+    const { pkg, flags } = SourceField.getPackage(uuid) ?? {};
```
Fyorl, line 41: Let's keep `getPackage` returning a package only. We can `parseUuid` here instead in order to retrieve the collection flags, and then modify `getPackage` to accept a `CompendiumCollection` alternatively in order to preserve its API and avoid doing the work twice.

```js
// prepareData
const { collection } = foundry.utils.parseUuid(uuid) ?? {};
const flags = collection?.metadata?.flags;
const pkg = SourceField.getPackage(collection);

// getPackage
if ( !uuidOrCollection ) return null;
const pack = typeof uuidOrCollection === "string" ? foundry.utils.parseUuid(uuidOrCollection)?.collection?.metadata : uuidOrCollection.metadata;
```

Fyorl (changes requested):

Fyorl (approved):

PR #5640, opened 2025-05-08
[#5605] Prevent blank values in duration & range units

### LINKED ISSUE #5605: Items have the "Self" tag even when Range is empty
Image

Image

(v13, v5.0.x)

arbron:
These fields weren't allowing blank values in V12, but a change in how V13 generates fields was adding a blank option. This change to the data models will prevent blank values from being entered and fix the generated inputs.

Closes #5605

### DIFF
--- a/module/data/shared/range-field.mjs
+++ b/module/data/shared/range-field.mjs
@@ -14,7 +14,7 @@ export default class RangeField extends SchemaField {
   constructor(fields={}, options={}) {
     fields = {
       value: new FormulaField({ deterministic: true }),
-      units: new StringField(),
+      units: new StringField({ required: true, blank: false, initial: "self" }),
       special: new StringField(),
       ...fields
     };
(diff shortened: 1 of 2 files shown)


module/data/shared/range-field.mjs
```diff
@@ -14,7 +14,7 @@ export default class RangeField extends SchemaField {
   constructor(fields={}, options={}) {
     fields = {
       value: new FormulaField({ deterministic: true }),
-      units: new StringField(),
+      units: new StringField({ required: true, blank: false, initial: "self" }),
```
krbz999, line 17: Won't this start throwing errors in `_validateSpecial` if anyone already set `units` to a blank value?

Fyorl: Data on disk that already has a blank units will get cleaned to the `initial` value on world load. It would only result in a validation error if there is no valid `initial` to clean to.

Fyorl (approved):

PR #5670, opened 2025-05-13
[#5465] Change "apply damage" to always remove hp, "apply healing" to always add hp

### LINKED ISSUE #5465: Player Right-Clicking Chat Card, "Apply Healing" Actually Causes Damage
If you do some kind of healing like casting a spell or using a potion of healing. Roll for the amount. The chat card lists how much healing was rolled. Then a player right-clicks on that number and gets the contextual menu. If they select "Apply Healing," they will actually take damage by that amount. If they select "Apply Damage," then they will heal by that amount.

This is terrible.

It's been explained to me that it's "a holdover" from how things used to be, but I have to say, it sucks. Reverse the labels or disable the contextual menu. Do something other than offer this as a choice. It's the absolute worst.

I apologize for being feisty about it. Thank you for listening.

roth-michael:
Closes #5465 
Before/After behavior:

| Damage | Apply Damage (Before) | Apply Damage (After) | Apply Healing (Before) | Apply Healing (After) | Apply Half Damage (Before) | Apply Half Damage (After) |
| --- | --- | --- | --- | --- | --- | --- |
| 10[slashing] | -10hp | -10hp | +10hp | +10hp | -5hp | -5hp |
| 10[healing] | +10hp | -10hp | -10hp | +10hp | +5hp | -5hp | 
| 10[temphp] |+10temphp | nothing | nothing | +10temphp | +5temphp | nothing |
| 5[healing] + 5[slashing] | nothing | -10hp | nothing | +10hp | nothing | -5hp |

Potentially disagreeable bits:
1. Half/double "damage" now can only be used for actual damage, rather than healing
2. Instances of combined damage & healing simply sum, rather than offsetting one another.

I don't see (1) as a dealbreaker, and (2) also seems rather unlikely. A "full" solution to (2) would involve showing a different context menu for heal cards where every instance of "Damage" and "Healing" is swapped. (I have this ready to go locally if that's a preferable fix).


arbron:
I'd probably think we should use the `only` option instead, that means any healing types will be ignored when you click "Apply Damage" and any damage types will be ignored when you click "Apply Healing":

```javascript
    return Promise.all(canvas.tokens.controlled.map(t => {
      return t.actor?.applyDamage(damages, { multiplier, only: multiplier < 0 ? "healing" : "damage" });
    }));
```

roth-michael:
Works for me. Only thing then is lost functionality of half/double healing via context menu (which also was lost in my original fix, but wouldn't be in the case of changing context menu options based on activity type)

Fyorl (approved):

Issue #5631, opened 2025-05-07
PhysicalItemTemplate#_renderContainers does not respect sheet permissions


JPMeehan:
The actor sheet render call in `_renderContainers` does not check permissions first with `sheet.isVisible`, resulting in warnings on players.

![Image](https://github.com/user-attachments/assets/e2f47a88-1c5c-4637-ad5a-3e65caee4192)

Screenshot and behavior find by Rob erred on the discord https://discord.com/channels/170995199584108546/670336046164213761/1369738212842541136

> During my weekend game my players said they were seeing an insufficient permissions banner popping when NPCs made an attack or damage roll. Thought it was a module until I saw the same thing in the ember game I play in last night, and I can replicate it with no modules enabled

PR #5675, opened 2025-05-15
[#5674] Fix spurious warnings on connected clients when updating items.

### LINKED ISSUE #5674: Using an item with ammunition triggers spurious warnings on connected clients

Fyorl:
- Closes #5674

Added an early return for an item that is not in a container or was not moved out of one. Let me know if you think there are any issues with this logic.


arbron:
Does this also resolve this issue: https://github.com/foundryvtt/dnd5e/issues/5631?

Fyorl:
> Does this also resolve this issue: #5631?

Seems so, I did wonder what that issue meant.

PR #5737, opened 2025-06-03
Add support for rotation region shapes with rotate region behavior

arbron:
Rotates the primary region's shapes when the rotate region behavior is activated, and adds support for rotating other linked regions at the same time. The region shape updates are performed at the middle of the animation time along with walls, lights, & sounds.

### DIFF
--- a/module/data/region-behavior/rotate-area.mjs
+++ b/module/data/region-behavior/rotate-area.mjs
@@ -54,6 +54,9 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
       lights: new SchemaField({
         ids: new SetField(new DocumentIdField())
       }),
+      regions: new SchemaField({
+        ids: new SetField(new DocumentIdField())
+      }),
       sounds: new SchemaField({
         ids: new SetField(new DocumentIdField())
       }),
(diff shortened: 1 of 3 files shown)


module/data/region-behavior/rotate-area.mjs
```diff
@@ -173,6 +176,11 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
       tiles: Array.from(this.tiles.ids).map(t => calculateRotationUpdate(this.scene.tiles.get(t))).filter(_ => _),
       tokens: Array.from(this.region.tokens).map(({ object: token }) => calculateRotationUpdate(token)),
       lights: Array.from(this.lights.ids).map(l => calculateRotationUpdate(this.scene.lights.get(l))).filter(_ => _),
+      regions: [this.region.id, ...this.regions.ids].map(r => {
+        const region = this.scene.regions.get(r);
+        return region
+          ? RotateAreaRegionBehaviorType.#rotateRegionShapes(this.scene.regions.get(r), angle, radians, pivot) : null;
```
Fyorl:
```suggestion
          ? RotateAreaRegionBehaviorType.#rotateRegionShapes(region, angle, radians, pivot) : null;
```

Fyorl (approved):

PR #5742, opened 2025-06-05
Adjust how rotating regions are animated to eliminate delays

arbron:
Adjust how the rotate area behavior handles animations. Instead of triggering separate animations for each token and tile based on their individual update hooks, this uses the update hook for the region behavior to trigger a single animation for all elements.

Also updated how tokens are refreshed to ensure token vision is updated throughout the animation on V13.

### DIFF
--- a/module/data/region-behavior/rotate-area.mjs
+++ b/module/data/region-behavior/rotate-area.mjs
@@ -160,7 +160,7 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
     // Prepare update data for each rotated document type
     const updates = {
       tiles: Array.from(this.tiles.ids).map(t => calculateRotationUpdate(this.scene.tiles.get(t))).filter(_ => _),
-      tokens: Array.from(this.region.tokens).map(({ object: token }) => calculateRotationUpdate(token)),
+      tokens: Array.from(this.region.tokens).map(t => calculateRotationUpdate(t)),
       lights: Array.from(this.lights.ids).map(l => calculateRotationUpdate(this.scene.lights.get(l))).filter(_ => _),
       regions: [this.region.id, ...this.regions.ids].map(r => {
         const region = this.scene.regions.get(r);
@@ -239,7 +239,14 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
    * @returns {Point}
    */
   static #placeableCenter(doc) {
-    if ( doc instanceof foundry.abstract.Document ) doc = doc.object;
+    if ( doc instanceof TokenDocument ) return doc.getCenterPoint();
+    if ( doc instanceof foundry.abstract.Document ) {
+      if ( doc.object ) doc = doc.object;
+      else {
+        const size = this.#placeableSize(doc);
+        return { x: doc.x + (size.width / 2), y: doc.y + (size.height / 2) };
+      }
+    }
     if ( "center" in doc ) return doc.center;
     return { x: doc.x, y: doc.y };
   }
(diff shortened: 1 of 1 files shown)


module/data/region-behavior/rotate-area.mjs
```diff
@@ -187,11 +173,26 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
...
+    setTimeout(async () => {
+      resolve(true);
+      await this.parent.update({ "system.status.rotating": false });
+    }, duration);
+    await animationComplete;
+
+    // Update all rotated documents
+    await canvas.scene.updateEmbeddedDocuments("AmbientLight", updates.lights);
+    await canvas.scene.updateEmbeddedDocuments("AmbientSound", updates.sounds);
+    await canvas.scene.updateEmbeddedDocuments("Tile", updates.tiles);
```
Fyorl: I guess I missed this earlier, but do we have any guarantee that `canvas.scene` is the scene that this region behaviour is in? Should this be `this.parent.parent` instead?

module/data/region-behavior/rotate-area.mjs
```diff
@@ -187,11 +173,26 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
...
+      { dnd5e: { rotateArea: { angle, duration, pivot } } }
+    );
+
+    // Wait for the visible animation to complete before performing document updates
+    const { promise: animationComplete, resolve } = Promise.withResolvers();
+    setTimeout(async () => {
+      resolve(true);
+      await this.parent.update({ "system.status.rotating": false });
+    }, duration);
+    await animationComplete;
```
Fyorl: Theoretically this shouldn't require several updates, right? It should be enough to call a single `scene.update` with all of the placeable updates and some `options` passed that detail how the animation is performed. Then we trigger the animation for each client in `onUpdateScene`. Or does that not work?

Fyorl (changes requested):

module/data/region-behavior/rotate-area.mjs
arbron: That was the old method, but it resulted in different elements animating out of sync because each `onUpdate` hook is called at slightly different times when the updates are complete.

Fyorl: The old method dispatched separate updates per embedded document also. I am suggesting we dispatch a single update to the scene and listen for `onUpdateScene`.

arbron: Hmm, I tried changing it to this:

```javascript
    await this.scene.update(updates, {
      animate: false,
      movement: updates.tokens.reduce((obj, { _id }) => {
        obj[_id] = {
          constrainOptions: { ignoreWalls: true, ignoreCost: true },
          showRuler: false
        };
        return obj;
      }, {})
    });
```

The update performed fine, but once it was done the "Loading Scene…" progress bar appears as the scene re-renders.

It also has the disadvantage of not calling the `preUpdate` and `update` hooks for each individual document, instead only calling them for the scene as a whole.

Fyorl: The current implementation is something like 10 server round-trips done in serial, so it would be great if we could avoid it. I'll take a look to see if we can do something about the loading scene behaviour. If not, then it's fine to do it this way until core gives us better tools.

> It also has the disadvantage of not calling the `preUpdate` and `update` hooks for each individual document, instead only calling them for the scene as a whole.

Arguably a benefit for what we're trying to do here.

arbron: Checked out `Scene#_onUpdate` in core and it looks like there isn't any way to prevent it from a complete canvas redraw if any placeable collections are updated.

Fyorl: Alright, if we can't do a bulk operation then I don't think there's a way to improve on this. It's possible that https://github.com/foundryvtt/foundryvtt/issues/11898 might offer some incidental improvement there.

Can you try at least batching all these `updateEmbeddedDocuments` calls into a single `Promise.all` so that they execute simultaneously, assuming the order doesn't matter. They will get queued on the server, but it means that the clients do not need to wait on a full server round-trip between each call.

arbron: Updated

Fyorl (approved):

PR #5749, opened 2025-06-07
[#5129] Add dialog for sending requests to certain players

arbron:
Adds a new `PartyRequestDialog` that can be used to prompt for which characters/players should receive a request. This dialog can be invoked from a specific group, in which case it will list the actors in that group, or it can use the primary party/assigned characters.

Party Request Dialog

The dialog presents the user with a list of actors that can be checked or unchecked as well as a list of owners for each actor. This list can be pre-filtered with the provided condition.

When the dialog is submitted a new `request` chat message will be created with the selected actors and users as well as any other provided data.

### DIFF
--- a/module/applications/actor/party-request-dialog.mjs
+++ b/module/applications/actor/party-request-dialog.mjs
@@ -1,0 +1,168 @@
+import Dialog5e from "../api/dialog.mjs";
+
+/**
+ * @type PartyRequestDialogOptions
+ * @property {object}                     request
+ * @property {PartyRequestCondition|null} request.condition  Callback used to determine if an actor should be included.
+ * @property {Actor5e|null}               request.group      Group actor to fetch the actor list from, otherwise uses
+ *                                                           the primary party if one is set or falls back to the
+ *                                                           assigned characters.
+ */
+
+/**
+ * @callback PartyRequestCondition
+ * @param {Actor5e} actor  An actor that might be able to receive the request.
+ * @returns {boolean}      Should the actor be shown in the dialog?
+ */
+
+/**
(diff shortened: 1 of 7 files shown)


module/applications/actor/party-request-dialog.mjs
```diff
@@ -0,0 +1,168 @@
...
+
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  #users;
+
+  /**
+   * Mapping of selected actors and their requested users.
+   * @type {Map<Actor5e, User>}
```
Fyorl:
```suggestion
   * @type {Map<Actor5e, User5e>}
```

module/applications/actor/party-request-dialog.mjs
```diff
@@ -0,0 +1,168 @@
...
+   * @returns {Promise<ChatMessage5e>}  Promise that resolves to the created request chat message.
+   * @throws if dialog is closed
+   */
+  static async sendRequest(handler, messageData={}, options={}) {
+    const recipients = await this.getRecipients(options);
+
+    messageData = foundry.utils.mergeObject({
+      system: {
+        handler,
+        targets: Array.from(recipients.entries()).map(([actor, user]) => ({ actor: actor.id, user: user?.id }))
```
Fyorl: I think this became a `DocumentUUIDField`


```suggestion
        targets: Array.from(recipients.entries()).map(([actor, user]) => ({ actor: actor.uuid, user: user?.id }))
```

Fyorl (approved):

Issue #5694, opened 2025-05-20
Tokens Marked Secret (Disposition Setting) Cannot Be Targeted


BillPitcher:
Systems Affected:
	•	D&D 5e 2.4.3 & 3.0.5
	•	D&D 4e 4.4.3
	•	Foundry VTT v12 & v13

Issue:
Tokens with the Disposition set to “Secret” cannot be targeted or interacted with by players:
	•	No hover effect
	•	No ability to target (even with keyboard shortcuts)
	•	No visual cue the token exists

Expected Behaviour:
Tokens with Secret disposition should remain neutral/ambiguous in faction but still be:
	•	Targetable by players (if visible and not hidden)
	•	Highlightable on hover to indicate presence

Impact:
Players think the token is just part of the map or background art. This causes confusion and undermines encounters designed to be subtle or deceptive.

Repro Steps:
	1.	Place a token and set Disposition to Secret
	2.	Ensure token is visible and not hidden
	3.	Log in as a player with line of sight
	4.	Attempt to hover or target
	5.	Result: no feedback or targeting possible

Workaround:
None that preserves the intended “neutral but mysterious” behaviour. Changing disposition to Neutral or Friendly restores interactivity but alters perceived faction.

PR #5759, opened 2025-06-12
[#3782] Add "Difficult Terrain" region behavior

### LINKED ISSUE #3782: Add support for difficult terrain using scene regions
Now that we have scene regions it would be good to have the ability to have a region be marked as difficult terrain and then adjust measured movement accordingly, though that might need to wait until the drag ruler improvements in V13.

arbron:
Adds a new region behavior for representing areas of difficult terrain in a scene. This differs from the "Modify Movement Cost" behavior provided by core in a few key areas:

- Doesn't stack with other areas of difficult terrain
- Doesn't allow customizing the multiplier
- Adds ignored token dispositions to support difficult terrain that only affects your enemies
- Adds type of difficult terrain to support spells and abilities that allow ignoring certain types of difficult terrain

Difficult Terrain Region

The default terrain types provided are Magical Effect (so Ranger's Land Stride can ignore nonmagical difficult terrain), Rock or Rubble (Ring of Elemental Command's Earth focus ignores this), and Ice or Snow (Boots of the Winterlands and many creatures ignore this type of difficult terrain). Actors gain ignored difficult terrain types in their movement config.

Closes #3782

### DIFF
--- a/module/data/region-behavior/difficult-terrain.mjs
+++ b/module/data/region-behavior/difficult-terrain.mjs
@@ -16,7 +16,7 @@ export default class DifficultTerrainRegionBehaviorType extends foundry.data.reg
     delete dispositions[CONST.TOKEN_DISPOSITIONS.SECRET];
     return {
       magical: new BooleanField(),
-      types: new SetField(new StringField({ choices: () => CONFIG.DND5E.difficultTerrainTypes })),
+      types: new SetField(new StringField()),
       ignoredDispositions: new SetField(new NumberField({ choices: dispositions }))
     };
   }
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -194,7 +194,11 @@ Hooks.once("init", function() {
   });
 
   DocumentSheetConfig.unregisterSheet(RegionBehavior, "core", foundry.applications.sheets.RegionBehaviorConfig, {
-    types: ["dnd5e.rotateArea"]
+    types: ["dnd5e.difficultTerrain", "dnd5e.rotateArea"]
+  });
+  DocumentSheetConfig.registerSheet(RegionBehavior, "dnd5e", applications.regionBehavior.DifficultTerrainConfig, {
+    label: "DND5E.SheetClass.DifficultTerrain",
+    types: ["dnd5e.difficultTerrain"]
   });
   DocumentSheetConfig.registerSheet(RegionBehavior, "dnd5e", applications.regionBehavior.RotateAreaConfig, {
     label: "DND5E.SheetClass.RotateArea",
--- a/lang/en.json
+++ b/lang/en.json
@@ -3402,6 +3402,7 @@
   "Character": "Default 5e Character Sheet",
   "ClassSummary": "Default 5e Class Summary Sheet",
   "Container": "Default 5e Container Sheet",
+  "DifficultTerrain": "Default 5e Difficult Terrain Sheet",
   "Group": "Default 5e Group Sheet",
   "Item": "Default 5e Item Sheet",
   "JournalEntry": "Default 5e Journal Entry Sheet",
--- a/module/applications/region-behavior/_module.mjs
+++ b/module/applications/region-behavior/_module.mjs
@@ -1,1 +1,2 @@
+export {default as DifficultTerrainConfig} from "./difficult-terrain-config.mjs";
 export {default as RotateAreaConfig} from "./rotate-area-config.mjs";
--- a/module/applications/region-behavior/difficult-terrain-config.mjs
+++ b/module/applications/region-behavior/difficult-terrain-config.mjs
@@ -1,0 +1,21 @@
+/**
+ * Config sheet for the Difficult Terrain region behavior.
+ */
+export default class DifficultTerrainConfig extends foundry.applications.sheets.RegionBehaviorConfig {
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _getFields() {
+    const fieldsets = super._getFields();
+    for ( const fieldset of fieldsets ) {
+      const typesField = fieldset.fields.find(f => f.field.name === "types")?.field;
+      if ( typesField ) {
+        typesField.element.choices = CONFIG.DND5E.difficultTerrainTypes;
+        break;
+      }
+    }
+    return fieldsets;
+  }
+}


krbz999:
Not sold on using disposition as a way of marking someone as an enemy. That ascribes factions to tokens, which it cannot do.

For example, imagine a three-way battle between the players and two other groups, all three groups being enemies of each other. The player characters would have the "Friendly" disposition, but both enemy groups would likely have the "Hostile" disposition.

If enemy group A creates difficult terrain that affects only their enemies, who is affected?

module/config.mjs
```diff
@@ -2478,6 +2475,21 @@ DND5E.healingTypes = {
...
+/* -------------------------------------------- */
+/*  Movement                                    */
+/* -------------------------------------------- */
+
+/**
+ * Types of terrain that can cause difficult terrain.
+ * @enum {string}
+ */
+DND5E.difficultTerrainTypes = {
+  magical: "DND5E.REGIONBEHAVIORS.DIFFICULTTERRAIN.Type.Magical",
```
krbz999: Likelihood of this having to be turned into an enum of objects in the future: estimated 99%

Zanderaf:
> Not sold on using disposition as a way of marking someone as an enemy. That ascribes factions to tokens, which it cannot do.
> 
> For example, imagine a three-way battle between the players and two other groups, all three groups being enemies of each other. The player characters would have the "Friendly" disposition, but both enemy groups would likely have the "Hostile" disposition.
> 
> If enemy group A creates difficult terrain that affects only their enemies, who is affected?

This might be a bigger problem in general with "Disposition" as a whole. I feel like the same problem might occur with Auras whenever those are attempted to be implemented in the future. Disposition as binary as it is doesnt feel all too helpful unless its a case of Red vs Blue / Party vs Monster fights

JPMeehan:
In fairness, I think 2 sided conflicts are the norm in TTRPGs and if core wants to expand the disposition options then they can do that later and the system can adjust

krbz999:
I was actually paraphrasing the comment here: https://github.com/foundryvtt/dnd5e/issues/5694#issuecomment-2893682139

SagaTympana:
Some additional potentially useful terrain types: "water," ignorable by anyone with a swim speed, "wall/cliff/etc.," ignorable by anyone with a climb speed.

arbron:
> Some additional potentially useful terrain types: "water," ignorable by anyone with a swim speed, "wall/cliff/etc.," ignorable by anyone with a climb speed.

Water & cliffs are to be handled by movement action costs, not by difficult terrain (see https://github.com/foundryvtt/dnd5e/pull/5780).

module/data/region-behavior/difficult-terrain.mjs
```diff
@@ -0,0 +1,84 @@
...
+  static LOCALIZATION_PREFIXES = ["DND5E.REGIONBEHAVIORS.DIFFICULTTERRAIN"];
+
+  /* ---------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    return {
+      type: new StringField({ choices: () => CONFIG.DND5E.difficultTerrainTypes }),
+      ignoredDispositions: new SetField(new NumberField({
+        choices: foundry.applications.sheets.TokenConfig.TOKEN_DISPOSITIONS
```
dev7355608: We should probably not include Secret.

module/data/region-behavior/difficult-terrain.mjs
```diff
@@ -0,0 +1,84 @@
...
+
+  /** @override */
+  static LOCALIZATION_PREFIXES = ["DND5E.REGIONBEHAVIORS.DIFFICULTTERRAIN"];
+
+  /* ---------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    return {
+      type: new StringField({ choices: () => CONFIG.DND5E.difficultTerrainTypes }),
```
dev7355608: Could "magical snow" or "snowy rocks" be a thing? Perhaps we should have a SetField of terrain properties.

module/data/terrain-data.mjs
```diff
@@ -0,0 +1,46 @@
...
+  /* -------------------------------------------- */
+
+  /** @override */
+  static getMovementCostFunction(token, options) {
+    return (from, to, distance, segment) => {
+      let finalDistance = distance * (segment.terrain?.difficulty ?? 1);
+      if ( segment.terrain?.difficultTerrain ) finalDistance += distance;
+      return finalDistance;
+    };
+  }
```
dev7355608: Alternatively, we could implement it like this because we test whether difficult terrain applies in `DifficultTerrainRegionBehaviorType#_getTerrainEffects` and therefore we can already add it to `TerrainData#difficulty`.

```suggestion
  /** @override */
  static resolveTerrainEffects(effects) {
    let data = super.resolveTerrainEffects(effects);
    if ( !effects.some(e => e.name === "difficultTerrain") ) return data;
    if ( !data ) data = new this({ difficulty: 2, difficultTerrain: true });
    else {
      let difficulty = data.difficulty + 1;
      if ( !Number.isFinite(difficulty) ) difficulty = null;
      data.updateSource({ difficulty });
    }
    return data;
  }
```

module/data/terrain-data.mjs
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Extension of terrain data with support for 5e concepts.
+ */
+export default class TerrainData5e extends foundry.data.TerrainData {
+
+  /** @inheritDoc */
+  static defineSchema() {
+    return {
+      ...super.defineSchema(),
+      difficultTerrain: new BooleanField({ required: true })
```
dev7355608:
```suggestion
      difficultTerrain: new BooleanField()
```

`BooleanField` is `required: true` by default.

module/data/region-behavior/difficult-terrain.mjs
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /* ---------------------------------------- */
+
+  /** @override */
+  _getTerrainEffects(token, segment) {
+    const ignoredTypes = token.actor?.system.attributes?.movement?.ignoredDifficultTerrain ?? new Set();
+    if ( this.ignoredDispositions.has(token.disposition) || ignoredTypes.has("all")
+      || (this.types.size && !this.types.difference(ignoredTypes).size)
+      || (ignoredTypes.has("nonmagical") && !this.types.has("magical")) ) return [];
```
dev7355608: I think we need special logic for this "ignores magical" too: if a creature ignores magical terrain, I assume it shouldn't matter whether it's magical rocks or magical snow.

```suggestion
      || (ignoredTypes.has("magical") && this.types.has("magical")) ) return [];
      || (ignoredTypes.has("nonmagical") && !this.types.has("magical")) ) return [];
```

dev7355608: Perhaps this special logic and behavior for magical means that `magical` should be its on boolean field on the `DifficultTerrainRegionBehaviorType` data model and not in `CONFIG.DND5E.difficultTerrainTypes`.

module/config.mjs
dev7355608: I agree, we should turn this into an enum of objects. 
```suggestion
  magical: { label: "DND5E.REGIONBEHAVIORS.DIFFICULTTERRAIN.Type.Magical" },
```

module/data/region-behavior/difficult-terrain.mjs
```diff
@@ -0,0 +1,89 @@
...
+    return {
+      types: new SetField(new StringField({ choices: () => CONFIG.DND5E.difficultTerrainTypes })),
+      ignoredDispositions: new SetField(new NumberField({
+        choices: () => {
+          const dispositions = { ...foundry.applications.sheets.TokenConfig.TOKEN_DISPOSITIONS };
+          delete dispositions[CONST.TOKEN_DISPOSITIONS.SECRET];
+          return dispositions;
+        }
+      }))
+    };
```
dev7355608: This would be a bit better for performance: a `choices` function is called every validation.

```suggestion
    const dispositions = { ...foundry.applications.sheets.TokenConfig.TOKEN_DISPOSITIONS };
    delete dispositions[CONST.TOKEN_DISPOSITIONS.SECRET];
    return {
      types: new SetField(new StringField({ choices: () => CONFIG.DND5E.difficultTerrainTypes })),
      ignoredDispositions: new SetField(new NumberField({ choices: dispositions }))
    };
```

dev7355608 (changes requested):

module/data/terrain-data.mjs
dev7355608: One advantage of this is that the speed is automatically adjusted for difficult terrain in `Token#_modifyAnimationMovementSpeed` without us needing to override it.

dev7355608 (approved):

module/data/region-behavior/difficult-terrain.mjs
```diff
@@ -0,0 +1,87 @@
...
+
+  /* ---------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    const dispositions = { ...foundry.applications.sheets.TokenConfig.TOKEN_DISPOSITIONS };
+    delete dispositions[CONST.TOKEN_DISPOSITIONS.SECRET];
+    return {
+      magical: new BooleanField(),
+      types: new SetField(new StringField({ choices: () => CONFIG.DND5E.difficultTerrainTypes })),
```
Fyorl: We typically can't use `choices` where the choices come from `CONFIG`. It would mean that the region behaviour gets marked as invalid if a key is deleted the config, or a module that was adding a key is disabled. Maybe that's ok for region behaviours, but it's not something we do elsewhere in the system.

module/data/region-behavior/difficult-terrain.mjs
```diff
@@ -0,0 +1,87 @@
...
+      ignoredDispositions: new SetField(new NumberField({ choices: dispositions }))
```
Fyorl, line 20: `choices` here is ok since it comes from a `CONST`.

Fyorl (changes requested):

module/data/region-behavior/difficult-terrain.mjs
arbron: I did that initially to take advantage of the automatic generation of the region behavior config, but we can use a custom config to show the type choices instead.

Fyorl: Yeah, makes sense. Like I said, maybe we don't care about this for the region behaviour because it's unlikely someone will modify this config. I see you've already made the change, but I'll leave it up to you if you want to revert that last commit or not.

Fyorl (approved):

module/data/region-behavior/difficult-terrain.mjs
arbron: I'll stick with the custom sheet for consistency with how we handled choices across the system. This would be a good time to support `choices` one data fields that aren't validated, but are available for generated inputs (and possibly AE suggestions in the future).

Issue #5376, opened 2025-03-14
American measurement systems into METRIC system setting


Drumanin:
Hello. Maybe I dont understand smth, but that setting is basically broken and doing literally nothing? No text or items values are changed. Ive test it on a clear world too.
Maybe I didnt do smth important for it? I dont know. But for not its not working. And im not from american country and want to play comfortable with the system we are using.
Its possible to do smth about it in the system? At least setting fixing would be great.

Ive made personal table of conversion one system into another (with rounding of course for better numbers). If that can be useful with update, I can share it too.

arbron:
The metric setting doesn't change the values on existing items, but it sets the default values for new items and how encumbrance is calculated. Compendium content would need to be converted over on its own.

Drumanin:
So, its only for the weight?

Is it possible to make some kind of script for system conversion for other aspects? Or maybe macro that people can use for adaptation? Would be greatly appretiated!

JPMeehan:
I wonder if this is something a preCreate hook could handle for importing content from the SRD/premium modules?

PR #5764, opened 2025-06-13
[#5376] Improve conversion functions, auto-convert weights

arbron:
Extends `convertLength` and `convertWeight` to support automatic target unit selection similar to `convertTime`, and adds the ability for all three of those methods to automatically select the equivalent unit in another measurement system (e.g. mi => km).

This change is then used during item preparation to automatically convert item weights into the proper measurement system depending on the "Use Metric Weight Units" option. Since this is done at data preparation rather than sheet rendering, this setting now requires a restart.

 ### Todo
- [ ] Automatically convert length units


pgervais70:
Could it be possible also to expose the convertLength and convertWeight functions to be used by macros or modules?

JPMeehan:
> Could it be possible also to expose the convertLength and convertWeight functions to be used by macros or modules?

pretty sure they would be available under dnd5e.utils

PR #5777, opened 2025-06-20
Add compendiumSource  when drop item

absymylyar:


module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -1691,6 +1691,17 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
+    const { uuid } =
+      TextEditor$1.implementation.getDragEventData(event);
+
+    item.flags = {
+      ...(item.flags ?? {}),
+      dnd5e: {
+        ...(item.flags?.dnd5e ?? {}),
+        sourceId: item.flags?.dnd5e?.sourceId ?? uuid
+      }
+    };
```
krbz999: `TextEditor$1` is not a thing - and this manipulation of the `item.flags` also won't work - it's an item, not raw item data.

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -1712,7 +1723,17 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
-      if ( !(item instanceof Item) ) item = await fromUuid(item.uuid);
+      if ( !(item instanceof Item) ) {
+        const { uuid } = item;
+        item = await fromUuid(uuid);
+        item.flags = {
+          ...(item.flags ?? {}),
+          dnd5e: {
+            ...(item.flags?.dnd5e ?? {}),
+            sourceId: item.flags?.dnd5e?.sourceId ?? uuid
+          }
```
krbz999: Same here.

module/applications/actor/api/base-actor-sheet.mjs
absymylyar: TextEditor$1 is a Foundry object, I didn't get any depreciation warning about it, but if there's a better solution, I'm on.

I can use the .setFlag function, but I wouldn't like it to affect the original items in compendiums.
This code works, and add correctly the flags, but would it be safe if I use .setFlag intead ?

absymylyar: I tried to use .setFlag. it works, but it affects the original item from compendium. I will need to clone the item first

arbron:
Before this goes any further, why do you want to add this data on drop?

Currently this is only used for the advancement system and is largely an artifact of the time when `flags.core.sourceId` wasn't as reliable. Now that core provides `_stats.compendiumSource` and we've made sure it is reliably added by the system there doesn't seem to be a reason to add it. The system doesn't use this data outside of advancement at all, so adding it won't be useful since dropped items cannot participate in those specific advancement workflows.

absymylyar:
That's simple, because many items in compendiums are missing the _stats.compendiumSource, and in those case there's no other way to identify the origin.
And yes, I need a reliable way to do that.

krbz999:
Is core not adding `_stats.compendiumSource` on drop?

absymylyar:
> Is core not adding `_stats.compendiumSource` on drop?

Nope

arbron:
> Nope

Then that problem should be fixed, rather than adding the `dnd5e.flags.sourceId` flag.

absymylyar:
Okay, I can target _stats.compendiumSource instead.

Fyorl:
If it's not being added then we're missing a call to `Items5e#fromCompendium` (or `Items5e#importFromCompendium`) as part of the `_onDropCreateItems` workflow.

krbz999:
Core actorsheetv2 doesnt do it either. Not sure if we've overridden this entirely (I didnt look)
```js
  async _onDropItem(event, item) {
    if ( !this.actor.isOwner ) return;
    if ( this.actor.uuid === item.parent?.uuid ) return this._onSortItem(event, item);
    const keepId = !this.actor.items.has(item.id);
    await Item$1.implementation.create(item.toObject(), {parent: this.actor, keepId});
  }
  ```

**Comment by absymylyar** [comment]
I think that there's an other issue: between to time the item is dropped and the moment it's copied in the Actor's inventory, it could be altered. I noticed that when I tried to use `.setFlag`, and the original item in compendium changed.
To keep it safe, I would clone the item right after the `fromUuid`

**Comment by krbz999** [comment]
> I think that there's an other issue: between to time the item is dropped and the moment it's copied in the Actor's inventory, it could be altered. I noticed that when I tried to use `.setFlag`, and the original item in compendium changed. To keep it safe, I would clone the item right after the `fromUuid`

Well yes, the `item` is the "original" item - not some ephemeral item that is (but isn't?) created on the actor.

**Comment by absymylyar** [comment]
Check my new version, if it better fits

File: module/applications/actor/api/base-actor-sheet.mjs [hunk 3]
```diff
@@ -1712,7 +1719,17 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
+
+        const { uuid } = item;
+        item = await fromUuid(uuid);
+
+        if (!item._stats?.compendiumSource) {
+          const { uuid } =
+            TextEditor.implementation.getDragEventData(event);
+          return item.clone({ "item._stats.compendiumSource": uuid }, { keepId: true });
+        }
+      }
```
krbz999: Seems kind of redundant. Even if making a clone is the correct call, an item doesn't have an `item.item._stats` property.

Secondly, `window.TextEditor` is deprecated (you should be seeing a warning in the console).

And thirdly, you're grabbing the uuid even though you already have it.

absymylyar: Hell yes! You're right! I already have the uuid, I don't need to search it.... Just one simple line at the good place. Thanks

Fyorl:
I did some limited testing with dropping items from compendia and they all seemed to be given a `compendiumSource` as expected. I think this PR would benefit from an associated bug report with clear reproduction steps so we can understand what it is attempting to address.

absymylyar:
My bad... I tested it, and it works indeed. I was focused on the wrong datas.
I close this PR.

PR #5780, opened 2025-06-20
[#5779] Add system-specific movement cost functions

### LINKED ISSUE #5779: Link actor movement speeds with token movement actions
Currently when switching to "Swim" or "Climb" speed using the token HUD's movement action control it will always cost twice the movement, event if the actor has swim or climb speeds. The cost functions for those movement types should be adjusted to take actor data into account.

arbron:
Replaces the default movement cost functions provided by core with versions that take the actor's movement types into account. If the actor type has movement of that type then it will use a normal cost. If the actor doesn't have that type but its walk speed can be used in place of it, then it will increase the cost by 1 foot for each foot. If walking speed isn't a valid fallback, then they will not be permitted to use that movement type.

This includes an expansion of the `CONFIG.DND5E.movementTypes` object to include the walk fallback data.

Closes #5779


module/documents/token.mjs
```diff
@@ -107,6 +107,35 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+  /**
+   * Return the movement cost function for a specific movement type.
+   * @param {string} type
+   * @param {TokenDocument5e} token
+   * @param {TokenMeasureMovementPathOptions} options
+   * @returns {TokenMovementCostFunction}
+   */
+  static getMovementCostFunction(type, token, options) {
+    const actorMovement = token.actor?.system.attributes?.movement ?? {};
+    if ( !(type in actorMovement) || actorMovement[type] ) return cost => cost;
```
dev7355608, line 141:
```suggestion
    const actorMovement = token.actor?.system.attributes?.movement;
    if ( !actorMovement || (actorMovement[type] > 0) ) return cost => cost;
```

I assume this is equivalent. Perhaps easier to understand.

JPMeehan:
Question, will 5.1 also cover roll data in movement? The current behavior of items like Gloves of Swimming and Climbing make it difficult to accurately track these values, which isn't an issue currently but will have negative effects with this new automation.

arbron:
> Question, will 5.1 also cover roll data in movement? The current behavior of items like Gloves of Swimming and Climbing make it difficult to accurately track these values, which isn't an issue currently but will have negative effects with this new automation.

Hopefully that will get in, but if it doesn't this shouldn't cause problems. Since this only checks that you have at least some of a certain movement type to use it without penalty, those items will work fine if you put in a default value or a value of `1` like is often suggested.

module/documents/token.mjs
```diff
@@ -107,6 +107,35 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+  /*  Movement                                    */
+  /* -------------------------------------------- */
+
+  /**
+   * Set up the cost functions for all movement types.
+   */
+  static registerMovementCosts() {
+    for ( const type of ["walk", "fly", "swim", "burrow", "climb"] ) {
+      CONFIG.Token.movement.actions[type].getCostFunction = (...args) => this.getMovementCostFunction(type, ...args);
+    }
```
dev7355608:
```suggestion
    for ( const type of Object.keys(CONFIG.DND5E.movementTypes) ) {
      const actionConfig = CONFIG.Token.movement.actions[type];
      if ( !actionConfig ) continue;
      actionConfig.getCostFunction = (...args) => this.getMovementCostFunction(type, ...args);
    }
```

module/documents/token.mjs
```diff
@@ -107,6 +107,35 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Return the movement cost function for a specific movement type.
+   * @param {string} type
+   * @param {TokenDocument5e} token
+   * @param {TokenMeasureMovementPathOptions} options
+   * @returns {TokenMovementCostFunction}
```
dev7355608: Incorrect type.

```suggestion
   * @returns {TokenMovementActionCostFunction}
```

module/documents/token.mjs
```diff
@@ -107,6 +107,35 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+   */
+  static getMovementCostFunction(type, token, options) {
```
dev7355608:
```suggestion
  static getMovementActionCostFunction(type, token, options) {
```

More accurate name.

module/documents/token.mjs
```diff
@@ -107,6 +107,35 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
   }
 
+  /* -------------------------------------------- */
+  /*  Movement                                    */
+  /* -------------------------------------------- */
+
+  /**
+   * Set up the cost functions for all movement types.
+   */
+  static registerMovementCosts() {
```
dev7355608: I'm not sure if this function will would live very long. I assume we'll configure a lot more about movement actions, and we'll probably want to do it all in one place. So perhaps it for the best to not add this function and just add the code to the init hook. Or we already add `_configureMovementActions` (like `_configureStatusEffects`).

module/documents/token.mjs
```diff
@@ -107,6 +107,35 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+    for ( const type of ["walk", "fly", "swim", "burrow", "climb"] ) {
+      CONFIG.Token.movement.actions[type].getCostFunction = (...args) => this.getMovementCostFunction(type, ...args);
+    }
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Return the movement cost function for a specific movement type.
```
dev7355608:
```suggestion
   * Return the movement action cost function for a specific movement type.
```

module/config.mjs
```diff
@@ -2481,18 +2481,37 @@ preLocalize("healingTypes", { keys: ["label"] });
 /* -------------------------------------------- */
 
 /**
- * The valid units of measure for movement distances in the game system.
- * By default this uses the imperial units of feet and miles.
- * @enum {string}
+ * @typedef MovementConfig
```
dev7355608:
```suggestion
 * @typedef MovementTypeConfig
```

Perhaps a more accurate name.

module/documents/token.mjs
dev7355608: Do we need this to be a static method? It effectively can't be overridden by a module extends `TokenDocument5e`, I believe, because it is called too early. So perhaps we just make non-class function.

module/documents/token.mjs
arbron: I specifically checked for the presence of a movement type to handle things like the group actors, which don't have the same movement type names as the others. If we don't check in this way then group actors will never be able to be moved.

module/documents/token.mjs
arbron: I did want to make it available to modules that might add their own movement actions, which would have happen after the system's `init` hook so we wouldn't have a chance to do the auto-registration based on the system's movement type config.

module/documents/token.mjs
arbron: Maybe I can rename it to a more generic name for preparing all movement actions for the system. I was interested in keeping it `TokenDocument5e` where it will be closer to other movement-related code, rather than out in `dnd5e.mjs`.

module/documents/token.mjs
```diff
@@ -107,6 +107,37 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+  /* -------------------------------------------- */
+
+  /**
+   * Set up the system's movement action customization.
+   */
+  static registerMovementActions() {
+    for ( const type of Object.keys(CONFIG.DND5E.movementTypes) ) {
+      const actionConfig = CONFIG.Token.movement.actions[type];
+      if ( !actionConfig ) continue;
+      actionConfig.getCostFunction = (...args) => this.getMovementActionCostFunction(type, ...args);
```
dev7355608, line 126: Only half speed if the actor doesn't have this movement type.

```suggestion
      actionConfig.getCostFunction = (...args) => this.getMovementActionCostFunction(type, ...args);
      actionConfig.getAnimationOptions = token => {
        const actorMovement = token.actor?.system.attributes?.movement ?? {};
        if ( !(type in actorMovement) || actorMovement[type] ) return {};
        return {movementSpeed: CONFIG.Token.movement.defaultSpeed / 2};
      };
```

Technically the half speed is bit wrong with difficult terrain (slower than it should be). But probably not super important.

dev7355608 (approved):

Issue #5778, opened 2025-06-20
Add functionality for tokens blocking/increasing cost of movement


roth-michael:
Legacy Rules:
You can move through allies or hostile creatures that are 2+ sizes different from you. Both those cases treat the movement as difficult terrain. Hostile creatures that are not 2+ sizes different from you block movement (unless you have Halfling Nimbleness, in which case you can move - with difficulty - through any creature larger than you).

Modern Rules:
Movement through allies is not difficult terrain. Hostile creatures that are Tiny can be moved through with no difficulty. Hostile creatures that are incapacitated, or 2+ sizes different from you, can be moved through _with difficulty_. Hostile creatures that are not 2+ sizes different from you block movement (unless you have Halfling Nimbleness, in which case you can move - with difficulty - through any hostile creature larger than you).

PR #5781, opened 2025-06-20
[#5778] Implement movement automation (tokens blocking/increasing cost of movement)

roth-michael:
Adds a setting (default enabled) which will take tokens in path of movement into account when determining cost. According to rules setting (and the presence of Halfling Nimbleness special trait), determines whether a token should block movement or count as difficult terrain.

TODO:
- [ ] Wait for next core stable to ensure bug from overriding `Token#findMovementPath` is resolved
- [ ] Determine whether setting belongs in a sub-menu
- [ ] Add a keybind to ignore movement automation while moving
- [ ] Ensure token-caused difficult terrain works for larger tokens moving through smaller tokens
- [ ] Add a way of recognizing something like "Freedom of Movement" to ignore the difficult terrain penalties imposed by tokens
- [ ] Maybe try to get something working in gridless


module/canvas/token.mjs
```diff
@@ -18,6 +18,46 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+  /** @inheritdoc */
+  _getMovementCostFunction(options) {
+    const costFunction = super._getMovementCostFunction(options);
+    const disableAutomationKey = foundry.helpers.interaction.KeyboardManager.MODIFIER_KEYS.ALT;
+    if ( !game.settings.get("dnd5e", "movementAutomation")
+      || game.keyboard.isModifierActive(disableAutomationKey) ) return costFunction;
+
+    const ignoredDifficultTerrain = this.actor?.system.attributes.movement.ignoredDifficultTerrain ?? new Set();
+    const ignoreDifficult = ["all", "nonmagical"].some(i => ignoredDifficultTerrain.has(i));
+    const unconstrainedMovement = game.user.isGM && ui.controls.controls.tokens.tools.unconstrainedMovement.active;
```
dev7355608, line 38:
```suggestion
```

dev7355608:
```suggestion
```

Unconstrained movement shouldn't change the cost: it just allows movement with infinite cost.

module/canvas/token.mjs
```diff
@@ -18,6 +18,46 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+    return (from, to, distance, segment) => {
+      const cost = costFunction(from, to, distance, segment);
+
+      // Check if blocked
+      if ( !unconstrainedMovement && this.layer.isOccupiedGridSpaceBlocking(to, this) ) return Infinity;
```
dev7355608:
```suggestion
      if ( this.layer.isOccupiedGridSpaceBlocking(to, this) ) return Infinity;
```

module/canvas/token.mjs
```diff
@@ -18,6 +18,46 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  _getMovementCostFunction(options) {
+    const costFunction = super._getMovementCostFunction(options);
+    const disableAutomationKey = foundry.helpers.interaction.KeyboardManager.MODIFIER_KEYS.ALT;
+    if ( !game.settings.get("dnd5e", "movementAutomation")
+      || game.keyboard.isModifierActive(disableAutomationKey) ) return costFunction;
```
dev7355608:
```suggestion
    if ( !game.settings.get("dnd5e", "movementAutomation") ) return costFunction;
```

We cannot do this: `_getMovementCostFunction` isn't called only during an interaction workflow. We don't have the API to support the addition of measurement `options` at the moment, that could enable overriding measurement behavior for a particular movement.

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,15 @@
+export default class TokenRuler5e extends foundry.canvas.placeables.tokens.TokenRuler {
+  refresh({passedWaypoints, pendingWaypoints, plannedMovement}) {
+    for (const movement of Object.values(plannedMovement)) {
+
+      // Show distance of furthest possible waypoint
+      if (movement.foundPath?.length) movement.foundPath.at(-1).explicit = true;
```
dev7355608:
```suggestion
```

The `rulerData` parameter is typed `DeepReadonly<TokenRulerData>`.

> Show distance of furthest possible waypoint

We would need to do this by overriding `_getWaypointStyle` and `_getWaypointLabelContext`, if desired. It is also probably important to style nonexplict waypoints that are shown a bit differently so that the user doesn't confuse it with waypoints they placed.

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Determine the set of tokens occupying the provided grid space which may be relevant for blocking/difficult terrain
+   * considerations
+   * @param {GridOffset3D} gridSpace  The grid space to check
+   * @param {Token5e} token           The token being moved
+   * @returns {Set<Token5e>} The set of potentially relevant tokens occupying the provided grid space
+   * @private
+   */
+  _getRelevantOccupyingTokens(gridSpace, token) {
```
dev7355608:
```suggestion
   */
  #getRelevantOccupyingTokens(gridSpace, token) {
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,96 @@
...
+    return found.filter(t => {
+      // Ignore self
+      if ( t === token ) return false;
+
+      // Ignore different elevation
+      const { k: occupiedElevation } = grid.getOffset(t.document);
+      if ( occupiedElevation !== gridSpace.k ) return false;
+
+      // Ensure space is actually occupied, not merely touching border of rectangle
+      const gridSpaces = t.document.getOccupiedGridSpaceOffsets().map(coord => grid.getTopLeftPoint(coord));
```
dev7355608:
```suggestion
      const gridSpaces = t.document.getOccupiedGridSpaceOffsets(this.document._source).map(coord => grid.getTopLeftPoint(coord));
```

Need to pass the `_source` because the prepared values are animated, which would lead to wrong results.

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,96 @@
...
+      return gridSpaces.some(coord => coord.x === topLeft.x && coord.y === topLeft.y);
```
dev7355608:
```suggestion
      const gridSpaces = t.document.getOccupiedGridSpaceOffsets(this.document._source);
      return gridSpaces.some(coord => (coord.i === gridSpace.i) && (coord.j === gridSpace.j) && (coord.k === gridSpace.k));
```
Better to work with grid offset coordinates here.

dnd5e.mjs
```diff
@@ -62,6 +62,7 @@ Hooks.once("init", function() {
   CONFIG.ActiveEffect.documentClass = documents.ActiveEffect5e;
   CONFIG.ActiveEffect.legacyTransferral = false;
   CONFIG.Actor.documentClass = documents.Actor5e;
+  CONFIG.Canvas.layers.tokens.layerClass = canvas.layers.TokenLayer5e;
```
dev7355608:
```suggestion
  CONFIG.Canvas.layers.tokens.layerClass = CONFIG.Token.layerClass = canvas.layers.TokenLayer5e;
```

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,15 @@
...
+
+      // Currently, infinite-cost waypoints come through socketing as null-cost. This fixes that
+      for (const waypoint of movement.unreachableWaypoints ?? []) {
+        waypoint.cost ??= Infinity;
+      }
```
dev7355608: I'm not sure we should fix this core bug in dnd5e. We also don't need to fix it if we handle occupied blocking spaces in `constrainMovementPath`.
```suggestion
```

module/canvas/token.mjs
```diff
@@ -18,6 +18,46 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+  findMovementPath(waypoints, options) {
+    waypoints = this.document.getCompleteMovementPath(waypoints);
+    return super.findMovementPath(waypoints, options);
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
+  _getMovementCostFunction(options) {
+    const costFunction = super._getMovementCostFunction(options);
```
dev7355608, line 32:
```suggestion
    const costFunction = super._getMovementCostFunction(options);
    const preview = options.preview && !game.user.isGM;
```

module/canvas/token.mjs
```diff
@@ -18,6 +18,46 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+    const disableAutomationKey = foundry.helpers.interaction.KeyboardManager.MODIFIER_KEYS.ALT;
+    if ( !game.settings.get("dnd5e", "movementAutomation")
+      || game.keyboard.isModifierActive(disableAutomationKey) ) return costFunction;
+
+    const ignoredDifficultTerrain = this.actor?.system.attributes.movement.ignoredDifficultTerrain ?? new Set();
+    const ignoreDifficult = ["all", "nonmagical"].some(i => ignoredDifficultTerrain.has(i));
+    const unconstrainedMovement = game.user.isGM && ui.controls.controls.tokens.tools.unconstrainedMovement.active;
+    return (from, to, distance, segment) => {
+      const cost = costFunction(from, to, distance, segment);
+
```
dev7355608, line 40:
```suggestion

  // Don't apply difficult terrain that the user doesn't know about
  if ( preview ) {
    const c = canvas.grid.getCenterPoint(to);
    if ( !(canvas.fog.isPointExplored(c) || canvas.visibility.testVisibility(c, {tolerance: 1})) ) return cost;
  }

```

I didn't test this and we might want to increase the tolerance for both the visibility and explored test.

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /**
+   * Determine whether the provided grid space is being occupied by a token which should cause difficult terrain for
+   * the provided token
+   * @param {GridOffset3D | TokenFindMovementPathWaypoint} gridSpace  The grid space to check
+   * @param {Token5e} token                                           The token being moved
+   * @returns {boolean} Whether the moving token should suffer difficult terrain
+   */
+  isOccupiedGridSpaceDifficult(gridSpace, token) {
```
dev7355608: If we have preview measurement, then tokens that are not visible should not apply difficult terrain.

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,96 @@
...
+   * @returns {boolean} Whether the moving token should be blocked
+   */
+  isOccupiedGridSpaceBlocking(gridSpace, token) {
+    const found = this._getRelevantOccupyingTokens(gridSpace, token);
+    const tokenSize = CONFIG.DND5E.actorSizes[token.actor?.system.traits.size]?.numerical ?? 2;
+    const modernRules = game.settings.get("dnd5e", "rulesVersion") === "modern";
+    const halflingNimbleness = token.actor?.getFlag("dnd5e", "halflingNimbleness");
+    return found.some(t => {
+      // Friendly tokens never block movement
+      if ( token.document.disposition === t.document.disposition ) return false;
```
dev7355608, line 18: Tokens with the Secret disposition should never affect movement: otherwise the ruler would reveal the secret.

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,96 @@
...
+
+      // Dead creatures never block movement, as they are objects
+      if ( t.actor?.statuses.has("dead") ) return false;
```
dev7355608: Petrified and Ethereal creatures should probably also not block movement.

module/canvas/token.mjs
```diff
@@ -18,6 +18,53 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+    // Get all grid spaces as waypoints so that running into a blocking token stops us immediately before it
+    waypoints = this.document.getCompleteMovementPath(waypoints);
+
+    // Drop all intermediate waypoints except those immediately before a blocking token
+    waypoints = waypoints.filter((waypoint, i) => {
+      return !waypoint.intermediate || this.layer.isOccupiedGridSpaceBlocking(waypoints[i + 1], this);
+    });
+    return super.findMovementPath(waypoints, options);
+  }
```
dev7355608: Instead of this here and giving occupied blocking spaces an infinite cost, we should implement token blocking in `Token#constrainMovementPath`. It has the advantages are that API moves that do not use `findMovementPath` stop in front of the blocking token as well and we don't have infinity cost in ruler waypoint labels.

module/canvas/token.mjs
dev7355608: We also don't need this once we handle occupied blocking spaces in `constrainMovementPath`.

module/canvas/layers/tokens.mjs
dev7355608, line 18: I suppose there's no better way to handle this at the moment and we have to assume all hostile tokens are on the same team and all neutral tokens are on the same team.

module/canvas/layers/tokens.mjs
dev7355608: We might need some API here for modules like Item Piles: an item shouldn't block movement. Perhaps we add the `"object"` status.

dev7355608 (changes requested):

module/canvas/token.mjs
Fyorl: Is there some other way we can do this? Can we check what context we're in? Can we disable this from some other location? This also needs to be unified with the difficult terrain region ideally, i.e. you use a single modifier to turn off all movement cost across tokens + difficult terrain. 

If there's no way to do it, then global settings to opt-into or out of movement automation entirely will have to do, but having a modifier toggle would allow the user to make use of movement automation in general, but force it off temporarily for various exceptions that are not covered by the system.

roth-michael: So, this means that currently there's no way to implement ignoring blocking/difficult tokens on a per-movement basis besides toggling unconstrained movement? Just want to make sure.

module/canvas/layers/tokens.mjs
roth-michael: This should be addressed by the suggested change to `_getMovementCostFunction`, right?

module/canvas/layers/tokens.mjs
roth-michael: It might also make sense to add something to `CONFIG.DND5E` for "statuses which should make a token never block movement" - that way it's an easy fix for someone who disagrees with any assumptions we make here.

module/canvas/layers/tokens.mjs
roth-michael: Yeah, we run into a similar issue in the difficult terrain PR. Without a more explicit implementation of "faction" I think this is the best we can do.

module/canvas/token.mjs
roth-michael: Extending `Token#constrainMovementPath` was my original plan of attack. The main issue I ran into is that it makes use of a couple of full-private functions, so I'd have to call `super` on it and then go through the returned path _again_ - plus something like `#getCollisionWaypoint` would have to be re-implemented in order to get the space prior to a blocking token. Will look into this again today, though.

module/canvas/layers/tokens.mjs
dev7355608: I think if we just test `token.visible` we shouldn't need the change I suggested in `_getMovementCostFunction`. If you see one square of a token, you see how large it is anyway and all spaces it occupies, even those that you don't see.

roth-michael:
~~(The conversations I've marked as "resolved" I have addressed locally, but haven't yet pushed because implementing _only_ those changes leaves the PR broken)~~

module/canvas/layers/tokens.mjs
roth-michael: That makes sense. Moving that logic into the `isOccupiedGridSpaceX` functions, do we still need to check for whether it's a preview measurement? If so, might make sense to throw in an `options` param I suppose.

module/canvas/token.mjs
dev7355608: > So, this means that currently there's no way to implement ignoring blocking/difficult tokens on a per-movement basis besides toggling unconstrained movement? Just want to make sure.

Constrain options can be customized for a particular movement. Measure options cannot at the moment. I logged an issue for that: https://github.com/foundryvtt/foundryvtt/issues/13127

> Is there some other way we can do this? Can we check what context we're in? Can we disable this from some other location?

No, there isn't. A toggle instead of a modifier would be better, but it still would have the same problems:
- If the user makes a multi waypoint movement and toggles the state before the movement completed, the measurement of the remaining waypoint changes.
- All movements via API would be subject to this toggle state.

module/canvas/token.mjs
dev7355608: Something like this should work:

```js
  /** @inheritdoc */
  constrainMovementPath(waypoints, options) {
    const {preview=false, ignoreTokens=false} = options; // Custom constrain option to ignore tokens

    if ( ignoreTokens ) return super.constrainMovementPath(waypoints, options);

    // Ignore preview if token vision is disabled or the current user is a GM
    if ( !canvas.visibility.tokenVision || game.user.isGM ) preview = false;

    const completePath = this.document.getCompleteMovementPath(waypoints);
    const blockedIndex = completePath.findIndex(waypoint => this.layer.isOccupiedGridSpaceBlocking(waypoint, this, {preview}));
    const blocked = blockedIndex  >= 0;

    if ( blockedIndex === 0 ) {} // ...

    waypoints = completePath.slice(0, blockedIndex - 1).filter(waypoint => !waypoint.intermediate);
    waypoints.push(completePath.at(blockedIndex - 1));

    const [path, constrained] = super.constrainMovementPath(waypoints, options);
    return [path, constrained || blocked];
  }
```

module/canvas/layers/tokens.mjs
dev7355608: If it isn't a preview measurement, we need test all tokens, otherwise the token might move through a token without difficult being applied if the token starts its move in a position where it cannot see the other token yet.

Options (`{preview=false}={})` parameter sounds good.

module/canvas/layers/tokens.mjs
roth-michael: This works fine when passing offsets, but when passing `TokenFindMovementPathWaypoint`s (e.g. from within `constrainMovementPath`, would need to convert the waypoint into offset. Better to do it that way, or to do it the way I had it originally (converting all to Point)?

module/canvas/token.mjs
roth-michael: Duh, yes this works wonderfully. Is the `blockedIndex === 0` check intended to ensure movement is blocked if, I suppose, the token is _inside_ a blocking token? That's the only instance I can think of where that'd come up, and honestly in that case I would opt for _not_ blocking movement.

roth-michael: https://github.com/user-attachments/assets/d86d8712-2289-4930-8e86-5ff8566f7550

Also - this is rather an edge case, but I'm not sure what to make of this strange interaction between tokens that block movement and walls that block movement.

dev7355608: > Also - this is rather an edge case, but I'm not sure what to make of this strange interaction between tokens that block movement and walls that block movement.

The snapping of the wall collision puts the token inside the other tokens space and the original path does not go through the other tokens space. So we probably need to check for blocking again if super constrained, and if blocked then super constrain again. In a loop. And hopefully it can't be an infinite loop.

roth-michael: I think it couldn't be - eventually it'd work its way back to the originating waypoint, right?

dev7355608: > Is the `blockedIndex === 0` check intended to ensure movement is blocked if, I suppose, the token is _inside_ a blocking token?

Yes.

> That's the only instance I can think of where that'd come up, and honestly in that case I would opt for _not_ blocking movement.

I think I'm lean towards not blocking too.

dev7355608: > I think it couldn't be - eventually it'd work its way back to the originating waypoint, right?

I think so. But perhaps there's some weird edge case. It's probably a good idea to be defensive and limit the number of iterations. If no valid path was found after all these iterations, we can just drop the collision waypoint to should get a valid path.

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * Determine the set of tokens occupying the provided grid space which may be relevant for blocking/difficult terrain
+   * considerations
+   * @param {GridOffset3D | TokenFindMovementPathWaypoint} gridSpace  The grid space to check
+   * @param {Token5e} token                                           The token being moved
+   * @param {object} options                                          Additional options
+   * @param {boolean} options.preview                                 Whether the movement in question is previewed
+   * @returns {Set<Token5e>} The set of potentially relevant tokens occupying the provided grid space
+   * @private
```
dev7355608:
```suggestion
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+   */
+  #getRelevantOccupyingTokens(gridSpace, token, {preview=false}={}) {
+    const grid = canvas.grid;
```
dev7355608, line 76:
```suggestion
    const grid = canvas.grid;
    if ( grid.isGridless ) return [];
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+      // ruleset. Modern ruleset incapacitated tokens should still be difficult even if within one size category.
+      // Dead creatures (objects) may be difficult terrain, for now that is the default treatment.
+      return friendlyToken || Math.abs(tokenSize - occupiedSize) >= 2 || mayHaveBlocked || t.actor?.statuses.has("dead");
+    });
+  }
+
+  /**
+   * Determine the set of tokens occupying the provided grid space which may be relevant for blocking/difficult terrain
+   * considerations
+   * @param {GridOffset3D | TokenFindMovementPathWaypoint} gridSpace  The grid space to check
```
dev7355608:
```suggestion
   * @param {GridOffset3D} gridSpace  The grid space to check
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+
+      // A size difference of less than 2 should block (adjust for Halfling Nimbleness)
+      return (Math.abs(tokenSize - occupiedSize) < 2);
+    });
+  }
+
+  /**
+   * Determine whether the provided grid space is being occupied by a token which should cause difficult terrain for
+   * the provided token
+   * @param {GridOffset3D | TokenFindMovementPathWaypoint} gridSpace  The grid space to check
```
dev7355608:
```suggestion
   * @param {GridOffset3D} gridSpace  The grid space to check
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
+export default class TokenLayer5e extends foundry.canvas.layers.TokenLayer {
+  /**
+   * Determine whether the provided grid space is being occupied by a token which should block the provided token
+   * @param {GridOffset3D | TokenFindMovementPathWaypoint} gridSpace  The grid space to check
```
dev7355608:
```suggestion
   * @param {GridOffset3D} gridSpace  The grid space to check
```

We won't need to support point coordinates after the changes in `constrainMovementPath`.

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+    return found.some(t => {
+      // Friendly tokens never block movement
+      if ( token.document.disposition === t.document.disposition ) return false;
+
+      // If creature has any statuses that should never block movement, don't block movement
+      if ( neverBlockStatuses.some(status => t.actor?.statuses.has(status)) ) return false;
+
+      const occupiedSize = CONFIG.DND5E.actorSizes[t.actor?.system.traits.size]?.numerical ?? 2;
+      // In modern rules, Tiny creatures can be moved through
+      if ( modernRules && occupiedSize === 0 ) return false;
```
dev7355608:
```suggestion
      if ( modernRules && (occupiedSize === 0) ) return false;
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+
+      // Halfling Nimbleness means no larger creature can block
+      if ( halflingNimbleness && occupiedSize > tokenSize ) return false;
```
dev7355608:
```suggestion
      if ( halflingNimbleness && (occupiedSize > tokenSize) ) return false;
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
+export default class TokenLayer5e extends foundry.canvas.layers.TokenLayer {
+  /**
+   * Determine whether the provided grid space is being occupied by a token which should block the provided token
+   * @param {GridOffset3D | TokenFindMovementPathWaypoint} gridSpace  The grid space to check
+   * @param {Token5e} token                                           The token being moved
+   * @param {object} options                                          Additional options
+   * @param {boolean} options.preview                                 Whether the movement in question is previewed
```
dev7355608:
```suggestion
   * @param {object} [options]                                          Additional options
   * @param {boolean} [options.preview=false]                                 Whether the movement in question is previewed
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+    });
+  }
+
+  /**
+   * Determine whether the provided grid space is being occupied by a token which should cause difficult terrain for
+   * the provided token
+   * @param {GridOffset3D | TokenFindMovementPathWaypoint} gridSpace  The grid space to check
+   * @param {Token5e} token                                           The token being moved
+   * @param {object} options                                          Additional options
+   * @param {boolean} options.preview                                 Whether the movement in question is previewed
```
dev7355608:
```suggestion
   * @param {object} [options]                                          Additional options
   * @param {boolean} [options.preview=false]                                 Whether the movement in question is previewed
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+    const neverBlockStatuses = CONFIG.statusEffects.filter(s => s.neverBlockMovement).map(s => s.id);
+    return found.some(t => {
+      const friendlyToken = token.document.disposition === t.document.disposition;
+
+      // In modern rules, friendly tokens are not difficult terrain
+      if ( modernRules && friendlyToken ) return false;
+      const occupiedSize = CONFIG.DND5E.actorSizes[t.actor?.system.traits.size]?.numerical ?? 2;
+
+      // In modern rules, Tiny creatures are not difficult terrain
+      if ( modernRules && occupiedSize === 0 ) return false;
```
dev7355608:
```suggestion
      if ( modernRules && (occupiedSize === 0) ) return false;
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+
+      // Halfling Nimbleness means a creature 1 size greater (normally would block) causes difficult terrain
+      if ( halflingNimbleness && occupiedSize === tokenSize + 1 ) return true;
```
dev7355608:
```suggestion
      if ( halflingNimbleness && (occupiedSize === tokenSize + 1) ) return true;
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+
+      // Tokens which may normally have blocked movement should still be considered difficult terrain
+      const mayHaveBlocked = neverBlockStatuses.some(status => t.actor?.statuses.has(status));
+
+      // Friendly means legacy, therefore difficult. A size difference of 2 or more is difficult terrain regardless of
+      // ruleset. Modern ruleset incapacitated tokens should still be difficult even if within one size category.
+      // Dead creatures (objects) may be difficult terrain, for now that is the default treatment.
+      return friendlyToken || Math.abs(tokenSize - occupiedSize) >= 2 || mayHaveBlocked || t.actor?.statuses.has("dead");
```
dev7355608:
```suggestion
      return friendlyToken || (Math.abs(tokenSize - occupiedSize) >= 2) || mayHaveBlocked || t.actor?.statuses.has("dead");
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+    });
+  }
+
+  /**
+   * Determine the set of tokens occupying the provided grid space which may be relevant for blocking/difficult terrain
+   * considerations
+   * @param {GridOffset3D | TokenFindMovementPathWaypoint} gridSpace  The grid space to check
+   * @param {Token5e} token                                           The token being moved
+   * @param {object} options                                          Additional options
+   * @param {boolean} options.preview                                 Whether the movement in question is previewed
```
dev7355608:
```suggestion
   * @param {object} [options]                                          Additional options
   * @param {boolean} [options.preview=false]                                 Whether the movement in question is previewed
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+      // Always ignore secret tokens
+      if ( t.document.disposition === CONST.TOKEN_DISPOSITIONS.SECRET ) return false;
+
+      // Ignore different elevation
+      const { k: occupiedElevation } = grid.getOffset(t.document);
+      if ( occupiedElevation !== (gridSpace.k) ) return false;
+
+      // Ensure space is actually occupied, not merely touching border of rectangle
+      const gridSpaces = t.document.getOccupiedGridSpaceOffsets(t.document._source);
+      return gridSpaces.some(coord => coord.i === gridSpace.i && coord.j === gridSpace.j);
```
dev7355608:
```suggestion
      return gridSpaces.some(coord => (coord.i === gridSpace.i) && (coord.j === gridSpace.j));
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+   * @param {GridOffset3D | TokenFindMovementPathWaypoint} gridSpace  The grid space to check
+   * @param {Token5e} token                                           The token being moved
+   * @param {object} options                                          Additional options
+   * @param {boolean} options.preview                                 Whether the movement in question is previewed
+   * @returns {Set<Token5e>} The set of potentially relevant tokens occupying the provided grid space
+   * @private
+   */
+  #getRelevantOccupyingTokens(gridSpace, token, {preview=false}={}) {
+    const grid = canvas.grid;
+    gridSpace = grid.getOffset(gridSpace);
```
dev7355608:
```suggestion
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+    const topLeft = grid.getTopLeftPoint(gridSpace);
+    const rect = new PIXI.Rectangle(topLeft.x, topLeft.y, grid.sizeX, grid.sizeY);
```
dev7355608, line 79:
```suggestion
    const rect = new PIXI.Rectangle(topLeft.x, topLeft.y, grid.sizeX, grid.sizeY);
    const lowerElevation = gridSpace.k * grid.distance;
    const upperElevation = (gridSpace.k + 1) * grid.distance;
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+
+      // If preview movement, don't reveal blocked or difficult terrain for non-visible tokens
+      if ( preview && !t.visible ) return false;
+
+      // Always ignore secret tokens
+      if ( t.document.disposition === CONST.TOKEN_DISPOSITIONS.SECRET ) return false;
+
+      // Ignore different elevation
+      const { k: occupiedElevation } = grid.getOffset(t.document);
+      if ( occupiedElevation !== (gridSpace.k) ) return false;
```
dev7355608:
```suggestion
      const occupiedElevation = t.document._source.elevation;
      if ( (occupiedElevation >= lowerElevation) && (occupiedElevation < upperElevation) ) return false;
```

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,19 @@
+export default class TokenRuler5e extends foundry.canvas.placeables.tokens.TokenRuler {
+  /** @inheritdoc */
+  _getWaypointStyle(waypoint) {
+    let modifiedWaypoint = waypoint;
+    if ( !waypoint.unreachable && waypoint.next?.unreachable ) {
+      modifiedWaypoint = { ...waypoint, explicit: true };
+    }
+    return super._getWaypointStyle(modifiedWaypoint);
```
dev7355608: We should probably have a different styling for nonexplicit waypoints so that the user doesn't confuse them with waypoints they placed themselves and right-click to remove it and are surprised it removes another waypoint. Not sure how likely that confusion would be though.

```suggestion
    const style = super._getWaypointStyle(modifiedWaypoint);
    if ( waypoint !== modifiedWaypoint ) style.alpha = (style.alpha ?? 1) * 0.5;
    return style;
```

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,19 @@
+export default class TokenRuler5e extends foundry.canvas.placeables.tokens.TokenRuler {
+  /** @inheritdoc */
+  _getWaypointStyle(waypoint) {
+    let modifiedWaypoint = waypoint;
+    if ( !waypoint.unreachable && waypoint.next?.unreachable ) {
```
dev7355608:
```suggestion
    if ( !waypoint.explicit && !waypoint.unreachable && waypoint.next?.unreachable ) {
```

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,19 @@
...
+      modifiedWaypoint = { ...waypoint, explicit: true };
+    }
+    return super._getWaypointStyle(modifiedWaypoint);
+  }
+
+  /** @inheritdoc */
+  _getWaypointLabelContext(waypoint, state) {
+    let modifiedWaypoint = waypoint;
+    if ( !waypoint.unreachable && waypoint.next?.unreachable ) {
```
dev7355608:
```suggestion
    if ( !waypoint.explicit && !waypoint.unreachable && waypoint.next?.unreachable ) {
```

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,19 @@
+export default class TokenRuler5e extends foundry.canvas.placeables.tokens.TokenRuler {
+  /** @inheritdoc */
+  _getWaypointStyle(waypoint) {
+    let modifiedWaypoint = waypoint;
+    if ( !waypoint.unreachable && waypoint.next?.unreachable ) {
+      modifiedWaypoint = { ...waypoint, explicit: true };
```
dev7355608: This works, but it violates `modifiedWaypoint.previous.next === modifiedWaypoint` and `modifiedWaypoint.next.previous === modifiedWaypoint`. But this isn't causing in the implementation of core's `TokenRuler`. Ideally we wouldn't do this, but then we had to copy the code from core's `TokenRuler`.

module/canvas/token.mjs
```diff
@@ -18,6 +18,86 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+
+    // If constrained, ensure we check for being blocked in the new path (repeat as necessary)
+    if (constrained) {
+      let numOldWaypoints;
+      do {
+        numOldWaypoints = waypoints.filter(w => !w.unreachable).length;
+        checkBlockedConstrained();
+      } while (numOldWaypoints !== waypoints.filter(w => !w.unreachable).length);
+    }
+    return [waypoints, constrained || blocked];
```
dev7355608: Rewritten the code with a couple of corrections. I didn't test it though.

- Ignore the first (origin) grid space, but if the first would be blocking and is ignored blocking grid spaces after should block.
- Check all occupied grid spaces by this token properly along the path.
- Only 10 tries, then constrain the path without the last waypoints (prevent potential infinite loops).
- Skip unnecessary `super.constrainMovementPath` in later iterations if nothing changed because of blocking.

```suggestion
    let path = waypoints;
    let constrained = false;

    for ( let k = 0; k < 10; k++ ) {
    
      // Apply blocking constraints
      const completePath = this.document.getCompleteMovementPath(path);
      let blockedIndex;
      for ( let i = 1; i < completePath.length; i++ ) {
        const waypoint = completePath[i];
        const occupiedGridSpaces = this.document.getOccupiedGridSpaceOffsets(waypoint);
        if ( occupiedGridSpaces.some(gridSpace => this.layer.isOccupiedGridSpaceBlocking(gridSpace, this, {preview})) ) {
          blockedIndex = i;
          break;
        }
      }
      const blocked = blockedIndex >= 1;
      if ( blocked ) {
        path = completePath.slice(0, blockedIndex - 1).filter(waypoint => !waypoint.intermediate);
        path.push(completePath.at(blockedIndex - 1));
        constrained = true;
      }

      // Test wall/cost constraints in the first iteration always and in later iterations only if the path changed due to blocking
      if ( (k === 0) || blocked ) {
        const [constrainedPath, wasConstrained] = super.constrainMovementPath(path, options);
        path = constrainedPath;
        if ( !wasConstrained ) return [path, constrained]; // No change: path is valid
        constrained = true;
      }
      
      // In a later iteration if there was no change due to blocking, we found a valid path
      else if ( !blocked ) return [path, constrained];
    }
    
    // After 10 failed attempts to find a valid path, remove the last waypoints and constrain this path
    [path] = this.constrainMovementPath(waypoints.slice(0, -1), options);
    return [path, true];
```

module/canvas/token.mjs
```diff
@@ -18,6 +18,86 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+    if ( !canvas.visibility.tokenVision || game.user.isGM ) preview = false;
+
+    let blocked = false;
+    let constrained = false;
+
+    // Check if blocked by token, then check if constrained
+    const checkBlockedConstrained = () => {
+      const completePath = this.document.getCompleteMovementPath(waypoints);
+      const blockedIndex = completePath.findIndex(waypoint =>
+        this.layer.isOccupiedGridSpaceBlocking(waypoint, this, {preview}));
```
dev7355608: We need to test the occupied grid space of this token.

module/canvas/token.mjs
```diff
@@ -18,6 +18,86 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+      [waypoints, constrained] = super.constrainMovementPath(waypoints, options);
+    };
+
+    checkBlockedConstrained();
+
+    // If constrained, ensure we check for being blocked in the new path (repeat as necessary)
+    if (constrained) {
+      let numOldWaypoints;
+      do {
+        numOldWaypoints = waypoints.filter(w => !w.unreachable).length;
```
dev7355608: `unreachable` isn't a property that should exist here.

module/config.mjs
```diff
@@ -3856,7 +3863,8 @@ DND5E.conditionTypes = {
   incapacitated: {
     name: "DND5E.ConIncapacitated",
     img: "systems/dnd5e/icons/svg/statuses/incapacitated.svg",
-    reference: "Compendium.dnd5e.content24.JournalEntry.phbAppendixCRule.JournalEntryPage.4i3G895hy99piand"
+    reference: "Compendium.dnd5e.content24.JournalEntry.phbAppendixCRule.JournalEntryPage.4i3G895hy99piand",
+    neverBlockMovement: true
```
dev7355608, line 3968: We should document this in the `_ConditionConfiguration` typedef above.

module/canvas/layers/tokens.mjs
dev7355608: Do we need `t.actor?.statuses.has("dead")` consider `dead` is `neverBlockMovement: true` and therefore `mayHaveBlocked` is already true?

roth-michael: No, you're right. Vestigial code, I'll remove it.

module/canvas/ruler.mjs
roth-michael: If copying over the core code in here and in `_getWaypointLabelContext` is ideal, we can do it that way instead. Whichever is preferred.

module/canvas/ruler.mjs
roth-michael: I was also considering setting `context.cssClass` to `"blocked"` or something in `_getWaypointLabelContext` and then throwing some styling into our css. The label itself getting styled rather than just the waypoint feels to me more informative.

module/canvas/layers/tokens.mjs
roth-michael: Just to make sure I understand this properly:
This should probably be
```js
if ( (occupiedElevation < lowerElevation) || (occupiedElevation >= upperElevation) ) return false;
``` 
right? And we're assuming earlier that each token takes up 1 grid space of elevation? We could I suppose make the assumption that width or height is equal to a token's "elevation height," though I'm happy to just give it one grid space to fudge with until there's a core concept of token "height" in the elevation axis.

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  /** @inheritdoc */
+  _getWaypointLabelContext(waypoint, state) {
+    let modifiedWaypoint = waypoint;
+    if ( !waypoint.explicit && !waypoint.unreachable && waypoint.next?.unreachable ) {
+      modifiedWaypoint = { ...waypoint, explicit: true };
+    }
+    const context = super._getWaypointLabelContext(modifiedWaypoint, state);
+    if ( waypoint !== modifiedWaypoint ) context.cssClass = "blocked";
```
dev7355608: CSS definition for `blocked` seems to be missing from this PR.

module/canvas/ruler.mjs
dev7355608: Let's copy it over then. I'll keep track of changes in to so that we can make the same here. Or remove the code entirely if we make this a core behavior (https://github.com/foundryvtt/foundryvtt/issues/13139).

With these changes I suggested, we apply this styling to any nonexplicit waypoint that is rendered. Core actually renders some: if the movement action changes during keyboard movement for example.

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,23 @@
...
+  /** @inheritdoc */
+  _getWaypointStyle(waypoint) {
+    let modifiedWaypoint = waypoint;
+    if ( !waypoint.explicit && !waypoint.unreachable && waypoint.next?.unreachable ) {
+      modifiedWaypoint = { ...waypoint, explicit: true };
+    }
+    const style = super._getWaypointStyle(modifiedWaypoint);
+    if ( waypoint !== modifiedWaypoint ) style.alpha = (style.alpha ?? 1) * 0.5;
+    return style;
+  }
```
dev7355608: I think this should be it:

```suggestion
  _getWaypointStyle(waypoint) {
    if ( !waypoint.explicit && waypoint.next && waypoint.previous && waypoint.actionConfig.visualize
      && waypoint.next.actionConfig.visualize && (waypoint.action === waypoint.next.action)
      && (waypoint.unreachable || !waypoint.next.unreachable) ) return {radius: 0};
    const user = game.users.get(waypoint.userId);
    const scale = canvas.dimensions.uiScale;
    return {radius: 6 * scale, color: user?.color ?? 0x000000, alpha: waypoint.explicit ? 1 : 0.5};
  }
```

Maybe the 50% doesn't look so good. If the styling of the waypoint label is sufficiently clear, maybe we don't need the reduced transparency here.

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,23 @@
...
+  /** @inheritdoc */
+  _getWaypointLabelContext(waypoint, state) {
+    let modifiedWaypoint = waypoint;
+    if ( !waypoint.explicit && !waypoint.unreachable && waypoint.next?.unreachable ) {
+      modifiedWaypoint = { ...waypoint, explicit: true };
+    }
+    const context = super._getWaypointLabelContext(modifiedWaypoint, state);
+    if ( waypoint !== modifiedWaypoint ) context.cssClass = "blocked";
+    return context;
+  }
```
dev7355608: I think this should be it:

```suggestion
  /** @override */
  _getWaypointLabelContext(waypoint, state) {
    const {index, elevation, explicit, next, previous, ray} = waypoint;
    state.hasElevation ||= (elevation !== 0);
    if ( !previous ) {
      state.previousElevation = elevation;
      return;
    }
    if ( !explicit && next && waypoint.actionConfig.visualize && next.actionConfig.visualize
      && (waypoint.action === next.action) && (waypoint.unreachable || !waypoint.next.unreachable) ) return;
    if ( (ray.distance === 0) && (elevation === previous.elevation) ) return;

    // Prepare data structure
    const context = {
      action: waypoint.actionConfig,
      cssClass: [
        waypoint.hidden ? "secret" : "",
        waypoint.next ? "" : "last",
        explicit ? "" : "nonexplicit"
      ].filterJoin(" "),
      secret: waypoint.hidden,
      units: canvas.grid.units,
      uiScale: canvas.dimensions.uiScale,
      position: {x: ray.B.x, y: ray.B.y + (next ? 0 : 0.5 * this.token.h) + (16 * canvas.dimensions.uiScale)}
    };

    // Segment Distance
    context.distance = {total: waypoint.measurement.distance.toNearest(0.01).toLocaleString(game.i18n.lang)};
    if ( index >= 2 ) context.distance.delta = waypoint.measurement.backward.distance.toNearest(0.01).signedString();

    // Segment Cost
    const cost = waypoint.measurement.cost;
    const deltaCost = waypoint.cost;
    context.cost = {total: Number.isFinite(cost) ? cost.toNearest(0.01).toLocaleString(game.i18n.lang) : "∞", units: canvas.grid.units};
    if ( index >= 2 ) context.cost.delta = Number.isFinite(deltaCost) ? deltaCost.toNearest(0.01).signedString() : "∞";

    // Elevation
    const deltaElevation = elevation - state.previousElevation;
    context.elevation = {total: elevation, icon: "fa-solid fa-arrows-up-down", hidden: !state.hasElevation};
    if ( deltaElevation !== 0 ) context.elevation.delta = deltaElevation.signedString();
    state.previousElevation = elevation;

    return context;
  }
```

Using more general `nonexplicit` instead of `blocked` CSS class name.

module/canvas/token.mjs
dev7355608: We need to make this change and remove the override.

module/canvas/token.mjs
```diff
@@ -18,6 +18,96 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+  /** @inheritdoc */
+  _getMovementCostFunction(options) {
+    const costFunction = super._getMovementCostFunction(options);
+    const disableAutomationKey = foundry.helpers.interaction.KeyboardManager.MODIFIER_KEYS.ALT;
+    if ( !game.settings.get("dnd5e", "movementAutomation")
+      || game.keyboard.isModifierActive(disableAutomationKey) ) return costFunction;
+
+    const ignoredDifficultTerrain = this.actor?.system.attributes.movement.ignoredDifficultTerrain ?? new Set();
+    const ignoreDifficult = ["all", "nonmagical"].some(i => ignoredDifficultTerrain.has(i));
+    const preview = options?.preview && !game.user.isGM;
```
dev7355608:
```suggestion
    const preview = options?.preview && canvas.visibility.tokenVision && !game.user.isGM;
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+   */
+  #getRelevantOccupyingTokens(gridSpace, token, {preview=false}={}) {
+    const grid = canvas.grid;
+    if ( grid.isGridless ) return [];
+    const topLeft = grid.getTopLeftPoint(gridSpace);
+    const rect = new PIXI.Rectangle(topLeft.x, topLeft.y, grid.sizeX, grid.sizeY);
+    const lowerElevation = gridSpace.k * grid.distance;
+    const upperElevation = (gridSpace.k + 1) * grid.distance;
+    const found = game.canvas.tokens.quadtree.getObjects(rect);
+    return found.filter(t => {
```
dev7355608: Instead of the `filter`, use the `collisionTest` optional parameter of `getObjects`. Just pass this filter function for `collisionTest` and change the signature of the function to `({t})`, because the `collisionTest` function is passed `QuadtreeObject`.
```js
/**
 * @typedef QuadtreeObject
 * @property {PIXI.Rectangle} r
 * @property {*} t
 * @property {Set<Quadtree>} [n]
 */
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,108 @@
...
+      // Ignore self
+      if ( t === token ) return false;
+
```
dev7355608, line 89:
```suggestion

      // Always ignore hidden tokens
      if ( t.document.hidden ) return false;

```

dev7355608 (changes requested):

module/canvas/ruler.mjs
roth-michael: Yeah, I figured I'd leave that piece of the implementation to the style pros, for now just ensuring that it has the class in place.

dnd5e.mjs
```diff
@@ -207,6 +208,14 @@ Hooks.once("init", function() {
...
   // Exhaustion handling
   documents.ActiveEffect5e.registerHUDListeners();
+
+  // Custom movement cost aggregator
+  CONFIG.Token.movement.costAggregator = (results, distance, segment) => {
+    return Math.max(...results.map(i => i.cost));
+  };
+
+  // Custom token ruler
+  CONFIG.Token.rulerClass = canvas.TokenRuler5e;
```
arbron: I'd suggest moving this up to where we adjust the other token config options, like after we set `CONFIG.Token.objectClass`.

lang/en.json
```diff
@@ -5037,6 +5039,10 @@
       "Hint": "Replaces all reference to lbs with kgs and updates the encumbrance calculations to use metric weight units."
     }
   },
+  "MOVEMENTAUTOMATION": {
```
arbron: I am thinking we might eventually want to move the various automation related options into a sub-app, so I would suggest renaming these `"SETTINGS.DND5E.AUTOMATION.Movement"` (then at some point the advancement automation and concentration automation settings can be renamed to match, though no need to do that now).

I'm also wondering if there might eventually be other movement-related settings that we add, such as restricting movement during combat if you have exceeded your move speed. In that case, it might be nice to have this be a more specific name (like `AUTOMATION.TokenBlocking`).

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,112 @@
...
+      const occupiedSize = CONFIG.DND5E.actorSizes[t.actor?.system.traits.size]?.numerical ?? 2;
+      // In modern rules, Tiny creatures can be moved through
+      if ( modernRules && (occupiedSize === 0) ) return false;
+
+      // Halfling Nimbleness means no larger creature can block
+      if ( halflingNimbleness && (occupiedSize > tokenSize) ) return false;
+
+      // A size difference of less than 2 should block (adjust for Halfling Nimbleness)
+      return (Math.abs(tokenSize - occupiedSize) < 2);
```
arbron:
```suggestion
      return Math.abs(tokenSize - occupiedSize) < 2;
```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,112 @@
...
+    });
+  }
+
```
arbron, line 34:
```suggestion

  /* -------------------------------------------- */

```

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,112 @@
...
+      // Tokens which may normally have blocked movement should still be considered difficult terrain
+      const mayHaveBlocked = neverBlockStatuses.some(status => t.actor?.statuses.has(status));
+
+      // Friendly means legacy, therefore difficult. A size difference of 2 or more is difficult terrain regardless of
+      // ruleset. Toknes which would have blocked if not for a certain status should be difficult.
+      return friendlyToken || (Math.abs(tokenSize - occupiedSize) >= 2) || mayHaveBlocked;
+    });
+  }
+
```
arbron, line 63:
```suggestion

  /* -------------------------------------------- */

```

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,56 @@
...
+  /** @inheritdoc */
+  _getWaypointStyle(waypoint) {
+    if ( !waypoint.explicit && waypoint.next && waypoint.previous && waypoint.actionConfig.visualize
+      && waypoint.next.actionConfig.visualize && (waypoint.action === waypoint.next.action)
+      && (waypoint.unreachable || !waypoint.next.unreachable) ) return {radius: 0};
+    const user = game.users.get(waypoint.userId);
+    const scale = canvas.dimensions.uiScale;
+    return {radius: 6 * scale, color: user?.color ?? 0x000000, alpha: waypoint.explicit ? 1 : 0.5};
+  }
+
```
arbron, line 11:
```suggestion

  /* -------------------------------------------- */

```

module/canvas/ruler.mjs
```diff
@@ -0,0 +1,56 @@
+export default class TokenRuler5e extends foundry.canvas.placeables.tokens.TokenRuler {
+  /** @inheritdoc */
```
arbron:
```suggestion
  /** @inheritDoc */
```

module/canvas/token.mjs
```diff
@@ -18,6 +18,94 @@ export default class Token5e extends foundry.canvas.placeables.Token {
 
   /* -------------------------------------------- */
 
+  /** @inheritdoc */
```
arbron:
```suggestion
  /** @inheritDoc */
```

module/canvas/token.mjs
```diff
@@ -18,6 +18,94 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+  _getDragConstrainOptions() {
+    const unconstrainedMovement = game.user.isGM
+      && ui.controls.controls.tokens.tools.unconstrainedMovement.active;
+    return { ...super._getDragConstrainOptions(), ignoreTokens: unconstrainedMovement };
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
```
arbron:
```suggestion
  /** @inheritDoc */
```

module/canvas/token.mjs
```diff
@@ -18,6 +18,94 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+      if ( !this.layer.isOccupiedGridSpaceDifficult(to, this, {preview}) ) return cost;
+
+      // Difficult terrain due to occupied grid space
+      return cost + distance;
+    };
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritdoc */
```
arbron:
```suggestion
  /** @inheritDoc */
```

module/config.mjs
```diff
@@ -1223,6 +1223,7 @@ preLocalize("activityConsumptionTypes", { key: "label" });
  * @property {number} hitDie                  Default hit die denomination for NPCs of this size.
  * @property {number} [token=1]               Default token size.
  * @property {number} [capacityMultiplier=1]  Multiplier used to calculate carrying capacities.
+ * @property {number} numerical               Numerical representation of size
```
arbron:
```suggestion
 * @property {number} numerical               Numerical representation of size.
```

dnd5e.mjs
roth-michael: Should the same be done for the custom movement cost aggregator?

lang/en.json
roth-michael: Would it make sense then to also change the registered setting from `dnd5e.movementAutomation` to `dnd5e.tokenBlocking` or similar?

arbron: Yeah, that would be reasonable.

dnd5e.mjs
arbron: I feel that is good where it is because it's a whole function being added, but once my other PR is merged we can keep this and my action cost function call grouped together.

module/settings.mjs
```diff
@@ -76,6 +76,16 @@ export function registerSystemSettings() {
...
 
+  // Movement automation
+  game.settings.register("dnd5e", "tokenBlocking", {
+    name: "SETTINGS.DND5E.AUTOMATION.TokenBlocking.Name",
+    hint: "SETTINGS.DND5E.AUTOMATION.TokenBlocking.Hint",
+    scope: "world",
+    config: true,
+    default: true,
+    type: Boolean
+  });
```
dev7355608: Perhaps we make this a setting with three options:
- Tokens neither apply difficult terrain nor block movement
- Tokens can apply difficult terrain but never block movement
- Tokens can apply difficult terrain and block movement

roth-michael: On this note, actually, something I've been pondering. Should `isOccupiedGridSpaceDifficult` return true for both difficult-terrain tokens _and_ blocking tokens? i.e. should it represent "at least difficult," or "precisely difficult"? Currently it's the latter, because precision felt appropriate when making a function which might be called outside the precise context of what we're doing here, but it would make implementing the "difficult only, no blocking" setting a bit easier to implement if it were "at least difficult."

dev7355608: I think "at least difficult" is correct considering that, if a hostile creature allowed you pass through its space, it would be difficult terrain for you.

roth-michael: Could probably do with better settings i18n, but figured I'd opt for overly-descriptive over unclear.

module/canvas/token.mjs
```diff
@@ -18,6 +18,96 @@ export default class Token5e extends foundry.canvas.placeables.Token {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _getMovementCostFunction(options) {
+    const costFunction = super._getMovementCostFunction(options);
+    if ( game.settings.get("dnd5e", "tokenBlocking") === "none" ) return costFunction;
+
+    const ignoredDifficultTerrain = this.actor?.system.attributes.movement.ignoredDifficultTerrain ?? new Set();
```
arbron: Might want to protect against actors that don't have movement data.

```suggestion
    const ignoredDifficultTerrain = this.actor?.system.attributes?.movement?.ignoredDifficultTerrain ?? new Set();
```

arbron (changes requested):

module/canvas/layers/tokens.mjs
```diff
@@ -0,0 +1,106 @@
...
+        if ( t === token ) return false;
+
+        // Always ignore hidden tokens
+        if ( t.document.hidden ) return false;
+
+        // If preview movement, don't reveal blocked or difficult terrain for non-visible tokens
+        if ( preview && !t.visible ) return false;
+
+        // Always ignore secret tokens
+        if ( t.document.disposition === CONST.TOKEN_DISPOSITIONS.SECRET ) return false;
```
krbz999, line 94: there's an isSecret getter for this

dev7355608: In this case `isSecret` would be wrong (it's not equivalent to that): using `isSecret` here would mean that if the GM is moving the player token the secret token counts as difficult terrain / blocks movement but it would be ignored completely if a player moved the token. Whether something is difficult terrain or block movement is must be the same for all users.

Definition of `isSecret`:
> Does this TokenDocument have the SECRET disposition and is the current user lacking the necessary permissions that would reveal this secret?

Fyorl (approved):
Impressive stuff

Kapuzenjoe:
I think there's still a bug here.

If the moved token has a higher elevation, then it should go through larger targets and not be blocked. 
When analyzing the code, I noticed that the function for `const occupiedGridSpaces = this.document.getOccupiedGridSpaceOffsets(waypoint);` doesn't get a k value, which means the elevation ` const lowerElevation = gridSpace.k * grid.distance;` isn't calculated. (for the difficult terrain part this is calculated correctly)


Is this intended?

roth-michael:
Yep seems we missed that, thank you. I'll open up another PR taking elevation into account for blocking.

Kapuzenjoe:
@roth-michael Hi, I know 5.1 isn't out yet, but I've noticed another bug. Group tokens can no longer be moved, and you get an error message when trying to.

> Uncaught TypeError: Cannot read properties of undefined (reading 'size')

The cause is probably the `isOccupiedGridSpaceBlocking()` function, since it asks for the `traits.size`, which group tokens don't have. 
I think group tokens should be excluded altogether and don't require blocking.

However, there is another problem: the new group token has unknown movement modes and is therefore also not movable, but I suspect the new group token is simply not fully developed yet.

PR #5786, opened 2025-06-23
Update magic items in legacy SRD to use XGE pricing

krbz999:
Updated all magic items in the `dnd5e.items` pack if
- they had a rarity
- `system.properties.has("mgc")`
- The price is halved for consumables.

Had to make a few judgement calls on what a "consumable" really is. Potions, oils, and ammunition were obvious, and I also opted for including anything that has a limited number of charges and then automatically get destroyed. I did not include things that have a *chance* to be destroyed, for example the Horn of Blasting, or other items where you can use them in different ways that also destroy them. Similarly, did not opt for including Decks.

Otherwise used this to help decide - the mapping of rarity to price is from XGE, which also states that consumables' prices are halved ("like a potion or scroll").

```js
const key = "dnd5e.items";
const pack = game.packs.get(key);
const items = await pack.getDocuments();

const pricing = {
  common: 100,
  uncommon: 400,
  rare: 4_000,
  veryRare: 40_000,
  legendary: 200_000,
};

const isConsumable = async (item) => {
  if (item.type !== "consumable") return false;
  return foundry.applications.api.Dialog.confirm({ content: `<p>Is consumable? ${item.toAnchor().outerHTML}</p>` });
};

const updates = [];

for (const item of items) {
  let price = pricing[item.system.rarity] ?? null;
  if (!price) continue;
  if (!item.system.properties?.has("mgc")) continue;

  const consumable = await isConsumable(item);
  if (consumable) price /= 2;

  updates.push({ _id: item.id, "system.price": { value: price, denomination: "gp" } });
}

await Item.updateDocuments(updates, { pack: key });
```

This was done on v13 and the 5.1 branch, not aiming for legacy 4.4.


Fyorl:
> Had to make a few judgement calls on what a "consumable" really is.

My own interpretation of 'consumable' in this context is anything that is consumed *immediately* upon use, but I agree that the rules are not particularly explicit. I'm ok with the interpretation used here too. We can always change it later if there's some more compelling arguments one way or the other.

krbz999:
> My own interpretation of 'consumable' in this context is anything that is consumed immediately upon use

That's pretty much what I went with.

Fyorl:
> That's pretty much what I went with.

That's not what I understood from this:

> I also opted for including anything that has a limited number of charges and then automatically get destroyed

That would be an item that is *eventually* destroyed after its last charge is used, not one that is *immediately* destroyed when used. But I can't find anywhere the rules offer an explicit definition of 'consumable' for the purposes of price/crafting costs, so I am fine with this interpretation until there's a compelling argument otherwise.

Fyorl (approved):

krbz999:
> That's not what I understood from this:

To clarify, I mean stuff like multi-use items that get destroyed when expending the last charge 100% of the time, specifically excluding things like... a wand that has a 20% chance to explode in your face when you expend its last charge. Those latter ones I do not interpret as "consumable".

So a potion with 1 use, or a gem that turns nonmagical after you use it once: Consumable.

Or in short, using a definition of '*not* consumable' to be: Has a chance to be reusable despite an arbitrary number of uses expended over any given period of time. (Excluding decks.)

PR #5790, opened 2025-06-24
[#5740] Disable Compendium Browser's Submit button if fewer than min/more than max documents are selected

### LINKED ISSUE #5740: Upon selecting more than the allowed number in limited-selection Compendium Browser, disable other checkboxes or submit button
Currently, trying to submit with more than the allowed number of selections throws an error and doesn't proceed with the selection. This flow could be improved by either:
A. Disabling all non-checked checkboxes when the number of selected results is equal to the maximum number of selected results.
B. Disabling the "Submit" button itself in the same circumstance.

I see no potential drawbacks to B (and it can also disable in the event that too _few_ entries are selected, if desired).
For A, I can potentially see a use case where someone wants to check a number of "all the things I might want" options, then narrow down the selection from there, which wouldn't be possible in that instance.

roth-michael:
Closes #5740 
Made `invalid` a boolean instead of a string, since now it's being checked in a second spot and `{{#if (eq invalid "invalid")}}` felt wrong.
Can _probably_ get rid of most of `#onHandleSubmit`, since it shouldn't be possible to trigger that at a time when it'd throw an error, but left it in for now just in case.

### DIFF
--- a/module/applications/compendium-browser.mjs
+++ b/module/applications/compendium-browser.mjs
@@ -481,11 +481,16 @@ export default class CompendiumBrowser extends Application5e {
     const { max, min } = this.options.selection;
 
     context.displaySelection = this.displaySelection;
-    context.invalid = (value < (min || -Infinity)) || (value > (max || Infinity)) ? "invalid" : "";
+    context.invalid = (value < (min || -Infinity)) || (value > (max || Infinity));
     const suffix = this.#selectionLocalizationSuffix;
     context.summary = suffix ? game.i18n.format(
       `DND5E.CompendiumBrowser.Selection.Summary.${suffix}`, { max, min, value }
     ) : value;
+    const pr = new Intl.PluralRules(game.i18n.lang);
+    context.invalidTooltip = game.i18n.format(`DND5E.CompendiumBrowser.Selection.Warning.${suffix}`, {
+      max, min, value,
+      document: game.i18n.localize(`DND5E.CompendiumBrowser.Selection.Warning.Document.${pr.select(max || min)}`)
+    });
     return context;
   }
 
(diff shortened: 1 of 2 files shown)


module/applications/compendium-browser.mjs
```diff
@@ -481,11 +481,16 @@ export default class CompendiumBrowser extends Application5e {
     const { max, min } = this.options.selection;
 
     context.displaySelection = this.displaySelection;
-    context.invalid = (value < (min || -Infinity)) || (value > (max || Infinity)) ? "invalid" : "";
+    context.invalid = (value < (min || -Infinity)) || (value > (max || Infinity));
     const suffix = this.#selectionLocalizationSuffix;
     context.summary = suffix ? game.i18n.format(
       `DND5E.CompendiumBrowser.Selection.Summary.${suffix}`, { max, min, value }
     ) : value;
+    const pr = new Intl.PluralRules(game.i18n.lang);
```
arbron: We're trying to reply on the utility `getPluralRules` from `utils.mjs` these days (didn't exist when `#onHandleSubmit` was written), so you can import that and swap this out:

```suggestion
// Above: import { getPluralRules } from "../utils.mjs";
    const pr = getPluralRules();
```

arbron (changes requested):
Just a minor change.

module/applications/compendium-browser.mjs
```diff
@@ -2,6 +2,7 @@ import * as Filter from "../filter.mjs";
 import SourceField from "../data/shared/source-field.mjs";
 import Application5e from "./api/application.mjs";
 import CompendiumBrowserSettingsConfig from "./settings/compendium-browser-settings.mjs";
+import { getPluralRules } from "../utils.mjs";
```
arbron: Just a little tweak to import order.

```suggestion
import { getPluralRules } from "../utils.mjs";
import Application5e from "./api/application.mjs";
import CompendiumBrowserSettingsConfig from "./settings/compendium-browser-settings.mjs";
```

arbron (changes requested):

arbron (approved):

PR #5791, opened 2025-06-24
[#5680] Ensure `config.isCritical` matches `isCritical`s of roll options for damage rolls

### LINKED ISSUE #5680: The `dnd5e.postRollConfiguration` hook does not display whether a roll was critical
`dialog.isCritical` is always `false` or `undefined`.
The value can be found on each individual roll, but I think it makes sense that this should be found in one location?

roth-michael:
Closes #5680 
Modify `config.isCritical` in two places:
1. `DamageRoll.applyKeybindings`: If advantage key held, `config.isCritical` is `true`, but if disadvantage key held, it's `false`
2. `DamageRollConfigurationDialog#_finalizeRolls`: If critical is clicked, `config.isCritical` is `true`, otherwise it's `false`
This ensures regardless of fast-forwarding, `config.isCritical` will always be accurate by the time it's passed in the `dnd5e.postRollConfiguration` hook

### DIFF
--- a/module/applications/dice/damage-configuration-dialog.mjs
+++ b/module/applications/dice/damage-configuration-dialog.mjs
@@ -81,8 +81,10 @@ export default class DamageRollConfigurationDialog extends RollConfigurationDial
 
   /** @override */
   _finalizeRolls(action) {
+    const isCritical = action === "critical";
+    this.config.isCritical = isCritical;
     return this.rolls.map(roll => {
-      roll.options.isCritical = action === "critical";
+      roll.options.isCritical = isCritical;
       roll.configureDamage({ critical: this.config.critical });
       return roll;
     });
--- a/module/applications/dice/damage-configuration-dialog.mjs
+++ b/module/applications/dice/damage-configuration-dialog.mjs
@@ -81,10 +81,9 @@ export default class DamageRollConfigurationDialog extends RollConfigurationDial
 
   /** @override */
   _finalizeRolls(action) {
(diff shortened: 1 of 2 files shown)


module/applications/dice/damage-configuration-dialog.mjs
```diff
@@ -81,8 +81,10 @@ export default class DamageRollConfigurationDialog extends RollConfigurationDial
 
   /** @override */
   _finalizeRolls(action) {
+    const isCritical = action === "critical";
+    this.config.isCritical = isCritical;
     return this.rolls.map(roll => {
-      roll.options.isCritical = action === "critical";
+      roll.options.isCritical = isCritical;
```
arbron: We can save this extra assignment by just applying `this.config.isCritical` to the roll options, since it will always be the same as `isCritical`.

```suggestion
      roll.options.isCritical = this.config.isCritical;
```

module/applications/dice/damage-configuration-dialog.mjs
```diff
@@ -81,8 +81,10 @@ export default class DamageRollConfigurationDialog extends RollConfigurationDial
 
   /** @override */
   _finalizeRolls(action) {
+    const isCritical = action === "critical";
+    this.config.isCritical = isCritical;
```
arbron:
```suggestion
    this.config.isCritical = action === "critical";
```

arbron (changes requested):

PR #5811, opened 2025-07-02
[#5300] Add feature prerequisites on other items

### LINKED ISSUE #5300: Item Prerequisites for other Items
Probably going to be blocked for a while, but it would be nice if when selecting items, additional restrictions can be set aside from level.

Certain newer feats (pre 5.5e / 2024 5e) require other feats, like `Soul of the Storm Giant` requiring 'Strike of the Giants' as well as the character being level 4.

arbron:
Adds a list of identifiers to feature prerequisites and require at least one of those other items to be present on the actor to take that feature using an Choose Items or ASI advancement.

To support this functionality `identifiedItems` has been added to actors along side `sourcedItems`. This groups items by their identifiers and optionally allows getting only items with a matching identifier and of a certain type.

Moves the prerequisite validation code into a new method on the feature item itself to avoid code duplication between the two advancement types.

Closes #5300

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -347,7 +347,6 @@
     },
     "Title": "Ability Score Improvement",
     "Warning": {
-      "Level": "Must be at least level {level} to take this feat.",
       "Type": "Only features with the \"feat\" type can be selected."
     }
   },
@@ -433,10 +432,8 @@
     },
     "Title": "Choose Items",
     "Warning": {
-      "FeatureLevel": "Must be at least level {level} to take this feature.",
       "InvalidType": "Only {type} items can be selected for this choice.",
       "NoOriginal": "Previously selected choice no longer available for replacement.",
-      "PreviouslyChosen": "This item has already been chosen at a previous level.",
       "SpellLevelAvailable": "Only {level} or lower spells can be chosen for this advancement.",
       "SpellLevelSpecific": "Only {level} spells can be chosen for this advancement.",
       "SpellList": "Only spells available on the {lists} spell list can be chosen for this advancement."
@@ -3008,16 +3005,26 @@
   "Header": "Feature Prerequisites",
   "FIELDS": {
     "prerequisites": {
+      "items": {
+        "hint": "Identifiers for items that the character must have before selecting this item.",
+        "label": "Required Items"
+      },
       "label": "Prerequisites",
       "level": {
-        "label": "Required Level",
-        "hint": "Character or class level required to select this feature when levelling up."
+        "hint": "Character or class level required to select this feature when levelling up.",
+        "label": "Required Level"
       },
       "repeatable": {
         "hint": "This feature can be chosen more than once.",
         "label": "Repeatable"
       }
     }
+  },
+  "Warning": {
+    "InvalidLevel": "must be at least level {level}",
+    "Message": "{actor} {requirements} in order to take this {type}.",
+    "MissingItem": "must have previously taken {items}",
+    "NotRepeatable": "must not have taken this before"
   }
 },
 "DND5E.Price": "Price",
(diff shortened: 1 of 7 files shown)


JPMeehan:
Is this intended to work as a way for path-based choice advancements to work? For example, could this be used to tie together subclasses that have you pick a damage type at 3 and then associate at later levels?

Zanderaf:
Long shot here, but is there any items that require multiple features / perquisite items to all be present instead of just one of them?

Zanderaf:
> Is this intended to work as a way for path-based choice advancements to work? For example, could this be used to tie together subclasses that have you pick a damage type at 3 and then associate at later levels?

I might be onto nothing but I think this mainly feels like something akin to the "Giant" feats from Bigby's where one feat is required for another in order to take it. Or how Pact of the Blade is required for some specific Warlock Invocations. I guess what you asked is about the million dragon themed subclasses / similar stuff where you pick a damage type and then roll with it for the whole progression?

JPMeehan:
Yeah, I get the intent is stuff like the new feat chains, but I'm wondering if this is also a way to handle the linked choices people keep asking about.

arbron:
> Long shot here, but is there any items that require multiple features / perquisite items to all be present instead of just one of them?

All of the examples I could find in official content either required only a single item (e.g. "Thirsting Blade" requires "Pact of the Blade") or one of multiple (e.g. "Initiate of High Sorcery" requires either Sorcerer class, Wizard class, or High Sorcery background).

If you have any 1st party examples that require multiple items that would be useful.

arbron:
> Is this intended to work as a way for path-based choice advancements to work? For example, could this be used to tie together subclasses that have you pick a damage type at 3 and then associate at later levels?

Not generally what this is designed for, but you could probably hack it together to work for that use case. One way to better support that would be to add the option to enforce these restrictions on Item Grant. Then you would be able to add an grant items advancement that has a separate feature for each damage type, but only one would appear that matches with the type selected at an earlier level. That would be a separate feature request to file after this one is done.

Zanderaf:
> > Long shot here, but is there any items that require multiple features / perquisite items to all be present instead of just one of them?
> 
> All of the examples I could find in official content either required only a single item (e.g. "Thirsting Blade" requires "Pact of the Blade") or one of multiple (e.g. "Initiate of High Sorcery" requires either Sorcerer class, Wizard class, or High Sorcery background).
> 
> If you have any 1st party examples that require multiple items that would be useful.

Not fully sure on any 1st party stuff but I'll give it a solid try once I have some free time after work to do a dive into it, you've covered most of the examples already with that though. You might be fully right on the end, ignoring any 3rd party premium content, I don't remember any off the top of my head.

Zanderaf:
> > Long shot here, but is there any items that require multiple features / perquisite items to all be present instead of just one of them?
> 
> All of the examples I could find in official content either required only a single item (e.g. "Thirsting Blade" requires "Pact of the Blade") or one of multiple (e.g. "Initiate of High Sorcery" requires either Sorcerer class, Wizard class, or High Sorcery background).
> 
> If you have any 1st party examples that require multiple items that would be useful.

So- I did find something:
Eldritch Adept from Tasha's lets you pick a warlock invocation when you take it but theres a part that mentions "you can't take invocations that have a prereq. unless you're a warlock that meets that prereq."

Theres a case here for "Agonizing Blast" and other "Eldritch Blast Invocations" that requires you to have the Eldritch Blast cantrip and be a Warlock, so both that cantrip and the warlock class would be needed to take it.

Fyorl (approved):
Should be easy enough to add some extra configuration at a future point for whether all of the items are required rather than just any of them, if it becomes necessary.

Issue #5672, opened 2025-05-14
Rework journal entry and journal entry pages for ApplicationV2


arbron:
- [x] `JournalSheet5e`
- [ ] `JournalClassPageSheet`
- [x] `JournalMapLocationPageSheet`
- [x] `JournalRulePageSheet`
- [ ] `JournalSpellListPageSheet`

arbron:
> Category Support 🙏

Presumably that would come automatically when moving `JournalEntry5e` to use the V2 journal entry as a base.

PR #5812, opened 2025-07-03
[#5672] Convert spell list page to ApplicationV2

Fyorl:


module/applications/journal/spells-page-sheet.mjs
```diff
@@ -244,6 +283,30 @@ export default class JournalSpellListPageSheet extends foundry.appv1.sheets.Jour
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Handle editing an unlinked spell.
+   * @this {JournalSpellListPageSheet}
+   * @param {PointerEvent} event  The originating event.
+   * @param {HTMLElement} target  The action target.
+   */
+  static #onEditUnlinked(event, target) {
```
arbron, line 272: Our general pattern for V2 apps is to separate the drag & drop handling into a separate section, so maybe put these two handlers above `_onDrop` and add in a heading:

```javascript
  /*  Drag & Drop                                 */
```

templates/journal/spell/config.hbs
```diff
@@ -0,0 +1,37 @@
...
+            {{ localize "JOURNALENTRYPAGE.DND5E.SpellList.Grouping.Hint" }}
+        </p>
+    </div>
+
+    <div class="form-group">
+        <label>{{ localize "DND5E.Description" }}</label>
+    </div>
+
+    <prose-mirror name="system.description.value" value="{{ system.description.value }}"
+                  {{#if isFirstRender}}toggled{{/if}}></prose-mirror>
```
arbron: Is this part of the core bug mentioned above?

It does lead to strange behavior where if you click the save button when first opening the sheet before making any other changes it will disable editing on the `<prose-mirror>` element, but not display any contents since we aren't providing the enriched description.

module/applications/journal/spells-page-sheet.mjs
```diff
@@ -3,30 +3,52 @@ import Items5e from "../../data/collection/items-collection.mjs";
...
+    position: {
+      width: 700
+    },
+    displayAsTable: false,
+    embedRendering: false,
+    grouping: null,
+    actions: {
+      "add-unlinked": JournalSpellListPageSheet.#onAddUnlinked,
+      delete: JournalSpellListPageSheet.#onDelete,
+      "edit-unlinked": JournalSpellListPageSheet.#onEditUnlinked
```
arbron: We use camel-cased action names everywhere else, so it is probably better to just adjust the template to match.

```suggestion
      addUnlinked: JournalSpellListPageSheet.#onAddUnlinked,
      delete: JournalSpellListPageSheet.#onDelete,
      editUnlinked: JournalSpellListPageSheet.#onEditUnlinked
```

templates/journal/spell/list.hbs
```diff
@@ -0,0 +1,34 @@
+<div>
+    <fieldset class="table card items">
+        <legend>{{ localize "TYPES.Item.spellPl" }}</legend>
+        <div class="header">
+            <button type="button" class="unbutton control-button" data-action="add-unlinked" data-tooltip
```
arbron:
```suggestion
            <button type="button" class="unbutton control-button" data-action="addUnlinked" data-tooltip
```

templates/journal/spell/list.hbs
```diff
@@ -0,0 +1,34 @@
...
+                <i class="fas fa-plus" inert></i>
+            </button>
+        </div>
+        <ol class="items-list">
+            {{#each spells}}
+            <li class="item flexrow" data-item-uuid="{{ spell.uuid }}" data-unlinked-id="{{ unlinked._id }}">
+                <div class="item-name">{{{ display }}}</div>
+                <div class="item-controls flexrow">
+                    {{#if unlinked}}
+                    <a class="item-control" data-action="edit-unlinked" data-tooltip
```
arbron:
```suggestion
                    <a class="item-control" data-action="editUnlinked" data-tooltip
```

templates/journal/spell/config.hbs
Fyorl: Yeah it's only necessary because of this bug https://github.com/foundryvtt/foundryvtt/issues/13143

Maybe there's a better workaround.

PR #5828, opened 2025-07-10
Add piety score

krbz999:
Related: https://github.com/foundryvtt/dnd5e/issues/2466

Adding piety data from the 'Mythic Odysseys of Theros' book.

- Adds a new setting, `pietyScore`.
- Adds `attributes.piety` on characters.
- Moves Loyalty and Piety into the 'Variant Rules' settings menu.

image

image

The initial value of Piety is 1. I was not able to find anything wrt whether piety can go below 1, though it *can* decrease. So I set it to be min 1 and non-nullable.

Like discussed on the relevant issue, it doesn't have a max and is not used for rolls, so it is not implemented as an ability score.

TODO and further ideas:
- [ ] Render on character sheets. (Would be very happy to just hand this off - I have been thinking either showing it similarly to loyalty is shown on NPC sheets, or as a badge next to XP / level, or a progress bar below the avatar showing your progress towards the thresholds).
- [ ] Extend `/award` enricher to grant piety.
- [x] Add feature subtype. I am not certain 'Supernatural Gift' (with its subtypes) fits the "boons" that can be granted by piety (they also don't seem to have an explicit name, referred to both as gifts, traits, and boons), though I don't feel strongly. "Pious Boon" might be good, with subtypes being the deity it belongs to.
- [ ] Add to own section on actor sheet or show 'Piety Trait' as subtitle?
- [ ] Add way to access piety config on character sheet.

### DIFF
--- a/module/applications/actor/config/piety-config.mjs
+++ b/module/applications/actor/config/piety-config.mjs
@@ -1,0 +1,51 @@
+import BaseConfigSheet from "../api/base-config-sheet.mjs";
+
+/**
+ * Configuration application for an actor's piety attribute.
+ */
+export default class PietyConfig extends BaseConfigSheet {
+  /** @override */
+  static DEFAULT_OPTIONS = {
+    position: {
+      width: 420
+    }
+  };
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static PARTS = {
+    config: {
+      template: "systems/dnd5e/templates/actors/config/piety-config.hbs"
+    }
+  };
+
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /** @override */
+  get title() {
+    return game.i18n.localize("DND5E.PIETY.Configure");
+  }
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preparePartContext(partId, context, options) {
+    context = await super._preparePartContext(partId, context, options);
+    const source = this.document.system._source;
+
+    context.data = source.attributes.piety;
+    context.fields = this.document.system.schema.getField("attributes.piety").fields;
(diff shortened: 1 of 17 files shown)


module/applications/actor/config/piety-config.mjs
```diff
@@ -0,0 +1,46 @@
...
+    }
+  };
+
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /** @override */
+  get title() {
+    return game.i18n.localize("DND5E.PIETY.Configure");
```
Fyorl:
```suggestion
    return _loc("DND5E.PIETY.Configure");
```

module/data/actor/character.mjs
```diff
@@ -77,7 +77,10 @@ export default class CharacterData extends CreatureTemplate {
             save: new FormulaField({ required: true, label: "DND5E.DeathSaveBonus" })
           })
         }, { label: "DND5E.DeathSave" }),
-        inspiration: new BooleanField({ required: true, label: "DND5E.Inspiration" })
+        inspiration: new BooleanField({ required: true, label: "DND5E.Inspiration" }),
+        piety: new SchemaField({
+          value: new NumberField({ min: 1, initial: 1, nullable: false, integer: true }),
```
Fyorl: For `getHumanReadableAttributeLabel`

```suggestion
          value: new NumberField({ min: 1, initial: 1, nullable: false, integer: true, label: "DND5E.PIETY.FIELDS.value.label" })
```

Also this should probably be `nullable: true, initial: null`. I don't think characters are guaranteed to be champions of a deity, and some can be Iconoclasts. But I only skimmed the rules.

Fyorl (approved):

module/data/actor/character.mjs
krbz999: Yes, that seems right. An iconoclast would gain features at certain *character* levels instead and explicitly has no piety value. There's also the oracle which has no deity but does have piety, but that doesn't seem relevant here.

PR #5838, opened 2025-07-14
[#4604] [#5408] Divorce spellcasting method from preparation state.

### LINKED ISSUE #4604: Simplify spellcasting CONFIG data and separate spell preparation modes from spell preparation states

### LINKED ISSUE #5408: [Bug] Warlock Spellcasting box doesn't treat Pack Magic spells as prepared
dnd5e Version: 4.3.6
Foundry Version: 12.331

When adding Pact spells to a Warlock character the Spellcasting summary box at the top of the sheet shows that "0/@scale.warlock.max-prepared" spells are prepared. I  suspect this is because spell preparation is being used to track the number of spells the Warlock is allowed to have, but spells configured as Pact Magic don't count as prepared in the summary box.

To Reproduce:
- Create a new character
- Add the Warlock class at level 1
- Add spells onto the Character sheet

Image

Fyorl:
- Closes #4604 
- Closes #5408 

Based on the most recent commit of yours I could find @krbz999 with additional refactoring and compatibility with the latest system changes.

- Separates the concept of a spellcasting method from its preparation state.
- Allows for definitions of arbitrary spellcasting methods with their own sub-progressions, or progressions based on character level rather than class level.
- Allows for spellcasting methods to indicate that their slots may only be used for spells that use the same method, or that spells that use the spellcasting method may only be cast with spells provided by that method.
- Moves spell slot progression and preparation logic out of `Actor5e` and into `SpellcastingModel`s.

### Breaking Changes
- `CONFIG.DND5E.spellPreparationModes` removed with deprecation.
- `CONFIG.DND5E.spellcastingTypes` removed with deprecation.
- `CONFIG.DND5E.spellProgression` read-only with no deprecation path.
- `SpellData#preparation#mode` deprecated in favour of `SpellData#method`.
- `SpellData#preparation#prepared` deprecated in favour of `SpellData#prepared`.
- `SpellConfigurationData#preparation` deprecated in favour of `SpellConfigurationData#method` and `SpellConfigurationData#prepared`.
- `Actor5e.computePactProgression` removed with deprecation.
- `Actor5e.computeLeveledProgression` removed with deprecation.
- `Actor5e.preparePactSlots` removed with deprecation.
- `Actor5e.prepareLeveledSlots` removed with deprecation.
- `Actor5e.prepareAltSlots` removed with deprecation.
- Spell item `labels#classes` type changed from `string` to `string[]`.

### DIFF
--- a/module/applications/actor/character-sheet.mjs
+++ b/module/applications/actor/character-sheet.mjs
@@ -721,9 +721,8 @@ export default class CharacterActorSheet extends BaseActorSheet {
   async _getFavoriteData(type, id) {
     // Spell slots
     if ( type === "slots" ) {
-      const [method] = id.match(/^\D+/) ?? [];
+      const { value, max, level, type: method } = this.actor.system.spells?.[id] ?? {};
       const model = CONFIG.DND5E.spellcasting[method];
-      const { value, max, level } = this.actor.system.spells[id] ?? {};
       const uses = { value, max, name: `system.spells.${id}.value` };
       if ( !model || model.isSingleLevel ) return {
         uses, level, method,
--- a/module/applications/actor/character-sheet.mjs
+++ b/module/applications/actor/character-sheet.mjs
@@ -697,7 +697,6 @@ export default class CharacterActorSheet extends BaseActorSheet {
         effectId: type === "effect" ? favorite.id : null,
         parentId: (type === "effect") && (favorite.parent !== favorite.target) ? favorite.parent.id: null,
         activityId: type === "activity" ? favorite.id : null,
-        preparationMode: (type === "slots") ? (/spell\d+/.test(id) ? "prepared" : id) : null,
         key: (type === "skill") || (type === "tool") ? id : null,
         toggle: toggle === undefined ? null : { applicable: true, value: toggle },
         quantity: quantity > 1 ? quantity : "",
@@ -722,25 +721,26 @@ export default class CharacterActorSheet extends BaseActorSheet {
   async _getFavoriteData(type, id) {
     // Spell slots
     if ( type === "slots" ) {
+      const [method] = id.match(/^\D+/) ?? [];
+      const model = CONFIG.DND5E.spellcasting[method];
       const { value, max, level } = this.actor.system.spells[id] ?? {};
       const uses = { value, max, name: `system.spells.${id}.value` };
-      if ( !/spell\d+/.test(id) ) return {
-        uses, level,
+      if ( !model || model.isSingleLevel ) return {
+        uses, level, method,
         title: game.i18n.localize(`DND5E.SpellSlots${id.capitalize()}`),
         subtitle: [
           game.i18n.localize(`DND5E.SpellLevel${level}`),
-          game.i18n.localize(`DND5E.Abbreviation${CONFIG.DND5E.spellcastingTypes[id]?.shortRest ? "SR" : "LR"}`)
+          game.i18n.localize(`DND5E.Abbreviation${model?.isSR ? "SR" : "LR"}`)
         ],
-        img: CONFIG.DND5E.spellcastingTypes[id]?.img || CONFIG.DND5E.spellcastingTypes.pact.img
+        img: model?.img || CONFIG.DND5E.spellcasting.pact.img
       };
 
       const plurals = new Intl.PluralRules(game.i18n.lang, { type: "ordinal" });
-      const isSR = CONFIG.DND5E.spellcastingTypes.leveled.shortRest;
       return {
-        uses, level,
+        uses, level, method,
         title: game.i18n.format(`DND5E.SpellSlotsN.${plurals.select(level)}`, { n: level }),
-        subtitle: game.i18n.localize(`DND5E.Abbreviation${isSR ? "SR" : "LR"}`),
-        img: CONFIG.DND5E.spellcastingTypes.leveled.img.replace("{id}", id)
+        subtitle: game.i18n.localize(`DND5E.Abbreviation${model.isSR ? "SR" : "LR"}`),
+        img: model.img.replace("{id}", id)
       };
     }
 
@@ -1123,14 +1123,14 @@ export default class CharacterActorSheet extends BaseActorSheet {
 
   /** @inheritDoc */
   _onDragStart(event) {
-    const modes = CONFIG.DND5E.spellPreparationModes;
+    const methods = CONFIG.DND5E.spellcasting;
     const { key } = event.target.closest("[data-key]")?.dataset ?? {};
-    const { level, preparationMode } = event.target.closest("[data-level]")?.dataset ?? {};
+    const { level, method } = event.target.closest("[data-level]")?.dataset ?? {};
     const isSlots = event.target.closest("[data-favorite-id]") || event.target.classList.contains("items-header");
     let type;
     if ( key in CONFIG.DND5E.skills ) type = "skill";
     else if ( key in CONFIG.DND5E.tools ) type = "tool";
-    else if ( modes[preparationMode]?.upcast && (level !== "0") && isSlots ) type = "slots";
+    else if ( methods[method]?.slots && (level !== "0") && isSlots ) type = "slots";
     if ( !type ) return super._onDragStart(event);
 
     // Add another deferred deactivation to catch the second pointerenter event that seems to be fired on Firefox.
@@ -1138,7 +1138,7 @@ export default class CharacterActorSheet extends BaseActorSheet {
     game.tooltip.deactivate();
 
     const dragData = { dnd5e: { action: "favorite", type } };
-    if ( type === "slots" ) dragData.dnd5e.id = (preparationMode === "prepared") ? `spell${level}` : preparationMode;
+    if ( type === "slots" ) dragData.dnd5e.id = methods[method].getSpellSlotKey(Number(level));
     else dragData.dnd5e.id = key;
     event.dataTransfer.setData("application/json", JSON.stringify(dragData));
     event.dataTransfer.effectAllowed = "link";
--- a/module/applications/api/primary-sheet-mixin.mjs
+++ b/module/applications/api/primary-sheet-mixin.mjs
@@ -503,8 +503,8 @@ export default function PrimarySheetMixin(Base) {
     static sortItemsPriority(a, b) {
       return a.system.linkedActivity?.item?.name.localeCompare(b.system.linkedActivity?.item?.name, game.i18n.lang)
         || ((a.system.level ?? 0) - (b.system.level ?? 0))
-        || (a.system.method ?? "").compare(b.system.method ?? "")
         || ((a.system.prepared ?? 0) - (b.system.prepared ?? 0))
+        || (a.system.method ?? "").compare(b.system.method ?? "")
         || a.name.localeCompare(b.name, game.i18n.lang);
     }
 
--- a/module/applications/api/primary-sheet-mixin.mjs
+++ b/module/applications/api/primary-sheet-mixin.mjs
@@ -503,8 +503,8 @@ export default function PrimarySheetMixin(Base) {
     static sortItemsPriority(a, b) {
       return a.system.linkedActivity?.item?.name.localeCompare(b.system.linkedActivity?.item?.name, game.i18n.lang)
         || ((a.system.level ?? 0) - (b.system.level ?? 0))
-        || a.system.preparation?.mode?.localeCompare(b.system.preparation?.mode, "en")
-        || ((b.system.preparation?.prepared ?? 0) - (a.system.preparation?.prepared ?? 0))
+        || (a.system.method ?? "").compare(b.system.method ?? "")
+        || ((a.system.prepared ?? 0) - (b.system.prepared ?? 0))
         || a.name.localeCompare(b.name, game.i18n.lang);
     }
 
--- a/module/applications/components/inventory.mjs
+++ b/module/applications/components/inventory.mjs
@@ -397,12 +397,13 @@ export default class InventoryElement extends HTMLElement {
       callback: li => this._onAction(li, "toggleCharge"),
       group: "state"
     }, {
-      name: `DND5E.ContextMenuAction${item.system.preparation?.prepared ? "Unprepare" : "Prepare"}`,
+      name: `DND5E.ContextMenuAction${item.system.prepared ? "Unprepare" : "Prepare"}`,
       icon: '<i class="fa-solid fa-sun fa-fw"></i>',
       condition: () => {
-        const isPrepared = ("preparation" in item.system) && (item.system.preparation.mode === "prepared");
+        const isPrepared = CONFIG.DND5E.spellcasting[item.system.method]?.prepares;
+        const isAlways = item.system.prepared === CONFIG.DND5E.spellPreparationStates.always.value;
         const canEdit = item.isOwner && !compendiumLocked;
-        return !item.hasRecharge && isPrepared && canEdit && !item.getFlag("dnd5e", "cachedFor");
+        return !item.hasRecharge && isPrepared && !isAlways && canEdit && !item.getFlag("dnd5e", "cachedFor");
       },
       callback: li => this._onAction(li, "prepare"),
       group: "state"
@@ -459,7 +460,7 @@ export default class InventoryElement extends HTMLElement {
     const { itemId } = target.closest("[data-item-id]")?.dataset ?? {};
     const { activityId } = target.closest("[data-activity-id]")?.dataset ?? {};
     const item = await this.getItem(itemId);
-    if ( !item ) return;
+    if ( !item || (target.ariaDisabled === "true") ) return;
     const activity = item.system.activities?.get(activityId);
 
     switch ( action ) {
@@ -786,7 +787,7 @@ export default class InventoryElement extends HTMLElement {
    * @protected
    */
   _onTogglePrepared(item) {
-    return item.update({ "system.preparation.prepared": !item.system.preparation?.prepared });
+    return item.update({ "system.prepared": Boolean(!item.system.prepared) });
   }
 
   /* -------------------------------------------- */
--- a/module/applications/components/inventory.mjs
+++ b/module/applications/components/inventory.mjs
@@ -787,7 +787,7 @@ export default class InventoryElement extends HTMLElement {
    * @protected
    */
   _onTogglePrepared(item) {
-    return item.update({ "system.prepared": Boolean(!item.system.prepared) });
+    return item.update({ "system.prepared": Number(!item.system.prepared) });
   }
 
   /* -------------------------------------------- */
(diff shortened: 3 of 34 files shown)


krbz999:
Hell yeah.

module/applications/components/inventory.mjs
```diff
@@ -786,7 +787,7 @@ export default class InventoryElement extends HTMLElement {
    * @protected
    */
   _onTogglePrepared(item) {
-    return item.update({ "system.preparation.prepared": !item.system.preparation?.prepared });
+    return item.update({ "system.prepared": Boolean(!item.system.prepared) });
```
krbz999: Redundant double casting to bool?

module/data/active-effect/enchantment.mjs
```diff
@@ -88,6 +88,25 @@ export default class EnchantmentData extends foundry.abstract.TypeDataModel {
           );
         }
         return false;
+
+      /** @deprecated since 5.1 */
+      case "system.preparation.mode":
+        foundry.utils.logCompatibilityWarning("system.preparation.mode is deprecated. Please instead use "
+          + "system.method to set the spellcasting method, or system.prepared to set the preparation state..",
```
krbz999:
```suggestion
          + "system.method to set the spellcasting method, or system.prepared to set the preparation state.",
```

module/data/item/spell.mjs
```diff
@@ -351,6 +382,12 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
         return { value, label, group: "DND5E.RangeDistance" };
       })
     ];
+
+    // Spellcasting
+    context.canPrepare = this.canPrepare;
+    context.spellcastingMethods = Object.values(CONFIG.DND5E.spellcasting).map(({ key, label }) => {
+      return { label, value: key };
+    });
```
krbz999, line 387: Consideration for some module compatibility here?
```js
const currentMethod = this.document.system.method;
if (currentMethod && !(currentMethod in CONFIG.DND5E.spellcasting)) {
  context.spellcastingMethods.push({ label: currentMethod, value: currentMethod });
}
```

This would mean temporarily disabling a module that adds a spellcasting method does not cause an issue just by viewing a sheet and making unrelated changes.

(With similar approach in class/subclass/advancement sheets)

module/data/spellcasting/spellcasting-model.mjs
```diff
@@ -0,0 +1,420 @@
...
+  /* -------------------------------------------- */
+
+  /**
+   * Initialize data models from global config.
+   */
+  static fromConfig() {
+    const { spellcasting } = CONFIG.DND5E;
+
+    // Give other packages a chance to define spellcasting.
+    Hooks.callAll("dnd5e.initSpellcasting", spellcasting);
```
krbz999: Is this breaking or will adding spellcasting data on `init` still work in the short term? I assume no?

module/data/spellcasting/spellcasting-model.mjs
```diff
@@ -0,0 +1,420 @@
...
+ * @extends {SlotSpellcasting}
+ */
+export class SingleLevelSpellcasting extends SlotSpellcasting {
+  /** @inheritDoc */
+  static defineSchema() {
+    return {
+      ...super.defineSchema(),
+      table: new TypedObjectField(new SchemaField({
+        slots: new NumberField({ required: true, nullable: false, integer: true, positive: true, initial: 1 }),
+        level: new NumberField({ required: true, nullable: false, integer: true, positive: true, initial: 1 })
```
krbz999, line 270: No fan of the accumulative approach, with each entry having deltas rather than overriding values?

module/data/spellcasting/spellcasting-model.mjs
```diff
@@ -0,0 +1,420 @@
...
+  /*  Validation                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Ensure spell slot table keys are numeric.
+   * @param {any} k  The key.
+   * @returns {boolean}
+   */
+  static #validateTableKey(k) {
+    return Number.isNumeric(k);
```
krbz999: Should also enforce being positive.

module/data/spellcasting/spellcasting-model.mjs
```diff
@@ -0,0 +1,420 @@
...
+   * Contribute to the actor's spellcasting progression for a spellcasting method that provides slots.
+   * @param {object} progression                      Spellcasting progression data. *Will be mutated.*
+   * @param {Actor5e|void} actor                      Actor for whom the data is being prepared, if any.
+   * @param {Item5e} [cls]                            Class for whom this progression is being computed.
+   * @param {SpellcastingDescription} [spellcasting]  Spellcasting descriptive object.
+   * @param {number} [count]                          Number of classes with this type of spellcasting.
+   */
+  computeProgression(progression, actor, cls, spellcasting, count) {
+    const prog = this.progression[spellcasting.progression];
+    if ( !prog ) return;
```
krbz999: It doesn't look like `spellcasting` is optional here, and it is called in `actor.mjs 415` without the last three params.

module/documents/item.mjs
```diff
@@ -365,26 +379,11 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
-  }
-
-  /* -------------------------------------------- */
-
   /**
    * Spellcasting details for a class or subclass.
    *
    * @typedef {object} SpellcastingDescription
-   * @property {string} type              Spellcasting type as defined in ``CONFIG.DND5E.spellcastingTypes`.
+   * @property {string} type              Spellcasting method as defined in ``CONFIG.DND5E.spellcasting`.
```
krbz999:
```suggestion
   * @property {string} type              Spellcasting method as defined in `CONFIG.DND5E.spellcasting`.
```

module/applications/components/inventory.mjs
Fyorl: Oops, that was supposed to be `Number`.

module/data/spellcasting/spellcasting-model.mjs
Fyorl: I suppose it wouldn't, which would mean the config deprecations might as well not exist. I will move it to `i18nInit` instead then. Maybe that's for the best anyway, as it allows `preLocalize` to run.

module/data/spellcasting/spellcasting-model.mjs
Fyorl: I found the existing table formats easier to read. You can input them directly by using the class tables from the books for reference without having to do any extra arithmetic. Also makes the logic for calculating available slots simpler.

module/data/spellcasting/spellcasting-model.mjs
Fyorl: You are right, making this optional was a later addition.

module/data/spellcasting/spellcasting-model.mjs
krbz999: Not sure I agree it's simpler since the additive approach means you simply iterate for each level. The current approach is simpler math (visually), though the additive approach makes adjustments easier. I don't feel strongly either way, though.

module/documents/actor/actor.mjs
```diff
@@ -385,29 +385,34 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+      const level = Object.values(progression).find(_ => _);
+      if ( level ) this.system.attributes.spell.level = level;
+      else progression.spell = this.system.attributes.spell.level ?? 0;
     }
 
-    for ( const type of Object.keys(CONFIG.DND5E.spellcastingTypes) ) {
+    for ( const [type, model] of Object.entries(CONFIG.DND5E.spellcasting) ) {
+      if ( !model.slots ) continue;
+      // Assume spellcasting methods without progression are based on character level rather than class level.
+      if ( foundry.utils.isEmpty(model.progression) ) model.computeProgression(progression, this);
```
krbz999, line 415: So this does not actually do anything, then. If the remaining parameters aren't given, the method simply returns. What is an example of character level based spellcasting anyhow?

Fyorl: It does not doing anything with the current implementations because the base system does not provide any spellcasting methods based on character level. If you were to implement one though, you could subclass `SlotSpellcasting` and provide a meaningful implementation.

module/applications/journal/class-page-sheet.mjs
```diff
@@ -292,27 +292,50 @@ export default class JournalClassPageSheet extends JournalEntryPageHandlebarsShe
...
+      }
+    } else if ( spellcastingModel?.slots ) {
+      const maxSpellLevel = Object.keys(CONFIG.DND5E.spellLevels).length - 1;
+      const spells = Object.fromEntries(Array.fromRange(maxSpellLevel, 1).map(l => {
+        return [spellcastingModel.getSpellSlotKey(l), {}];
+      }));
 
       let largestSlot;
       for ( const level of Array.fromRange(CONFIG.DND5E.maxLevel, 1).reverse() ) {
         const progression = { slot: 0 };
```
arbron: Need the adjusted initial value to get the spell table to show correctly.

```suggestion
        const progression = { [spellcasting.type]: 0 };
```

module/applications/journal/class-page-sheet.mjs
```diff
@@ -337,42 +360,16 @@ export default class JournalClassPageSheet extends JournalEntryPageHandlebarsShe
...
-       * @param {SpellcastingDescription} spellcasting  Spellcasting descriptive object.
-       * @function dnd5e.buildSpellcastingTable
-       * @memberof hookEvents
-       */
-      Hooks.callAll(
-        `dnd5e.build${spellcasting.type.capitalize()}SpellcastingTable`, table, item, spellcasting
-      );
-    }
+    /**
+     * A hook event that fires to generate the table for custom spellcasting types.
```
arbron: Since this calls for all spellcasting types now:

```suggestion
     * A hook event that fires to generate the table for spellcasting types.
```

module/data/item/spell.mjs
```diff
@@ -133,12 +130,28 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
...
+   * @deprecated since 5.1
+   * @ignore
+   */
+  get preparation() {
+    foundry.utils.logCompatibilityWarning("SpellData#preparation is deprecated. Please use SpellData#method in "
+      + "place of preparation.mode and SpellData#prepared in place of preparation.prepared.",
+    { since: "DnD5e 5.1", until: "DnD5e 5.4" });
+    if ( this.prepared === 2 ) return { mode: "always", prepared: 1 };
+    if ( this.method === "spell" ) return { mode: "prepared", prepared: this.prepared };
+    return { mode: this.method, prepared: this.prepared };
```
arbron: Before `preparation.prepared` was a boolean, not a number:

```suggestion
    if ( this.prepared === 2 ) return { mode: "always", prepared: true };
    if ( this.method === "spell" ) return { mode: "prepared", prepared: Boolean(this.prepared) };
    return { mode: this.method, prepared: Boolean(this.prepared) };
```

module/data/item/spell.mjs
```diff
@@ -190,6 +203,30 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
     }
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * Migrate preparation data.
+   * @since 5.1
```
arbron: Linter wants this to be an actual SemVer version. So we could adjust it or disable the `jsdoc/check-values` rule (which is only used for a handful of tags anyway: https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/check-values.md)

```suggestion
   * @since 5.1.0
```

templates/items/details/details-spell.hbs
```diff
@@ -43,21 +43,20 @@
     </div>
```
arbron, line 43: Some adjustments will need to be made to the spell header because "Class Spellcasting" is long enough to wrap, leading to the header elements to overlap the tabs:

Screenshot 2025-07-15 at 10 23 18

module/applications/actor/character-sheet.mjs
```diff
@@ -722,25 +721,26 @@ export default class CharacterActorSheet extends BaseActorSheet {
   async _getFavoriteData(type, id) {
     // Spell slots
     if ( type === "slots" ) {
+      const [method] = id.match(/^\D+/) ?? [];
```
arbron: Seems like we should be able to use `actor.system.spells[id].type` to get the method rather than having to do regex on the ID.

module/applications/api/primary-sheet-mixin.mjs
```diff
@@ -503,8 +503,8 @@ export default function PrimarySheetMixin(Base) {
     static sortItemsPriority(a, b) {
       return a.system.linkedActivity?.item?.name.localeCompare(b.system.linkedActivity?.item?.name, game.i18n.lang)
         || ((a.system.level ?? 0) - (b.system.level ?? 0))
-        || a.system.preparation?.mode?.localeCompare(b.system.preparation?.mode, "en")
-        || ((b.system.preparation?.prepared ?? 0) - (a.system.preparation?.prepared ?? 0))
+        || (a.system.method ?? "").compare(b.system.method ?? "")
+        || ((a.system.prepared ?? 0) - (b.system.prepared ?? 0))
```
arbron, line 506: So I reversed these when fixing a different bug recently because it made more sense to me that when using "Priority" mode, prepared and always prepared spells would sort above non-prepared spells.

arbron (changes requested):

templates/items/details/details-spell.hbs
krbz999: This is also currently a minor issue on 5.0.4 for spells with many classes.

Zanderaf: running into this consistently when making monsters as well- innate spellcasting is long enough to cause the elements to always wrap into the tabs.

module/applications/api/primary-sheet-mixin.mjs
Fyorl: I guess it only matters for cantrips, since otherwise spells with different methods are never mixed together in the same section. But I agree you'd want to see your prepared cantrips first I'd imagine.

templates/items/details/details-spell.hbs
Fyorl: I've changed it to just 'Spellcasting' for now. I had thought that 'Spellcasting Method: Spellcasting' looked a little tautological, but it's probably for the best since I think we would prefer the spellcasting method name to match the name of the feature that grants the spellcasting.

If @arbron doesn't hate it, we can move the list of classes to the tags at the bottom instead. Or just remove it entirely. I'm not sure how useful the class list is, and it generates some support burden from people asking how to get their custom classes into the list.

module/applications/actor/character-sheet.mjs
Fyorl: Good idea

templates/items/details/details-spell.hbs
krbz999: Would be sad to see it go. :)

module/applications/api/primary-sheet-mixin.mjs
arbron: Aw, wasn't quite clear about this. I rather than moving the method after prepared, I was thinking of reversing the sorting order of prepared itself.

```suggestion
        || ((b.system.prepared ?? 0) - (a.system.prepared ?? 0))
```

Current order:
Screenshot 2025-07-15 at 16 25 33

Adjusted order:
Screenshot 2025-07-15 at 16 25 43

templates/items/details/details-spell.hbs
arbron: The bottom looks good:

Screenshot 2025-07-15 at 16 30 59

arbron (approved):
Looking good

PR #5841, opened 2025-07-16
[#4895] Add automatic self enchanting option for Enchant activity

### LINKED ISSUE #4895: [Feature Request] Add a self-enchant checkbox to pre-populate the enchant chat message with the activity's parent item
There are items such as Dagger of Venom in the DMG '24 module and Lightbringer in the PBSO module that come with "self"-enchanting activities. It would be a nice feature to save a few drag and drops with the mouse from the inventory tab to the chat messages if a checkbox could be checked on the Enchant activity and have the "drag and drop" happen at chat message creation automatically.

arbron:
Adds an option to the Behavior section of Enchant activities that allows them to automatically enchant their parent item when used. When activated with this option, the Enchant activity will check to see if an enchantment from this item is already on its item. If one is found, it will be removed. If one isn't found, or the profile of the found enchantment doesn't match the configured profile, then a new enchantment will be created. This allows for easy toggling of an enchantment on an item, or swapping between multiple enchantments on a single item.

Enchant - Self Enchant Config

When multiple profiles are available, the usage dialog will select the currently applied one by default and indicate that it is active in the list.

Enchant - Self Enchant Selection

To avoid code duplication, some of the enchantment application code has been moved from `EnchantmentApplicationElement` into the `EnchantActivity`.

There is also some changes to how the behavior section of the activity sheets are created to avoid having to create a new template every time a new behavior option is required.

Closes #4895

### DIFF
--- a/module/documents/activity/enchant.mjs
+++ b/module/documents/activity/enchant.mjs
@@ -122,13 +122,13 @@ export default class EnchantActivity extends ActivityMixin(EnchantActivityData) 
 
   /**
    * Apply an enchantment to the provided item.
-   * @param {string} profile            ID of the enchantment profile to apply.
-   * @param {Item5e} item               Item to which to apply the enchantment.
+   * @param {string} profile                  ID of the enchantment profile to apply.
+   * @param {Item5e} item                     Item to which to apply the enchantment.
    * @param {object} [options={}]
    * @param {ChatMessage5e} [options.chatMessage]     Chat message used to make the enchantment, if applicable.
    * @param {ActiveEffect5e} [options.concentration]  Concentration active effect to associate with this enchantment.
-   * @param {boolean} [options.strict]  Display UI errors and prevent creation if enchantment isn't allowed.
-   * @returns {ActiveEffect5e|null}     Created enchantment effect if the process was successful.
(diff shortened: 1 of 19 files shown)


module/documents/activity/enchant.mjs
```diff
@@ -56,33 +67,103 @@ export default class EnchantActivity extends ActivityMixin(EnchantActivityData)
...
 
+  /**
+   * Apply an enchantment to the provided item.
+   * @param {string} profile            ID of the enchantment profile to apply.
+   * @param {Item5e} item               Item to which to apply the enchantment.
+   * @param {object} [options={}]
+   * @param {ChatMessage5e} [options.chatMessage]     Chat message used to make the enchantment, if applicable.
+   * @param {ActiveEffect5e} [options.concentration]  Concentration active effect to associate with this enchantment.
+   * @param {boolean} [options.strict]  Display UI errors and prevent creation if enchantment isn't allowed.
+   * @returns {ActiveEffect5e|null}     Created enchantment effect if the process was successful.
```
Fyorl:
```suggestion
   * @returns {Promise<ActiveEffect5e|null>}     Created enchantment effect if the process was successful.
```

Fyorl (approved):

PR #5842, opened 2025-07-16
[#5806] Alleviate chat forced reflow/target selection performance issues

### LINKED ISSUE #5806: Chat message updating when selecting tokens causes noticable performance issues
FVTT 346, DND 5.0.4, Chrome

When selecting a token, the targeting update for chat messages often causes a forced reflow. This issue is exponentially worse the more messages are in the chat and even worse the more modules adding ui elements there are as everything is forced to repaint

Image

The image showcases the lag spike (with a couple of UI modules). The same can be observed with 100 damage chat messages, the performance hit is between 10 to 40ms on my machine with no modules (which is noticable but not terrible) and around 100 with just a couple modules added that have ui elements. Generally speaking the more stuff you have on screen the worse it is (such as open windows and so on)

robertaird:
This proposes a two-pronged approach to alleviate the performance issues with target selection while there are a lot of chat messages:

1. Applies [`content-visibility: auto`](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility) to the chat message card/effect/target trays on all but the most recent 5 chat messages. 
2. `buildTargetsList` will only run when the chat tray is open. Also registers an event listener for the `toggled` event so any expanded sections are correctly updated.

Each change on its own made a noticable difference. The second is the bigger performance boost as long as you don't have too many expanded target selections in chat, and the first helps alleviate the rest.

I don't have a ton of experience with Mixins, but I suspect my changes to `TargetedApplicationMixin` are an anti-pattern and Mixins shouldn't be concerned with implementation details of their parent class. I considered a separate `TargetedApplicationChatTrayElement` but I wanted to keep changes minimal for easier review and any potential refactoring.

Closes #5806

### DIFF
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -494,9 +494,6 @@
     margin-top: 6px;
     &:not(.collapsed) > label { padding-bottom: 4px; }
   }
-  &.collapsed {
-    content-visibility: auto;
-  }
 
   > label {
     cursor: var(--cursor-pointer);
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -483,12 +483,19 @@
 }
 
 /* Damage & Effects Trays */
+.message.dnd5e2:not(:nth-last-child(-n + 5)) :is(.card-tray, .effects-tray, .targets-tray) {
+  content-visibility: auto;
+}
+
 .dnd5e2 :is(.card-tray, .effects-tray, .targets-tray) {
   margin-top: .125rem;
   &.damage-tray, &.effects-tray { margin-block-start: 8px; }
   &.targets-tray {
     margin-top: 6px;
     &:not(.collapsed) > label { padding-bottom: 4px; }
+  }
+  &.collapsed {
+    content-visibility: auto;
   }
 
   > label {
--- a/module/applications/chat-log.mjs
+++ b/module/applications/chat-log.mjs
@@ -4,6 +4,25 @@ import ChatMessage5e from "../documents/chat-message.mjs";
  * Custom implementation of the chat log to support saving tray states.
  */
 export default class ChatLog5e extends foundry.applications.sidebar.tabs.ChatLog {
+  /**
+   * The active intersection observer.
+   * @type {IntersectionObserver}
+   */
+  #intersections;
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _onFirstRender(context, options) {
+    const scroll = this.element.querySelector(".chat-scroll");
+    const log = this.element.querySelector(".chat-log");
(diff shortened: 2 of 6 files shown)


gambit07:
> This proposes a two-pronged approach to alleviate the performance issues with target selection while there are a lot of chat messages:
> 
> 1. Applies [`content-visibility: auto`](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility) to the chat message card/effect/target trays on all but the most recent 5 chat messages. 
> 2. `buildTargetsList` will only run when the chat tray is open. Also registers an event listener for the `toggled` event so any expanded sections are correctly updated.
> 
> Each change on its own made a noticable difference. The second is the bigger performance boost as long as you don't have too many expanded chat, and the first helps alleviate the rest.
> 
> I don't have a ton of experience with Mixins, but I suspect my changes to `TargetedApplicationMixin` are an anti-pattern and Mixins shouldn't be concerned with implementation details of their parent class. I considered a separate `TargetedApplicationChatTrayElement` but I wanted to keep changes minimal for easier review and any potential refactoring.
> 
> Closes #5806

How much of a drop in latency did you see?

robertaird:
> How much of a drop in latency did you see?

I tested with between 100 and 200 damage and effect application messages. 

When most of the target selections are collapsed, I can sometimes still see a small delay in the list populating but the latency is otherwise (to me) negligible. I also no longer see a hitch in the animated turn marker I'm using as I select and deselect targets.

If you do have a bunch expanded for some reason, the latency is still noticable but reduced, and it varies on how many targets you're selecting at once. I have a half baked thought on how this could still be improved further but I'm not sure it's necessary since these auto-collapse when you apply the damage or effects.

Here are some benchmarks I took when dragging to select 9 targets at once with one expanded damage message.

5.1.x branch:
5 1 x

and with these changes:
improvements

Fyorl:
I spent some time testing performance but I could not reproduce these profiling results. With 200 damage & effect trays open I never observed a recalculate style event of more than 5ms. Regardless, 5ms is still quite a lot, and building these targets lists is wasteful if the card is not visible.

I've pushed an additional commit here that will prevent the target list from being built at all if the card is not visible, as we have added a new 'never collapse' option for trays in 5.1, and so we can't safely assume that most trays will be collapsed anymore.

Since I had some trouble reproducing the profiling results, it would be useful if you could test these changes and let me know if you still experience the same performance gains as with the original commit.

robertaird:
Ah yeah, I missed the never collapse option. 

This looks perfect. I tested your changes with never collapse enabled, and even with all the trays expanded I'm now getting performance on par with my previous "most trays collapsed" tests 🥳

latest-commit

PR #5845, opened 2025-07-17
[#2187] Allow recovery periods to be set by uses scale value

### LINKED ISSUE #2187: Add `uses` scale value type

arbron:
Adds a scale value option to the recovery periods on items or activities. For non-embedded items, selected the "Scale Value" option will swap the input to a freeform text input with `@scale` filled in, allowing for entering any scale value. For items on actors, the list will be filled with any scale values on the actor (unless the current value isn't found, in which case it will fall back to showing the text input).

Uses Scale Value Options

Uses Scale Value Text

Uses Scale Value Rage

Closes #2187

### DIFF
--- a/module/data/shared/uses-field.mjs
+++ b/module/data/shared/uses-field.mjs
@@ -1,3 +1,4 @@
+import { ScaleValueTypeUsage } from "../../data/advancement/scale-value.mjs";
 import { formatNumber, formatRange, getPluralRules, prepareFormulaValue } from "../../utils.mjs";
 import FormulaField from "../fields/formula-field.mjs";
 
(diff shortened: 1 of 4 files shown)


module/data/shared/uses-field.mjs
```diff
@@ -46,6 +47,12 @@ export default class UsesField extends SchemaField {
 
     const periods = [];
     for ( const recovery of this.uses.recovery ) {
+      recovery.isScaleValue = recovery.period.startsWith("@scale");
+      if ( recovery.isScaleValue ) {
+        const scaleValue = foundry.utils.getProperty(rollData, recovery.period.replace("@", ""));
```
Fyorl:
```suggestion
        const scaleValue = foundry.utils.getProperty(rollData, recovery.period.slice(1));
```

Fyorl (approved):

PR #5850, opened 2025-07-17
[#5847] Add `identifier` to unlinked spells on spell list page

### LINKED ISSUE #5847: Add identifier to unlinked spells config in spell lists
Since all spells have identifiers now, it would be useful to be able to specify the identifier when adding spells to the unlinked spells in a spell list journal page. It would also be helpful to add a method to the spell list registry to fetch all the identifiers from linked or unlinked spells. This would allow for creating a method that easily fetches all spells on an actor's sheet from a certain spell list, regardless of their source.

arbron:
Adds `identifier` on unlinked spells in the config and adds a new `identifiers` getter in the spell list registry that returns all the identifiers for linked & unlinked spells in the list.

Modifies the spell list filtering in the compendium browser to use identifiers rather than UUIDs to allow for finding spells from multiple sources. This would allow a module to create a single spell list referencing SRD spells, but still allow the PHB versions of those spells to appear if that module is present.

Closes #5847

### DIFF
--- a/module/registry.mjs
+++ b/module/registry.mjs
@@ -269,6 +269,14 @@ class SpellListRegistry {
   /* -------------------------------------------- */
 
   /**
+   * IDs of compendiums that have been re-indexed during loading.
+   * @type {Set<string>}
+   */
+  static #compendiumsIndexed = new Set();
+
+  /* -------------------------------------------- */
+
+  /**
    * UUIDs of spell lists in the process of being loaded.
    * @type {Set<string>}
    */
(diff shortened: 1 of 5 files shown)


module/registry.mjs
```diff
@@ -351,10 +359,16 @@ class SpellListRegistry {
...
     const list = type.get(page.system.identifier);
-    list.contribute(page).forEach(uuid => {
+    await Promise.all(Array.from(list.contribute(page)).map(uuid => {
       if ( !SpellListRegistry.#bySpell.has(uuid) ) SpellListRegistry.#bySpell.set(uuid, new Set());
       SpellListRegistry.#bySpell.get(uuid).add(list);
-    });
+      const { collection } = foundry.utils.parseUuid(uuid);
+      if ( (collection instanceof CompendiumCollection) && !this.#compendiumsIndexed.has(collection.metadata.id) ) {
+        this.#compendiumsIndexed.add(collection.metadata.id);
+        this.#loading.add(collection.metadata.id);
```
Fyorl, line 368: Need to update docstring for `#loading`.

PR #5866, opened 2025-07-21
[#5787] Automatically create advancement on classes & backgrounds

### LINKED ISSUE #5787: Update default advancements on Background and Species when using Modern rules
### Species
When creating a new Species item, it adds a few advancements automatically. One of them is the Ability Score Improvement, but in the Modern rules the starting ASI is in the Background. Can we remove this from the Species when using the Modern rules?

### Background
When creating a new Background item, it doesn't have any default advancements. Could we add some to make it easier to make them?

- Ability Score Improvement, with 3 Points and 2 in Point Cap
- Grant Items for the Origin Feat
- Traits for the Skill Proficiencies
- another Traits for the Tool Proficiency

arbron:
In addition to the default advancement created on Species, this adds Trait & Grant Item advancement to backgrounds (and ASI with the modern rules) and Hit Points, ASI, and a single Grant Items advancement for classes.

Creates a new `AdvancementTemplate` for item data models that encapsulates the default creation code so that individual types just need to override `_advancementToCreate` to indicate what should be created and call `preCreateAdvancement` during the `_preCreate` method.

Closes #5787

### DIFF
--- a/module/data/item/background.mjs
+++ b/module/data/item/background.mjs
@@ -1,18 +1,17 @@
 import ItemDataModel from "../abstract/item-data-model.mjs";
-import AdvancementField from "../fields/advancement-field.mjs";
+import AdvancementTemplate from "./templates/advancement.mjs";
 import ItemDescriptionTemplate from "./templates/item-description.mjs";
 import StartingEquipmentTemplate from "./templates/starting-equipment.mjs";
 
-const { ArrayField } = foundry.data.fields;
-
 /**
  * Data definition for Background items.
+ * @mixes AdvancementTemplate
  * @mixes ItemDescriptionTemplate
  * @mixes StartingEquipmentTemplate
- *
- * @property {object[]} advancement  Advancement objects for this background.
  */
-export default class BackgroundData extends ItemDataModel.mixin(ItemDescriptionTemplate, StartingEquipmentTemplate) {
+export default class BackgroundData extends ItemDataModel.mixin(
+  AdvancementTemplate, ItemDescriptionTemplate, StartingEquipmentTemplate
+) {
 
   /* -------------------------------------------- */
   /*  Model Configuration                         */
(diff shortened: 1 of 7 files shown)


module/data/item/class.mjs
```diff
@@ -246,6 +247,29 @@ export default class ClassData extends ItemDataModel.mixin(ItemDescriptionTempla
...
+  /** @override */
+  _advancementToCreate(options) {
+    return [
+      { type: "HitPoints" },
+      { type: "AbilityScoreImprovement", level: 4 },
+      { type: "AbilityScoreImprovement", level: 8 },
+      { type: "AbilityScoreImprovement", level: 12 },
+      { type: "AbilityScoreImprovement", level: 16 },
+      { type: "AbilityScoreImprovement", level: 19 },
+      { type: "ItemGrant", title: game.i18n.localize("DND5E.ADVANCEMENT.Defaults.ClassFeatures"), level: 1 }
```
Fyorl: If all classes got features at the same levels, I think we could fill them in here. Since they don't, and we're only filling in the level 1 feature because it's the only one we know about for sure, I would prefer we didn't attempt it at all.

arbron: I was figuring it might be convenient to have one pre-created that you could then duplicate across levels, but I'm fine leaving it out.

arbron:
> I noticed that with modern rules, background & species both offer 'Common' as a language, which is probably not correct.

That was an oversight, I meant to leave it just on the Background to match how we implemented Backgrounds in the PHB/SRD.

module/data/item/background.mjs
```diff
@@ -60,6 +50,35 @@ export default class BackgroundData extends ItemDataModel.mixin(ItemDescriptionT
...
+
+    return [
+      { type: "AbilityScoreImprovement", configuration: { points: 3 } },
+      { type: "Trait", title: game.i18n.localize("DND5E.ADVANCEMENT.Defaults.BackgroundProficiencies") },
+      {
+        type: "Trait",
+        title: game.i18n.localize("DND5E.ADVANCEMENT.Defaults.ChooseLanguages"),
+        configuration: { grants: ["languages:standard:common"] }
+      },
+      { type: "ItemGrant", title: game.i18n.localize("DND5E.ADVANCEMENT.Defaults.BackgroundFeat") }
```
Fyorl, line 68: #5138 is approved so maybe worth adding `configuration` here to restrict to origin feats.

arbron: Well this is an Item Grant because backgrounds only offer a single feat rather than a choice.

PR #5871, opened 2025-07-21
Fix some dialogs and concentration

### LINKED ISSUE #5870: Dialog prompt for choosing which concentration effect to end always shows fallback name

krbz999:
Moved some of the more basic uses of `Dialog` over to `DialogV2`.

Fixed the concentration ending prompt's select showing the fallback name in all cases.

Closes #5870.

### DIFF
--- a/module/documents/mixins/pseudo-document.mjs
+++ b/module/documents/mixins/pseudo-document.mjs
@@ -235,7 +235,7 @@ export default function PseudoDocumentMixin(Base) {
       const type = game.i18n.localize(this.metadata.label);
       return foundry.applications.api.Dialog.confirm({
         window: { title: `${game.i18n.format("DOCUMENT.Delete", { type })}: ${this.name || this.title}` },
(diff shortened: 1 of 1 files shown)


module/applications/advancement/advancement-manager.mjs
```diff
@@ -572,25 +572,30 @@ export default class AdvancementManager extends Application5e {
...
+          title: `${game.i18n.localize("DND5E.ADVANCEMENT.Manager.ClosePrompt.Title")}: ${this.actor.name}`
+        },
+        position: { width: 400 },
         content: game.i18n.localize("DND5E.ADVANCEMENT.Manager.ClosePrompt.Message"),
-        buttons: {
-          close: {
-            icon: '<i class="fas fa-times" inert></i>',
+        buttons: [
+          {
+            action: "abort",
```
krbz999: Controversial change changing this from "close" to "abort", I'm sure. How ever will we cope.

(Using `action: "close"` prevented the callback from working at all.)

arbron: No objection to "abort", but perhaps "stop" to be more in line with the label?

PR #5875, opened 2025-07-22
[#5422] Allow referencing target CR in Transform activity

### LINKED ISSUE #5422: Add support for checking target's CR when determining CR limit in transform activity
Polymorph relies on limiting the creature selection based on the target's CR or character level, but currently it can only reference the caster's stats in the CR formula. When calculating the final CR, it should be possible to use `@target.details.cr` to use the target's stats rather than the caster's. In situations where there is more than one targeted actor, this would either use the first CR or find the lowest CR of the group.

arbron:
Adds `@target.details.cr` to the roll data for the Transform activity allowing it to be used when selecting the CR restriction in the compendium browser. If multiple targets are selected, then this number is the lowest of all of them.

Closes #5422

### DIFF
--- a/module/documents/activity/transform.mjs
+++ b/module/documents/activity/transform.mjs
@@ -190,7 +190,7 @@ export default class TransformActivity extends ActivityMixin(TransformActivityDa
   getRollData(options) {
     const rollData = super.getRollData(options);
     if ( game.user.targets.size ) {
-      foundry.utils.setProperty(rollData, "target.cr", game.user.targets.reduce((final, token) => {
+      foundry.utils.setProperty(rollData, "target.details.cr", game.user.targets.reduce((final, token) => {
         const details = token.actor?.system.details ?? {};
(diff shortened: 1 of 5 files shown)


module/documents/activity/transform.mjs
```diff
@@ -181,4 +181,21 @@ export default class TransformActivity extends ActivityMixin(TransformActivityDa
...
+
+  /* -------------------------------------------- */
+  /*  Helpers                                     */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  getRollData(options) {
+    const rollData = super.getRollData(options);
+    if ( game.user.targets.size ) {
+      foundry.utils.setProperty(rollData, "target.cr", game.user.targets.reduce((final, token) => {
```
Fyorl: Not ready to open the floodgates on including `@target` in roll data. Conceivably this might be the target's full roll data in its final implementation, so this would have to change to `@target.details.cr`.

PR #5876, opened 2025-07-22
[#5581] Add direct link mode for selection in Transform activity

### LINKED ISSUE #5581: Allow world actors to be linked directly to the transform activity
Allow world actors to be linked directly to the transform activity (compared to the summon activity), which makes it possible to only allow certain world creatures to be transformed instead of opening the compendium browser.

The reason for this is, firstly, that my players don't have read permissions to the Monster Manual, which causes the compendium browser to appear empty, and secondly, that I sometimes make changes to creatures (e.g., token images) that aren't taken into account during import.
Furthermore, a druid, for example, only has a certain number of known forms. To maintain overview, it would also be useful to link these directly to the activity.

With the summon activity, I can simply drag the creatures onto the activity. This is currently not possible with the transform activity.

Tested with:
- Foundry V13 Stable1 441 (Node Js Version)
- DnD5e v5.0.1
- FireFox Mac v138.0

arbron:
Adds mode to the Transform activity to select between the default CR mode and a new Direct Link mode (similar to how the Summon activity is set up). When the direct link mode is selected a UUID field replaces the CR field allowing adding a specific actor which will be used when transforming.

Transform - Direct Link

Closes #5581

### DIFF
--- a/module/data/activity/transform-data.mjs
+++ b/module/data/activity/transform-data.mjs
@@ -4,7 +4,8 @@ import TransformationSetting from "../settings/transformation-setting.mjs";
 import BaseActivityData from "./base-activity.mjs";
 
 const {
-  ArrayField, BooleanField, DocumentIdField, EmbeddedDataField, NumberField, SchemaField, SetField, StringField
+  ArrayField, BooleanField, DocumentIdField, DocumentUUIDField, EmbeddedDataField,
+  NumberField, SchemaField, SetField, StringField
 } = foundry.data.fields;
 
 /**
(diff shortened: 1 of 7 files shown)


module/data/activity/transform-data.mjs
```diff
@@ -32,6 +35,7 @@ const {
  * @property {boolean} transform.customize     Should any customized settings be respected or should the default
  *                                             settings for the selected profile be used instead.
  * @property {string} transform.identifier     Class identifier that will be used to determine applicable level.
+ * @property {string} transform.mode           Method of determining what type of creature to transform into.
```
Fyorl:
```suggestion
 * @property {""|"cr"} transform.mode           Method of determining what type of creature to transform into.
```

Fyorl (approved):
UI looks good but see my earlier comment on the docstring.

Issue #5643, opened 2025-05-08
Add setting to configure calendar used in the system


arbron:
Currently the calendar can only be set programmatically, but there should be a mechanism in the system for selecting which calendar to use. This will allow modules to register their own calendars that can then be selected by the GM.

JPMeehan:
Feels like this should be a core setting, if we're talking about adjusting `CONFIG.time.worldCalendarConfig`?

PR #5877, opened 2025-07-24
[#5643] Add calendar registration system & setting

arbron:
Adds a list of available calendar systems within `DND5E.calendar` which includes the Gregorian calendar by default. The calendar settings app allows for changing the selected calendar system.

Because calendar setup needs to occur during the system's `init` hook, a new `dnd5e.setupCalendar` hook has been added to allow modules to customize the calendar configuration before it is finalized.


module/_types.mjs
```diff
@@ -146,12 +146,20 @@
...
+/**
+ * @typedef {FormSelectOption} CalendarOption
+ * @property {object} config                Calendar configuration data.
+ * @property {typeof CalendarData} [class]  Data model class.
+ */
+
 /**
  * @typedef {FormSelectOption} CalendarTimeFormatter
  * @property {string|TimeFormatter} formatter  The formatter name on the current calendar or a formatter function.
+ * @property {string[]} [calendars]            Calendars that support this formatter, or blank for all calendars.
```
Fyorl: Slightly awkward if you want to register support for your calendar with a particular formatter, which ends up then removing support for every other calendar.

Should it be the other way around that calendars have a list of supported formatters instead? Or should calendars just take care to check this isn't empty before pushing to it?

arbron: I'm thinking I may just remove this part and have all formatters always available. A calendar that wants to disable a specific formatter can use the hook to remove it from the config during initialization and can add their own formatters if their calendar is enabled.

PR #5878, opened 2025-07-24
[#5411] Allow transformation to retain spells in certain spell list

### LINKED ISSUE #5411: Add option to retain spells during transformation based on spell list
The Circle of the Moon Druid is able to retain and cast their circle spells while in Wild Shape. Adding an option to select a spell list that is retained after transformation would allow this.

arbron:
Add a new transformation option that allows specifying one or more spell lists. If included and the player has an item matching the specified spell list identifier (so `subclass:moon` would search for a subclass with the `moon` identifier), then any spells on that list will be retained during transformation.

Screenshot 2025-07-24 at 13 59 48

Adds spell lists to the SRD for each of the included subclasses to allow for them to be selected.

Closes #5411

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -3779,7 +3779,7 @@ DND5E.transformation = {
       label: "DND5E.TRANSFORM.Preset.WildShape.Label",
       settings: {
         effects: new Set(["otherOrigin", "origin", "feat", "spell", "class", "background"]),
-        keep: new Set(["bio", "class", "feats", "hp", "languages", "mental", "spells", "type"]),
+        keep: new Set(["bio", "class", "feats", "hp", "languages", "mental", "type"]),
         merge: new Set(["saves", "skills"]),
         minimumAC: "(13 + @abilities.wis.mod) * sign(@subclasses.moon.levels)",
         spellLists: new Set(["subclass:moon"]),
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -2807,7 +2807,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     delete d.system.resources; // Don't change your resource pools
     delete d.system.currency; // Don't lose currency
     delete d.system.bonuses; // Don't lose global bonuses
-    if ( settings.keep.has("spells") ) delete d.system.attributes.spellcasting; // Keep spellcasting ability if retaining spells.
+    if ( settings.keep.has("spells") || settings.spellLists.size ) delete d.system.attributes.spellcasting; // Keep spellcasting ability if retaining spells.
 
     // Specific additional adjustments
     d.system.details.alignment = o.system.details.alignment; // Don't change alignment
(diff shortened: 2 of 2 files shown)


module/config.mjs
```diff
@@ -3771,9 +3779,10 @@ DND5E.transformation = {
       label: "DND5E.TRANSFORM.Preset.WildShape.Label",
       settings: {
         effects: new Set(["otherOrigin", "origin", "feat", "spell", "class", "background"]),
-        keep: new Set(["bio", "class", "feats", "hp", "languages", "mental", "type"]),
+        keep: new Set(["bio", "class", "feats", "hp", "languages", "mental", "spells", "type"]),
```
Fyorl: Only the moon druid can cast a selection of spells while Wild Shaped, not other Druids, right? So should this really be enabled for Wild Shape in general?

Fyorl: I've made this change on the commit but let me know regardless.

Fyorl (changes requested):
Having the subclasses appear in the compendium browser alongside the classes is a bit noisy when I would expect the main use case to be selecting spells from a class list for preparation.

I tried adding a separate filter to `SpellData` to separate out class and subclass lists, but that results in the selections being ANDed together, when instead you would probably want to OR them to see the full list of spells available to prepare.

I think separating out these filters instead with a post-render hook would work fine, but at that point I considered if we really need to show them at all in the compendium browser. They are always prepared so I don't know why you would need to see them when preparing spells.

The only thing I could think of was legacy Warlocks with their expanded spell lists that aren't always prepared, but legacy Warlock lists are not included in this PR.

For now I've gone with hiding them since that was the easier change, but let me know what you think.

module/config.mjs
arbron: We need "Keep Spells" checked to transfer spell slots appropriately both directions. For non Moon druids, the transformation system will see the Moon druid spell list, check and see that the actor doesn't have the `moon` subclass, and thus not copy over any spells.

arbron:
> Having the subclasses appear in the compendium browser alongside the classes is a bit noisy when I would expect the main use case to be selecting spells from a class list for preparation.
> 
> I tried adding a separate filter to `SpellData` to separate out class and subclass lists, but that results in the selections being ANDed together, when instead you would probably want to OR them to see the full list of spells available to prepare.
> 
> I think separating out these filters instead with a post-render hook would work fine, but at that point I considered if we really need to show them at all in the compendium browser. They are always prepared so I don't know why you would need to see them when preparing spells.
> 
> The only thing I could think of was legacy Warlocks with their expanded spell lists that aren't always prepared, but legacy Warlock lists are not included in this PR.
> 
> For now I've gone with hiding them since that was the easier change, but let me know what you think.

This should be fine for now, but I have a thought on how to split them out for final display. Its probably a bit bigger than we want in this PR, so this should work as a good enough solution for now.

module/config.mjs
Fyorl: That's only if you have the moon druid list in your World though, right? The PHB isn't updated yet to include it, and those people who don't have the PHB will get all spells copied, no?

Could we adjust the `revertOriginalForm` logic instead to check for `transformOptions.spellLists` as well as `keep.has("spell")`?

arbron: Aw, that is because when the spell list isn't registered the value doesn't get populated at all. We could alter the behavior of `transformInto` and `revertOriginalForm` instead.

Fyorl (approved):

PR #5880, opened 2025-07-25
[#2094] Move actor socket events into data models where applicable

### LINKED ISSUE #2094: Shift type-specific logic into data models where possible

arbron:
- Move HP socket events to `AttributesFields`
- Move size socket events to `TraitsFields`
- Move other character events into `CharacterData`
- Rewrite `Actor5e#_updateExhaustion` using `_preUpdateOperation`
- Reorder methods in `Actor5e` for more clear grouping, add more section headings

Closes #2094

### DIFF
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -521,9 +521,9 @@ export default class AttributesFields {
   /* -------------------------------------------- */
 
   /**
-   * Display concentration challenge if necessary, set bloodied status, and throw damage hook.
+   * Display concentration challenge if necessary, set bloodied status, and fire damage hook.
    * @this {CharacterData|NPCData|VehicleData}
-   * @param {object} changed  The differential data that was changed relative to the documents prior values.
+   * @param {object} changed  The differential data that was changed relative to the document's prior values.
    * @param {object} options  Additional options which modify the update request.
    * @param {string} userId   The id of the User requesting the document update.
    */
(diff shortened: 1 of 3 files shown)


module/data/actor/templates/attributes.mjs
```diff
@@ -497,4 +497,69 @@ export default class AttributesFields {
...
+      foundry.utils.setProperty(changes, "system.attributes.death.success", 0);
+      foundry.utils.setProperty(changes, "system.attributes.death.failure", 0);
+    }
+    foundry.utils.setProperty(options, "dnd5e.hp", { ...this.attributes.hp });
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Display concentration challenge if necessary, set bloodied status, and throw damage hook.
```
Fyorl:
```suggestion
   * Display concentration challenge if necessary, set bloodied status, and fire damage hook.
```

module/data/actor/templates/attributes.mjs
```diff
@@ -497,4 +497,69 @@ export default class AttributesFields {
...
+   * @this {CharacterData|NPCData|VehicleData}
+   * @param {object} changed  The differential data that was changed relative to the documents prior values.
```
Fyorl:
```suggestion
   * @param {object} changed  The differential data that was changed relative to the document's prior values.
```

module/data/actor/templates/attributes.mjs
```diff
@@ -497,4 +497,69 @@ export default class AttributesFields {
...
+   * @param {object} options  Additional options which modify the update request.
+   * @param {string} userId   The id of the User requesting the document update.
+   */
+  static async onUpdateHP(changed, options, userId) {
+    if ( !changed.system?.attributes?.hp ) return;
+    if ( userId === game.userId ) await this.parent.updateBloodied(options);
+
+    const hp = options.dnd5e?.hp;
+    if ( hp && !options.isRest && !options.isAdvancement ) {
```
Fyorl: Seems better to early return here and avoid some rightward-shift.

module/documents/actor/actor.mjs
```diff
@@ -264,6 +264,33 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+   * Format a type object into a string.
+   * @param {object} typeData          The type data to convert to a string.
+   * @returns {string}
+   */
+  static formatCreatureType(typeData) {
+    if ( typeof typeData === "string" ) return typeData; // Backwards compatibility
+    let localizedType;
+    if ( typeData.value === "custom" ) {
+      localizedType = typeData.custom;
+    } else if ( typeData.value in CONFIG.DND5E.creatureTypes ) {
```
```suggestion
    if ( typeData.value === "custom" ) localizedType = typeData.custom;
    else if ( typeData.value in CONFIG.DND5E.creatureTypes ) {
```

module/documents/actor/actor.mjs
```diff
@@ -264,6 +264,33 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+      localizedType = game.i18n.localize(typeData.swarm ? code.plural : code.label);
+    }
+    let type = localizedType;
+    if ( typeData.swarm ) {
+      type = game.i18n.format("DND5E.CreatureSwarmPhrase", {
+        size: game.i18n.localize(CONFIG.DND5E.actorSizes[typeData.swarm].label),
+        type: localizedType
+      });
+    }
+    if (typeData.subtype) type = `${type} (${typeData.subtype})`;
```
```suggestion
    if ( typeData.subtype ) type = `${type} (${typeData.subtype})`;
```

module/documents/actor/actor.mjs
```diff
@@ -3387,47 +3314,39 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+      const level = foundry.utils.getProperty(operation.updates[index], "system.attributes.exhaustion");
+      if ( !Number.isFinite(level) ) continue;
+      const effect = actor.effects.get(ActiveEffect5e.ID.EXHAUSTION);
+
+      // Delete any existing effect is no exhaustion level
+      if ( level < 1 ) await effect?.delete();
+
+      // Adjust stored exhaustion value in existing effect
+      else if ( effect ) {
+        await effect.update({ "flags.dnd5e.exhaustionLevel": level }, { dnd5e: { originalExhaustion } });
```
Fyorl: Probably shouldn't perform actual CRUD operations in this method, it's still possible for the update to fail (maybe it's rejected by the server for whatever reason). I think the intention behind my original comment on `_onUpdateExhaustion` was that we amend the appropriate AE and send the update along with the update delta as part of the `effects` collection. Similarly, you can inject a new effect into the collection to create the AE. Deletion should probably remain in `_onUpdate`. However, if we need certain hooks to fire on creation/update of these AEs, then we have to fire full CRUD operations for those exhaustion AEs.

We have a proposal for a generalised framework around AEs with multiple levels in #6699, so maybe it makes more sense to stick with `_onUpdate` for now and revisit during or after #6699.

module/data/actor/templates/traits.mjs
```diff
@@ -128,4 +128,43 @@ export default class TraitsField {
...
+   * Update the prototype token size for newly created actors.
+   * @this {CharacterData|NPCData|VehicleData}
+   * @param {object} data     The initial data object provided to the document creation request.
+   * @param {object} options  Additional options which modify the creation request.
+   */
+  static async preCreateSize(data, options) {
+    if ( this.parent._stats?.compendiumSource?.startsWith("Compendium.") ) return;
+    const prototypeToken = {};
+    if ( "size" in (this.system.traits || {}) ) {
+      const size = CONFIG.DND5E.actorSizes[this.system.traits.size || "med"].token ?? 1;
```
Fyorl:
```suggestion
    if ( "size" in (this.traits || {}) ) {
      const size = CONFIG.DND5E.actorSizes[this.traits.size || "med"].token ?? 1;
```

Fyorl (changes requested):

PR #5881, opened 2025-07-25
[#5025] Switch legendaries to use `spent` rather than `value`

### LINKED ISSUE #5025: Use "spent" rather than "value" for legendary actions & resistances
With the new "In Lair" toggle, the maximum value for legendary actions and resistances can change quickly, potentially during combat. This can cause an issue because those uses are tracked by the number of remaining uses. If you have "3" maximum uses set and no spent uses and toggle the "In Lair" toggle, the actor will end up with 3 out of 4 uses even through one hasn't been spent yet. Conversely, if the actor starts in a lair with all 4 uses available and then leaves the lair, they will be left with 4 uses remaining even though their maximum is now 3.

Tracking through spent uses rather than remaining uses would solve both of these problems, so the creature entering its lair would end up with 4 remaining uses and the one leaving with 3.

Some accommodation would have to be made for existing consumption that targets the `value` property rather than spent, and a migration would have to be added to move one to the other (luckily it is simpler than the uses migration due to the maximum not supporting formulas).

arbron:
Change legendary actions and resistances from storing a `value` to storing the number of `spent`. This leads to the correct number of available legendaries when an NPC enters or leaves their lair.

Some adjustments were made to `consumeAttribute` to allow for adjusting `spent` when the consumption target is `value`. This helps ensure the consumption limits are correct and that no migration is needed for consumption targets.

Closes #5025

### DIFF
--- a/module/applications/actor/api/base-actor-sheet.mjs
+++ b/module/applications/actor/api/base-actor-sheet.mjs
@@ -1571,9 +1571,10 @@ export default class BaseActorSheet extends PrimarySheetMixin(
   static #togglePip(event, target) {
     const n = Number(target.closest("[data-n]")?.dataset.n);
     const prop = target.dataset.prop ?? target.closest("[data-prop]")?.dataset.prop;
-    if ( !n || Number.isNaN(n) || !prop ) return;
+    if ( !Number.isNumeric(n) || !prop ) return;
     let value = foundry.utils.getProperty(this.actor, prop);
-    if ( value === n ) value--;
+    if ( (value === n) && prop.endsWith(".value") ) value--;
+    else if ( (value === n) && prop.endsWith(".spent") ) value++;
     else value = n;
     this.submit({ updateData: { [prop]: value } });
   }
--- a/module/data/actor/npc.mjs
+++ b/module/data/actor/npc.mjs
@@ -254,7 +254,8 @@ export default class NPCData extends CreatureTemplate {
   static _migrateData(source) {
     super._migrateData(source);
     NPCData.#migrateEnvironment(source);
-    NPCData.#migrateLegendaries(source);
+    NPCData.#migrateLegendaries(source, "legact");
+    NPCData.#migrateLegendaries(source, "legres");
     NPCData.#migrateSource(source);
     NPCData.#migrateSpellLevel(source);
     NPCData.#migrateTypeData(source);
@@ -277,15 +278,14 @@ export default class NPCData extends CreatureTemplate {
   /**
    * Convert legendary action & resistance `value` to `spent`.
    * @param {object} source  The candidate source data from which the model will be constructed.
+   * @param {string} prop    The property to migrate.
    * @since 5.1.0
    */
-  static #migrateLegendaries(source) {
-    for ( const k of ["legact", "legres"] ) {
-      if ( !foundry.utils.hasProperty(source, `resources.${k}.value`)
-        || !foundry.utils.hasProperty(source, `resources.${k}.max`) ) continue;
-      foundry.utils.setProperty(source, `resources.${k}.spent`, source.resources[k].max - source.resources[k].value);
-      delete source.resources[k].value;
-    }
+  static #migrateLegendaries(source, prop) {
+    const resource = source.resources?.[prop];
+    if ( !resource || !("max" in resource) || !("value" in resource) || ("spent" in resource) ) return;
+    source.resources[prop].spent = resource.max - resource.value;
+    delete resource.value;
   }
 
   /* -------------------------------------------- */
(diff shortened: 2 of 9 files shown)


krbz999:
Do token attribute bars still work?

arbron:
> Do token attribute bars still work?

Yep, they rely on the derived `value` so nothing changes for them.

krbz999:
> > Do token attribute bars still work?
> 
> Yep, they rely on the derived `value` so nothing changes for them.

Just tested. They're not editable on the token any more cus they're single values.

For some reason I also do not seem able to edit LegAct or LegRes directly on the NPC sheet, but only if opened as a synthetic actor.

arbron:
> Just tested. They're not editable on the token any more cus they're single values.

Fixed.

> For some reason I also do not seem able to edit LegAct or LegRes directly on the NPC sheet, but only if opened as a synthetic actor.

Not able to reproduce this issue.

krbz999:
> Not able to reproduce this issue.

Same-sies, now. :)

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -1571,9 +1571,10 @@ export default class BaseActorSheet extends PrimarySheetMixin(
   static #togglePip(event, target) {
     const n = Number(target.closest("[data-n]")?.dataset.n);
     const prop = target.dataset.prop ?? target.closest("[data-prop]")?.dataset.prop;
-    if ( !n || Number.isNaN(n) || !prop ) return;
+    if ( !Number.isNumeric(n) || !prop ) return;
```
Fyorl, line 1574: Checking `Number.isNaN` is not sufficient anymore?

module/data/actor/npc.mjs
```diff
@@ -256,6 +254,7 @@ export default class NPCData extends CreatureTemplate {
   static _migrateData(source) {
     super._migrateData(source);
     NPCData.#migrateEnvironment(source);
+    NPCData.#migrateLegendaries(source);
```
Fyorl: Perf tweaks

```suggestion
    NPCData.#migrateLegendaries(source, "legact");
    NPCData.#migrateLegendaries(source, "legres");
```

module/data/actor/npc.mjs
```diff
@@ -275,6 +274,22 @@ export default class NPCData extends CreatureTemplate {
...
+   * @since 5.1.0
+   */
+  static #migrateLegendaries(source) {
+    for ( const k of ["legact", "legres"] ) {
+      if ( !foundry.utils.hasProperty(source, `resources.${k}.value`)
+        || !foundry.utils.hasProperty(source, `resources.${k}.max`) ) continue;
+      foundry.utils.setProperty(source, `resources.${k}.spent`, source.resources[k].max - source.resources[k].value);
+      delete source.resources[k].value;
+    }
+  }
```
Fyorl: Perf tweaks (avoid throwaway array) and a check for whether the field has already been migrated.

```suggestion
  /**
   * Convert legendary action & resistance `value` to `spent`.
   * @param {object} source  The candidate source data from which the model will be constructed.
   * @param {string} prop    The property to migrate.
   * @since 5.1.0
   */
  static #migrateLegendaries(source, prop) {
    const resource = source.resources?.[prop];
    if ( !resource || !("max" in resource) || !("value" in resource) || ("spent" in resource) ) return;
    source.resources[prop].spent = resource.max - resource.value;
    delete resource.value;
  }
```

module/data/actor/npc.mjs
```diff
@@ -460,6 +475,10 @@ export default class NPCData extends CreatureTemplate {
     };
     AttributesFields.prepareHitPoints.call(this, this.attributes.hp, hpOptions);
 
+    // Legendary Actions & Resistances
+    const clampValue = source => Math.clamp(source.max - source.spent, 0, source.max);
+    this.resources.legact.value = clampValue(this.resources.legact);
+    this.resources.legres.value = clampValue(this.resources.legres);
```
Fyorl: Avoid throwaway closure in data prep

```suggestion
    const { legact, legres } = this.resources;
    legact.value = Math.clamp(legact.max - legact.spent, 0, legact.max);
    legres.value = Math.clamp(legres.max - legres.spent, 0, legres.max);
```

Fyorl (approved):
Noticed these show up as recovering 'spent legendary actions +1' in the recovery card. Very unlikely to come up in normal play though I'd imagine.

module/applications/actor/api/base-actor-sheet.mjs
arbron: When using `spent` `0` becomes a valid option, so I wanted to distinguish between getting a `0` and getting `undefined`.

Fyorl: Checking for `Number.isNaN` still accomplishes that though, no?

arbron: Well while `undefined` is not a number, it is not not a number, if you catch my drift :)

Screenshot 2025-07-28 at 21 13 36

(so using `Number.isNaN(n)` would result in `undefined` being passed through, while `!Number.isNumeric(n)` doesn't)

PR #5882, opened 2025-07-25
[#5869] Rework enricher event handlers to use `onRender`

### LINKED ISSUE #5869: Prismatic Layers & Prismatic Rays roll tables include baked-out enrichers
Rather than including the original enricher data, these tables include the enrichers baked out as HTML which doesn't render with the correct styles and may affect future functionality.

arbron:
Removes the global event listeners previously used to handle enricher events and replaces them with adding events in the `onRender` callback. This involved some changes to how the chat listeners for roll requests are handled because they were using the global listeners.

Closes #5869


module/enrichers.mjs
```diff
@@ -1350,23 +1547,63 @@ function createRollLink(label, dataset={}, { classes="roll-link", tag="a" }={})
...
   _addDataset(link, dataset);
+  if ( tag === "a" ) link.dataset.action = "roll";
   return link;
 }
 
 /* -------------------------------------------- */
 /*  Actions                                     */
 /* -------------------------------------------- */
 
+const _addListeners = (buttons, handler) =>
```
Fyorl: Might as well be declared as a function, and go under Helpers.

module/enrichers.mjs
```diff
@@ -1497,106 +1673,41 @@ function createRequestButton(dataset) {
...
+async function handleRoll(event, target) {
+  const dataset = getRollActionDataset(target);
+  const link = target.closest("a") ?? target;
+  link.disabled = true;
+  window.getSelection().empty();
+  try {
+    switch ( dataset.type ) {
+      case "attack": return await rollAttack(dataset, event);
+      case "damage": return await rollDamage(dataset, event);
+      case "item": return await useItem(dataset, event);
```
Fyorl, line 1687: `useItem` doesn't take `event`. Maybe it should, and forward it to `activity.use`?

module/enrichers.mjs
```diff
@@ -1497,106 +1670,61 @@ function createRequestButton(dataset) {
...
-          messageType: "roll",
-          roll: { type: rollType },
-          targets: getTargetDescriptors()
-        }
-      },
-      flavor: game.i18n.localize(`DND5E.${rollType === "healing" ? "Healing" : "Damage"}Roll`),
-      speaker: ChatMessage.implementation.getSpeaker()
-    }
-  };
+const LISTENER = Symbol("enricher listener");
```
Fyorl:
```suggestion
const LISTENER = Symbol("enricherListener");
```

Issue #4425, opened 2024-09-25
Add "Ongoing Effect" active effect type


arbron:
This new active effect type would be used by the system for tracking any ongoing effects. This would be created automatically by activities whenever one with a non-instantaneous duration is used. This effect would be a superset of the functionality currently handled by concentration effects:
- Track concentration
- Track created measured templates
- Track summoned creatures
- Track effects applied to actors
- Track enchantments applied

There would be some sort of UI on the actor sheet to make it clear what is currently ongoing, see changes caused by that ongoing effect, and end it. Ending the effect would remove any templates, summons, or applied effects.

Kapuzenjoe:
I think the feature is extremely good and can hardly wait!

But I have an additional suggestion, if it doesn't already exist as a separate feature request: Would it then also be possible to trigger activities from the active effects?
For example, the spell "Tasha's Caustic Brew" where a creature that is covered always gets a 2d4 acid at the start of its turn. Then you could roll the damage directly with all parameters from the active effect

PR #5892, opened 2025-07-28
[#4963] Add visibility options to activities

### LINKED ISSUE #4963: Adding visibility options for Item Activities
In particular, adding a visibility toggle for users that are attuned/unattuned or when an item is unidentified, so an unattuned player using an unidentified item doesn't see unavailable-unless-attuned or identification-required activities in their options, but I can see other possible visibility options being useful as well.

With regard to identification, could also toggle labeling unidentified activities as "Unidentified Activity" when the item is unidentified.

arbron:
Adds several options to activities that control the ability to use or configure them:
- Requires Attunement (not usable if not attuned)
- Requires Magic (not usable in antimagic field)
- Requires Identification (not usable or visible if not identified)
- Level range (not usable if player is outside the specified range, can be set to character level or specific class level)

Activity Visibility - Config

When an activity isn't usable it will not appear in the activity selection dialog, in the list of activities beneath an item in inventory, or in the item's summary in inventory.

Activity Visibility - Inventory (not attuned)

Activity Visibility - Inventory (attuned)

When an activity isn't configurable it won't appear in the list of activities on the item (currently the activities tab is hidden on unidentified items, but this might allow us to show it).

Some options aren't always available:
- Require attunement is only available if the item can be attuned, and will be locked to the same value as requires magic if the item requires attunement because by the rules, if attunement is required then all magical features of the item are locked behind attunement
- Requires magic is only available if the item has the "magical" property, unless magical isn't in the list of valid properties
- Requires identification is only available for items that can be identified
- Cast activity always has "Requires Magic" checked

Since `TransformActivity` and `SummonActivity` already had level limits, the associated identifier has been migrated to match the new visibility identifier available on all activities (the level limits are still available on each profile, since those are still useful).

Closes #4963


Zanderaf:
Could we maybe get a "Concentration" option for visibility gating? Some activities dont make sense on spells like moving moonbeam or repeating damage if a spell needs concentration to be active. Would be nice to have "concentration" an activity visibility gate on all items honestly.

arbron:
> Could we maybe get a "Concentration" option for visibility gating? Some activities dont make sense on spells like moving moonbeam or repeating damage if a spell needs concentration to be active. Would be nice to have "concentration" an activity visibility gate on all items honestly.

That will be part of the improvements to item activation tracking: https://github.com/foundryvtt/dnd5e/issues/4425

Zanderaf:
> > Could we maybe get a "Concentration" option for visibility gating? Some activities dont make sense on spells like moving moonbeam or repeating damage if a spell needs concentration to be active. Would be nice to have "concentration" an activity visibility gate on all items honestly.
> 
> That will be part of the improvements to item activation tracking: #4425

Got it! Was just curious on whether or not we'd be getting something along those lines, theres a fair few features / spells that could use something like that, stuff like Vampiric Touch, Ennervation, Moonbeam, and basically anything you can activate and repeat.

lang/en.json
```diff
@@ -285,6 +285,35 @@
     },
     "name": {
       "label": "Name"
+    },
+    "visibility": {
+      "identifier": {
+        "hint": "Identifier used to determine whether the character level or a specific class level should be used for visibility level limits.",
```
```suggestion
        "hint": "The identifier of the class that level limits apply to. If left blank, the character level is used.",
```

module/applications/activity/activity-sheet.mjs
```diff
@@ -336,6 +337,29 @@ export default class ActivitySheet extends PseudoDocumentSheet {
...
+      input: context.inputs.createCheckboxInput,
+      value: lockedValue ?? context.source.visibility[name]
+    } : null;
+    const itemSystem = this.activity.item.system;
+    const isRider = this.activity.isRider;
+    context.visibilityFields = [
+      // Only show "Require Attunement" if item has an attunement option
+      ["required", "optional"].includes(itemSystem.attunement) || isRider ? addField("requireAttunement",
+        // If item requires attunement, then the "Require Attunement" option is locked to the "Require Magic" option
+        !isRider && itemSystem.attunement === "required"
```
Fyorl:
```suggestion
        !isRider && (itemSystem.attunement === "required")
```

module/applications/activity/cast-sheet.mjs
```diff
@@ -75,6 +75,8 @@ export default class CastSheet extends ActivitySheet {
       input: context.inputs.createCheckboxInput
     }];
     if ( context.spell ) context.placeholder = { name: context.spell.name, img: context.spell.img };
+    const requireMagicField = context.visibilityFields.find(f => f.field.name === "requireMagic");
+    if ( requireMagicField ) requireMagicField.disabled = true;
```
Fyorl: Do we also need to force `value` to `true`?


```suggestion
    if ( requireMagicField ) Object.assign(requireMagicField, { disabled: true, value: true });
```

module/data/item/templates/equippable-item.mjs
```diff
@@ -43,6 +43,16 @@ export default class EquippableItemTemplate extends SystemDataModel {
   /*  Properties                                  */
   /* -------------------------------------------- */
 
+  /**
+   * This item is capable of being attuned.
+   * @type {boolean}
+   */
+  get canAttune() {
+    return ["required", "optional"].includes(this.attunement);
```
Fyorl: We can avoid some throwaway arrays where it's not so arduous to do so:


```suggestion
    return (this.attunement === "required") || (this.attunement === "optional");
```

PR #5904, opened 2025-07-30
[#5374] Convert Group sheet to Application V2.

### LINKED ISSUE #5374: Redesign group sheets and convert to ApplicationV2

Fyorl:
- Closes #5374 

image


module/data/actor/group.mjs
```diff
@@ -168,16 +199,17 @@ export default class GroupActor extends ActorDataModel.mixin(CurrencyTemplate) {
 
   /** @inheritDoc */
   prepareDerivedData() {
-    const system = this;
     Object.defineProperty(this.details.xp, "derived", {
       get() {
-        return system.type.value === "encounter" ? system.members.reduce((xp, { actor, quantity }) =>
+        return this.type.value === "encounter" ? this.members.reduce((xp, { actor, quantity }) =>
```
arbron: Inside the `defineProperty` getter, `this` is equal to the object it is define on (so in this case, `this.details.xp`), that is why system was defined separately. Currently this throws an error because of that.

templates/actors/group/header.hbs
```diff
@@ -0,0 +1,111 @@
...
+                <div class="header">
+                    <h3>{{ localize "DND5E.AdvancementTitle" }}</h3>
+                    <button type="button" class="config-button unbutton" data-action="award" data-tooltip
+                            aria-label="{{ localize "DND5E.Group.Distribute" }}">
+                        <i class="fa-solid fa-trophy" inert></i>
+                    </button>
+                </div>
+                <div class="info">
+                    <div>
+                        <span class="label">{{ localize "DND5E.ExperiencePoints.Pool" }}</span>
```
arbron, line 90: We can probably hide the whole XP part if the Leveling without XP setting is selected.

dnd5e.mjs
```diff
@@ -594,6 +594,12 @@ Hooks.on("renderJournalEntryPageSheet", applications.journal.JournalEntrySheet5e
 
 Hooks.on("renderActiveEffectConfig", documents.ActiveEffect5e.onRenderActiveEffectConfig);
 
+Hooks.on("renderDocumentSheetConfig", (app, html) => {
+  if ( (app.options.document instanceof foundry.documents.Actor) && (app.options.document.type === "group") ) {
+    dnd5e.applications.actor.GroupActorSheet.addDocumentSheetConfigOptions(app, html);
```
arbron: Can be a bit shorter since we've already imported `applications` here:

```suggestion
    applications.actor.GroupActorSheet.addDocumentSheetConfigOptions(app, html);
```

less/v2/items.less
```diff
@@ -103,8 +104,8 @@
       display: block;
       width: 96px;
       height: 96px;
-      border-radius: 100%;
-      border: 4px solid var(--dnd5e-color-gold);
+      border-radius: 8px;
```
arbron, line 108: Removing the circle generally looks nice, but it does put the portrait very close to the mode toggle. Perhaps it would be worth it to give the top bar a few more pixels of margin before the header section starts.

module/applications/actor/group-sheet.mjs
```diff
@@ -1,433 +1,487 @@
...
+    actions: {
+      award: GroupActorSheet.#onAward,
+      editDescription: GroupActorSheet.#onEditDescription,
+      placeMembers: GroupActorSheet.#onPlaceMembers,
+      removeMember: GroupActorSheet.#onRemoveMember,
+      roll: GroupActorSheet.#onRoll
+    },
+    tab: "members"
+  };
+
```
arbron, line 26:
```suggestion

  /* -------------------------------------------- */

```

module/applications/actor/group-sheet.mjs
```diff
@@ -1,433 +1,487 @@
...
+      container: { classes: ["tab-body"], id: "tabs" },
+      template: "systems/dnd5e/templates/actors/group/biography.hbs",
+      scrollable: [""]
+    },
+    tabs: {
+      id: "tabs",
+      classes: ["tabs-right"],
+      template: "systems/dnd5e/templates/shared/sidebar-tabs.hbs"
+    }
+  };
```
arbron, line 60:
```suggestion

  /* -------------------------------------------- */

```

module/applications/actor/group-sheet.mjs
```diff
@@ -1,433 +1,487 @@
...
 
   /**
-   * IDs for items on the sheet that have been expanded.
-   * @type {Set<string>}
-   * @protected
+   * Application tabs.
+   * @type {SheetTabDescriptor5e}
```
arbron: `PrimarySheetMixin` provides the TABS, so this can be `/** @override */`.

module/applications/actor/group-sheet.mjs
```diff
@@ -1,433 +1,487 @@
...
+   * Description currently being edited.
+   * @type {string|null}
    */
-  static unsupportedItemTypes = new Set(["background", "race", "class", "subclass", "feat"]);
+  editingDescriptionTarget = null;
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  tabGroups = {
```
arbron, line 83: The character and NPC sheets include this above with the static properties, but since it is an instance property it would probably make more sense to adjust those other two sheets to move this down to match this one.

module/applications/actor/group-sheet.mjs
```diff
@@ -1,433 +1,487 @@
...
-    const members = [];
-    if ( stats.nMembers ) {
-      members.push(`${stats.nMembers} ${game.i18n.localize(`DND5E.Group.Member.${rule.select(stats.nMembers)}`)}`);
-    }
-    if ( stats.nVehicles ) {
-      members.push(`${stats.nVehicles} ${game.i18n.localize(`DND5E.Group.Vehicle.${rule.select(stats.nVehicles)}`)}`);
-    }
-    if ( !members.length ) return game.i18n.localize("DND5E.GroupSummaryEmpty");
-    return game.i18n.format("DND5E.GroupSummary", {members: formatter.format(members)});
+  async _prepareCharacterContext(actor, context, options) {
```
arbron: For the broad organization of the other new actor sheets, we've moved these more specific preparation methods into their own section beneath the "Rendering" section:

- Rendering (for part context preparation and other general application rendering methods)
- Actor Preparation Helpers (for preparation methods related to the actor itself)
- Item Preparation Helpers (for preparation methods related to individual items on the actor)

module/applications/actor/npc-sheet.mjs
```diff
@@ -180,8 +180,8 @@ export default class NPCActorSheet extends BaseActorSheet {
       secrets: this.actor.isOwner, relativeTo: this.actor, rollData: context.rollData
     };
     context.enriched = {
-      public: await TextEditor.enrichHTML(this.actor.system.details.biography.public, enrichmentOptions),
-      value: await TextEditor.enrichHTML(this.actor.system.details.biography.value, enrichmentOptions)
+      public: await CONFIG.ux.TextEditor.enrichHTML(this.actor.system.details.biography.public, enrichmentOptions),
+      value: await CONFIG.ux.TextEditor.enrichHTML(this.actor.system.details.biography.value, enrichmentOptions)
```
arbron: We do have `const TextEditor = foundry.applications.ux.TextEditor.implementation;` at the top of this file, so we aren't calling the global here.

module/data/actor/group.mjs
```diff
@@ -3,6 +3,8 @@ import ActorDataModel from "../abstract/actor-data-model.mjs";
 import FormulaField from "../fields/formula-field.mjs";
 import CurrencyTemplate from "../shared/currency.mjs";
 import GroupSystemFlags from "./group-system-flags.mjs";
+import { convertLength } from "../../utils.mjs";
+import MovementField from "../shared/movement-field.mjs";
```
arbron:
```suggestion
import { convertLength } from "../../utils.mjs";
import FormulaField from "../fields/formula-field.mjs";
import CurrencyTemplate from "../shared/currency.mjs";
import MovementField from "../shared/movement-field.mjs";
import GroupSystemFlags from "./group-system-flags.mjs";
```

(the utils import should actually be one line higher beneath `TokenPlacement`, but GitHub won't let me suggest that line ¯\_(ツ)_/¯)

module/data/actor/group.mjs
```diff
@@ -73,9 +77,24 @@ export default class GroupActor extends ActorDataModel.mixin(CurrencyTemplate) {
...
+          air: new NumberField({
+            nullable: false, min: 0, step: 0.1, initial: 0, speed: true, label: "DND5E.MovementAir"
+          }),
+          pace: new StringField({
+            required: true, blank: false, initial: "normal", choices: () => CONFIG.DND5E.travelPace,
+            label: "DND5E.Travel.Label"
+          }),
+          units: new StringField({
+            required: true, nullable: true, blank: false, label: "DND5E.MovementUnits", initial: () => {
+              return game.settings.get("dnd5e", "metricLengthUnits") ? "km" : "mi";
```
arbron, line 94: We should probably add another entry to `CONFIG.DND5E.defaultUnits` for travel pace. That would allow us to use the `defaultUnits` utility method to match elsewhere.

module/data/actor/vehicle.mjs
```diff
@@ -5,6 +5,7 @@ import AttributesFields from "./templates/attributes.mjs";
 import CommonTemplate from "./templates/common.mjs";
 import DetailsFields from "./templates/details.mjs";
 import TraitsFields from "./templates/traits.mjs";
+import MovementField from "../shared/movement-field.mjs";
```
arbron: Should be the first import above the other import from `"../shared"`.

module/utils.mjs
```diff
@@ -49,10 +49,12 @@ export function formatModifier(mod) {
  * @param {boolean} [options.numerals]  Format the number as roman numerals.
  * @param {boolean} [options.ordinal]   Use ordinal formatting.
  * @param {boolean} [options.words]     Write out number as full word, if possible.
+ * @param {string} [options.blank]      Format a zero or otherwise empty value as the given string.
  * @returns {string}
  */
-export function formatNumber(value, { numerals, ordinal, words, ...options }={}) {
+export function formatNumber(value, { numerals, ordinal, words, blank, ...options }={}) {
```
arbron: Love me some alphabetization.

```suggestion
export function formatNumber(value, { blank, numerals, ordinal, words, ...options }={}) {
```

module/utils.mjs
```diff
@@ -49,10 +49,12 @@ export function formatModifier(mod) {
  * @param {boolean} [options.numerals]  Format the number as roman numerals.
  * @param {boolean} [options.ordinal]   Use ordinal formatting.
  * @param {boolean} [options.words]     Write out number as full word, if possible.
+ * @param {string} [options.blank]      Format a zero or otherwise empty value as the given string.
```
arbron:
```suggestion
 * @param {string} [options.blank]      Format a zero or otherwise empty value as the given string.
 * @param {boolean} [options.numerals]  Format the number as roman numerals.
 * @param {boolean} [options.ordinal]   Use ordinal formatting.
 * @param {boolean} [options.words]     Write out number as full word, if possible.
```

templates/actors/group/inventory.hbs
```diff
@@ -0,0 +1,44 @@
...
+            <div class="meter progress" role="meter" aria-valuemin="0" aria-valuenow="{{ pct }}"
+                 aria-valuetext="{{ value }}" aria-valuemax="{{ max }}" style="--bar-percentage: {{ pct }}%">
+                <div class="label">
+                    <i class="fa-solid fa-weight-hanging" inert></i>
+                    <span class="value">{{ value }}</span>
+                    <span class="separator">&sol;</span>
+                    <span class="max">{{ max }}</span>
+                </div>
+            </div>
+            {{/with}}
```
arbron: Looks like we could use `systems/dnd5e/templates/inventory/encumbrance.hbs` for this, we'd just need to add a options to hide the breakpoints.

templates/actors/group/member.hbs
```diff
@@ -0,0 +1,116 @@
...
+            <img src="{{ portrait.src }}" alt="{{ name }}">
+            {{/if}}
+        </div>
+        <div class="bars">
+            <strong class="name">
+                <a class="rollable" data-action="showDocument" data-uuid="{{ uuid }}">{{ name }}</a>
+            </strong>
+            {{#with system.attributes.hp}}
+            <div class="hp-bar bar">
+                <label>HP</label>
```
arbron:
```suggestion
                <label>{{ localize "DND5E.HP" }}</label>
```

templates/actors/group/member.hbs
```diff
@@ -0,0 +1,116 @@
...
+                <div class="values">
+                    <span class="value">{{ value }}</span>
+                    <span class="separator">&sol;</span>
+                    <span class="max">{{ max }}</span>
+                </div>
+            </div>
+            {{/with}}
+            {{#with system.attributes.hd}}
+            <div class="hd-bar bar">
+                <label>HD</label>
```
arbron: Strangely we don't have a hit dice abbreviation localization yet, so you'll need to add this one:

```suggestion
                <label>{{ localize "DND5E.HITDICE.Abbreviation" }}</label>
```

templates/actors/group/member.hbs
```diff
@@ -0,0 +1,116 @@
...
+                </li>
+                {{/each}}
+            </ul>
+            {{/if}}
+        </div>
+        <div class="classes">
+            {{#if @root.editable}}
+            <button type="button" class="config-button unbutton" data-action="removeMember" data-tooltip
+                    aria-label="{{ localize "DND5E.Group.Action.Remove" }}">
+                <i class="fa-solid fa-square-xmark"></i>
```
arbron:
```suggestion
                <i class="fa-solid fa-square-xmark" inert></i>
```

templates/actors/group/tabs.hbs
```diff
@@ -0,0 +1,9 @@
+<nav data-group="primary">
```
arbron: Is this any different than the `systems/dnd5e/templates/shared/sidebar-tabs.hbs` we use for the other sheets?

arbron (changes requested):

module/data/actor/group.mjs
Fyorl: I also took this opportunity to rename this class to `GroupData`. It was kind of the odd-one-out. I don't actually know if a deprecation is even necessary. I struggle to imagine anything is subclassing this. I added one anyway, just in case.

module/applications/actor/group-sheet.mjs
Fyorl: Not 100% on the demarcation between these methods. I've re-arranged them. Let me know if it's what you had in mind.

module/data/actor/group.mjs
Fyorl: Seems this import wasn't actually used.

module/data/actor/group.mjs
Fyorl: `defaultUnits` method also checks the type against a system setting, so we would need to add a new setting too, no? But having separate setting for metric length units and metric 'travel' units seems like too much.

templates/actors/group/member.hbs
Fyorl: Thank you for catching these leftovers from prototyping.

templates/actors/group/tabs.hbs
Fyorl: Another leftover from before the other sheets even became V2. It's not actually used in `GroupActorSheet.PARTS` so it has been deleted.

module/data/actor/group.mjs
arbron: I suppose we could do some special casing in `defaultUnits` to use: `metric${(type === "travelPace" ? "length" : type).capitalize()}Units` for the settings key and piggyback on the normal length units. Mainly I think it would be good to have them configurable in case people want to add some special fantasy units.

module/applications/actor/group-sheet.mjs
arbron: Yeah, that matches what I did on the other sheets.

templates/actors/group/header.hbs
```diff
@@ -78,12 +78,15 @@
             <div class="xp card">
                 <div class="header">
                     <h3>{{ localize "DND5E.AdvancementTitle" }}</h3>
+                    {{#if showXP}}
```
arbron: It would be nice to keep the award button, I use that when I'm running my game because it creates a chat message record of the award (as opposed to no record with the normal "Transfer Currency" control), which is useful for Adventurer's League logs.

Fyorl: Ah, did not realise the `Award` dialog already had logic for this. I've re-added the trophy button but modified the tooltip appropriately depending on levelling mode.

Zarkoncheg:
Does the new group sheet support specifying the quantity?
The Tomb of annihilation module implements encounters with a group of monsters, which are implemented through the group character. I don't see a setting a quantity of monsters.

PR #5905, opened 2025-07-30
Add support for requesting group skill checks via group actor rollSkill.

Fyorl:
API-only at the moment


templates/chat/request-card.hbs
```diff
@@ -10,7 +10,9 @@
                     <span class="title">{{ actor.name }}</span>
                 </span>
                 <div class="status">
-                    {{#if completed}}
+                    {{#if total includeZero=true}}
+                    <span class="total">{{ total }}</span>
```
arbron, line 14: It would be nice to have some success/failure coloring if a target is provided when the skill is rolled.

module/data/chat-message/request-message-data.mjs
```diff
@@ -87,7 +90,7 @@ export default class RequestMessageData extends ChatMessageDataModel {
    */
   static async #handleRequest(event, target) {
     const actor = fromUuidSync(target.closest("[data-uuid]").dataset.uuid);
-    const result = await CONFIG.DND5E.requests[this.handler](actor, this.parent, this.data);
+    const result = await CONFIG.DND5E.requests[this.handler](actor, this.parent, this.data, { event });
```
arbron, line 121: Right now the type for `CONFIG.DND5E.requests` is just `Record<string, Function>`, it would be good to properly document the expected API for the callback.

arbron (approved):
Looks good, nothing urgent in my comments.

templates/chat/request-card.hbs
Fyorl: Added this

image

module/data/chat-message/request-message-data.mjs
```diff
@@ -87,27 +118,49 @@ export default class RequestMessageData extends ChatMessageDataModel {
    */
   static async #handleRequest(event, target) {
     const actor = fromUuidSync(target.closest("[data-uuid]").dataset.uuid);
-    const result = await CONFIG.DND5E.requests[this.handler](actor, this.parent, this.data);
-    if ( result instanceof ChatMessage ) {
+    const result = await CONFIG.DND5E.requests[this.handler](actor, this.parent, this.data, { event });
+    if ( (result instanceof ChatMessage) && !result.getFlag("dnd5e", "requestResult") ) {
```
Fyorl, line 122: Workaround for https://github.com/foundryvtt/foundryvtt/issues/13067. We try to add the flags during message creation now, and I've added an associated `createChatMessage` hook.

module/config.mjs
```diff
@@ -4892,12 +4892,26 @@ DND5E.defaultArtwork = {
 /*  Requests                                    */
 /* -------------------------------------------- */
 
+/**
+ * @callback RequestCallback5e
+ * @param {Actor5e} actor               The actor fulfilling the request.
+ * @param {ChatMessage5e} request       The request message.
+ * @param {object} config               Additional request configuration.
+ * @param {RequestOptions5e} [options]  Additional options provided at fulfillment time.
+ */
```
arbron, line 4902:
```suggestion
 * @returns {Promise<ChatMessage5e>}    Result chat message that will be associated with request.
 */
```

arbron (approved):

PR #5941, opened 2025-08-10
Modify `base` AE type with system-specific data

arbron:
This new effect type replaces the base type provided by core that adds a `magical` toggle and moves rider statuses from flags into the system data. An in-memory migration handles changing the type and moving the status riders from flags, while system migration handles setting a logical default for the new `magical` property.

Creates `ActiveEffectDataModel` that new can extend to gain some default functionality. This new model includes getters for the actor and item containing the effect, as well as `applicableType` which returns the document type to which the effect should apply. This allows some of the previous explicit checks for enchantments and rider effects to be handled by the specific types and supports  module-provided AE types that can apply to actors, items, or neither.
    
Also moves some of the enchantment-specific creation & deletion logic into the system data model instead.

### DIFF
--- a/module/data/active-effect/_types.mjs
+++ b/module/data/active-effect/_types.mjs
@@ -1,0 +1,11 @@
+/**
+ * @typedef EnchantmentActiveEffectSystemData
+ * @property {boolean} magic               Does this effect originate from a magical source?
+ */
+
+/**
+ * @typedef StandardActiveEffectSystemData
+ * @property {boolean} magic               Does this effect originate from a magical source?
+ * @property {object} rider
+ * @property {Set<string>} rider.statuses  Additional status effects that are separately applied when effect is applied.
+ */
--- a/module/data/active-effect/_types.mjs
+++ b/module/data/active-effect/_types.mjs
@@ -1,11 +1,13 @@
 /**
- * @typedef EnchantmentActiveEffectSystemData
- * @property {boolean} magic               Does this effect originate from a magical source?
- */
-
-/**
- * @typedef StandardActiveEffectSystemData
+ * @typedef baseActiveEffectSystemData
+ * @property {EffectChangeData[]} changes  Changes to apply to the actor.
  * @property {boolean} magic               Does this effect originate from a magical source?
  * @property {object} rider
  * @property {Set<string>} rider.statuses  Additional status effects that are separately applied when effect is applied.
+ */
+
+/**
+ * @typedef EnchantmentActiveEffectSystemData
+ * @property {EffectChangeData[]} changes  Changes to apply to the item.
+ * @property {boolean} magic               Does this effect originate from a magical source?
  */
--- a/module/data/active-effect/_types.mjs
+++ b/module/data/active-effect/_types.mjs
@@ -1,5 +1,5 @@
 /**
- * @typedef baseActiveEffectSystemData
+ * @typedef BaseActiveEffectSystemData
  * @property {EffectChangeData[]} changes  Changes to apply to the actor.
  * @property {boolean} magic               Does this effect originate from a magical source?
  * @property {object} rider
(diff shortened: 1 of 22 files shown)


arbron:
### Todo
- Add something to handle `temporary` flag added here: https://github.com/foundryvtt/dnd5e/pull/6949

module/data/active-effect/_types.mjs
```diff
@@ -0,0 +1,13 @@
+/**
+ * @typedef BaseActiveEffectSystemData
+ * @property {EffectChangeData[]} changes  Changes to apply to the actor.
+ * @property {boolean} magic               Does this effect originate from a magical source?
```
Fyorl:
```suggestion
 * @property {boolean} magical              Does this effect originate from a magical source?
```

module/data/active-effect/base.mjs
```diff
@@ -0,0 +1,53 @@
+import ActiveEffectDataModel from "../abstract/active-effect-data-model.mjs";
+
+const { BooleanField, SchemaField, SetField, StringField } = foundry.data.fields;
+
+/**
+ * @import { BaseActiveEffectSystemData } from "./types.mjs";
```
Fyorl:
```suggestion
 * @import { BaseActiveEffectSystemData } from "./_types.mjs";
```

module/data/active-effect/enchantment.mjs
```diff
@@ -95,6 +154,62 @@ export default class EnchantmentData extends (ActiveEffectTypeDataModel ?? TypeD
...
+    toRemove.forEach(f => f.remove());
+  }
+
+  /* -------------------------------------------- */
+  /*  Socket Event Handlers                       */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _preCreate(data, options, user) {
+    if ((await super._preCreate(data, options, user)) === false) return false;
```
Fyorl:
```suggestion
    if ( (await super._preCreate(data, options, user)) === false ) return false;
```

module/data/active-effect/enchantment.mjs
```diff
@@ -95,6 +154,62 @@ export default class EnchantmentData extends (ActiveEffectTypeDataModel ?? TypeD
...
+        .forEach(element => element.buildItemList());
+    }
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _onDelete(options, userId) {
+    super._onDelete(options, userId);
+    if ( this.isApplied ) dnd5e.registry.enchantments.untrack(this.origin, this.uuid);
```
Fyorl:
```suggestion
    if ( this.isApplied ) dnd5e.registry.enchantments.untrack(this.parent.origin, this.parent.uuid);
```

module/documents/active-effect.mjs
```diff
@@ -444,11 +464,7 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
    * @returns {Promise<ActiveEffect5e[]>}      Created rider effects.
    */
   async createRiderConditions() {
-    const riders = new Set();
-
-    for ( const status of this.getFlag("dnd5e", "riders.statuses") ?? [] ) {
-      riders.add(status);
-    }
+    const riders = new Set(this.system.riders?.statuses ?? []);
```
```suggestion
    const riders = new Set(this.system.rider?.statuses ?? []);
```

module/migration.mjs
```diff
@@ -739,8 +739,11 @@ export function migrateEffectData(effect, migrationData, { parent }={}) {
   const updateData = {};
   _migrateDocumentIcon(effect, updateData, {...migrationData, field: "img"});
   _migrateEffectArmorClass(effect, updateData);
+  if ( foundry.utils.isNewerVersion("5.2.0", effect._stats?.systemVersion ?? parent?._stats?.systemVersion) ) {
```
Fyorl:
```suggestion
  if ( foundry.utils.isNewerVersion("6.0.0", effect._stats?.systemVersion ?? parent?._stats?.systemVersion) ) {
```

Fyorl (approved):
Good cleanup

PR #5948, opened 2025-08-11
[#5429] Use transformation source to-hit bonus when transformed.

### LINKED ISSUE #5429: Wild Shaped actors incorrectly uses proficiency bonus of the druid rather than beast
Foundry 12.331
dnd5e 4.3.6

Repro:
1. Open a level 5 druid PC sheet
2. Drag a Wolf sheet from PHB or SRD onto the druid sheet
3. Leave all the transform settings as default or unchecked 
4. Have transformed Wolf make a bite attack

Expected:
* Bite attack roll is 1d20 + 2 + 2
* 2 from STR and 2 from Wolf PB

Actual:
* Bite attack roll is 1d20 + 2 + 3
* 2 from STR and 3 from Druid PB

Fyorl:
- Closes #5429

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -2870,7 +2870,16 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       if ( settings.keep.has("languages") ) d.system.traits.languages = o.system.traits.languages;
 
       // Keep specific items from the original data
-      d.items = d.items.concat(o.items.filter(i => {
+      const profDiff = source.system.attributes.prof - this.system.attributes.prof;
+      d.items = d.items.map(i => {
+        if ( settings.keep.has("class") && ((i.type === "feat") || (i.type === "weapon")) ) {
+          // Items gained from the source should use the source's proficiency bonus.
+          Object.values(i.system.activities).forEach(activity => {
+            if ( activity.type === "attack" ) activity.attack.bonus += ` ${profDiff < 0 ? "" : "+"}${profDiff}`;
+          });
+        }
+        return i;
+      }).concat(o.items.filter(i => {
         switch ( i.type ) {
           case "class":
           case "subclass": return settings.keep.has("class") || settings.keep.has("hp");


module/documents/actor/actor.mjs
```diff
@@ -2870,7 +2870,17 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
 
       // Keep specific items from the original data
-      d.items = d.items.concat(o.items.filter(i => {
+      const profDiff = source.system.attributes.prof - this.system.attributes.prof;
+      const numberFormat = new Intl.NumberFormat(game.i18n.lang, { signDisplay: "always" });
+      d.items = d.items.map(i => {
+        if ( settings.keep.has("class") && ((i.type === "feat") || (i.type === "weapon")) ) {
+          // Items gained from the source should use the source's proficiency bonus.
+          Object.values(i.system.activities).forEach(activity => {
+            if ( activity.type === "attack" ) activity.attack.bonus += numberFormat.format(profDiff);
```
arbron: Seems like this could result in some numbers we cannot parse depending on the language the user has selected:

```javascript
(new Intl.NumberFormat("ne", { signDisplay: "always" })).format(5) // Result: +५
// (for all our Nepali users)
```

arbron (changes requested):

module/documents/actor/actor.mjs
```diff
@@ -2870,7 +2870,16 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       if ( settings.keep.has("languages") ) d.system.traits.languages = o.system.traits.languages;
 
       // Keep specific items from the original data
-      d.items = d.items.concat(o.items.filter(i => {
+      const profDiff = source.system.attributes.prof - this.system.attributes.prof;
+      d.items = d.items.map(i => {
+        if ( settings.keep.has("class") && ((i.type === "feat") || (i.type === "weapon")) ) {
+          // Items gained from the source should use the source's proficiency bonus.
+          Object.values(i.system.activities).forEach(activity => {
+            if ( activity.type === "attack" ) activity.attack.bonus += `${profDiff < 0 ? "" : "+"}${profDiff}`;
```
arbron: Space needed in case the existing bonus ends in a `@` value (`@prof-13` won't evaluate properly):

```suggestion
            if ( activity.type === "attack" ) activity.attack.bonus += ` ${profDiff < 0 ? "" : "+"}${profDiff}`;
```

arbron (changes requested):

module/documents/actor/actor.mjs
Fyorl: These are good catches, thank you

arbron (approved):

PR #5949, opened 2025-08-11
[#5419] When keep HP or spells transformation options are set, retain these values when reverting form.

### LINKED ISSUE #5419: Wild Shaped druid reverted wont retain the transformed actors' hit points
Since you retain Hit Points with the '24 Wild Shape, it would be nice if modern rules could also change that the hit points in the transformed actor are retained when reverting.

Fyorl:
- Closes #5419

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -3100,12 +3100,23 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
      */
     Hooks.callAll("dnd5e.revertOriginalForm", this, options);
 
+    const transformOptions = this.getFlag("dnd5e", "transformOptions");
     const previousActorIds = this.getFlag("dnd5e", "previousActorIds") ?? [];
     const isOriginalActor = !previousActorIds.length;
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -3100,13 +3100,24 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
     // Obtain a reference to the original actor
     const original = game.actors.get(this.getFlag("dnd5e", "originalActor"));
 
+    const update = {};
+    if ( transformOptions?.keep?.includes("hp") ) {
+      foundry.utils.setProperty(update, "system.attributes.hp.value", this.system.attributes.hp.value);
+    }
+    if ( transformOptions?.keep?.includes("spells") ) {
+      Object.entries(this.system.spells ?? {}).forEach(([k, v]) => {
+        if ( v.max ) update[k] = v.value;
```
```suggestion
        if ( v.max ) update[`system.spells.${k}.value`] = v.value;
```

arbron (changes requested):

arbron (approved):

PR #5953, opened 2025-08-12
Store message request targets by UUID rather than ID.

Fyorl:
This is to allow for requesting rolls for synthetic Actors, not compendium Actors. It should theoretically still be possible then to evaluate request targets synchronously with `fromUuidSync` if necessary.

Let me know if you foresee any issues with us porting the ToA requests to this API with these changes.

### DIFF
--- a/module/data/chat-message/request-message-data.mjs
+++ b/module/data/chat-message/request-message-data.mjs
@@ -97,16 +97,15 @@ export default class RequestMessageData extends ChatMessageDataModel {
       content: await foundry.applications.ux.TextEditor.implementation.enrichHTML(
         this.parent.content, { rollData: this.parent.getRollData() }
       ),
-      targets: (await Promise.all(this.targets.map(async t => {
-        if ( !t.actor ) return null;
-        const actor = await fromUuid(t.actor);
+      targets: this.targets.map(t => {
+        const actor = fromUuidSync(t.actor);
         if ( !actor ) return null;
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -2384,7 +2384,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       }
     };
     if ( config.request ) foundry.utils.setProperty(chatData, "flags.dnd5e.requestResult", {
-      actorId: this.id, requestId: config.request.id
+      actorUUID: this.uuid, requestId: config.request.id
```
krbz999: Could we do `actorUuid` rather than `actorUUID`, I feel like that's used more often

module/data/chat-message/request-message-data.mjs
```diff
@@ -36,7 +38,7 @@ export default class RequestMessageData extends ChatMessageDataModel {
       data: new ObjectField(),
       handler: new StringField({ required: true, blank: false }),
       targets: new ArrayField(new SchemaField({
-        actor: new ForeignDocumentField(foundry.documents.BaseActor),
+        actor: new DocumentUUIDField({ type: "Actor" }),
```
arbron: I feel like we should add a validation here to prevent compendium UUIDs from being used. That would allow us to stick with `fromUuidSync` throughout the code while still allowing synthetic actors to be referenced. I can't think of a reason that compendium actors would be necessary since they generally can't roll anyway.

arbron:
Swapped back to `fromUuidSync` in a few places and removed `async` code where possible.

PR #5996, opened 2025-08-14
Item Choice Flow: Any Available Level option is not working correctly

### LINKED ISSUE #5995: Item Choice Flow: Any Available Level option is not working correctly
It looks like in `ItemChoiceFlow._maxSpellSlotLevel`, the `progression` variable initializes a `slots` member to 0 instead of the `spell` member, making this fail to return anything > 0 for non-pact casters.

righthandofvecna:
Fix #5995

### DIFF
--- a/module/applications/advancement/item-choice-flow.mjs
+++ b/module/applications/advancement/item-choice-flow.mjs
@@ -325,7 +325,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
 
     // For advancements on classes or subclasses, use the largest slot available for that class
     if ( spellcasting?.type ) {
-      const progression = { spell: 0, pact: 0 };
+      const progression = Object.fromEntries(Object.keys(CONFIG.DND5E.spellcasting).map(k => [k, 0]));
       const maxSpellLevel = Object.keys(CONFIG.DND5E.spellLevels).length - 1;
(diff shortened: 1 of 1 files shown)


module/applications/advancement/item-choice-flow.mjs
```diff
@@ -325,7 +325,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
 
     // For advancements on classes or subclasses, use the largest slot available for that class
     if ( spellcasting?.type ) {
-      const progression = { slot: 0, pact: 0 };
+      const progression = { spell: 0, pact: 0 };
```
krbz999: hm, if this doesnt work for regular spell slots, is it currently also not functional for types added by modules?

righthandofvecna: Very likely, yeah. This fixes it for the normal types that I've tested, but maybe it should be updated to instead initialise based on that `CONFIG.DND5E.spellcasting` constant

righthandofvecna: Something like [7583d0d](https://github.com/foundryvtt/dnd5e/pull/5996/commits/7583d0db3e012ab7b8259e6f13dece145305202f)?

Fyorl (approved):

PR #5998, opened 2025-08-14
[#5934] Add setting for players to use effect application tray

### LINKED ISSUE #4529: Enchantment on other players features
A player of mine use a spell with a Enchantment that target a weapon of another player. When the other player tried to enchant his weapon, he couldn't do it and the error bellow was showed.

image

Makes no sense that only the GM can apply enchantment on items that was not from his features. There's a lot of spells that a player can enchant other players items and with this, the GM has always to manually enchant everyones items.

### LINKED ISSUE #5934: Allow players access to the Effect Application tray under some circumstances
Similar to #3300, but for effect application rather than damage application.
Related (perhaps a different solution to the same root issue): #2847 

Breaking up various types of effect application and how they might better be treated below (assuming just changing the application tray, not auto-granting or socketing anything).
- Activity used with `self` set as target type (example: Rage): Should offer tray to owners of the actor, similar to current behavior with Rage as a transfer effect but without actually needing it to be a transfer effect.
- Activity with non-`self` target type, but self-targeted (potential example: Mage Armor): Same as above.
- Activity with non-`self` target type, only unowned tokens targeted (example: also Mage Armor, Bless, most "hostile" spells): Only show tray for GM (current behavior).
- Activity with non-`self` target type, a mix of owned & unowned tokens targeted (example: also Bless): Entirely possible this gets rolled up with the "non-`self`, but self-targeted" scenario. Not sure the "perfect" behavior here, but perhaps a variation of the effect application tray which force-unchecks any unowned targets. If both player and GM try to apply effect, no real issue, since effects can't "stack." If ever that did become an issue, special treatment to either update the tray to un-check already-applied-to targets on other machines, or to prevent multi-application from the same originating chat card.

Naturally, if it _were_ determined that socketing effect application should exist as a setting, all scenarios would boil down to "show effect tray if they have permissions to see it."

Potential sticking point: Concentration-dependent effects. Bless as the example, if an ally tried to apply a bless which targeted them, that would require updating the concentration effect on the _caster_, which they might not have ownership of. I think _that_ would be an entirely valid use of socketing (and in the event that no valid owner is online, either throw an error as is currently coded, or simply ignore the concentration effect).

roth-michael:
Closes #5934 
Closes #4529 by happenstance.

Setting enabled by default.
Players will see the effect application tray if:
- It's a transfer effect & they own the chat card (existing behavior)
- The setting is enabled, the chat card is owned by the user, and the associated activity targets self
- The setting is enabled and the chat card's targets include at least one owned actor of the user

Other changes to existing behavior:
Concentration-dependent effects (including enchantments) no longer update the concentration effect when applied via chat tray (or chat enchant drag-drop). Instead, the effects are given a `flags.dnd5e.dependsOn` flag with the UUID of the concentration effect. Then, in `_preCreate`, if that flag is populated but there is no active GM, the creation is blocked with a warning (similar to how breaking concentration is treated when there are dependent effects on unowned actors). In `_onCreate`, if the flag is populated and the user is the active GM, _then_ the concentration effect is updated to reflect its dependent(s). This update is done via Semaphore to avoid race conditions when multiple concentration-dependent effects are created in quick succession.

### DIFF
--- a/module/data/active-effect/base.mjs
+++ b/module/data/active-effect/base.mjs
@@ -35,6 +35,16 @@ export default class BaseEffectData extends ActiveEffectDataModel {
   }
 
   /* -------------------------------------------- */
+  /*  Data Preparation                            */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  prepareBaseData() {
+    super.prepareBaseData();
+    if ( this.isOnActivity ) this.parent.transfer = this.parent.disabled = false;
+  }
+
+  /* -------------------------------------------- */
   /*  Properties                                  */
   /* -------------------------------------------- */
 
@@ -46,11 +56,37 @@ export default class BaseEffectData extends ActiveEffectDataModel {
   /* -------------------------------------------- */
 
   /**
+   * Is this effect selected by any activity on its parent item?
+   * @type {boolean}
+   */
+  get isOnActivity() {
+    return this.item
+      && (this.item.system.activities?.contents ?? []).some(a => a.effects.some(e => e._id === this.parent._id));
+  }
+
+  /* -------------------------------------------- */
+
+  /**
    * Is this effect a rider for a non-applied enchantment?
    * @type {boolean}
    */
   get isRider() {
     return this.item?.getFlag("dnd5e", "riders.effect")?.includes?.(this.parent.id) ?? false;
+  }
+
+  /* -------------------------------------------- */
+  /*  Event Listeners & Handlers                  */
+  /* -------------------------------------------- */
+
+  /** @override */
+  onRenderActiveEffectConfig(app, html, context) {
+    if ( !this.isOnActivity ) return;
+    const checkboxes = html.querySelectorAll('dnd5e-checkbox[name="transfer"], dnd5e-checkbox[name="disabled"]');
+    for ( const checkbox of checkboxes ) {
+      if ( app.isEditable ) checkbox.dataset.tooltip = `DND5E.EFFECT.DisabledTooltip.${checkbox.name.capitalize()}`;
+      checkbox.disabled = true;
+      checkbox.checked = false;
+    }
   }
 
   /* -------------------------------------------- */
--- a/module/data/chat-message/usage-message-data.mjs
+++ b/module/data/chat-message/usage-message-data.mjs
@@ -114,8 +114,11 @@ export default class UsageMessageData extends ItemMessageData {
     }
 
     const item = this.parent.getAssociatedItem();
+    const activity = await fromUuid(this.activity.uuid);
+    const allowPlayerApplication = this.targets?.some(t => fromUuidSync(t.token).isOwner)
+      || ((this.parent.author?.id === game.user.id) && (activity.target.affects.type === "self"));
     context.effects = (await Promise.all(this.effects.map(uuid => fromUuid(uuid, { relative: item }))))
-      .filter(e => e && (game.user.isGM || (e.transfer & (this.parent.author?.id === game.user.id))));
+      .filter(e => e && ( game.user.isGM || dnd5e.settings.allowPlayerEffectsTray && allowPlayerApplication));
     return context;
   }
 
--- a/module/data/chat-message/usage-message-data.mjs
+++ b/module/data/chat-message/usage-message-data.mjs
@@ -117,9 +117,9 @@ export default class UsageMessageData extends ItemMessageData {
     const item = this.parent.getAssociatedItem();
     const activity = this.parent.getAssociatedActivity();
     const allowPlayerApplication = this.targets?.some(t => TargetsField.resolve(t).token?.isOwner)
-      || ((this.parent.author?.id === game.user.id) && (activity.target.affects.type === "self"));
+      || ((this.parent.author?.id === game.user.id) && (activity?.target.affects.type === "self"));
     context.effects = (await Promise.all(this.effects.map(uuid => fromUuid(uuid, { relative: item }))))
-      .filter(e => e && (game.user.isGM || dnd5e.settings.allowPlayerEffectsTray && allowPlayerApplication));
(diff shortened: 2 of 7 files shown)


roth-michael:
Rewritten to make use of #6312 
Effectively, the changes to current 5.2.x branch are:
- The above bullets for players seeing the tray
- The warning & prevention of concentration-dependent effect/enchant application moved from `EffectApplicationElement#_applyEffectToActor` and `EnchantActivity#applyEnchantment`, respectively, to `ActiveEffect5e#_preCreate`

The only loss of functionality, I think, is that players can no longer apply concentration-dependent enchantments & effects to themselves unless there's an active GM. Well, technically they could _never_ apply concentration-dependent _effects_ to themselves, since they wouldn't have had access to the application tray, so really it's just an inability to apply concentration-dependent enchantments without an active GM present. That seems acceptable to me, given that they are currently unable to break concentration on such effects without an active GM present.

Fyorl (commented):
I won't merge this now because I think there are two moving targets that we need to hit first:
1. Chat card redesign
2. Smoothing out the transferred-but-disabled toggle effects (i.e. we will remove support for this from the tray and prefer that effects are always applied fresh from the tray)

But please remind me so that we make sure this goes into 6.0.

roth-michael:
Okay, 3 commits.
1. Intended behavior of this PR. Nothing much has changed since my last comment here except that I no longer check the "transfer effect and self-authored" condition, because of commits 2 and 3.
2. Migration: Effects selected by one or more of their parent item's activities will get `disabled: false, transfer: false`.
3. UX improvements around the new intended logic: Local AEs added to an activity will get `disabled: false`, such effects will have their transfer checkbox disabled w/ an explanatory tooltip, such effects will have `transfer: true` updates amended to not include that.

Open questions/notes:
- If we set `transfer: false` during data prep for activity-borne effects, that would obviate the need for a migration, for use of `onUpdateActivities` for this, and for the `_preCreate` logic.
- This and #7319 have a good amount of similarity between them. Whichever goes in first, the other can probably be amended to simplify logic.

**EDIT:** Reduced to 2 commits. 2 and 3 merged into one which implements the first open question.

module/data/active-effect/base.mjs
```diff
@@ -45,6 +55,17 @@ export default class BaseEffectData extends ActiveEffectDataModel {
 
   /* -------------------------------------------- */
 
+  /**
+   * Is this effect selected by any activity on its parent item?
+   * @type {boolean}
+   */
+  get isOnActivity() {
+    return this.item
```
Fyorl, line 63: I went digging and turned up one exception we have here, which is the 'select form' mode of the Transform activity. It's basically the same thing as the `transfer: true`, toggle `disabled` style that we are trying to eliminate. I will note it here so that it's recorded as it must be addressed before release, but probably doesn't require action in this PR. I think the two options are:

1. Add something to activity `metadata` and check it here so we don't force `transfer: false, disabled: true` on those effects.
2. Use the same mechanism as every other effect: When applying a form we create a new copy of the effect with the `dependentOn` flag, and delete any sibling effect from a different form. Formless then just deletes.

module/data/chat-message/usage-message-data.mjs
```diff
@@ -114,8 +114,11 @@ export default class UsageMessageData extends ItemMessageData {
     }
 
     const item = this.parent.getAssociatedItem();
+    const activity = await fromUuid(this.activity.uuid);
+    const allowPlayerApplication = this.targets?.some(t => fromUuidSync(t.token).isOwner)
```
Fyorl: Use `TargetsField.resolve` for recorded targets, rather than `fromUuidSync`, and make sure to check for `null` returns.

module/data/chat-message/usage-message-data.mjs
```diff
@@ -114,8 +114,11 @@ export default class UsageMessageData extends ItemMessageData {
     }
 
     const item = this.parent.getAssociatedItem();
+    const activity = await fromUuid(this.activity.uuid);
```
Fyorl:
```suggestion
    const activity = this.parent,getAssociatedActivity();
```

module/documents/active-effect.mjs
```diff
@@ -766,6 +766,13 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     if ( await super._preCreate(data, options, user) === false ) return false;
     if ( options.keepOrigin === false ) this.updateSource({ origin: this.parent.uuid });
 
+    // Cannot apply a concentration-dependent effect without an active GM
+    const dependentUuid = this.getFlag("dnd5e", "dependentOn");
```
Fyorl: I guess actually we don't need this anymore since the `dependentOn` work. We blocked this because we would later write to the concentration effect in `_onCreate` and you can't write to an effect you don't own. But we don't do that anymore since we use the registry.

module/settings.mjs
```diff
@@ -251,6 +251,16 @@ export function registerSystemSettings() {
     })
   });
 
+  // Allow Player use of Effect Application Tray
+  game.settings.register("dnd5e", "allowPlayerEffectsTray", {
+    name: "SETTINGS.DND5E.PERMISSIONS.AllowEffects.Name",
+    hint: "SETTINGS.DND5E.PERMISSIONS.AllowEffects.Hint",
+    scope: "world",
+    config: true,
+    default: true,
```
Fyorl: I still maintain that it is better to have the GM apply effects to all targets at once rather than have players piecemeal apply effects to themselves. I know some disagree with that, but it's difficult to understand where the majority lies. I would prefer we maintain the status quo and have this default to `false`, and those who feel strongly can turn this back on.

Fyorl (changes requested):

module/documents/active-effect.mjs
roth-michael: Trying to remember why I kept this even after the change to `dependentOn`. Failing, though, so will remove this following some testing to make sure I didn't miss anything.

Fyorl (approved):

PR #6001, opened 2025-08-14
Add compendium browser link to encounter actor sheet

arbron:
Replaces the existing drop area on Encounter sheets with a button to open the compendium browser filtered to NPCs and makes it visible when the encounter is empty or in edit mode.

Encounter Browser Button

### DIFF
--- a/module/applications/actor/encounter-sheet.mjs
+++ b/module/applications/actor/encounter-sheet.mjs
@@ -261,7 +261,7 @@ export default class EncounterActorSheet extends MultiActorSheet {
       }
     });
     if ( results ) {
-      const actors = await Promise.all(results.map(r => fromUuid(r)));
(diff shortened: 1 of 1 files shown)


module/applications/actor/encounter-sheet.mjs
```diff
@@ -232,6 +241,33 @@ export default class EncounterActorSheet extends MultiActorSheet {
...
+          documentClass: "Actor",
+          types: new Set(["npc"])
+        }
+      },
+      selection: {
+        min: 1
+      }
+    });
+    if ( results ) {
+      const actors = await Promise.all(results.map(r => fromUuid(r)));
```
Fyorl:
```suggestion
      const actors = await Promise.all(results.map(fromUuid));
```

Fyorl (approved):
Nice polish

PR #6017, opened 2025-08-16
[#6016] Encounter and Vehicle actors don't have `traits`

### LINKED ISSUE #6016: Encounter and Vehicle typed actors do not have `traits`
```js
tokens.mjs:12 
 Uncaught TypeError: Cannot read properties of undefined (reading 'size')
[Detected 1 package: system:dnd5e(5.1.1)]
    at TokenLayer5e.isOccupiedGridSpaceBlocking (tokens.mjs:12:73)
```

thatlonelybugbear:
Closes #6016 

Not sure if it's *better* to return early and not care about those types, instead.

### DIFF
--- a/module/canvas/token.mjs
+++ b/module/canvas/token.mjs
@@ -22,10 +22,9 @@ export default class Token5e extends foundry.canvas.placeables.Token {
   findMovementPath(waypoints, options) {
 
     // Normal behavior if movement automation is disabled or this.actor is not a creature
-    if (
-        game.settings.get("dnd5e", "disableMovementAutomation")
-        || !this.document.actor?.system.isCreature
-    ) return super.findMovementPath(waypoints, options);
+    if ( game.settings.get("dnd5e", "disableMovementAutomation") || !this.document.actor?.system.isCreature ) {
+      return super.findMovementPath(waypoints, options);
+    }
 
     // Get all grid spaces as waypoints so that running into a blocking token stops us immediately before it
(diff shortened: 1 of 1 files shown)


roth-michael:
Elsewhere this has been solved by checking `Actor5e#isCreature` prior to calling these functions - probably the same could be done here in `Token5e#findMovementPath`

thatlonelybugbear:
Opted for the least invasive solution.

I guess 
```js
 /** @inheritDoc */
  findMovementPath(waypoints, options) {

    // Normal behavior if movement automation is disabled or !Actor5e#isCreature
    if (
        game.settings.get("dnd5e", "disableMovementAutomation")
        || !this.document.actor?.system.isCreature
    ) return super.findMovementPath(waypoints, options);

    // Get all grid spaces as waypoints so that running into a blocking token stops us immediately before it
    waypoints = this.document.getCompleteMovementPath(waypoints);

    // Drop all intermediate waypoints except those immediately before a blocking token
    const grid = this.document.parent.grid;
    waypoints = waypoints.filter((waypoint, i) => {
      return !waypoint.intermediate || this.layer.isOccupiedGridSpaceBlocking(grid.getOffset(waypoints[i + 1]), this);
    });
    return super.findMovementPath(waypoints, options);
  }
   ```
   would do it too if preferable.

roth-michael:
Maybe both, honestly. It shouldn't check for non-creatures, but in theory it also shouldn't throw errors if somebody manually calls it with a non-creature token.

PR #6030, opened 2025-08-16
[#5874] Add antimagic region behavior, tie into `magicAvailable`

### LINKED ISSUE #5874: Add antimagic region behavior
Add a new antimagic region behavior that would prevent magic from working within its bounds. This would apply a special status effect (similar to bloodied or encumbered) that would indicate the character is under this effect and allow the system to make adjustments:
- Spells cannot be cast
- `ItemDataModel#magicAvailable` would return `false`, disabling magic bonuses to weapons, ammunition, and armor
- Active effects marked as "magical" would become unavailable (requires additional work)
- Activities marked as "Requires Magic" wouldn't be available (requires additional work, see https://github.com/foundryvtt/dnd5e/issues/4963)

arbron:
Creates a new `antimagic` region behavior. When a token enters this region it gains the new `antimagic` status effect. The effect is then checked for in the `magicAvailable` method, disabling anything that requires magic. It also hides any spells in the spells tab.

The `magicAvailable` method is moved from `EquippableItemTemplate` to `ItemDataModel` so it is available on all items and the logic has been updated to ensure it respects the new status effect and works regardless of whether the item has properties or attunement.

Closes #5874


Zanderaf:
Any plans to support something like spells and magical effects ignoring this region effect? The Antimagic Field spell does say that there are some instances where magic can still occur inside it.

'Spells and other magical effects, except those created by an artifact or a deity, are suppressed in the sphere and can't protrude into it."

Issue #6029, opened 2025-08-16
Tokens block each other when dragged together


Saibot393:
With the new enemy token block movement feature dragging two tokens of opposite disposition at the same time can cause one of them to block the other.
E.g. dragging the two tokens in the screenshot (the left one being "friendly", the right one "hostile") to the right at the same time causes the left one to stay at its current position as it collides with the right tokens pre-drag position

<img width="784" height="525" alt="Image" src="https://github.com/user-attachments/assets/74cbf121-dec1-4414-b8ff-863aff42ea8c" />

thatlonelybugbear:
In a similar vein, if the two token started their movement some grid squares apart, the movement of the following token would be stopped at the point of the initial collision.

Those cases can be fixed by #6031 but it then introduces some wall collision logic issues, shown in the video uploaded in the PR.

PR #6031, opened 2025-08-16
Ignore tokens moving together

thatlonelybugbear:
Should be closing #6029 

Will be testing some more for edge cases, but seems to deal with the issue at hand.


thatlonelybugbear:
Edge cases with the suggested change.
- Acolyte set as Hostile, rest as Friendly 

[edgeCases.webm](https://github.com/user-attachments/assets/b37de49a-602f-4a05-aa07-2d52eb78102c)

PR #6043, opened 2025-08-18
[#6015, #6042] Always show movement ruler highlight, change movement automation setting from a boolean to a dropdown

### LINKED ISSUE #6015: Add option to allow moving through enemies
On version 5.1 and 5.1.1, characters are unable to move on to enemies, or move through them, if they are within 1 size of each other. This behavior can be disabled, but disabling it also disables all other movement automation.

Ideally there would be fine grained controls for each type of movement automation, but at least separate out automatically blocking movement from displaying movement costs.

### LINKED ISSUE #6042: Split ruler measurement and enforced movement automation into separate settings
(or just 1 setting that enforces only the higher automation)

There is currently no way to get the colored highlighting when measuring with the drag-ruler without also getting the other features that are the 'controversial' features that necessitated the setting in the first place.

Tables who do not want to have the system govern their token movement should in my opinion be allowed to use this simpler quality-of-life feature on its own.

These are what I would consider quality-of-life features since they do not prevent or inhibit users but only provide more information:
- Colored drag ruler (green, yellow, red depending on distance). This is simply information added onto the default core drag ruler measurement.
- The `Difficult Terrain` region behavior's modifiers. The core `Modify Movement Cost` region behavior has no setting, so it makes less sense that this behavior is locked behind one.
- Movement cost modifiers due to other tokens in the path.

This is what I would consider higher automation that far from all tables would want to play with:
- Preventing movement of tokens due to other tokens blocking them. This one also has more edge cases that need to be covered compared to the above three.
- The slowed movement animation due to a lack of a Movement Speed on the actor (eg when under the effect of the `grappled` status).

roth-michael:
Closes #6015 
Closes #6042 
Existing "Disable Movement Automation" setting covers only difficult terrain & token-caused difficult terrain. Replacement  "Movement Automation" setting has, in addition to a `none` (equivalent of current `true`) and a `full` (equivalent of current `false`), a `noBlocking`, which is the full automation _except for_ enforced blocking of movement due to enemy tokens. The old setting will be migrated to the new one according to the above-listed equivalents. Movement distance highlighting now always shown (but can be disabled by setting `CONFIG.DND5E.tokenRulerColors`'s values to `null`.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -5513,7 +5513,11 @@
   "AUTOMATION": {
     "Movement": {
       "Name": "Disable Movement Automation",
-      "Hint": "Do not apply difficult terrain movement penalties, or prevent creatures from moving through each other according to the rules version."
+      "Hint": "Do not apply difficult terrain movement penalties, including tokens counting as difficult terrain according to the rules version."
+    },
+    "TokenBlocking": {
+      "Name": "Disable Token Blocking",
+      "Hint": "Do not prevent creatures from moving through each other according to the rules version."
     }
   },
   "BLOODIED": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -5513,10 +5513,10 @@
   "AUTOMATION": {
     "Movement": {
       "Name": "Movement Automation",
-      "Hint": "Configure the amount of movement automation used by the system.",
-      "Full": "Full: Tokens may block or hinder movement of other tokens",
-      "NoBlocking": "Partial: Tokens may act as difficult terrain but cannot block each other",
-      "None": "None: Tokens can move freely through each other, difficult terrain does nothing"
+      "Hint": "Configure the amount of movement automation used by the system: Full automation including token blocking, partial automation of difficult terrain only, or off entirely.",
+      "Full": "Full",
+      "NoBlocking": "Partial",
+      "None": "None"
     }
   },
   "BLOODIED": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -5512,12 +5512,11 @@
 "SETTINGS.DND5E": {
   "AUTOMATION": {
     "Movement": {
-      "Name": "Disable Movement Automation",
-      "Hint": "Do not apply difficult terrain movement penalties, including tokens counting as difficult terrain according to the rules version."
-    },
-    "TokenBlocking": {
-      "Name": "Disable Token Blocking",
-      "Hint": "Do not prevent creatures from moving through each other according to the rules version."
+      "Name": "Movement Automation",
+      "Hint": "Configure the amount of movement automation used by the system.",
+      "Full": "Full: Tokens may block or hinder movement of other tokens",
+      "NoBlocking": "Partial: Tokens may act as difficult terrain but cannot block each other",
+      "None": "None: Tokens can move freely through each other, difficult terrain does nothing"
     }
   },
   "BLOODIED": {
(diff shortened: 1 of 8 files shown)


arbron:
@roth-michael Could you comment on https://github.com/foundryvtt/dnd5e/issues/6015 :)

robertaird:
could this also add an additional setting for enabling token blocking only during combat, or is that better as a separate feature request? in modules that encourage players controlling own their tokens during exploration it can get in the way a bit (especially in scenes with lots of NPCs), but i appreciate having it for combat

Fyorl (commented):
I preferred the proposal that had the single drop-down of different levels. Can we go with that and revisit if we add more granular options (such as the one requested in this PR)? The problem with adding extra settings is that the current list of system settings is already too long and overwhelming for new users. Moving settings into their own dialogs solves this somewhat, but then individual settings become unsearchable, which is also a problem for new users.

Until we have a better solution for managing settings, we need to be highly selective about which new settings we add.

module/canvas/ruler.mjs
```diff
@@ -85,10 +85,9 @@ export default class TokenRuler5e extends foundry.canvas.placeables.tokens.Token
    * @returns {object} The adjusted style, or existing style if no adjustment is necessary
    */
   #getSpeedBasedStyle(waypoint, style) {
-    // If movement automation disabled, or if showing a different client's measurement, use default style
-    const noAutomation = game.settings.get("dnd5e", "disableMovementAutomation");
+    // If showing a different client's measurement, or if teleporting, use default style
     const isSameClient = game.user.id in this.token._plannedMovement;
-    if ( noAutomation || !isSameClient || CONFIG.Token.movement.actions[waypoint.action]?.teleport ) return style;
+    if ( !isSameClient || CONFIG.Token.movement.actions[waypoint.action]?.teleport ) return style;
```
Fyorl: Does this need a check for `none`? Or is this movement colouring always-on now?

lang/en.json
```diff
@@ -5458,8 +5211,11 @@
 "SETTINGS.DND5E": {
   "AUTOMATION": {
     "Movement": {
-      "Name": "Disable Movement Automation",
-      "Hint": "Do not apply difficult terrain movement penalties, or prevent creatures from moving through each other according to the rules version."
+      "Name": "Movement Automation",
+      "Hint": "Configure the amount of movement automation used by the system.",
+      "Full": "Full: Tokens may block or hinder movement of other tokens",
+      "NoBlocking": "Partial: Tokens may act as difficult terrain but cannot block each other",
+      "None": "None: Tokens can move freely through each other, difficult terrain does nothing"
```
Fyorl: Not enough space for these

```suggestion
      "Name": "Movement Automation",
      "Hint": "Configure the amount of movement automation used by the system: Full automation including token blocking, partial automation of difficult terrain only, or off entirely.",
      "Full": "Full",
      "NoBlocking": "Partial",
      "None": "None"
```

Fyorl (changes requested):

module/canvas/ruler.mjs
roth-michael: Always on. Figure since it's entirely non-prohibitive, no reason not to keep it (and can be disabled by nulling everything under `CONFIG.DND5E.tokenRulerColors`. But will defer to you on this; gating it behind the setting _would_ match the current "off" behavior.

Fyorl: I'm not fussed

Fyorl (approved):

PR #6053, opened 2025-08-19
[#4708] Rebuild `AdvancementFlow`, `SubclassFlow`, & `TraitFlow` for V2

arbron:
Adds a new version of `AdvancementFlow` based on `ApplicationV2` and implements `SubclassFlow` & `TraitFlow` using it. Adds some modified rendering for `AdvancementManager` to support either application version.

The design of `SubclassFlow` remains the same, but `TraitFlow` has been reworked to use a separated list and display icons for each trait.

Trait Flow V2 Saves
Trait Flow V2 Skills

The V2 version of advancement flows are designed to submit on change, rather than only submitting during the forward process. This will generally require changes to the coresponding advancement's application method to handle piecemeal application. In order to keep everything in the correct state, this also means that `apply` is called once before the initial rendering of the flow to give the advancement a chance to populate initial data.

This change involves some modification to the `Advancement` API for the `apply`, `restore`, `reverse`, and `automaticApplicationValue` methods:

- `apply`, `restore`, and `reverse` get a new `options` object
- `automatic` option is passed when the automatic application process is being applied
- `initial` option is passed when the `apply` method is called initially before the flow is first rendered
- `automaticApplicationValue` has been made `async` (this change applies when using both V1 or V2 flows)

All of these changes have been made in a way to allow V1 and V2 advancement flows to coexist for a time, but should support future changes to the advancement manager allowing for grouped flows.

### DIFF
--- a/module/documents/advancement/advancement.mjs
+++ b/module/documents/advancement/advancement.mjs
@@ -247,15 +247,6 @@ export default class Advancement extends PseudoDocumentMixin(BaseAdvancementData
   }
 
   /* -------------------------------------------- */
-
-  /** @inheritDoc */
-  updateSource(updates) {
-    super.updateSource(updates);
-    this.prepareData();
-    return this;
-  }
-
-  /* -------------------------------------------- */
   /*  Application Methods                         */
   /* -------------------------------------------- */
 
--- a/module/documents/mixins/pseudo-document.mjs
+++ b/module/documents/mixins/pseudo-document.mjs
@@ -204,6 +204,7 @@ export default function PseudoDocumentMixin(Base) {
      */
     updateSource(updates) {
       super.updateSource(updates);
+      this.prepareData?.();
       return this;
     }
 


arbron:
Adjusted the design to be more compact like the proposed new item choice flow design:

Trait V2 Tweaks 2

module/applications/advancement/advancement-flow-v2.mjs
```diff
@@ -0,0 +1,261 @@
...
+    return context;
+  }
+
+  /* -------------------------------------------- */
+  /*  Life-Cycle Handlers                         */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _onFirstRender(context, options) {
+    super._onFirstRender(context, options);
```
Fyorl:
```suggestion
  async _onFirstRender(context, options) {
    await super._onFirstRender(context, options);
```

module/applications/advancement/advancement-flow-v2.mjs
```diff
@@ -0,0 +1,261 @@
...
+      level: this.level,
+      type: this.advancement.constructor.typeName
+    });
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _onRender(context, options) {
+    super._onRender(context, options);
```
Fyorl:
```suggestion
  async _onRender(context, options) {
    await super._onRender(context, options);
```

module/applications/advancement/advancement-flow-v2.mjs
```diff
@@ -0,0 +1,261 @@
...
+  /**
+   * Handle clicking on an item to open its sheet.
+   * @this {SubclassFlow}
+   * @param {Event} event         Triggering click event.
+   * @param {HTMLElement} target  Button that was clicked.
+   */
+  static async #viewItem(event, target) {
+    const uuid = target.closest("[data-uuid]")?.dataset.uuid;
+    const item = await fromUuid(uuid);
+    item?.sheet.render(true);
```
Fyorl:
```suggestion
    item?.sheet.render({ force: true });
```

module/applications/advancement/advancement-manager.mjs
```diff
@@ -552,9 +570,15 @@ export default class AdvancementManager extends Application5e {
...
-    this.step.flow.options.manager ??= this;
-    await this.step.flow._render(true, options);
+    if ( this.step.flow instanceof Application ) {
+      this.step.flow._element = null;
+      this.step.flow.options.manager ??= this;
+      await this.step.flow._render(true, options);
+    } else {
+      this.step.flow.manager ??= this;
+      await this.step.flow.render({ force: true });
+      this.step.flow._insertElement(this.step.flow.element);
```
Fyorl, line 580: I wonder why this is needed

module/applications/advancement/subclass-flow.mjs
```diff
@@ -1,116 +1,128 @@
...
 
   /** @inheritDoc */
-  async retainData(data) {
-    await super.retainData(data);
-    const uuid = foundry.utils.getProperty(data, "flags.dnd5e.sourceId");
-    if ( uuid ) this.subclass = await fromUuid(uuid);
+  async _prepareContext(options) {
+    const uuid = foundry.utils.getProperty(this.retainedData ?? {}, "flags.dnd5e.sourceId");
+    if ( uuid ) await this.advancement.apply(this.level, { retainedData: this.retainedData, uuid });
+    return await super._prepareContext(options);
```
Fyorl:
```suggestion
    return super._prepareContext(options);
```

module/data/advancement/item-grant.mjs
```diff
@@ -40,7 +40,7 @@ export default class ItemGrantConfigurationData extends foundry.abstract.DataMod
 
   /** @inheritDoc */
   static migrateData(source) {
-    if ( "items" in source ) {
+    if ( foundry.utils.getType(source.items) === "Array" ) {
```
Fyorl:
```suggestion
    if ( Array.isArray(source.items) ) {
```

module/documents/advancement/subclass.mjs
```diff
@@ -56,26 +56,34 @@ export default class SubclassAdvancement extends Advancement {
...
+  /**
+   * @typedef SubclassAdvancementApplicationData
+   * @property {object} [retainedData]  Retained data object for a previous subclass.
+   * @property {string} [uuid]          UUID of subclass to add.
+   */
+
   /** @inheritDoc */
-  async apply(level, data, retainedData) {
-    const useRetained = data.uuid === foundry.utils.getProperty(retainedData, "flags.dnd5e.sourceId");
+  async apply(level, { retainedData={}, uuid }={}, options={}) {
```
Fyorl, line 64: Doesn't seem to match the signature of `super.apply`. Maybe `uuid` should move to `options`?

module/documents/item.mjs
```diff
@@ -1000,7 +1000,7 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     const advancement = this.system.advancement.get(id);
     const update = { [`system.advancement.${id}`]: updates };
     if ( source ) {
-      this.updateSource(update);
+      advancement.updateSource(update);
```
Fyorl:
```suggestion
      advancement.updateSource(updates);
```

module/documents/advancement/subclass.mjs
```diff
@@ -56,26 +56,34 @@ export default class SubclassAdvancement extends Advancement {
...
     let itemData = useRetained ? retainedData : null;
     if ( !itemData ) {
-      itemData = await this.createItemData(data.uuid);
+      itemData = await this.createItemData(uuid);
       delete itemData.flags?.dnd5e?.advancementOrigin;
       delete itemData.flags?.dnd5e?.advancementRoot;
       foundry.utils.setProperty(itemData, "system.classIdentifier", this.item.identifier);
     }
     if ( itemData ) {
       this.actor.updateSource({ items: [itemData] });
```
Fyorl, line 76: So, while updating the advancement's source will not cause any references to break, updating the Actor or Item's source will. You can see this in the subclass flow where adding or removing a subclass will cause the title to disappear because only the new subclass advancement instance has been re-prepared.

If there's no way around this (I assume any advancement that grants an item needs to grant it to the actor clone, so probably not), then we'll need to look into that patch I made for #6236, or if there are still problems with that, then your full implementation.

module/documents/advancement/subclass.mjs
arbron: Hmm, the signature is number, object, object, so it seems like it should match. Is it because I am destructuring it here rather than accessing the properties later?

Plus UUID is part of the main application data for the advancement so I would expect it to be in `data` rather than `options`.

module/applications/advancement/advancement-manager.mjs
arbron: Not sure, testing it without seems to render correctly, so maybe there was something weird going on 6 months ago when I originally wrote this.

module/documents/advancement/subclass.mjs
arbron: Looks like it is an issue with `Advancement.prepareData` not being called after `updateSource` is called.

module/documents/advancement/subclass.mjs
Fyorl: Right, I see that you're actually intending to change the signature so that all advancements accept `retainedData` in their 2nd parameter. I got it the other way around and thought `SubclassAdvancement` was the odd-one-out.

module/applications/advancement/advancement-manager.mjs
Fyorl: I discovered why: When going backwards, the flow is removed from the DOM but internally still considers itself rendered, so `_insertElement` isn't called again, resulting in the user seeing a blank step. Maybe calling `close` on the flows is the technically correct solution but I am ok with adding this line back in for now.

module/documents/advancement/advancement.mjs
```diff
@@ -244,27 +246,36 @@ export default class Advancement extends PseudoDocumentMixin(BaseAdvancementData
...
   }
 
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  updateSource(updates) {
+    super.updateSource(updates);
+    this.prepareData();
+    return this;
+  }
```
Fyorl: I assume activities don't need the same treatment? If so, we can move this into the `PseudoDocument` override.

Fyorl (approved):

PR #6054, opened 2025-08-20
Fix crawling cost in difficult terrain

dev7355608:


module/documents/token.mjs
```diff
@@ -127,6 +127,7 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
       };
       actionConfig.getCostFunction = (...args) => this.getMovementActionCostFunction(type, ...args);
     }
+    CONFIG.Token.movement.actions.crawl.getCostFunction = () => (cost, from, to, distance) => cost + distance;
```
dev7355608: Core uses `cost * 2`, which makes the cost of crawling 4 feet instead of 3 feet per foot.

Fyorl: Do we need to check the `disableMovementAutomation` setting here? If that's set, then all other cost functions are `cost => cost`.

PR #6055, opened 2025-08-20
[#4708] Rebuild `ScaleValueFlow` & `SizeFlow` for V2

arbron:
Reworked `ScaleValueFlow` & `SizeFlow` for application V2. Scale value flow just has some minor tweaks to move the scale value changes to the center. Size flow gets the select dropdown moved into a fieldset to match other advancement types with a single choice.

Scale Value V2 Tweaks

Size V2 Tweaks

### DIFF
--- a/module/documents/advancement/size.mjs
+++ b/module/documents/advancement/size.mjs
@@ -102,7 +102,9 @@ export default class SizeAdvancement extends Advancement {
 
   /** @inheritDoc */
   async reverse(level, options={}) {
+    const data = { ...this.value };
     this.actor.updateSource({ "system.traits.size": "med" });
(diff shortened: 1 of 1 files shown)


Zanderaf:
Is the text on the Size advancement preset from the system? If so, think there might be a spelling mistake 😭 
image

krbz999:
> Is the text on the Size advancement preset from the system? If so, think there might be a spelling mistake 😭

No, that's a custom hint in an advancement.

https://github.com/foundryvtt/dnd5e/pull/6056

Fyorl (approved):
Found one issue but good to go otherwise. If you advance past a size advancement and then return back to it, the form does restore the size that was picked.

PR #6067, opened 2025-08-23
[#6063] Check for `neverBlockMovement` when constraining movement & finding movement path

### LINKED ISSUE #6063: Ethereal does not allow movement through other tokens
A token marked as ethereal allows other tokens to move through it, but it does not allow the token itself to move through other tokens.

Image

Image

roth-michael:
Closes #6063 
Makes me think we could consider a `Actor5e#canBlockMovement` getter which would be something like
```js
get canBlockMovement() {
  const neverBlockStatuses = CONFIG.statusEffects.filter(s => s.neverBlockMovement).map(s => s.id);
  return this.isCreature && !neverBlockStatuses.some(status => this.statuses.has(status));
}
```

### DIFF
--- a/module/canvas/token.mjs
+++ b/module/canvas/token.mjs
@@ -20,9 +20,11 @@ export default class Token5e extends foundry.canvas.placeables.Token {
 
   /** @inheritDoc */
   findMovementPath(waypoints, options) {
+    const neverBlockStatuses = CONFIG.statusEffects.filter(s => s.neverBlockMovement).map(s => s.id);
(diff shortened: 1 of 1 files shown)


arbron:
Whoops, somehow I managed to miss this PR and wrote my own: https://github.com/foundryvtt/dnd5e/pull/6079

It's basically the same but I added caching to `neverBlockStatuses` so we aren't constantly having to recalculate it and used a set intersection check.

Issue #6135, opened 2025-09-03
Add opportunity to gain proficiency in "all" skills


Dewi-Jones:
Right now, the Boon of Skill (PHB) upgrades every single skill proficiency to `1`. This does not behave well with any modules that add additional skills.
For these cases, a possibility to add proficiency in "all" skills would be necessary.
Related: https://github.com/foundryvtt/dnd5e/issues/4644

<img width="532" height="693" alt="Image" src="https://github.com/user-attachments/assets/e7d5d50f-903d-4a4e-a7ba-f8b836427e46" />

krbz999:
Would pair well with the new Tool Expertise flag.

In *Aetherial Expanse*, we add a new Navigation skill that is currently not benefiting from this boon.

arbron:
Feels like we could add a bit to the data model, rather than using flags here:

```javascript
roll: {
  proficiency: {
    checks: new NumberField({ choices: [0, 0.5, 1, 2] }),
    saves: new NumberField({ choices: [0, 1] }),
    skills: new NumberField({ choices: [0, 0.5, 1, 2] }),
    tools: new NumberField({ choices: [0, 0.5, 1, 2] })
  }
}
```

These would basically set the "proficiency floor" for different roll types. Then the AE would be be: `system.roll.proficiency.skills | UPGRADE | 1`.

krbz999:
> These would basically set the "proficiency floor" for different roll types. Then the AE would be be: `system.roll.proficiency.skills | UPGRADE | 1`.

I also wonder if there would be value in a field that handles `CONST.ACTIVE_EFFECT_MODES.ADD` in a special way, with -1 or 1 moving the proficiency up or down a "stage". E.g. if the current value is 0, `ADD 1` becomes 0.5; 0.5 becomes 1; and 1 becomes 2.

UPGRADE, DOWNGRADE, and OVERRIDE would work the same, but this would allow for simply upping the proficiency level without having to consider what the current value might be.

PR #6136, opened 2025-09-03
Flag: Proficiency in all Skills

Dewi-Jones:
Adds a new Flag `allAroundAdept` and extends `calculateAbilityCheckProficiency` to grant a multiplier of `1` to all skills if this flag is set.
Closes https://github.com/foundryvtt/dnd5e/issues/6135

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -4713,6 +4713,12 @@ DND5E.characterFlags = {
     section: "DND5E.Feats",
     type: Boolean
   },
+  allAroundAdept: {
+    name: "DND5E.FlagsAllAroundAdept",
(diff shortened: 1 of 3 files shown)


module/config.mjs
```diff
@@ -4713,6 +4713,12 @@ DND5E.characterFlags = {
     section: "DND5E.Feats",
     type: Boolean
   },
+  allAroundAdept: {
```
krbz999, line 4716: Maybe it's just me but I'd go with just a simple name like "All Skills".

Dewi-Jones: I chose this because the Boon of Skill names this explicitely "All-Around Adept".

PR #6150, opened 2025-09-07
[#6126] Display icon for rider activities and effects on item

### LINKED ISSUE #6126: Visually indicate on item sheets what activities & effects are riders
For people not familiar with the enchantment system, it can be confusing why some effects & activities are listed on the item but not actually usable when the item is used. Adding an icon next to activities and effects that are riders would help indicate what is occurring in those situations.

arbron:
Adds a small paperclip icon next to rider activities and effects to make it clear that they are riders.

Rider Activities Icon
Rider Effects Icon

Closes #6126

### DIFF
--- a/module/data/activity/base-activity.mjs
+++ b/module/data/activity/base-activity.mjs
@@ -179,6 +179,16 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
   /* -------------------------------------------- */
 
   /**
+   * Is this activity a rider for a non-applied enchantment?
+   * @type {boolean}
+   */
+  get isRider() {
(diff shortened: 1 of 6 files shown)


module/data/activity/base-activity.mjs
```diff
@@ -178,6 +178,16 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
 
   /* -------------------------------------------- */
 
+  /**
+   * Is this activity a rider for a non-applied enchantment?
+   * @type {boolean}
```
krbz999, line 183: This can be `undefined`.

PR #6153, opened 2025-09-07
[#3514] Allow AC formulae to reference spell mod

### LINKED ISSUE #3514: Can't use spellmod in AC formula?
I cannot use `12 + @attributes.spellmod` in the ac formula in an active effect. As you can see in the image below, the formula completely fails to modify the AC (spellmod is 3, btw)
image

If I manually set it to `12 + @abilities.cha.mod` it works exactly as expected.
image

This makes it pretty likely that the issue is in the effect calculation with `@attributes.spellmod` Unless I'm doing something wrong here?

krbz999:
Rather than simplifying `ac.min`, `ac.bonus`, and then deriving `ac.value` immediately during data prep, the first two are never simplified (don't think there's much reason to), and `ac.value` is moved to a getter that is derived only once when needed.

This allows for using any roll data properties as an AC bonus, despite data prep order.

Tested with an effect:
```
system.attributes.ac.bonus | ADD      | @attributes.spell.mod
system.attributes.ac.min   | OVERRIDE | 5 * @attributes.spell.mod
```

Closes #3514.

### DIFF
--- a/module/data/actor/character.mjs
+++ b/module/data/actor/character.mjs
@@ -235,12 +235,12 @@ export default class CharacterData extends CreatureTemplate {
     this.prepareCurrency();
     this.prepareSkills({ rollData, originalSkills });
     this.prepareTools({ rollData });
+    AttributesFields.prepareSpellcastingAbility.call(this);
     AttributesFields.prepareArmorClass.call(this, rollData);
     AttributesFields.prepareConcentration.call(this, rollData);
     AttributesFields.prepareEncumbrance.call(this, rollData);
     AttributesFields.prepareInitiative.call(this, rollData);
     AttributesFields.prepareMovement.call(this, rollData);
-    AttributesFields.prepareSpellcastingAbility.call(this);
     TraitsFields.prepareLanguages.call(this);
     TraitsFields.prepareResistImmune.call(this);
 
--- a/module/data/actor/npc.mjs
+++ b/module/data/actor/npc.mjs
@@ -421,13 +421,13 @@ export default class NPCData extends CreatureTemplate {
     this.prepareCurrency();
     this.prepareSkills({ rollData, originalSkills });
     this.prepareTools({ rollData });
+    AttributesFields.prepareSpellcastingAbility.call(this);
     AttributesFields.prepareArmorClass.call(this, rollData);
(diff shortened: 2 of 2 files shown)


Fyorl (changes requested):
Converting every call to `ac.value` in the codebase into a closure invocation defeats the purpose of data preparation. If there is a data dependency that `ac.value` requires, then `prepareArmorClass` should be moved later in the preparation order to a point where that dependency is satisfied.

krbz999:
> Converting every call to `ac.value` in the codebase into a closure invocation defeats the purpose of data preparation. If there is a data dependency that `ac.value` requires, then `prepareArmorClass` should be moved later in the preparation order to a point where that dependency is satisfied.

It would be at most once (per data prep cycle), not every call. I wanted to avoid rearranging the data prep order as that could open up many more similar issues.

Fyorl:
> It would be at most once (per data prep cycle), not every call.

It looks like every access of `ac.value` would invoke the closure. What am I missing?

krbz999:
> > It would be at most once (per data prep cycle), not every call.
> 
> It looks like every access of `ac.value` would invoke the closure. What am I missing?

Perhaps I don't understand what you mean, but `ac._value` is assigned a non-nullish value only once.

(Though `getRollData` is retrieved each time; I've changed this.)

Fyorl:
> Perhaps I don't understand what you mean, but `ac._value` is assigned a non-nullish value only once.

Yes, so `simplifyBonus` is only called once, but every time `ac.value` is accessed the `get: () => {}` function is invoked, no?

I see you've moved the `getRollData` call, which is better, but what's the benefit of incurring the overhead of a function call on every access vs. simply calculating this value later? Have we verified that it actually does cause issues?

krbz999:
> Have we verified that it actually does cause issues?

I want to say that it is unlikely, but that'd require more testing than I have the time to dedicate.

Spellcasting itself relies only on the Abilities having been prepared (specifically only `abilities.*.mod` seems relevant).

For PCs (and NPCs), it *looks* like it can be moved to before AC, conc, encumbrance, (and exhaustion), init, and movement with no issues, which it was previously immediately after.

Fyorl (approved):

PR #6169, opened 2025-09-10
[#5237, #5293] Improve currency in starting equipment labels

### LINKED ISSUE #5237: The Class Summary page does not currently show the optional starting wealth option from the class item
For example, the Barbarian's Starting Equipment in the Class Summary page:

Image

The Class item does contain the optional gold value though so it would be great if we could display that as an "Or":

Image

### LINKED ISSUE #5293: The Class item sheet details tab starting equipment doesn't cater for the small purse of gold included with the kit
It doesn't cater for the small purse included with the kit, only the large purse if you eschew the kit. 

Image

I.e. The `and 5gp` bit – there's no way to add that in.

Image

My suggestion would be to add an "Choose Award" type to the select

arbron:
Add a new entry type to starting equipment that allows entering currency directly. SRD 5.2 classes have been updated to include the missing gold from their starting equipment.

Starting Equipment Gold Config

Make use of this new currency entry to display the starting equipment list with the starting wealth included as an option when using the modern rules formatting.

Starting Equipment Gold Modern

Also adds letter prefixes to OR-lists to more accurately reflect how classes are displayed in both the mordern and legacy styles.

Starting Equipment Gold Legacy

Closes #5237
Closes #5293

### DIFF
--- a/module/data/item/templates/starting-equipment.mjs
+++ b/module/data/item/templates/starting-equipment.mjs
@@ -30,14 +30,48 @@ export default class StartingEquipmentTemplate extends SystemDataModel {
    * @type {string}
    */
   get startingEquipmentDescription() {
+    return this.getStartingEquipmentDescription();
+  }
+
+  /* -------------------------------------------- */
+  /*  Helpers                                     */
+  /* -------------------------------------------- */
+
+  /**
+   * Create a HTML formatted description of the starting equipment on this item.
+   * @param {object} [options={}]
+   * @param {boolean} [options.modernStyle]       Should this be formatted according to modern rules or legacy.
+   * @returns {string}
+   */
+  getStartingEquipmentDescription({ modernStyle }={}) {
     const topLevel = this.startingEquipment.filter(e => !e.group);
     if ( !topLevel.length ) return "";
 
-    // If more than one entry, display as an unordered list (like for classes)
-    if ( topLevel.length > 1 ) return `<ul>${topLevel.map(e => `<li>${e.label}</li>`).join("")}</ul>`;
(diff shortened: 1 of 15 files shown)


module/data/item/templates/starting-equipment.mjs
```diff
@@ -216,4 +230,78 @@ export class EquipmentEntryData extends foundry.abstract.DataModel {
...
+   * @param {string[]} entries                    Entries to prefix.
+   * @param {object} [options={}]
+   * @param {number} [options.depth=1]            Current depth of the OR entry (1 or 2).
+   * @param {boolean} [options.modernStyle=true]  Capitalized first level markers rather than lowercased.
+   * @param {string[]} [options.usedPrefixes]     Prefixes that were used.
+   * @returns {string[]}
+   */
+  static prefixOrEntries(entries, { depth=1, modernStyle=true, usedPrefixes }={}) {
+    let letters = game.i18n.localize("DND5E.StartingEquipment.Prefixes");
+    if (!letters) return entries;
```
Fyorl:
```suggestion
    if ( !letters ) return entries;
```

module/data/item/templates/starting-equipment.mjs
```diff
@@ -216,4 +230,78 @@ export class EquipmentEntryData extends foundry.abstract.DataModel {
...
+    if ( (modernStyle && (depth == 1)) || (!modernStyle && (depth === 2)) ) letters = letters.toUpperCase();
```
Fyorl:
```suggestion
    if ( (modernStyle && (depth === 1)) || (!modernStyle && (depth === 2)) ) letters = letters.toUpperCase();
```

PR #6175, opened 2025-09-11
Add dist script for manual builds.

Fyorl:


package.json
```diff
@@ -7,6 +7,7 @@
     "build:db": "node ./utils/packs.mjs package pack",
     "build:json": "node ./utils/packs.mjs package unpack",
     "build:source": "node ./utils/packs.mjs package unpack",
+    "dist": "node dist.mjs",
```
arbron: Maybe this in the `utils` folder with the other build tooling to keep the top-level folder clearner.

```suggestion
    "dist": "node ./utils/dist.mjs",
```

PR #6176, opened 2025-09-11
[#4354, #4772] Hide/suppress riders for disabled enchantments

### LINKED ISSUE #4354: Feature Request: Allow storage of arbitrary data on Activity pseudo-documents
In essence, flags, but for Activities. Now that so much configuration has shifted from being on the item directly to being on Activities, it would be convenient for modules to be able to store data on the Activity pseudo-documents. 

Example to follow:

Context: 
- Region Attacher, a module which (in dnd5e) allows the GM to pre-configure a Region's behavior and enable it to "attach" to an item-created template. 
- Simple example: a spell which shelters a 30ft sphere from rain. With 30ft sphere as the template type, and the "Attach Region" checkbox enabled, and the region itself configured with a behavior to suppress weather effects, casting the spell would result in a visual confirmation of the behavior described in the spell.

Previous Behavior: 
- On an Item sheet, when target type was one which would result in a template prompt, would show a checkbox & button to enable attaching & configure behavior.
- Flags for whether to attach, as well as what the configured behavior should be, were stored on the Item under `flags.region-attacher`.
- When the template was created, the origin Item (if it existed) was checked for those flags, and logic proceeded accordingly (creating a region with the region behavior(s) if necessary).

Desired New Behavior:
- On an _Activity sheet_, when Area has a shape chosen, show checkbox & button in Area subsection to enable attaching & configure behavior.
- Flags (or pseudo-flags, or whatever format of storing arbitrary data makes sense) for whether to attach, as well as configured behavior, will be stored _on the Activity_ under `???.region-attacher`.
- When the template is created, the origin _Activity_ (if it exists) is checked for those arbitrary data, and logic proceeds accordingly.

Potential workaround:
- Store arbitrary data on the Item still, under `flags.region-attacher.<activity_id>`.
- Difficulty here is that this would require either a more intrusive UX (a new application to handle this) or interception of the default update behavior (adding a checkbox to an Activity sheet and changing its value will attempt to update the Activity - would have to intercept this and instead update the parent item).

Thanks for your consideration.

### LINKED ISSUE #4772: Activities granted by an enchantment are available when the enchantment is disabled
While making an item with multiple "forms," I wanted to use 4 enchantments, one for each form. The difference between each form is only that they each have 4 cast activities.

Thinking an enchantment could be used to add these activities, that turned out fine, except the activities are still available when any of the enchantments are toggled off.

Secondly, the enchantments can also not be deleted as the origin points to a uuid in a compendium (`Compendium.<pack key>.Item.EuDrnySNeo8TofOw.Activity.83zwntpxRs3fljWw`) and is using `fromUuidSync`.

### LINKED ISSUE #6198: Enchantments with Rider Activity that has Start of Turn activation time still triggers when the enchantment is not active.
I've been making a Swallow feature for this whale here, and it has a portion where at the start of each of the whale's turns while it has swallowed a creature, it deals 2d6 acid damage to the creature inside.

I made the feature self-enchanting with the Start of Turn Damage activity as a rider activity, yet it still seems to trigger the start of turn even when it shouldn't be since the item isn't enchanted yet?

Image

Not sure if this happens for other combat activation times, only tested start of turn.

arbron:
When an enchantment that contains rider items or effects is disabled, hide any rider items from the sheet, any rider activities from the item, and suppress and hide rider effects.

Adds `riderOrigin` flag to rider activities, effects, and items. This entailed adding support for flags on activities.

Closes #4354 
Closes #4772
Closes #6198

### DIFF
--- a/module/applications/actor/api/base-actor-sheet.mjs
+++ b/module/applications/actor/api/base-actor-sheet.mjs
@@ -761,7 +761,7 @@ export default class BaseActorSheet extends PrimarySheetMixin(
    * @protected
    */
   _assignItemCategories(item) {
-    if ( item.riderOrigin?.disabled ) return [];
+    if ( item.dependentOrigin?.active === false ) return [];
     if ( item.type === "container" ) return new Set(["containers", "inventory"]);
     if ( item.type === "spell" ) return new Set(["spells"]);
     if ( "inventorySection" in item.system.constructor ) return new Set(["inventory"]);
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -51,6 +51,19 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
   static LOCALIZATION_PREFIXES = [...super.LOCALIZATION_PREFIXES, "DND5E.ACTIVEEFFECT"];
 
   /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Another effect that granted this effect as a rider.
+   * @type {ActiveEffect5e|null}
+   */
+  get dependentOrigin() {
+    if ( !(this.parent instanceof Item) ) return null;
+    return this.parent.effects.get(this.flags.dnd5e?.dependentOn) ?? null;
+  }
+
+  /* -------------------------------------------- */
 
   /**
    * Is this effect an enchantment on an item that accepts enchantment?
@@ -85,7 +98,7 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     if ( this.type === "enchantment" ) return false;
     if ( this.parent instanceof dnd5e.documents.Item5e ) {
       if ( this.parent.areEffectsSuppressed ) return true;
-      if ( this.riderOrigin?.disabled ) return true;
+      if ( this.dependentOrigin?.active === false ) return true;
     }
     return false;
   }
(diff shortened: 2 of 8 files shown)


krbz999:
> This entailed adding support for flags on activities.

https://github.com/foundryvtt/dnd5e/issues/4354

module/documents/active-effect.mjs
```diff
@@ -91,6 +94,17 @@ export default class ActiveEffect5e extends ActiveEffect {
 
   /* -------------------------------------------- */
 
+  /**
+   * Another effect that granted this effect as a rider.
+   * @type {ActiveEffect5e|null}
+   */
+  get riderOrigin() {
+    if (!(this.parent instanceof Item)) return null;
```
```suggestion
    if ( !(this.parent instanceof Item) ) return null;
```

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -747,6 +747,7 @@ export default class BaseActorSheet extends PrimarySheetMixin(
    * @protected
    */
   _assignItemCategories(item) {
+    if ( item.riderOrigin?.disabled ) return [];
```
Fyorl: Should we be checking `disabled`? Or `active`?


```suggestion
    if ( item.riderOrigin?.active === false ) return [];
```

arbron: At the moment we don't suppress enchantments, but if the anti-magic stuff comes into play that will change so I'll swap it over.

Issue #3194, opened 2024-03-06
Using the token portrait over actor portrait on the character sheet does not respect wildcard tokens


krbz999:
It should probably pick a default.

Or if being extra, pick a random one on each render, which could be kinda funny.

BillPitcher:
Or the toggle could be disabled if wildcard. Anything is better than broken img links.

PR #6178, opened 2025-09-11
[#3077] Swap chat icon based on preferred sheet artwork

### LINKED ISSUE #3077: Feature Request: Add option to display token artwork instead of character artwork in chat messages
Currently, the character artwork is always used in chat messages. That makes certain scenarios very challenging.
To be more specific, I have an adventure where scarecrows are disguised as humans. I have one scarecrow actor and different token images for their human appearance. When I roll something with the disguised scarecrows, it always shows the scarecrow in the chat message.
This extends to other scenarios. Even having one "commoner" actor with different token images for NPCs, the chat messages will always show the one character artwork set on the commoner actor.
Having the option to show token artwork instead of character artwork would solve all of this.

arbron:
If a character or NPC has their token displayed on the sheet rather than their portrait, display that in the chat card also. Moves the logic for selecting the artwork into its own method so it can be shared between the sheet and the chat card, and cache the result so each chat card doesn't end up with a different random token.

Closes #3077

### DIFF
--- a/less/v2/chat.less
+++ b/less/v2/chat.less
@@ -46,16 +46,22 @@
         place-content: center;
       }
 
-      img {
+      img, video {
         width: 38px;
         height: 38px;
-        border-radius: 4px;
-        box-shadow: 0 0 6px rgb(0 0 0 / 85%);
-        object-fit: cover;
-        object-position: top;
         border: none;
         flex: none;
+      }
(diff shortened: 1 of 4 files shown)


krbz999:
https://github.com/foundryvtt/dnd5e/issues/3194

Issue #4934, opened 2024-12-27
"Spellcasting" for Check ability in `CheckActivity` should use roller's spellcasting


arbron:
Currently if "Spellcasting" is selected for the Check Ability then it will use the ability of the person who uses the activity. It would make more sense if it uses the ability for the person clicking the roll button in chat.

Blocked by https://github.com/foundryvtt/dnd5e/pull/4883

PR #6183, opened 2025-09-12
Add `spellcastingAbility` that takes highest ability

arbron:
Adds `spellcastingAbility` property to actors that returns the highest ability from among their spellcasting classes or their spellcasting attribute if no spellcasting classes are found.

Unblocks #4934
Unblocks #5092

### DIFF
--- a/module/data/activity/summon-data.mjs
+++ b/module/data/activity/summon-data.mjs
@@ -66,7 +66,7 @@ export default class BaseSummonActivityData extends BaseActivityData {
 
   /** @inheritDoc */
   get ability() {
-    return this.match.ability || super.ability || this.item.abilityMod || this.actor?.system.attributes?.spellcasting;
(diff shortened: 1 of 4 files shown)


module/data/activity/summon-data.mjs
```diff
@@ -91,7 +91,7 @@ export default class SummonActivityData extends BaseActivityData {
 
   /** @inheritDoc */
   get ability() {
-    return this.match.ability || super.ability || this.item.abilityMod || this.actor?.system.attributes?.spellcasting;
+    return this.match.ability || super.ability || this.item.abilityMod || this.actor?.system.spellcastingAbility;
```
krbz999: the new getter is on the `actor` tho

Fyorl (approved):
Should `ConsumableData#_typeAbilityMod` also use this?

PR #6188, opened 2025-09-15
[#6187] Add hint to consumption target field when not embedded

### LINKED ISSUE #6187: Add hints for consumption target fields on non-embedded items
There is no explanation for the expected values of the Item Uses, Material, or Attribute consumption targets on an activity on a non-embedded item. Adding a hint to those fields that makes it clear what is needed would be a useful feature for new users.

arbron:
Allows different consumption types to specify their own hints that will be displays above the target field when used on an item that isn't embedded.

Non-Embedded Hint

Closes #6187


lang/en.json
```diff
@@ -1627,6 +1628,7 @@
     },
     "ItemUses": {
       "Label": "Item Uses",
+      "NonEmbeddedHint": "UUID of compendium item that will be linked automatically to matching item on sheet.",
```
Fyorl: Needs updating now that this field also support identifiers.

PR #6192, opened 2025-09-15
[#6027, #6190] Build standard document creation dialog in AppV2

### LINKED ISSUE #6027: When an activity createDialog is rendered, cannot tab between inputs in other applications
All tabs focus on the Create Activity button.

### LINKED ISSUE #6190: Create shared document creation dialog used by the system
Currently the system has 3 separate document creation dialogs depending on what sort of dialogs are being created. If you are creating an item, it uses an implementation through `Item5e#createDialog` with the system's style, for creating activities and advancement it uses `PseudoDocument#createDialog` with the system's style, plus support for per-type hints, and for other documents types like actors and journal entry pages it uses core's standard dialog that is lacking the styling provided by the system.

These three methods should be standardized on one dialog that uses the system's dialog style and supports icons and hints for each document type. This would help avoid the large amount of duplicate logic we have now and standardize the styling of these dialogs. Since the current methods still call to the old AppV1 dialogs, this would be a good time to make this change.

arbron:
Creates a standard `CreateDocumentDialog` that replaces core's `Document#createDialog` to give a shared implementation using the system's design across actors, items, activities, and advancement. This allows items and actors to take advantage of the `_createDialogTypes` and `_createDialogData` methods that are used by activites and advancement to provide icons and hints to options.

Closes #6027
Closes #6190

 ### Todo
- [ ] Provide type-specific icons for each actor type
- [ ] Provide hints for item types
- [ ] Expand dialog to `JournalEntryPage` & `RegionBehavior`


arbron:
Added default artwork for the various actor types:

Screenshot 2025-11-11 at 11 14 38

module/applications/create-document-dialog.mjs
```diff
@@ -0,0 +1,224 @@
...
+    config.buttons.unshift(foundry.utils.mergeObject({
+      action: "ok", label: title, icon: "fa-solid fa-check", default: true
+    }, ok));
+
+    const { promise, resolve } = Promise.withResolvers();
+    const dialog = new this(config);
+    dialog.addEventListener("close", event => {
+      if ( !dialog.submitted ) return;
+      const createData = dialog.options.createData;
+      const createOptions = dialog.options.createOptions;
```
Fyorl:
```suggestion
      const { createData, createOptions } = dialog.options;
```

module/applications/create-document-dialog.mjs
```diff
@@ -0,0 +1,224 @@
...
+   * @param {object} createOptions
+   * @param {object} dialogOptions
+   */
+  static migrateOptions(createOptions, dialogOptions) {
+    const applicationOptions = {
+      top: "position", left: "position", width: "position", height: "position", scale: "position", zIndex: "position",
+      title: "window", id: "", classes: "", jQuery: ""
+    };
+
+    for (const [k, v] of Object.entries(createOptions)) {
```
Fyorl:
```suggestion
    for ( const [k, v] of Object.entries(createOptions) ) {
```

module/applications/create-document-dialog.mjs
```diff
@@ -0,0 +1,224 @@
...
+      if (k in applicationOptions) {
```
Fyorl:
```suggestion
      if ( k in applicationOptions ) {
```

module/applications/create-document-dialog.mjs
```diff
@@ -0,0 +1,224 @@
...
+        foundry.utils.logCompatibilityWarning("The ClientDocument.createDialog signature has changed. "
+          + "It now accepts database operation options in its second parameter, "
+          + "and options for DialogV2.prompt in its third parameter.", { since: 13, until: 15, once: true });
+        const dialogOption = applicationOptions[k];
+        if (dialogOption) foundry.utils.setProperty(dialogOptions, `${dialogOption}.${k}`, v);
```
Fyorl:
```suggestion
        if ( dialogOption ) foundry.utils.setProperty(dialogOptions, `${dialogOption}.${k}`, v);
```

lang/en.json
```diff
@@ -1203,7 +1204,9 @@
       }
     }
   },
-  "ThisTool": "This Tool"
+  "Hint": "Perform an ability check possibly assisted by a skill or tool.",
```
```suggestion
  "Hint": "Request an ability check possibly assisted by a skill or tool.",
```

lang/en.json
```diff
@@ -3499,7 +3502,12 @@
       }
     }
   },
-  "OnSave": "On Save"
+  "Hint": "Impose a saving throw roll with the option of rolling damage.",
```
```suggestion
  "Hint": "Request a saving throw with the option of rolling damage.",
```

Fyorl (changes requested):
I added some dark mode styles. I also noticed that the restrictions on types we place on certain tabs of the character/npc sheets are not being respected. Also some other minor comments. Other than that, this is looking good. Having a bespoke application is much nicer.

PR #6195, opened 2025-09-16
[#4398] Allow specification of `additional` initial filters when initializing Compendium Browser

### LINKED ISSUE #4398: Bug: Unable to pre-filter the Compendium Browser display
It is not possible to pre-filter the Compendium Browser to display, for example, only spells from a certain spell list, despite using the `initial` option when opening the Compendium Browser:

```
const filters = { locked: { types: new Set([ `spell` ]) } };
filters.initial = {
    documentClass: `Item`,
    types: new Set([ `spell` ]),
    additional: { spelllist: { class: { [`wizard`]: 1 } } }
};

const result = await game.dnd5e.applications.CompendiumBrowser.select(foundry.utils.mergeObject({ filters }, { selection: { min: 1, max: 2 } }, { inplace: false } ));
if (result) {
    console.log(result);
}
```

Despite setting the filters explicitly to state that only Wizard spells should display, all spells still show in the Compendium Browser when opened. I believe this relates to lines 782 and 801 of [CompendiumBrowser](https://github.com/foundryvtt/dnd5e/blob/4.0.x/module/applications/compendium-browser.mjs) (`_applyModeFilters` and `_applyTabFilters`) which actively delete the `additional` property, meaning it cannot be set using the `initial` property.

I have tested removing the two lines that delete the property, and the list can be pre-filtered successfully; however, the checkboxes in the sidebar do not correlate to this fact. The screenshot below shows that the list has been pre-filtered to only show Wizard spells (purposely set to **no** spells in my test environment) but the checkboxes do not respect this filter:

### Actual Result
Actual

### Expected Result
Expected

I am unclear as to the rationale behind deleting the `additional` key so have raised as an issue rather than as a Pull Request.

roth-michael:
Closes #4398
Current behavior deletes `this.#filters.additional` on tab filter application (which _always_ gets called at end of constructor unless there are locked filters), to keep filters from one tab from erroneously applying to another. 

Added an argument to `_applyTabFilters`, `keepFilters` (defaults to false) which, if true, _won't_ delete those `additional` filters. This makes it possible to do, for instance:
```js
new dnd5e.applications.CompendiumBrowser({
    tab: "spells",
    filters: {
        initial: {
            additional: {
                spelllist: {
                    "class:cleric": 1
                },
                level: {
                    max: 5
                }
            }
        }
    }
}).render(true)
```
and get an appropriately initially-filtered (but not locked) CB:
image

Additional logic provided for the special handling of the "feats" tab, so as to avoid clobbering `additional` filters there, too.

### DIFF
--- a/module/applications/compendium-browser.mjs
+++ b/module/applications/compendium-browser.mjs
@@ -69,7 +69,7 @@ export default class CompendiumBrowser extends Application5e {
       const isAdvanced = this._mode === this.constructor.MODES.ADVANCED;
       const tab = this.constructor.TABS.find(t => t.tab === this.options.tab);
       if ( !tab || (!!tab.advanced !== isAdvanced) ) this.options.tab = isAdvanced ? "actors" : "classes";
-      this._applyTabFilters(this.options.tab, true);
+      this._applyTabFilters(this.options.tab, { keepFilters: true });
     }
   }
 
@@ -828,11 +828,12 @@ export default class CompendiumBrowser extends Application5e {
 
   /**
    * Apply filters based on the selected tab.
-   * @param {string} id                   The tab ID.
-   * @param {boolean} [keepFilters=false] Whether to keep the existing additional filters
+   * @param {string} id                           The tab ID.
+   * @param {object} [options]                    Additional options
+   * @param {boolean} [options.keepFilters=false] Whether to keep the existing additional filters
    * @protected
    */
-  _applyTabFilters(id, keepFilters=false) {
+  _applyTabFilters(id, { keepFilters=false }={}) {
     const tab = this.constructor.TABS.find(t => t.tab === id);
     if ( !tab ) return;
     const { documentClass, types } = tab;


module/applications/compendium-browser.mjs
```diff
@@ -69,7 +69,7 @@ export default class CompendiumBrowser extends Application5e {
       const isAdvanced = this._mode === this.constructor.MODES.ADVANCED;
       const tab = this.constructor.TABS.find(t => t.tab === this.options.tab);
       if ( !tab || (!!tab.advanced !== isAdvanced) ) this.options.tab = isAdvanced ? "actors" : "classes";
-      this._applyTabFilters(this.options.tab);
+      this._applyTabFilters(this.options.tab, true);
```
Fyorl: Since this is a `@protected` method, and we want to avoid breaking the API later, we should use a full options object in case we need to add more later.

```suggestion
      this._applyTabFilters(this.options.tab, { keepFilters: true });
```

PR #6212, opened 2025-09-18
[#4962] Allow dynamic compendium browser filter choices

### LINKED ISSUE #4962: Add compendium browser filter for feat types
When in the simple mode in the compendium browser on the Feats tab, there is no way to filter between Origin, Epic Boon, General, or Fighting Style feats.

arbron:
Allows the choices for a `set` filter in the compendium browser to be a function rather than a static object. If a function is provided, then it will be passed the currently set of filters so it can be dynamically adjusted. This allows filtering the list of feature types down to just the relevant ones, such as just the feat types when viewing the "Feats" tab in standard mode.

Closes #4962

### DIFF
--- a/module/applications/compendium-browser.mjs
+++ b/module/applications/compendium-browser.mjs
@@ -554,7 +554,7 @@ export default class CompendiumBrowser extends Application5e {
     else if ( partId === "filters" ) {
       context.additional = Array.from(context.filterDefinitions?.entries() ?? []).reduce((arr, [key, data]) => {
         // Special case handling for 'Feats' tab in basic mode.
-        if ( (types[0] === "feat") && ((key === "category")) ) return arr;
+        if ( (types[0] === "feat") && (key === "category") ) return arr;
 
         let sort = 0;
         switch ( data.type ) {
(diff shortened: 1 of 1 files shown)


module/applications/compendium-browser.mjs
```diff
@@ -557,7 +554,7 @@ export default class CompendiumBrowser extends Application5e {
     else if ( partId === "filters" ) {
       context.additional = Array.from(context.filterDefinitions?.entries() ?? []).reduce((arr, [key, data]) => {
         // Special case handling for 'Feats' tab in basic mode.
-        if ( (types[0] === "feat") && ((key === "category") || (key === "subtype")) ) return arr;
```
Fyorl, line 514: Some weirdness if you toggle a filter on the feats tab of the compendium browser in non-advanced mode, all of the advanced filters for `feat` type items re-appear.

Fyorl (changes requested):

module/applications/compendium-browser.mjs
arbron: Aw, its because I'm re-rendering the `filters` part in `#onSetFilter` without passing in the `dnd5e.browser.types`, so it doesn't have the data it needs.

PR #6215, opened 2025-09-19
[#6210] Limit group travel pace if any members are below half speed

### LINKED ISSUE #6210: Incorporate rules on Slower Travelers into group sheet travel pace calculation
From the DMG:
> **Slower Travelers.** The group must move at a Slow pace if any group member’s Speed is reduced to half or less of normal.

arbron:
Limits the group's pace to "Slow" if any members are below half their normal walking speed. This is done by adding a new `slowed` boolean to actor's movement data. Currently this only takes into account speed reductions from status effects, not from any active effects that might be reducing their speed.

Created a new `effectivePace` getter and moved the pace label into a getter also to ensure that changes to actor's speeds can be reflected on the group actor after a render, rather than requiring a complete re-preparation of the group actor data.

Closes #6210

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -4603,7 +4603,8 @@
     "Fast": "Fast",
     "Normal": "Normal",
     "Slow": "Slow"
-  }
+  },
+  "Slowed": "Slowed"
 },
 
 "DND5E.Treasure": {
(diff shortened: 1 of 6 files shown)


Fyorl:
Closing in favour of #6316

PR #6219, opened 2025-09-19
[#4957] Automatically scale down titles that don't fit

### LINKED ISSUE #4957: [UI] Activity selection title looks bad for items with long names
image

arbron:
Adds some resizing of the title in all system applications to try and prevent the title from being cut off. The title can be scaled down up to 50%.

Title Resizing

This is achieved by creating a second copy of the title that isn't visible and isn't constrained by the window's width. On render or resize the width of this invisible copy is compared against the width of the window, and if it is wider then a scaling factor is applied via CSS to the window's title.

Closes #4957

### DIFF
--- a/less/v2/apps.less
+++ b/less/v2/apps.less
@@ -1133,7 +1133,7 @@
 
     &:has(h2:not(:empty)) { margin-block-end: calc(var(--font-size-24) + 12px); }
 
-    h1, h2 {
+    h1, h2, .title-measure {
       position: absolute;
       inset-inline: 0;
       color: var(--color-text-primary);
@@ -1144,13 +1144,19 @@
     }
     h1 {
       inset-block-start: calc(var(--header-height) - 20px);
-      font-size: var(--font-size-24);
+      font-size: calc(var(--font-size-24) * var(--font-size-scale, 1));
     }
     h2 {
       inset-block-start: calc(var(--header-height) + var(--font-size-16));
       margin: 0;
       font-size: var(--font-size-16);
       text-shadow: none;
+    }
+    .title-measure {
+      font-size: var(--font-size-24);
+      inset: unset;
+      overflow: visible;
+      text-wrap: nowrap;
     }
 
     .header-control:not(.unbutton) { background: var(--dnd5e-background-25); }
@@ -1170,6 +1176,7 @@
       border-radius: 4px;
     }
   }
+  .title-measure { visibility: hidden; }
 
   &.minimized .window-header {
     h1 { all: revert-layer; }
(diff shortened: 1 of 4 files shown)


less/v2/apps.less
```diff
@@ -1144,16 +1160,22 @@
     }
     h1 {
       inset-block-start: calc(var(--header-height) - 20px);
-      font-size: var(--font-size-24);
+      font-size: calc(var(--font-size-24) * var(--font-size-scale, 1));
```
Fyorl, line 1147: ~~I wonder if we can use container units for this, which would save us the extra element and `requestAnimationFrame` stuff.~~

```css
.window-header {
  container-type: inline-size;
  h1 { font-size: min(var(--font-size-18), 90cqi); }
}
```

Fyorl (changes requested):

less/v2/apps.less
Fyorl: Nevermind, I've very much misunderstood what we were trying to do here.

module/applications/api/application-v2-mixin.mjs
```diff
@@ -280,16 +254,30 @@ export default function ApplicationV2Mixin(Base, { handlebars=true }={}) {
...
+     * Scale the font size of the app's title to ensure it fits.
+     */
+    #scaleTitle() {
+      requestAnimationFrame(() => {
+        const maxWidth = this.element.offsetWidth - 60;
+        const fullWidth = this.window.titleMeasure.offsetWidth;
+        if ( fullWidth > maxWidth ) {
+          this.window.title.style.setProperty("--font-size-scale", Math.clamp(maxWidth / fullWidth, 0.5, 1));
+        }
+      });
```
Fyorl, line 330: Was trying to see if there's a way to not have to use the `.title-measure` because it's kind of a lot for just this purpose. The issue is we need to know the unconstrained size of the header text, which is difficult to do with the existing header styles. Turns out there is an API for reading pure text box. Since the title doesn't wrap, it just clips, we can use this:

```js
requestAnimationFrame(() => {
  const { title } = this.window;
  if ( !this.element || !title ) return;
  // Divide by existing font scale to recover original size.
  const scale = parseFloat(title.style.getPropertyValue("--font-size-scale")) || 1;
  const range = document.createRange();
  range.selectNodeContents(title);
  const fullWidth = range.getBoundingClientRect().width / scale;
  const maxWidth = this.element.offsetWidth - 60;
  if ( fullWidth ) title.style.setProperty("--font-size-scale", Math.clamp(maxWidth / fullWidth, 0.5, 1));
});
```

krbz999: Could clamp the font-size using `font-size: clamp(x, y, z)` with the parent element having `container-type: inline-size`.

Fyorl (approved):

PR #6227, opened 2025-09-21
[#6226] Migrate advancement data to being stored in object

### LINKED ISSUE #6226: Migrate advancement data storage from array to object to match activities
Standardizing storage of advancement and activities should allow these two pseudo-document types to share more code, and will make performing CRUD operations on advancement documents simpler.

arbron:
Changes the advancement data on an item to be stored in an object rather than in an array and modifies the various methods for CRUD operations on advancement data to reflect the change.

Closes #6226


module/data/item/templates/advancement.mjs
```diff
@@ -28,7 +28,24 @@ export default class AdvancementTemplate extends SystemDataModel {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Migrate advancement data to be stored in an object, not an array.
+   * @param {object} source  Candidate source data to migrate.
+   */
+  static #migrateStorage(source) {
+    if ( foundry.utils.getType(source.advancement) === "Array" ) {
```
Fyorl: `Array.isArray` is faster.
```suggestion
    if ( Array.isArray(source.advancement) ) {
```

Fyorl (approved):
Looks good, the only issue I was able to find during testing was that I was able to swap out a prerequisite invocation at a lower level (via modify choices, not via the replacement function of the ItemChoice advancement), and the following invocation choices weren't appropriately invalidated. I think we probably have always had this issue but I wanted to flag it in case it's a regression.

PR #6231, opened 2025-09-23
Anticipate possibility of `null` author in `ChatMessage5e`

stwlam:
Noticed during semi-related core testing


arbron:
Should probably mark https://github.com/foundryvtt/foundryvtt/issues/13105 as breaking.

stwlam:
> Should probably mark [foundryvtt/foundryvtt#13105](https://github.com/foundryvtt/foundryvtt/issues/13105) as breaking.

This was to address a V13 bug: `author` data can be initialized to `null` even if the field isn't nullable

PR #6233, opened 2025-09-23
[#2291, #5770] Tie item uses recovery into calendar system

### LINKED ISSUE #2291: Add a function on actor sheets to recharge items on a 'new day' without resting
A new day might come without a player taking a rest at all.

### LINKED ISSUE #5770: Item recharge interaction with gritty realism settings
### Problem
When setting the variant rest rules for gritty realism, it causing items that recharge on dawn/dusk/day to multiply their recharge formula.

For example, an item that should recharge 1d4 charges at dawn, will roll 7d4 instead if you select short reset and use the new day button.
Disabling the gritty realism rest rules makes the item recharge correctly.

This does not happen if the item is configured to recharge on short rest.

### Expected behaviour
When selecting short rest, and clicking the new day checkbox, items configured to roll dice for recharge on dawn/day/dusk roll the number of dice configured on the item.

arbron:
Allows recovery for for periods of "day", "dawn", or "dusk" to be triggered by the passage of world time rather that through the rest dialog. This is gated by a setting that by default is tied to whether the calendar is enabled, but can be set to manual if needed (for example, if using a module-provided calendar).

Some changes were made to the `recoverUses` API to support the passage of multiple relevant recovery periods if the time was advanced more than one day. This also fixes the bug with short rests when using the gritty realism rest setting.

When a time passes the system checks to see if any relevant recovery periods should trigger and if so, triggers recovery for all world and actor items. If any changes were made, it then creates a chat message to indicate the time passage and display any deltas. These deltas are gated by observer permissions.

Time Passage Messages

The created chat message also has a description of the time that passed (e.g. "The sun has risen." or "Two days have passed."). In order to support better localization of these times, some updates have been made to the number formatting system to support the `words` (and `numerals`) settings when formatting with units.

One potential point of improvement is how we handle selection of recovery periods. At the moment, we only select the first recovery profile that matches a provided period. This may cause issues for cases where a "dawn" and "dusk" may have passed in one update. If you had an item that was only usable during the daytime, you might say it recovers all uses at dawn and loses them at dusk. This works fine so long as you do not advance past both periods at the same time. In the future the recovery uses method might be update with the option of selecting multiple profiles so long as they are not marked as mutually exclusive.

Closes #2291
Closes #5770

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -28,6 +28,7 @@
 "TYPES.ChatMessage": {
   "request": "Request Message",
   "rest": "Rest Message",
+  "timePassed": "Time Passed Message",
   "turn": "Combat Turn Message",
   "usage": "Usage Message"
 },
@@ -1279,6 +1280,13 @@
       "hint": "Allow the calendar UI to be viewed & customized by all users.",
       "label": "Enabled"
     },
+    "dailyRecovery": {
+      "calendar": "Calendar Recovery (via time passing)",
+      "default": "Default (tied to calendar enabled setting)",
+      "hint": "Control whether limited uses with day, dawn, and dusk recovery are recovered based on the world calendar or through the rest dialog.",
+      "label": "Daily Recovery Mode",
+      "manual": "Manual Recovery (via rest dialog)"
+    },
     "formatters": {
       "date": {
         "label": "Date Format"
@@ -1468,6 +1476,17 @@
       "Autumn": "Autumn",
       "Winter": "Winter"
     }
+  },
+  "TimePassage": {
+    "Midnight": "A new day has begun.",
+    "Sunrise": "The sun has risen.",
+    "Sunset": "The sun has set.",
+    "SunsetSunrise": "The sun has set and a new dawn of a new day has come.",
+    "TimePassed": {
+      "one": "{number} has passed.",
+      "other": "{number} have passed."
+    },
+    "Title": "Passage of Time"
   }
 },
 
--- a/lang/en.json
+++ b/lang/en.json
@@ -1481,7 +1481,7 @@
     "Midnight": "A new day has begun.",
     "Sunrise": "The sun has risen.",
     "Sunset": "The sun has set.",
-    "SunsetSunrise": "The sun has set and a new dawn of a new day has come.",
+    "SunsetSunrise": "The sun has set and the dawn of a new day has come.",
     "TimePassed": {
       "one": "{number} has passed.",
       "other": "{number} have passed."
--- a/module/applications/actor/rest/base-rest-dialog.mjs
+++ b/module/applications/actor/rest/base-rest-dialog.mjs
@@ -102,7 +102,7 @@ export default class BaseRestDialog extends Dialog5e {
    */
   get promptNewDay() {
     // Only prompt if rest is longer than 10 minutes and less than 24 hours
-    return (this.duration > 10) && (this.duration < 1440);
+    return dnd5e.settings.calendarConfig.manualRecovery && (this.duration > 10) && (this.duration < 1440);
   }
 
   /* -------------------------------------------- */
--- a/module/data/calendar/calendar-data.mjs
+++ b/module/data/calendar/calendar-data.mjs
@@ -1,7 +1,8 @@
-import { formatNumber } from "../../utils.mjs";
+import { convertTime, formatNumber, formatTime, getPluralRules } from "../../utils.mjs";
+import { IndividualDeltaField } from "../chat-message/fields/deltas-field.mjs";
 
 /**
- * @import { CalendarFormattingContext, CalendarTimeDeltas } from "./_types.mjs";
+ * @import { CalendarFormattingContext, CalendarTimeDeltas, TimePassageData } from "./_types.mjs";
  */
 
 /**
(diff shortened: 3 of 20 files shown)


module/applications/actor/rest/base-rest-dialog.mjs
```diff
@@ -71,7 +71,7 @@ export default class BaseRestDialog extends Dialog5e {
     const duration = CONFIG.DND5E.restTypes[this.config.type]
       ?.duration?.[game.settings.get("dnd5e", "restVariant")] ?? 0;
     // Only prompt if rest is longer than 10 minutes and less than 24 hours
-    return (duration > 10) && (duration < 1440);
+    return game.settings.get("dnd5e", "calendarConfig").manualRecovery && (duration > 10) && (duration < 1440);
```
krbz999: Maybe I've overlooked some context but should the day (1440) not be derived from the calendar?

arbron: Well, this is only used when you are not using calendar-based recovery anyway, so I'm not sure how important it is.

lang/en.json
```diff
@@ -1358,6 +1377,17 @@
       "Autumn": "Autumn",
       "Winter": "Winter"
     }
+  },
+  "TimePassage": {
+    "Midnight": "A new day has begun.",
+    "Sunrise": "The sun has risen.",
+    "Sunset": "The sun has set.",
+    "SunsetSunrise": "The sun has set and a new dawn of a new day has come.",
```
```suggestion
    "SunsetSunrise": "The sun has set and the dawn of a new day has come.",
```

module/data/calendar/calendar-data.mjs
```diff
@@ -327,6 +328,122 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+          if ( !foundry.utils.isEmpty(result?.updates) ) {
+            deltas.push({ uuid: item.uuid, deltas: IndividualDeltaField.getDeltas(item, result.updates) });
+            updates.push({ _id: item.id, ...result.updates });
+            rolls.push(...result.rolls);
+          }
+        }
+      };
+
+      const itemUpdates = [];
+      await performItemUpdates(game.items, itemUpdates);
```
Fyorl: An item not on an Actor I would imagine is kind of 'inert', no? When would we expect that such an item recovers its uses over the passage of time?

module/data/calendar/calendar-data.mjs
```diff
@@ -327,6 +328,122 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+   * Trigger recovery for any world items based on the passage of time.
+   * @param {TimePassageData} timePassageData
+   */
+  static async handleTimePassage(timePassageData) {
+    const periods = new Map();
+    for ( const [d, p] of CONFIG.DND5E.calendarDeltasRecoveryMapping.entries() ) {
+      if ( timePassageData[d] ) periods.set(p, timePassageData[d]);
+    }
+
+    const deltas = [];
```
Fyorl: Avoid sweeping `game.items` and `game.actors` if there's nothing to do.
```suggestion
    if ( !periods.size ) return;
    const deltas = [];
```

module/data/calendar/calendar-data.mjs
```diff
@@ -327,6 +328,122 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Trigger recovery for any world items based on the passage of time.
+   * @param {TimePassageData} timePassageData
+   */
+  static async handleTimePassage(timePassageData) {
+    const periods = new Map();
```
Fyorl, line 353: Seems we only create a message if we have `deltas` and we only have `deltas` if we are doing calendar recovery.

```suggestion
    if ( !dnd5e.settings.calendarConfig.manualRecovery ) return;
    const periods = new Map();
```

Then we don't need the if-statement

module/data/calendar/calendar-data.mjs
```diff
@@ -327,6 +328,122 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+    CalendarData5e.handleTimePassage(timePassageData);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Trigger recovery for any world items based on the passage of time.
+   * @param {TimePassageData} timePassageData
+   */
+  static async handleTimePassage(timePassageData) {
```
Fyorl, line 352: I think so long as we bail early if `!periods.size` then scanning all items on all world actors is probably fine, since we would only ever do it if there's a long rest or the DM is advancing time a lot.

But even a normal adventure will import a fair few actors. I had some considerations for how we could improve performance so as to not have to worry about this method at all:
1. Only scan `character`s, `vehicle`s, and `npc`s that are in `group`s or flagged as `important`.
2. Move the `performItemUpdates` closure out to a static helper.
3. `getRollData` is pretty expensive, but we call it indiscriminately even though it's only necessary for formula-type recovery. Could we make the `rollData` param optional and have the item call `getRollData` on itself only when absolutely required?

module/data/calendar/calendar-data.mjs
```diff
@@ -327,6 +328,122 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+  static generateTimePassageMessage(timePassageData) {
+    let message;
+
+    // X days have passed
+    if ( timePassageData.midnights > 1 ) {
+      const pr = getPluralRules();
+      const { value, unit } = convertTime(timePassageData.midnights, "day", { strict: false });
+      const number = formatTime(value, unit, { words: true });
+      message = game.i18n.format(`DND5E.CALENDAR.TimePassage.TimePassed.${pr.select(value)}`, {
+        number: number.capitalize(), numberLc: number
```
Fyorl, line 425: What's `numberLc` for?

module/data/chat-message/_types.mjs
```diff
@@ -60,6 +60,19 @@
...
 
+/**
+ * @typedef TimePassedMessageSystemData
+ * @property {DocumentDeltasData[]} changes  Item recovery from this time change.
+ */
+
+/**
+ * @typedef DocumentDeltasData
+ * @param {IndividualDeltaData} deltas  Data deltas for an document update.
+ * @param {string} uuid                 UUID of the document to which the deltas apply.
```
Fyorl:
```suggestion
 * @property {IndividualDeltaData[]} deltas  Data deltas for a document update.
 * @property {string} uuid                 UUID of the document to which the deltas apply.
```

module/data/settings/calendar-setting.mjs
```diff
@@ -19,9 +19,26 @@ export class CalendarConfigSetting extends foundry.abstract.DataModel {
...
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Should item usage recovery be handled manually through the rest dialog?
+   * @type {boolean}
+   */
+  get manualRecovery() {
+    return this.dailyRecovery === "manual" || (!this.dailyRecovery && !this.enabled);
```
Fyorl:
```suggestion
    return (this.dailyRecovery === "manual") || (!this.dailyRecovery && !this.enabled);
```

module/data/calendar/calendar-data.mjs
```diff
@@ -327,6 +328,122 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+   * `onUpdateWorldTime` and registered late to allow modules to modify the time deltas with their
+   * own entries before time passage is handled.
+   * @param {number} worldTime
+   * @param {number} deltaTime
+   * @param {object} options
+   * @param {string} userId
+   */
+  static onTimePassage(worldTime, deltaTime, options, userId) {
+    if ( !game.user.isActiveGM || (deltaTime <= 0) ) return;
+    const timePassageData = { worldTime, deltaTime, ...options.dnd5e.deltas };
```
Fyorl:
```suggestion
    const timePassageData = { worldTime, deltaTime, ...(options.dnd5e?.deltas ?? {}) };
```

module/data/calendar/calendar-data.mjs
```diff
@@ -327,6 +328,122 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+    const deltas = [];
+    const rolls = [];
+
+    if ( !game.settings.get("dnd5e", "calendarConfig").manualRecovery ) {
+      const performItemUpdates = async (items, updates) => {
+        for ( const item of items ) {
+          const result = await item.system.recoverUses?.(periods, item.getRollData()) ?? {};
+          if ( !foundry.utils.isEmpty(result?.updates) ) {
+            deltas.push({ uuid: item.uuid, deltas: IndividualDeltaField.getDeltas(item, result.updates) });
+            updates.push({ _id: item.id, ...result.updates });
```
Fyorl, line 375: Do we need to handle `result.destroy`?

module/data/chat-message/time-passed-message-data.mjs
```diff
@@ -0,0 +1,64 @@
+import ChatMessageDataModel from "../abstract/chat-message-data-model.mjs";
+import ActivationsField from "./fields/activations-field.mjs";
```
Fyorl: Not used

Fyorl (changes requested):

module/data/calendar/calendar-data.mjs
arbron: So languages that don't put the number at the start of the sentence can control capitalization.

module/data/calendar/calendar-data.mjs
arbron: My intention here is to give modules a chance to use the message even if nothing is recovered. In that case the `dnd5e.preCreateTimePassedMessage` hook is fired and modules can set `create: true` and make any changes to the message they like.

Fyorl: Yeah, we can still call the hook. Maybe it's a bit neater to check `manualRecovery` and `periods.size` near the top here and call the hook before early-return.

module/data/calendar/calendar-data.mjs
arbron: I've eliminated world item updates so I can inline the `performItemUpdates` helper and I've added on-demand `getRollData`. I've also moved the updates themselves into a batch update.

module/data/calendar/calendar-data.mjs
arbron: Yep, another feature that is newer than this PR :)

module/data/calendar/calendar-data.mjs
```diff
@@ -327,6 +328,121 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+  static onTimePassage(worldTime, deltaTime, options, userId) {
+    if ( !game.user.isActiveGM || (deltaTime <= 0) ) return;
+    const timePassageData = { worldTime, deltaTime, ...(options.dnd5e?.deltas ?? {}) };
+    CalendarData5e.handleTimePassage(timePassageData);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Trigger recovery for any world items based on the passage of time.
```
Fyorl:
```suggestion
   * Trigger recovery for any world actors based on the passage of time.
```

module/data/item/templates/activities.mjs
```diff
@@ -349,6 +358,9 @@ export default class ActivitiesTemplate extends SystemDataModel {
       }
     };
 
+    if ( !rollData && (this.activities.some(a => a.uses?.recovery.some(u => u.type === "formula"))
+      || this.uses.recovery.some(u => u.type === "formula")) ) rollData = this.parent.getRollData();
```
Fyorl: I think we can make this a teeny bit faster.

```suggestion
    if ( !rollData && this.uses.recovery.some(u => u.type === "formula")) ) rollData = this.parent.getRollData();
```

Then in the `for` loop below:
```js
if ( activity.dependentOrigin?.active === false ) continue;
if ( !rollData && activity.uses?.recovery.some(u => u.type === "formula") ) rollData = this.parent.getRollData();
```

module/documents/actor/actor.mjs
```diff
@@ -2571,17 +2571,20 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
     const restConfig = CONFIG.DND5E.restTypes[config.type];
-    const recovery = Array.from(restConfig.recoverPeriods ?? []);
-    if ( recoverShortRestUses ) recovery.unshift("sr");
-    if ( recoverLongRestUses ) recovery.unshift("lr");
-    if ( recoverDailyUses || config.newDay ) recovery.unshift("day", "dawn", "dusk");
+    let recovery = Array.from(restConfig.recoverPeriods ?? []).map(p => [p, 1]);
+    if ( recoverShortRestUses ) recovery.unshift(["sr", 1]);
+    if ( recoverLongRestUses ) recovery.unshift(["lr", 1]);
+    if ( (recoverDailyUses || config.newDay) && game.settings.get("dnd5e", "calendarConfig").manualRecovery ) {
+      const days = (game.settings.get("dnd5e", "restVariant") === "gritty") && (config.type === "long") ? 7 : 1;
```
```suggestion
    if ( (recoverDailyUses || config.newDay) && dnd5e.settings.calendarConfig.manualRecovery ) {
      const days = (dnd5e.settings.restVariant === "gritty") && (config.type === "long") ? 7 : 1;
```

Fyorl (approved):

PR #6236, opened 2025-09-24
Rework pseudo document collections to use stable instances

arbron:
Changes how pseudo document collections act to use a store of stable instances stored on the item in the same way that normal embedded documents are handled. This fixes a number of issues where updates to the pseudo documents would result in disconnected instances in applications.


module/data/fields/pseudo-document-collection-field.mjs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    // Reconstruct the source array, retaining object references
+    for ( const [id, d] of Object.entries(diff) ) {
+      if ( foundry.utils.isDeletionKey(id) ) {
+        if ( id[0] === "-" ) {
+          delete source[key][id.slice(2)];
+          continue;
+        }
+        id = id.slice(2);
```
krbz999, line 61: Reassigning a const.

module/data/fields/psuedo-document-field.mjs
```diff
@@ -0,0 +1,76 @@
+/**
+ * Data field that selects the appropriate PseudoDocument data model if available, otherwise defaults to generic
+ * `ObjectField` to prevent issues with custom pseudo-document types that aren't currently loaded.
+ */
+export default class PsuedoDocumentField extends foundry.data.fields.ObjectField {
```
krbz999: Psue -> Pseu

module/data/abstract/pseudo-document-collection.mjs
```diff
@@ -0,0 +1,350 @@
...
+
+  /**
+   * Add a document to the collection.
+   * @param {string} key                           The embedded Document ID.
+   * @param {PseudoDocument} value                 The embedded Document instance.
+   * @param {object} [options]                     Additional options to the set operation.
+   * @param {boolean} [options.modifySource=true]  Whether to modify the collection's source as part of the operation.
+   * */
+  set(key, value, {modifySource=true, ...options}={}) {
+    if ( modifySource ) this._set(key, value, options);
```
krbz999, line 260: `_set` doesnt actually take a third param.

module/data/abstract/pseudo-document-collection.mjs
```diff
@@ -0,0 +1,351 @@
...
+
+  /* -------------------------------------------- */
+  /*  Collection Initialization                   */
+  /* -------------------------------------------- */
+
+  /**
+   * Initialize the EmbeddedCollection by synchronizing its Document instances with existing _source data.
+   * Importantly, this method does not make any modifications to the _source array.
+   * It is responsible for creating, updating, or removing Documents from the Collection.
+   * @param {DataModel} parent                       Parent containing this collection.
```
krbz999, line 121: called "parent" here but is "model"

module/data/fields/pseudo-document-field.mjs
```diff
@@ -0,0 +1,76 @@
+/**
+ * Data field that selects the appropriate PseudoDocument data model if available, otherwise defaults to generic
+ * `ObjectField` to prevent issues with custom pseudo-document types that aren't currently loaded.
+ */
+export default class PseudoDocumentField extends foundry.data.fields.ObjectField {
```
krbz999, line 5: This file is also named "psuedo"

module/data/fields/_module.mjs
```diff
@@ -1,9 +1,13 @@
...
-export {default as AdvancementCollectionField} from "./advancement-collection-field.mjs";
 export {default as AdvancementDataField} from "./advancement-data-field.mjs";
-export {default as AdvancementField} from "./advancement-field.mjs";
 export {default as AdvantageModeField} from "./advantage-mode-field.mjs";
 export {default as FormulaField} from "./formula-field.mjs";
 export {default as IdentifierField} from "./identifier-field.mjs";
 export {default as LocalDocumentField} from "./local-document-field.mjs";
 export {default as MappingField} from "./mapping-field.mjs";
+export {default as PseudoDocumentCollectionField} from "./pseudo-document-collection-field.mjs";
+export {default as PsuedoDocumentField} from "./pseudo-document-field.mjs";
```
thatlonelybugbear, line 8: typo in `PsuedoDocumentField` => `PseudoDocumentField`

Fyorl (commented):
I would need to understand a bit better why the scope of these changes are necessary to support the upcoming work. Can you share an example of the way in which advancements or activity instances are currently not stable?

arbron:
> I would need to understand a bit better why the scope of these changes are necessary to support the upcoming work. Can you share an example of the way in which advancements or activity instances are currently not stable?

The main issue I was having was with applying changes using `updateSource`. When `updateSource` is called from the advancement itself, either on the advancement or the actor containing it, it would trigger a re-preparation of the item data which completely recreates the advancement instances. This would lead to the `this` in the apply method referring to an out of date instance of the advancement, and any other operations through the end of the method would be working with stale data.

This became a particular problem with `HitPointsAdvancement`, where in order to do the dynamic updating I was first attempting to un-apply the previously applied hit point changes and then applying the new changes.

There are potentially other ways to solve this in the advancement, such as re-fetching the advancement from the item after every single update rather than calling `this`, but they all lead to a lot of manual work for something that should just work as expected and could lead to a lot of bugs.

Fyorl:
> The main issue I was having was with applying changes using `updateSource`. When `updateSource` is called from the advancement itself, either on the advancement or the actor containing it, it would trigger a re-preparation of the item data which completely recreates the advancement instances. This would lead to the `this` in the apply method referring to an out of date instance of the advancement, and any other operations through the end of the method would be working with stale data.
> 
> This became a particular problem with `HitPointsAdvancement`, where in order to do the dynamic updating I was first attempting to un-apply the previously applied hit point changes and then applying the new changes.
> 
> There are potentially other ways to solve this in the advancement, such as re-fetching the advancement from the item after every single update rather than calling `this`, but they all lead to a lot of manual work for something that should just work as expected and could lead to a lot of bugs.

That clears things up, thank you. I think for `updateSource`, now that we have migrated advancements to a `MappingField`, I would expect calling `updateSource` on the advancement directly to work. It seems to not break any references when I test on v13 & v14. We would also need to update `Item5e#updateAdvancement` to call `updateSource` on the advancement directly in the `source: true` case. If that works and gives us no trouble then I think that would be the most preferable solution.

If there are still issues with that method, then I have attached a [patch](https://github.com/user-attachments/files/25025291/stable-advancements.patch) of what I think are the minimal changes required in order to maintain stable references. In summary:
- Write the `pseudoCollections` as you have done here.
- Adjust `AdvancementCollectionField#initialize`.
- Adjust `AdvancementField#_updateCommit`.

If there are still problems that will mean we cannot support the future work then we can go with this PR, but I'd rather avoid it if there are more lightweight solutions. We are copying a lot of code from `EmbeddedCollection`, much of which is not useful for these pseudo-documents, as well as copying other field implementations that we will have to maintain in the future. I can see that already some of these field implementations need adjusting for v14.

arbron:
> That clears things up, thank you. I think for `updateSource`, now that we have migrated advancements to a `MappingField`, I would expect calling `updateSource` on the advancement directly to work. It seems to not break any references when I test on v13 & v14. We would also need to update `Item5e#updateAdvancement` to call `updateSource` on the advancement directly in the `source: true` case. If that works and gives us no trouble then I think that would be the most preferable solution.
> 
> If there are still issues with that method, then I have attached a [patch](https://github.com/user-attachments/files/25025291/stable-advancements.patch) of what I think are the minimal changes required in order to maintain stable references. In summary:
> 
> * Write the `pseudoCollections` as you have done here.
> * Adjust `AdvancementCollectionField#initialize`.
> * Adjust `AdvancementField#_updateCommit`.
> 
> If there are still problems that will mean we cannot support the future work then we can go with this PR, but I'd rather avoid it if there are more lightweight solutions. We are copying a lot of code from `EmbeddedCollection`, much of which is not useful for these pseudo-documents, as well as copying other field implementations that we will have to maintain in the future. I can see that already some of these field implementations need adjusting for v14.

Did some testing of my other advancement branches and they seem to be working better with just the `updateSource` change. Maybe it would be best for you to start reviewing the UI stuff (starting here: https://github.com/foundryvtt/dnd5e/pull/6053, which I've reworked to not be based on this PR) and we can reevaluate them as we go through to make sure everything is still working as expected.

PR #6237, opened 2025-09-24
[#5886] Allow consumption targets to link with identifiers

### LINKED ISSUE #5886: Add support for identifiers in addition to UUIDs in item use consumption
It would be easier if item use consumption would allow to enter an identifier and when the item is added to an actor, the identifier is looked up in the actor’s items like it is done with item’s UUID. 
This would facilitate preparing items in a Compendium, especially when an item use consumption needs to refer to an item that could come from different Compendiums.

arbron:
Adds support for re-linking consumption using item identifiers in addition to UUIDs. When using identifiers, they can use either a straight identifier (e.g. `channel-divinity`) or an identifier prefixed with an item type (e.g. `spell:divine-strike`).

Closes #5886

### DIFF
--- a/module/data/activity/base-activity.mjs
+++ b/module/data/activity/base-activity.mjs
@@ -756,7 +756,7 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
    * @internal
    */
   _remapConsumptionTarget(target) {
-    if ( !this.actor || this.actor.items.has(target) ) return target;
+    if ( !target || !this.actor || this.actor.items.has(target) ) return target;
 
     // Re-link UUID target
     if ( target.startsWith("Compendium.") ) {


module/data/activity/base-activity.mjs
```diff
@@ -752,6 +749,35 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
...
   /* -------------------------------------------- */
 
+  /**
+   * Remap a UUID or identifier in a consumption target to the ID of an item on the actor.
+   * @param {string} target
+   * @returns {string}
+   * @internal
+   */
+  _remapConsumptionTarget(target) {
+    if ( !this.actor || this.actor.items.has(target) ) return target;
```
Fyorl: Just to be safe
```suggestion
    if ( !target || !this.actor || this.actor.items.has(target) ) return target;
```

Fyorl (approved):

PR #6239, opened 2025-09-24
[#4708] Rebuild `HitPointsFlow` for V2

arbron:
Rework `HitPointsFlow` for application V2. The new interface now displays the hit points calculation that is occuring including the ability modifier and per-level bonuses.

Screenshot 2025-09-24 at 15 44 17


arbron:
Stuck the checkbox into a fieldset:

Screenshot 2025-09-24 at 19 35 07

templates/advancement/hit-points-flow.hbs
```diff
@@ -1,27 +1,80 @@
...
-    {{#if isFirstClassLevel}}
+            {{!-- Selected HP Increase --}}
+            {{#if isFirstClassLevel}}
+            <div class="form-group label-top">
+                <label>{{ localize "DND5E.ADVANCEMENT.HitPoints.Breakdown.Max" }}</label>
+                <span>{{ hp.max }}</span>
+            </div>
+            {{else if manual}}
+            <div class="form-group label-top">
+                <label>value</label>
```
Fyorl: Not localised. Also, I think something like 'roll' might be a better label. I think 'value' might be confusing as to what value is expected to be entered and where it comes from.

module/applications/advancement/hit-points-flow.mjs
```diff
@@ -1,84 +1,105 @@
...
+        max: this.advancement.hitDieValue,
+        modifier: {
+          label: CONFIG.DND5E.abilities[abilityId]?.abbreviation ?? "",
+          value: mod
+        },
+        previous: Object.keys(this.advancement.value).reduce((total, level) => {
+          if ( parseInt(level) === this.level ) return total;
+          return total + Math.max(this.advancement.valueForLevel(parseInt(level)) + mod, 1) + bonus;
+        }, 0),
+        total: hp.max
```
Fyorl: If `value` is left blank, the calculation we display looks incorrect:

Image

`35 + ? + 3 = 35`

module/applications/advancement/hit-points-flow.mjs
```diff
@@ -1,84 +1,105 @@
...
+      hp: {
+        average: this.advancement.average,
+        bonus,
+        max: this.advancement.hitDieValue,
+        modifier: {
+          label: CONFIG.DND5E.abilities[abilityId]?.abbreviation ?? "",
+          value: mod
+        },
+        previous: Object.keys(this.advancement.value).reduce((total, level) => {
+          if ( parseInt(level) === this.level ) return total;
```
Fyorl, line 54: It's possible this advancement could have levels populated beyond the current level. For example, if the user is modifying choices. In these cases the calculation for `previous` will be incorrect.

There is another bug related to this, in that a character's max HP still includes values from HP advancements for levels beyond their current level. Repro steps are:
1. Level a character to 3.
2. Note their max HP.
3. Level them back down to 1.
4. Observe their max HP remains unchanged.

arbron: Hmm, I'm not able to reproduce either of these issues. When you use "Modify Choices", it should be removing all advancement value data for future levels, so none will be displayed. Same with leveling down.

Fyorl: I am investigating further. I think this might be a v14 bug.

Fyorl: It was a v14 bug. No issues here with that bug fixed.

Fyorl (approved):
Found one issue, but good to go once that's resolved. If you advance past the hit points advancement and then go back, clicking the 'take average' checkbox will update the calculation, but the checkbox itself stays checked. You should notice the same if you advance several levels and then attempt to modify choices.

XXDarrienX:
A suggestion to this, an extra button to auto set max hp instead of rolling or taking average?

PR #6249, opened 2025-09-26
[#4708, #6246] Rebuild `ItemChoiceFlow` for V2

### LINKED ISSUE #6246: Item Choice Advancement: When replacing choices, choices that you no longer meet the prereqs of are removed, but newly attained prereqs do not show additional choices
Like the title says, in this instance, Celerity I is required for Celerity II. Selecting to replace Celerity I also removes the option for taking Celerity II (as intended)
Image 

Image

Meanwhile replacing Celerity I and taking say, Iron Durability I does not show Iron Durability II as a possible choice. I would guess that because the advancement isn't fully done, Iron Durability I isnt technically a fulfilled prereq for choosing Iron Durability II.

Image

arbron:
Rework `ItemChoiceFlow` for application V2 on top of the rebuilt `ItemGrantFlowV2`. Uses a more compact style for the item list to accommodate the potentially larger list of items.

Item Choice V2 Tweaks

Added the ability to take a repeatable choice more than once at a single level, previously you could only select a repeatable choice again at a higher level.

Closes #6246

### DIFF
--- a/module/data/item/feat.mjs
+++ b/module/data/item/feat.mjs
@@ -328,12 +328,8 @@ export default class FeatData extends ItemDataModel.mixin(
     const messages = [];
 
     // If a feature has item pre-requisites, make sure the other items exist on the actor
-    const pendingAddition = new Set(added.map(i => i.system.identifier ?? formatIdentifier(i.name)));
-    const pendingRemoval = new Set(removed.map(i => i.system.identifier ?? formatIdentifier(i.name)));
-    const someExist = !this.prerequisites.items.size || Array.from(this.prerequisites.items).some(i => {
-      return (actor.identifiedItems.get(i)?.size || pendingAddition.has(i)) && !pendingRemoval.has(i);
-    });
-    if ( !someExist ) {
+    if ( this.prerequisites.items.size
+      && !Array.from(this.prerequisites.items).some(i => actor.identifiedItems.get(i)?.size) ) {
       messages.push(game.i18n.format("DND5E.Prerequisites.Warning.MissingItem", {
         items: game.i18n.getListFormatter({ type: "disjunction" }).format(Array.from(this.prerequisites.items))
       }));


arbron:
This one is a bear with all the different combinations of selection and replacement. I'd appreciate some additional eyes testing this especially for things like modifying previous levels and what-not.

Zanderaf:
> This one is a bear with all the different combinations of selection and replacement. I'd appreciate some additional eyes testing this especially for things like modifying previous levels and what-not.

I will try my absolute hardest to break this and find any bugs I can once I have time

Zanderaf:
speaking of which, does this solve the issue I found in #6246 where default identifiers were causing issues in regards to the advancement flow?

arbron:
> speaking of which, does this solve the issue I found in #6246 where default identifiers were causing issues in regards to the advancement flow?

It does.

module/data/item/feat.mjs
```diff
@@ -343,8 +339,14 @@ export default class FeatData extends ItemDataModel.mixin(
       }));
     }
 
+    // Check to ensure the item doesn't already exist on actor if it is not repeatable
+    if ( !this.prerequisites.repeatable && actor.sourcedItems?.get(this.parent.uuid)?.size
+      && !added.find(a => a.uuid === this.parent.uuid) ) {
```
Fyorl, line 340: I think we want to use similar logic to the above `someExist` check if I understand correctly. We want to know if the item is not pending removal and either: the item already exists on the actor, or the item is pending addition.

```suggestion
    if ( !this.prerequisites.repeatable && !removed.find(i => i.uuid === this.parent.uuid)
      && (actor.sourcedItems?.get(this.parent.uuid)?.size || added.find(a => a.uuid === this.parent.uuid)) ) {
```

module/documents/advancement/item-choice.mjs
```diff
@@ -72,31 +72,68 @@ export default class ItemChoiceAdvancement extends ItemGrantAdvancement {
...
   storagePath(level) {
     return `value.added.${level}`;
   }
 
   /* -------------------------------------------- */
 
+  /**
+   * @typedef {ItemGrantAdvancementApplicationData} ItemChoiceAdvancementApplicationData
+   * @property {Record<number, Record<string, Item5e>>} previousItems  Copies of items added at earlier levels.
+   * @property {string} [replace]                       Actor ID of the item being replaced.
```
Fyorl:
```suggestion
   * @property {string} [replace]                       ID of the item being replaced.
```

module/documents/advancement/item-choice.mjs
```diff
@@ -109,11 +146,13 @@ export default class ItemChoiceAdvancement extends ItemGrantAdvancement {
...
-  restore(level, data) {
+  async restore(level, data, options={}) {
     const original = this.actor.items.get(data.replaced?.original);
-    if ( data.replaced && !original ) data.items = data.items.filter(i => i._id !== data.replaced.replacement);
+    if ( data.replaced && !original ) Object.entries(data.items).forEach(([uuid, i]) => {
+      if ( i._id !== data.replaced.replacement ) delete data.items[uuid];
+    });
 
-    super.restore(level, data);
+    await super.restore(level, data, options={});
```
Fyorl:
```suggestion
    await super.restore(level, data, options);
```

module/documents/advancement/item-choice.mjs
```diff
@@ -72,31 +72,68 @@ export default class ItemChoiceAdvancement extends ItemGrantAdvancement {
...
-    replacement ??= Object.keys(updates).pop();
-    if ( original && replacement ) {
+    if ( original ) {
+      const replaced = this.value.replaced[level] ?? {};
+      await this._restoreReplacedItem(replaced.level, replaced.original, previousItems ?? retainedData.items);
       const replacedLevel = Object.entries(this.value.added).reverse().reduce((level, [l, v]) => {
         if ( (original in v) && (Number(l) > level) ) return Number(l);
         return level;
       }, 0);
       if ( Number.isFinite(replacedLevel) ) {
```
Fyorl, line 119: Looking at this more closely, is it possible for this to ever be false? If `original` isn't found, then `replacedLevel` is `0`, which is finite, so we still attempt to delete the item and record a replacement. I wonder if it's supposed to be:

```suggestion
      if ( replacedLevel ) {
```

Or is `0` maybe a valid replacement level since feats might have their advancements granted at 'any level', i.e. `0`?

module/documents/advancement/item-choice.mjs
Fyorl: `typeLabel` not defined


```suggestion
      return handleError("DND5E.ADVANCEMENT.ItemChoice.Warning.InvalidType", { type });
```

Fyorl:
```suggestion
      type = game.i18n.localize(CONFIG.Item.typeLabels[restriction.type]);
```

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -49,112 +49,138 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
...
-      }
-    }
+  /**
+   * Source UUIDs of all currently selected items for this level with how many times they have been selected.
+   * @type {Map<string, number>}
+   */
+  get selected() {
+    return Object.values(this.advancement.value.added[this.level] ?? {}).reduce((map, uuid) => {
+      if ( !map.has(uuid) ) map.set(uuid, 1);
+      else map.set(uuid, map.get(uuid) + 1);
```
```suggestion
      map.set(uuid, (map.get(uuid) ?? 0) + 1);
```

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -49,112 +49,138 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
...
+    const dropped = [];
+    const selected = this.selected;
+    for ( const [id, uuid] of Object.entries(value.added[this.level] ?? {}) ) {
+      const item = await fromUuid(uuid);
+      if ( item ) {
+        added.push(item);
+        if ( !this.pool.find(p => p.uuid === uuid) ) dropped.push(item);
+      }
+    }
+    const removed = counts.replaced ? actor.items.get(config.choices[this.level].replacement) : [];
```
```suggestion
    const removed = counts.replaced ? actor.items.get(value.replaced[this.level]?.original) : [];
```

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -49,112 +49,138 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
...
-        };
+      const added = value.added[level];
+      if ( added ) context.sections.set(level, {
+        header: game.i18n.format(`DND5E.AdvancementLevel${level === "0" ? "AnyHeader" : "Header"}`, { level }),
+        items: Object.entries(added).map(([id, uuid]) => {
+          const { name, img } = actor.items.get(id) ?? fromUuidSync(uuid);
+          previouslySelected.add(uuid);
+          return {
+            id, img, name, uuid,
+            checked: id === config.choices[this.level].replacement,
```
```suggestion
            checked: id === value.replaced[this.level]?.original,
```

module/documents/advancement/item-choice.mjs
```diff
@@ -126,28 +165,95 @@ export default class ItemChoiceAdvancement extends ItemGrantAdvancement {
...
+   * @param {object} [options={}]
+   * @param {boolean} [options.showMessage=false]  Show a UI message if the validation fails.
+   */
+  async _evaluatePrerequisites(level, { showMessage=false }={}) {
+    const added = Object.values(this.value.added[level] ?? {}).map(uuid => fromUuidSync(uuid));
+    const replaced = this.actor.items.get(this.value.replaced[level]?.original);
+    const removed = replaced ? [replaced] : [];
+
+    for ( let i = 0; i < 100; i++ ) {
+      const itemsBefore = Object.keys(this.value.added).length;
```
Fyorl:
```suggestion
      const itemsBefore = Object.keys(this.value.added[level] ?? {}).length;
```

module/documents/advancement/item-choice.mjs
```diff
@@ -126,28 +165,95 @@ export default class ItemChoiceAdvancement extends ItemGrantAdvancement {
...
+      for ( const [id, uuid] of Object.entries(this.value.added[level] ?? {}) ) {
+        const item = this.actor.items.get(id);
+        if ( !item ) continue;
+        const isValid = item.system.validatePrerequisites?.(this.actor, {
+          added, removed, level: level || this.actor.system.details?.level, showMessage
+        }) ?? true;
+        if ( isValid !== true ) await this.reverse(level, { skipEvaluation: true, uuid });
+      }
+      if ( itemsBefore === Object.keys(this.value.added).length ) break;
```
Fyorl:
```suggestion
      if ( itemsBefore === Object.keys(this.value.added[level] ?? {}).length ) break;
```

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -231,59 +265,58 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
...
-    event.preventDefault();
-    const uuidToDelete = event.currentTarget.closest("[data-uuid]")?.dataset.uuid;
-    if ( !uuidToDelete ) return;
-    this.dropped.findSplice(i => i.uuid === uuidToDelete);
-    this.selected.delete(uuidToDelete);
-    this._evaluatePrerequisites();
-    this.render();
+  /** @override */
+  async _handleForm(event, form, formData) {
+    const retainedData = { items: {} };
```
Fyorl: Unused

module/documents/advancement/item-choice.mjs
```diff
@@ -109,11 +146,13 @@ export default class ItemChoiceAdvancement extends ItemGrantAdvancement {
   /* -------------------------------------------- */
 
   /** @inheritDoc */
-  restore(level, data) {
+  async restore(level, data, options={}) {
     const original = this.actor.items.get(data.replaced?.original);
-    if ( data.replaced && !original ) data.items = data.items.filter(i => i._id !== data.replaced.replacement);
+    if ( data.replaced && !original ) Object.entries(data.items).forEach(([uuid, i]) => {
```
Fyorl: It seems that if you select no items, then `data.items` doesn't exist at all, and this line throws an error. Possibly the use of `data.items` in the `super` implementation might also potentially throw. Quick repro to see what I mean: Add a Warlock, advance to Invocation selection, advance past it without picking anything, then attempt to go back.

templates/advancement/item-choice-flow.hbs
```diff
@@ -1,79 +1,64 @@
...
-                        data-tooltip aria-label="{{ localize 'DND5E.ItemDelete' }}">
-                    <i class="fa-solid fa-trash" inert></i>
-                </button>
-                <input type="hidden" name="{{ uuid }}" value="true" data-dtype="Boolean">
-                {{else}}
-                <dnd5e-checkbox name="{{ uuid }}" {{ checked checked }} {{ disabled disabled }}></dnd5e-checkbox>
+        <ul class="item-list unlist {{~#if isCurrentLevel}} current-level{{/if}}">
+            {{#each items}}
+            <li class="item flexrow {{~#if replaced}} replaced{{/if}}" data-uuid="{{ uuid }}">
+                <div class="item-name flexrow">
```
Fyorl: Should these be clickable? I think they were before.

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -49,112 +49,138 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
...
+    for ( const [id, uuid] of Object.entries(value.added[this.level] ?? {}) ) {
+      const item = await fromUuid(uuid);
+      if ( item ) {
+        added.push(item);
+        if ( !this.pool.find(p => p.uuid === uuid) ) dropped.push(item);
+      }
+    }
+    const removed = counts.replaced ? actor.items.get(config.choices[this.level].replacement) : [];
+
+    context.sections.set(this.level, {
```
Fyorl: Maybe a slight improvement could be that we don't show the section for selecting a new option if there would otherwise be no choices available if the user opts to not make a replacement. Then it would be a bit clearer that the user is only being presented with this flow in case they want to make a replacement (and it makes the dialog not so long).

module/data/item/feat.mjs
arbron: What issue is this trying to fix? I applied the changes as suggested and it leads to whatever option you select immediately disappearing from the item choice dialog.

arbron:
Fixed the last bug in your list relating to using Modify Choices with repeatable choices. The first three can probably be handled in a separate PR because I don't think we've ever dealt with invalidating prerequisites before, so those aren't regressions.

module/data/item/feat.mjs
Fyorl: No issue I noticed, it just looks wrong. I don't understand the current logic. If the item is not repeatable, and the item already exists on the actor, then it should throw, that part makes sense. But why check for `!added.find(a => a.uuid === this.parent.uuid)`? If the item is pending addition then surely it should throw as well, because then you'll end up with 3 items on the actor (the existing one, the pending one, and this one), no?

Fyorl:
> Fixed the last bug in your list relating to using Modify Choices with repeatable choices. The first three can probably be handled in a separate PR because I don't think we've ever dealt with invalidating prerequisites before, so those aren't regressions.

I am ok with a follow-up PR but (1) is definitely a regression. If I test on `5.3.x` right now, the result is that, as soon as I choose to replace Pact of the Blade, Eldritch Smite is removed from the list of available options, because we call `_evaluatePrerequisites` and then `render` in `ItemChoiceFlow#_onChangeInput` in order to handle such cases.

For (2) I thought we threw a special error which invalidated future advancements and forced them to be re-run, but I can't reproduce that behaviour on `5.3.x` so I am probably misremembering.

arbron:
> > Fixed the last bug in your list relating to using Modify Choices with repeatable choices. The first three can probably be handled in a separate PR because I don't think we've ever dealt with invalidating prerequisites before, so those aren't regressions.
> 
> I am ok with a follow-up PR but (1) is definitely a regression. If I test on `5.3.x` right now, the result is that, as soon as I choose to replace Pact of the Blade, Eldritch Smite is removed from the list of available options, because we call `_evaluatePrerequisites` and then `render` in `ItemChoiceFlow#_onChangeInput` in order to handle such cases.

Aw, looks like because in the `apply` method we aren't applying `actor.reset()` which is needed to re-populate `identifiedItems` after deleting items in this manner.

module/data/item/feat.mjs
arbron: In this case, `added` is the list of items that have been chosen at this level specifically. Those items have already been created on the actor so they appear in `sourcedItems`, but we want to keep them in the list of items so they are available to be unchecked.

Fyorl: I see, thanks. So if this method is now run after any pending additions or removals have already been applied to the actor, then I think some of the logic needs to be adjusted, because it originally assumed that these pending additions or removals hadn't happened yet (thus the 'pending' classifier).

For example, with `someExist`, it should be sufficient to check `identifiedItems` for the prerequisite, and we don't need to check `pendingAddition` or `pendingRemoval` at all.

module/documents/advancement/item-choice.mjs
```diff
@@ -109,45 +141,107 @@ export default class ItemChoiceAdvancement extends ItemGrantAdvancement {
...
+  /**
+   * Restore a previously replaced item that had been removed through the flow process.
+   * @param {number} level       Level of the item that was replaced.
+   * @param {string} originalId  ID of the item that was replaced.
+   * @param {object[]|Record<number, Record<string, Item5e>>} [retainedItems={}]  Original item data.
+   */
+  async _restoreReplacedItem(level, originalId, retainedItems={}) {
+    const uuid = this.value.added[level]?.[originalId];
+    const itemData = (Array.isArray(retainedItems) ? retainedItems.find(i => i._id === originalId)
+      : retainedItems[level]?.[originalId]?.toObject()) ?? await this.createItemData(uuid, originalId);
```
Fyorl, line 234:
```suggestion
    const itemData = ((Array.isArray(retainedItems) ? retainedItems.find(i => i._id === originalId)
      : retainedItems[level]?.[originalId]?.toObject())) ?? await this.createItemData(uuid, originalId);
```

arbron: Why a second set of parens? The current parenthesis are already around the whole ternary statement.

Fyorl: Right you are

PR #6256, opened 2025-09-29
[#4708] Rebuild `AbilityScoreImprovementFlow` for V2

arbron:
Rework `AbilityScoreImprovementFlow` for application V2 while retaining the existing design.

The value data for ASIs has changed so `value.type` no longer defaults to `asi`, it is instead set on the initial application if relevant. This PR updates all of the ASIs in the SRD to get rid of that initial value, but that isn't a required migration.


module/applications/advancement/ability-score-improvement-flow.mjs
```diff
@@ -1,30 +1,69 @@
...
+  /**
+   * Data on points that can be assigned.
+   * @type {{ assigned: number, available: number, cap: number, total: number }}
+   */
+  get points() {
+    const { configuration, value } = this.advancement;
+    const points = {
+      assigned: Object.keys(CONFIG.DND5E.abilities).reduce((assigned, key) => {
+        if ( !this.advancement.canImprove(key) || configuration.locked.has(key) ) return assigned;
+        return assigned + (value.assignments?.[key] ?? 0) - (configuration.fixed[key] ?? 0);
```
Fyorl: The logic here is different, we're taking the fixed points away where we didn't before. I assume it's needed to fix something, but it's not quite working right. You can test yourself by starting with a blank character and adding the legacy High-Elf race to them. When you're prompted for the ASI you can increase something like Dex and note how the fixed Cha points decrease by 1, and the total points remaining stays at 2. If you then decrease Dex again, the points remaining will go to 3.

module/applications/advancement/ability-score-improvement-flow.mjs
```diff
@@ -187,58 +195,69 @@ export default class AbilityScoreImprovementFlow extends AdvancementFlow {
...
-  async _updateObject(event, formData) {
-    await this.advancement.apply(this.level, {
-      type: (this.feat && !this.feat.isASI) ? "feat" : "asi",
-      assignments: this.assignments,
-      featUuid: this.feat?.uuid,
-      retainedItems: this.retainedData?.retainedItems
-    });
+  /** @override */
+  async _handleForm(event, form, formData) {
+    switch (event.target?.name) {
```
Fyorl:
```suggestion
    switch ( event.target?.name ) {
```

module/documents/advancement/ability-score-improvement.mjs
```diff
@@ -168,55 +168,71 @@ export default class AbilityScoreImprovementAdvancement extends Advancement {
   /*  Application Methods                         */
   /* -------------------------------------------- */
 
+  /**
+   * @typedef SubclassAdvancementApplicationData
```
Fyorl:
```suggestion
   * @typedef AbilityScoreImprovementAdvancementApplicationData
```

packs/_source/feats24/epic-boon-feats/boon-of-combat-prowess.yml
```diff
@@ -44,8 +44,7 @@ system:
         locked: []
         points: 1
         max: 30
-      value:
-        type: asi
+      value: {}
```
Fyorl, line 47: These ASIs on feats don't themselves allow you to pick a feat so I would have thought `type: "asi"` was correct, but maybe I am misunderstanding some of these changes. Same with the barbarian's Primal Champion, and the backgrounds.

arbron: The initial `value.type` was removed simply because it doesn't have an effect anymore. Now when the advancement is called with `initial: true` in the options it sets the type to `asi` automatically if `allowFeat` is `false`.

module/applications/advancement/ability-score-improvement-flow.mjs
arbron: Looks like the issue you are seeing is actually down in `_handleForm` relating to how the maximum allowed score was calculated.

This change is to back out the fixed improvements because unlike before, those fixed improvements are already in `value.assignments` (before the fixed improvements weren't applied until after the flow was complete).

There does seem to be another issue here were if you have a score with a fixed improvement that can't be fully applied (e.g. Charisma score of 19 before adding the Half Elf, so it can only get a +1 to charisma before hitting the 20 maximum) then it will add an extra point to the pool.

PR #6260, opened 2025-10-01
[#6259] Add level limits to activity's applied effects

### LINKED ISSUE #6259: Apply level limits to all applied effects on activities
Currently enchantments have level limits to the effects they apply, and activities as a whole are getting level limits with https://github.com/foundryvtt/dnd5e/pull/5892, but it would be good to be able to control the levels at which effects are available for all activities. This would basically entail moving some of the level limits handling from `EnchantActivity` out into the base activity type.

arbron:
Adds the ability for any effect applied by an activity to have a level limit set similar to the way the Enchant Activity currently functions.

Custom activities should get his behavior by default unless they are providing their own `additionalSettings` template for effects (similar to the system's `SaveActivity`). In that case, those activities will need to add the level limits partial to their own effects settings template.

Closes #6259

### DIFF
--- a/module/data/activity/base-activity.mjs
+++ b/module/data/activity/base-activity.mjs
@@ -131,7 +131,10 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
    * @type {ActiveEffect5e[]|null}
    */
   get applicableEffects() {
-    return this.effects?.map(e => e.effect).filter(e => e) ?? null;
+    const level = this.relevantLevel;
+    return this.effects?.filter(e =>
+      e.effect && ((e.level?.min ?? -Infinity) <= level) && (level <= (e.level?.max ?? Infinity))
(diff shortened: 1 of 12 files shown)


module/data/activity/base-activity.mjs
```diff
@@ -162,7 +165,10 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
    * @type {ActiveEffect5e[]|null}
    */
   get applicableEffects() {
```
Fyorl, line 133: Are we already filtering out rider effects somewhere?

Fyorl (approved):

module/data/activity/base-activity.mjs
arbron: This is only looking at the effects defined in the activity, so we do that filter in the application when we provide the list of effects to choose from.

PR #6265, opened 2025-10-02
[#6264] Allow `type:identifier` format in feature prerequisites

### LINKED ISSUE #6264: Accept `type:identifier` format in feature prerequisites field
For the consumption field the system uses the format of `type:identifier` to narrow down the identified item selection to a specific item type. This functionality would also be useful for feature prerequisites to disambiguate potential conflicts.

arbron:
Adds the ability to an `IdentifierField` to accept an item type before the `:`. Only the identifier after the colon will be fully validated.

The `identifiedItems#get` method will now check to see if a colon is included in the provided key and split them up automatically to make using it directly easier.

Closes #6264


arbron:
Leaving this as a draft until the consumption changes from `5.1.10` are rebased onto this branch so some of the no longer needed type splitting code can be removed from it.

Fyorl (approved):

PR #6274, opened 2025-10-05
[#3060] Convert movement fields to formulas, standardize ordering

### LINKED ISSUE #3060: Add parsing for rollData applied to actor `movement` properties
Currently the `system.attributes.movement` object only accepts integer/null values, and can't be hit with rollData via AEs. Let's fix that. :)

arbron:
Converts movement type data into formula fields. Each movement type can refer to resolved speed data from movement types prepared earlier, so the ordering of movement data has been modified to ensure these fields are prepared in a useful order.

Movement Formulas

Movement types are no longer sorted alphabetically because the sort order of the config directly affects the evaluation order, so if we sorted them on localization that could cause effects to no longer properly evaluate.

Closes #3060

### ToDo
- [x] Add `@attributes.movement.speed` & AE redirect to handle that

### DIFF
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -425,7 +425,6 @@ export default class AttributesFields {
     reduction = convertLength(reduction, CONFIG.DND5E.defaultUnits.length.imperial, units);
     const bonus = simplifyBonus(this.attributes.movement.bonus, rollData);
     this.attributes.movement.max = 0;
-    let slowed = false;
     for ( const type of Object.keys(CONFIG.DND5E.movementTypes) ) {
       let speed = Math.max(0, simplifyBonus(this.attributes.movement[type], rollData) - reduction);
       if ( noMovement || (crawl && (type !== "walk")) ) speed = 0;


module/config.mjs
```diff
@@ -2554,6 +2554,9 @@ preLocalize("difficultTerrainTypes", { key: "label", sort: true });
  * @enum {MovementTypeConfig}
  */
 DND5E.movementTypes = {
+  walk: {
```
krbz999, line 2371: If the order matters, should this object perhaps be frozen?

arbron: That would prevent potentially customizing movement types (see https://github.com/foundryvtt/dnd5e/pull/4976).

krbz999: Yeah what I am asking is whether that should be allowed, then. I don't really like the idea of being able to change the order in which the movement attributes are prepared by eg deleting one and adding it back.

roth-michael: Feels like if somebody is going out of their way to break intended functionality there's only so much we can do. As far as potentially accidental messing up of the order, could maybe add a priority/order property to `MovementTypeConfig` that's used whenever order is relevant. Though I don't imagine it would be a particularly common issue; just feels a maybe cleaner when order actually matters to not just rely on which keys were written in first.

module/data/actor/templates/attributes.mjs
Fyorl, line 444: Once this is rebased, we will need to update the logic for detecting slowed creatures. Now that the empty source value is `""` rather than `undefined`, every creature is considered slowed because their base speeds are evaluating to `0` instead of `NaN`.

The rules state only creatures whose Speed is less than half are considered slowed, and the term 'Speed' has replaced the concept of 'walking' speed (see previous comment). So we probably should move the slowed logic out of this loop and check only against walk speed.

module/data/actor/templates/attributes.mjs
```diff
@@ -424,12 +424,10 @@ export default class AttributesFields {
       ? (this.attributes.exhaustion ?? 0) * (CONFIG.DND5E.conditionTypes.exhaustion?.reduction?.speed ?? 0) : 0;
     reduction = convertLength(reduction, CONFIG.DND5E.defaultUnits.length.imperial, units);
     const bonus = simplifyBonus(this.attributes.movement.bonus, rollData);
-    const field = this.schema.getField("attributes.movement");
     this.attributes.movement.max = 0;
     let slowed = false;
```
Fyorl: I think this is no longer used

PR #6285, opened 2025-10-08
[#5373] Convert vehicle sheets to ApplicationV2.

Fyorl:
image

*Ship art by [Epic Isometric](https://www.patreon.com/cw/epicisometric)*

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -189,6 +189,10 @@
       "Label": "Bonus Action"
     },
     "Crew": {
+      "Counted": {
+        "one": "{number} crew action",
+        "other": "{number} crew actions"
+      },
       "Header": "Crew Actions",
       "Label": "Crew Action"
     },
--- a/module/applications/actor/vehicle-sheet.mjs
+++ b/module/applications/actor/vehicle-sheet.mjs
@@ -556,18 +556,6 @@ export default class VehicleActorSheet extends BaseActorSheet {
   }
 
   /* -------------------------------------------- */
-
-  /** @override */
-  _showConfiguration(event, target) {
-    if ( target.dataset.config === "movement" ) {
-      new MovementSensesConfig({
-        document: this.actor, type: "movement", withResolution: true
-      }).render({ force: true });
-      return false;
-    }
-  }
-
-  /* -------------------------------------------- */
   /*  Drag & Drop                                 */
   /* -------------------------------------------- */
 
--- a/module/data/actor/vehicle.mjs
+++ b/module/data/actor/vehicle.mjs
@@ -413,6 +413,20 @@ export default class VehicleData extends CommonTemplate {
       results.actor["system.attributes.actions.spent"] = 0;
     }
   }
+
+  /* -------------------------------------------- */
+  /*  Helpers                                     */
+  /* -------------------------------------------- */
+
+  /**
+   * Whether the given activity should prompt for auto-consumption of a crew action.
+   * @param {Activity} activity  The activity.
+   * @returns {boolean|void}
+   */
+  static canConsumeCrewAction(activity) {
+    const { actor } = activity;
+    return actor?.system.attributes?.actions?.stations === false;
+  }
 }
 
 /* -------------------------------------------- */
--- a/module/applications/activity/activity-usage-dialog.mjs
+++ b/module/applications/activity/activity-usage-dialog.mjs
@@ -243,33 +243,36 @@ export default class ActivityUsageDialog extends Dialog5e {
     context.fields = [];
     context.notes = [];
 
-    const containsLegendaryConsumption = this.activity.consumption.targets
-      .find(t => (t.type === "attribute") && (t.target === "resources.legact.value"));
-    if ( (this.activity.activation.type === "legendary") && this.actor.system.resources?.legact
-      && this._shouldDisplay("consume.action") && !containsLegendaryConsumption ) {
-      const pr = new Intl.PluralRules(game.i18n.lang);
-      const value = (this.config.consume !== false) && (this.config.consume?.action !== false);
-      const warn = (this.actor.system.resources.legact.value < this.activity.activation.value) && value;
-      context.fields.push({
-        field: new BooleanField({
-          label: game.i18n.format("DND5E.CONSUMPTION.Type.Action.Prompt", {
-            type: game.i18n.localize("DND5E.LegendaryAction.Label")
-          }),
-          hint: game.i18n.format("DND5E.CONSUMPTION.Type.Action.PromptHint", {
-            available: game.i18n.format(
-              `DND5E.ACTIVATION.Type.Legendary.Counted.${pr.select(this.actor.system.resources.legact.value)}`,
-              { number: `<strong>${formatNumber(this.actor.system.resources.legact.value)}</strong>` }
-            ),
-            cost: game.i18n.format(
-              `DND5E.ACTIVATION.Type.Legendary.Counted.${pr.select(this.activity.activation.value)}`,
-              { number: `<strong>${formatNumber(this.activity.activation.value)}</strong>` }
-            )
-          })
-        }),
-        input: context.inputs.createCheckboxInput,
-        name: "consume.action",
-        value, warn
+    const activationConfig = CONFIG.DND5E.activityActivationTypes[this.activity.activation.type];
+    if ( activationConfig?.consume && this._shouldDisplay("consume.action") ) {
+      const { property } = activationConfig.consume;
+      const containsConsumption = this.activity.consumption.targets.find(t => {
+        return (t.type === "attribute") && (t.target === `${property}.value`);
       });
+      const current = foundry.utils.getProperty(this.actor.system, property);
+      if ( current && !containsConsumption ) {
+        const plurals = new Intl.PluralRules(game.i18n.lang);
+        const value = (this.config.consume !== false) && (this.config.consume?.action !== false);
+        const warn = (current.value < this.activity.activation.value) && value;
+        context.fields.push({
+          value, warn,
+          field: new BooleanField({
+            label: game.i18n.format("DND5E.CONSUMPTION.Type.Action.Prompt", {
+              type: activationConfig.label
+            }),
+            hint: game.i18n.format("DND5E.CONSUMPTION.Type.Action.PromptHint", {
+              available: game.i18n.format(`${activationConfig.counted}.${plurals.select(current.value)}`, {
+                number: `<strong>${formatNumber(current.value)}</strong>`
+              }),
+              cost: game.i18n.format(`${activationConfig.counted}.${plurals.select(this.activity.activation.value)}`, {
+                number: `<strong>${formatNumber(this.activity.activation.value)}</strong>`
+              })
+            })
+          }),
+          input: context.inputs.createCheckboxInput,
+          name: "consume.action"
+        });
+      }
     }
 
     if ( this.activity.requiresSpellSlot && this.activity.consumption.spellSlot
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -6,6 +6,7 @@ import Actor5e from "./documents/actor/actor.mjs";
 import * as advancement from "./documents/advancement/_module.mjs";
 import { preLocalize } from "./utils.mjs";
 import MappingField from "./data/fields/mapping-field.mjs";
+import VehicleData from "./data/actor/vehicle.mjs";
 
 /**
  * @import { TravelPace5e } from "./data/shared/movement-field.mjs";
@@ -1022,11 +1023,27 @@ preLocalize("abilityActivationTypes");
 
 /**
  * @typedef ActivityActivationTypeConfig
+ * @property {string} [counted]         Localized label for the countable activation type.
  * @property {string} label             Localized label for the activation type.
  * @property {string} [header]          Localized label for the activation type header.
  * @property {string} [group]           Localized label for the presentational group.
  * @property {boolean} [passive=false]  Classify this item as a passive feature on NPC sheets.
  * @property {boolean} [scalar=false]   Does this activation type have a numeric value attached?
+ * @property {ActivityActivationAutoConsumptionConfig} [consume]  Configuration for automatically consuming this
+ *                                                                resource.
+ */
+
+/**
+ * @typedef ActivityActivationAutoConsumptionConfig
+ * @property {ActivityActivationAutoConsumptionPredicate} [canConsume]  A predicate to check if this usage qualifies
+ *                                                                      for auto-consumption.
+ * @property {string} property  The path to the property that is consumed.
+ */
+
+/**
+ * @callback ActivityActivationAutoConsumptionPredicate
+ * @property {Activity} activity  The activity with the consumption.
+ * @returns {boolean|void}        Return explicit false to block auto-consumption.
  */
 
 /**
@@ -1093,6 +1110,10 @@ DND5E.activityActivationTypes = {
     passive: true
   },
   legendary: {
+    counted: "DND5E.ACTIVATION.Type.Legendary.Counted",
+    consume: {
+      property: "resources.legact"
+    },
     label: "DND5E.ACTIVATION.Type.Legendary.Label",
     header: "DND5E.ACTIVATION.Type.Legendary.Header",
     group: "DND5E.ACTIVATION.Category.Monster",
@@ -1110,6 +1131,11 @@ DND5E.activityActivationTypes = {
     group: "DND5E.ACTIVATION.Category.Monster"
   },
   crew: {
+    counted: "DND5E.ACTIVATION.Type.Crew.Counted",
+    consume: {
+      canConsume: VehicleData.canConsumeCrewAction,
+      property: "attributes.actions"
+    },
     label: "DND5E.ACTIVATION.Type.Crew.Label",
     header: "DND5E.ACTIVATION.Type.Crew.Header",
     group: "DND5E.ACTIVATION.Category.Vehicle",
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -485,7 +485,9 @@ export default function ActivityMixin(Base) {
 
       const ignoreLinkedConsumption = this.isSpell && !this.consumption.spellSlot;
       if ( config.consume !== false ) {
-        const hasActionConsumption = this.activation.type === "legendary";
+        const activationConfig = CONFIG.DND5E.activityActivationTypes[this.activation.type] ?? {};
+        const hasActionConsumption = activationConfig.consume
+          && (activationConfig.consume.canConsume?.(this) !== false);
         const hasResourceConsumption = this.consumption.targets.length > 0;
         const hasLinkedConsumption = (linked?.consumption.targets.length > 0) && !ignoreLinkedConsumption;
         const hasSpellSlotConsumption = this.requiresSpellSlot && this.consumption.spellSlot;
@@ -606,25 +608,29 @@ export default function ActivityMixin(Base) {
       if ( config.consume === false ) return updates;
       const errors = [];
 
-      // Handle action economy
-      if ( ((config.consume === true) || config.consume.action) && (this.activation.type === "legendary") ) {
-        const containsLegendaryConsumption = this.consumption.targets
-          .find(t => (t.type === "attribute") && (t.target === "resources.legact.value"));
+      // Handle auto consumption.
+      const activationConfig = CONFIG.DND5E.activityActivationTypes[this.activation.type];
+      if ( ((config.consume === true) || config.consume.action) && activationConfig?.consume ) {
+        const { property } = activationConfig.consume;
+        const valueProperty = `${property}.value`;
+        const containsConsumption = this.consumption.targets.find(t => {
+          return (t.type === "attribute") && (t.target === valueProperty);
+        });
         const count = this.activation.value ?? 1;
-        const legendary = this.actor.system.resources?.legact;
-        if ( legendary && !containsLegendaryConsumption ) {
+        const current = foundry.utils.getProperty(this.actor.system, property);
+        if ( current && !containsConsumption ) {
           let message;
-          if ( legendary.value === 0 ) message = "DND5E.ACTIVATION.Warning.NoActions";
-          else if ( count > legendary.value ) message = "DND5E.ACTIVATION.Warning.NotEnoughActions";
+          if ( current.value < 1 ) message = "DND5E.ACTIVATION.Warning.NoActions";
+          else if ( count > current.value ) message = "DND5E.ACTIVATION.Warning.NotEnoughActions";
           if ( message ) {
             const err = new ConsumptionError(game.i18n.format(message, {
-              type: game.i18n.localize("DND5E.LegendaryAction.Label"),
+              type: activationConfig.label,
               required: formatNumber(count),
-              available: formatNumber(legendary.value)
+              available: formatNumber(current.value)
             }));
             errors.push(err);
           } else {
-            updates.actor["system.resources.legact.spent"] = legendary.spent + count;
+            updates.actor[`system.${property}.spent`] = current.spent + count;
           }
         }
       }
--- a/templates/inventory/encumbrance.hbs
+++ b/templates/inventory/encumbrance.hbs
@@ -8,7 +8,7 @@
         {{else}}
         <i class="fa-solid fa-weight-hanging" inert></i>
         {{/if}}
-        <span class="value">{{ value }}</span>
+        <span class="value">{{ dnd5e-numberFormat value }}</span>
         <span class="separator">&sol;</span>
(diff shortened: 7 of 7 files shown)


module/applications/actor/vehicle-sheet.mjs
```diff
@@ -1,326 +1,656 @@
...
+    const subtitles = [];
+    if ( COVER[cover] ) subtitles.push(game.i18n.localize(COVER[cover]));
+    if ( armor?.value ) subtitles.push(`${game.i18n.localize("DND5E.AC")} ${armor.value}`);
+    if ( hp?.dt ) subtitles.push(`${game.i18n.localize("DND5E.HITPOINTS.DT.abbr")} ${hp.dt}`);
+    if ( subtitles.length ) ctx.subtitle = subtitles.join(" • ");
+    if ( item.type === "weapon" ) {
+      const enrichmentOptions = { secrets: item.isOwner, relativeTo: item, rollData: item.getRollData() };
+      ctx.enriched = {
+        attack: await CONFIG.ux.TextEditor.enrichHTML("[[/attack extended]]", enrichmentOptions),
+        damage: await CONFIG.ux.TextEditor.enrichHTML("[[/damage extended]]", enrichmentOptions)
```
krbz999: Overriding weapon descriptions on the sheet?

module/data/actor/vehicle.mjs
```diff
@@ -178,16 +281,84 @@ export default class VehicleData extends CommonTemplate {
 
     this.prepareAbilities({ rollData, originalSaves });
     AttributesFields.prepareArmorClass.call(this, rollData);
-    AttributesFields.prepareEncumbrance.call(this, rollData, { validateItem: item =>
-      (item.flags.dnd5e?.vehicleCargo === true) || !["weapon", "equipment"].includes(item.type)
-    });
+    if ( this.attributes.ac.value ) {
+      this.attributes.ac.motionless = this.attributes.ac.value - Math.max(0, this.abilities.dex.mod);
```
krbz999: `abilities.dex` might not exist.

templates/actors/vehicle/sidebar.hbs
```diff
@@ -0,0 +1,178 @@
...
+    <label class="slide-toggle header-interactable">
+        <input type="checkbox" name="flags.dnd5e.showVehicleAbilities" {{ checked showAbilities }}>
+        {{ localize "DND5E.VEHICLE.ShowVehicleAbilities" }}
+        <i class="fa-solid fa-toggle-{{ ifThen showAbilities "on" "off" }}" inert></i>
+    </label>
+    {{/if}}
+
+    <div class="portrait">
+        <img src="{{ document.img }}" alt="{{ document.name }}"
+             {{#if editable}}data-action="editImage" data-edit="img"{{/if}}>
```
arbron:
```suggestion
        <img src="{{ document.img }}" alt="{{ document.name }}"
             {{#if editable}}data-action="editImage" data-edit="img"{{else}}data-action="showArtwork"{{/if}}>
```

Should probably also support toggling to display the token.

templates/actors/vehicle/sidebar.hbs
```diff
@@ -0,0 +1,178 @@
+<aside class="sheet-sidebar">
+
+    {{!-- Portrait --}}
+    {{#if editable}}
+    <label class="slide-toggle header-interactable">
+        <input type="checkbox" name="flags.dnd5e.showVehicleAbilities" {{ checked showAbilities }}>
+        {{ localize "DND5E.VEHICLE.ShowVehicleAbilities" }}
+        <i class="fa-solid fa-toggle-{{ ifThen showAbilities "on" "off" }}" inert></i>
+    </label>
+    {{/if}}
```
arbron: If we add support for toggling to show the token, perhaps this could be moved over a column to be above the abilities.

templates/actors/vehicle/sidebar.hbs
```diff
@@ -0,0 +1,178 @@
...
+    {{!-- Armor Class --}}
+    {{#if (or editable system.attributes.ac.value)}}
+    {{> "systems/dnd5e/templates/actors/parts/actor-trait-line.hbs" icon="fa-solid fa-shield" label="DND5E.ArmorClass"
+        configAction="armorClass" configLabel="DND5E.ArmorConfig" data=system.attributes.ac }}
+    {{/if}}
+
+    {{!-- Speed --}}
+    {{> "systems/dnd5e/templates/actors/parts/actor-trait-line.hbs" icon="fa-solid fa-shoe-prints"
+        label="DND5E.Speed" configAction="movement" configLabel="DND5E.MovementConfig"
+        value=system.attributes.movement.max }}
```
arbron: It would be good to include the unit here in some way to distinguish between `ft/round` and `mi/day`.

templates/actors/vehicle/stations.hbs
```diff
@@ -0,0 +1,210 @@
...
+        {{#dnd5e-itemContext item as |ctx|}}
+        <li class="station draggable" data-item-id="{{ item.id }}" data-uuid="{{ item.uuid }}">
+
+            {{!-- Header --}}
+            <div class="station-header">
+
+                {{!-- Name, AC, & DT --}}
+                <div class="name-stacked item-tooltip {{ @root.rollableClass }}" data-action="useItem">
+                    <span class="title">{{ item.name }}</span>
+                    <span class="subtitle">{{ ctx.subtitle }}</span>
```
arbron, line 67: Subtitle can contain HTML:

Screenshot 2025-10-08 at 14 08 07 2

```suggestion
                    <span class="subtitle">{{{ ctx.subtitle }}}</span>
```

templates/actors/vehicle/stations.hbs
```diff
@@ -0,0 +1,210 @@
...
+                        <i class="fa-solid fa-circle-xmark"></i>
+                    </a>
+                    {{/if}}
+                </div>
+                {{/each}}
+            </div>
+            {{/if}}
+
+        </li>
+        {{/dnd5e-itemContext}}
```
arbron, line 157: For items with limited uses it would be good to represent them here somewhere.

templates/actors/vehicle/stations.hbs
```diff
@@ -0,0 +1,210 @@
+<section class="sheet-stations" data-area="draft">
+
+    {{!-- Abilities --}}
+    {{#if showAbilities}}
+    <div class="abilities">
+        {{> ".header" }}
+        {{> ".header" }}
+        {{#each abilities}}
+        {{> ".ability" path=(concat "system.abilities." key) }}
+        {{/each}}
```
arbron, line 10: It would be nice to place these column-first, so the three physical stats go down one column and the mental stats go down the other.

templates/actors/vehicle/stations.hbs
```diff
@@ -0,0 +1,210 @@
...
+
+            </div>
+
+            {{!-- Controls --}}
+            {{#if @root.editable}}
+            <div class="controls">
+                <button type="button" class="config-button unbutton" data-action="removeMember" data-tooltip
+                        aria-label="{{ localize "DND5E.Group.Action.Remove" }}">
+                    <i class="fa-solid fa-square-xmark" inert></i>
+                </button>
```
arbron, line 45: The delete button feels a bit odd in its current position. Could it be moved over to the corner of the card? I also think most of the other delete buttons like this have rounded corners.

Screenshot 2025-10-08 at 14 16 28

templates/actors/vehicle/crew.hbs
```diff
@@ -0,0 +1,81 @@
...
+            <span class="quantity">{{ passengers.total }}</span>
+        </header>
+        <ol class="unlist">
+            {{#each passengers.value}}
+            {{> ".crew" draggable=true adjustable=(not @root.locked) }}
+            {{/each}}
+        </ol>
+    </section>
+    {{else if system.passengers.max}}
+    <div class="drop-zone drop-zone-lg roboto-upper passengers" data-area="passengers">
```
arbron: Elsewhere in the code we've used `drop-area` instead of `drop-zone`. Perhaps some style consolidation could also take place with those.

```css
    .drop-area {
      border: 1px dashed var(--color-border-light-2);
      border-radius: 4px;
      padding: 6px;
    }
```

lang/en.json
```diff
@@ -4873,14 +4881,83 @@
 "DND5E.Value": "Value",
 
 "DND5E.VEHICLE": {
+  "Actions": {
```
arbron: We generally use a singular "Action" here:

```suggestion
  "Action": {
```

module/applications/actor/_module.mjs
```diff
@@ -1,8 +1,8 @@
 export {default as CharacterActorSheet} from "./character-sheet.mjs";
-export {default as ActorSheet5eVehicle} from "./vehicle-sheet.mjs";
 export {default as EncounterActorSheet} from "./encounter-sheet.mjs";
 export {default as GroupActorSheet} from "./group-sheet.mjs";
 export {default as NPCActorSheet} from "./npc-sheet.mjs";
+export {default as VehicleActorSheet} from "./vehicle-sheet.mjs";
 export {default as TransformDialog} from "./transform-dialog.mjs";
```
arbron:
```suggestion
export {default as TransformDialog} from "./transform-dialog.mjs";
export {default as VehicleActorSheet} from "./vehicle-sheet.mjs";
```

module/applications/api/primary-sheet-mixin.mjs
```diff
@@ -274,6 +274,7 @@ export default function PrimarySheetMixin(Base) {
     /** @inheritDoc */
     async _onRender(context, options) {
       await super._onRender(context, options);
+      this._renderSource();
```
arbron, line 277: This means we can probably pull this out of NPC sheet to avoid doing it twice.

module/applications/shared/movement-senses-config.mjs
```diff
@@ -84,7 +85,16 @@ export default class MovementSensesConfig extends BaseConfigSheet {
...
     }));
 
-    context.unitsOptions = Object.entries(CONFIG.DND5E.movementUnits).map(([value, { label }]) => ({ value, label }));
+    context.unitsOptions = Object.entries(CONFIG.DND5E.movementUnits).map(([value, { label, travelResolution }]) => {
+      if ( !this.options.withResolution || !travelResolution ) return { value, label };
+      return {
+        value,
+        label: game.i18n.format("DND5E.UNITS.TRAVEL.UnitLabel", {
+          unit: label,
+          resolution: game.i18n.format(`DND5E.UNITS.TRAVEL.${travelResolution.capitalize()}`)
```
arbron, line 94: One issue with this is that the PHB vehicles all define their speeds in mph rather than m/day.

less/v2/vehicle.less
```diff
@@ -0,0 +1,510 @@
...
+    &.active { display: grid; }
+  }
+
+  .sheet-crew {
+    padding-top: 8px;
+    container: crew / inline-size;
+    grid-template: "a a" min-content "c c" min-content;
+
+    &.has-assigned {
+      grid-template: "a b" min-content "c c" min-content;
```
arbron: Feels a bit odd to have mismatched columns here:

Screenshot 2025-10-08 at 14 43 47

```suggestion
      grid-template: "a b" min-content "c c" min-content;
      grid-template-columns: 1fr 1fr;
```

module/applications/actor/vehicle-sheet.mjs
arbron: This can lead to non-rendered enrichers or invalid values if we just add it. This is what I see for a newly created weapon:

Screenshot 2025-10-08 at 14 08 07

The `NaN` range should be fixed on the enricher side, but we should also check to ensure there is actually damage defined for the item before adding the damage enricher.

module/data/actor/vehicle.mjs
```diff
@@ -178,16 +281,84 @@ export default class VehicleData extends CommonTemplate {
...
+   * Compute the update required in order to adjust the crew by some delta.
+   * @param {string} area          The crew area.
+   * @param {string} uuid          The crew member's UUID.
+   * @param {string|number} delta  The delta.
+   * @returns {object}
+   */
+  getCrewUpdates(area, uuid, delta) {
+    const roster = this[area].value;
+    const quantity = roster.reduce((acc, u) => acc + (u === uuid), 0);
+    const target = Math.max(0, typeof delta === "number" ? quantity + delta : parseDelta(delta, quantity));
```
arbron: Bug with adjusting quantities when a crew member has some assigned and some unassigned. Since the quantity in the unassigned list only shows the unassigned count and not the total count, any adjustments there result in the final crew number being off by however many assigned crew of that type there are.

So if you have 2 sailors in the crew, one assigned to a station and the other not, and then you decrease the count on the unassigned side by 1 it will completely remove the sailor from the vehicle (though they will still appear listed at the station at which they were assigned, just not in the Crew tab any longer).

module/applications/actor/vehicle-sheet.mjs
```diff
@@ -1,326 +1,656 @@
-import ActorSheet5e from "./deprecated/base-sheet.mjs";
+import { formatCR, formatNumber } from "../../utils.mjs";
+import CompendiumBrowser from "../compendium-browser.mjs";
+import ContextMenu5e from "../context-menu.mjs";
+import MovementSensesConfig from "../shared/movement-senses-config.mjs";
+import BaseActorSheet from "./api/base-actor-sheet.mjs";
```
arbron, line 6:
```suggestion

const TextEditor = foundry.applications.ux.TextEditor.implementation;

```

module/applications/actor/vehicle-sheet.mjs
```diff
@@ -1,326 +1,656 @@
...
+   */
+  async _prepareStationsContext(context, options) {
+    const Inventory = customElements.get(this.options.elements.inventory);
+    const columns = Inventory.mapColumns(["uses", "controls"]);
+    const sections = [{
+      columns, id: "features", label: "DND5E.Features", order: 100, minWidth: 170,
+      items: context.itemCategories.features
+    }];
+    (context.itemCategories.stations || []).sort((a, b) => a.sort - b.sort);
+    (context.itemCategories.features || []).sort((a, b) => a.sort - b.sort);
```
arbron: Feels a bit cleaner since we aren't doing any assignment here:

```suggestion
    context.itemCategories.stations?.sort((a, b) => a.sort - b.sort);
    context.itemCategories.features?.sort((a, b) => a.sort - b.sort);
```

module/applications/actor/vehicle-sheet.mjs
```diff
@@ -1,326 +1,656 @@
...
+   * @param {object} context  Shared context provided by _prepareContext.
+   * @param {object} options  Application render options.
+   * @returns {Promise<object>}
+   * @protected
+   */
+  async _prepareCrewContext(context, options) {
+    const { groupCrew, resolveCrew } = this.constructor;
+    if ( context.system.crew.value.length ) {
+      const crew = groupCrew(context.system.crew.value);
+      const assigned = context.itemCategories.stations.flatMap(i => i.system.crew.value).reduce((obj, uuid) => {
```
arbron: Causes the sheet to fail to render if crew is present but no stations:

```suggestion
      const assigned = (context.itemCategories.stations ?? []).flatMap(i => i.system.crew.value).reduce((obj, uuid) => {
```

module/applications/actor/vehicle-sheet.mjs
```diff
@@ -1,326 +1,656 @@
...
+    const unitSystem = game.settings.get("dnd5e", "metricWeightUnits") ? "metric" : "imperial";
+    const units = baseUnits.default[unitSystem];
+    return Promise.all(this.actor.system.draft.value.map(async uuid => {
+      const actor = await fromUuid(uuid);
+      const { system } = actor;
+      const capacity = (actor.system.attributes?.encumbrance?.max || 0) * draftMultiplier;
+      const subtitle = [
+        CONFIG.DND5E.actorSizes[system.traits?.size]?.label,
+        system.details?.type?.label,
+        capacity ? `${formatNumber(capacity)} ${units}` : null
```
arbron:
```suggestion
        capacity ? formatWeight(capacity, units) : null
```

module/applications/actor/vehicle-sheet.mjs
```diff
@@ -1,326 +1,656 @@
...
+   */
+  static #onUncrew(event, target) {
+    const { uuid } = target.closest("[data-uuid]")?.dataset ?? {};
+    const { itemId } = target.closest("[data-item-id]")?.dataset ?? {};
+    const item = this.actor.items.get(itemId);
+    if ( !item ) return;
+    const crew = [...item.system.crew.value];
+    crew.findSplice(u => u === uuid);
+    item.update({ "system.crew.value": crew });
+
```
arbron:
```suggestion
```

module/data/actor/vehicle.mjs
```diff
@@ -65,12 +74,11 @@ export default class VehicleData extends CommonTemplate {
   /** @inheritDoc */
   static defineSchema() {
     return this.mergeSchema(super.defineSchema(), {
-      vehicleType: new StringField({ required: true, initial: "water", label: "DND5E.VehicleType" }),
       attributes: new SchemaField({
         ...AttributesFields.common,
         ac: new SchemaField({
           ...AttributesFields.armorClass,
-          motionless: new StringField({ required: true, label: "DND5E.ArmorClassMotionless" })
+          calc: new StringField({ initial: "flat", label: "DND5E.ArmorClassCalculation" })
```
arbron, line 82: Do we want to get rid of motionless AC? That is still required for Infernal War Machines and having it a separate number will help with stat block creation if we do that.

module/applications/actor/vehicle-sheet.mjs
krbz999: and that &bull.

module/data/actor/vehicle.mjs
Fyorl: It still exists, it's just derived now.

module/applications/shared/movement-senses-config.mjs
Fyorl: Group travel pace uses miles-per-day, and a group using a vehicle should inherit the vehicle's move speed. I could modify the movement config to show miles-per-day if spawned on the group sheet, and miles-per-hour if spawned on the vehicle sheet, but the underlying unit we store would be `mi` in both cases, which seems wrong.

Converting from mph to mpd requires some assumptions too. Can a Galley maintain its speed for 24h per day? Or does it require enough crew to man the oars in 3 8-hour shifts to do that?

Allowing the user to enter the mpd value directly seemed like the most straightforward solution, but I am open to either improvements to our units system, and/or adding some extra data to groups, like hours travelled.

module/applications/actor/vehicle-sheet.mjs
Fyorl: > Overriding weapon descriptions on the sheet?

Not sure I understand the question, but I don't think so.

Fyorl: > but we should also check to ensure there is actually damage defined for the item before adding the damage enricher.

It would be pretty strange for a weapon to not have any damage defined for it. I don't think this is a real problem, but I can add a guard I suppose.

templates/actors/vehicle/sidebar.hbs
Fyorl: I will add it if there's a demand. It doesn't seem like a useful feature for vehicles.

templates/actors/vehicle/stations.hbs
Fyorl: I don't mind, but this ordering matches that of the character sheet saving throws so whichever we choose, it should be consistent.

templates/actors/vehicle/stations.hbs
Fyorl: This is the exact same positioning as on the group & encounter sheets. It also does have rounded corners.

module/data/actor/vehicle.mjs
Fyorl: Good catch

Fyorl:
> Design is looking good!
> 
> For legacy vehicles with Action Thresholds, we probably want to display that somewhere (maybe a toggle making use of the existing `attributes.actions.stations` boolean).

I addressed most of these. Notably, I have added a uses bar to stations that require it. I also added options for many of the legacy naval rules, such as the quality score and vehicle initiative.

The vehicle actions implementation should now be more complete. This involved migrating `attributes.actions.value` (apparently this represented maximum actions, which is quite unidiomatic) to `attributes.actions.max`, and I have added `attributes.actions.spent`. I added pips for vehicle actions, which replenish automatically every turn, and the ability to target these actions for consumption.

I also added a banner for the dark mode version.

image

*Ship art by [Epic Isometric](https://www.patreon.com/cw/epicisometric)*

templates/actors/vehicle/stations.hbs
arbron: I guess it just looks a bit different when the picture is on that same side, so there is less room around the remove button. As far as I can tell this is the only spot where the art is on the right instead of the left.

Perhaps swapping that would work, which would also potentially open up the right side in case we wanted to add quantity to draft animals which would may be useful (and match a bit more with how crew are set up).

module/applications/shared/movement-senses-config.mjs
arbron: Looking at some airship stat blocks, those list speed like this: `Speed: 10 mph (Fly Speed 100 ft. [hover])`, so what we may need to do is do what I did over on Black Flag and split up combat movement speed and travel pace into two separate entries.

Over there I implemented a new set of "Travel Pace" units that could then be represented as miles/hour or miles/day. Black Flag has a set conversion formula for calculating daily travel pace, I'm not sure dnd5e has one. If not, we'll have to settle on a distance-per-day calculation (if we implement it in the vehicles, then we bake that calculation in at the data level, changing what the books say, while if we implement it on the group sheets then we have the flexibility to tweak it as needed and aren't changing the stat blocks in officially published content).

If you want to leave it with just the combat speed in this PR, I can work on the travel pace similar to what I have over on Black Flag as a follow up so we don't have to settle on something right this second.

arbron:
My only remaining concern beyond the comments above about draft animals and travel pace is the lack of an Effects tab. Equipment on the sheet can still have effects that modify the vehicle, and activities can still apply effects to it, but there is no way to view any of these effects.

module/applications/shared/movement-senses-config.mjs
Fyorl: Having a separate travel pace works for me. D&D does have a calculation for travel pace:

image

As you can see though, it is a function of the number of hours travelled in a day. For groups on foot, or something like a wagon being pulled by horses, then it's fine to assume 8 hours, but that's probably not a reasonable assumption for a vehicle.

templates/actors/vehicle/stations.hbs
Fyorl: I can move it to the left

Fyorl: I take it back, I don't like it.

Fyorl:
> My only remaining concern beyond the comments above about draft animals and travel pace is the lack of an Effects tab. Equipment on the sheet can still have effects that modify the vehicle, and activities can still apply effects to it, but there is no way to view any of these effects.

I added an effects tab. Also noticed some IWMs have a reaction block, so I've added some partitioning for bonus actions and reactions to the features sections.

module/applications/actor/vehicle-sheet.mjs
```diff
@@ -259,13 +275,21 @@ export default class VehicleActorSheet extends BaseActorSheet {
...
-    const sections = [{
-      columns, id: "features", label: "DND5E.Features", order: 100, minWidth: 170,
-      items: context.itemCategories.features
-    }];
+    const sections = Object.fromEntries(["features", "bonus", "reaction"].map((id, i) => {
+      const { header } = CONFIG.DND5E.activityActivationTypes[id] ?? {};
+      const label = id === "features" ? "DND5E.Features" : header;
+      return [id, { columns, id, label, order: (i + 1) * 100, items: [], minWidth: 170 }];
+    }));
+    context.itemCategories.features?.forEach(i => {
```
arbron, line 283: Maybe hide sections that don't have any items.

Fyorl: I was hiding them with CSS. Is it not working?

arbron: Oh yeah, I made this comment before noticing the CSS hadn't properly reloaded. Checked again and it is working fine.

Fyorl:
> One final thought I had was about the vehicle actions. Since we have the "Crew Action" activation type, maybe we could hook it up like we do with "Legendary Action" activation and have it consume vehicle actions automatically (provided "Action Stations" isn't set) rather than requiring consumption.

Good idea. I have generalised our auto-consumption logic now and added crew actions to it. Worth reviewing just those changes to activity consumption if you can.

I also removed the travel resolution from the vehicle movement config per our earlier discussion so speed is now feet-per-round or miles-per-hour, and we can add a separate travel speed units later.

PR #6290, opened 2025-10-09
Fix issues with negative consumption preview when using dice

arbron:
Applies a few changes to how the preview for activity recovery that includes die rolls. Previously if the first term was negative and a die then the formula would display `(1d4 * -1) - 1` and it would not scale properly using automatic mode. This adds `0 +` to the start of formulas before scaling them to ensure the dice are not multiplied by `-1`.

**Before:**
Negative Consumption Before

**After:**
Negative Consumption After

This also includes a new `BasicRoll#invert` method that takes the roll's formula and inverts the operators so that the total from the new formula will be the old total times negative one. This allows the system to properly display non-deterministic recovery formulas.

### DIFF
--- a/module/dice/basic-roll.mjs
+++ b/module/dice/basic-roll.mjs
@@ -357,11 +357,13 @@ export default class BasicRoll extends Roll {
    * @returns {BasicRoll}
    */
   invert() {
-    // Add or remove "0 +" from the start of formulas that don't being with a numeric term
+    // Add "0 +" to the start of formulas that don't begin with a numeric term
     if ( !(this.terms[0] instanceof foundry.dice.terms.NumericTerm) ) this.terms.unshift(
       new foundry.dice.terms.NumericTerm({ number: 0 }),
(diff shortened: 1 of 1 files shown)


module/dice/basic-roll.mjs
```diff
@@ -347,6 +347,40 @@ export default class BasicRoll extends Roll {
...
+  /*  Inversion Methods                           */
+  /* -------------------------------------------- */
+
+  /**
+   * Invert the roll's formula, resulting in a formula whose total, if multiplied by -1, will be the same as
+   * the evaluated total of the original formula.
+   * @returns {BasicRoll}
+   */
+  invert() {
+    // Add or remove "0 +" from the start of formulas that don't being with a numeric term
```
Fyorl:
```suggestion
    // Add or remove "0 +" from the start of formulas that don't begin with a numeric term
```

Also, doesn't seem to match what is actually occurring, which is that we are removing `0 -` in the second branch.

Fyorl (approved):

PR #6294, opened 2025-10-10
[#4570] Allow consumed hit dice to be used in roll formulas

### LINKED ISSUE #4570: Store data of consumed properties when an activity consumes uses or hit dice
Saving this in flags, or in the message, or somewhere else, would allow for an item to consume a hit die, and then also roll said hit dice properly. Currently this is not technically possible, since consuming the "largestAvailable" hit die might result in what is the largest available now being something different, making the subsequent roll incorrect.

Hit dice is, I think, a very special case here, where limited uses might just be an integer, it could be a more complicated formula (eg if you consume 6 hit dice, which could be something like "2d12 + 2d8 + 2d6").

Related issue: https://github.com/foundryvtt/foundryvtt-premium-content/issues/718

arbron:
When clicking on a chat card action or triggering subsequent actions a flag will be set on the item that contains a formatted version of the activity's consumption. Currently this only includes a formula for the hit dice consumed at `@consumed.hd`, allowing any hit dice consumed to be rolled.

Closes #4570

### DIFF
--- a/module/documents/activity/mixin.mjs
+++ b/module/documents/activity/mixin.mjs
@@ -8,6 +8,7 @@ import PseudoDocumentMixin from "../mixins/pseudo-document.mjs";
 
 /**
  * @import { FavoriteData5e } from "../../data/actor/_types.mjs";
+ * @import { ActorDeltasData } from "../../data/chat-message/fields/_types.mjs";
  * @import {
  *   BasicRollDialogConfiguration, BasicRollMessageConfiguration, DamageRollProcessConfiguration
  * } from "../../dice/_types.mjs";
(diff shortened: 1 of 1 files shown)


Zanderaf:
Does this allow for multiple different types of consumed Hit Dice to be rolled at once? Like 2d10 Fighter Hit Dice + 1d6 Wizard hit dice being 2d10 + 1d6?

krbz999:
> Does this allow for multiple different types of consumed Hit Dice to be rolled at once? Like 2d10 Fighter Hit Dice + 1d6 Wizard hit dice being 2d10 + 1d6?

Looks like it stores the formula. So the stored property would be `"2d10 + 1d6"` in that case.

PR #6297, opened 2025-10-11
[#6295] Add separate travel speeds to vehicles & party

### LINKED ISSUE #6295: Add separate travel pace fields to vehicle sheet, travel pace unit types

arbron:
Adds a new `attributes.travel` entry to vehicles to represent speeds in miles/kilometers per hour and store automatically calculated travel pace (miles/kilometers per day). Vehicles can contain both travel speeds and combat speeds, while the party has lost combat speeds and only has travel speeds.

Movement Details

Existing movement speeds in parties are migrated to travel speeds based on the original unit type (units with per-round resolution are divided by ten while those with per-day resolution are divided by eight). Speeds on parties are now only set in per-hour values which are automatically converted to per-day values based on the travel pace.

Movement Config - Vehicle
Movement Config - Party

Currently everything is hard-coded as having 8 hours of travel per day, but in the future it would be useful to allow that to be customized if a specific vehicle can move all day long.

Closes #6295

### DIFF
--- a/module/data/actor/fields/travel-field.mjs
+++ b/module/data/actor/fields/travel-field.mjs
@@ -70,23 +70,25 @@ export default class TravelField extends foundry.data.fields.SchemaField {
     for ( const type of Object.keys(travel.speeds) ) {
       let pace = travel.paces[type] = Math.max(0, simplifyBonus(travel.paces[type], rollData)) * multiplier;
       const speed = travel.speeds[type] = Math.max(0, simplifyBonus(travel.speeds[type], rollData)) * multiplier;
-      if ( pace > travel.paces.max ) travel.paces.max = pace;
-      if ( speed > travel.speeds.max ) travel.speeds.max = speed;
       if ( speed && !pace ) pace = travel.paces[type] = speed * TravelField.#HOURS_PER_DAY;
       if ( pace && !speed ) travel.speeds[type] = Math.floor(pace / TravelField.#HOURS_PER_DAY);
       if ( pace && paceMode ) travel.paces[type] = TravelField.applyPaceMultiplier(
         pace, paceMode, CONFIG.DND5E.travelUnits[units]?.type
       );
+      if ( pace > travel.paces.max ) travel.paces.max = pace;
+      if ( speed > travel.speeds.max ) travel.speeds.max = speed;
     }
 
     if ( !movement ) return;
     for ( const [type, { travel: travelType="land" }] of Object.entries(CONFIG.DND5E.movementTypes) ) {
       if ( !movement[type] ) continue;
       const speed = TravelField.convertMovementToTravel(movement[type], movement.units, units);
-      travel.paces[travelType] ||= TravelField.applyPaceMultiplier(
+      const pace = travel.paces[travelType] ||= TravelField.applyPaceMultiplier(
         speed * TravelField.#HOURS_PER_DAY, paceMode, CONFIG.DND5E.movementUnits[movement.units]?.type
       );
       travel.speeds[travelType] ||= speed;
+      if ( pace > travel.paces.max ) travel.paces.max = pace;
+      if ( speed > travel.speeds.max ) travel.speeds.max = speed;
     }
   }
 


module/applications/shared/movement-senses-config.mjs
```diff
@@ -77,34 +75,29 @@ export default class MovementSensesConfig extends BaseConfigSheet {
...
+        const automaticUnit = CONFIG.DND5E.movementUnits[placeholderData?.units ?? defaultUnits("length")]?.label ?? "";
+        context.unitsOptions.blank = true;
+        context.unitsOptions.unshift(
+          { value: "", label: game.i18n.format("DND5E.AutomaticValue", { value: automaticUnit.toLowerCase() }) },
+          { rule: true }
+        );
+      }
     }
 
+    if ( this.keyPath.endsWith("movement") ) this._prepareTravelFields(context);
```
Fyorl:
```suggestion
    if ( this.options.type === "movement" ) this._prepareTravelFields(context);
```

module/data/actor/fields/travel-field.mjs
```diff
@@ -0,0 +1,133 @@
...
+ * @property {Record<string, string>} speeds  Formulas for various travel speeds.
+ * @property {string} units                   Movement used to measure the various travel speeds.
+ */
+
+/**
+ * Field for storing travel data.
+ */
+export default class TravelField extends foundry.data.fields.SchemaField {
+  constructor(fields={}, { initialUnits=null, ...options }={}) {
+    const numberConfig = { required: true, nullable: true, min: 0, step: 0.1, initial: null };
```
Fyorl: Not used

module/data/actor/fields/travel-field.mjs
```diff
@@ -0,0 +1,133 @@
...
+    super(fields, options);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Number of hours of travel per day.
+   * @type {number}
+   */
+  static #HOURS_PER_DAY = 8; // TODO: Allow this to be configured
```
Fyorl, line 50: Doesn't have to be in this PR, but I don't think we should release 5.2 without this addressed. For groups it's fine, but for vehicles it's probably going to be incorrect about as often as it's correct. Adding an 'hours per day' field (default 24 for vehicles, 8 for groups) to this `TravelField` seems like a pretty straightforward solution.

module/data/actor/fields/travel-field.mjs
```diff
@@ -0,0 +1,133 @@
...
+      else if ( halfMovement ) speed *= 0.5;
+      travel.speeds[type] = speed;
+      if ( speed > travel.max ) travel.max = speed;
+      if ( speed ) travel.paces[type] = TravelField.convertSpeedToPace(speed, pace);
+    }
+
+    if ( !movement ) return;
+    for ( const [type, { travel: travelType="land" }] of Object.entries(CONFIG.DND5E.movementTypes) ) {
+      if ( !movement[type] || travel.speeds[travelType] ) continue;
+      const speed = TravelField.convertMovementToTravel(movement[type], movement.units, units);
```
Fyorl, line 85: We can write the derived speed to the model. This would allow us to then show these values as placeholders in the movement config. It will give users a bit of a hint that they don't necessarily need to fill out those fields unless their stat block deviates from the auto-calculated values.

```suggestion
      const speed = TravelField.convertMovementToTravel(movement[type], movement.units, units);
      travel.speeds[travelType] = speed;
```

module/data/actor/fields/travel-field.mjs
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Convert combat movement speed in feet or meters to travel speed in mph or km/h.
+   * @param {number} value
+   * @param {string} initialUnit
+   * @param {string} finalUnit
+   * @returns {number}
+   */
+  static convertMovementToTravel(value, initialUnit, finalUnit) {
+    value = convertLength(value, initialUnit, "ft", { strict: false });
+    return convertTravelSpeed(value / 10, "mph", { strict: false, to: finalUnit }).value;
```
Fyorl: MrAioros sent me the formula for metric, so we can preserve some accuracy here.

Image


```suggestion
    const fromConfig = CONFIG.DND5E.movementUnits[initialUnit];
    const toConfig = CONFIG.DND5E.movementUnits[finalUnit];
    if ( (fromConfig.type === "metric") && (toConfig.type === "metric") ) {
      value = convertLength(value, initialUnit, "m", { strict: false });
      return convertTravelSpeed(value / 2, "kph", { strict: false, to: finalUnit }).value;
    } else {
      value = convertLength(value, initialUnit, "ft", { strict: false });
      return convertTravelSpeed(value / 10, "mph", { strict: false, to: finalUnit }).value;
    }
```

module/data/actor/fields/travel-field.mjs
```diff
@@ -0,0 +1,133 @@
...
+   */
+  static prepareData(rollData) {
+    const { movement, travel } = this.attributes;
+    const { pace, units } = travel;
+    travel.paces = {};
+    travel.max = 0;
+
+    const noMovement = this.parent.hasConditionEffect("noMovement");
+    const halfMovement = this.parent.hasConditionEffect("halfMovement");
+    const paceConfig = CONFIG.DND5E.travelPace[pace];
```
Fyorl: Not used.

module/data/actor/fields/travel-field.mjs
```diff
@@ -0,0 +1,133 @@
...
+    for ( const type of Object.keys(travel.speeds) ) {
+      let speed = Math.max(0, simplifyBonus(travel.speeds[type], rollData));
+      if ( noMovement ) speed = 0;
+      else if ( halfMovement ) speed *= 0.5;
+      travel.speeds[type] = speed;
+      if ( speed > travel.max ) travel.max = speed;
+      if ( speed ) travel.paces[type] = TravelField.convertSpeedToPace(speed, pace);
```
Fyorl: Not passing units here means we can't apply the weird logic for PHB 'standard' travel pace for groups.

module/data/actor/vehicle.mjs
```diff
@@ -247,20 +237,26 @@ export default class VehicleData extends CommonTemplate {
...
-   * vehicle's type.
+   * Migrate movement speeds to travel pace by taking the previous max speed and assigning it to a movement type
+   * based on the vehicle's type.
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
   static #migrateMovement(source) {
     const movement = source.attributes?.movement;
     const { vehicleType } = source;
-    if ( !vehicleType || !movement || !("walk" in movement) ) return;
+    const newUnits = { mi: "mph", km: "kph" }[movement?.units];
```
Fyorl: Trying to avoid creating throwaway objects in these in-memory migrations until I've had a chance to actually profile data prep performance.

```suggestion
    let newUnits;
    if ( movement?.units === "mi" ) newUnits = "mph";
    else if ( movement?.units === "km" ) newUnits = "kph";
```

module/data/shared/movement-field.mjs
```diff
@@ -69,26 +63,9 @@ export default class MovementField extends foundry.data.fields.SchemaField {
...
-        perDay = Math.round(inFeet * .8);
-      }
-      if ( pace ) {
-        const isStandard = (unitConfig.type === "imperial") && (perDay === CONFIG.DND5E.travelPace.normal.standard);
-        if ( isStandard ) this.paces[k] = paceConfig?.standard ?? perDay;
-        else this.paces[k] = Math.floor(perDay * (paceConfig?.multiplier ?? 1));
-      }
-      else this.paces[k] = perDay;
-    });
+    foundry.utils.logCompatibilityWarning(
```
Fyorl, line 66: Should we be calling `TravelPace.prepareData` here in order to complete the deprecation path?

module/utils.mjs
```diff
@@ -669,6 +685,22 @@ export function convertTime(value, from, { combat=false, strict=true, to }={}) {
...
+/**
+ * Convert the provided travel speed to another unit.
+ * @param {number} value                    The travel speed being converted.
+ * @param {string} from                     The initial unit.
+ * @param {object} options
+ * @param {boolean} [options.strict=false]  Throw an error if either unit isn't found.
+ * @param {string} options.to               The final unit.
+ * @returns {{ value: number, unit: string }}
+ */
+export function convertTravelSpeed(value, from, { strict=false, to }) {
```
Fyorl, line 697: I don't really understand when we use a signature like `convertTime` vs. a signature like `convertLength`. This function looks more like the latter than the former to me.

templates/shared/config/movement-senses-config.hbs
```diff
@@ -1,4 +1,5 @@
 <section class="flexcol">
+    {{#if fields}}
     <fieldset class="card">
```
Fyorl, line 3: Maybe give this a "Combat" legend?

module/data/actor/group.mjs
```diff
@@ -152,6 +130,26 @@ export default class GroupData extends GroupTemplate {
...
+  /* -------------------------------------------- */
+
+  /**
+   * Migrate travel speeds from `attributes.movement` to `attributes.travel`.
+   * @param {object} source  The candidate source data from which the model will be constructed.
+   */
+  static #migrateTravel(source) {
+    if ( source.attributes?.travel || !source.attributes?.movement ) return;
+    let { pace, units, ...speeds } = source.attributes.movement;
+    const finalUnit = { ft: "mph", mi: "mph", m: "kph", km: "kph" }[units];
```
Fyorl:
```suggestion
    const config = CONFIG.DND5E.movementUnits[units];
    const finalUnit = config?.type === "metric" ? "kph" : "mph";
```

module/data/actor/fields/travel-field.mjs
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    if ( !movement ) return;
+    for ( const [type, { travel: travelType="land" }] of Object.entries(CONFIG.DND5E.movementTypes) ) {
+      if ( !movement[type] || travel.speeds[travelType] ) continue;
+      const speed = TravelField.convertMovementToTravel(movement[type], movement.units, units);
+      const travelPace = TravelField.convertSpeedToPace(speed, pace, movement.units);
+      travel.paces[travelType] = Math.max(travelPace, travel.paces[travelType] ?? -Infinity);
```
Fyorl: Since we're going to the effort of adding the travel field here, perhaps we should go all the way and add daily travel pace data to cover the full spectrum of speeds that crop up in various publications and which don't necessarily follow the DMG formula.

So for example, instead of exclusively deriving `travel.paces`, we add it as another `MappingField`, alongside `speeds`, and then we do the same thing we're doing here where we fill in a derived value if an explicit one is not provided.

That has the added benefit of making the group travel pace migration simpler since we can migrate any `mi` paces directly to `mpd`, etc. without any conversion or assumptions.

I imagine the biggest difficulty will be naming for our various configs and conversion utils since we have already used `travel` somewhat ambiguously. But perhaps 'travel speed' vs. 'travel pace' is a good enough distinction.

module/data/shared/movement-field.mjs
arbron: I didn't because if they are still calling the old method, then they probably haven't added the travel data to the actor, so calling `TravelPace.prepareData` will throw an error.

module/utils.mjs
arbron: I have some future PRs that modify more of the methods to match the `convertTime` for better automatic conversion between imperial and metric. I decided to write this one using the future proposed API so it wouldn't require a deprecation path if that PR comes along.

templates/shared/config/movement-senses-config.hbs
```diff
@@ -1,26 +1,60 @@
...
+            <label>{{ label }}</label>
+            <div class="form-fields">
+                {{ formInput speed.field name=speed.name value=speed.value placeholder=speed.placeholder }}
+                {{ formInput pace.field name=pace.name value=pace.value placeholder=pace.placeholder }}
+            </div>
+        </div>
+        {{/each}}
+        <hr>
+        {{#if fields.units}}
+        {{ formField fields.units value=data.units options=unitsOptions localize=true blank=false rootId=@root.partId }}
```
Fyorl, line 52: This is currently reading as 'Miles per hour', but should probably just read 'Miles' since the 'per hour' and 'per day' parts are moved to the header now.

templates/actors/vehicle/sidebar.hbs
```diff
@@ -158,9 +158,17 @@
     {{/if}}
 
     {{!-- Speed --}}
+    {{#if showSpeed}}
     {{> "systems/dnd5e/templates/actors/parts/actor-trait-line.hbs" icon="fa-solid fa-shoe-prints"
         label="DND5E.Speed" configAction="movement" configLabel="DND5E.MOVEMENT.Action.Configure" unitsConfig=CONFIG.movementUnits
         data=(dnd5e-object value=system.attributes.movement.max units=system.attributes.movement.units) }}
+    {{/if}}
+    {{#if showTravel}}
```
Fyorl: GoS ships show mph and mpd for speed so it would be nice to have an extra line here if the travel pace is calculated. But if messing around with units is too much of a pain, we can leave it for now.

module/data/actor/vehicle.mjs
```diff
@@ -247,20 +237,28 @@ export default class VehicleData extends CommonTemplate {
...
-   * vehicle's type.
+   * Migrate movement speeds to travel pace by taking the previous max speed and assigning it to a movement type
+   * based on the vehicle's type.
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
   static #migrateMovement(source) {
     const movement = source.attributes?.movement;
     const { vehicleType } = source;
-    if ( !vehicleType || !movement || !("walk" in movement) ) return;
+    let newUnits;
```
Fyorl, line 247: I think this value was actually unit-less on the old sheets. In which case, we would never migrate since `newUnits` will always be `undefined`. Should we just assume mph?

module/data/actor/fields/travel-field.mjs
```diff
@@ -0,0 +1,148 @@
...
+ * Field for storing travel data.
+ */
+export default class TravelField extends foundry.data.fields.SchemaField {
+  constructor(fields={}, { initialUnits=null, ...options }={}) {
+    fields = {
+      pace: new StringField({
+        required: true, blank: false, initial: "normal", choices: () => CONFIG.DND5E.travelPace,
+        label: "DND5E.Travel.Label"
+      }),
+      paces: new MappingField(new FormulaField({ deterministic: true }), {
```
Fyorl, line 29: What sort of formulae might we expect in these fields? Any examples from official content?

module/data/actor/fields/travel-field.mjs
```diff
@@ -0,0 +1,148 @@
...
+    Object.defineProperty(travel.speeds, "max", { value: 0, writable: true });
+
+    const noMovement = this.parent.hasConditionEffect("noMovement");
+    const halfMovement = this.parent.hasConditionEffect("halfMovement");
+    const multiplier = noMovement ? 0 : halfMovement ? 0.5 : 1;
+    for ( const type of Object.keys(travel.speeds) ) {
+      let pace = travel.paces[type] = Math.max(0, simplifyBonus(travel.paces[type], rollData)) * multiplier;
+      const speed = travel.speeds[type] = Math.max(0, simplifyBonus(travel.speeds[type], rollData)) * multiplier;
+      if ( pace > travel.paces.max ) travel.paces.max = pace;
+      if ( speed > travel.speeds.max ) travel.speeds.max = speed;
```
Fyorl: I think we need to check these after we've done the derivations below.

module/data/shared/movement-field.mjs
Fyorl: I suppose we could make `TravelPace.prepareData` early return or otherwise be resilient to a lack of travel data. But I doubt there's actually any downstream code calling this method so we don't need to expend extra effort on the deprecation path.

module/utils.mjs
Fyorl: It seems like `_convertSystemUnits` will immediately throw if `to` is `undefined`, so it didn't seem like `to` was really optional. If there are changes to `_convertSystemUnits` as well in that future work, then that's fine. Definitely prefer avoiding a future deprecation if we can.

templates/actors/vehicle/sidebar.hbs
arbron: I can add that, the only difficulty is finding yet another Font Awesome icon :).

module/data/actor/fields/travel-field.mjs
arbron: Because we use these for displaying on the sheet, I decided to stick them before the derivation so that only speeds you explicitly enter are displayed on the sheet to give creators a bit more control over the presentation.

templates/actors/vehicle/sidebar.hbs
Fyorl: Maybe use `fa-sun` for now and I can tweak it later.

module/data/actor/vehicle.mjs
arbron: Movement had units on the old sheet:

Screenshot 2025-10-16 at 10 56 23

module/data/actor/fields/travel-field.mjs
arbron: None that I can think of, I was just setting it up as formula for parity across all movement/travel speeds when https://github.com/foundryvtt/dnd5e/pull/6274 goes in.

module/data/actor/vehicle.mjs
Fyorl: OK all good then

module/data/actor/fields/travel-field.mjs
Fyorl: Let's stick with FormulaField then

module/data/actor/fields/travel-field.mjs
Fyorl: Seems wrong to have a derived `max` available in the data that is not the correct value, it might be needed in other contexts. We should check source if we want to know if there's an explicit value defined. But even then, if I put 5 mph as the speed for a GoS ship, should I not see its mpd on the sheet? Or I have to manually enter the mpd even though the system calculates it?

arbron: Adjusted so derived speeds and paces are used in max and shown on the sheet.

PR #6298, opened 2025-10-12
Add `adv` and `dis` die modifiers for advantage handling

arbron:
Adds a new pair of die modifiers: `adv` for rolling with advantage, and `dis` for rolling with disadvantage. These modifiers simplify the normal setup so giving a roll advantage no longer requires changing the number of dice rolled.

Advantage Modifier - D20 Rolls

The modifier also supports a number, which indicates the number of additional times the dice should be rolled, so `adv2` can be used to handle Elven Accuracy, indicating that the original die formula should be rolled 2 additional times and only the highest should be taken.

Advantage Modifier - Elven Accuracy

The modifier also works for larger die pools, so if you wanted to roll your damage with advantage you would just write `4d6adv` and it would roll the dice twice, and take the set of four rolled that has the highest total (different than changing it to `8d6kh4`, which would select the highest four dice from all of those rolled and would result in a different total).

Advantage Modifier - Damage

### DIFF
--- a/module/dice/d20-die.mjs
+++ b/module/dice/d20-die.mjs
@@ -68,7 +68,7 @@ export default class D20Die extends BasicDie {
    */
   applyAdvantage(advantageMode) {
     this.options.advantageMode = advantageMode;
-    this.modifiers.findSplice(m => m.startsWith("adv") || m.startsWith("dis") || (m === kh) || (m === kl));
+    this.modifiers.findSplice(m => m.startsWith("adv") || m.startsWith("dis") || (m === "kh") || (m === "kl"));
     this.number = 1;
     if ( advantageMode === CONFIG.Dice.D20Roll.ADV_MODE.NORMAL ) return;
     const isAdvantage = advantageMode === CONFIG.Dice.D20Roll.ADV_MODE.ADVANTAGE;


module/dice/d20-die.mjs
```diff
@@ -68,13 +68,10 @@ export default class D20Die extends Die {
    */
   applyAdvantage(advantageMode) {
     this.options.advantageMode = advantageMode;
-    this.modifiers.findSplice(m => ["kh", "kl"].includes(m));
```
Fyorl, line 71: If we process an existing advantage formula, like `2d20kh`, don't we now end up with `2d20khadv`, which will give a strange result.

arbron: True, but I'm not sure there are many cases that will happen since `applyAdvantage` is only called during the rolling process, so it shouldn't be happening on any rolls that still have `kh` or `kl` unless they specifically modified to have `options.configured = false`.

That said its pretty easy to strip those out so we can just do that anyway.

module/dice/d20-die.mjs
```diff
@@ -68,13 +68,11 @@ export default class D20Die extends Die {
...
-    if ( advantageMode === CONFIG.Dice.D20Roll.ADV_MODE.NORMAL ) this.number = 1;
-    else {
-      const isAdvantage = advantageMode === CONFIG.Dice.D20Roll.ADV_MODE.ADVANTAGE;
-      this.number = (isAdvantage && this.options.elvenAccuracy) ? 3 : 2;
-      this.modifiers.push(isAdvantage ? "kh" : "kl");
-    }
+    this.modifiers.findSplice(m => m.startsWith("adv") || m.startsWith("dis") || (m === kh) || (m === kl));
```
Fyorl:
```suggestion
    this.modifiers.findSplice(m => m.startsWith("adv") || m.startsWith("dis") || (m === "kh") || (m === "kl"));
```

arbron:
> Thanks, and do we need to make any changes to `BasicRoll.preCalculateTerm`?

I don't believe so, because we are not modifying the die count when using this new term so it won't affect the min/max calculations that `preCalculateTerm` is designed to accommodate.

PR #6299, opened 2025-10-13
[#4610] Allow certain enrichers to be invoked from chat

### LINKED ISSUE #4610: Make damage/healing enrichers available as chat commands
I just had my mind blown when I found out that /award is an actual valid chat command. In the favor of saving clicks, I think every enrichers should be considered as an option for a chat command, it was super cool to figure out that I award like this for quick things.

With damage/healing, it could open the roll window immediately. I think it'd be fine as well if it still posted the chat message with the enricher, in case you need to reroll it, but would help with the flow if it opened up the roll dialog immediately. 

I can't think of use-cases with the other enrichers right now, but I think it's worth the consideration for similar usage.

arbron:
Allows for `/check`, `/save`, `/attack` and `/damage` commands to be used from chat using the same options as if they were used in the enricher format.

For the `/check` and `/save` commands, a new `request` option can be included to create a roll request card directly.

Closes #4610

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -9,8 +9,10 @@ const slugify = value => value?.slugify().replaceAll("-", "").replaceAll("(", ""
 const logWarning = (msg, options) =>
   log(options.relativeTo ? `${msg} [${options.relativeTo.uuid}]` : msg, { level: "warn" });
 
-const VALID_CHAT_COMMANDS = ["attack", "check", "concentration", "damage", "heal", "healing", "save", "skill", "tool"];
-const VALID_COMMANDS = [...VALID_CHAT_COMMANDS, "award", "item"];
+const VALID_CHAT_COMMANDS = [
+  "attack", "award", "check", "concentration", "damage", "heal", "healing", "save", "skill", "tool"
+];
+const VALID_COMMANDS = [...VALID_CHAT_COMMANDS, "item"];
 const makeCommandPattern = commands => `/(?<type>${commands.join("|")})(?<config> .*?)?`;
 const CHAT_REGEX = new RegExp(`^${makeCommandPattern(VALID_CHAT_COMMANDS)}$`, "i");
 
@@ -51,6 +53,7 @@ export function chatMessage(message) {
   config = parseConfig(config, { multiple: ["damage", "heal", "healing"].includes(type) });
   switch (type) {
     case "attack": handleAttackCommand(config); break;
+    case "award": Award.handleAward(message); break;
     case "heal":
     case "healing": config._isHealing = true;
     case "damage": handleDamageCommand(config); break;
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -60,6 +60,7 @@ export function chatMessage(message) {
     case "check":
     case "skill":
     case "tool": handleCheckCommand(config); break;
+    case "concentration": config._isConcentration = true;
     case "save": handleSaveCommand(config); break;
   }
   return false;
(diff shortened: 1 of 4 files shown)


module/enrichers.mjs
```diff
@@ -7,16 +7,20 @@ import { rollItem } from "./documents/macro.mjs";
 
 const slugify = value => value?.slugify().replaceAll("-", "").replaceAll("(", "").replaceAll(")", "");
 
+const VALID_CHAT_COMMANDS = [
+  "attack", "award", "check", "concentration", "damage", "heal", "healing", "save", "skill", "tool"
```
Fyorl, line 13: `concentration` doesn't have a handler in `chatMessage`, did you mean to include it?

module/enrichers.mjs
```diff
@@ -597,14 +642,30 @@ export async function enrichCheck(config, label, options) {
...
 /* -------------------------------------------- */
 
+/**
+ * Handle a check command in chat.
+ * @param {object} config  Configuration data.
+ */
+function handleCheckCommand(config) {
+  config = parseCheckConfig(config);
+  config.type = "check";
+  if ( config.request ) return handlePostRequest(config);
```
Fyorl, line 657: When we use a check enricher, i.e. `[[/check ath 15]]`, there is some logic in the enricher renderer that selects the ability from the system's configured default skill abilities, and injects that information into the request button. We don't do that here, so the `handlePostRequest` has no ability and you end up with a request button for 'DC 15 undefined'.

module/enrichers.mjs
```diff
@@ -35,6 +39,32 @@ export function registerCustomEnrichers() {
...
+ * @param {string} message  Text of the message being posted.
+ * @returns {false|void}    Returns `false` to prevent the message from continuing to parse.
+ */
+export function chatMessage(message) {
+  const match = message.match(CHAT_REGEX);
+  if ( !match ) return;
+  let { type, config } = match.groups;
+  config = parseConfig(config, { multiple: ["damage", "heal", "healing"].includes(type) });
+  switch (type) {
+    case "attack": handleAttackCommand(config); break;
```
Fyorl, line 55: Attempting a blank `/attack` in the chat throws an error. Maybe that's a degenerate case, but I suppose it should just roll a flat d20?

Fyorl (changes requested):

module/enrichers.mjs
```diff
@@ -597,15 +643,37 @@ export async function enrichCheck(config, label, options) {
...
  * @param {object} dataset
  * @returns {object[]}
  */
 function createCheckRequestButtons(dataset) {
-  const skills = dataset.skill?.split("|") ?? [];
-  const tools = dataset.tool?.split("|") ?? [];
-  if ( (skills.length + tools.length) <= 1 ) return [createRequestButton(dataset)];
+  const skills = foundry.utils.getType(dataset.skill) === "string" ? dataset.skill.split("|") : dataset.skill ?? [];
+  const tools = foundry.utils.getType(dataset.tool) === "string" ? dataset.tool.split("|") : dataset.tool ?? [];
+  if ( (skills.length + tools.length) <= 1 ) {
```
Fyorl: Looks good but I think we are still missing full parity with the `/check` enricher. For example `[[/check slt thief]]` is a valid enricher, but pressing the roll request for that enricher gives a different result to using `/check slt thief request` in chat. We probably need to extract more of that logic into `parseCheckConfig`.

Fyorl (changes requested):

module/enrichers.mjs
arbron: The "using tool" use case should be fixed, plus I've made full keys work properly (e.g. `/check stealth`).

Fyorl (approved):

PR #6302, opened 2025-10-13
[#5029] Add `[[language]]` enricher

### LINKED ISSUE #5029: [Feature suggestion] Language enricher
In certain instances where players encounter text in exotic languages, saying "the text is in Abyssal, can any of you read it?" can be problematic and somewhat gives away the content of the text (a message in Abyssal is most likely evil or dangerous to read). So I think it would be valuable to have a /language enricher, something like the passive perception enricher, that displays which of the players meets the requirement (in this case, being proficient with the language in question). Something like this:

Image

arbron:
Adds a new passive enricher that displays which members of the party have proficiency in a certain language when hovered. Reworks the tooltip to handle more generic passive enrichers in the future.

Language Enricher

Closes #5029

### DIFF
--- a/module/tooltips.mjs
+++ b/module/tooltips.mjs
@@ -175,7 +175,7 @@ export default class Tooltips5e {
         const abilityMod = systemData.abilities?.[ability]?.mod;
         if ( abilityMod !== undefined ) passive = 10 + abilityMod;
       }
-      if ( !passive ) return false;
+      if ( passive === undefined ) return false;
       return {
         passive, status: dc !== undefined ? passive >= dc ? "success" : "failure" : ""
       };
@@ -198,7 +198,7 @@ export default class Tooltips5e {
    * @protected
    */
   async _onHoverPassive({ label }, memberCallback) {
-    const party = game.actors.party;
+    const { party } = game.actors;
     if ( !party ) {
       this.tooltip.innerHTML = label;
       return;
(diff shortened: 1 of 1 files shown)


module/tooltips.mjs
```diff
@@ -139,33 +160,55 @@ export default class Tooltips5e {
...
-        const customSkillData = member.actor?._prepareSkill(skill, { ability });
+        const customSkillData = actor._prepareSkill(skill, { ability });
         passive = customSkillData.passive;
       } else {
         // Passive ability check
-        const abilityMod = systemData?.abilities?.[ability]?.mod;
+        const abilityMod = systemData.abilities?.[ability]?.mod;
         if ( abilityMod !== undefined ) passive = 10 + abilityMod;
       }
+      if ( !passive ) return false;
```
Fyorl: Might have a passive of `0`


```suggestion
      if ( passive === undefined ) return false;
```

module/tooltips.mjs
```diff
@@ -139,33 +160,55 @@ export default class Tooltips5e {
...
+
+  /**
+   * Handle hovering a passive tooltip to display results for primary party.
+   * @param {object} context
+   * @param {string} context.label                  Header or fallback label if no party is defined.
+   * @param {PassiveMemberCallback} memberCallback  Method called for each member to get additional context information.
+   * @protected
+   */
+  async _onHoverPassive({ label }, memberCallback) {
+    const party = game.actors.party;
```
Fyorl:
```suggestion
    const { party } = game.actors;
```

module/tooltips.mjs
```diff
@@ -139,33 +160,55 @@ export default class Tooltips5e {
...
+    if ( !party ) {
+      this.tooltip.innerHTML = label;
+      return;
+    }
+
+    const context = { label, party: [] };
+    for ( const member of party.system.members ) {
+      if ( !member.actor ) continue;
+      const c = memberCallback(member.actor);
+      if ( c ) context.party.push({ name: member.actor.name, img: member.actor.img, ...c });
```
Fyorl:
```suggestion
      const ctx = memberCallback(member.actor);
      if ( ctx ) context.party.push({ name: member.actor.name, img: member.actor.img, ...ctx });
```

Fyorl (approved):

PR #6303, opened 2025-10-14
[#5266] Add flag to replace values in stat blocks for "templates"

### LINKED ISSUE #5266: Add method for annotating "template" summoning actors that will modify embedded stat block
Right now there is no mechanism to make an embedded stat block match the ones presented in spells like "Summon Beast" because concepts like `AC: 11 + the spell's level` and `HP: 20 (Air only) or 30 (Land and Water only) + 5 for each spell level above 2` aren't representable easily in actor data.

Theoretically this could be done as an embed for the summon activity itself, but that would be a very complex endeavor. It would require comparing the stat blocks for each summoning profile and generating differences between them (e.g. any item that appears only on one of the profiles will be marked as `Pack Tactics (Land Only)` and different HP and speeds will need to be merged), plus representing any changes entered in the summon changes tab as human-readable (evaluating `@item.level` in the AC bonus as `+ the spell's level`).

Another option is adding in a more limited system using flags that allows for overriding any normal stat block fields with a custom value. So the template actor for Summon Beast might have this flag data:

```javascript
{
  "dnd5e.statBlockOverride": {
    ac: "11 + the spell's level",
    hp: "20 (Air only) or 30 (Land and Water only) + 5 for each spell level above 2",
    speed: "30 ft.; Climb 30 ft. (Land only); Fly 60 ft. (Air only); Swim 30 ft. (Water only)",
    pb: "equals your Proficiency Bonus"
  }
}
```

Then during the creation of the embedded stat block if would check these flags and display their values rather than the auto-calculated values.

arbron:
Adds a new `dnd5e.statBlockOverride` flag that allows for defining values that will replace the default stat block values when a NPC is embedded. This makes it possible to display summoned "template" actors accurately in journals using embeds.

Bestial Spirit Overrides

```javascript
{
  "dnd5e.statBlockOverride": {
    ac: "11 + the spell's level",
    hp: "20 (Air only) or 30 (Land and Water only) + 5 for each spell level above 2",
    speed: "30 ft.; Climb 30 ft. (Land only); Fly 60 ft. (Air only); Swim 30 ft. (Water only)",
    pb: "equals your Proficiency Bonus"
  }
}
```

Current keys are: `tag`, `ac`, `hp`, `speed`, `senses`, `cr`, `xp`, `pb`, `gear`, `languages`, `vulnerabilities`, `resistances`, `immunities`, `conditionImmunities`, `initiative`, and `skills`. Not all of these will be used depending on the rules version used when embedding.

Closes #5266

### DIFF
--- a/module/data/actor/npc.mjs
+++ b/module/data/actor/npc.mjs
@@ -616,6 +616,7 @@ export default class NPCData extends CreatureTemplate {
       ...Array.from(value).map(t => Trait.keyLabel(t, { trait })).filter(_ => _),
       ...splitSemicolons(custom ?? "")
     ].sort((lhs, rhs) => lhs.localeCompare(rhs, game.i18n.lang)));
+    const o = this.parent.flags.dnd5e?.statBlockOverride ?? {};
 
     const prepareSpeed = () => {
       const standard = formatter.format([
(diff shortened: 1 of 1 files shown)


Zanderaf:
Not sure how much of an ask this is, but any shot we could have a key for `size`, `name` and `type`?

Fyorl (approved):

arbron:
> Not sure how much of an ask this is, but any shot we could have a key for `size`, `name` and `type`?

Name can already be customized using the label on the enricher itself (e.g. `@Embed[...]{Ape}`), the other two can be handled using the `tag` replacement, but I suppose we could split them up.

PR #6305, opened 2025-10-14
[#6247] Add speed reduction from armor's strength requirement

### LINKED ISSUE #6247: Heavy armor does not reduce speed, despite not meeting the strength requirement.
According to SRD 5.2, certain types of armor have strength requirements, and if these are not met, the character receives a 10-feet penalty to their speed.

Image

Image

Current items in the SRD compendium, such as Plate Armor, have these strength requirements set.

Image

However, when equipping an item, the speed penalty does not apply even if the requirements are not met.

Image

Image

Tested on Foundry 13.348 and system version 5.1.9

arbron:
Add speed reduction when wearing armor and not meeting its strength requirement. The reduction is defined in the config at `DND5E.armorSpeedReduction` in feet and converted to meters if necessary.

Closes #6247

### DIFF
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -165,7 +165,7 @@ export default class AttributesFields {
     }, { armors: [], shields: [] });
 
     // Set stealth disadvantage
-    if ( armors[0]?.system.properties.has("stealthDisadvantage") ) {
+    if ( armors[0]?.system.properties.has("stealthDisadvantage") && this.skills ) {
       AdvantageModeField.setMode(this, "skills.ste.roll.mode", -1);
     }
 
@@ -410,7 +410,7 @@ export default class AttributesFields {
 
   /**
    * Modify movement speeds taking exhaustion and any other conditions into account.
-   * @this {CharacterData|NPCData}
+   * @this {CharacterData|NPCData|VehicleData}
    * @param {ActorRollData} rollData  The Actor's roll data.
    */
   static prepareMovement(rollData=this.parent.getRollData()) {
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -430,6 +430,9 @@ export default class AttributesFields {
     const units = this.attributes.movement.units ??= defaultUnits("length");
     let reduction = dnd5e.settings.rulesVersion === "modern" && !this.traits?.ci?.value?.has("exhaustion")
       ? (this.attributes.exhaustion ?? 0) * (CONFIG.DND5E.conditionTypes.exhaustion?.reduction?.speed ?? 0) : 0;
+    if ( (this.attributes.ac?.equippedArmor?.system.strength ?? 0) > (this.abilities?.str?.value ?? Infinity) ) {
+      reduction += CONFIG.DND5E.armorSpeedReduction;
+    }
     reduction = convertLength(reduction, CONFIG.DND5E.defaultUnits.length.imperial, units);
     const bonus = simplifyBonus(this.attributes.movement.bonus, rollData);
     this.attributes.movement.max = 0;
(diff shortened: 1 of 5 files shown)


Zanderaf:
There any way to ignore this with the 2014 PHB Dwarf's feature?
`Your speed is not reduced by wearing heavy armor.` from their `Speed` entry

Fyorl (commented):
Perhaps this should generate an AE the same way that encumbrance does. That would allow it to be toggled off in the case of various exceptions.

arbron:
> Perhaps this should generate an AE the same way that encumbrance does. That would allow it to be toggled off in the case of various exceptions.

Biggest potential issue with that is it wouldn't respond to any AEs that modify your strength score (like a Belt of Giant Strength) because the speed-reducing AE would need to be created on document update. Not sure that plus the complexity of monitoring all actor updates would be worth it rather than just adding a flag.

Fyorl:
> Biggest potential issue with that is it wouldn't respond to any AEs that modify your strength score (like a Belt of Giant Strength) because the speed-reducing AE would need to be created on document update. Not sure that plus the complexity of monitoring all actor updates would be worth it rather than just adding a flag.

A flag is fine. Though I suppose this issue also must exist for encumbrance in that case.

module/data/actor/templates/attributes.mjs
```diff
@@ -428,6 +428,10 @@ export default class AttributesFields {
     const units = this.attributes.movement.units ??= defaultUnits("length");
     let reduction = dnd5e.settings.rulesVersion === "modern"
       ? (this.attributes.exhaustion ?? 0) * (CONFIG.DND5E.conditionTypes.exhaustion?.reduction?.speed ?? 0) : 0;
+    if ( ((this.attributes.ac?.equippedArmor?.system.strength ?? 0) > (this.abilities?.str?.value ?? Infinity))
```
Fyorl, line 433: This will apply to NPCs and Vehicles too. Probably fine for vehicles since they use AC `flat`, but not technically required anymore. Also wouldn't put it past WotC to have NPCs wearing armour heavier than their strength score allows so this might also subtly affect some NPCs out there. At the same time, sidekicks or other long-term NPCs might want these rules applied appropriately.

Maybe we wait to see if this hits any official NPCs. Probably should exclude vehicles though to be safe.

packs/_source/races/dwarf/hill-dwarf.yml
```diff
@@ -44,8 +44,10 @@ system:
       id="secret-bkbaaEWqAGbU8ZgK"><p><strong>Foundry
       Note</strong></p><p>Dwarven Toughness should be configured by adding 1 to
       the "Per Level" bonus in HP configuration on your character sheet
-      (accessible using the cog next to your character's hit
-      points).</p></section>
+      (accessible using the cog next to your character's hit points).</p><p>The
```
Fyorl, line 47: We could put this as an AE on one of their features (I know Speed doesn't have a feature), like we do for Halfling Nimbleness. But I notice we have some inconsistency there because we don't do this for the Halfling Lucky trait.

Fyorl (approved):
Looks good. I verified that we do have this same problem you described with the encumbrance AEs, so we should look at moving those away from AEs and making them internal to actor prep.

module/data/actor/templates/attributes.mjs
arbron: Not much worry at the moment, because equipping armor complete breaks data preparation for vehicles 😭. I'll see if I can fix that.

arbron: Fixed the data preparation error, but right now I'm leaning towards not ignoring vehicles. Since putting a strength requirement in armor, equipping it on a vehicle, and then changing the preparation mode for the vehicle is a very deliberate set of steps we could probably just assume people doing that are doing it with intention.

Fyorl: We can assume they're doing it with intention, but the rules on equipped armour apply to creatures, not vehicles.

PR #6306, opened 2025-10-14
[#6304] Add movement bonus that applies only to existing speeds

### LINKED ISSUE #6304: Add movement bonus that only applies to movement types that have existing value
Currently it is difficult to implement an effect like Longstrider. The spell should add 10 feet to all of the actor's movement speeds, but if an AE is added that includes `system.attributes.movement.* | ADD | 10` for each different movement type it will end up giving the character movement speeds that they don't already possess. To handle this properly, there should be an additional movement bonus field that would be added to each of the movement types only so long as they have existing movement of that type.

arbron:
Introduces the `attributes.movement.bonus` formula field that allows users to enter a bonus that is applied to all movement speeds so long as a given movement type already has a speed.

Closes #6304

### DIFF
--- a/module/data/shared/movement-field.mjs
+++ b/module/data/shared/movement-field.mjs
@@ -1,3 +1,5 @@
+import FormulaField from "../fields/formula-field.mjs";
+
 const { BooleanField, NumberField, SetField, StringField } = foundry.data.fields;
 
 /**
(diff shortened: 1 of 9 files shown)


module/data/shared/movement-field.mjs
```diff
@@ -31,6 +33,7 @@ export default class MovementField extends foundry.data.fields.SchemaField {
       fly: new NumberField({ ...numberConfig, label: "DND5E.MOVEMENT.Type.Fly", speed: true }),
       swim: new NumberField({ ...numberConfig, label: "DND5E.MOVEMENT.Type.Swim", speed: true }),
       walk: new NumberField({ ...numberConfig, label: "DND5E.MOVEMENT.Type.Walk", speed: true }),
+      bonus: new FormulaField({ deterministic: true, label: "DND5E.MOVEMENT.FIELDS.bonus.label" }),
```
krbz999, line 21: What's the use case for having it in the schema and not just initialized during `prepareBaseData`?

arbron: In case you wanted to just type the bonus on your sheet?

PR #6312, opened 2025-10-16
[#6311] Convert dependent document tracking to registry

### LINKED ISSUE #6311: Convert dependent tracking to use flag on dependent document & registry, rather than flag on active effect
Currently if a dependent item is created as part of a rider enchantment and then moved to a different actor, the item is no longer linked as a dependent and will not be deleted when the enchantment is removed because its UUID no longer matches the one stored in the enchantment flags. By changing dependent items to use a flag on the created document instead and linking them up with a registry, the system will be able to track these documents even if they are moved/duplicated.

arbron:
Switches to using a `dnd5e.dependentOn` flag on dependent documents and a `dnd5e.registry.dependents` registry to track and handle removing dependent documents, rather than a flag on the active effect to which they are dependent.

For dependent activities and effects created directly on the item the `dependentOn` flag is stored as a relative UUID to ensure that duplicating an item with an enchantment will no lead to riders from that enchantment being linked to the incorrect document.

Closes #6311

### DIFF
--- a/module/registry.mjs
+++ b/module/registry.mjs
@@ -41,26 +41,26 @@ class DependentsRegistry {
 
   /**
    * Resolve an active effect ID into an absolute UUID.
-   * @param {string} id            ID or UUID of active effect.
+   * @param {string} idOrUuid      ID or UUID of active effect.
    * @param {Document} dependent   Document to track as a dependent.
    * @returns {string}
    */
-  static #resolveDependentID(id, dependent) {
-    if ( id.length > 16 ) return id;
+  static #resolveDependentID(idOrUuid, dependent) {
+    if ( idOrUuid.length > 16 ) return idOrUuid;
     let relative = dependent.parent;
     if ( relative && !(relative instanceof Item) ) relative = relative.parent;
-    return relative.effects.get(id)?.uuid;
+    return relative.effects.get(idOrUuid)?.uuid;
   }
 
   /* -------------------------------------------- */
 
   /**
    * Add a dependent document to the registry.
-   * @param {string} uuid          ID or UUID of active effect.
+   * @param {string} idOrUuid      ID or UUID of active effect.
    * @param {Document} dependent   Document to track as a dependent.
    */
-  static track(uuid, dependent) {
-    uuid = DependentsRegistry.#resolveDependentID(uuid, dependent);
+  static track(idOrUuid, dependent) {
+    const uuid = DependentsRegistry.#resolveDependentID(idOrUuid, dependent);
     if ( !uuid ) return;
     if ( !DependentsRegistry.#dependents.has(uuid) ) DependentsRegistry.#dependents.set(uuid, new Set());
     DependentsRegistry.#dependents.get(uuid).add(dependent.uuid);
@@ -70,11 +70,11 @@ class DependentsRegistry {
 
   /**
    * Remove a dependent document from the registry.
-   * @param {string} uuid         ID or UUID of active effect.
+   * @param {string} idOrUuid     ID or UUID of active effect.
    * @param {Document} dependent  Dependent document to stop tracking.
    */
-  static untrack(uuid, dependent) {
-    uuid = DependentsRegistry.#resolveDependentID(uuid, dependent);
+  static untrack(idOrUuid, dependent) {
+    const uuid = DependentsRegistry.#resolveDependentID(idOrUuid, dependent);
     DependentsRegistry.#dependents.get(uuid)?.delete(dependent.uuid);
   }
 }


module/documents/mixins/dependent.mjs
```diff
@@ -0,0 +1,29 @@
...
+ * @returns {typeof DependentDocument}
+ * @mixin
+ */
+export default function DependentDocumentMixin(Base) {
+  class DependentDocument extends Base {
+    /** @inheritDoc */
+    prepareData() {
+      super.prepareData();
+      if ( this.flags?.dnd5e?.dependentOn && this.uuid ) {
+        dnd5e.registry.dependents.track(this.flags.dnd5e?.dependentOn, this);
```
Fyorl:
```suggestion
        dnd5e.registry.dependents.track(this.flags.dnd5e.dependentOn, this);
```

module/documents/mixins/dependent.mjs
```diff
@@ -0,0 +1,29 @@
...
+      }
+    }
+
+    /* -------------------------------------------- */
+
+    /** @inheritDoc */
+    _onDelete(options, userId) {
+      super._onDelete(options, userId);
+      if ( this.flags?.dnd5e?.dependentOn && this.uuid ) {
+        dnd5e.registry.dependents.untrack(this.flags.dnd5e?.dependentOn, this);
```
Fyorl:
```suggestion
        dnd5e.registry.dependents.untrack(this.flags.dnd5e.dependentOn, this);
```

module/registry.mjs
```diff
@@ -1,6 +1,83 @@
 import CompendiumBrowser from "./applications/compendium-browser.mjs";
 import { formatIdentifier } from "./utils.mjs";
 
+/* -------------------------------------------- */
+/*  Dependents                                  */
+/* -------------------------------------------- */
+
+class DependentsRegistry {
+  /**
+   * Registration of documents that are dependent of an active effect. The map is keyed by the UUID of
```
Fyorl:
```suggestion
   * Registration of documents that are dependent on an active effect. The map is keyed by the UUID of
```

module/registry.mjs
```diff
@@ -1,6 +1,83 @@
...
+   * the active effect upon which the document is dependent and contains a set of UUIDs for that effect's
+   * dependents. All UUIDs are expected to be world UUIDs or UUIDs of documents with the same ancestor
+   * document as the effect they are dependent on.
+   * @type {Map<string, Set<string>}
```
Fyorl:
```suggestion
   * @type {Map<string, Set<string>>}
```

module/registry.mjs
```diff
@@ -1,6 +1,83 @@
...
+   */
+  static #dependents = new Map();
+
+  /* -------------------------------------------- */
+
+  /**
+   * Fetch dependent documents for an active effect.
+   * @param {ActiveEffect|string} effect  Active effect for which to get the dependent documents or string for an
```
Fyorl:
```suggestion
   * @param {ActiveEffect|string} effect  Active effect for which to get the dependent documents or UUID for an
```

module/registry.mjs
```diff
@@ -1,6 +1,83 @@
...
+        }
+        return fromUuidSync(uuid, { strict: false });
+      })
+      .filter(_ => _);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Convert relative UUID into concrete UUID.
```
Fyorl:
```suggestion
   * Resolve a relative UUID into an absolute UUID.
```

module/documents/active-effect.mjs
```diff
@@ -450,23 +453,25 @@ export default class ActiveEffect5e extends ActiveEffect {
       }
       return effectData;
     }));
-    riderEffects = riderEffects.filter(_ => _);
+    riderEffects = riderEffects.filter(_ => _)
+    riderEffects.forEach(e => foundry.utils.setProperty(e, "flags.dnd5e.dependentOn", relativeUUID));
```
Fyorl: Maybe activities have some idiosyncratic `getRelativeUUID` definition that means this isn't possible, but the way relative UUIDs are intended to be used is that they are relative to the Document that they appear on. So if we are marking an AE here as dependent on another AE, we should store the relative UUID like so:

```suggestion
    riderEffects.forEach(e => foundry.utils.setProperty(e, "flags.dnd5e.dependentOn", this.getRelativeUUID(e)));
```

Then resolving the UUID becomes straightforward as `fromUuid(document.getFlag("dnd5e", "dependentOn"), { relative: document })`

module/registry.mjs
```diff
@@ -1,6 +1,83 @@
...
+   * @param {string} uuid          UUID of active effect.
+   * @param {Document} dependent   Document to track as a dependent.
+   * @returns {string}
+   */
+  static #resolveRelativeUUID(uuid, dependent) {
+    if ( !uuid.startsWith(".") ) return uuid;
+    let relative = dependent.parent;
+    if ( relative && !(relative instanceof Item) ) relative = relative.parent;
+    return foundry.utils.parseUuid(uuid, { relative }).uuid;
```
Fyorl: See my earlier comment but I would expect this to be more straightforward:


```suggestion
    return foundry.utils.parseUuid(uuid, { relative: dependent }).uuid;
```

module/registry.mjs
```diff
@@ -1,6 +1,83 @@
...
+
+  /**
+   * Add a dependent document to the registry.
+   * @param {string} uuid          UUID of active effect.
+   * @param {Document} dependent   Document to track as a dependent.
+   */
+  static track(uuid, dependent) {
+    uuid = DependentsRegistry.#resolveRelativeUUID(uuid, dependent);
+    if ( !DependentsRegistry.#dependents.has(uuid) ) DependentsRegistry.#dependents.set(uuid, new Set());
+    const set = DependentsRegistry.#dependents.get(uuid).add(dependent.uuid);
```
Fyorl:
```suggestion
    DependentsRegistry.#dependents.get(uuid).add(dependent.uuid);
```

module/documents/active-effect.mjs
arbron: It causes issues with `_resolveRelativeUuid`, because it checks that `relative` is an instance of `foundry.abstract.Document` and that will fail since Activities don't extend documents. Also, if we went with the shorting relative UUID format (`.vXqIWIc0PAiEspIw` as opposed to `.ActiveEffect.vXqIWIc0PAiEspIw`), event if activities could work with relative UUID parsing, then there would still be an issue when it tries to fetch the dependent active effect relative to the activity because with the shorter format it assumes the type is the same as the relative document (e.g. if only `.vXqIWIc0PAiEspIw` is provided with an activity, then the type would also be assumed to be an activity).

Fyorl: Well you wouldn't use the same UUID between the rider effects and activities. You resolve the relative UUID against the Document the UUID is being stored on, and that resulting UUID allows you to retrieve the original Document when given the storing Document.

```js
fromUuidSync(this.getRelativeUUID(activity), { relative: activity }) === this;
fromUuidSync(this.getRelativeUUID(effect), { relative: effect }) === this;
```

But if relative UUIDs don't work properly for activities then I suppose we don't have much choice. I would consider it at least for the effects, that way we have a clear exception path for activities-only because they don't support proper relative UUIDs.

arbron: But the issue is, in both cases the relative UUID is pointing to an Active Effect, so its more like:

```javascript
fromUuidSync(this.getRelativeUUID(effect1), { relative: activity }); // Error
fromUuidSync(this.getRelativeUUID(effect1), { relative: effect2 }); // Works
```

We could just store the ID of the active effect and fetch it directly rather than using a UUID, since we always know it will be an active effect. In that case there would still need to be logic for grabbing the parent item that differs per type (in activities, `activity.parent !== activity.item`, and active effects won't have an getter `effect.item` until V14, so we need to handle both cases slightly differently no matter what).

Fyorl: Hmm, maybe something I'm not understanding or I'm not getting across correctly. Let's discount the activity case for now because we know that can't work due to how activities are not real documents.

If we were doing this just for rider effects, I'm saying that the correct way to store and use this relative UUID would be like so:

```js
riderEffects.forEach(e => foundry.utils.setProperty(e, "flags.dnd5e.dependentOn", this.getRelativeUUID(e)));
```

Then, if we have that rider effect and we need to retrieve the AE that it is dependent on, such as in `DependentsRegistry##resolveRelativeUUID`, it's a very simple case of resolving against the document the UUID is stored on, e.g.:

```js
foundry.utils.parseUuid(uuid, { relative: dependent }); // Returns the 'this' from our earlier code
```

Is there something I'm missing that would mean this doesn't work?

arbron: Yeah, that would work. Regardless of the way we went, it is always going to require special casing for activities.

Anyways, to bypass this whole issue I've just pushed up a commit that switches to using the ID for the `dependentOn` flag for same-item documents.

module/registry.mjs
```diff
@@ -1,6 +1,84 @@
...
+        return fromUuidSync(uuid, { strict: false });
+      })
+      .filter(_ => _);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Resolve an active effect ID into an absolute UUID.
+   * @param {string} id            ID or UUID of active effect.
```
Fyorl: Let's make this change for `track` and `untrack` as well then.

```suggestion
   * @param {string} idOrUuid            Active effect UUID, or ID if sharing the same parent.
```

Fyorl (approved):

PR #6316, opened 2025-10-17
Allow groups to set a primary vehicle.

Fyorl:
This PR contains implementations for:
- Marking a vehicle as the primary vehicle for a group.
- Inheriting travel pace from the primary vehicle.
- Automation for the 'Slower Travelers' rule from the DMG.
- An inventory toggle to allow the group to use the primary vehicle's inventory as their own.

image

### DIFF
--- a/module/applications/actor/group-sheet.mjs
+++ b/module/applications/actor/group-sheet.mjs
@@ -316,7 +316,9 @@ export default class GroupActorSheet extends MultiActorSheet {
   /** @inheritDoc */
   _onClose(options) {
     super._onClose(options);
-    this.actor.system.members.forEach(({ actor }) => delete actor.apps[this.id]);
+    this.actor.system.members.forEach(({ actor }) => {
+      if ( actor ) delete actor.apps[this.id];
+    });
   }
 
   /* -------------------------------------------- */
@@ -324,7 +326,9 @@ export default class GroupActorSheet extends MultiActorSheet {
   /** @inheritDoc */
   async _onFirstRender(context, options) {
     await super._onFirstRender(context, options);
-    this.actor.system.members.forEach(({ actor }) => actor.apps[this.id] = this);
+    this.actor.system.members.forEach(({ actor }) => {
+      if ( actor ) actor.apps[this.id] = this;
+    });
   }
 
   /* -------------------------------------------- */
@@ -399,8 +403,9 @@ export default class GroupActorSheet extends MultiActorSheet {
     const { type, key } = target.dataset;
     if ( type !== "skill" ) return;
     const { uuid } = target.closest("[data-uuid]")?.dataset ?? {};
+    const { pace } = this.actor.system.getTravelPace();
     const actor = await fromUuid(uuid);
-    actor?.rollSkill({ event, skill: key, pace: this.actor.system.attributes.travel.pace });
+    actor?.rollSkill({ event, pace, skill: key });
   }
 
   /* -------------------------------------------- */
--- a/module/data/actor/group.mjs
+++ b/module/data/actor/group.mjs
@@ -194,7 +194,7 @@ export default class GroupData extends GroupTemplate {
       writable: false
     });
     if ( !memberIds.has(this.primaryVehicle?.id) || (this.primaryVehicle?.type !== "vehicle") ) {
-      this.primaryVehicle = null;
+      Object.defineProperty(this, "primaryVehicle", { value: null });
     }
   }
 
@@ -249,17 +249,7 @@ export default class GroupData extends GroupTemplate {
   getTravelPace() {
     let { pace } = this.attributes.travel;
     const slowed = this.members.some(({ actor }) => {
-      if ( !actor?.system.isCreature ) return false;
-      const source = actor.system._source.attributes?.movement;
-      const { movement } = actor.system.attributes ?? {};
-      const { fromSpecies } = movement ?? {};
-      for ( const key of Object.keys(CONFIG.DND5E.movementTypes) ) {
-        const base = source?.[key] ?? fromSpecies?.[key];
-        if ( !base ) continue;
-        const current = movement?.[key];
-        if ( current <= (base / 2) ) return true;
-      }
-      return false;
+      return actor?.system.isCreature && actor?.system.attributes?.movement?.slowed;
     });
     const travelPaces = Object.keys(CONFIG.DND5E.travelPace);
     const slow = travelPaces.indexOf("slow");
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -328,7 +328,7 @@ export default class AttributesFields {
     encumbrance.max = encumbrance.thresholds.maximum;
     encumbrance.mod = (sizeMod * maximumMultiplier).toNearest(0.1);
     encumbrance.stops = {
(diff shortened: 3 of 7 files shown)


module/data/actor/group.mjs
```diff
@@ -137,17 +149,20 @@ export default class GroupData extends GroupTemplate {
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
   static #migrateTravel(source) {
-    if ( source.attributes?.travel || !source.attributes?.movement ) return;
+    if ( !source.attributes?.movement ) return;
```
Fyorl, line 154: Fixed an issue with the migration here. We cannot bail if we detect `attributes.travel` because a group might have updated one part of the `travel` field, but not others, which remain un-migrated on disk.

templates/actors/group/member.hbs
```diff
@@ -1,5 +1,9 @@
 <li class="member" data-uuid="{{ uuid }}">
     <div class="vitals card" {{#if underlay}}style="--underlay: {{ underlay }}"{{/if}}>
+        {{#if isPrimaryVehicle}}
```
arbron, line 3: This ends up not being visible if you use a square portrait on the vehicle (like the artwork set for the PHB vehicles):

Screenshot 2025-10-17 at 15 16 32

After moving it in front, it is still difficult to make out against the artwork:

Screenshot 2025-10-17 at 15 17 32

module/data/actor/group.mjs
```diff
@@ -22,32 +22,44 @@ const { ArrayField, ForeignDocumentField, NumberField, SchemaField, StringField
...
       details: new SchemaField({
         xp: new SchemaField({
           value: new NumberField({ integer: true, min: 0, label: "DND5E.ExperiencePoints.Current" })
         }, { label: "DND5E.ExperiencePoints.Label" })
-      }, { label: "DND5E.Details" })
+      }, { label: "DND5E.Details" }),
+      members: new ArrayField(new SchemaField({
+        actor: new ForeignDocumentField(foundry.documents.BaseActor)
+      }), { label: "DND5E.GroupMembers" }),
+      primaryVehicle: new ForeignDocumentField(foundry.documents.BaseActor)
```
arbron, line 64: Removing a member doesn't unset the primary vehicle.

module/applications/actor/group-sheet.mjs
```diff
@@ -250,14 +300,41 @@ export default class GroupActorSheet extends MultiActorSheet {
...
+    super._onClose(options);
+    this.actor.system.members.forEach(({ actor }) => delete actor.apps[this.id]);
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _onFirstRender(context, options) {
+    await super._onFirstRender(context, options);
+    this.actor.system.members.forEach(({ actor }) => actor.apps[this.id] = this);
```
arbron: If the actor for a member is deleted from the world but not removed from members then this throws an error because `actor` is null (same issue in `_onClose`). This only persists until the data is re-prepared for the vehicle.

module/data/actor/group.mjs
```diff
@@ -223,6 +240,50 @@ export default class GroupData extends GroupTemplate {
...
+      const { movement } = actor.system.attributes ?? {};
+      const { fromSpecies } = movement ?? {};
+      for ( const key of Object.keys(CONFIG.DND5E.movementTypes) ) {
+        const base = source?.[key] ?? fromSpecies?.[key];
+        if ( !base ) continue;
+        const current = movement?.[key];
+        if ( current <= (base / 2) ) return true;
+      }
+      return false;
+    });
```
arbron, line 253: Looks like you hit the slowed changes that I did in another PR: https://github.com/foundryvtt/dnd5e/pull/6215 :)

One thing I did in that PR is just added a `slowed` boolean to `attributes.movement` that is added as part of data preparation on the members. That looks to be a bit simpler than this because we have the original & modified speeds available at that point:

```javascript
      // In `prepareMovement` in attributes.mjs, right before `this.attributes.movement[type] = speed;`
      if ( (type === "walk") && this.attributes.movement.walk ) {
        this.attributes.movement.slowed = speed <= Math.floor(this.attributes.movement.walk / 2);
      }
```

One other thing I did in that PR is added an icon after the speed of the slowed member in the members list to make it clear who was slowing the group:

```html
	                <span class="value">
                    {{ system.attributes.movement.walk }}
                    {{#if system.attributes.movement.slowed}}
                    <i class="fa-solid fa-triangle-exclamation" data-tooltip
                       aria-label="{{ localize 'DND5E.Travel.Slowed' }}"></i>
                    {{/if}}
                </span>
```

module/data/actor/templates/attributes.mjs
```diff
@@ -328,8 +328,12 @@ export default class AttributesFields {
     encumbrance.max = encumbrance.thresholds.maximum;
     encumbrance.mod = (sizeMod * maximumMultiplier).toNearest(0.1);
     encumbrance.stops = {
-      encumbered: Math.clamp((encumbrance.thresholds.encumbered * 100) / encumbrance.max, 0, 100),
-      heavilyEncumbered: Math.clamp((encumbrance.thresholds.heavilyEncumbered * 100) / encumbrance.max, 0, 100)
+      encumbered: Number.isFinite(encumbrance.max )
```
arbron:
```suggestion
      encumbered: Number.isFinite(encumbrance.max)
```

module/data/actor/group.mjs
Fyorl: Yeah, looks like I missed that PR.

> One thing I did in that PR is just added a `slowed` boolean to `attributes.movement` that is added as part of data preparation on the members. That looks to be a bit simpler than this because we have the original & modified speeds available at that point:

I agree it's simpler and would probably go with that implementation, though the current one in this PR does not suffer from this limitation:

>Currently this only takes into account speed reductions from status effects, not from any active effects that might be reducing their speed.

I would expect that status effects are the only modifiers relevant during travel (exhaustion primarily), but maybe there are some I'm not considering. The *slow* spell would cut a character's speed via AE, but that's not going to last long enough to be relevant to travel pace, for example. If we can't think of any, then maybe we go with the `slowed` implementation.

module/data/actor/group.mjs
Fyorl: It doesn't need to

templates/actors/group/member.hbs
Fyorl: Vehicles don't have classes, so I guess I can move this to the right.

module/data/actor/group.mjs
arbron: But if I have a vehicle set as primary, and then say "Remove Member", the travel pace and inventory remain linked to that vehicle and there is no way to remove it without re-adding the vehicle or adding a new vehicle and setting it as primary.

Fyorl: Ah, there is code to handle that in `GroupData#prepareBaseData`, but I see it's not working because of some interaction with the `ForeignDocumentField`.

module/data/actor/group.mjs
Fyorl: I think I was able to move the `slowed` calculation into actor preparation but also retain the ability to account for AE speed modifications. Let me know if that looks correct or not to you.

arbron: Looks like it will work.

PR #6320, opened 2025-10-20
[#6138] Allow setting disposition of summoned tokens

### LINKED ISSUE #6138: Summons controllability permissions and disposition
Feature suggestion: Make summoned companion actors (i.e imps) have the same controllability as the player that summoned them, have the player be the owner by default and make the token disposition friendly if summoned by a player.

Currently if a player summons an imp lets say, the token by default is not owned by the player and the token is bought in as an hostile

arbron:
Adds a "Disposition" option to summons changes to set that value on the summoned token. Modifies several SRD spells to have that value set as appropriate.

Closes #6138

### DIFF
--- a/module/data/activity/summon-data.mjs
+++ b/module/data/activity/summon-data.mjs
@@ -130,4 +130,14 @@ export default class BaseSummonActivityData extends BaseActivityData {
       }
     });
   }
+
+  /* -------------------------------------------- */
+  /*  Socket Event Handlers                       */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _preCreate(data) {
+    super._preCreate(data);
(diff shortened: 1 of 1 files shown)


module/data/activity/summon-data.mjs
```diff
@@ -59,6 +60,7 @@ export default class SummonActivityData extends BaseActivityData {
       }),
       creatureSizes: new SetField(new StringField()),
       creatureTypes: new SetField(new StringField()),
+      disposition: new NumberField({ nullable: true, choices: Object.values(CONST.TOKEN_DISPOSITIONS) }),
```
Fyorl, line 32: Would like to have `initial: 1`, since that seems like a pretty sensible default for most summons. But that would mean all existing summons suddenly are summoned with friendly disposition after updating to 5.2 which is not ideal.

Could we put it in `SummonActivity#_preCreate` maybe?

Fyorl (approved):

PR #6325, opened 2025-10-21
[#6322] Add support for "All Damage" in resistances/etc.

### LINKED ISSUE #6322: Add support for ALL in damage resistances, immunities, and vulnerabilities
Treats the character as if they have resistance/etc in all damage types. Will require some changes to damage calculation and the damage application tray. If a character has resistance/etc to all damage, then their sheet shouldn't also show individual damage types.

arbron:
Adds an "All Damage" option in resistances, immunities, and vulnerabilties that supersedes any other damage type entered in those configs. When the actor is prepared the system will clear any other damage types in those values if "All Damage" is set. It will also check and remove any damage types in reistances that are also included in immunities to help with issues in the damage application tray and to clean up the sheet.

Changes had to made to the damage calculation system to separate the `active` state into type-specific activations VS. "All" activations. This allows a new icon to be displayed in the tray to control resistances/etc. for "All Damage". Because this change made the `hasEffect` and `ignore` functions bigger and more confusing, this PR moves them into separate private methods so they can be better documented.

Also adds "All Damage" option in damage modification, but this is applied in addition to any damage-specific modification. The process of applying modifications has been changed to take into account that a damage operation may have multiple entries with the same type. Previously, the system would apply the same damage modification to each separate entry of a damage type. It will now keep a running total of the modification applied to avoid over applying.

Closes #6322

### DIFF
--- a/module/data/actor/fields/damage-trait-field.mjs
+++ b/module/data/actor/fields/damage-trait-field.mjs
@@ -8,7 +8,7 @@ export default class DamageTraitField extends SimpleTraitField {
   constructor(fields={}, { initialBypasses=[], ...options }={}) {
     super({
       bypasses: new SetField(new StringField(), {
-        label: "DND5E.DamagePhysicalBypass", hint: "DND5E.DAMAGE.PhysicalBypass.Hint", initial: initialBypasses
+        label: "DND5E.DAMAGE.PhysicalBypass.Label", hint: "DND5E.DAMAGE.PhysicalBypass.Hint", initial: initialBypasses
       })
     }, options);
   }
--- a/module/data/actor/templates/traits.mjs
+++ b/module/data/actor/templates/traits.mjs
@@ -27,7 +27,7 @@ export default class TraitsField {
       dm: new SchemaField({
         amount: new MappingField(new FormulaField({ deterministic: true }), { label: "DND5E.DamMod" }),
         bypasses: new SetField(new StringField(), {
-          label: "DND5E.DamagePhysicalBypass", hint: "DND5E.DAMAGE.PhysicalBypass.Hint"
+          label: "DND5E.DAMAGE.PhysicalBypass.Label", hint: "DND5E.DAMAGE.PhysicalBypass.Hint"
         })
       }),
       ci: new SimpleTraitField({}, { label: "DND5E.ConImm" })
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -31,7 +31,8 @@ import * as Trait from "./trait.mjs";
  *   SkillToolRollDialogConfiguration, SkillToolRollProcessConfiguration
  * } from "../../dice/_types.mjs";
  * @import {
- *   DamageApplicationOptions, DamageDescription, DamageSummary, RestConfiguration, RestResult, SpellcastingDescription
+ *   DamageAffectCategory, DamageApplicationOptions, DamageDescription, DamageSummary,
+ *   RestConfiguration, RestResult, SpellcastingDescription
  * } from "../_types.mjs";
  */
 
@@ -862,7 +863,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       }
 
       // Apply damage modification
-      if ( !dm.bypasses.intersection(d.properties).size ) {
+      if ( !dm.bypasses?.intersection(d.properties).size ) {
         applyModification(d);
         applyModification(d, "ALL");
       }
@@ -919,7 +920,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /**
    * Determine whether a specific type of change to a damage value will have an effect.
-   * @param {string} category                                Type of change that should be considered (e.g. resistance).
+   * @param {DamageAffectCategory} category                  Type of change that should be considered.
    * @param {DamageDescription|string} damage                Damage description to consider or a specific type.
    * @param {object} [options={}]
    * @param {DamageApplicationOptions} [options.options={}]  Damage application options.
@@ -936,7 +937,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       }
       return true;
     };
-    const type = foundry.utils.getType(damage) === "string" ? damage : damage.type;
+    const type = typeof damage === "string" ? damage : damage.type;
 
     // If category is resistance, check for downgraded immunities
     if ( category === "resistance" ) {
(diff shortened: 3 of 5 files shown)


module/documents/actor/actor.mjs
```diff
@@ -918,6 +913,85 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    if ( (options.ignore?.[category] === true) || (options.ignore?.[category]?.has?.(type)) ) return true;
+
+    // When downgrading, always ignore immunities unless `skipDowngrade` option is set
+    if ( (category === "immunity") && downgrade(type) && !skipDowngrade ) return true;
+
+    // When downgrading, resistances should be decided by whether immunity is applied
+    if ( (category === "resistance") && downgrade(type) && !this.#changeHasEffect("immunity", type) ) return true;
+
+    return false;
+  }
```
krbz999, line 996: You weren't kidding, these are ever so slightly more complicated!

module/applications/components/damage-application.mjs
```diff
@@ -131,10 +131,14 @@ export default class DamageApplicationElement extends TargetedApplicationMixin(C
     for ( const [change, values] of Object.entries(active) ) {
       if ( foundry.utils.getType(values) !== "Set" ) continue;
       for ( const type of values ) {
-        const config = CONFIG.DND5E.damageTypes[type] ?? CONFIG.DND5E.healingTypes[type];
-        if ( !config ) continue;
-        const data = { type, change, icon: config.icon };
-        types.push(data);
+        if ( type === "ALL" ) {
+          types.push({ type, change, icon: "systems/dnd5e/icons/svg/damage/all.svg" });
```
Fyorl, line 135: I suppose it's possible that a creature might have resistance to all damage, and a character has a feature that ignores resistance to, e.g. necrotic. In that case, an Eldritch Blast + Hex would halve the force damage but deal full necrotic damage, but the DM has no way to input that via the damage tray since they can only toggle off resistance to all damage, which also gives the wrong result.

It's pretty niche so I don't think we should do anything about it now, but I wanted to note it so we can maybe think of some potential solutions.

module/data/actor/fields/damage-trait-field.mjs
```diff
@@ -8,7 +8,7 @@ export default class DamageTraitField extends SimpleTraitField {
   constructor(fields={}, { initialBypasses=[], ...options }={}) {
     super({
       bypasses: new SetField(new StringField(), {
-        label: "DND5E.DamagePhysicalBypass", hint: "DND5E.DamagePhysicalBypassHint", initial: initialBypasses
+        label: "DND5E.DamagePhysicalBypass", hint: "DND5E.DAMAGE.PhysicalBypass.Hint", initial: initialBypasses
```
Fyorl:
```suggestion
        label: "DND5E.DAMAGE.PhysicalBypass.Label", hint: "DND5E.DAMAGE.PhysicalBypass.Hint", initial: initialBypasses
```

module/data/actor/templates/traits.mjs
```diff
@@ -27,7 +27,7 @@ export default class TraitsField {
       dm: new SchemaField({
         amount: new MappingField(new FormulaField({ deterministic: true }), { label: "DND5E.DamMod" }),
         bypasses: new SetField(new StringField(), {
-          label: "DND5E.DamagePhysicalBypass", hint: "DND5E.DamagePhysicalBypassHint"
+          label: "DND5E.DamagePhysicalBypass", hint: "DND5E.DAMAGE.PhysicalBypass.Hint"
```
Fyorl:
```suggestion
          label: "DND5E.DAMAGE.PhysicalBypass.Label", hint: "DND5E.DAMAGE.PhysicalBypass.Hint"
```

module/documents/actor/actor.mjs
```diff
@@ -830,65 +830,53 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
 
-      // Apply type-specific damage reduction
-      if ( !ignore("modification", d.type) && traits.dm?.amount[d.type]
-        && !traits.dm.bypasses.intersection(d.properties).size ) {
-        const modification = simplifyBonus(traits.dm.amount[d.type], rollData);
-        if ( Math.sign(d.value) !== Math.sign(d.value + modification) ) d.value = 0;
-        else d.value += modification;
-        d.active.modification = true;
+      // Apply damage modification
+      if ( !dm.bypasses.intersection(d.properties).size ) {
```
Fyorl: `dm` might be an empty object.


```suggestion
      if ( !dm.bypasses?.intersection(d.properties).size ) {
```

module/documents/actor/actor.mjs
```diff
@@ -929,6 +917,85 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /* -------------------------------------------- */
 
+  /**
+   * Determine whether a specific type of change to a damage value will have an effect.
+   * @param {string} category                                Type of change that should be considered (e.g. resistance).
```
Fyorl: I guess we should have a `@typedef` for this.

module/documents/actor/actor.mjs
```diff
@@ -929,6 +917,85 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    const config = this.system.traits?.[`d${category.slice(0, 1)}`];
+    const downgrade = type => options.downgrade === true || options.downgrade?.has?.(type);
+    const setActive = type => {
+      if ( damage.active ) {
+        damage.active[type] ??= {};
+        damage.active[type][category] = true;
+      }
+      return true;
+    };
+    const type = foundry.utils.getType(damage) === "string" ? damage : damage.type;
```
Fyorl: `typeof` is faster.

```suggestion
    const type = typeof damage === "string" ? damage : damage.type;
```

module/documents/actor/actor.mjs
```diff
@@ -929,6 +917,85 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+        return setActive("all");
+      }
+      if ( downgrade(type) && this.#changeHasEffect("immunity", type, { skipDowngrade: true }) ) {
+        return setActive("type");
+      }
+    }
+
+    // If damage type is physical and bypass present in properties, skip further checks
+    if ( CONFIG.DND5E.damageTypes[type]?.isPhysical && damage.properties?.size
+      && config.bypasses?.intersection(damage.properties)?.size ) return false;
```
```suggestion
      && config?.bypasses?.intersection(damage.properties)?.size ) return false;
```

Fyorl (approved):

Issue #6338, opened 2025-10-29
Reference for `Falling` has no `&reference` tag for the Prone condition


Zanderaf:
See below, would probably be nice to have a `&reference[prone]` in there 
<img width="954" height="313" alt="Image" src="https://github.com/user-attachments/assets/a9462ad2-cbf7-41a8-90fc-5225c07da2d4" />

Sayshal:
Can be fixed via #6352

Issue #6291, opened 2025-10-10
Make use of the `lookup` enricher's fallback in Monster Features 


Dewi-Jones:
Would be nice to utilize the fallback functionality of the `lookup` enricher in the Monster Features Compendium.
Example `[[lookup @name lowercase]]` vs `[[lookup @name lowercase]]{creature}`

<img width="482" height="83" alt="Image" src="https://github.com/user-attachments/assets/33011bca-9092-47e0-95af-b3074a646160" />

Makes it simpler to read the features in the compendium without sacrificing any functionality.

Sayshal:
I'm going through the compendium content tags and making PRs for any easy task I can help get done. Is there any reason this shouldn't be 'creature' in some instances? Seems like a fairly easy find/replace otherwise. @Dewi-Jones 

Edit: Can be fixed via #6353

Issue #6137, opened 2025-09-03
Disciplined Survivor doesn't give proficiency in all saving throws


Dewi-Jones:
The level 14 Feature Disciplined Survivor of the 2024 Monk assumes Proficiency in Strength and Dexterity is present, which is cases of Multiclassing might not be the case. Also it should give proficiency in death saves, as they are considered saving throws.

<img width="334" height="234" alt="Image" src="https://github.com/user-attachments/assets/060e9564-91c1-4414-b9eb-e46f89ffbbd9" />

Dewi-Jones:
Should probably just give the `diamondSoul` Flag

henry-malinowski:
It would be better to support Death Save prof for homebrew subclass or feats that grant Death Save prof. Diamond Soul wouldn't work for them.

Sayshal:
Can be fixed via #6354

PR #6351, opened 2025-10-30
[TASK] Pack fixes

Sayshal:
Going through all compendium fixes/minor corretions that I think would be easy to implement without disturbing or breaking anything. If any of these are deemed too destructive I can remove them individually.

- Add `&Reference` to Falling Rule. #6338
- Add `creature` fallback in lookup enrichers. #6291 
- Added `str` and `dex` save proficiencies to guard against multiclass that don't grant these proficiencies.Added passive effect to additionally grant `adv` on death saves. #6137

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2479,6 +2479,7 @@
 "DND5E.Loot.Junk": "Junk",
 "DND5E.Loot.Material": "Material",
 "DND5E.Loot.Resource": "Resource",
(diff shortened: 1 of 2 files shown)


arbron:
I'd recommend slitting these up into separate PRs because we might not want to apply all of these changes.

Sayshal:
> I'd recommend slitting these up into separate PRs because we might not want to apply all of these changes.

Will do this evening.

PR #6353, opened 2025-10-30
Add `creature` fallback in lookup enrichers

### LINKED ISSUE #6291: Make use of the `lookup` enricher's fallback in Monster Features

Sayshal:
Closes #6291

### DIFF
--- a/packs/_source/actors24/aberration/aboleth.yml
+++ b/packs/_source/actors24/aberration/aboleth.yml
@@ -959,7 +959,7 @@ items:
           activity=clFFG8uDBSFzxIo8]] in a [[lookup @target.template.size
           activity=clFFG8uDBSFzxIo8]]-[[lookup @target.template.units
           activity=clFFG8uDBSFzxIo8]] Emanation originating from the [[lookup
-          @name lowercase]] at the end of the [[lookup @name lowercase]]{monster}'s
+          @name lowercase]]{monster} at the end of the [[lookup @name lowercase]]{monster}'s
           turn.</p><p><em>Failure:</em> The target is cursed. Until the curse
           ends, the target's skin becomes slimy, the target can breathe air and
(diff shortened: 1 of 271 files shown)


arbron:
I'm wondering if we want to change this to be `{monster}` rather than `{creature}` because monster features regularly use creature to mean the target of the ability, so using creature as the fallback may result in some confusing wording.

See Swallow:
"If the creature takes 30 damage or more on a single turn from the swallowed creature, the creature must succeed…"

vs

"If the monster takes 30 damage or more on a single turn from the swallowed creature, the monster must succeed…"

Sayshal:
> I'm wondering if we want to change this to be `{monster}` rather than `{creature}` because monster features regularly use creature to mean the target of the ability, so using creature as the fallback may result in some confusing wording.

That makes sense, yeah. I can submit an update in a little bit here.

Sayshal:
> Do the features on the monsters themselves need this fallback? I would think probably not since they should always resolve. But I suppose if new monsters are created by dragging the monster features then they would end up with these fallbacks anyway.
> 
> At the very least it would reduce the size of this diff I suppose.

I don't feel strongly either way and can make the revisions if that's what we want. Having the fallback is nice so when imported and then dragged to an actor it's more readable.


Either way - not ready for merge yet as I found a few more instances of lookups without fallbacks.

Sayshal:
Alright. 474 more adjustments - my regex searching wasn't catching the way YAML wraps lines. ALL monsters & monster features now have a fallback. Features I think is a good choice as mentioned above.

PR #6354, opened 2025-10-30
Fix 2024 Monk (Disciplined Survivor)

### LINKED ISSUE #6137: Disciplined Survivor doesn't give proficiency in all saving throws

Sayshal:
Added `str` and `dex` save proficiencies to guard against multiclass that don't grant these proficiencies. Added passive effect to additionally grant `adv` on death saves. Closes #6137

### DIFF
--- a/packs/_source/classes24/monk/class-features/disciplined-survivor.yml
+++ b/packs/_source/classes24/monk/class-features/disciplined-survivor.yml
@@ -94,9 +94,9 @@ effects:
     type: base
     system: {}
     changes:
-      - key: system.attributes.death.roll.mode
-        mode: 2
-        value: '1'
+      - key: flags.dnd5e.diamondSoul
+        mode: 5
+        value: 'true'
         priority: null
     disabled: false
     duration:
--- a/packs/_source/classes24/monk/monk.yml
+++ b/packs/_source/classes24/monk/monk.yml
@@ -399,24 +399,6 @@ system:
       value: {}
       title: Unarmored Movement
       hint: ''
-    - _id: wOFoztcLEFPTTmK7
-      type: Trait
-      configuration:
-        mode: default
-        allowReplacements: false
-        grants:
-          - saves:str
-          - saves:dex
-          - saves:con
-          - saves:int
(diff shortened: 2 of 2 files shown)


packs/_source/classes24/monk/class-features/disciplined-survivor.yml
```diff
@@ -86,7 +86,50 @@ system:
...
-effects: []
+effects:
+  - name: Disciplined Survivor
+    img: icons/skills/wounds/injury-pain-body-orange.webp
+    origin: Compendium.dnd5e.classes24.Item.phbmnkDiscipline
+    _id: Js4phYMTL714onqn
+    type: base
+    system: {}
+    changes:
+      - key: system.attributes.death.roll.mode
```
krbz999: Advantage on death saves?

Sayshal: I was going based off [this](https://x.com/JeremyECrawford/status/688437911725805569) and the comments on the original issue. Fine with reverting if the consensus is death saving throws aren't saving throws. Diamoned Soul has virtually the same wording as Disciplined Survivor:

> Additionally, whenever you make a saving throw and fail, you can spend 1 ki point to reroll it and take the second result.

> Additionally, whenever you make a saving throw and fail, you can expend 1 Focus Point to reroll it, and you must use the new roll.

krbz999: Proficiency is not advantage.

krbz999: This item has no business granting an advancement, it should just be an effect with the Diamond Soul flag.

Sayshal: Will fix.

PR #6359, opened 2025-10-31
[#6238, #6329, #6349] Actor transformation fixes

### LINKED ISSUE #6238: TokenDocument modifications are lost when reverting transformation on unlinked tokens
Image
^Before Transformation

Image
^After Transformation

Image
^Restore Transformation

When transform the token to another form with Appearance Only options, the Token data did not retain the Token Identity data, which replace the Token Name, Display Name settings, etc.
When Restore Transformation, the Token Identity data still did not return and also the Token image path was different than before Transformation but rather different one (due to that the Prototype Token has a Randomize Wildcard Images enabled).

### LINKED ISSUE #6329: Transform activity and wildcard
Hi ! When I use the transform activity on a monster with a base token with wildcard, when I revert the transformation, the token don't have the same image than before the transformation, cause wildcard making his job. Is it possible in this case to force the token image to be the same than before ?
Tried without any module.

### LINKED ISSUE #6349: Reverting a 2+-deep transformation on an unlinked token's delta is only semi-functional
Specifically, with Actor 1, Actor 2, and Actor 3, transforming 1 -> 2 -> 3 leads to, in order:
- Actor 1
- Actor 1 (Actor 2)
- Actor 1 (Actor 2 (Actor 3))
All good. Then upon reversion, the delta becomes equivalent to that of `Actor 1 (Actor 2)`, while the token is that of `Actor 1`. The actor thinks it has a remaining reversion (because it ought to) but reverting does nothing. This is because `flags.dnd5e.previousActorData` gets clobbered in `Actor5e#transformInto` for unlinked tokens, so the _previous_ transformation's data is the only thing ever saved. The token appears properly because _that_ data is generated from the prototype of the base actor.
Now, what exactly the bug is depends on what the intended functionality is. For a linked actor, reverting transformation always reverts to the _original_ actor, there's no such thing as going back one layer of transformation. If this is how it should work for unlinked as well, then it's just a matter of
```diff
+ if ( !this.token.flags.dnd5e?.previousActorData ) {
    const previousActorData = this.token.delta.toObject();
    foundry.utils.setProperty(tokenData, "flags.dnd5e.previousActorData", previousActorData);
+ }
```
in `Actor5e#transformInto`. If the desired is to support multi-stage reversions, then something more in-depth would be needed.

Aioros:
This includes:

a) Fix for reverting multi-stage transformation of unlinked tokens (closes #6349, thanks @roth-michael).
b) Reviewed logic to clean up flags on revert. To describe it shortly:
  - the `isOriginalActor` bool was always false;
  - most flags (but not all) were deleted anyway when their document was;
  - flags on polymorphed world actors were not deleted when restored by a player.

c) Added a new `previousTokenData` flag to store the original `texture.src` to be restored on revert instead of using the prototype one (closes #6329).
d) Use own token's data for `bar1`, `bar2`, `displayBars`, `displayName`, `disposition`, `rotation`, `elevation` instead of the prototype's. Also store the original token name in the same `previousTokenData` flag to be restored on revert. Closes #6238, _but_ it's worth noting that the issue actually exists for linked tokens too, and an approach like this one could not work in that case (if you have multiple linked tokens for the same actor with different token names/displayBars/etc, which properties do you use?). I do think that it's a much rarer situation though.

I'm aware that the transform functions are delicate. I'll continue testing on my end, but let me know if you notice any issue.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -2835,7 +2835,9 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     // Token appearance updates
     const tokenPropsFromSource = ["width", "height", "alpha", "lockRotation", "ring"];
     const tokenTexturePropsFromSource = ["offsetX", "offsetY", "scaleX", "scaleY", "src", "tint"];
-    const tokenPropsFromSelf = ["bar1", "bar2", "displayBars", "displayName", "disposition", "rotation", "elevation"];
+    const tokenPropsFromSelf = [
+      "bar1", "bar2", "displayBars", "displayName", "disposition", "rotation", "elevation", "hidden"
+    ];
 
     for ( const k of tokenPropsFromSource ) {
       d.prototypeToken[k] = sourceData.prototypeToken[k];
@@ -3039,7 +3041,6 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     if ( this.isToken ) {
       const tokenData = d.prototypeToken;
       delete d.prototypeToken;
-      tokenData.hidden = this.token.hidden;
       for ( const k of tokenPropsFromSelf ) {
         tokenData[k] = this.token[k];
       }
@@ -3101,7 +3102,6 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       newTokenData._id = t.id;
       newTokenData.actorId = newActor.id;
       newTokenData.actorLink = true;
-      newTokenData.hidden = t.document.hidden;
       for ( const k of tokenPropsFromSelf ) {
         newTokenData[k] = t.document[k];
       }


Fyorl:
Could you leave a comment on #6238 and #6329 so that those issues can be assigned to you?

module/documents/actor/actor.mjs
```diff
@@ -3175,11 +3184,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       await this.sheet.close();
       const token = await TokenDocument.implementation.create(tokenUpdate, { parent: this.token.parent, render: true });
       await this.token.delete({ replacements: { [this.token._id]: token.uuid } });
-      if ( isOriginalActor ) {
-        await this.unsetFlag("dnd5e", "isPolymorphed");
-        await this.unsetFlag("dnd5e", "previousActorIds");
-        await this.token.unsetFlag("dnd5e", "previousActorData");
-      }
+      await token.unsetFlag("dnd5e", "previousActorData");
```
Fyorl: Is it necessary to delete this here? Can we remove the flag as part of the `tokenUpdate` to avoid another round-trip?

module/documents/actor/actor.mjs
```diff
@@ -2833,6 +2833,8 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     d.system.attributes.ac.flat = source.system.attributes.ac.value; // Override AC
 
     // Token appearance updates
+    if (settings.keep.has("self") && this.isToken) d.prototypeToken.name = this.token.name;
```
```suggestion
    if ( settings.keep.has("self") && this.isToken ) d.prototypeToken.name = this.token.name;
```

module/documents/actor/actor.mjs
```diff
@@ -3201,27 +3206,31 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    } else {
+      // Remove the flags
+      const actorUpdates = polymorphedActorIds.map(p => {
+        return { _id: p, "flags.dnd5e": { "-=isPolymorphed": null, "-=previousActorIds": null } };
+      });
+      await Actor.implementation.updateDocuments(actorUpdates);
+
+      if ( isRendered ) {
+        this.sheet?.close();
+      }
```
```suggestion
      if ( isRendered ) this.sheet?.close();
```

module/documents/actor/actor.mjs
```diff
@@ -3201,27 +3206,31 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-        && game.actors?.get(id) // Actor still exists
-      ).concat([this.id]); // Add this id
-
-      await Actor.implementation.deleteDocuments(idsToDelete);
-    } else if ( isRendered ) {
-      this.sheet?.close();
+      await Actor.implementation.deleteDocuments(polymorphedActorIds);
+    } else {
+      // Remove the flags
+      const actorUpdates = polymorphedActorIds.map(p => {
```
Fyorl: I suppose it's technically possible ownership might have changed between transformation and reversion. Will need to check ownership before attempting to update these actors, otherwise it will error.

module/documents/actor/actor.mjs
```diff
@@ -2833,6 +2833,8 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     d.system.attributes.ac.flat = source.system.attributes.ac.value; // Override AC
 
     // Token appearance updates
+    if (settings.keep.has("self") && this.isToken) d.prototypeToken.name = this.token.name;
+
     for ( const k of ["width", "height", "alpha", "lockRotation"] ) {
```
Fyorl: While we're here:


```suggestion
    for ( const k of ["width", "height", "alpha", "lockRotation", "ring"] ) {
```

And then we can delete line 2844 (2842 originally).

module/documents/actor/actor.mjs
```diff
@@ -3039,8 +3041,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       tokenData.elevation = this.token.elevation;
       tokenData.hidden = this.token.hidden;
       tokenData.rotation = this.token.rotation;
-      const previousActorData = this.token.delta.toObject();
-      foundry.utils.setProperty(tokenData, "flags.dnd5e.previousActorData", previousActorData);
+      if ( !this.token.flags.dnd5e?.previousActorData ) {
+        const previousActorData = this.token.delta.toObject();
+        foundry.utils.setProperty(tokenData, "flags.dnd5e.previousActorData", previousActorData);
+        foundry.utils.setProperty(tokenData, "flags.dnd5e.previousTokenData.texture.src", this.token.texture.src);
+        foundry.utils.setProperty(tokenData, "flags.dnd5e.previousTokenData.name", this.token.name);
```
Fyorl: Looks like there are a few more things beyond just name and texture that get changed during transform: `width`, `height`, `texture.tint`, etc. Presumably we would want to restore those as well.

Fyorl (changes requested):

Aioros:
Just updated with a more organized approach to the token properties that we transform and restore, and all other suggested changes where still applicable. The new approach also has a more robust solution for #6238 when linked tokens are involved.

module/documents/actor/actor.mjs
```diff
@@ -2833,14 +2833,17 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     d.system.attributes.ac.flat = source.system.attributes.ac.value; // Override AC
 
     // Token appearance updates
-    for ( const k of ["width", "height", "alpha", "lockRotation"] ) {
+    const tokenPropsFromSource = ["width", "height", "alpha", "lockRotation", "ring"];
+    const tokenTexturePropsFromSource = ["offsetX", "offsetY", "scaleX", "scaleY", "src", "tint"];
+    const tokenPropsFromSelf = ["bar1", "bar2", "displayBars", "displayName", "disposition", "rotation", "elevation"];
```
Fyorl: If we add `hidden` to this list we can remove lines 3042 and 3104


```suggestion
    const tokenPropsFromSelf = [
      "bar1", "bar2", "displayBars", "displayName", "disposition", "rotation", "elevation", "hidden"
    ];
```

PR #6373, opened 2025-11-06
Consolidate Activity rendering. Add ActivityElement.

Fyorl:
A few related changes in this PR:
- Adjusts activities rendering so they resemble most other lists. It will allow us to place some useful columns for each activity row in the future. At the moment, only the activation cost is displayed.
  - Also adds a marker to indicate which activity is the 'primary' activity.
- Adjusts effects to display riders as subordinate to the effect they accompany.
- Adjusts activities to display riders as subordinate to the activity that has configured them as a rider.
- Adds `ActivityElement` which can be used to make activity rendering more portable in the future.
- Cleans up some various v1 code that we no longer need.

image

Since the activity/effect -> rider relationship is many:many, displaying the riders as subordinate is not a perfect fit. If a rider is associated with more than one activity/effect, it will render under all of them. I think it is an improvement to take them out of the normal flow since they do behave differently, but if it turns out that associating the same rider with multiple activities/effects is a common pattern, then it leads to a lot of redundancy in the display, so we may need to re-think this design.

### DIFF
--- a/templates/shared/activities.hbs
+++ b/templates/shared/activities.hbs
@@ -70,8 +70,14 @@
 <div class="item-detail item-controls activity-controls always-visible">
 
     {{#if @root.editable}}
+    {{!-- Edit --}}
+    <button type="button" class="unbutton control-button item-control item-action" data-action="editDocument"
+            data-tooltip aria-label="{{ localize "DND5E.ACTIVITY.Action.Edit" }}">
+        <i class="fa-solid fa-pen-to-square" inert></i>
+    </button>
+
     {{!-- Delete --}}
-    <button type="button" class="unbutton control-button" data-action="deleteDocument"
+    <button type="button" class="unbutton control-button item-control" data-action="deleteDocument"
             data-tooltip aria-label="{{ localize "DND5E.ACTIVITY.Action.Delete" }}">
         <i class="fas fa-trash" inert></i>
     </button>
--- a/lang/en.json
+++ b/lang/en.json
@@ -263,7 +263,8 @@
 "DND5E.ACTIVITY": {
   "Action": {
     "Create": "Create Activity",
-    "Delete": "Delete Activity"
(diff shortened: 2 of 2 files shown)


templates/shared/activities.hbs
```diff
@@ -0,0 +1,88 @@
...
+    </div>
+
+</div>
+
+{{!-- Controls --}}
+<div class="item-detail item-controls activity-controls always-visible">
+
+    {{#if @root.editable}}
+    {{!-- Delete --}}
+    <button type="button" class="unbutton control-button" data-action="deleteDocument"
```
arbron: Noticeably larger that the controls in the effects & advancement tabs:

buttons

Looks like it needs the `item-control` class:

```suggestion
    <button type="button" class="unbutton control-button item-control" data-action="deleteDocument"
```

templates/shared/activities.hbs
```diff
@@ -0,0 +1,88 @@
...
+        {{#if subtitle}}<span class="subtitle">{{ subtitle }}</span>{{/if}}
+    </div>
+
+</div>
+
+{{!-- Controls --}}
+<div class="item-detail item-controls activity-controls always-visible">
+
+    {{#if @root.editable}}
+    {{!-- Delete --}}
```
arbron, line 79: Would be good to have an edit button here for parity with the effects & advancement tabs:

```suggestion
    {{!-- Edit --}}
    <button type="button" class="unbutton control-button item-control item-action" data-action="editDocument"
            data-tooltip aria-label="{{ localize "DND5E.ACTIVITY.Action.Edit" }}">
        <i class="fa-solid fa-pen-to-square" inert></i>
    </button>

    {{!-- Delete --}}
```

(the "DND5E.ACTIVITY.Action.Edit" localization will need to be added)

arbron (approved):

PR #6377, opened 2025-11-07
[#6376] Redesign Roll Table sheet using dnd5e2 styling

### LINKED ISSUE #6376: Roll table sheet doesn't get proper styling for system enrichers
Image

arbron:
Implements a system version of the roll table sheet that uses the system's styling. This doesn't bother rewriting the HTML, but makes certain targeted changes on render to ensure the proper classes are added and the dnd5e checkboxes are used.

Roll Table Re-design

Core's edit/view toggle is replaced with an edit slider in the header like other 5e sheets. In order to make this easier, some of the system's mode toggle code was changed to be more similar to how core's sheet is implemented (adding `isEditMode` property and using a `changeMode` action rather than a direct listener. None of these should be breaking changes.

Some changes were also made to sheet styles to make them more widely available and item's `.item-image` header styling was made into a more generic `.document-image` to be usable across sheets.

Closes #6376

### DIFF
--- a/module/applications/roll-table-sheet.mjs
+++ b/module/applications/roll-table-sheet.mjs
@@ -25,7 +25,6 @@ export default class RollTableSheet5e extends ApplicationV2Mixin(RollTableSheet,
     this.element.querySelector(".sheet-header [data-action=changeMode]")?.remove();
     this.element.querySelectorAll("tbody .inline-control").forEach(c => c.classList.add("unbutton", "control-button"));
     this._replaceElements("input[type=checkbox]", "dnd5e-checkbox");
-    this._replaceElements('table img[src$=".svg"]', "dnd5e-icon");
   }
 
   /* -------------------------------------------- */
--- a/less/v2/sheets.less
+++ b/less/v2/sheets.less
@@ -291,13 +291,6 @@
     tbody {
       scrollbar-gutter: auto;
     }
-
-    td.image > dnd5e-icon {
-      display: block;
-      --icon-size: 32px;
-      block-size: 32px;
-      inline-size: 32px;
-    }
   }
 
   &.edit-mode table .controls { flex: 0 0 90px; }


module/applications/api/primary-sheet-mixin.mjs
Fyorl, line 11: `RollTable` is a primary document as well so I wonder if it shouldn't use `PrimarySheetMixin` rather than `ApplicationV2Mixin`. Not a big deal though.

module/applications/api/application-v2-mixin.mjs
```diff
@@ -14,7 +14,8 @@ const { HandlebarsApplicationMixin } = foundry.applications.api;
  * @mixin
  */
 export default function ApplicationV2Mixin(Base) {
-  class BaseApplication5e extends HandlebarsApplicationMixin(Base) {
+  const _BaseApplication5e = "PARTS" in Base ? Base : HandlebarsApplicationMixin(Base);
+  class BaseApplication5e extends _BaseApplication5e {
```
Fyorl: `ApplicationV2Mixin` is a function, so maybe it would be cleaner to do something like this:
```js
export default function ApplicationV2Mixin(Base, { handlebars=true }={}) {
  const _BaseApplication5e = handlebars ? HandlebarsApplicationMixin(Base) : Base;
  class BaseApplication5e extends _BaseApplication5e {
```

module/applications/roll-table-sheet.mjs
```diff
@@ -0,0 +1,33 @@
...
+    classes: ["titlebar"]
+  };
+
+  /* -------------------------------------------- */
+  /*  Life-Cycle Handlers                         */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _onRender(context, options) {
+    super._onRender(context, options);
```
Fyorl:
```suggestion
  async _onRender(context, options) {
    await super._onRender(context, options);
```
Notice we have a fair few `_onRender` overrides that are not `async` in the codebase. I think it was a change from v12 -> v13.

module/applications/api/primary-sheet-mixin.mjs
arbron: I don't think it has anything we need, `PrimarySheetMixin` gives us tab handling, sub-document editing/creation, and the source button, none of which roll tables use, as well as a conflicting implementation of edit/play mode.

arbron:
> These row backgrounds don't go to the end of the row on the right:
> 
> Image

This is the result of `scrollbar-gutter: stable` on the scrollable table. I swapped it over to `auto` since I'm not sure how much of a concern that is here.

> Can we use `dnd5e-icon`s for these SVGs so that we can give them more appropriate colours for dark mode?
> 
> Image

I made this swap, but it doesn't help with the default icons because the core SVG that is used doesn't have the necessary `--icon-fill` variable. I took a look and `BaseTableResult` doesn't have a place for setting a default icon, so we can't swap it that way. Perhaps we could have a system migration to swap any usage of the core D20 icon with one of our own so it can be colored properly.

arbron:
Hmm, swapping to `dnd5e-icon` also means that `editImage` doesn't work because it deliberately throw an error when used on anything but an `` element.

PR #6379, opened 2025-11-08
[#4214] Keybinds: Open `CompendiumBrowser`, Toggle Sheet Lock

### LINKED ISSUE #4214: Add key toggle for edit mode
Hello!

It would be amazing if we could have a keybind that turns edit mode on/off for the currently focused item.
It was bearable on the main character sheet because you would not make so many adjustments at once, but It is a bit time consuming on items since you have to turn it on for every single item you want to edit.

Or at least please make it so if an item is opened from a character sheet on edit mode then the item also shows on edit mode.
I know this won't matter to people whom use the official source material that foundry provides, but for people that homebrew and whose language is not english, editing item names is a common occurrence where the edit button just slows things down rather than keeping it clean.

Sayshal:
This PR adds 2 keybinds:

1. **Shift + B**: Opens Compendium Browser in default state.
2. **L**: While in an application with a sheet MODE toggle (Edit/View), toggles that status.

Closes #4214 

Open to discussion on keys.

### DIFF
--- a/module/settings.mjs
+++ b/module/settings.mjs
@@ -44,10 +44,11 @@ export function registerSystemKeybindings() {
 
   game.keybindings.register("dnd5e", "toggleSheetMode", {
     name: "KEYBINDINGS.DND5E.ToggleSheetMode",
-    editable: [{ key: "KeyL" }],
+    editable: [{ key: "KeyE", modifiers: ["Shift"] }],
     onDown: () => {
       const app = ui.activeWindow;
-      if ( app?.rendered && app.changeMode ) app.changeMode();
+      if ( !app?.rendered || !app.changeMode || !app.isEditable ) return false;
+      app.changeMode();
       return true;
     }
   });
@@ -62,6 +63,18 @@ export function registerSystemKeybindings() {
       else new CompendiumBrowser().render({ force: true });
       return true;
     }
+  });
+
+  // TODO: Workaround for foundryvtt#14564
+  Hooks.on("openDetachedWindow", (id, win) => {
+    const { KeyboardManager } = foundry.helpers.interaction;
+    const forward = up => event => {
+      const el = win.document.activeElement;
+      if ( el && (["INPUT", "SELECT", "TEXTAREA"].includes(el.tagName) || el.isContentEditable) ) return;
+      game.keyboard._processKeyboardContext(KeyboardManager.getKeyboardEventContext(event, up), { force: true });
+    };
+    win.addEventListener("keydown", forward(false));
+    win.addEventListener("keyup", forward(true));
   });
 }
 
--- a/module/settings.mjs
+++ b/module/settings.mjs
@@ -2,7 +2,6 @@ import CompendiumBrowser from "./applications/compendium-browser.mjs";
 import BastionSettingsConfig from "./applications/settings/bastion-settings.mjs";
 import CalendarSettingsConfig from "./applications/settings/calendar-settings.mjs";
 import CombatSettingsConfig from "./applications/settings/combat-settings.mjs";
-import CompendiumBrowser from "./applications/compendium-browser.mjs";
 import CompendiumBrowserSettingsConfig from "./applications/settings/compendium-browser-settings.mjs";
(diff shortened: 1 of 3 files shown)


module/settings.mjs
```diff
@@ -2,6 +2,7 @@ import CompendiumBrowser from "./applications/compendium-browser.mjs";
 import BastionSettingsConfig from "./applications/settings/bastion-settings.mjs";
 import CalendarSettingsConfig from "./applications/settings/calendar-settings.mjs";
 import CombatSettingsConfig from "./applications/settings/combat-settings.mjs";
+import CompendiumBrowser from "./applications/compendium-browser.mjs";
```
arbron: This is already imported at the top.


```suggestion
```

module/settings.mjs
```diff
@@ -41,6 +42,32 @@ export function registerSystemKeybindings() {
...
+  game.keybindings.register("dnd5e", "toggleSheetMode", {
+    name: "KEYBINDINGS.DND5E.ToggleSheetMode",
+    editable: [{ key: "KeyL" }],
+    onDown: () => {
+      const app = ui.activeWindow;
+      if ( app?.rendered && app.constructor?.MODES && app._mode !== undefined ) {
+        const { MODES } = app.constructor;
+        app._mode = app._mode === MODES.EDIT ? MODES.PLAY : MODES.EDIT;
+        app.render();
+      }
```
arbron: There is an issue with this method in that it doesn't properly update the tooltip/aria label for the play/edit toggle. The `_onChangeSheetMode` method currently handles this, but it isn't really designed to be called externally. We should instead probably introduce a new `changeMode` public method onto `PrimaryDocumentMixin` and let that handle the label updates:

```javascript
// `applications/api/primary-sheet-mixin.mjs` line 315
    /**
     * Handle changing the sheet mode.
     * @this {PrimarySheet5e}
     * @param {Event} event         Triggering click event.
     * @param {HTMLElement} target  Button that was clicked.
     */
    static #changeMode(event, target) {
      this.changeMode();
    }

    /**
     * Handle the user toggling the sheet mode.
     * @param {Event} event         Triggering click event.
     * @param {HTMLElement} target  Button that was clicked.
     * @protected
     */
    async _onChangeSheetMode(event, target=event.currentTarget) {
        "The `_onChangeSheetMode` method has been moved to `changeMode`.",
        { since: "DnD5e 6.0", until: "DnD5e 6.2" }
      );
      this.changeMode();
    }

    /**
     * Change the sheet mode.
     * @param {PrimarySheetMixin.MODES} [mode]  Mode to set. If not provided, mode will be toggled.
     */
    async changeMode(mode) {
      const { MODES } = this.constructor;
      this._mode = mode ?? (this.isEditMode ? MODES.PLAY : MODES.EDIT);
      const button = this.element?.querySelector('[data-action="changeMode"]');
      if ( button ) {
        const label = _loc(`DND5E.SheetMode${this.isEditMode ? "Edit" : "Play"}`);
        button.checked = this.isEditMode;
        button.dataset.tooltip = label;
        button.setAttribute("aria-label", label);
      }
      await this.submit();
      this.render();
    }
```

That greatly simplifies what we call here:

```suggestion
      if ( app?.rendered && app.changeMode ) app.changeMode();
```

arbron (changes requested):
I notice these aren't working in detached windows, it seems the keybinding handlers aren't being called at all. That seems like a core issue, but I'll let Kim way in when he reviews.

module/settings.mjs
Sayshal: Resolved.

Fyorl (approved):
Added a workaround for that core detached window bug. I also changed the edit mode toggle from `L` to `Shift+E` for parity with `Shift+B`. Maybe `L` was just one of the few unreserved single letters? Didn't seem related to edit toggle. Maybe edit toggLe? I also considered `F2` since that's used in lots of software for Edit but if you have a keyboard without its own function row that's a bit annoying. Probably doesn't matter too much since you can just rebind it anyway.

Issue #2837, opened 2024-02-02
[Feature Request] Allow "Group by Origin" to sort Features not granted by a specific source, such as manually-added features


TremendoDude:
As of 3.0.0, the new Default 5e Character Sheet now allows players to Group by Origin, separating features by Class, Race, Background, and Other. However, at the moment, it only groups features that were actively granted by the listed Class/Race/Background through automated granting of features.

There are many situations where someone will wish to add a custom feature or swap out a feature in accordance with these categories without having to go back through the automated feature-granting stages. Beyond that, when using a source that needs to be recreated by hand (such as a Race or Class from a physical sourcebook), the current Group by Origin option is unusable unless the source is first recreated with all automation intact, which is a shame when it'd often be so much easier to just create the features as needed.

It would be great if it were possible to Group by Origin in a less strict way, to allow making use of this useful sorting method with features not granted through automated recreated sources.
- One idea would be to allow features to group in accordance with Feature Type. Race Features would be grouped along with features granted by the Race, Class features would be grouped along with features granted by the class (perhaps with manual sorting allowing between classes in Multiclass situations), etc.
- Another idea would be to allow manual sorting between the Class, Race, Background, and other lists. By default features would appear in the list appropriate to their source, but players would be able to drag and drop a feature from one list to another. So for example, a manually added Class feature, which would by default appear in "Other Features", could be then dragged manually to the Class Features list.

etiquettestartshere:
I agree. Its particularly frustrating because some features that may be added after the fact, as the story or play dictates, often make the most sense logically to be added at level 1, since they aren't mean to represent something that was gained by level up, but just be part of that category, and level 1 tends to have the most steps of Advancement that must be replayed when selecting Modify Choices. The UX for adding even one new feature to a class category at level 1, for instance, is pretty grueling.

If drag and drop is too difficult, another option would be to have the feature groups be populated on items after they are already on the actor, similar to how items will populate resource consumption after the item is added to an actor.

M0nk3yy:
I would definitely be in favour of this. Upon first encountering such a situation, my default assumptions were that I'd either be able to: 

1. Drag and drop them into the section I wanted if they were not correctly categorized. 

or

2. Go to the details tab and change something near the Class Feature drop-down. Such as potentially a second drop-down that shows classes currently on the character sheet. 


[...]

PR #6382, opened 2025-11-08
[#6342] Convert `sourceClass` property to `sourceItem`

### LINKED ISSUE #6342: Consider adding Race/Species and or Background as label underneath spells granted by a Race/Species or Background.
Couldn't find an issue already made for it, so sorry if this is a duplicate. Spells granted through class & subclass have their subclass labels applied to the item, while spells granted by a race/species or background doesn't. Not sure if strictly necessary but it would sometimes be handy to know what spell comes from what part of your character.

For reference, the `Bone Chill` cantrip is from a race item, in this case "High Elf".

Image

Sayshal:
This PR replaces the old `sourceClass` field with a new `sourceItem` field that tracks where a spell comes from using an identifier that works across all item types. Closes #6342

### Core Data Model (`spell.mjs`)
- **Added `sourceItem` field**
  ```js
  sourceItem: new IdentifierField({ allowType: true, label: "DND5E.SourceItem.Label" })
  ```
  Stores the `type:identifier` of the source item (class, subclass, species, background, feat, etc.).
- **Fetch source items** via `actor.identifiedItems.get(sourceItem)?.first()` — automatically handles all item types.
- **Added `classIdentifier` getter** — resolves class/subclass sources to a class identifier for spellcasting ability and preparation tracking.
- **Deprecated `sourceClass` getter** with compatibility warning (5.3 → 6.0).

### UI Updates
- **Spell source dropdown**
  - Class spells: editable dropdown showing spellcasting classes
  - Other spells: locked field showing granting item name with hint
- **Spell subtitle** displays source name (e.g., "Fire Bolt • High Elf")

### Setting Source Item
- **Advancements** (`spell-config.mjs`):
  ```js
  setProperty(itemData, "system.sourceItem", `${this.item.type}:${this.item.identifier}`);
  ```
- **Item-granted spells** (`cast.mjs`):
  ```js
  "system.sourceItem": `${this.item.type}:${this.item.identifier}`
  ```
- **Spell creation** (`_preCreate`): Auto-sets to `class:{identifier}` when adding spells to actors.

### Display Logic (`base-actor-sheet.mjs`)
- Uses `actor.identifiedItems.get(sourceItem)?.first()` to fetch and display source item name.

### Migration
Automatically converts existing data:
```
sourceClass → sourceItem (prefixed with class:)
```

---

**Result:** Spell sources are now tracked consistently across all item types using the `identifiedItems` infrastructure. The generic `sourceItem` field name allows future extension to features and other item types.

## Before & After
Before
After

### DIFF
--- a/module/applications/actor/api/base-actor-sheet.mjs
+++ b/module/applications/actor/api/base-actor-sheet.mjs
@@ -965,8 +965,8 @@ export default class BaseActorSheet extends PrimarySheetMixin(
     // Subtitle
     let sourceLabel;
     if ( linked ) sourceLabel = linked.name;
-    else if ( item.system.spellSource ) {
-      const grantingItem = item.parent.identifiedItems.get(item.system.spellSource)?.first();
+    else if ( item.system.sourceItem ) {
+      const grantingItem = item.parent.identifiedItems.get(item.system.sourceItem)?.first();
       sourceLabel = grantingItem?.name;
     } else {
       // Check spells added from advancements
--- a/module/data/item/spell.mjs
+++ b/module/data/item/spell.mjs
@@ -54,7 +54,7 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
       properties: new SetField(new StringField(), { label: "DND5E.SpellComponents" }),
       range: new RangeField(),
       school: new StringField({ required: true, label: "DND5E.SpellSchool" }),
-      spellSource: new IdentifierField({ allowType: true, label: "DND5E.SpellSource.Label" }),
+      sourceItem: new IdentifierField({ allowType: true, label: "DND5E.SourceItem.Label" }),
       target: new TargetField()
     });
   }
@@ -140,8 +140,8 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
    * @type {string}
    */
   get classIdentifier() {
-    if ( !this.spellSource ) return "";
-    const sourceItem = this.parent?.actor?.identifiedItems.get(this.spellSource)?.first();
+    if ( !this.sourceItem ) return "";
+    const sourceItem = this.parent?.actor?.identifiedItems.get(this.sourceItem)?.first();
     if ( sourceItem?.type === "class" ) return sourceItem.identifier;
     if ( sourceItem?.type === "subclass" ) return sourceItem.system.classIdentifier ?? "";
     return "";
@@ -154,7 +154,7 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
    * @ignore
    */
   get sourceClass() {
-    foundry.utils.logCompatibilityWarning("SpellData#sourceClass is deprecated. Please use SpellData#spellSource "
+    foundry.utils.logCompatibilityWarning("SpellData#sourceClass is deprecated. Please use SpellData#sourceItem "
       + "instead.", { since: "DnD5e 5.3", until: "DnD5e 6.0" });
     return this.classIdentifier ?? "";
   }
@@ -310,7 +310,7 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
     SpellData.#migrateActivation(source);
     SpellData.#migrateTarget(source);
     SpellData.#migratePreparation(source);
-    SpellData.#migrateSpellSource(source);
+    SpellData.#migrateSourceItem(source);
   }
 
   /* -------------------------------------------- */
@@ -389,13 +389,13 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
   /* -------------------------------------------- */
 
   /**
-   * Migrate sourceClass to spellSource.
+   * Migrate sourceClass to sourceItem.
    * @since 5.3.0
    * @param {object} source  The candidate source data from which the model will be constructed.
    */
-  static #migrateSpellSource(source) {
+  static #migrateSourceItem(source) {
     if ( "sourceClass" in source ) {
-      if ( source.sourceClass ) source.spellSource = `class:${source.sourceClass}`;
+      if ( source.sourceClass ) source.sourceItem = `class:${source.sourceClass}`;
       delete source.sourceClass;
     }
   }
@@ -507,13 +507,13 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
     ];
 
     context.parts = ["dnd5e.details-spell", "dnd5e.field-uses"];
-    context.spellSourceLocked = false;
+    context.sourceItemLocked = false;
 
     // Default Ability & Spellcasting Classes
     if ( this.parent.actor ) {
       // Get spell source item.
-      const sourceItem = this.spellSource
-        ? this.parent.actor.identifiedItems.get(this.spellSource)?.first()
+      const sourceItem = this.sourceItem
+        ? this.parent.actor.identifiedItems.get(this.sourceItem)?.first()
         : null;
 
       const ability = CONFIG.DND5E.abilities[
@@ -548,10 +548,10 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
             label: grantingItem.name
           });
 
-          if ( !this.spellSource ) context.source.spellSource = `${grantingItem.type}:${grantingItem.identifier}`;
+          if ( !this.sourceItem ) context.source.sourceItem = `${grantingItem.type}:${grantingItem.identifier}`;
 
-          context.spellSourceLocked = true;
-          context.spellSourceHint = "DND5E.SpellSource.LockedHint";
+          context.sourceItemLocked = true;
+          context.sourceItemHint = "DND5E.SourceItem.LockedHint";
         }
       }
     }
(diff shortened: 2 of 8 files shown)


module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -975,10 +973,28 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
+      sourceLabel = linked.name;
+    } else if ( item.system.spellSource?.identifier ) {
+      const { type, identifier } = item.system.spellSource;
+      if ( type === "class" ) {
+        sourceLabel = item.parent.classes[identifier]?.name;
+      } else {
+        // For all other sources (subclass, race, background, item), look up the granting item by identifier
+        const grantingItem = item.parent.items.find(i => i.identifier === identifier);
+        sourceLabel = grantingItem?.name;
+      }
```
roth-michael: Feels like this could be
```suggestion
      const grantingItem = item.parent.itemTypes[type].find(i => i.identifier === identifier);
      sourceLabel = grantingItem?.name;
```
assuming `type` is always defined. This way a same-identifier background & item (or something) wouldn't end up with the wrong `grantingItem`.

module/data/item/spell.mjs
```diff
@@ -132,10 +135,30 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
...
   /* -------------------------------------------- */
 
+  /**
+   * @deprecated since 5.2
+   * @ignore
+   */
+  get sourceClass() {
+    foundry.utils.logCompatibilityWarning("SpellData#sourceClass is deprecated. Please use SpellData#spellSource "
+      + "instead.", { since: "DnD5e 5.2", until: "DnD5e 6.0" });
+    return this.spellSource?.identifier;
```
roth-michael: Maybe a check for `this.spellSource?.type === "class"`, returning empty string if not. (Or, cleverer, if `type` is a subclass, getting the class identifier corresponding to it). I could be convinced either way, dunno how much needs to go into a deprecated getter.

module/data/item/spell.mjs
```diff
@@ -454,14 +510,65 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
 
     // Default Ability & Spellcasting Classes
     if ( this.parent.actor ) {
+      // For subclass spells, resolve to the parent class identifier.
+      let classIdentifier = this.spellSource?.identifier;
+      if ( this.spellSource?.type === "subclass" ) {
+        const subclassItem = this.parent.actor.items.find(i => i.identifier === this.spellSource.identifier);
```
roth-michael:
```suggestion
        const subclassItem = this.parent.actor.subclasses[classIdentifier];
```

module/data/item/spell.mjs
```diff
@@ -454,14 +510,65 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
...
+        if ( advancementOrigin ) {
+          const [itemId] = advancementOrigin.split(".");
+          grantingItem = this.parent.actor.items.get(itemId);
+        }
+
+        // Check for item-granted spells.
+        if ( !grantingItem ) {
+          const linked = this.linkedActivity;
+          grantingItem = linked?.item;
+        }
```
roth-michael, line 543:
```suggestion
        grantingItem ??= this.linkedActivity?.item;
```

module/data/item/spell.mjs
```diff
@@ -454,14 +510,65 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
...
-        this.parent.actor.spellcastingClasses[this.sourceClass]?.spellcasting.ability
+        this.parent.actor.spellcastingClasses[classIdentifier]?.spellcasting.ability
           ?? this.parent.actor.system.attributes?.spellcasting
       ]?.label?.toLowerCase();
       if ( ability ) context.defaultAbility = game.i18n.format("DND5E.DefaultSpecific", { default: ability });
       else context.defaultAbility = game.i18n.localize("DND5E.Default");
       context.spellcastingClasses = Object.entries(this.parent.actor.spellcastingClasses ?? {})
         .map(([value, cls]) => ({ value, label: cls.name }));
+
+      const sourceType = this.spellSource?.type;
```
roth-michael: Since this is only used on line 533, I'd just remove this and have 533 be 
```js
if ( this.spellSource?.type === "class" ) {
```

module/data/item/spell.mjs
```diff
@@ -454,14 +510,65 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
...
+      let sourceIdentifier = this.spellSource?.identifier;
```
roth-michael: Similarly, this is only used below where it will be redefined anyway.

module/data/item/spell.mjs
```diff
@@ -454,14 +510,65 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
...
+        }
+
+        // Check for item-granted spells.
+        if ( !grantingItem ) {
+          const linked = this.linkedActivity;
+          grantingItem = linked?.item;
+        }
+
+        if ( grantingItem ) {
+          sourceIdentifier = grantingItem.identifier;
```
roth-michael: As mentioned on line 530, might as well just make this `const sourceIdentifier = grantingItem.identifier`

roth-michael (commented):
I wonder whether it doesn't make sense to move some of the logic in `getSheetData` (specifically, populating `spellSource` in the context using info from `advancementOrigin`) into a data preparation method. Ideally that information would be accessible programmatically for non-class-type spells, I think.

module/documents/activity/cast.mjs
```diff
@@ -128,6 +128,8 @@ export default class CastActivity extends ActivityMixin(BaseCastActivityData) {
           cachedFor: this.relativeUUID
         }
       },
+      "system.spellSource.type": "item",
```
pgervais70: Why not use `this.item.type` instead of hardcoded value, this way it would be similar to how consumption allows identifying an item, type and identifier.

module/applications/actor/api/base-actor-sheet.mjs
Sayshal: Resolved in latest commit.

module/data/item/spell.mjs
Sayshal: Resolved in latest commit.

module/data/item/spell.mjs
Sayshal: Resolved in latest commit.

module/data/item/spell.mjs
Sayshal: Resolved in latest commit.

module/data/item/spell.mjs
Sayshal: Resolved in latest commit.

module/data/item/spell.mjs
Sayshal: Resolved in latest commit.

module/data/item/spell.mjs
Sayshal: Resolved in latest commit.

module/documents/activity/cast.mjs
Sayshal: Resolved in latest commit.

Sayshal:
> I wonder whether it doesn't make sense to move some of the logic in `getSheetData` (specifically, populating `spellSource` in the context using info from `advancementOrigin`) into a data preparation method. Ideally that information would be accessible programmatically for non-class-type spells, I think.

Should this be a new data preparation method or can it move into an existing one, do you think? Trying to avoid disrupting the existing code structure as much as possible.

module/data/item/spell.mjs
```diff
@@ -52,7 +52,10 @@ export default class SpellData extends ItemDataModel.mixin(ActivitiesTemplate, I
       properties: new SetField(new StringField(), { label: "DND5E.SpellComponents" }),
       range: new RangeField(),
       school: new StringField({ required: true, label: "DND5E.SpellSchool" }),
-      sourceClass: new StringField({ label: "DND5E.SpellSourceClass" }),
+      spellSource: new SchemaField({
+        type: new StringField({ label: "DND5E.SpellSource.Type" }),
+        identifier: new StringField({ label: "DND5E.SpellSource.Identifier" })
+      }, { label: "DND5E.SpellSource.Label" }),
```
arbron: Rather than changing this to a nested structure, we probably want to use the new `allowType` option on `IdentifierField` that was added in `5.2` so this can be a single field with simpler migration:

```suggestion
      spellSource: new IdentifierField({ allowType: true, label: "DND5E.SpellSource.Label" }),
```

Then whenever you need to fetch the item you can just do `actor.identifiedItems.get(spellSource)?.first()` and it will handle figuring out the type and fetching the correct documents for you.

Sayshal:
Ready for re-review, I think i got everything this time. Humbly await any more feedback.

Sayshal:
Hoping this can be prioritized into 5.2.X or 5.3.

roth-michael:
A note on this. If implemented, this new field could be extended to features as well, thereby helping fulfill #2837 (specifically, sharing a mechanism for defining the "origin"/source of a feature/spell)

Sayshal:
> Further to Michael's comments, if we want to extend this work to include things like allowing users to manually fill the source class for any given feature, then we should consider renaming this field to something like `sourceItem`. Better to do it now with one migration rather than having to migrate `sourceClass` -> `spellSource` -> `sourceItem` later.

I will make adjustments later this afternoon to `sourceItem`, I think a seperate PR to open this up for potentially resolving #2837 would make the most sense. I'm on board with the adjustments made.

Sayshal:
@Fyorl changes made, ready for final review.

PR #6400, opened 2025-11-13
[#6384] Adds habitat/treasure block & pull quote styling

### LINKED ISSUE #6384: Add habitat/treasure block and bestiary quote styling to system
These two sections are used in the Monster Manual and it would be good to be able to use these in other pieces of content without having to reproduce the styling:

Image

arbron:
Moves the habitat/treasure block and pull quote styling from the Monster Manual into the system to be more widely used. Changes the classes used there to remove the positioning from the quote (instead of `quote` and `quote-left`, use `quote-lg float-right` and `quote-lg float-left`).

Habitat   Quote Styling Light

Habitat   Quote Styling Dark

Closes #6384

### DIFF
--- a/less/v2/blocks.less
+++ b/less/v2/blocks.less
@@ -218,7 +218,7 @@
   }
 
   /* Big Quotes */
-  .quote {
+  .quote-lg {
     --size: 50%;
     --side-gutter: 10%;
     --vert-gutter: 1em;


less/v2/blocks.less
```diff
@@ -194,4 +194,85 @@
...
+  /* NPC Habitat and Treasure Blocks */
+  .habitat-treasure {
+    padding: 2px;
+    border-block-start: 1px solid var(--dnd5e-habitat-treasure-color-border);
+    background-color: var(--dnd5e-habitat-treasure-color-background);
+    display: table;
+  }
+
+  /* Big Quotes */
+  .quote {
```
Fyorl: Is this the style we want for *all* quotes? Or just ones that are matching the MM style? Maybe something like `.pull-quote`, `.quote-lg`, or `.quote-mm`?

PR #6415, opened 2025-11-17
Add Eberron, Faerûn, and Greyhawk calendars, approximate formatters

arbron:
Adds three new calendars into the system for use in popular settings: the Calendar of Greyhawk (for Greyhawk), the Calendar of Harptos (for Forgotten Realms), and the Calendar of Khorvaire (for Eberron).

Calendar - Greyhawk
Calendar - Harptos (normal)
Calendar - Khorvaire

The Calendar of Harptos includes some special formatting for the six festval days that occur throughout the year outside of the normal months. For those, it will just display the name of the festival rather than the day and month.

Calendar - Harptos (festival)

This also includes two new formatters for approximate dates and times. The date formatter uses the current season with "Early", "Mid", and "Late" to give a hint as to when in the season it is. For time it uses the time of sunrise and sunset and selects a descriptive term depending on the time of the day. These terms include Sunrise, Sunset, Noon, Midnight, Dawn, Dusk, Morning, Afternoon, Evening, and Night.

Calendar - Approximate Date   Time

### DIFF
--- a/module/data/calendar/calendar-data.mjs
+++ b/module/data/calendar/calendar-data.mjs
@@ -20,7 +20,7 @@ export default class CalendarData5e extends foundry.data.CalendarData {
    * @returns {number}                      Number of hours since the start of the day as a decimal.
    */
   static hoursOfDay(time=game.time.components, calendar=game.time.calendar) {
-    const components = typeof time === "number" ? this.timeToTimeComponents(time) : time;
+    const components = typeof time === "number" ? this.timeToComponents(time) : time;
     const minutes = components.minute + (components.second / calendar.days.secondsPerMinute);
     return components.hour + (minutes / calendar.days.minutesPerHour);
   }


JPMeehan:
Thank you for addressing the year zero property, going from 0 YK to 998 was a bit of a pain...

module/data/calendar/calendar-data.mjs
```diff
@@ -127,12 +132,70 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+    let seasonEnd = season.dayEnd;
+    let days = 0;
+    for ( const month of calendar.months.values ) {
+      if ( (seasonStart !== null) && (seasonEnd !== null) ) break;
+      if ( season.monthStart === month.ordinal ) seasonStart = days;
+      if ( season.monthEnd === month.ordinal ) seasonEnd = days;
+      days += components.leapYear ? month.leapDays ?? month.days : month.days;
+    }
+    if ( seasonEnd < seasonStart ) seasonEnd += calendar.days.daysPerYear;
+    if ( day < seasonStart ) day += calendar.days.daysPerYear;
```
Fyorl, line 154: Seems like this should be impossible. Can `components.day` be before the start day of the current `components.season`? Doesn't hurt to be defensive I suppose.

module/data/calendar/calendar-of-greyhawk.mjs
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+/* -------------------------------------------- */
+
+export const CALENDAR_OF_GREYHAWK = {
+  name: "Calendar of Greyhawk",
+  years: {
+    yearZero: 576,
+    firstWeekday: 0
```
Fyorl, line 35: Would be nice to offer a way to configure `yearZero` and `firstWeekday` via settings that we then merge into the chosen calendar config. I imagine most campaigns don't take place at the start of that world's calendar. I'm not sure if we are currently taking `firstWeekday` into account anywhere either.

module/data/calendar/calendar-data.mjs
```diff
@@ -88,28 +90,31 @@ export default class CalendarData5e extends foundry.data.CalendarData {
 
   /**
    * Prepared date parts passed to the localization.
-   * @param {CalendarData} calendar      The configured calendar
-   * @param {TimeComponents} components  Time components to format
```
Fyorl, line 92: Is the typedef not `TimeComponents`?

module/data/calendar/calendar-of-harptos.mjs
```diff
@@ -0,0 +1,167 @@
...
+      { name: "DND5E.CALENDAR.Harptos.Day.Nine", ordinal: 9 },
+      { name: "DND5E.CALENDAR.Harptos.Day.Ten", ordinal: 10 }
+    ],
+    daysPerYear: 365,
+    hoursPerDay: 24,
+    minutesPerHour: 60,
+    secondsPerMinute: 60
+  },
+  festivals: [
+    { name: "DND5E.CALENDAR.Harptos.Festival.Midwinter", month: 1, day: 31 },
```
Fyorl, line 152: I was wondering how you were going to do the intercalary days. I did the same thing for the Harptos calendar implementation on my website by tacking them onto the previous month.

module/data/calendar/calendar-data.mjs
arbron: That happens when the season wraps around the end of the year, so if Winter starts in December and you are in January, this is the keep the percentage consistent. Without this as soon as you pass the new years it will jump from "Mid-Winter" to "Early Winter".

module/data/calendar/calendar-of-greyhawk.mjs
arbron: We don't use `firstWeekday` anywhere at the moment, but we'll probably want to be able to display the weekday in a formatter as some point (and we'll need it if we want to display a whole calendar).

I'll add a setting for starting year though.

arbron: I suppose for starting year, do we want to change `yearZero` or just provide an interface for easily setting the initial year by changing the game time? A sort of "Jump to Year" control in settings. That would ensure that the day of week wouldn't need to be customized to match a change in year if the days of the week don't align neatly with the days of the year.

JPMeehan: Jump to year would generally be helpful — time travel and historical campaigns are frequent topics in published material

Fyorl: A simple dialog that offers year, month, & day inputs to jump to would be good and seems relatively straightforward.

module/data/calendar/calendar-data.mjs
```diff
@@ -16,12 +15,12 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
-   * @param {number|Components} [time]  The time to use, by default the current world time.
-   * @param {CalendarData} [calendar]   Calendar to use, by default the current world calendar.
-   * @returns {number}                  Number of hours since the start of the day as a decimal.
+   * @param {number|TimeComponents} [time]  The time to use, by default the current world time.
+   * @param {CalendarData} [calendar]       Calendar to use, by default the current world calendar.
+   * @returns {number}                      Number of hours since the start of the day as a decimal.
    */
   static hoursOfDay(time=game.time.components, calendar=game.time.calendar) {
-    const components = typeof time === "number" ? this.timeToComponents(time) : time;
+    const components = typeof time === "number" ? this.timeToTimeComponents(time) : time;
```
Fyorl: Looks like this got hit with the find & replace.

PR #6432, opened 2025-11-21
[#6366] Add `PoolTerm` damage roll simplification for damage tooltip

### LINKED ISSUE #6366: Damage roll with advantage does not display dice rolls in the chat card
Rolling damage with advantage (e.g. `{1d8,1d8}kh`) fails to display the actual dice rolls in the chat card:

Image

(the example is from an enricher i manually put into chat, but actual weapon attack damage rolls have the same problem)

`/roll`ing the same term works as expected:

Image

This is with 0 modules on 13.350 and 5.1.10.

Aioros:
Allow `_simplifyDamageRoll` to process `PoolTerm`s. Individual dice will be added to the tooltip with the appropriate CSS classes.

Closes #6366

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -612,9 +612,7 @@ export default class ChatMessage5e extends ChatMessage {
           const result = term.results[i];
           // Apply main result classes to individual dice
           simplified.dice.forEach(die => {
-            const resultClasses = termResultClasses
-              .filter(c => result[c])
-              .join(" ");
+            const resultClasses = termResultClasses.filter(c => result[c]).join(" ");
             if ( resultClasses.length ) die.classes += ` ${resultClasses}`;
           });
           aggregate.dice.push(...simplified.dice);


module/documents/chat-message.mjs
```diff
@@ -587,30 +587,48 @@ export default class ChatMessage5e extends ChatMessage {
...
         aggregate.method ??= tooltipData.method;
       }
+      if ( term instanceof PoolTerm ) {
+        term.rolls.forEach((poolTermRoll, i) => {
+          // Get simplified data for each roll
+          const simplified = this._simplifyDamageRoll(poolTermRoll);
+          const result = term.results[i];
+          // Apply main result classes to individual dice
+          simplified.dice.forEach(die => {
+            const resultClasses = ["success", "failure", "rerolled", "exploded", "discarded"]
```
Fyorl: Did you observe that some entries were missing the appropriate classes? I would have expected `DiceTerm#getTooltipData` to handle this already.

Fyorl (changes requested):

module/documents/chat-message.mjs
Aioros: Let's say the pool term is `{2d8kl,3d6kl}kh`; each inner term's `getTooltipData` would provide the expected classes for their specific `kl` modifier, but that wouldn't account for the outer `kh`, which would discard one entire inner term. So I'm checking the results on the main PoolTerm itself and applying that to the respective DiceTerm's dice (if the 2d8kl term was discarded, apply `discarded` to both d8s). Let me know if that makes sense or if there's something I'm missing.

Fyorl: Got it, thank you

module/documents/chat-message.mjs
```diff
@@ -587,30 +587,48 @@ export default class ChatMessage5e extends ChatMessage {
...
+              .filter(c => !!result[c] && !die.classes.includes(c))
+              .filterJoin(" ");
+            if (resultClasses.length) die.classes += ` ${resultClasses}`;
```
Fyorl:
```suggestion
            if ( resultClasses.length ) die.classes += ` ${resultClasses}`;
```

module/documents/chat-message.mjs
Fyorl: Can hoist this array to the top of the method since it's not being mutated.

module/documents/chat-message.mjs
```diff
@@ -587,30 +587,48 @@ export default class ChatMessage5e extends ChatMessage {
...
       }
+      if ( term instanceof PoolTerm ) {
+        term.rolls.forEach((poolTermRoll, i) => {
+          // Get simplified data for each roll
+          const simplified = this._simplifyDamageRoll(poolTermRoll);
+          const result = term.results[i];
+          // Apply main result classes to individual dice
+          simplified.dice.forEach(die => {
+            const resultClasses = ["success", "failure", "rerolled", "exploded", "discarded"]
+              .filter(c => !!result[c] && !die.classes.includes(c))
```
Fyorl: `.filter` only cares about truthiness, not exact booleans, so we can skip a coercion. Equally the browser doesn't care about duplicate class names.


```suggestion
              .filter(c => result[c])
```

module/documents/chat-message.mjs
```diff
@@ -587,30 +587,48 @@ export default class ChatMessage5e extends ChatMessage {
...
+              .filterJoin(" ");
```
Fyorl: Don't need to filter again.


```suggestion
              .join(" ");
```

PR #6433, opened 2025-11-21
Add dialog for setting the current date

arbron:
Adds a new dialog for setting the game time to a specific date using the year, month, and day. Because core's `GameTime#set` method doesn't support using `month` or `dayOfMonth` to set the date, this adds a new `CalendarData5e#jumpToDate` method that takes those values and calculates the correct values before passing them to core.

Set Date Dialog

### DIFF
--- a/module/data/calendar/calendar-data.mjs
+++ b/module/data/calendar/calendar-data.mjs
@@ -95,7 +95,7 @@ export default class CalendarData5e extends foundry.data.CalendarData {
    * @param {number} [components.day]    Day within the month.
    */
   async jumpToDate({ year, month, day }) {
-    const components = game.time.components;
+    const components = { ...game.time.components };
     year ??= components.year + this.years.yearZero;
     month ??= components.month;
     day ??= components.dayOfMonth;
@@ -111,6 +111,7 @@ export default class CalendarData5e extends foundry.data.CalendarData {
       dayOfYear += leapYear ? (m.leapDays ?? m.days) : m.days;
     }
     components.day = dayOfYear;
+    components.month = month;
 
     await game.time.set(components);
   }


module/data/calendar/calendar-data.mjs
```diff
@@ -83,6 +83,38 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+
+  /**
+   * Set the date to a specific year, month, or day. Any values not provided will remain the same.
+   * @param {object} components
+   * @param {number} [components.year]   Visible year (with `yearZero` added in).
+   * @param {number} [components.month]  Index of month.
+   * @param {number} [components.day]    Day within the month.
+   */
+  async jumpToDate({ year, month, day }) {
+    const components = game.time.components;
```
Fyorl: Should we be mutating `game.time.components`? Maybe it doesn't matter if we're about to set `game.time` anyway.


```suggestion
    const components = { ...game.time.components };
```

module/data/calendar/calendar-data.mjs
```diff
@@ -83,6 +83,38 @@ export default class CalendarData5e extends foundry.data.CalendarData {
...
+    components.year = year - this.years.yearZero;
+    const { leapYear } = this._decomposeTimeYears(this.componentsToTime(components));
+
+    // Convert days within month to day of year
+    let dayOfYear = day - 1;
+    for ( let idx=0; idx<month; idx++ ) {
+      const m = this.months.values[idx];
+      dayOfYear += leapYear ? (m.leapDays ?? m.days) : m.days;
+    }
+    components.day = dayOfYear;
```
Fyorl, line 113: We don't set `components.month` anywhere.


```suggestion
    components.day = dayOfYear;
    components.month = month;
```

arbron: `componentsToTime` doesn't use month for calculating the timestamp anywhere, that's why we have to convert to days of year instead.

Issue #6410, opened 2025-11-15
Jumping movement type uses twice as much movement as it should


Yamiash102:
When using the Jump movement option, it calculates movement as though you move 2 ft for every 1 ft traveled:

<img width="447" height="227" alt="Image" src="https://github.com/user-attachments/assets/7e035aee-b031-41b2-927f-96ede72ca60d" />

Tested in Foundry v13.351 and DnD 5e v5.1.10 with and without modules enabled.

I don't expect this to work exactly as the rules indicate, with the math for dashes and such becoming increasingly complicated. However, the rules in both 2014 and 2024 don't say jumping should cost twice as much movement.

2014 SRD:

<img width="854" height="263" alt="Image" src="https://github.com/user-attachments/assets/ddf7f988-8022-45ca-b1b5-1e54ef39dc0c" />

2024 SRD:

<img width="863" height="296" alt="Image" src="https://github.com/user-attachments/assets/becbee9b-e51f-49fb-8a74-29e6a5e079c6" />

Both from DnDBeyond.

Fyorl:
Fixed in @Aioros PR #6435 if you would like to comment for assignment.

Aioros:
Yep, thanks

PR #6435, opened 2025-11-21
Derive Jump movement

Aioros:
This might be more opinionated than I'm comfortable with, let me know what you think. The safer part of this PR is a correction of the movement cost for Jump to bring it in line with the rules (as described in #6410).

Then, to evaluate the ruler's styling, instead of hardcoding a specific behavior for `jump` I'm proposing a callback configuration property called `deriveSpeed`, which would allow defining a movement type with its own custom calculation for the base speed used by the ruler. This works well enough for Jump, where green (corresponding to half the strength score) would be ok, and yellow (full strength score) would suggest that the movement might have other requirements (the 10ft run-up). But it's definitely possible that this is too specific to be actually useful.

### DIFF
--- a/module/_types.mjs
+++ b/module/_types.mjs
@@ -371,19 +371,12 @@
 
 /**
  * @typedef MovementTypeConfiguration
- * @property {DeriveSpeedFunction} [deriveSpeed]  Function used to derive the max action speed.
  * @property {boolean} [hidden=false]             Whether this movement speed is displayed in the actor's sheet.
  * @property {string} label                       Localized label for the movement type.
  * @property {string} [travel]                    Travel type in `CONFIG.DND5E.travelTypes` to map this movement
  *                                                speed to. If not provided, then `land` is assumed.
  * @property {boolean} [walkFallback]             When this special movement type runs out, can the actor fall back
  *                                                to using their walk speed at 2x cost?
- */
-
-/**
- * @callback DeriveSpeedFunction
- * @param {Token5e} token  The token attempting the movement.
- * @returns {number}
  */
 
 /* -------------------------------------------- */
--- a/module/_types.mjs
+++ b/module/_types.mjs
@@ -371,11 +371,19 @@
 
 /**
  * @typedef MovementTypeConfiguration
- * @property {string} label            Localized label for the movement type.
- * @property {string} [travel]         Travel type in `CONFIG.DND5E.travelTypes` to map this movement speed to. If not
- *                                     provided, then `land` is assumed.
- * @property {boolean} [walkFallback]  When this special movement type runs out, can the actor fall back to using their
- *                                     walk speed at 2x cost?
+ * @property {DeriveSpeedFunction} [deriveSpeed]  Function used to derive the max action speed.
+ * @property {boolean} [hidden=false]             Whether this movement speed is displayed in the actor's sheet.
+ * @property {string} label                       Localized label for the movement type.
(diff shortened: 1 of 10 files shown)


module/_types.mjs
```diff
@@ -370,11 +370,18 @@
...
- * @property {string} [travel]         Travel type in `CONFIG.DND5E.travelTypes` to map this movement speed to. If not
- *                                     provided, then `land` is assumed.
- * @property {boolean} [walkFallback]  When this special movement type runs out, can the actor fall back to using their
- *                                     walk speed at 2x cost?
+ * @property {string} label                       Localized label for the movement type.
+ * @property {string} [travel]                    Travel type in `CONFIG.DND5E.travelTypes` to map this movement
+ *                                                speed to. If not provided, then `land` is assumed.
+ * @property {boolean} [walkFallback]             When this special movement type runs out, can the actor fall back
+ *                                                to using their walk speed at 2x cost?
+ * @property {DeriveSpeedFunction} [deriveSpeed]  Function used to derive the max action speed.
```
Fyorl: Prefer to order these properties in alphabetical order.

Fyorl (commented):
No objection to these changes. I think a more forward-looking solution would be to derive a `movement.jump` value from the strength score (after AEs have applied). In v14 you can then target this field with the `final` AE phase allowing for effects that increase jump distance.

Aioros:
I definitely like the derived `movement.jump` and the active effect flexibility. I'll switch to that and update, thanks.

(I'm assuming the whole `deriveSpeed` thing can be safely scrapped, but let me know if you think it could be useful on its own).

Aioros:
What I landed on for now:

- Kept the `deriveSpeed` function property in the config;
- Derived a `movement.jump` attribute, used it in the `deriveSpeed` for jump;
- Realized that adding `jump` to the movement types caused it to be displayed with the other speeds in the sheet; added optional `hidden` property and the necessary tweaks to avoid that.

module/applications/actor/npc-sheet.mjs
```diff
@@ -371,7 +371,7 @@ export default class NPCActorSheet extends BaseActorSheet {
 
     // Speed
     context.speed = [
-      ...Object.entries(CONFIG.DND5E.movementTypes).map(([k, { label }]) => {
+      ...Object.entries(CONFIG.DND5E.movementTypes).filter(([k, m]) => !m.hidden).map(([k, { label }]) => {
```
Fyorl:
```suggestion
      ...Object.entries(CONFIG.DND5E.movementTypes).filter(([, m]) => !m.hidden).map(([k, { label }]) => {
```

module/canvas/ruler.mjs
```diff
@@ -100,6 +100,11 @@ export default class TokenRuler5e extends foundry.canvas.placeables.tokens.Token
       currActionSpeed = Math.max(currActionSpeed, movement.walk);
     }
 
+    // If current action has a derived speed, use that instead
+    if ( CONFIG.DND5E.movementTypes[waypoint.action]?.deriveSpeed ) {
+      currActionSpeed = CONFIG.DND5E.movementTypes[waypoint.action].deriveSpeed(this.token);
+    }
```
Fyorl: I think this should be unnecessary now. We are retrieving the `currActionSpeed` from `movement[waypoint.action]` on line 95 which should give the correct result already since `movement.jump` exists now.

Fyorl (changes requested):

Fyorl (approved):

PR #6445, opened 2025-11-24
[#3060] Add `@attributes.movement.speed`, add AE shim

arbron:
Adds `system.attributes.movement.speed` that is populated as soon as walking speed is calculated to allow other speed formulas to reference it.

Also adds an AE shim to redirect and effects targeting that field towards `movement.walk`.


module/documents/active-effect.mjs
```diff
@@ -47,6 +47,17 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
 
   /* -------------------------------------------- */
 
+  /**
+   * Active effect fields that should be redirected to another field, optionally with a compatibility warning.
+   * Optional warning object contains options passed to `foundry.utils.logCompatibilityWarning`.
+   * @type {Record<string, { path: string, [warning]: object }>}
```
```suggestion
   * @type {Record<string, { key: string, [warning]: object }>}
```

PR #6446, opened 2025-11-24
[#6364] Improve ownership handling in group sheet.

### LINKED ISSUE #6364: Group Actor Visibility Issues
Pre the revised group sheet, Limited/Observer permissions only let players see a generic scale of health for each group member.
Now it seems that it's all or nothing, with the group sheet Limited view showing nothing but the name and token which is effectively useless.
We used to like the ability to see the sheet but not know in absolute terms how much HP each PC had remaining.
The DM has now reduced the permission level to prevent us from seeing this, however this also locks us out of all of the shared inventory features.

Mechanical information about group members should be hidden from users who do not have ownership of that actor.

Fyorl:
- Closes #6364

### DIFF
--- a/templates/actors/group/member.hbs
+++ b/templates/actors/group/member.hbs
@@ -13,7 +13,9 @@
         </div>
         <div class="bars">
             <strong class="name">
-                <a class="rollable" data-action="showDocument" data-uuid="{{ uuid }}">{{ name }}</a>
+                <a {{#if canView}}data-action="showDocument"{{/if}} data-uuid="{{ uuid }}">
+                    {{ name }}
+                </a>
             </strong>
             {{#with system.attributes.hp}}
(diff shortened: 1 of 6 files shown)


templates/actors/group/member.hbs
```diff
@@ -13,7 +13,9 @@
         </div>
         <div class="bars">
             <strong class="name">
-                <a class="rollable" data-action="showDocument" data-uuid="{{ uuid }}">{{ name }}</a>
+                <a {{#if canView}}data-action="showDocument"{{/if}} data-uuid="{{ uuid }}">
```
Fyorl, line 16: Some assumptions about the `rollable` class inherited from other actor sheets that don't really apply to the group sheet with individual member ownership.

Issue #6403, opened 2025-11-14
Update wiki for 5.2 release


arbron:
- [ ] Activities
  - [ ] Visibility changes
- [ ] Hooks
  - [ ] Movement token blocking hook: https://github.com/foundryvtt/dnd5e/pull/6038

PR #6450, opened 2025-11-24
[#6403] Update activity wiki pages based on 5.2 changes

arbron:


wiki/Activity-Type-Enchant.md
```diff
@@ -30,17 +30,17 @@ The "Enchantments" tab contains a list of potential enchantments that can be app
...
+#### Additional Activities, Effects, & Items
 
-When enchanting profiles can be given additional effects and items that are added alongside the enchantment. Because enchantments can only affect the item to which they are applied, these *Additional Effects* can be used to give the item and Active Effect that changes the actor. The *Additional Items* can be used to specify additional items that are given to the actor when this enchantment is applied. Both the effects and items will be removed from the actor when the enchantment is removed.
+When enchanting profiles can be given additional activities, effects, and items that are added alongside the enchantment. The *Additional Activities* can be used to automatically add new activities to the item that is enchanted. Because enchantments can only affect the item to which they are applied, these *Additional Effects* can be used to give the item and Active Effect that changes the actor. The *Additional Items* can be used to specify additional items that are given to the actor when this enchantment is applied. All of these activities, effects, and items will be removed from the actor when the enchantment is removed.
```
Fyorl:
```suggestion
When enchanting, profiles can be given additional activities, effects, and items that are added alongside the enchantment. The *Additional Activities* can be used to automatically add new activities to the item that is enchanted. Because enchantments can only affect the item to which they are applied, these *Additional Effects* can be used to give the item an Active Effect that changes the actor. The *Additional Items* can be used to specify additional items that are given to the actor when this enchantment is applied. All of these activities, effects, and items will be removed from the actor when the enchantment is removed.
```

Issue #6344, opened 2025-10-29
Change the way a creature's `Size` is done to allow for modifying it via active effects.


Zanderaf:
Currently, theres no way to increase or decrease an actor's size by a specific step, such as a Medium creature becoming Large / a Small creature becoming tiny.

When creating an "Enlarge / Reduce" style active effect where a creature changes size by a certain step up or down, you need to specify the exact size category the active effect changes the actor into, which isn't really optimal.

One solution could be to make `size` similar to a skills proficiency level (0, 0.5, 1, 2)- with it going from 1- Tiny to 6 - Gargantuan, allowing for effects to specifically do changes like 

`key for creature size / add / 1`

to increase the size of the creature by one step or ``key for creature size / upgrade / 5` to upgrade a creature's size to `Huge`

krbz999:
To avoid migrating actor data and effects to use numeric values (and to avoid users having to remember seemingly arbitrary numbers), the AE documentClass or the size data field itself could have special handling for what the different `ACTIVE_EFFECT_MODES` mean.

eg `UPGRADE | med` would change only small/tiny actors to Medium, and `ADD | 1` would go one step 'up'.

(To be even more helpful, this could also be extended to support keys like "medium" or "large" instead of the somewhat less intuitive "med" and "lg" but that's an aside.)

Zanderaf:
that feels like a better solution now t hat I think of it, would be able to support custom sizes as well

Aioros:
Commenting for assigning purpose (I'll probably see if we can add `multiply` to the new setup once the new PR is merged)

PR #6456, opened 2025-11-25
Actor size active effects

Aioros:
Humble proposal for a `SizeField` actor field with active effect integration as discussed in #6344. I'm basing the implementation on the `numerical` property that was added to actor sizes in 5.1.10. The assumption is that these are consecutive integers, which seemed reasonable judging by how they are used in the token movement automation.

For `MULTIPLY`, I opted to base it on token size: `MULTIPLY | 2` would bring `tiny` to `sm`, `sm` and `med` to `lg`, and so on. `MULTIPLY | 0.5` brings `lg` to `medium`, `medium` and `small` to `tiny`. The closest size available is chosen if the target size is not an integer.

### DIFF
--- a/module/data/actor/fields/size-field.mjs
+++ b/module/data/actor/fields/size-field.mjs
@@ -1,0 +1,81 @@
+/**
+ * Subclass of StringField that tracks an actor's size.
+ */
+export default class SizeField extends foundry.data.fields.StringField {
+
+  #actorSizes = Object.entries(CONFIG.DND5E.actorSizes).reduce((obj, [key, size]) => {
+    obj[key] = obj[size.numerical] = { ...size, key };
+    obj.length = (obj.length ?? 0) + 1;
+    return obj;
+  }, {});
+
+  /* -------------------------------------------- */
+  /*  Active Effect Integration                   */
+  /* -------------------------------------------- */
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeOverride(value, delta, model, change) {
+    // Force a given actor size.
+    if ( Number.isNumeric(delta) ) delta = Math.clamp(delta, 0, this.#actorSizes.length - 1);
+    return this.#actorSizes[delta]?.key ?? value;
+  }
+
+  /** @override */
+  _applyChangeAdd(value, delta, model, change) {
+    // Increment actor size by a numeric value
+    if ( !Number.isNumeric(delta) ) return value;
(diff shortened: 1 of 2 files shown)


module/data/actor/fields/size-field.mjs
```diff
@@ -0,0 +1,81 @@
...
+  _applyChangeDowngrade(value, delta, model, change) {
+    // Downgrade actor size to a smaller value
+    const currentNumerical = this.#actorSizes[value].numerical;
+    if ( Number.isNumeric(delta) ) delta = Math.clamp(delta, 0, this.#actorSizes.length - 1);
+    let targetNumerical = this.#actorSizes[delta]?.numerical ?? currentNumerical;
+    const newNumerical = Math.min(currentNumerical, targetNumerical);
+    return this.#actorSizes[newNumerical].key;
+  }
+
+}
```
krbz999:
```suggestion
  }
}
```

module/data/actor/templates/traits.mjs
```diff
@@ -20,7 +21,7 @@ export default class TraitsField {
    */
   static get common() {
     return {
-      size: new StringField({ required: true, initial: "med", label: "DND5E.Size" }),
+      size: new SizeField({ required: true, initial: "med", label: "DND5E.Size" }),
```
krbz999: Could move these defaults to the field itself (though not the `label`).

arbron:
@Aioros Didn't notice this PR and ended up writing my own size AE PR: https://github.com/foundryvtt/dnd5e/pull/7035

Mine handled the V14 active effect changes but doesn't handle the `multiply` change mode, so maybe once mine is merged you could adjust this one to just add multiply support for separate consideration.

(also maybe give a comment on the original issue so I can get you added as a second Assignee)

Aioros:
I've actually been thinking about the `multiply` change mode for this and how I was trying to implement it, and I'm not sure it fully makes sense anymore.
My original thought for size multiplication was something like "double size means that a 1x1 token becomes 2x2, and a 2x2 becomes 4x4", which... is pretty weird all around, and would still be weird if it tried to double the area instead of the side.
I'm thinking of the old phrasing for Enlarge/Reduce, for example, where "doubles in all dimensions" translates to "go up one category", and leaning towards considering `multiply` fairly useless for this field, as I think the other modes cover pretty much any first- or third-party features that I can remember. If that sounds right, feel free to close this PR.

PR #6458, opened 2025-11-26
[#6368] Allow for non-physical ammo properties

### LINKED ISSUE #6368: Allow modules to define non-physical ammo properties.
As a developer, I would like to be able to flag an item property as being valid for ammunition without it also being a physical property for damage resistances.

Currently, the dnd5e system hard codes the "returning" property as the only non-physical ammo property. Premium 3rd party modules like the Grim Hollow Player's Guide add more non-physical ammo properties like Armor Piercing, which cannot be accurately modeled in 5.1. It would be helpful if there was a set or boolean that helped flag if something is valid for ammo.

Sayshal:
Added a `validForAmmunition` property to allow for non-physical weapon properties like 'Armor Piercing' from 3rd party premium module "Grim Hollow Player's Guide". Closes #6368

### DIFF
--- a/module/_types.mjs
+++ b/module/_types.mjs
@@ -331,13 +331,12 @@
  * Configuration data for item properties.
  *
  * @typedef ItemPropertyConfiguration
- * @property {string} label                 Localized label.
- * @property {string} [abbreviation]        Localized abbreviation.
- * @property {string} [icon]                Icon that can be used in certain places to represent this property.
- * @property {string} [reference]           Reference to a rule page describing this property.
- * @property {boolean} [isPhysical]         Is this property one that can cause damage resistance bypasses?
- * @property {boolean} [isTag]              Is this spell property a tag, rather than a component?
- * @property {boolean} [validForAmmunition] Can this property be applied to ammunition without being physical?
+ * @property {string} label           Localized label.
+ * @property {string} [abbreviation]  Localized abbreviation.
+ * @property {string} [icon]          Icon that can be used in certain places to represent this property.
+ * @property {string} [reference]     Reference to a rule page describing this property.
+ * @property {boolean} [isPhysical]   Is this property one that can cause damage resistance bypasses?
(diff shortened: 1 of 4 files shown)


arbron:
I wonder if we could make use of the existing `CONFIG.DND5E.validProperties` config rather than adding a new properly to `ItemPropertyConfiguration`. Maybe we could abuse Javascript a bit and stick it as an attribute on the valid properties set:

```javascript
// In `config.mjs`
DND5E.validProperties.consumable.ammunition = new Set(["ret"]);

// In `data/item/templates/item-description.mjs`
  get validProperties() {
    const config = CONFIG.DND5E.validProperties[this.parent.type] ?? [];
    const valid = new Set([...config, ...(config[this.type?.value] ?? [])]);
    if ( this.parent.actor?.system.isNPC && this.schema.has("quantity") ) valid.add("gear");
    return valid;
  }
```

krbz999:
with most items having a subtype in the same location (`system.type.value`), one (new) object for all subtypes would also make sense.

Sayshal:
> I wonder if we could make use of the existing `CONFIG.DND5E.validProperties` config rather than adding a new properly to `ItemPropertyConfiguration`. Maybe we could abuse Javascript a bit and stick it as an attribute on the valid properties set:
> 
> ```js
> // In `config.mjs`
> DND5E.validProperties.consumable.ammunition = new Set(["ret"]);
> 
> // In `data/item/templates/item-description.mjs`
>   get validProperties() {
>     const config = CONFIG.DND5E.validProperties[this.parent.type] ?? [];
>     const valid = new Set([...config, ...(config[this.type?.value] ?? [])]);
>     if ( this.parent.actor?.system.isNPC && this.schema.has("quantity") ) valid.add("gear");
>     return valid;
>   }
> ```

Reworked per your suggestion and moved it onto `CONFIG.DND5E.validProperties` as a subtype set and generalized the base `validProperties` getter to union `config[type.value]`.

Fyorl (approved):

PR #6459, opened 2025-11-26
[#6330] Allow TableOfContentsCompendium entries to be explicitly hidden

### LINKED ISSUE #6330: Allow TableOfContentsCompendium entries to be explicitly hidden
Previously to hide a journal entry in the table of contents app you would remove all dnd5e flags. But since navigation has moved to the the dnd5e flag object it isn't possible to hide those entries in the journal anymore if there is navigation on the pages (Something that is quite useful for organization/navigation).

Solution: Have the table of contents app ignore the navigation object and its keys for the purose of displaying with something like
`
      if ( !flags ) continue;
      const keys = Object.keys(flags);
      if ( (keys.length === 1) && (keys[0] === "navigation") ) continue;
      const type = flags.type ?? "chapter";
`

Nice addition: Have tocHidden apply to both pages and entries.

Sayshal:
Closes #6330

### DIFF
--- a/module/applications/journal/table-of-contents.mjs
+++ b/module/applications/journal/table-of-contents.mjs
@@ -87,6 +87,9 @@ export default class TableOfContentsCompendium extends foundry.applications.side
     for ( const entry of documents ) {
       const flags = entry.flags?.dnd5e;
       if ( !flags ) continue;
+      const keys = Object.keys(flags);
+      if ( (keys.length === 1) && (keys[0] === "navigation") ) continue;
+      if ( flags.tocHidden ) continue;
       const type = flags.type ?? "chapter";
 
       if ( type === "header" ) {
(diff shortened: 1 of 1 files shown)


module/applications/journal/table-of-contents.mjs
```diff
@@ -87,6 +87,9 @@ export default class TableOfContentsCompendium extends foundry.applications.side
     for ( const entry of documents ) {
       const flags = entry.flags?.dnd5e;
       if ( !flags ) continue;
+      const keys = Object.keys(flags);
+      if ( (keys.length === 1) && (keys[0] === "navigation") ) continue;
+      if ( flags.tocHidden ) continue;
```
arbron:
```suggestion
      if ( flags.tocHidden || !keys.length || ((keys.length === 1) && (keys[0] === "navigation")) ) continue;
```

arbron (changes requested):

arbron (approved):

PR #6483, opened 2025-11-29
Pass `chatMessage` to all enchantment related Hooks

### LINKED ISSUE #6484: Pass the `chatMessage` data in Enchantment relevant Hooks.
Would allow for more and easier control over enchantments if users wanted to add to the system restrictions already present.

thatlonelybugbear:
`chatMessage` holds lots of data that we could use to affect if an Enchantment should go through.

Passing in the relevant Hooks the `chatMessage` will allow for more niche control and restrictions.

Closes #6484

### DIFF
--- a/module/documents/activity/enchant.mjs
+++ b/module/documents/activity/enchant.mjs
@@ -134,7 +134,7 @@ export default class EnchantActivity extends ActivityMixin(BaseEnchantActivityDa
 
     // Validate against the enchantment's restraints on the origin item
     if ( strict ) {
-      const errors = this.canEnchant(item);
+      const errors = this.canEnchant(item, chatMessage);
       if ( errors?.length ) {
         errors.forEach(err => ui.notifications.error(err.message, { console: false }));
         return null;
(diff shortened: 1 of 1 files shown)


module/documents/activity/enchant.mjs
```diff
@@ -203,10 +205,11 @@ export default class EnchantActivity extends ActivityMixin(BaseEnchantActivityDa
 
   /**
    * Determine whether the provided item can be enchanted based on this enchantment's restrictions.
-   * @param {Item5e} item  Item that might be enchanted.
+   * @param {Item5e} item                Item that might be enchanted.
+   * @param {ChatMessage5e} chatMessage  Chat message used to make the enchantment, if applicable.
```
Fyorl, line 209: 'if applicable' implies it's optional (and it is, in the case of the call in `ActiveEffect#_preCreate`), so we should put it in an options bag.

module/documents/activity/enchant.mjs
```diff
@@ -257,8 +260,9 @@ export default class EnchantActivity extends ActivityMixin(BaseEnchantActivityDa
      * @param {EnchantmentError[]} errors  List of errors containing failed restrictions. The item will be enchanted
      *                                     so long as no errors are listed, otherwise the provided errors will be
      *                                     displayed to the user.
+     * @param {ChatMessage5e} chatMessage  Chat message used to make the enchantment, if applicable.
      */
-    Hooks.callAll("dnd5e.canEnchant", this, item, errors);
+    Hooks.callAll("dnd5e.canEnchant", this, item, errors, chatMessage);
```
Fyorl, line 265: Same comment as above, and gives us some future-proofing.

PR #6503, opened 2025-12-01
[#6482] Add damage rounding between application of resistances and vulnerabilities

### LINKED ISSUE #6482: [BUG] Damage calculation sometimes incorrect when target has both reistance and vulnerability
When a creature has both resistance and vulnerability to a particular damage type, and the damage rolled is an odd number, the resultant damage should be one less than originally rolled, because it's halved for resistance, rounded down, and _then_ doubled, as laid out in the [Basic Rules](https://www.dndbeyond.com/sources/dnd/br-2024/playing-the-game#OrderofApplication). However, the Foundry system seems not to be rounding off that half point of damage before doubling it again, leading the creature to take one more damage than they should. A very minor issue to be sure, but it _is_ incorrect so I thought it worth at least documenting.

Aioros:
Round down damage immediately when evaluating damage resistances instead of using an aggregated multiplier. The calculated `multiplier`, even if not directly involved in the calculation, is still kept and included in each `DamageSummary` item in case we prefer to maintain compatibility, but I don't think it's used anywhere else by the system.

Closes #6482.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -847,18 +847,28 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       }
 
       let damageMultiplier = multiplier;
+      let appliedDamage = d.value * multiplier;
 
       // Apply damage resistance
-      if ( this.#changeHasEffect("resistance", d, { options }) ) damageMultiplier /= 2;
+      if ( this.#changeHasEffect("resistance", d, { options }) ) {
+        damageMultiplier /= 2;
+        appliedDamage = Math.trunc(appliedDamage / 2);
+      }
 
       // Apply damage vulnerability
-      if ( this.#changeHasEffect("vulnerability", d, { options }) ) damageMultiplier *= 2;
+      if ( this.#changeHasEffect("vulnerability", d, { options }) ) {
+        damageMultiplier *= 2;
+        appliedDamage *= 2;
+      }
 
       // Negate healing types
       if ( (options.invertHealing !== false) && ((d.type === "healing")
-        || ((d.type === "maximum") && (treatAs === "healing"))) ) damageMultiplier *= -1;
+        || ((d.type === "maximum") && (treatAs === "healing"))) ) {
+        damageMultiplier *= -1;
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -878,29 +878,35 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       }
 
       let damageMultiplier = multiplier;
+      let appliedDamage = Math.trunc(d.value * multiplier);
```
Fyorl: I think we only need to round at the point of computing resistance. If we round here then every damage part gets rounded individually after applying the global multiplier before being totalled, which is not correct. If you take half damage on a save, that's half the total damage, not the sum of half of each damage type.


```suggestion
      let appliedDamage = d.value * multiplier;
```

Aioros:
Maybe force pushing on this wasn't the cleanest idea, let me know if that's an issue. Just wanted to have it ready for 5.3+. Removed the unnecessary rounding, everything else is still the same.

Yeah, I think I messed the branch up. I'll clean up, apologies.

arbron:
> Maybe force pushing on this wasn't the cleanest idea, let me know if that's an issue. Just wanted to have it ready for 5.3+. Removed the unnecessary rounding, everything else is still the same.
> 
> Yeah, I think I messed the branch up. I'll clean up, apologies.

No, its fine, you just needed to change the target branch from 5.2 to 5.3 on GitHub. Fixed it for you.

thatlonelybugbear:
I was away and just catching up.

Not sure if any input is still needed, or where the best place for that might be

Fyorl:
> Not sure if any input is still needed, or where the best place for that might be

Sure, input is still fine, wherever you would like to leave it (that I can see). Just can't guarantee it will be acted upon expediently.

As of this PR `d.active.multiplier` is no longer reliable to determine exactly how the final amount was calculated. As far as I can tell, this data has no internal uses, only downstream uses. So I wondered if this change would be disruptive and, if so, what the desired resolution might look like. Also maybe they are related to some existing damage PRs.

Issue #6513, opened 2025-12-03
Damage threshold incorrectly applied to healing


Aioros:
Steps to reproduce:

1) create new actor;
2) set their HP using the HitPointsConfig dialog;
3) try to apply healing damage.

Seems like the issue is undesired precedence in this condition:
```js
damages.amount < (this.system.attributes?.hp?.dt) ?? -Infinity
```
where the < is evaluated before the coalescing. When setting the HP on the actor, `dt` becomes `null` (instead of the original `undefined`), and funnily enough `-1 < undefined === false` but `-1 < null === true`. In any case, just fixing the parenthesis should fix it. I'll make a quick PR in a few.

roth-michael:
That fix alone won't be sufficient to resolve the issue - with any _set_ damage threshold, `damages.amount` will still be negative for _healing_, which means it'll always be "less" than the threshold. So while the above fixes healing on creatures with un-set damage thresholds, healing will continue to be 0 for any creatures with set ones.

Aioros:
Good point, thanks, a little embarrassed I missed that. I imagine that the actual condition should instead be something like this?
```js
(damages.amount > 0 && damages.amount < (this.system.attributes?.hp?.dt ?? -Infinity))
```

roth-michael:
Yeah, I think that'd make sense. Assuming damage threshold and healing _shouldn't_ have any interaction (which I think is the case).

PR #6515, opened 2025-12-03
[#6514] Fix healing when damage threshold is set

Aioros:
Addendum to PR #6513 to fully fix #6514, apologies for having missed it the first time (and thanks @roth-michael for bringing it up). As mentioned in the issue's discussion, this assumes that damage thresholds do not apply in any way to healing damage.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -903,7 +903,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     damages.amount = damages.amount > 0 ? Math.floor(damages.amount) : Math.ceil(damages.amount);
 
     // Apply damage threshold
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -903,7 +903,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     damages.amount = damages.amount > 0 ? Math.floor(damages.amount) : Math.ceil(damages.amount);
 
     // Apply damage threshold
-    if ( (damages.amount < (this.system.attributes?.hp?.dt ?? -Infinity))
+    if ( (damages.amount > 0 && damages.amount < (this.system.attributes?.hp?.dt ?? -Infinity))
```
```suggestion
    if ( ((damages.amount > 0) && (damages.amount < (this.system.attributes?.hp?.dt ?? -Infinity)))
```

arbron (changes requested):

Issue #5726, opened 2025-05-25
Dice Pools do not work with critical hits


NonchalantWombat:
When using custom formula for weapon attack damage that contains a dice pool, such as: 

{10d6, 10d6}kh + @mod

When rolling critical hit damage, the base string does not double the dice. Adding extra critical hit damage seems to roll three times, instead of adding that into the base.

![Image](https://github.com/user-attachments/assets/b9e83009-5354-425d-8135-2b0084ad8759)
![Image](https://github.com/user-attachments/assets/d9b7584b-8ebf-47dd-bcce-2196f8af13a8)

Aioros:
Commenting for assignment purposes

Issue #5060, opened 2025-01-26
Activities: Damage -> Formulas not applying critical when there is a roundup or down


Somel1978:
When you use a formula on the damage the activity does not process criticals:

![Image](https://github.com/user-attachments/assets/4e727690-c849-4f9d-acc9-419acd544245)

![Image](https://github.com/user-attachments/assets/e8721182-614d-43eb-9cb1-fccc68dd30f0)

This was replicated on a vanila setup without any other modules installed.
With versions 4.1.2 and 4.2.2. On foundry v12. and on Foundry v13.
Was replicated with max critical damage and regular critical (double dice).

Aioros:
Commenting for assignment purposes

Issue #4910, opened 2024-12-20
Only one Critical Damage Die is being maximized with dnd5e setting on


ctbritt:
Possibly related to https://github.com/foundryvtt/dnd5e/issues/3192, when setting the dnd5e system setting to maximize critical damage dice results, only lone die is being maximized. 

![CleanShot 2024-12-20 at 13 40 13](https://github.com/user-attachments/assets/f315ac8a-ebbc-4fc3-afb9-6ae2d50453f3)

(i've manually lowered the critical threshold for testing purposes, also the 15/barbarian has brutal critical _and_ savage attacks for +3 critical damage dice.)

In this case, the roll is `4d12+4+1+12`, when it should be `1d12+4+1+48`

Aioros:
Commenting for assignment purposes

Issue #4752, opened 2024-11-19
Critical Damage possibly weird with dice formulas that include modifiers like "kh"


Zanderaf:
Applying critical damage to a formula that goes something like "2d10kh" outputs a resulting formula of "4d10kh". Not exactly sure on everyone else's view on it, but I feel like it should be outputting a formula of "2d10kh + 2d10kh"
![image](https://github.com/user-attachments/assets/3a1825c9-f9a3-41a8-a59e-39f744290c98)

krbz999:
When a term has *any* modifier, maybe it's reasonable that the term should be duplicated rather than doubled in size. To simplify things. This is probably only relevant for the `kh`, `kl`, `dh`, and `dl` modifiers, but that's already 4 variants not to mention the variants with a numeric value appended (e.g. `kh3`).

Aioros:
Commenting for assignment purposes

PR #6538, opened 2025-12-09
Critical damage fixes

Aioros:
This PR tries to provide a reasonable implementation of critical damage configuration for complex damage rolls.
The first, simpler commit is a fix for #4910, to allow maximization of all bonus critical dice.
The second takes a larger opinionated general approach, targeting issues #4752, #5060, #5726:

- `NumericTerm`s and `DiceTerm`s with no modifiers keep the original logic;
- `NumericTerm`s with modifiers and any other non-deterministic term are duplicated (or more precisely, added to the formula multiple times according to the critical multiplier and bonus dice available).

Some work on `BasicRoll` was necessary to correctly sync-evaluate maximized pool terms with modifiers.
A few examples of critical transformations:

- `2d6 + 3` -> `4d6 + 3`
- `2d6kh + 3` -> `2d6kh + 3 + 2d6kh`
- `floor(3d6/2) + 1` -> `floor(3d6/2) + 1 + floor(3d6/2)`
- `2d6kh + 3`, with 2 bonus dice and powerful criticals -> `2d6kh + 3 + 18`
- `{1d10,3d4}kl + 2` , with multiply numeric and powerful critical -> `{1d10,3d4}kl + 4 + 10`

### DIFF
--- a/module/applications/dice/d20-configuration-dialog.mjs
+++ b/module/applications/dice/d20-configuration-dialog.mjs
@@ -63,14 +63,13 @@ export default class D20RollConfigurationDialog extends RollConfigurationDialog 
   /* -------------------------------------------- */
 
   /** @override */
-  _finalizeRolls(action) {
+  _finalizeConfig(config, action) {
     let advantageMode = CONFIG.Dice.D20Roll.ADV_MODE.NORMAL;
     if ( action === "advantage" ) advantageMode = CONFIG.Dice.D20Roll.ADV_MODE.ADVANTAGE;
     else if ( action === "disadvantage" ) advantageMode = CONFIG.Dice.D20Roll.ADV_MODE.DISADVANTAGE;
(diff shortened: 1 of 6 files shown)


Fyorl:
I got to this too late for it to go into 5.3 but I think it's important because some of these formulas we use in official content. I'll tag it to 6.0 to make sure I look at it. Worth also re-evaluating in the wake of the `adv`/`dis` modifiers.

Fyorl (approved):
Sorry to somewhat bulldoze your PR. Felt that we were slapping too many band-aids on this damage roll code so I tried to build it back up from first principles and just kept pulling on the thread. It should handle any formula pretty robustly now, including some nonsensical ones that we wouldn't expect anyway but fall out from the code for free.

Also some overdue cleanup of v10 baggage that we were lugging around and no longer need to deal with.

PR #6543, opened 2025-12-09
Add application for configuring overall Table of Contents flags

arbron:
Adds a new configuration application available from the header of the table of contents window that displays all of the journal entries in the pack grouped by folder and allows for setting the `type` and `position` flag.

Table of Contents Compendium Config

Configuration of the other entry flags (`order`, `title`, & `showPages`) as well as the individual page flags (`tocHidden`) will be left to a separate config application accessible from the individual entry sheets.

### DIFF
--- a/module/applications/journal/config/compendium-toc-config.mjs
+++ b/module/applications/journal/config/compendium-toc-config.mjs
@@ -166,7 +166,7 @@ export default class CompendiumTOCConfig extends Application5e {
     for ( const [id, flags] of Object.entries(submitData) ) {
       const update = { _id: id };
       for ( const key of ["type", "position", "append"] ) {
-        if ( flags[key] ) update[`flags.dnd5e.${key}`] = flags[key];
+        if ( flags[key] || (flags[key] === 0) ) update[`flags.dnd5e.${key}`] = flags[key];
         else update[`flags.dnd5e.-=${key}`] = null;
       }
       updates.push(update);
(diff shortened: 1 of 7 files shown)


module/applications/journal/config/compendium-toc-config.mjs
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _prepareContext(options) {
+    const context = await super._prepareContext(options);
+
+    // TODO: Ideally this would be `getIndex({ fields: ["flags.dnd5e.type", "flags.dnd5e.position"] })`
+    // but that doesn't seem to properly update the flags when they are updated
```
krbz999: Relevant issue: <https://github.com/foundryvtt/foundryvtt/issues/9984>

module/applications/journal/config/compendium-toc-config.mjs
```diff
@@ -0,0 +1,176 @@
...
+      .sort((lhs, rhs) => lhs.sort - rhs.sort)
+      .map((option, value) => ({ ...option, value }));
+
+    context.folders = [];
+    const traverse = node => {
+      if ( !node ) return;
+      context.folders.push({
+        name: node.folder?.name ?? game.i18n.localize("DND5E.TABLEOFCONTENTS.NoFolder"),
+        entries: node.entries.map(o => {
+          const entry = this.compendium.get(o._id);
```
Fyorl: I think in practice this works because the compendium has been fully populated by opening the ToC sheet. But in that case, we don't need the `getIndex` call above. So I assume this was intending to read from the index.


```suggestion
          const entry = this.compendium.index.get(o._id);
```

Fyorl (approved):

PR #6553, opened 2025-12-11
[#6554] - Wiki: Active Effects Guide. Clarification of ability checks/saves roll modes

### LINKED ISSUE #6554: [Wiki] - Add ability checks roll modes in the active effects guide and clarify change modes usage
Image

thatlonelybugbear:
- Added Ability Checks roll modes
- Added all relevant change modes explanation

Closes #6554

### DIFF
--- a/wiki/Active-Effect-Guide.md
+++ b/wiki/Active-Effect-Guide.md
@@ -24,6 +24,8 @@ Useful examples:
 > When using formulas in an Active Effect Value, the actor sheet display that corresponds to the changed value will not always display the evaluated formula, but it will be applied when rolled.
 > E.g. When adding `@abilities.cha.mod` to `system.bonuses.abilities.save` to simulate a Paladin's Aura of Protection, the actor sheet will not display that bonus applied to saving throws. The bonus will be present when the saving throw is rolled.
 
+## Change Modes
+### General description
 | Change Mode | Description |
 |------------ | ------------|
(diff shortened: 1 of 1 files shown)


thatlonelybugbear:
End result should be something like this
image

Fyorl (changes requested):
These change mode behaviours apply to any `RollConfigField`, of which there are several more in addition to ability checks and saving throws: initiative, concentration, skills, tools, death saves.

I don't think the appropriate place for this information is under ability checks and saving throws specifically.

thatlonelybugbear:
Other than reworking the whole guide to cluster entries about `.XYZ.roll.mode` together, I don't see where they could all go.
They are now close to one another based on both being `system.abilities...`

The same could be added for other system keys which lack specific examples.

One solution could be to add something like:
<hr>


For both Ability Checks and Saves, the following Change Modes are supported:

* `Add`: Adds one instance of the specified Advantage Mode to the roll.
* `Override`: Forces the roll to the specified Advantage Mode.
* `Downgrade` with Effect Value `0`: The roll cannot have advantage.
* `Upgrade` with Effect Value `0`: The roll cannot have disadvantage.
>💡 The same logic can be applied to 
>`system.attributes.concentration.roll.mode`,
>`system.attributes.death.roll.mode`,
>`system.attributes.init.roll.mode`,
> `system.skills.[abbreviation].roll.mode`, 
>`system.tools.[abbreviation].roll.mode`

krbz999:
I'd say add a section specifically just for Advantage Modes and refer to it for each relevant section. It has enough features on its own that it could deserve its own section with in-depth explanations of how ADD/OVERRIDE/UPGRADE/DOWNGRADE work.

E: The field works the same for each entry of it, so for each section the guide would only need to describe the attribute key and then refer to the Advantage Mode section for further elaboration.

PR #6562, opened 2025-12-13
[#1267] Add `@details.tier` to actor data

### LINKED ISSUE #1267: Add `tier` to actor's roll data
Many player features scale up with character level depending on what tier they fall in. For example, the Dragonborn breath weapon starts at 2d6 damage and increased by one for each tier the player advances to (3d6 at 6th level, 4d6 at 11th, 5d6 at 16th). Currently the best way to automatically determine that damage is to write `((ceil((@details.level - 5)/5)) +2)d6` into the roll formula, which isn't very clean on its purpose. Including `tier` in the roll data would allow this to be simplified to `(@details.tier + 1)d6`.

Tier could either be calculated using a formula just like above (which would automatically determine post level 20 tiers if a module enabled them) or it could be an option configured in `config.js`.

arbron:
Uses a formula `ceil((level - 4) / 6) + 1` to calculate the tier.

Closes #1267

### DIFF
--- a/module/data/item/class.mjs
+++ b/module/data/item/class.mjs
@@ -205,6 +205,7 @@ export default class ClassData extends ItemDataModel.mixin(
   prepareDerivedData() {
     super.prepareDerivedData();
     this.prepareDescriptionData();
+    this.tier = Math.ceil((this.levels - 4) / 6) + 1;
   }
 
   /* -------------------------------------------- */


arbron:
I've added `tier` to individual classes also because I've had some content that scales in the same way as cantrips, but based off a class level rather than the overall character level.

PR #6572, opened 2025-12-16
[#6488] Add AC and speed attribution tooltips for members in a group sheet

### LINKED ISSUE #6488: [Feature] Group sheet display AC breakdown
It would be helpful as the DM to be able to see at a glance on the group sheet how a player's AC is being generated (shield is equipped, spells, feat, etc) without having to open their sheet or ask the player. When I mouse over the AC tag in the group sheet it just displays "Armor Class", but when you mouse over the "AC" tag in a player's sheet it breaks it down. 

Image

Aioros:
Closes #6488.

### DIFF
--- a/templates/actors/group/member.hbs
+++ b/templates/actors/group/member.hbs
@@ -86,11 +86,17 @@
         {{else}}
         <div class="info">
             <div>
-                <i class="fa-solid fa-shield" data-tooltip aria-label="{{ localize "DND5E.ArmorClass" }}"></i>
+                <i class="fa-solid fa-shield"
+                   data-attribution="attributes.ac" data-reference-tooltip="{{ uuid }}"
+                   data-attribution-caption="DND5E.ArmorClass" aria-label="{{ localize "DND5E.ArmorClass" }}">
+                </i>
                 <span class="value">{{ system.attributes.ac.value }}</span>
             </div>
             <div>
-                <i class="fa-solid fa-shoe-prints" data-tooltip aria-label="{{ localize "DND5E.Speed" }}"></i>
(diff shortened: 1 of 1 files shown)


templates/actors/group/member.hbs
```diff
@@ -86,11 +86,17 @@
         {{else}}
         <div class="info">
             <div>
-                <i class="fa-solid fa-shield" data-tooltip aria-label="{{ localize "DND5E.ArmorClass" }}"></i>
+                <i class="fa-solid fa-shield"
```
arbron: Right now this is a pretty small hover area, it would probably be nicer if this was on the containing `<div>` so that hovering over the icon or the value would show the tooltip. That would also prevent the tooltip from concealing the value when hovered.

templates/actors/group/member.hbs
```diff
@@ -86,11 +86,17 @@
...
+                   data-attribution="attributes.ac" data-reference-tooltip="{{ uuid }}"
+                   data-attribution-caption="DND5E.ArmorClass" aria-label="{{ localize "DND5E.ArmorClass" }}">
+                </i>
                 <span class="value">{{ system.attributes.ac.value }}</span>
             </div>
             <div>
-                <i class="fa-solid fa-shoe-prints" data-tooltip aria-label="{{ localize "DND5E.Speed" }}"></i>
+                <i class="fa-solid fa-shoe-prints"
```
arbron: Same comment as with AC.

arbron (changes requested):

Fyorl (approved):

PR #6575, opened 2025-12-17
[#6495] Allow for effects to adjust the bloodied threshold

### LINKED ISSUE #6495: Allow the threshhold for being "Bloodied" to be set via an active effect or somehow changed.
There are some features in 3rd party premium content (Grim Hollow 2024) that allows you to be considered Bloodied when below your current hit point maximum. Currently no way to change it afaik since Bloodied seems to be hardcoded at 50% hit points.

Being able to override that threshold with some sort of flag would be helpful.

krbz999:
Stores a numeric value in `CharacterData/NPCData#attributes.bloodiedThreshold` during the base data prep cycle, grabbing the initial value from the CONFIG, which has been adjusted to be 50 instead of 0.5 for simplicity.

Adjusted to compare HP percentage using `attributes.hp.pct` instead of a calculation in the `updateBloodied` method using `hp.effectiveMax` (unsure why this was the way it was).

To allow for the edge case of "bloodied when missing *any* hit points" there's a comparison to `hp.pct === 100` such that an effect can be applied with `system.attributes.bloodiedThreshold | OVERRIDE | 100` to satisfy this requirement.

Closes #6495.

### DIFF
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -148,6 +148,9 @@ export default class AttributesFields {
         }),
         value: new NumberField({ integer: true, min: 0, initial: 0, persisted: false })
       }, { label: "DND5E.Attunement" }),
+      bloodiedThreshold: new NumberField({
+        nullable: false, min: 0, max: 100, persisted: false, initial: () => CONFIG.DND5E.bloodied.threshold
+      }),
       senses: new SensesField(),
       spell: new SchemaField({
         attack: new NumberField({ integer: true }),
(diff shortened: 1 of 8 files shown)


module/data/actor/templates/attributes.mjs
```diff
@@ -239,6 +239,16 @@ export default class AttributesFields {
 
   /* -------------------------------------------- */
 
+  /**
+   * Prepare the bloodied threshold.
+   * @this {CharacterData|NPCData}
+   */
+  static prepareBaseBloodied() {
+    this.attributes.bloodiedThreshold = CONFIG.DND5E.bloodied.threshold;
```
arbron: `bloodiedThreshold: NumberField({ min: 0, max: 100, initial: () => CONFIG.DND5E.bloodied.threshold, persisted: false });` perhaps?

krbz999: yes, could be simplified a lot with v14.

Fyorl (approved):

PR #6588, opened 2025-12-23
Makes `originatingMessage` available to calculateDamage and applyDamage workflows

### LINKED ISSUE #6587: Pass the `messageId` into the options object of the apply damage button clicked.
That way we can associate the button clicked with a message allowing to retrieve more data during `dnd5e.preApplyDamage` and `dnd5e.applyDamage` hooks.

thatlonelybugbear:
Closes #6587

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -865,7 +865,7 @@ export default class ChatMessage5e extends ChatMessage {
       properties: new Set(roll.options.properties ?? [])
     }));
     return Promise.all(canvas.tokens.controlled.map(t => {
-      return t.actor?.applyDamage(damages, { multiplier, isDelta: true, originatingMessage: this.chatMessage });
+      return t.actor?.applyDamage(damages, { multiplier, isDelta: true, originatingMessage: this });
     }));
   }
 


module/applications/components/damage-application.mjs
```diff
@@ -312,9 +312,11 @@ export default class DamageApplicationElement extends TargetedApplicationMixin(C
    */
   async _onApplyDamage(event) {
     event.preventDefault();
+    const messageId = event.target?.closest?.('[data-message-id]')?.dataset?.messageId;
```
arbron: No need to do this manually, you can just use `this.chatMessage` which will always have the current chat message.

module/applications/components/damage-application.mjs
```diff
@@ -312,9 +312,11 @@ export default class DamageApplicationElement extends TargetedApplicationMixin(C
...
     for ( const target of this.targetList.querySelectorAll("[data-target-uuid]") ) {
       const token = fromUuidSync(target.dataset.targetUuid);
       const options = this.getTargetOptions(target.dataset.targetUuid);
+      options.messageId = messageId;
```
arbron: Rather than setting it here, you should set this in `getTargetOptions` so that the message data is also available during the earlier calls to `calculateDamage` and the hooks that that message calls.

thatlonelybugbear:
> Rather than passing the `messageId`, I was thinking we could just pass the whole chat message to save everyone the extra work of fetching it on their own (maybe as `options.origin`).

Should we maybe preserve the `originatingMessage` nomenclature in this case, using `options.originatingMessage`?

thatlonelybugbear:
> > Rather than passing the `messageId`, I was thinking we could just pass the whole chat message to save everyone the extra work of fetching it on their own (maybe as `options.origin`).
> 
> Should we maybe preserve the `originatingMessage` nomenclature in this case, using `options.originatingMessage`?

I see in <https://github.com/foundryvtt/dnd5e/pull/6662/changes#diff-753ee83a2ebd7a18eea5612c5e96c010e8a12005fb0fb72c4da510385888de7aR333> that you already added it in as `origin`, so disregard.

arbron:
> Should we maybe preserve the `originatingMessage` nomenclature in this case, using `options.originatingMessage`?

That might make more sense. Looking at how we use `origin` throughout the system at the moment, I would probably expect that to be the activity that triggered the damage rather than the chat message.

(don't worry about my other PR, I'll adjust that to whatever we settle on, I just needed something to get it working)

thatlonelybugbear:
> > Should we maybe preserve the `originatingMessage` nomenclature in this case, using `options.originatingMessage`?
> 
> That might make more sense. Looking at how we use `origin` throughout the system at the moment, I would probably expect that to be the activity that triggered the damage rather than the chat message.
> 
> (don't worry about my other PR, I'll adjust that to whatever we settle on, I just needed something to get it working)

OK then, I will stick to `originatingMessage` and see how that goes

module/documents/chat-message.mjs
```diff
@@ -865,7 +865,7 @@ export default class ChatMessage5e extends ChatMessage {
       properties: new Set(roll.options.properties ?? [])
     }));
     return Promise.all(canvas.tokens.controlled.map(t => {
-      return t.actor?.applyDamage(damages, { multiplier, isDelta: true });
+      return t.actor?.applyDamage(damages, { multiplier, isDelta: true, originatingMessage: this.chatMessage });
```
Fyorl:
```suggestion
      return t.actor?.applyDamage(damages, { multiplier, isDelta: true, originatingMessage: this });
```

Fyorl (approved):

PR #6589, opened 2025-12-23
[#5154] Create custom message types for bastion attacks & turns

### LINKED ISSUE #5154: Add custom message type for bastions
Currently the `Bastion` singleton is responsible for creating the chat messages for bastions and handling events triggered from the bastion chat cards. Adding a new custom chat message type would allow for moving some of this logic out of that monolithic class.

This could either be handled by a single bastion card, or there could be two types to handle the bastion turns and the bastion attacks, since they have different data and rendering requirements.

The turn cards currently store all of their data in `flags.dnd5e.bastion` as an object containing `orders` (object with facility ID and order results), `items` (array of objects with item UUIDs and quantity), and `gold` (object with gold quantity and claimed boolean).

The attack cards store their data in the same place and include `deaths` (number of defenders killed) or `undefended` (no defenders before the attack occurred).

arbron:
Creates two new message types, `bastionAttack` and `bastionTurn`, and moves the logic and event handlers into those types rather than having it all in `Bastion`.

These types now dynamically render their content directly from the data rather than relying on pre-rendered `content`.

Closes #5154

### DIFF
--- a/module/data/chat-message/bastion-turn-message-data.mjs
+++ b/module/data/chat-message/bastion-turn-message-data.mjs
@@ -138,7 +138,7 @@ export default class BastionTurnMessageData extends ChatMessageDataModel {
     const { gp } = this.actor?.system.currency ?? {};
     if ( !this.gold.value || this.gold.claimed || (gp === undefined) ) return;
     await this.actor.update({ "system.currency.gp": gp + this.gold.value });
-    this.parent.update({ content, system: { gold: { claimed: true } } });
+    this.parent.update({ "system.gold.claimed": true } } });
   }
 
   /* -------------------------------------------- */


module/data/chat-message/bastion-turn-message-data.mjs
```diff
@@ -0,0 +1,176 @@
...
+   * Handle claiming generated gold.
+   * @this {BastionTurnMessageData}
+   * @param {Event} event         Triggering click event.
+   * @param {HTMLElement} target  Button that was clicked.
+   */
+  static async #onClaimGold(event, target) {
+    const { gp } = this.actor?.system.currency ?? {};
+    if ( !this.gold.value || this.gold.claimed || (gp === undefined) ) return;
+    await this.actor.update({ "system.currency.gp": gp + this.gold.value });
+    this.parent.update({ content, system: { gold: { claimed: true } } });
```
Fyorl: `content` not defined
```suggestion
    this.parent.update({ "system.gold.claimed": true });
```

Fyorl (commented):
Good stuff, thank you

Fyorl (approved):

Issue #5647, opened 2025-05-09
Prompt to advance bastion turns automatically based on calendar


arbron:
With calendar integration, the system should be able to track when the next bastion turn should occur and somehow prompt the GM to trigger one. This could be by either placing a GM-only message into chat with the button, placing a button in the calendar HUD, or automatically selecting "Advance Bastion Turn" when resting from a group actor (or some combination of the three).

This would need to track on what day the previous bastion turn occurred.

PR #6591, opened 2025-12-24
[#5647] Advance bastion orders based on calendar

arbron:
Automatically advance bastion orders based on when new days occur using the built-in calendar. Allows orders to complete without needing a bastion turn to be triggered. Each facility with an order in progress stores the time it was last updated and that is used to advance or complete the order when time passes.

 ### Todo
- [ ] Prompt for bastion turns on regular interval

### DIFF
--- a/module/data/settings/bastion-setting.mjs
+++ b/module/data/settings/bastion-setting.mjs
@@ -24,4 +24,31 @@ export default class BastionSetting extends foundry.abstract.DataModel {
       })
     };
   }
+
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Cached version of the minimum level required to have a bastion.
+   * @param {number}
+   */
+  static #threshold;
+
+  /* -------------------------------------------- */
+  /*  Helpers                                     */
+  /* -------------------------------------------- */
+
+  /**
+   * Determine whether bastions are available for a specific actor.
+   * @param {Actor5e} actor  Actor that may have a bastion.
+   * @returns {boolean}
+   */
+  availableForActor(actor) {
+    if ( !BastionSetting.#threshold ) {
+      const { basic, special } = CONFIG.DND5E.facilities.advancement;
+      BastionSetting.#threshold = Math.min(...Object.keys(basic), ...Object.keys(special));
+    }
+    return this.enabled && (actor.system.details?.level >= BastionSetting.#threshold);
+  }
 }
--- a/module/data/settings/bastion-setting.mjs
+++ b/module/data/settings/bastion-setting.mjs
@@ -49,6 +49,6 @@ export default class BastionSetting extends foundry.abstract.DataModel {
       const { basic, special } = CONFIG.DND5E.facilities.advancement;
       BastionSetting.#threshold = Math.min(...Object.keys(basic), ...Object.keys(special));
     }
-    return this.enabled && (actor.system.details?.level >= BastionSetting.#threshold);
+    return this.enabled && actor.system.isCharacter && (actor.system.details?.level >= BastionSetting.#threshold);
   }
 }
--- a/module/data/settings/bastion-setting.mjs
+++ b/module/data/settings/bastion-setting.mjs
@@ -31,7 +31,7 @@ export default class BastionSetting extends foundry.abstract.DataModel {
 
   /**
    * Cached version of the minimum level required to have a bastion.
-   * @param {number}
+   * @type {number}
    */
   static #threshold;
 
--- a/module/documents/actor/bastion.mjs
+++ b/module/documents/actor/bastion.mjs
@@ -1,3 +1,4 @@
+import CalendarData5e from "../../data/calendar/calendar-data.mjs";
 import BastionAttackDialog from "../../applications/actor/bastion-attack-dialog.mjs";
 import BastionAttackMessageData from "../../data/chat-message/bastion-attack-message-data.mjs";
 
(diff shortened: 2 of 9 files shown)


module/data/settings/bastion-setting.mjs
```diff
@@ -24,4 +24,31 @@ export default class BastionSetting extends foundry.abstract.DataModel {
...
     };
   }
+
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Cached version of the minimum level required to have a bastion.
+   * @param {number}
```
Fyorl:
```suggestion
   * @type {number}
```

module/documents/actor/bastion.mjs
```diff
@@ -29,64 +30,84 @@ export default class Bastion {
...
-  async advanceTurn(facility, { duration=7 }={}) {
+  async advanceTurn(facility, { duration=7, performUpdates=true }={}) {
     const { disabled, progress, type } = facility.system;
 
     // Case 1 - No order in progress.
     if ( !progress.max && !disabled ) {
-      await facility.update({ "system.progress.order": "" });
-      if ( type.value === "basic" ) return {}; // Basic facilities do nothing.
-      return { order: "maintain" }; // Special facilities are considered to have been issued the maintain order.
+      const updates = { "system.progress.order": "" };
```
Fyorl: Since `handleTimePassage` is sensitive to whether the delta is empty or not, we can return an empty delta in some cases. An idle facility already has `order: ""` for example.


```suggestion
      const updates = progress.order ? { "system.progress.order": "" } : {};
```

module/documents/actor/bastion.mjs
```diff
@@ -29,64 +30,84 @@ export default class Bastion {
...
+    // Calculate duration if automatic
+    if ( !duration && progress.updated ) {
+      const days = CalendarData5e.dayDifference(
+        game.time.calendar.timeToComponents(progress.updated),
+        game.time.components
+      );
+      if ( days > 0 ) duration = days;
     }
 
     const newProgress = Math.min(progress.value + duration, progress.max);
```
Fyorl, line 100: It's possible for `duration` to still be `null` here.


```suggestion
    duration ??= 0;
    const newProgress = Math.min(progress.value + duration, progress.max);
```

module/data/settings/bastion-setting.mjs
```diff
@@ -24,4 +24,31 @@ export default class BastionSetting extends foundry.abstract.DataModel {
...
+   */
+  static #threshold;
```
Fyorl, line 36: What's the earliest we might cache this? Is it possible a module updates `CONFIG.DND5E.facilities.advancement` later?

module/documents/actor/bastion.mjs
```diff
@@ -29,64 +30,84 @@ export default class Bastion {
...
-      return { order: "maintain" }; // Special facilities are considered to have been issued the maintain order.
+      const updates = { "system.progress.order": "" };
+      if ( performUpdates ) await facility.update(updates);
+      if ( type.value === "basic" ) return { updates }; // Basic facilities do nothing.
+      return { order: "maintain", updates }; // Special facilities are considered to have been issued the maintain order.
+    }
+
+    // Calculate duration if automatic
+    if ( !duration && progress.updated ) {
+      const days = CalendarData5e.dayDifference(
```
Fyorl: Is it not sufficient to pass `timePassageData.midnights` as the `duration`? Do we need to store an `updated` timestamp on the individual facilities?

Fyorl (changes requested):
Thanks for doing this. I think we are able to model Bastions much better now with these changes.

As part of my review, I went over the rules again to re-familiarise myself. That's why you'll see I made a small update to change the behaviour of disabled facilities. Briefly, the rules say that a disabled facility cannot be *used*, but they don't say progress is interrupted. That was a logical leap that isn't actually mentioned anywhere in the rules that I could tell. So without a rule that says 'a disabled facility loses its progress', there's no reason to assume that's the case.

But it revealed an issue in the PR. Because `advanceTurn` is called every day now, we would actually repair a facility in one day rather than one bastion turn.

So for the most part, this PR is actually much more correct than what we had before. All durations in the bastion rules are specified in days, they are totally disconnected from bastion turns. A bastion turn is something a DM calls for, at their discretion, which the book *recommends* is every 7 days.

A bastion turn elapsing does very few things: (a) It issues the Maintain order to the whole Bastion for each character that is not present at the Bastion or unable to communicate with it. This prompts a roll on the events table. (b) It causes damaged facilities to be repaired.

So what we need, I think, is a clear distinction between *day-quantised* events, and *turn-quantised* events. We have the day-quantised events down pretty much already, it's just the turn-quantised events that need some additional work in this PR then.

I would propose then, if you have the calendar enabled, that the 'Advance Bastion Turn' button becomes a 'Maintain Bastion' button instead. It issues the Maintain order to all idle facilities and (later) it can roll on the events table, possibly triggering an attack, which we already have code for. It will also cause damaged facilities to be repaired.

If you don't have the calendar enabled, then the button just works exactly as it did before and it rolls progress + bastion turn all together. That means our `// TODO: Should this advance game.time` is 'no', and we can delete that comment.

module/data/settings/bastion-setting.mjs
arbron: It is cached either when the first character sheet is opened or when time is advanced past midnight, so both would generally be past the `ready` hook so it shouldn't be an issue.

module/documents/actor/bastion.mjs
arbron: My intention here was to avoid double-triggering if time was reversed and then advanced again. Since we don't do that for recovery I'm fine not doing the extra work here.

arbron:
@Fyorl Changed it so the advance turn button adapts depending on calendar settings and facilities are only repaired if a full turn is issued by clicking that button.

module/data/settings/bastion-setting.mjs
```diff
@@ -24,4 +24,31 @@ export default class BastionSetting extends foundry.abstract.DataModel {
...
+   * Determine whether bastions are available for a specific actor.
+   * @param {Actor5e} actor  Actor that may have a bastion.
+   * @returns {boolean}
+   */
+  availableForActor(actor) {
+    if ( !BastionSetting.#threshold ) {
+      const { basic, special } = CONFIG.DND5E.facilities.advancement;
+      BastionSetting.#threshold = Math.min(...Object.keys(basic), ...Object.keys(special));
+    }
+    return this.enabled && (actor.system.details?.level >= BastionSetting.#threshold);
```
Fyorl:
```suggestion
    return this.enabled && system.isCharacter && (actor.system.details?.level >= BastionSetting.#threshold);
```

module/documents/actor/bastion.mjs
Fyorl: Right, we would need a timestamp to handle that, makes sense. Either that or we make recovery/progress reverse when time reverses I suppose.

Fyorl (approved):

Issue #6600, opened 2025-12-31
Add Highest Spell Slot Level to Rolldata


Dewi-Jones:
Right now, it is not easy to get the highest spell slot level in a formula, for example when setting up an activity that consumes spell slots, but is not itself on a spell. (Related: https://github.com/foundryvtt/dnd5e/issues/4568)

**Proposal**
Expand `attributes.spell` to include the maximum possible spell slot level (based on spellcasting progression), and the maximum current spell slot level (if highest level spell slots are expended).

PR #6604, opened 2026-01-02
[#6600] Add spell slot data to `system.attributes.spell.slots`

krbz999:
Since spell slots are prepared in the `Actor5e` class and as the very last thing, and since this extra data needs to be prepared after slots, the only way I could think of was to call `AttributesFields.prepareSpellSlotData` (which is arguably a very terrible name) in the `Actor5e` class as well, gated behind the actor subtype.

Leaving as draft for discussion etc etc.


Fyorl (commented):
I think if we want to do this in data prep there's not really a better place than what you have. I can't necessarily think of a better name for the method. The only thing I would tweak is to use `this.isCreature()` rather than testing `character` and `npc` specifically.

Something you could consider is a similar design to `HitDice`, which I believe you also authored mostly. The `largestAvailable`, etc. concepts are pretty similar to what we have in `HitDice`.

One final consideration is that class features typically (not always) when referring to 'of a level for which you have spell slots' refer to the class' single-classed spell slot table rather than referring to the *actual* spell slots the character has available. It seems somewhat related to this issue that if you're looking for the 'largest' spell slot, then you might actually need the largest spell slot *for a given class* some of the time. For example, the calculation we do in `ItemChoiceFlow#_maxSpellSlotLevel`.

But they are not totally related. The maximum class spell slot level is available as soon as item prep finishes, rather than at the end of `Actor5e#prepareData`, and we would probably want that data on `@classes.*.spellcasting.slots` rather than under `@attributes`. Also, the concept of `largestAvailable` wouldn't exist in such a context since actual spellcasting slots are shared between classes.

Whether we address it in this PR or later, I think it needs to be kept in mind for this work.

PR #6609, opened 2026-01-07
Feature: Add "Available (No Cantrips)" option to Item Choice Advancement

CATFOOL:
This PR adds a new option Available (No Cantrips) (availableNoCantrips) to the Spell Level restriction in the Item Choice Advancement configuration.

Currently, the Available option allows players to choose spells of any level for which they have spell slots, but this includes Cantrips (0-level spells). There are cases (e.g., specific feats or class features) where a player should be allowed to choose from any leveled spell they can cast, but not Cantrips. This new option fills that gap.

### DIFF
--- a/module/applications/advancement/item-choice-flow.mjs
+++ b/module/applications/advancement/item-choice-flow.mjs
@@ -129,7 +129,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
 
     const spellLevel = config.restriction.level;
     const maxSlot = this._maxSpellSlotLevel();
-    const minSpellLevel = spellLevel === "availableNoCantrips" ? 1 : 0;
+    const minSlot = spellLevel === "availableNoCantrips" ? 1 : 0;
     const validateSpellLevel = (config.type === "spell") && ["available", "availableNoCantrips"].includes(spellLevel);
 
     const added = [];
@@ -154,7 +154,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
           this.advancement.actor, { added, removed, level: this.featureLevel }
         ) === true);
         const validSpell = !validateSpellLevel
-          || ((item.system.level <= maxSlot) && (item.system.level >= minSpellLevel));
+          || ((item.system.level >= minSlot) && (item.system.level <= maxSlot));
         if ( validFeature && validSpell ) {
           const data = {
             id, img, name, uuid,
@@ -347,8 +347,8 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
     const spellLevel = this.advancement.configuration.restriction.level;
     if ( (this.advancement.configuration.type === "spell") && ["available", "availableNoCantrips"].includes(spellLevel) ) {
       const maxSlot = this._maxSpellSlotLevel();
-      const minLevel = spellLevel === "availableNoCantrips" ? 1 : 0;
-      if ( (item.system.level > maxSlot) || (item.system.level < minLevel) ) {
+      const minSlot = spellLevel === "availableNoCantrips" ? 1 : 0;
+      if ( (item.system.level < minSlot) || (item.system.level > maxSlot) ) {
         ui.notifications.error("DND5E.ADVANCEMENT.ItemChoice.Warning.SpellLevelAvailable", {
           format: { level: CONFIG.DND5E.spellLevels[maxSlot] }
         });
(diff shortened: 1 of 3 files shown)


module/applications/advancement/item-choice-flow.mjs
```diff
@@ -140,7 +142,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
         const validFeature = !i.system.validatePrerequisites || (i.system.validatePrerequisites(
           this.advancement.actor, { added, removed, level: this.featureLevel }
         ) === true);
-        const validSpell = !validateSpellLevel || (i.system.level <= maxSlot);
+        const validSpell = !validateSpellLevel || (i.system.level <= maxSlot && i.system.level >= minSpellLevel);
```
arbron:
```suggestion
        const validSpell = !validateSpellLevel || ((i.system.level >= minSlot) && (i.system.level <= maxSlot));
```

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -128,7 +128,9 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
 
     const spellLevel = this.advancement.configuration.restriction.level;
     const maxSlot = this._maxSpellSlotLevel();
-    const validateSpellLevel = (this.advancement.configuration.type === "spell") && (spellLevel === "available");
+    const minSpellLevel = spellLevel === "availableNoCantrips" ? 1 : 0;
```
arbron: I'd probably just name this `minSlot` to match with the existing `maxSlot`. Alternatively you could rename the existing `maxSlot` to `maxLevel` if you think that would make more sense.

```suggestion
    const minSlot = spellLevel === "availableNoCantrips" ? 1 : 0;
```

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -214,9 +216,12 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
     if ( config.type === "feat" ) {
       filters.locked.arbitrary = [{ k: "system.prerequisites.level", o: "lte", v: this.featureLevel }];
     } else if ( (config.type === "spell") && (config.restriction.level !== "") ) {
+      const isAvailable = ["available", "availableNoCantrips"].includes(config.restriction.level);
       filters.locked.additional.level = {
-        min: config.restriction.level === "available" ? undefined : Number(config.restriction.level),
-        max: config.restriction.level === "available" ? this._maxSpellSlotLevel() : Number(config.restriction.level)
+        min: config.restriction.level === "availableNoCantrips" ? 1
+          : config.restriction.level === "available" ? 0
+          : Number(config.restriction.level),
```
arbron: We should probably leave the min level as undefined in cases where we don't care about it so that players are still free to set the filter to a higher level if they want to only look at higher levels spells.

```suggestion
          : config.restriction.level === "available" ? undefined : Number(config.restriction.level),
```

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -308,9 +313,10 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
 
     // If spell level is restricted to available level, ensure the spell is of the appropriate level
     const spellLevel = this.advancement.configuration.restriction.level;
-    if ( (this.advancement.configuration.type === "spell") && spellLevel === "available" ) {
+    if ( (this.advancement.configuration.type === "spell") && ["available", "availableNoCantrips"].includes(spellLevel) ) {
       const maxSlot = this._maxSpellSlotLevel();
-      if ( item.system.level > maxSlot ) {
+      const minLevel = spellLevel === "availableNoCantrips" ? 1 : 0;
+      if ( item.system.level > maxSlot || item.system.level < minLevel ) {
```
arbron: Same naming comment as above.

```suggestion
      const minSlot = spellLevel === "availableNoCantrips" ? 1 : 0;
      if ( (item.system.level < minSlot) || (item.system.level > maxSlot)  ) {
```

arbron (changes requested):

arbron:
There is also going to be a re-writing of `ItemChoiceFlow` very shortly, so be prepared to re-apply these changes to `ItemChoiceFlowV2` once https://github.com/foundryvtt/dnd5e/pull/6249 is merged.

Fyorl (approved):

Issue #6608, opened 2026-01-07
Allow for rolldata in attack/saveDC overrides on cast activities


krbz999:
Working on a premium module which has a feature that grants some spells (using Cast activities) whose DC scales with some custom stat we add to roll data, as well as `@prof`.

In the Cast activity, we can specify only a flat numeric value for the DC override, which creates an enchantment on the granted spells with `activities[save].save.dc.formula` as the attribute key. However, this field *does* allow rolldata, so I don't foresee there being an issue with allowing it in the cast activity.

krbz999:
Relatedly, the field in question is currently a `NumberField` but allows non-integer values. I'm wondering if it was always meant to be a `StringField` or `FormulaField` and this was just a hasty oversight.

https://github.com/foundryvtt/dnd5e/blob/5.2.x/module/data/activity/cast-data.mjs#L25

krbz999:
Probably resolved with v14, though the above Q re: integer still stands

PR #6616, opened 2026-01-08
[#6608] Allow for rolldata in cast activity spell overrides

krbz999:


Fyorl (approved):
Still need to make this change to even allow entering formula data. But now we support v14-only these fields will apply directly to the Summon activity's `flat` fields without us having to run `simplifyBonus` first.

Issue #6610, opened 2026-01-07
Add `data-no-toc` to header elements of NPC statblock embeds


krbz999:
Doesn't really seem necessary to have the monster's name, `Actions`, and `Bonus Actions` in the journal entry toc.

PR #6617, opened 2026-01-08
[#6610] Add `data-no-toc` to header elements of npc stat block embeds

krbz999:
Using a text page with title disabled and nothing but an embed...

Before:

image

After:

image


arbron:
@krbz999 Any chance of giving this a quick rebase onto the latest `5.3.x`?

PR #6632, opened 2026-01-16
[#6630] Implement `_applyChangeSubtract` for `FormulaField`

### LINKED ISSUE #6630: Add `_applyChangeSubtract` method to `FormulaField` to properly handle new change mode

arbron:
The subtract change method will subtract the whole provided value in parenthesis to properly handle more complex formulas (e.g. `1 - (3 + @prof)`).

Closes #6630

### DIFF
--- a/module/data/fields/formula-field.mjs
+++ b/module/data/fields/formula-field.mjs
@@ -50,7 +50,7 @@ export default class FormulaField extends foundry.data.fields.StringField {
 
   /** @override */
   _applyChangeSubtract(value, delta, model, change) {
-    if ( !value ) return delta;
+    if ( !value ) return `-(${delta})`;
     return `${value} - (${delta})`;
   }
 


module/data/fields/formula-field.mjs
```diff
@@ -48,6 +48,14 @@ export default class FormulaField extends foundry.data.fields.StringField {
 
   /* -------------------------------------------- */
 
+  /** @override */
+  _applyChangeSubtract(value, delta, model, change) {
+    if ( !value ) return delta;
+    return `${value} - (${delta})`;
+  }
```
krbz999, line 55: shouldnt this be `-delta`

arbron: I was thinking if you have an delta like `-5 + 5` subtracted from `1`, then using that would result in `1 - -5 + 5 = 11` but it should be `1 - (-5 + 5) = 1`.

krbz999: My bad, I was referring to the line above; if `!value`, would a delta of `5` not result in `5` instead of the expected value of `-5`?

arbron: Aw yeah, that makes sense.

Fyorl (approved):

PR #6634, opened 2026-01-16
[#6633] Use `<formula-input>` for `FormulaField`

### LINKED ISSUE #6633: Set `FormulaField` to use `<formula-input>` by default

arbron:
`FormulaField` will automatically produce a `<formula-input>` unless it has choices set. Also styled `<formula-input>` to match the system's form styling.

Formula Input

Closes #6633

### DIFF
--- a/less/v2/forms.less
+++ b/less/v2/forms.less
@@ -404,6 +404,8 @@
       padding-right: 20px;
     }
 
+    &.split-group .form-group.formula-input { flex: 1 1 min-content; }
+
     /* Full-Width Groups */
     &.full-width {
       padding: 6px;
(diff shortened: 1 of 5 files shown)


less/v2/forms.less
```diff
@@ -459,6 +461,32 @@
     font-size: var(--font-size-11);
   }
 
+  formula-input {
+    overflow: hidden;
+    &:focus-within { outline: none; }
+    &:enabled {
+      &:focus-within, &:hover {
```
Fyorl, line 468: `:focus-within` I think is cheaper than `:has(> :focus-visible)` but maybe they have different semantics?

PR #6658, opened 2026-01-23
[#6121, #6646] Add support for multiple AC formulas

### LINKED ISSUE #6121: Add another property to armor calculations that allow/disallow shields to be added
I was creating a new armor calculation for something similar to unarmored defense, the only change being that the class specifies the **bonus will still apply if the character is using a shield.** Upon further investigation, Unarmored Defense doesn't currently prevent an Actor from equipping a shield and gaining the bonus, despite having a specific armor calculation. 

Image
Image

```
/**
 * Common armor class calculations.
 * @enum {{ label: string, [formula]: string }}
 */
DND5E.armorClasses = {
  flat: {
    label: "DND5E.ArmorClassFlat",
    formula: "@attributes.ac.flat"
  },
  natural: {
    label: "DND5E.ArmorClassNatural",
    formula: "@attributes.ac.flat"
  },
  default: {
    label: "DND5E.ArmorClassEquipment",
    formula: "@attributes.ac.armor + @attributes.ac.dex"
  },
  unarmoredMonk: {
    label: "DND5E.ArmorClassUnarmoredMonk",
    formula: "10 + @abilities.dex.mod + @abilities.wis.mod"
  },
  unarmoredBarb: {
    label: "DND5E.ArmorClassUnarmoredBarbarian",
    formula: "10 + @abilities.dex.mod + @abilities.con.mod"
  },
  unarmoredBard: {
    label: "DND5E.ArmorClassUnarmoredBard",
    formula: "10 + @abilities.dex.mod + @abilities.cha.mod"
  },
  custom: {
    label: "DND5E.ArmorClassCustom"
  }
};
preLocalize("armorClasses", { key: "label" });

```

### LINKED ISSUE #6646: Add AC “floor” AE flag (`flags.dnd5e.minimumACBase`) that does not override `system.attributes.ac.calc`
To support a particular feature in Ember and for D&D5e in general, we need a system-supported way to implement features like “Your AC can’t be lower than 16” via Active Effects without forcing `system.attributes.ac.calc = custom`. 

Today, the only reliable approach is overriding calc mode + using a `max()` formula, which breaks/supersedes other calc modes (e.g., Monk/Barbarian Unarmored Defense, Mage Armor, Natural Armor modes, etc.).

I am requesting a new AE-usable flag that applies a **minimum base AC** after the current calc mode computes base AC, while still allowing shields and other bonuses to stack normally.

## Steps to reproduce (current limitation)

1. Create an Active Effect intended to enforce “AC minimum 16”.
2. Implement via:
   * `system.attributes.ac.calc` → override → `custom`
   * `system.attributes.ac.formula` → custom → `max(16, @attributes.ac.armor + @attributes.ac.dex)`
3. Observe: actors using other AC calculation modes (Mage Armor / Unarmored Defense / etc.) are forced into `custom` and lose intended behavior.

## Proposal

Add `flags.dnd5e.minimumACBase` (or `flags.dnd5e.acFloor` or a attribute name you prefer)

* Type: number (optionally allow formula strings)
* Behavior: system applies `ac.base = max(ac.base, minimumACBase)` **after** calc-mode base computation, **before** additive modifiers.

Conceptually:

* Compute base via current `ac.calc`
* Apply floor to base
* Continue normal pipeline so `ac.shield`, `ac.bonus`, `ac.cover` still add on top

## Acceptance criteria

1. Setting `flags.dnd5e.minimumACBase` via AE enforces a minimum **base** AC.
2. Does not modify/override `system.attributes.ac.calc`.
3. Works across existing calc modes (default armor, Mage Armor, Monk/Barbarian UD, Natural Armor, etc.).
4. Shield/bonus/cover continue to stack normally on top of the floored base.
5. Multiple floors stack by taking the highest value.
6. Documented in the Active Effect Guide.

## Test cases

* No armor, Dex +2 → base 12 → floor 16 → final 16
* Leather + Dex +3 → base 14 → floor 16 → final 16
* Plate → base 18 → floor 16 → final 18
* Mage Armor yields 14 → floor 16 → final 16
* Monk UD yields 15 → floor 16 → final 16
* Floor 16 + shield +2 + ring +1 → final 19
* Floors 16 and 17 → effective 17

arbron:
Modifies how armor class data is stored to use multiple formulas rather than only having a single formula that is evaluated. This change means that all enabled formulas are evaluated fully and the one that provides the highest base AC is used.

This also allows armor class formulas to define whether they are usable when the character is wearing armor or wielding a shield. This makes all of the existing unarmoed defense formulas work as expected.

In order to ensure that the previous `flat` calculation properly overrides the full AC it has been moved to a separate `override` value. The existing `flat` input is retained to provide a place for players to manually enter a value that can be used in formulas (like the Natural Armor formula).

Armor Class Configuration Formulas

Migrations are in place to move the existing actor data to use the new data structure and shims are added to ensure existing active effects that target AC will continue working. The removed `calc` and `formula` fields are still set by the winning calculation so no shims are required.

Closes #6121
Closes #6646

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -826,22 +826,74 @@
   }
 },
 "DND5E.Armor": "Armor",
+
+"DND5E.ARMORCLASS": {
+  "Action": {
+    "Configure": "Configure Armor",
+    "CreateFormula": "Create Custom Formula",
+    "DeleteFormula": "Delete Custom Formula"
+  },
+  "Calculation": {
+    "Armored": "Equipped Armor",
+    "Draconic": "Draconic Resilience",
+    "Label": "Calculation",
+    "Mage": "Mage Armor",
+    "Natural": "Natural Armor",
+    "Unarmored": "Unarmored",
+    "UnarmoredBarbarian": "Unarmored Defense (Barbarian)",
+    "UnarmoredBard": "Unarmored Defense (Bard)",
+    "UnarmoredMonk": "Unarmored Defense (Monk)"
+  },
+  "Condition": {
+    "Positive": "Must have {type}",
+    "Negative": "Must not have {type}"
+  },
+  "Configuration": "Configuration",
+  "CurrentFormula": "Current Formula",
+  "CustomFormulas": {
+    "Empty": "No custom formulas. Use the <i class=\"fas fa-plus\"></i> button above to create one."
+  },
+  "FIELDS": {
+    "attributes": {
+      "ac": {
+        "flat": {
+          "hint": "Number that can be used in an armor class by using @attributes.ac.flat in the formula.",
+          "label": "Flat Value"
+        },
+        "formulas": {
+          "element": {
+            "armored": {
+              "label": "Armored"
+            },
+            "formula": {
+              "label": "Formula"
+            },
+            "label": {
+              "label": "Label"
+            },
+            "shielded": {
+              "label": "Shielded"
+            }
+          },
+          "label": "Formulas"
+        },
+        "label": "Armor Class",
+        "override": {
+          "hint": "Value that will be used in place of calulated armor class, ignoring cover, shield, and any other bonuses.",
+          "label": "Override"
+        },
+        "selectedFormulas": {
+          "hint": "These formulas will be used when calculating armor class.",
+          "label": "Selected Formulas"
+        }
+      }
+    }
+  },
+  "Flat": "Flat",
+  "Formula": "Formula"
+},
+
 "DND5E.ArmorClass": "Armor Class",
-"DND5E.ArmorClassEquipment": "Equipped Armor",
-"DND5E.ArmorClassFlat": "Flat",
-"DND5E.ArmorClassUnarmored": "Unarmored",
-"DND5E.ArmorClassNatural": "Natural Armor",
-"DND5E.ArmorClassMage": "Mage Armor",
-"DND5E.ArmorClassMotionless": "Armor Class while Motionless",
-"DND5E.ArmorClassDraconic": "Draconic Resilience",
-"DND5E.ArmorClassUnarmoredMonk": "Unarmored Defense (Monk)",
-"DND5E.ArmorClassUnarmoredBarbarian": "Unarmored Defense (Barbarian)",
-"DND5E.ArmorClassUnarmoredBard": "Unarmored Defense (Bard)",
-"DND5E.ArmorClassCustom": "Custom Formula",
-"DND5E.ArmorConfig": "Configure Armor",
-"DND5E.ArmorConfigHint": "Fill in the above box to override automatically calculated armor class.",
-"DND5E.ArmorClassCalculation": "Calculation",
-"DND5E.ArmorClassFormula": "Formula",
 "DND5E.ArmorHeavyProficiency": "Heavy",
 "DND5E.ArmorLightProficiency": "Light",
 "DND5E.ArmorMediumProficiency": "Medium",
--- a/lang/en.json
+++ b/lang/en.json
@@ -934,7 +934,9 @@
         "formulas": {
           "element": {
             "armored": {
-              "label": "Armored"
+              "excluded": "Not Armored",
+              "label": "Armored",
+              "required": "Armor Required"
             },
             "formula": {
               "label": "Formula"
@@ -943,7 +945,9 @@
               "label": "Label"
             },
             "shielded": {
-              "label": "Shielded"
+              "excluded": "No Shield",
+              "label": "Shielded",
+              "required": "Shield Required"
             }
           },
           "label": "Formulas"
--- a/lang/en.json
+++ b/lang/en.json
@@ -897,22 +897,74 @@
   }
 },
 "DND5E.Armor": "Armor",
+
+"DND5E.ARMORCLASS": {
+  "Action": {
+    "Configure": "Configure Armor",
+    "CreateFormula": "Create Custom Formula",
+    "DeleteFormula": "Delete Custom Formula"
+  },
+  "Calculation": {
+    "Armored": "Equipped Armor",
+    "Draconic": "Draconic Resilience",
+    "Label": "Calculation",
+    "Mage": "Mage Armor",
+    "Natural": "Natural Armor",
+    "Unarmored": "Unarmored",
+    "UnarmoredBarbarian": "Unarmored Defense (Barbarian)",
+    "UnarmoredBard": "Unarmored Defense (Bard)",
+    "UnarmoredMonk": "Unarmored Defense (Monk)"
+  },
+  "Condition": {
+    "Positive": "Must have {type}",
+    "Negative": "Must not have {type}"
+  },
+  "Configuration": "Configuration",
+  "CurrentFormula": "Current Formula",
+  "CustomFormulas": {
+    "Empty": "No custom formulas. Use the <i class=\"fas fa-plus\"></i> button above to create one."
+  },
+  "FIELDS": {
+    "attributes": {
+      "ac": {
+        "flat": {
+          "hint": "Number that can be used in an armor class by using @attributes.ac.flat in the formula.",
+          "label": "Flat Value"
+        },
+        "formulas": {
+          "element": {
+            "armored": {
+              "label": "Armored"
+            },
+            "formula": {
+              "label": "Formula"
+            },
+            "label": {
+              "label": "Label"
+            },
+            "shielded": {
+              "label": "Shielded"
+            }
+          },
+          "label": "Formulas"
+        },
+        "label": "Armor Class",
+        "override": {
+          "hint": "Value that will be used in place of calulated armor class, ignoring cover, shield, and any other bonuses.",
+          "label": "Override"
+        },
+        "selectedFormulas": {
+          "hint": "These formulas will be used when calculating armor class.",
+          "label": "Selected Formulas"
+        }
+      }
+    }
+  },
+  "Flat": "Flat",
+  "Formula": "Formula"
+},
+
 "DND5E.ArmorClass": "Armor Class",
-"DND5E.ArmorClassEquipment": "Equipped Armor",
-"DND5E.ArmorClassFlat": "Flat",
-"DND5E.ArmorClassUnarmored": "Unarmored",
-"DND5E.ArmorClassNatural": "Natural Armor",
-"DND5E.ArmorClassMage": "Mage Armor",
-"DND5E.ArmorClassMotionless": "Armor Class while Motionless",
-"DND5E.ArmorClassDraconic": "Draconic Resilience",
-"DND5E.ArmorClassUnarmoredMonk": "Unarmored Defense (Monk)",
-"DND5E.ArmorClassUnarmoredBarbarian": "Unarmored Defense (Barbarian)",
-"DND5E.ArmorClassUnarmoredBard": "Unarmored Defense (Bard)",
-"DND5E.ArmorClassCustom": "Custom Formula",
-"DND5E.ArmorConfig": "Configure Armor",
-"DND5E.ArmorConfigHint": "Fill in the above box to override automatically calculated armor class.",
-"DND5E.ArmorClassCalculation": "Calculation",
-"DND5E.ArmorClassFormula": "Formula",
 "DND5E.ArmorHeavyProficiency": "Heavy",
 "DND5E.ArmorLightProficiency": "Light",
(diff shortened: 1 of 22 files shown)


arbron:
Currently a draft to show the change to the data models and migrations required. Will need to re-write the AC configuration dialog before this can be merged.

module/applications/property-attribution.mjs
```diff
@@ -102,11 +103,13 @@ export default class PropertyAttribution extends Application5e {
...
+    } else if ( property.startsWith("attributes.ac.clamped.") ) {
+      return CONFIG.DND5E.abilities[parts[3]]?.label ?? property;
     } else if ( (property === "attributes.ac.dex") && CONFIG.DND5E.abilities.dex ) {
       return CONFIG.DND5E.abilities.dex.label;
     } else if ( (parts[0] === "prof") || (property === "attributes.prof") ) {
-      return _loc("DND5E.Proficiency");
+      return game.i18n.localize("DND5E.Proficiency");
     }
-    return property;
+    return getHumanReadableAttributeLabel(property);
```
Fyorl, line 113: Maybe related to this, not sure. When testing I observed that NPCs using `natural` display attribution a bit strangely, even though the resulting value is correct.

Image

module/data/actor/templates/attributes.mjs
```diff
@@ -99,9 +120,42 @@ export default class AttributesFields {
...
+    switch ( ac.calc ) {
+      case "custom":
+        ac.formulas ??= [];
+        ac.formulas.push({ formula: ac.formula });
+        break;
+      case "flat":
+        ac.override = ac.flat;
+        break;
+      case "default": break;
+      default:
```
Fyorl, line 166: I don't know if any official monsters do this, and if they do, it's probably a mistake, but it might be possible for a creature to have a natural armour value less than what it would have with 10 + dex. In those cases, after this PR, its AC would no longer match the stat block. Adding a `natural` case should be enough:

```js
case "natural":
  ac.selectedFormulas = ["natural"];
  break;
```

module/data/actor/templates/attributes.mjs
```diff
@@ -122,6 +176,7 @@ export default class AttributesFields {
   static prepareBaseArmorClass() {
     const ac = this.attributes.ac;
     ac.armor = 10;
+    ac.base = -Infinity;
     ac.shield = ac.cover = 0;
     ac.min = ac.bonus = "";
```
Fyorl: I guess these can be `persist: false` fields now.

module/data/actor/templates/attributes.mjs
```diff
@@ -147,14 +202,13 @@ export default class AttributesFields {
...
 
-    // Apply automatic migrations for older data structures
-    let cfg = CONFIG.DND5E.armorClasses[ac.calc];
-    if ( !cfg ) {
-      ac.calc = "flat";
-      if ( Number.isNumeric(ac.value) ) ac.flat = Number(ac.value);
-      cfg = CONFIG.DND5E.armorClasses.flat;
-    }
+    // Add selected formulas
+    const baseFormulas = Object.entries(CONFIG.DND5E.armorClasses)
```
Fyorl: `foundry.utils.iterateEntries` can avoid the intermediate array creation.

module/data/actor/templates/attributes.mjs
```diff
@@ -164,74 +218,74 @@ export default class AttributesFields {
...
 
-    // Compute cover.
-    ac.cover = Math.max(ac.cover, this.parent.coverBonus);
+    // If armor is equipped, prepare clamped abilities
+    const isHeavy = ac.equippedArmor?.system.type.value === "heavy";
+    ac.clamped = Object.entries(this.abilities).reduce((obj, [k, v]) => {
+      obj[k] = isHeavy ? 0 : Math.min(v.mod, ac.equippedArmor?.system.armor.dex ?? Infinity);
+      return obj;
+    }, {});
+    ac.dex = ac.clamped.dex;
```
Fyorl, line 267: Don't really understand this `ac.clamped` and how it differs from `ac.dex`. We are clamping every ability score by the armour's max dex bonus? Doesn't really make sense to me.

module/data/actor/templates/attributes.mjs
```diff
@@ -164,74 +218,74 @@ export default class AttributesFields {
...
+      return true;
+    });
+
+    for ( const [index, config] of validFormulas.entries() ) {
+      try {
+        const replaced = replaceFormulaData(config.formula, rollData, {
+          actor: this, missing: null, property: game.i18n.localize("DND5E.ArmorClass")
+        });
+        const result = replaced ? new Roll(replaced).evaluateSync().total : 0;
+        if ( result > ac.base ) {
```
Fyorl, line 282: If an AE sets `ac.base` to something high, like `25`, then I think `ac.formula/calc/label` are never set, but it looks like our code mostly assumes they are always set.

module/data/actor/templates/attributes.mjs
```diff
@@ -99,9 +120,42 @@ export default class AttributesFields {
...
   /* -------------------------------------------- */
 
+  /**
+   * Migrate the old single armor formula into formulas.
+   * @param {object} [source]  The source attributes object.
+   * @internal
+   */
+  static _migrateArmorClass(source) {
+    const ac = source?.ac ?? {};
+    if ( Number.isNumeric(ac.value) && !CONFIG.DND5E.armorClasses[ac.calc] ) {
```
Fyorl, line 148: Since we renamed `default` -> `armored`, I think every creature with `ac.calc: "default"` is going to end up with an override, no?

module/data/actor/templates/attributes.mjs
```diff
@@ -164,74 +218,74 @@ export default class AttributesFields {
...
+      ac.base = ac.flat ?? 0;
+      ac.calc = "natural";
+      ac.formula = "@attributes.ac.flat";
+    }
+
+    ac.cover = Math.max(ac.cover, this.parent.coverBonus);
     ac.min = simplifyBonus(ac.min, rollData);
     ac.bonus = simplifyBonus(ac.bonus, rollData);
-    ac.value = Math.max(ac.min, ac.base + ac.shield + ac.bonus + ac.cover);
+    if ( ac.override ) ac.value = ac.override;
```
Fyorl: Allows for `0` as a valid `ac.override`.
```suggestion
    if ( Number.isFinite(ac.override) ) ac.value = ac.override;
```

lang/en.json
```diff
@@ -868,22 +880,74 @@
...
+            },
+            "shielded": {
+              "label": "Shielded"
+            }
+          },
+          "label": "Formulas"
+        },
+        "label": "Armor Class",
+        "override": {
+          "hint": "Value that will be used in place of calulated armor class, ignoring cover, shield, and any other bonuses.",
```
```suggestion
          "hint": "Value that will be used in place of calculated armor class, ignoring cover, shield, and any other bonuses.",
```

module/documents/actor/actor.mjs
```diff
@@ -2651,51 +2596,36 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const cfg = CONFIG.DND5E.armorClasses[ac.calc];
     const attribution = [];
 
-    if ( ac.calc === "flat" ) {
+    if ( ac.override ) {
```
```suggestion
    if ( Number.isFinite(ac.override) ) {
```

module/documents/active-effect.mjs
```diff
@@ -61,9 +61,11 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
   /**
    * Active effect fields that should be redirected to another field, optionally with a compatibility warning.
    * Optional warning object contains options passed to `foundry.utils.logCompatibilityWarning`.
-   * @type {Record<string, { key: string, [warning]: object }>}
+   * @type {Record<string, { key: string, [mode]: CONST.ACTIVE_EFFECT_MODES, [warning]: object }>}
    */
   static SHIM_FIELDS = {
+    "system.attributes.ac.calc": { key: "system.attributes.ac.selectedFormulas", mode: CONST.ACTIVE_EFFECT_MODES.ADD },
+    "system.attributes.ac.formula": { key: "system.attributes.ac.formulas", mode: CONST.ACTIVE_EFFECT_MODES.ADD },
```
Fyorl: I believe this would just append the raw string to the formulas array, which is an array of objects.

Fyorl (changes requested):
With regards to the AE shim, I wonder if it would be neater to not shim those fields but instead add them as `persist: false` fields. Then we treat them as another formula to add to the pile.

```js
if ( ac.calc === "flat" ) ac.override ??= ac.flat;
else if ( ac.calc === "custom" ) ac.formulas.push({ formula: ac.formula });
else if ( CONFIG.DND5E.armorClasses[ac.calc]?.formula ) {
  ac.formulas.push({ ...CONFIG.DND5E.armorClasses[ac.calc], id: ac.calc });
}
```

It maintains the winner-takes-all behaviour of these AEs prior to this PR, but does mean we are stuck with this legacy path forever. It's one option to address the issues with the current shim that I flagged below, and feels neater to me.

It does mean we still need a solution going forward for AEs that want to add a new AC calculation. For that I would propose that they target `system.attributes.ac.formulas | ADD | <formula>` and we intercept it at application time to transform it to `{ formula: <formula> }`. Then our conditional AE work will take care of the armoured/shielded conditions.

Finally then, that leaves the question of do we need these `armoured`/`shielded` fields in the model at all? I would think they are still useful for manual input of AC formulae (i.e. not via AE), but they are currently not exposed in the UI. I'm not sure but it looks like they were maybe intended to be all along anyway.

module/data/actor/templates/attributes.mjs
arbron: There have been requests to support 3rd party content where the AC is calculated using a different ability than dexterity but it is still clamped like dex. This adds support for this by not special-casing dexterity.

module/data/actor/templates/attributes.mjs
arbron: `ac.value` is what we used for AC back before we added AC formulas the first time. This should only apply to super-ancient data.

arbron:
> Finally then, that leaves the question of do we need these `armoured`/`shielded` fields in the model at all? I would think they are still useful for manual input of AC formulae (i.e. not via AE), but they are currently not exposed in the UI. I'm not sure but it looks like they were maybe intended to be all along anyway.

I had left them off in want of a good three-state checkbox-style input, but I can just add them using dropdowns.

module/data/actor/templates/attributes.mjs
```diff
@@ -163,14 +223,19 @@ export default class AttributesFields {
...
-    let cfg = CONFIG.DND5E.armorClasses[ac.calc];
-    if ( !cfg ) {
-      ac.calc = "flat";
-      if ( Number.isNumeric(ac.value) ) ac.flat = Number(ac.value);
-      cfg = CONFIG.DND5E.armorClasses.flat;
-    }
+    // Add formulas set by old-style AEs
+    if ( ac.calc === "flat" ) ac.override = ac.flat;
+    else if ( (ac.calc === "custom") && ac.formula ) ac.formulas.push({ formula: ac.formula });
+    else if ( ac.calc in CONFIG.DND5E.armorClasses ) ac.selectedFormulas.push(ac.calc);
```
Fyorl:
```suggestion
    else if ( ac.calc in CONFIG.DND5E.armorClasses ) ac.selectedFormulas.add(ac.calc);
```

Fyorl (changes requested):
Still need a solution for this:
> It does mean we still need a solution going forward for AEs that want to add a new AC calculation. For that I would propose that they target `system.attributes.ac.formulas | ADD | <formula>` and we intercept it at application time to transform it to `{ formula: <formula> }`. Then our conditional AE work will take care of the armoured/shielded conditions.

Unless you plan to defer or prefer requiring JSON here.

module/data/actor/fields/ac-formulas-field.mjs
```diff
@@ -0,0 +1,35 @@
...
+    }));
+  }
+
+  /* -------------------------------------------- */
+  /*  Active Effect Integration                   */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  applyChange(value, model, change, options) {
+    if ( (change.type === "add") && (foundry.utils.getType(change.value) === "string") ) change.value = {
```
Fyorl: Plain `typeof` is a bit faster.
```suggestion
    if ( (change.type === "add") && (typeof change.value === "string") ) change.value = {
```

Fyorl (approved):
Nice work.

PR #6662, opened 2026-01-26
[#3939] Add Maximum Hit Points damage & healing type

### LINKED ISSUE #3939: Add max HP healing/damage type
Currently we have normal HP and temp HP healing types, but there are certain spells like Aid and Heroes' Feast that grant a temporary increase in max HP. A new max HP healing type would allow for rolling for those spells, and then when applied using the damage application interface could increase the temp max HP on the character by that amount and also increase their current HP to match.

arbron:
Adds a new option to both damage and healing options that applies to the target's temporary maximum hit points. This option is responsive to what is applying it, so when used in a Heal activity it increases the maximum hit points (and current hit points by the same amount) while in any other activity it reduces the max (but doesn't modify current HP except by clamping).

Closes #3939

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -900,12 +900,13 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       }
 
       // Negate healing types
-      if ( (options.invertHealing !== false) && (d.type === "healing") ) damageMultiplier *= -1;
+      if ( (options.invertHealing !== false) && ((d.type === "healing")
+        || ((d.type === "maximum") && (treatAs === "healing"))) ) damageMultiplier *= -1;
 
       d.value = d.value * damageMultiplier;
       d.active.multiplier = (d.active.multiplier ?? 1) * damageMultiplier;
       if ( d.type === "temphp" ) damages.temp += d.value;
-      else if ( d.type === "maximum" ) damages.tempMax += d.value * (treatAs === "healing" ? -1 : 1);
+      else if ( d.type === "maximum" ) damages.tempMax += d.value;
       else damages.amount += d.value;
     });
 


Zanderaf:
Any shot we could get some sort of "sources" of maximum hit point reduction / increase as some form of book-keeping? Maybe some sort of display in the Hit Point dialog on actor sheets?

Would help to know what specific effect to end with something like Greater Restoration:
`• One effect reducing the target's hit point maximum`

module/data/item/weapon.mjs
```diff
@@ -545,12 +545,11 @@ export default class WeaponData extends ItemDataModel.mixin(
...
-      return {
-        value, label,
-        selected: context.source.damage.base.types.includes?.(value) ?? context.source.damage.base.types.has(value)
-      };
-    });
+    context.damageTypes = Object.entries(CONFIG.DND5E.damageTypes).map(([value, { label }]) => ({ value, label }));
+    if ( this.type.value === "natural" ) context.damageTypes.push(
+      { rule: true },
+      { value: "maximum", label: "DND5E.HEAL.Type.Maximum" }
+    );
```
Fyorl: Probably fine to just add this indiscriminately.

module/documents/actor/actor.mjs
```diff
@@ -885,10 +891,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       d.value = d.value * damageMultiplier;
       d.active.multiplier = (d.active.multiplier ?? 1) * damageMultiplier;
       if ( d.type === "temphp" ) damages.temp += d.value;
+      else if ( d.type === "maximum" ) damages.tempMax += d.value * (treatAs === "healing" ? -1 : 1);
       else damages.amount += d.value;
     });
 
     damages.amount = damages.amount > 0 ? Math.floor(damages.amount) : Math.ceil(damages.amount);
+    if ( damages.tempMax < 0 ) damages.amount += damages.tempMax;
```
Fyorl: `damages.amount` would become unrounded again if `damages.tempMax` is also a decimal. Also, not sure if it's correct to add maximum hp damage to the total damage amount. If a creature takes 5 damage and 5 maximum hp damage, then I think they've still only taken 5 damage overall, no?

Should it be something like:
```js
if ( damages.tempMax < 0 ) damages.amount = Math.max(damages.amount, Math.abs(damages.tempMax));`
```

module/documents/_types.mjs
```diff
@@ -51,6 +51,7 @@
  * Options for damage application.
  *
  * @typedef DamageApplicationOptions
+ * @property {"damage"|"healing"} [as]       Treat maximum hit points types as healing or damage.
```
Fyorl: Is this option used anywhere?

module/documents/actor/actor.mjs
```diff
@@ -829,8 +831,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     if ( Hooks.call("dnd5e.preCalculateDamage", this, damages, options) === false ) return false;
 
     const multiplier = options.multiplier ?? 1;
+    const treatAs = options.origin?.flags?.dnd5e?.roll?.type
```
Fyorl: Do we need this? Can we determine if it's temp max damage or healing based on the sign?

module/applications/components/damage-application.mjs
```diff
@@ -275,15 +278,19 @@ export default class DamageApplicationElement extends TargetedApplicationMixin(C
...
    */
   refreshListEntry(token, entry, options) {
-    const { active, temp, total } = this.calculateDamage(token, options);
+    const { active, temp, tempMax, total } = this.calculateDamage(token, options);
     const calculatedDamage = entry.querySelector(".calculated.damage");
     calculatedDamage.innerText = formatNumber(-total, { signDisplay: "exceptZero" });
     calculatedDamage.classList.toggle("healing", total < 0);
     calculatedDamage.dataset.tooltip = `DND5E.${total < 0 ? "Healing" : "Damage"}`;
-    calculatedDamage.hidden = !total && !!temp;
+    calculatedDamage.hidden = !total && !!temp && !!tempMax;
```
Fyorl:
```suggestion
    calculatedDamage.hidden = !total && (!!temp || !!tempMax);
```

module/documents/actor/actor.mjs
arbron: This is actually handling increases to the temporary max, so adding +5 to your temporary max hit points will also increase your current hit points by the same amount (healing is a negative number the same way it is with the amount, since this is "applying damage").

Currently decreases to your temporary max don't decrease your current hit points because we are just relying on clamping in data preparation to handle that.

module/documents/actor/actor.mjs
arbron: None of the damage amounts are ever negative until they go through this method. For healing, we specifically check for `type === "healing"` and then multiply that by `-1` to invert it, but for maximum we need a way to distinguish because the type isn't sufficient.

I was also considering adding two distinct types, `maximum-damage` and `maximum-healing`, but that would move the complexity elsewhere.

module/applications/components/damage-application.mjs
```diff
@@ -72,7 +72,7 @@ export default class DamageApplicationElement extends TargetedApplicationMixin(C
    * @returns {DamageApplicationOptions}
    */
   getTargetOptions(uuid) {
-    if ( !this.#targetOptions.has(uuid) ) this.#targetOptions.set(uuid, { multiplier: 1 });
+    if ( !this.#targetOptions.has(uuid) ) this.#targetOptions.set(uuid, { multiplier: 1, origin: this.chatMessage });
```
Fyorl: I merged #6588 by the way, so be sure to align these changes with that, or vice versa.

module/documents/actor/actor.mjs
```diff
@@ -885,9 +891,11 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       d.value = d.value * damageMultiplier;
       d.active.multiplier = (d.active.multiplier ?? 1) * damageMultiplier;
       if ( d.type === "temphp" ) damages.temp += d.value;
+      else if ( d.type === "maximum" ) damages.tempMax += d.value * (treatAs === "healing" ? -1 : 1);
```
Fyorl: Does this interact with the `invertHealing` option at all?

module/documents/actor/actor.mjs
Fyorl: Yeah I considered having two distinct types as well when reviewing this, but I agree that it wouldn't necessarily be an improvement.

PR #6663, opened 2026-01-27
[#6660] Add transform mode for changing forms on single actor

### LINKED ISSUE #1783: Add `form` field for shapechanging creatures
It would be nice to view and select the form of creatures. See the image below for an example.

Suggestions :
- Ability to change statistics/data based on form
  - Statistics, e.g movement for the Wereraven
  - Token art and image
  - Features and spells
- Creatures like `Faerie Dragons` could also benefit from this, e.g. the form could be a color which grants additionnal spells
  - Or one could create Faerie Dragons for each color. The Monster Manual seperates them as `Young` and `Older`...

image

### LINKED ISSUE #6660: Add "Change Form" mode to transform activity
New mode for the transform activity that behaves differently than the current methods. Rather than using stats from an external creature to transform the actor, this instead swaps between several active effects to configure the actor. This would allow an NPC to have several forms that it transforms between (e.g. Humanoid form, Hybrid form, and Wolf form for a Werewolf). When the activity is used it will apply the active effect for the selected form and remove the active effect for other forms. With the token-changing effects coming in V14 this would make it easy to change the token with the transformation.

In order to properly implement forms, this would also need the ability to selectively show/hide certain items. Most creatures that transform in this manner also have items only available in certain form (e.g. "Bite (Hybrid or Wolf only)"). This item hiding could be handled in one of two ways: A) The transform profile itself defines what items are hidden in that form, or B) Active effects gain the ability to hide items (maybe `items.hidden | ADD | <item id or identifier>`).

The activity would have a setting for whether the creature supports a "form-less" mode. Most NPCs wouldn't support this because every mode needs to have at least one item hidden, but creatures that just change their appearance or players using a spell like Alter or Disguise Self this would make sense.

This method would probably only make sense when applied to the actor using the activity (at least the item hiding/revealing part).

arbron:
Adds a new mode to the Transform activity that allows for changing between different forms defined using active effects on a single actor. When the form is changed, the effect for the selected form is enabled and transfered to the actor while the effects for all other forms are diabled and no longer transfered.

This mode also adds a "Formless" option, which allows for disabling all forms. Using the formless option is good for spells like Disguise Self where the actor should go back to normal once the spell ends while creatures like Werewolves should not use this option, since they are always in one of three distinct forms.

Closes #6660

### DIFF
--- a/module/applications/activity/transform-sheet.mjs
+++ b/module/applications/activity/transform-sheet.mjs
@@ -75,7 +75,8 @@ export default class TransformSheet extends ActivitySheet {
 
     context.profileModes = [
       { value: "", label: _loc("DND5E.TRANSFORM.FIELDS.transform.mode.Direct") },
-      { value: "cr", label: _loc("DND5E.TRANSFORM.FIELDS.transform.mode.CR") }
+      { value: "cr", label: _loc("DND5E.TRANSFORM.FIELDS.transform.mode.CR"), rule: true },
+      { value: "form", label: _loc("DND5E.TRANSFORM.FIELDS.transform.mode.Form") }
     ];
     context.profiles = context.source.profiles.map((data, index) => ({
       data, index,
@@ -85,6 +86,24 @@ export default class TransformSheet extends ActivitySheet {
       prefix: `profiles.${index}.`,
       source: context.source.profiles[index] ?? data
     })).sort((lhs, rhs) => (lhs.name || "").localeCompare(rhs.name || "", game.i18n.lang));
+
+    context.effects = {
+      labels: {
+        legend: "DND5E.TRANSFORM.FIELDS.profiles.label"
+      },
+      fields: [
+        {
+          field: context.fields.transform.fields.mode,
+          options: context.profileModes,
+          value: context.source.transform.mode
+        },
+        {
+          field: context.fields.transform.fields.formless,
+          input: dnd5e.applications.fields.createCheckboxInput,
+          value: context.source.transform.formless
+        }
+      ]
+    };
 
     return context;
   }
@@ -116,7 +135,7 @@ export default class TransformSheet extends ActivitySheet {
   /** @inheritDoc */
   async _onRender(context, options) {
     await super._onRender(context, options);
-    this.element.querySelector(".activity-profiles").addEventListener("drop", this.#onDrop.bind(this));
+    this.element.querySelector(".activity-profiles")?.addEventListener("drop", this.#onDrop.bind(this));
   }
 
   /* -------------------------------------------- */
--- a/module/data/activity/transform-data.mjs
+++ b/module/data/activity/transform-data.mjs
@@ -37,6 +37,7 @@ export default class BaseTransformActivityData extends BaseActivityData {
       settings: new EmbeddedDataField(TransformationSetting, { nullable: true, initial: null }),
       transform: new SchemaField({
         customize: new BooleanField(),
+        formless: new BooleanField(),
         mode: new StringField({ initial: "cr" }),
         preset: new StringField()
       })
@@ -47,20 +48,25 @@ export default class BaseTransformActivityData extends BaseActivityData {
   /*  Properties                                  */
   /* -------------------------------------------- */
 
-  /** @override */
-  get applicableEffects() {
-    return null;
+  /**
+   * Transform profiles that can be performed based on spell/character/class level.
+   * @type {EffectApplicationData[]|TransformProfile[]}
+   */
+  get availableProfiles() {
+    if ( this.transform.mode === "form" ) return this.applicableEffects;
+    const level = this.relevantLevel;
+    return this.profiles.filter(e => ((e.level.min ?? -Infinity) <= level) && (level <= (e.level.max ?? Infinity)));
   }
 
   /* -------------------------------------------- */
 
   /**
-   * Transform profiles that can be performed based on spell/character/class level.
-   * @type {TransformProfile[]}
+   * When using the form mode, what is the ID of the currently active form if any.
+   * @type {string|null}
    */
-  get availableProfiles() {
-    const level = this.relevantLevel;
-    return this.profiles.filter(e => ((e.level.min ?? -Infinity) <= level) && (level <= (e.level.max ?? Infinity)));
+  get currentProfile() {
+    if ( this.transform.mode !== "form" ) return null;
+    return this.effects.find(e => !e.effect?.disabled && e.effect?.transfer)?._id ?? "";
   }
 
   /* -------------------------------------------- */
--- a/module/data/activity/transform-data.mjs
+++ b/module/data/activity/transform-data.mjs
@@ -66,7 +66,11 @@ export default class BaseTransformActivityData extends BaseActivityData {
    */
   get currentProfile() {
     if ( this.transform.mode !== "form" ) return null;
-    return this.effects.find(e => !e.effect?.disabled && e.effect?.transfer)?._id ?? "";
+    const active = this.effects.find(e => {
+      const effect = e.getEffect();
+      return !effect?.disabled && effect?.transfer;
+    })?._id;
+    return active ?? (this.transform.formless ? "" : null);
   }
 
   /* -------------------------------------------- */
--- a/module/documents/activity/transform.mjs
+++ b/module/documents/activity/transform.mjs
@@ -34,6 +34,7 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
         actions: {
           transformActor: TransformActivity.#transformActor
         },
+        applyEffectsInChat: false,
         dialog: TransformUsageDialog
       }
     }, { inplace: false })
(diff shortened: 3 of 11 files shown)


Zanderaf:
Guessing this also closes #1783 ?

module/data/activity/transform-data.mjs
```diff
@@ -47,22 +48,27 @@ export default class BaseTransformActivityData extends BaseActivityData {
   /*  Properties                                  */
   /* -------------------------------------------- */
 
-  /** @override */
-  get applicableEffects() {
```
Fyorl, line 51: If I am following the code correctly, we are deleting this so that we can use the `super.applicableEffects` behaviour to level-gate profiles in `availableProfiles`. This has the side-effect of making these effects visible in the effects tray. So to work-around that, we are deleting `messageConfig.data.flags.dnd5e.use.effects` (which is currently broken due to us migrating that flag). 

Rather than trying to repurpose `applicableEffects` for this, it might be preferable to instead use bare `this.effects` in `availableProfiles` and just duplicate the level-gate logic from `applicableEffects`. That way we don't have to mess with anything downstream. If we want to share code, it might be better to factor out the internals of `applicableEffects` instead.

module/applications/activity/transform-sheet.mjs
```diff
@@ -86,6 +87,22 @@ export default class TransformSheet extends ActivitySheet {
       source: context.source.profiles[index] ?? data
     })).sort((lhs, rhs) => (lhs.name || "").localeCompare(rhs.name || "", game.i18n.lang));
 
+    context.effects = {
+      title: _loc("DND5E.TRANSFORM.FIELDS.profiles.label"),
```
Fyorl: Already localised in the template
```suggestion
      title: "DND5E.TRANSFORM.FIELDS.Profiles.label",
```

module/documents/activity/transform.mjs
```diff
@@ -173,4 +195,27 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Handle transforming the actor to a specific form.
+   * @param {string} [profileId]  ID of the profile into which to transform.
+   */
+  async #transformToForm(profileId) {
+    if ( this.transform.mode !== "form" ) return;
+    const profile = this.applicableEffects.find(e => e._id === profileId);
```
Fyorl, line 207: Not used. Maybe we intend to bail if we can't find the profile?

module/documents/activity/transform.mjs
```diff
@@ -173,4 +195,27 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
...
+    // TODO: Not sure this is the best actor selection logic
+    const targets = getSceneTargets(this.actor).map(t => t.actor);
+    if ( !targets.length ) targets.push(this.actor);
+    for ( const target of targets ) {
+      const item = target.items.get(this.item.id);
+      if ( !item ) continue;
+      await item.updateEmbeddedDocuments("ActiveEffect", this.effects
```
Fyorl: Not necessary but might be nice to unroll this into a for loop so that we can detect if any change would be made and provide some feedback otherwise, like an `info`/`warn` for "Already in this form" or something.

module/documents/activity/transform.mjs
```diff
@@ -139,6 +146,16 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
...
   }
 
+  /* -------------------------------------------- */
+
+  /** @override */
+  async _triggerSubsequentActions(config, results) {
+    if ( this.transform.mode !== "form" ) return;
+    const profile = results.message?.flags?.dnd5e?.transform?.profile
+      ?? results.message?.data?.flags?.dnd5e?.transform?.profile
+    this.#transformToForm(profile);
```
Fyorl, line 156:
```suggestion
    return this.#transformToForm(profile);
```

module/documents/activity/transform.mjs
```diff
@@ -151,6 +168,11 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
    * @param {ChatMessage5e} message  Message associated with the activation.
    */
   static async #transformActor(event, target, message) {
+    if ( this.transform.mode === "form" ) {
+      this.#transformToForm(message.getFlag("dnd5e", "transform.profile"));
```
Fyorl:
```suggestion
      await this.#transformToForm(message.getFlag("dnd5e", "transform.profile"));
```

module/documents/activity/transform.mjs
```diff
@@ -139,6 +146,16 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
     return CompendiumBrowser.selectOne({ filters: { locked }});
   }
 
+  /* -------------------------------------------- */
+
+  /** @override */
+  async _triggerSubsequentActions(config, results) {
+    if ( this.transform.mode !== "form" ) return;
+    const profile = results.message?.flags?.dnd5e?.transform?.profile
+      ?? results.message?.data?.flags?.dnd5e?.transform?.profile
```
Fyorl:
```suggestion
      ?? results.message?.data?.flags?.dnd5e?.transform?.profile;
```

module/data/activity/transform-data.mjs
```diff
@@ -47,22 +48,27 @@ export default class BaseTransformActivityData extends BaseActivityData {
...
 
+  /* -------------------------------------------- */
+
+  /**
+   * When using the form mode, what is the ID of the currently active form if any.
+   * @type {string|null}
+   */
+  get currentProfile() {
+    if ( this.transform.mode !== "form" ) return null;
+    return this.effects.find(e => !e.effect?.disabled && e.effect?.transfer)?._id ?? "";
```
Fyorl: So `_prepareUsageConfig` can correctly fall back to the first profile if the transform is not formless.
```suggestion
    const active = this.effects.find(e => !e.effect?.disabled && e.effect?.transfer)?._id;
    if ( active ) return active;
    return this.transform.formless ? "" : null;
```

module/utils.mjs
```diff
@@ -675,7 +675,8 @@ export function getTargetDescriptors() {
  * @returns {Token5e[]}
  */
 export function getSceneTargets(actor) {
-  let targets = canvas.tokens?.controlled.filter(t => t.actor && (!actor || t.actor === actor)) ?? [];
+  let targets = canvas.tokens?.controlled
+    .filter(t => t.actor && (!actor || (t.actor === actor) || (t.document.baseActor === actor))) ?? [];
```
Fyorl: Not sure about this. Is it expected that you would be running an unlinked NPC via their sidebar actor? That would mean none of the usage tracking would be accurate for example. Unless I'm misunderstanding this change.

Fyorl (changes requested):

module/documents/activity/transform.mjs
arbron: `_triggerSubsequentActions` is intentionally not awaited and discards any return value.

module/utils.mjs
arbron: The logic was to allow for transforming multiple unlinked actors at once (say you have 3 werewolves in the scene, you could select them and use the Change Shape on the sidebar actor to transform all three at once). Only really useful when transforming multiple unlinked actors when the transformation doesn't have associated consumption, so I can pull it out to simplify things a bit.

Fyorl: Ah ok, that makes sense. I think the transformation controls where it calls `getSceneTargets` so maybe we add a parameter for this `baseActor` check. I'm a bit wary of changing this `getSceneTargets` behaviour globally.

module/data/activity/transform-data.mjs
arbron: This is the same method we are using in Enchant and Summon activities, but perhaps we can just add a metadata option that indicates whether the applicable effects are passed to the chat message or not.

module/documents/activity/transform.mjs
```diff
@@ -173,4 +195,37 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
...
+  /* -------------------------------------------- */
+
+  /**
+   * Handle transforming the actor to a specific form.
+   * @param {string} [profileId]  ID of the profile into which to transform.
+   */
+  async #transformToForm(profileId) {
+    if ( this.transform.mode !== "form" ) return;
+    const profile = this.applicableEffects.find(e => e._id === profileId);
+    if ( !profile && !this.transform.forrmless ) return;
```
Fyorl:
```suggestion
    if ( !profile && !this.transform.formless ) return;
```

module/utils.mjs
```diff
@@ -726,11 +726,15 @@ export function getTargetDescriptors() {
 
 /**
  * Get currently selected tokens in the scene or user's character's tokens.
- * @param {Actor5e} [actor]  Only allow tokens associated with this specific actor.
+ * @param {Actor5e} [actor]                   Only allow tokens associated with this specific actor.
+ * @param {object} [options={}]
+ * @param {boolean} [options=checkBaseActor]  Also include tokens whose base actor matches the provided actor.
```
Fyorl:
```suggestion
 * @param {boolean} [options.checkBaseActor]  Also include tokens whose base actor matches the provided actor.
```

Fyorl (approved):

PR #6666, opened 2026-01-27
[#6665] Add ability to hide items using active effect

### LINKED ISSUE #6665: Add active effect key for hiding items on sheet
To implement the "Change Form" transform mode (see https://github.com/foundryvtt/dnd5e/issues/6660) there needs to be a way to conditionally show and hide items on the sheet. Rather than baking that in to the transform activity itself, it would be useful if that could be done using an active effect such as `items.hidden | ADD | <item id or identifier>`, allowing this functionality to be used outside of just that context.

arbron:
Adds `items.hidden` as an active effect key that can be used to hide items on the sheet. This value accepts either a specific item ID or an identifier (with support for type prefixes).

When an item is hidden in this way it is not displayed on the sheet, any of its effects are suppressed, and it is hidden in the character's favorites.

Closes #6665

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -196,13 +196,32 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
   /** @inheritDoc */
   static applyChange(model, change, options={}) {
     change = change.effect._applyChangeShim(change);
+
+    // Handle special actor flags
     if ( change.key.startsWith("flags.dnd5e.") ) change = change.effect._prepareFlagChange(model, change);
+
+    // Properly handle formulas that don't exist as part of the data model
     if ( ActiveEffect5e.FORMULA_FIELDS.has(change.key) ) {
       const field = new FormulaField({ deterministic: change.key !== "system.damage.bonus" });
       return { [change.key]: this.applyChangeField(model, change, { field }) };
     }
+
+    // Handle activity-targeted changes
     if ( (change.key.startsWith("activities[") || change.key.startsWith("system.activities."))
       && (model instanceof Item) ) return change.effect.applyActivity(model, change);
+
+    // Handle hiding items
+    if ( change.key === "items.hidden" && (doc instanceof Actor) ) {
+      if ( change.type === "add" ) {
+        if ( doc.items.has(change.value) ) doc.hiddenItems.add(change.value);
+        else doc.identifiedItems.get(change.value)?.forEach(i => doc.hiddenItems.add(i.id));
+      } else if ( change.mode === "subtract" ) {
+        if ( doc.items.has(change.value) ) doc.hiddenItems.delete(change.value);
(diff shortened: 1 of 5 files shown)


module/documents/active-effect.mjs
```diff
@@ -196,13 +196,32 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
       const field = new FormulaField({ deterministic: change.key !== "system.damage.bonus" });
       return { [change.key]: this.applyChangeField(model, change, { field }) };
     }
+
+    // Handle activity-targeted changes
     if ( (change.key.startsWith("activities[") || change.key.startsWith("system.activities."))
       && (model instanceof Item) ) return change.effect.applyActivity(model, change);
+
+    // Handle hiding items
+    if ( change.key === "items.hidden" && (doc instanceof Actor) ) {
```
```suggestion
    if ( (change.key === "items.hidden") && (doc instanceof Actor) ) {
```

module/documents/active-effect.mjs
```diff
@@ -196,13 +196,32 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+      if ( change.type === "add" ) {
```
Fyorl, line 215:
```suggestion
      if ( change.mode === "add" ) {
```

module/documents/active-effect.mjs
```diff
@@ -196,13 +196,32 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+        if ( doc.items.has(change.value) ) doc.hiddenItems.add(change.value);
+        else doc.identifiedItems.get(change.value)?.forEach(i => doc.hiddenItems.add(i.id));
+      } else if ( change.mode === "subtract" ) {
+        if ( doc.items.has(change.value) ) doc.hiddenItems.delete(change.value);
+        else doc.identifiedItems.get(change.value)?.forEach(i => doc.hiddenItems.delete(i.id));
```
```suggestion
        if ( model.items.has(change.value) ) model.hiddenItems.add(change.value);
        else model.identifiedItems.get(change.value)?.forEach(i => model.hiddenItems.add(i.id));
      } else if ( change.mode === "subtract" ) {
        if ( model.items.has(change.value) ) model.hiddenItems.delete(change.value);
        else model.identifiedItems.get(change.value)?.forEach(i => model.hiddenItems.delete(i.id));
```

Fyorl (approved):
Being able to hide items using wildcard identifiers would be a pretty powerful feature, i.e. `leomunds-*`, but we would probably have to maintain a `WordTree` alongside the `IdentifiedItemsMap`. Not sure if it's worth the extra memory overhead for each Actor.

module/documents/active-effect.mjs
arbron: Think you got this one backward, needed to adjust the `"subtract"` line instead ;-)

PR #6667, opened 2026-01-27
Display limited uses for activities on item sheet

arbron:
Adds a "Charges" column to the activities list on item sheet to display the uses assigned to each activity. This column behaves the same as the similar column in the inventory.

Activity Uses

### DIFF
--- a/module/applications/components/activities.mjs
+++ b/module/applications/components/activities.mjs
@@ -119,7 +119,7 @@ export default class ActivitiesElement extends (foundry.applications.elements.Ad
       if ( activity && (input.dataset.name === "uses.value") ) {
         this.document.updateActivity(activityId, { "uses.spent": activity.uses.max - result });
       }
-      else this.document.update({ [input.dataset.name]: result });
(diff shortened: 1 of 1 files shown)


module/applications/components/activities.mjs
```diff
@@ -48,4 +37,103 @@ export default class ActivitiesElement extends (foundry.applications.elements.Ad
...
+    const activity = this.document.system.activities?.get(activityId);
+    if ( !activity ) return;
+    const result = parseInputDelta(input, activity);
+    if ( (result !== undefined) && input.dataset.name ) {
+      event.stopPropagation();
+      // Special case handling for uses.
+      if ( activity && (input.dataset.name === "uses.value") ) {
+        this.document.updateActivity(activityId, { "uses.spent": activity.uses.max - result });
+      }
+      else this.document.update({ [input.dataset.name]: result });
```
Fyorl: I don't think we update anything but uses at the moment which is probably why this got missed.


```suggestion
      else this.document.updateActivity(activityId, { [input.dataset.name]: result });
```

Fyorl (approved):

PR #6668, opened 2026-01-27
Add duration control to group rest dialog, "Until Sunrise" button

arbron:
Adds a control for setting the amount of time that will be advanced during a rest that appears when the calendar is enabled. This is only shown when resting from the group sheet and will cause the rest to advance time by the set amount.

Rest Duration

There is a helper button that will automatically set the rest duration to the following sunrise.

### DIFF
--- a/module/applications/actor/rest/base-rest-dialog.mjs
+++ b/module/applications/actor/rest/base-rest-dialog.mjs
@@ -1,7 +1,7 @@
-import { filteredKeys } from "../../../utils.mjs";
+import { convertTime, filteredKeys } from "../../../utils.mjs";
 import Dialog5e from "../../api/dialog.mjs";
 
-const { BooleanField } = foundry.data.fields;
+const { BooleanField, NumberField, StringField } = foundry.data.fields;
 
 /**
  * @import { RestConfiguration } from "../../../documents/_types.mjs";
@@ -22,6 +22,9 @@ export default class BaseRestDialog extends Dialog5e {
 
   /** @override */
   static DEFAULT_OPTIONS = {
+    actions: {
+      setDurationSunrise: BaseRestDialog.#onSetDurationSunrise
+    },
     classes: ["rest"],
     config: null,
     document: null,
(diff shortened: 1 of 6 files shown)


module/applications/actor/rest/base-rest-dialog.mjs
```diff
@@ -207,11 +243,30 @@ export default class BaseRestDialog extends Dialog5e {
       data.targets = filteredKeys(data.targets ?? {});
       this.actor.setFlag("dnd5e", "restSettings", data);
     }
+    if ( foundry.utils.getType(data.duration) === "Object" ) {
```
Fyorl:
```suggestion
    if ( foundry.utils.isPlainObject(data.duration) ) {
```

module/applications/actor/rest/base-rest-dialog.mjs
```diff
@@ -207,11 +243,30 @@ export default class BaseRestDialog extends Dialog5e {
...
+   * @this {BaseRestDialog}
+   * @param {Event} event         Triggering click event.
+   * @param {HTMLElement} target  Button that was clicked.
+   */
+  static #onSetDurationSunrise(event, target) {
+    const sunrise = Math.round(game.time.calendar.sunrise());
+    const now = game.time.components.hour;
+    const hoursUntilSunrise = sunrise - now + (now > sunrise ? game.time.calendar.days.hoursPerDay : 0);
+    this.element.querySelector('[name="duration.value"]').value = hoursUntilSunrise;
+    this.element.querySelector('[name="duration.unit"]').value = "hour";
```
Fyorl:
```suggestion
    this.form.elements["duration.value"].value = hoursUntilSunrise;
    this.form.elements["duration.unit"].value = "hour";
```

module/applications/actor/rest/base-rest-dialog.mjs
```diff
@@ -71,6 +75,17 @@ export default class BaseRestDialog extends Dialog5e {
 
   /* -------------------------------------------- */
 
+  /**
+   * Duration of the rest in minutes.
+   * @type {number}
+   */
+  get duration() {
+    return this.config.duration ?? CONFIG.DND5E.restTypes[this.config.type]
+      ?.duration?.[game.settings.get("dnd5e", "restVariant")] ?? 0;
```
Fyorl: Save a `Setting` creation every time we hit this getter.


```suggestion
      ?.duration?.[dnd5e.settings.restVariant] ?? 0;
```

Fyorl (approved):

PR #6673, opened 2026-01-29
Swap roll table SVGs with `dnd5e-icon`, make editable

arbron:
Automatically swaps all SVG icons on the roll table sheet with `<dnd5e-icon>` elements to allow for coloring them to match. Swaps any instances of the default icon with a version that the system provides that supports variable color. Also applies the `gold-icon` class to any non-SVG icons.

Roll Table Sheet Icons

In order to support editing these icons directly on the sheet, this adds a hidden `<input>` element inside `<dnd5e-icon>` when the `data-edit` attribute is set and keeps its value synchronized with the icon src. This also replaces core's `onEditImage` method with one that doesn't have a restriction to only work with `` elements.

### DIFF
--- a/module/applications/components/icon.mjs
+++ b/module/applications/components/icon.mjs
@@ -151,8 +151,7 @@ export default class IconElement extends AdoptedStyleSheetMixin(MaybeAdoptable) 
     if ( this.dataset.edit ) {
       if ( !input ) {
         input = document.createElement("input");
-        input.name = this.dataset.edit;
-        input.type = "hidden";
+        Object.assign(input, { name: this.dataset.edit, type: "hidden" });
         this.append(input);
       }
       input.value = src;
--- a/module/applications/roll-table-sheet.mjs
+++ b/module/applications/roll-table-sheet.mjs
@@ -28,22 +28,23 @@ export default class RollTableSheet5e extends ApplicationV2Mixin(RollTableSheet,
     this.element.querySelector(".sheet-header [data-action=changeMode]")?.remove();
     this.element.querySelectorAll("tbody .inline-control").forEach(c => c.classList.add("unbutton", "control-button"));
     this._replaceElements("input[type=checkbox]", "dnd5e-checkbox");
-    this._replaceElements('table img[src$=".svg"]', "dnd5e-icon", icon => {
+    this._replaceElements('table td.image img[src$=".svg"]', "dnd5e-icon", icon => {
       if ( icon.src === "icons/svg/d20-black.svg" ) icon.src = "systems/dnd5e/icons/svg/dice/d20.svg";
     });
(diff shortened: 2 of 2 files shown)


module/applications/components/icon.mjs
```diff
@@ -131,4 +127,44 @@ export default class IconElement extends AdoptedStyleSheetMixin(MaybeAdoptable)
...
+    if ( element instanceof Promise ) element.then(insertElement);
+    else insertElement(element);
+
+    // Add input if `data-edit` is set
+    let input = this.querySelector("input");
+    if ( this.dataset.edit ) {
+      if ( !input ) {
+        input = document.createElement("input");
+        input.name = this.dataset.edit;
+        input.type = "hidden";
```
Fyorl:
```suggestion
        Object.assign(input, { name: this.dataset.edit, type: "hidden" });
```

module/applications/roll-table-sheet.mjs
```diff
@@ -25,23 +28,29 @@ export default class RollTableSheet5e extends ApplicationV2Mixin(RollTableSheet,
...
    * Replace all matching elements with a new tag, keeping all existing attributes.
-   * @param {string} selector  CSS selector to find elements to replace.
-   * @param {string} tagName   Tag name for the new element to use.
+   * @param {string} selector      CSS selector to find elements to replace.
+   * @param {string} tagName       Tag name for the new element to use.
+   * @param {Function} [callback]  Method called for each new element before it replaces the old one.
    * @protected
    */
-  _replaceElements(selector, tagName) {
+  _replaceElements(selector, tagName, callback) {
```
Fyorl: Maybe wrap in `options` to allow extending later since this is extensible API.

module/applications/roll-table-sheet.mjs
```diff
@@ -25,23 +28,29 @@ export default class RollTableSheet5e extends ApplicationV2Mixin(RollTableSheet,
     this.element.querySelector(".sheet-header [data-action=changeMode]")?.remove();
     this.element.querySelectorAll("tbody .inline-control").forEach(c => c.classList.add("unbutton", "control-button"));
     this._replaceElements("input[type=checkbox]", "dnd5e-checkbox");
+    this._replaceElements('table img[src$=".svg"]', "dnd5e-icon", icon => {
+      if ( icon.src === "icons/svg/d20-black.svg" ) icon.src = "systems/dnd5e/icons/svg/dice/d20.svg";
+    });
+    this.element.querySelectorAll("table img").forEach(icon => icon.classList.add("gold-icon"));
```
Fyorl: This selector is too broad. It will match images inside entries.

```suggestion
    this.element.querySelectorAll("table td.image img").forEach(icon => icon.classList.add("gold-icon"));
```

Fyorl (approved):

Issue #6641, opened 2026-01-19
Allow standalone Active Effect documents to be linked to an activity


arbron:
Currently activities can only include Active Effects contained on the same Item in this applied effects list. With the addition of standalone Active Effect documents, it would be great to be able to link to an AE in its own compendium. This would allow for common effects to be re-used between different features.

It would also be nice to add a active effects compendium to the system that contains a number of useful effects:
- Ability check/save advantage (one per ability)
- Damage Resistance (one per damage type)
- Status effects (one for each status effect that may be applied)

These could all be pre-configured with SVG icons set and the changes prepared to make this easier for content creators.

PR #6679, opened 2026-01-30
[#6641] Add active effects compendium with stock effects

arbron:
Adds a new "Active Effects" compendium that contains stock effects for common use cases. These effects give advantage/disadvantage on ability checks, ability saves, ability checks & saves, and skill checks, resistance, immunity, and vulnerability to damage types, immunity to conditions, applying individual conditions, and giving special speeds equal to the actor's base speed.


Fyorl (approved):
Initiative advantage/disadvantage would be nice in the ability check section. Didn't notice any issues except some harmless ones: Some `flags.dnd5e.riders.statuses: []` in a couple of places, and some changes have `priority: null` while others just omit the key entirely.

Issue #6408, opened 2025-11-15
Generalize the handling for Exhaustion to general conditions with "levels"


krbz999:
- Store the 'level' of the status on the ActiveEffect document.
- Make exhaustion levels on the actor derived rather than storing it on the actor data models.
- Generalize the handling for exhaustion to easily allow for other statuses with levels (such as 'Contamination' from the Drakkenheim setting).
- Set the level max (i.e. 6 for exhaustion) in `DND5E.statusEffects` as an optional property.
- For compatibility, redirect updates targeting `attributes.exhaustion` to create/update/delete an exhaustion effect.
- Reuse the method that retrieves the exhaustion icon; it's a likely assumption that a module-added condition with levels stores all its icons in the same directory, and it's reasonable to require that the files are named identically (with only a numeric value in the name being different). This should then be used for the token HUD in a more general manner.

Stretch goal:
- Implement new status/condition effect subtype? The scrolling text on the canvas could then be moved into a function on this subtype.
- Store the additional statuses of "legacy" exhaustion in `DND5E.statusEffects` rather than hardcoding during data prep? Generalize this concept as well such that even Drakkenheim contamination can apply additional statuses easily*?
- Add property for leveled conditions, even module-added conditions, in actors' `attributes` during data preparation?
- Adjust `Actor#toggleStatusEffect` to allow passing a level increase/delta?

\*_(current workaround would be to hook onto a spell slot preparation method, or monkeypatch `prepareDerivedData`)_

Zanderaf:
would like this as well, trying to implement some sort of tracking to Delayed Blast Fireball leads to some ugly active effect tab viewing when applying the same effect more than once. AEs having "Levels" would be a nice way to handle cumulative/stacking effects

<img width="777" height="367" alt="Image" src="https://github.com/user-attachments/assets/857a28d1-9b10-4e26-afe3-5b5ef49e6bb3" />

Dewi-Jones:
Active effects are not conditions. Stackable AE's would surely be their own feature request.

PR #6699, opened 2026-02-06
[#6408] Generalize leveled conditions

krbz999:
Adds a "Condition" ActiveEffect subtype which is responsible for storing levels and other data related to that on the actor.

Generalizes most of the handling surrounding exhaustion. The `attributes.exhaustion` property remains for backwards compatibility and for the ability to update the Exhaustion condition via an actor update. For other conditions with levels, one can call `actor.toggleStatusEffects(statusId, { levels: 1 })`.

For example, adding a new condition,
```js
DND5E.conditionTypes.contamination = {
  name: "Contamination",
  img: "modules/drakkenheim-scgd/assets/layout/meteor.svg",
  levels: 6,
  conditions: {
    4: ["weakened"],
    5: ["dazed", "incapacitated"],
  },
};
```
will add a new status with 6 levels, where "Weakened" (an unrelated example condition) is applied at 4 leves of Contamination or higher, and Dazed and Incapaciated are applied at 5 levels or higher.

The `reduction` property is also generalized and can be used by any condition, not just leveled ones.

### DIFF
--- a/module/applications/hud/token-hud.mjs
+++ b/module/applications/hud/token-hud.mjs
@@ -1,0 +1,69 @@
+import ActiveEffect5e from "../../documents/active-effect.mjs";
+import ConditionData from "../../data/active-effect/condition.mjs";
+
+export default class TokenHUD5e extends foundry.applications.hud.TokenHUD {
+  /** @inheritdoc */
+  static DEFAULT_OPTIONS = {
+    actions: {
+      effect: {
+        handler: TokenHUD5e.#onToggleEffect,
+        buttons: [0, 2]
+      }
+    }
+  };
+
+  /* -------------------------------------------------- */
+
+  /** @inheritDoc */
+  async _onRender(context, options) {
+    await super._onRender(context, options);
+    this.#adjustConditionIcons();
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Replace condition icons with leveled icons.
+   */
+  #adjustConditionIcons() {
+    const actor = this.object.actor;
+    if ( !actor ) return;
+
+    const conditions = actor.effects.documentsByType.condition;
+    for ( const c of conditions ) {
+      if ( !c.system.hasLevels ) continue;
+      const { level, type } = c.system;
+      const img = ConditionData.getIconByLevel(type, level);
+      const icon = this.element.querySelector(`[data-status-id="${type}"]`);
+      if ( !icon ) continue;
+      icon.style.objectPosition = "-100px";
+      icon.style.background = `url("${img}") no-repeat center / contain`;
+    }
+  }
+
+  /* -------------------------------------------------- */
+  /*   Event handlers                                   */
+  /* -------------------------------------------------- */
+
+  /**
+   * Toggle a condition via the HUD.
+   * @param {PointerEvent} event    The triggering event.
+   * @param {HTMLElement} target    The element that defined the [data-action].
+   */
+  static async #onToggleEffect(event, target) {
+    if ( !this.actor ) {
+      ui.notifications.warn("HUD.WarningEffectNoActor", { localize: true });
+      return;
+    }
+
+    const id = target.dataset.statusId;
+
+    let resolved = false;
+    if ( id === "concentrating" ) resolved = ActiveEffect5e._manageConcentration(event, this.actor);
+
+    if ( !resolved ) this.actor.toggleStatusEffect(id, {
+      overlay: event.button === 2,
+      levels: event.button === 0 ? 1 : -1
+    });
+  }
+}
--- a/module/applications/hud/token-hud.mjs
+++ b/module/applications/hud/token-hud.mjs
@@ -47,6 +47,7 @@ export default class TokenHUD5e extends foundry.applications.hud.TokenHUD {
 
   /**
    * Toggle a condition via the HUD.
+   * @this TokenHUD5e
    * @param {PointerEvent} event    The triggering event.
    * @param {HTMLElement} target    The element that defined the [data-action].
    */
--- a/module/data/active-effect/condition.mjs
+++ b/module/data/active-effect/condition.mjs
@@ -1,0 +1,169 @@
+const { ActiveEffectTypeDataModel } = foundry.data;
+const { TypeDataModel } = foundry.abstract;
+
+const { NumberField, StringField } = foundry.data.fields;
+
+/**
+ * System data model for condition active effects.
+ */
+export default class ConditionData extends (ActiveEffectTypeDataModel ?? TypeDataModel) {
+  /** @override */
+  static defineSchema() {
+    const schema = ActiveEffectTypeDataModel ? super.defineSchema() : {};
+    return Object.assign(schema, {
+      level: new NumberField({ nullable: true, integer: true, initial: null, min: 1 }),
+      type: new StringField({ required: true, blank: false })
+    });
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Does the given status have levels?
+   * @param {string} type   The primary status id.
+   * @returns {boolean}
+   */
+  static hasLevels(type) {
+    const config = CONFIG.DND5E.conditionTypes[type];
+    return !!config && ("levels" in config);
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Does this condition have levels?
+   * @type {boolean}
+   */
+  get hasLevels() {
+    return ConditionData.hasLevels(this.type);
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * The max level of this condition.
+   * @type {number|null}
+   */
+  get maxLevel() {
+    return this.hasLevels
+      ? CONFIG.DND5E.conditionTypes[this.type].levels
+      : null;
+  }
+
+  /* -------------------------------------------------- */
+
+  /** @inheritDoc */
+  prepareDerivedData() {
+    super.prepareDerivedData();
+    if ( !this.hasLevels ) {
+      this.level = null;
+      return;
+    }
+
+    this.level = Math.clamp(this.level, 1, this.maxLevel);
+    this.parent.name = `${CONFIG.DND5E.conditionTypes[this.type].name} (${this.level})`;
+    this.parent.img = this.constructor.getIconByLevel(this.type, this.level);
+
+    for (let i=1; i <= this.level; i++) {
+      const statuses = CONFIG.DND5E.conditionTypes[this.type].conditions?.[i] ?? [];
+      statuses.forEach(s => this.parent.statuses.add(s));
+    }
+
+    const actor = this.parent.parent;
+    if ( this.parent.active && (actor instanceof Actor) ) {
+      actor.system.conditions ??= {};
+      actor.system.conditions[this.type] = this.level;
+    }
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Get the image used to represent a status of a given level.
+   * @param {string} type     The condition type.
+   * @param {number} level    The level of the condition.
+   * @returns {string}
+   */
+  static getIconByLevel(type, level) {
+    const { img } = CONFIG.DND5E.conditionTypes[type];
+    const split = img.split(".");
+    const ext = split.pop();
+    const path = split.join(".");
+    return `${path}-${level}.${ext}`;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Decrease the level of this condition.
+   * @param {number} [levels=1]   The increase in levels.
+   * @returns {Promise<ActiveEffect5e>}   A promise that resolves to the updaeed effect.
+   */
+  async decrease(levels=1) {
+    if ( !this.hasLevels ) return this;
+    await this.parent.update({ "system.level": this.level - levels }, { dnd5e: { originalLevel: this.level }});
+    return this.parent;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Increase the level of this condition.
+   * @param {number} [levels=1]   The increase in levels.
+   * @returns {Promise<ActiveEffect5e>}   A promise that resolves to the updaeed effect.
+   */
+  async increase(levels=1) {
+    if ( !this.hasLevels ) return this;
+    await this.parent.update({ "system.level": this.level + levels }, { dnd5e: { originalLevel: this.level }});
+    return this.parent;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Manage a change to levels of an actor's conditions.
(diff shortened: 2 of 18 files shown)


module/applications/hud/token-hud.mjs
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  /* -------------------------------------------------- */
+  /*   Event handlers                                   */
+  /* -------------------------------------------------- */
+
+  /**
+   * Toggle a condition via the HUD.
+   * @param {PointerEvent} event    The triggering event.
```
Fyorl, line 51:
```suggestion
   * @this {TokenHUD5e}
   * @param {PointerEvent} event    The triggering event.
```

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
+const { ActiveEffectTypeDataModel } = foundry.data;
+const { TypeDataModel } = foundry.abstract;
+
+const { NumberField, StringField } = foundry.data.fields;
+
+/**
+ * System data model for condition active effects.
+ */
+export default class ConditionData extends (ActiveEffectTypeDataModel ?? TypeDataModel) {
```
Fyorl: We can drop this now in 6.0
```suggestion
export default class ConditionData extends ActiveEffectTypeDataModel {
```

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
+const { ActiveEffectTypeDataModel } = foundry.data;
+const { TypeDataModel } = foundry.abstract;
```
Fyorl:
```suggestion
```

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
...
+ * System data model for condition active effects.
+ */
+export default class ConditionData extends (ActiveEffectTypeDataModel ?? TypeDataModel) {
+  /** @override */
+  static defineSchema() {
+    const schema = ActiveEffectTypeDataModel ? super.defineSchema() : {};
+    return Object.assign(schema, {
+      level: new NumberField({ nullable: true, integer: true, initial: null, min: 1 }),
+      type: new StringField({ required: true, blank: false })
+    });
```
Fyorl:
```suggestion
    return {
      ...super.defineSchema(),
      level: new NumberField({ nullable: true, integer: true, initial: null, min: 1 }),
      type: new StringField({ required: true, blank: false })
    };
```

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
...
+    this.parent.img = this.constructor.getIconByLevel(this.type, this.level);
+
+    for (let i=1; i <= this.level; i++) {
+      const statuses = CONFIG.DND5E.conditionTypes[this.type].conditions?.[i] ?? [];
+      statuses.forEach(s => this.parent.statuses.add(s));
+    }
+
+    const actor = this.parent.parent;
+    if ( this.parent.active && (actor instanceof Actor) ) {
+      actor.system.conditions ??= {};
```
Fyorl: We can add this to `CommonTemplate` as `persisted: false` now.

```js
conditions: new MappingField(new NumberField({ integer: true, min: 1 }), { persisted: false })
```

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
...
+    const ext = split.pop();
+    const path = split.join(".");
+    return `${path}-${level}.${ext}`;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Decrease the level of this condition.
+   * @param {number} [levels=1]   The increase in levels.
```
Fyorl:
```suggestion
   * @param {number} [levels=1]           The decrease in levels.
```

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
...
+    if ( !this.hasLevels ) return this;
+    await this.parent.update({ "system.level": this.level - levels }, { dnd5e: { originalLevel: this.level }});
+    return this.parent;
+  }
+
+  /* -------------------------------------------------- */
+
+  /**
+   * Increase the level of this condition.
+   * @param {number} [levels=1]   The increase in levels.
```
Fyorl:
```suggestion
   * @param {number} [levels=1]           The increase in levels.
```

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
...
+
+  /* -------------------------------------------------- */
+
+  /** @inheritDoc */
+  _onUpdate(data, options, userId) {
+    super._onUpdate(data, options, userId);
+
+    if ( !this.hasLevels ) return;
+    const delta = this.level - options.dnd5e?.originalLevel;
+    if ( delta ) this.parent._displayScrollingStatus(delta);
```
Fyorl: `_displayScrollingStatus` accepts a boolean.
```suggestion
    this.parent._displayScrollingStatus(delta > 0);
```

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
...
+      });
+      return getDocumentClass("ActiveEffect").create(data, { parent: actor, keepId: true });
+    }
+
+    // Decrease level (possibly delete).
+    if ( effect && (levels < 0) ) {
+      const nLevel = effect.system.level + levels;
+      if ( nLevel <= 0 ) {
+        await effect.delete();
+        return false;
```
Fyorl: JSDoc says we return the deleted effect.


```suggestion
        return effect;
```

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
...
+    if ( !this.hasLevels ) {
+      this.level = null;
+      return;
+    }
+
+    this.level = Math.clamp(this.level, 1, this.maxLevel);
+    this.parent.name = `${CONFIG.DND5E.conditionTypes[this.type].name} (${this.level})`;
+    this.parent.img = this.constructor.getIconByLevel(this.type, this.level);
+
+    for (let i=1; i <= this.level; i++) {
```
Fyorl:
```suggestion
    for ( let i = 1; i <= this.level; i++ ) {
```

module/data/actor/templates/attributes.mjs
```diff
@@ -501,12 +500,19 @@ export default class AttributesFields {
     const heavilyEncumbered = statuses.has("heavilyEncumbered");
     const exceedingCarryingCapacity = statuses.has("exceedingCarryingCapacity");
     const units = this.attributes.movement.units ??= defaultUnits("length");
-    let reduction = dnd5e.settings.rulesVersion === "modern" && !this.traits?.ci?.value?.has("exhaustion")
-      ? (this.attributes.exhaustion ?? 0) * (CONFIG.DND5E.conditionTypes.exhaustion?.reduction?.speed ?? 0) : 0;
+
+    let reduction = statuses.reduce((acc, status) => {
```
Fyorl, line 544: Probably rebase artefact but we lost the check against `this.traits.ci.value`.

module/documents/actor/actor.mjs
```diff
@@ -527,37 +528,39 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    const conditions = this.system.conditions ?? {};
     const imms = this.system.traits?.ci?.value ?? new Set();
-    const applyExhaustion = (level !== null) && !imms.has("exhaustion")
-      && (dnd5e.settings.rulesVersion === "legacy");
-    const statuses = this.statuses;
-    const isActiveSource = k => {
-      const l = Number(k.split("-").pop());
-      return (statuses.has(k) && !imms.has(k)) || (applyExhaustion && Number.isInteger(l) && (level >= l));
+
+    const applyCondition = prop => {
```
Fyorl, line 550: Probably a rebase artefact, but we lost the special-casing around dodging.

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,169 @@
...
+      const effect = await getDocumentClass("ActiveEffect").fromStatusEffect(statusId);
+      const data = foundry.utils.mergeObject(effect.toObject(), {
+        _id: dnd5e.utils.staticID(`dnd5e${statusId}`),
+        "system.level": levels
+      });
+      return getDocumentClass("ActiveEffect").create(data, { parent: actor, keepId: true });
+    }
+
+    // Decrease level (possibly delete).
+    if ( effect && (levels < 0) ) {
```
Fyorl, line 144: What should happen if there is no effect and the user attempts to decrease? I think at the moment it falls through to default core behaviour, which would apply it as an overlay effect. Do we want to suppress that? Do we suppress it here or at the TokenHUD layer?

module/documents/actor/actor.mjs
```diff
@@ -3494,9 +3497,27 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-    const created = await super.toggleStatusEffect(statusId, options);
+  /**
+   * Toggle a configured status effect for the Actor.
+   * @override
+   * @param {string} statusId                 A status effect ID defined in CONFIG.statusEffects
+   * @param {object} [options={}]             Additional options which modify how the effect is created.
+   * @param {boolean} [options.active]        Force the effect to be active or inactive regardless of its current state.
+   * @param {boolean} [options.overlay=false] Display the toggled effect as an overlay.
+   * @param {number} [options.levels=1]       A potential level increase.
+   * @returns {Promise<ActiveEffect|boolean|undefined>}  A promise which resolves to one of the following values:
```
```suggestion
   * @returns {Promise<ActiveEffect|boolean|void>}  A promise which resolves to one of the following values:
```

module/documents/active-effect.mjs
```diff
@@ -720,31 +702,6 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
 
   /* -------------------------------------------- */
 
-  /**
-   * Determine whether this effect applies a status that should prompt concentration to end.
-   * @returns {boolean}
-   * @protected
-   */
-  _shouldPromptConcentrationEnd() {
```
Fyorl, line 728: The branch moved underneath you here and we have concentration handling in `_onCreateOperation` which still uses this method.

module/settings.mjs
```diff
@@ -744,6 +744,15 @@ export function applyLegacyRules() {
   DND5E.conditionEffects.initiativeDisadvantage.delete("incapacitated");
   DND5E.conditionEffects.initiativeDisadvantage.delete("surprised");
 
+  // Add exhaustion effects.
+  DND5E.conditionEffects.noMovement.add("exhaustion-5");
```
Fyorl, line 748: Good idea

module/documents/actor/actor.mjs
```diff
@@ -3518,18 +3539,10 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
    * @protected
    */
   async _onUpdateExhaustion(data, options) {
```
Fyorl, line 3575: I think at this point `system.conditions.exhaustion` should be equivalent to `originalExhaustion` since we haven't run `_applyDelta` yet. In which case we can use that to diff against, and then we can delete the `_preUpdate` override that records `originalExhaustion` entirely.

module/documents/actor/actor.mjs
Fyorl, line 1: We have generalised `reduction.speed`, which is good, but I think `reduction.roll` remains hard-coded. Needs something like this instead of `addRollExhaustion`:

```js
addConditionRollReduction(parts, data) {
  const imms = this.system.traits?.ci?.value ?? new Set();
  for ( const status of this.statuses ) {
    if ( imms.has(status) ) continue;
    const { rolls } = CONFIG.DND5E.conditionTypes[status]?.reduction ?? {};
    if ( !rolls ) continue;
    const levels = ConditionData.hasLevels(status) ? (this.system.conditions?.[status] ?? 0) : 1;
    const penalty = level * rolls;
    if ( penalty ) {
      parts.push(`@reductions.${status}`);
      data.reductions ??= {};
      data.reductions[status] = penalty;
    }
  }
}
```

Fyorl (changes requested):
Good consolidation

module/data/active-effect/condition.mjs
krbz999: I did run into this initially with exhaustion becoming an overlay and I decided I was fine with that.

module/documents/actor/actor.mjs
krbz999: I have implemented this and deprecated the exhaustion function.

module/documents/actor/actor.mjs
krbz999: Could do with a second pair of eyes on my logic here.

module/data/actor/templates/attributes.mjs
krbz999: It was indeed. Should be re-implemented now.

module/data/active-effect/condition.mjs
```diff
@@ -0,0 +1,164 @@
...
+  /* -------------------------------------------------- */
+
+  /**
+   * Does the given status have levels?
+   * @param {string} type   The primary status id.
+   * @returns {boolean}
+   */
+  static hasLevels(type) {
+    const config = CONFIG.DND5E.conditionTypes[type];
+    return !!config && ("levels" in config);
```
krbz999, line 24: Could be that `Number.isFinite(config?.levels) && config.levels > 0` is an improvement, but I'm not aware of it being meaningfully faster.

module/documents/actor/actor.mjs
Fyorl: LGTM. Might recurse infinitely if someone were to add `dodging` to `noMovement`. But that would be a strange thing to do.

module/documents/actor/actor.mjs
```diff
@@ -1185,18 +1194,40 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+  addConditionRollReduction(parts, data) {
+    const immunities = this.system.traits?.ci?.value ?? new Set();
+    this.statuses.difference(immunities).forEach(status => {
+      const reduction = CONFIG.DND5E.conditionTypes[status]?.reduction?.rolls ?? 0;
+      if ( !reduction ) return;
+
+      const levels = ConditionData.hasLevels(status) ? (this.system.conditions[status] ?? 0) : 1;
+      const penalty = levels * reduction;
+
+      if (penalty) {
```
```suggestion
      if ( penalty ) {
```

module/documents/actor/actor.mjs
Fyorl: It occurred to me that even without this PR we could have done the same thing by reading the level off the existing exhaustion AE. So what we have at the moment is probably cleaner, and I would retract this suggestion.

Fyorl (approved):

Issue #6357, opened 2025-10-30
Move enchantment riders from activity to effect itself


arbron:
Currently rider activities, effects, and items are defined on the Enchant activity for each rider profile. Moving these onto the effect itself would allow enchantments with riders to be created without the Enchant activity, and would move enchantment effects in line with what is required to implement https://github.com/foundryvtt/dnd5e/issues/4706 (since normal effects don't necessarily have an associated activity to store rider data).

Zanderaf:
Somewhat related, would this inevitably solve https://github.com/foundryvtt/dnd5e/issues/5799 ? The problem I was running into with it seems to sort of be solved with the actual riders being on the enchantment if I'm understanding this issue right.

PR #6705, opened 2026-02-09
[#6335] Add `ModifyItem` advancement for enchanting existing items

### LINKED ISSUE #6335: Advancement Type: Enchant/Modify Item
This advancement type allows the modification of an existing item by automatically adding an enchantment.

## Level Up Interface

The interface will display any matching items found on the sheet that will be modified. It may also have the ability to choose whether an item is modified or not.

## Configuration

The `changes` array contains one or more changes that will be applied.

Within it, the `enchantment` property contains an ID for a single enchantment that will be applied to any matching items.

The `identifiers` array contains typed identifiers used to find items that will be modified.

```json
{
  "changes": [{
    "enchantment": "vEKsOhMFEZeSb3dF",
    "identifiers": ["spell:mage-hand"]
  }]
}
```

## Value

The `modified` array contains data on all items that were modified by the advancement. It contains `item`, the ID of the modified item, and `effect`, the ID of the created enchantment effect.

```json
{
  "modified": [
    { "item": "0f8yGfhBaQazWKqR", "effect": "EqtYkfVJzdDI5Oz5" }
  ]
}
```

## Future Directions

In the future, when a more powerful tag-based filtering is provided by core (V14 perhaps), the simple `identifiers` list could be expanded to support more filtering options (perhaps it only modifies the "Mage Hand" spell if it is assigned to a specific class).

One other consideration is how to handle new items added after the advancement has been run. Should there be a mechanism to more easily apply the modification if "Mage Hand" is added later?

arbron:
Adds a new advancement type that allows for applying an enchantment to an existing item on an actor's sheet. This advancement contains a list of enchantments to apply, and a list of identifiers that are used to find the items to apply those enchantments to.

Modify Items Config (feature)

When configured directly on a class, species, or background, the configuration dialog doesn't include the selection dropdown nor the "Disassociate" button because there is not "Effects" tab visible on those item types.

Modify Items Config (class)

During the advancement process, each enchantment will be displayed with a list of matching items below it.

Modify Items Flow

### Future Improvement
- Allow user control over what items are actually enchanted
- Support rider activities & effects once #6357 is resolved

Closes #6335


pgervais70:
What if a matching item is added later?
You would need to reapply the advancement?

arbron:
> What if a matching item is added later? You would need to reapply the advancement?

Yep

Zanderaf:
Any plans for this to be retroactive in the future like mentioned in the original issue?

maragondi:
It would be great if an enchantment could automatically apply to the appropriate items when they appear on an actor, and automatically disappear if the item that provides the enchantment is removed.

This would make it possible to create a lot of cool things. For example, I could imagine a homebrew feature that improves a whole group of cold school abilities, or even a full talent system.

I realize this is probably a pretty niche use case, but automatic application would still be extremely convenient.

icons/LICENSE
```diff
@@ -46,6 +46,7 @@ The dnd5e system for Foundry Virtual Tabletop includes icon artwork licensed fro
 /svg/actors/group.svg - "Tattered banner" by Lorc under CC BY 3.0
 /svg/actors/npc.svg - "Spiked dragon head" by Delapouite under CC BY 3.0
 /svg/actors/vehicle.svg - "Ship's wheel" by Delapouite under CC BY 3.0
+/svg/advacement/modify-item.svg - "Anvil impact" by Lorc under CC BY 3.0
```
Fyorl:
```suggestion
/svg/advancement/modify-item.svg - "Anvil impact" by Lorc under CC BY 3.0
```

module/applications/advancement/modify-item-config.mjs
```diff
@@ -0,0 +1,132 @@
...
+   * Handle creating a new effect and associating with a change.
+   * @this {ModifyItemConfig}
+   * @param {Event} event         Triggering click event.
+   * @param {HTMLElement} target  Button that was clicked.
+   */
+  static async #addChange(event, target) {
+    const effectData = {
+      name: this.advancement.title,
+      img: this.advancement.icon,
+      origin: this.item.uuid,
```
Fyorl: We need to leave origin blank and only set it at the point of applying the enchantments to the modified items.

```js
/**
 * Is this effect an enchantment on an item that accepts enchantment?
 * @type {boolean}
 */
get isAppliedEnchantment() {
  return (this.type === "enchantment") && !!this.origin && (this.origin !== this.parent.uuid);
}
```

`this.origin !== this.parent.uuid` is `false` when the Item is at rest, but once it is placed on an Actor, it's given a new ID (most of the time), which breaks this check and results in the enchantment suddenly applying to its container Item. Leaving it blank means it will never be applied to the container Item.

module/data/advancement/modify-item-data.mjs
```diff
@@ -0,0 +1,53 @@
...
+  /** @override */
+  static LOCALIZATION_PREFIXES = ["DND5E.ADVANCEMENT.ModifyItem"];
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  static defineSchema() {
+    return {
+      changes: new ArrayField(new SchemaField({
+        _id: new DocumentIdField(), // TODO: Allow for ID or UUID in V14
```
Fyorl: Should we do this now?

module/data/advancement/modify-item-data.mjs
```diff
@@ -0,0 +1,53 @@
...
+ * Value data for the Modify Item advancement.
+ * @extends {foundry.abstract.DataModel<ModifyItemAdvancementValueData>}
+ * @mixes ModifyItemAdvancementValueData
+ */
+export class ModifyItemValueData extends foundry.abstract.DataModel {
+  /** @inheritDoc */
+  static defineSchema() {
+    return {
+      modified: new ArrayField(new SchemaField({
+        change: new DocumentIdField(), // TODO: Allow for ID or UUID in V14
```
Fyorl: Same as above

module/documents/advancement/modify-item.mjs
```diff
@@ -0,0 +1,91 @@
...
+      const effect = this.item.effects.get(change._id);
+      if ( !effect ) continue;
+      const itemsToChange = change.identifiers
+        .values()
+        .flatMap(i => this.actor.identifiedItems.get(i) ?? new Set())
+        .filter(i => !alreadyModified.has(`${change._id}.${i._id}`))
+        .toArray();
+      for ( const item of itemsToChange ) {
+        const clone = effect.clone({
+          _id: foundry.utils.randomID(),
```
Fyorl, line 49: We would set `origin` here.

module/documents/advancement/modify-item.mjs
```diff
@@ -0,0 +1,91 @@
...
+
+  /** @override */
+  restore(level, data, options={}) {
+    this.apply(level, data, options);
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  reverse(level, options={}) {
```
Fyorl: These methods are all `async` now.

Fyorl (changes requested):

module/applications/advancement/modify-item-config.mjs
```diff
@@ -0,0 +1,136 @@
...
+      : this.advancement.configuration.toObject().changes;
+    if ( foundry.utils.hasProperty(submitData, "selectedLocalEffects")
+      || foundry.utils.hasProperty(submitData, "selectedRemoteEffects") ) {
+      for ( const _id of submitData.selectedLocalEffects ?? [] ) {
+        if ( changes.find(e => e._id === _id) ) continue;
+        changes.push({ _id });
+      }
+      for ( const uuid of submitData.selectedRemoteEffects ?? [] ) {
+        const effect = fromUuidSync(uuid, { strict: false });
+        if ( changes.find(e => e.uuid === uuid) || (effect.type !== "enchantment") ) continue;
```
Fyorl:
```suggestion
        if ( changes.find(e => e.uuid === uuid) || (effect?.type !== "enchantment") ) continue;
```

module/documents/advancement/modify-item.mjs
```diff
@@ -0,0 +1,100 @@
...
+  /** @override */
+  automaticApplicationValue(level) {
+    return {};
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  async restore(level, data, options={}) {
+    this.apply(level, data, options);
```
Fyorl:
```suggestion
    return this.apply(level, data, options);
```

templates/advancement/modify-item-config-changes.hbs
```diff
@@ -0,0 +1,51 @@
...
+        {{/each}}
+    </ul>
+</fieldset>
+
+{{#*inline ".change"}}
+<div class="details flexrow">
+    {{ dnd5e-icon effect.img alt="" classes="gold-icon" }}
+    {{{ contentLink }}}
+    <div class="list-controls flexrow">
+        {{#if @root.hasEffectsTab}}
```
Fyorl: Otherwise we have no way to remove remote AEs.


```suggestion
        {{#if (or data.uuid @root.hasEffectsTab)}}
```

Fyorl (approved):

PR #6707, opened 2026-02-09
[#6627] Fix `getActorCurrencyUpdates` not working in some instances

### LINKED ISSUE #6627: deductActorCurrency doesn't look at higher currencies
deductActorCurrency fails to deduct currency when there are insufficient funds from lower denominations, even if higher denominations exist.

ie.
This will fail:
Player: 1pp 0gp 0sp 0cp
dnd5e.applications.CurrencyManager.deductActorCurrency(a, 1, "gp")

But cascading down works:
Player: 0pp 0gp 10sp 0cp
dnd5e.applications.CurrencyManager.deductActorCurrency(a, 1, "gp")

-- update --

It actually does seem to work with whole numbers, but doesn't work with remainders.

ie if I have 1pp I can deduct 10gp sucessfully, but not 1gp

arbron:
Modifies how `getActorCurrencyUpdates` works to now perform additional passes in an attempt to find the correct currency to deduce without resulting in a remainder. Previously, if you had a character with 100 gp, 5 sp, and 20 cp, and you asked to deduce 100 cp, it would return a remainder of 30 despite having enough currency to complete the request. This is because the loop would start at the lowest currency (100 - 20 = 80), then go to the next currency up (80 - (5 * 10) = 30), and then continue up through the denominations not finding any more that would be able to be deducted.

This change causes it to instead perform several loops over the list of currencies. If on the first pass ends with a remainder, then it will move the first entry in the currencies list to the end and re-perform the calculation. It will continue until it ends up with a remainder of `0` or has used each currency denomination as its first entry.

Closed #6627

### DIFF
--- a/module/applications/currency-manager.mjs
+++ b/module/applications/currency-manager.mjs
@@ -275,15 +275,16 @@ export default class CurrencyManager extends Application5e {
     const { currency } = actor.system;
     if ( amount <= 0 ) return { system: { currency: { ...currency } }, remainder: amount, item: [] };
 
-    const currencies = Object.entries(CONFIG.DND5E.currencies).map(([denom, { conversion }]) => {
-      return [denom, conversion];
-    }).sort(([, a], [, b]) => priority === "high" ? a - b : b - a);
+    const currencies = Object.entries(CONFIG.DND5E.currencies)
+      .filter(([denom]) => !exact || (denom !== denomination))
+      .map(([denom, { conversion }]) => [denom, conversion])
+      .sort(([, a], [, b]) => priority === "high" ? a - b : b - a);
     const baseConversion = CONFIG.DND5E.currencies[denomination].conversion;
     if ( exact ) currencies.unshift([denomination, baseConversion]);
 
     let passes = currencies.length;
     let updates;
-    while ( passes > 0 ) {
+    while ( passes ) {
       updates = { system: { currency: { ...currency } }, remainder: amount, item: [] };
       for ( const [denom, conversion] of currencies ) {
         const multiplier = conversion / baseConversion;
@@ -292,7 +293,6 @@ export default class CurrencyManager extends Application5e {
         updates.system.currency[denom] -= deduct;
         if ( !updates.remainder ) return updates;
       }
-      if ( !updates.remainder ) break;
       currencies.push(currencies.shift());
       passes--;
     }


module/applications/currency-manager.mjs
```diff
@@ -273,21 +273,28 @@ export default class CurrencyManager extends Application5e {
...
-    for ( const [denom, conversion] of currencies ) {
-      const multiplier = conversion / baseConversion;
-      const deduct = Math.min(updates.system.currency[denom], Math.floor(updates.remainder * multiplier));
-      updates.remainder -= deduct / multiplier;
-      updates.system.currency[denom] -= deduct;
-      if ( !updates.remainder ) return updates;
+
+    let passes = currencies.length;
+    let updates;
+    while ( passes > 0 ) {
```
Fyorl:
```suggestion
    while ( passes ) {
```

module/applications/currency-manager.mjs
```diff
@@ -273,21 +273,28 @@ export default class CurrencyManager extends Application5e {
...
+      updates = { system: { currency: { ...currency } }, remainder: amount, item: [] };
+      for ( const [denom, conversion] of currencies ) {
+        const multiplier = conversion / baseConversion;
+        const deduct = Math.min(updates.system.currency[denom], Math.floor(updates.remainder * multiplier));
+        updates.remainder -= deduct / multiplier;
+        updates.system.currency[denom] -= deduct;
+        if ( !updates.remainder ) return updates;
+      }
+      if ( !updates.remainder ) break;
```
Fyorl: We already return in the loop so this can never be true.

module/applications/currency-manager.mjs
```diff
@@ -273,21 +273,28 @@ export default class CurrencyManager extends Application5e {
...
-    const updates = { system: { currency: { ...currency } }, remainder: amount, item: [] };
-    if ( amount <= 0 ) return updates;
+    if ( amount <= 0 ) return { system: { currency: { ...currency } }, remainder: amount, item: [] };
 
     const currencies = Object.entries(CONFIG.DND5E.currencies).map(([denom, { conversion }]) => {
       return [denom, conversion];
     }).sort(([, a], [, b]) => priority === "high" ? a - b : b - a);
     const baseConversion = CONFIG.DND5E.currencies[denomination].conversion;
-
     if ( exact ) currencies.unshift([denomination, baseConversion]);
```
Fyorl: If `exact` is true, we add a duplicate denomination, so I think we want to filter that out first.

```js
const currencies = Object.entries(CONFIG.DND5E.currencies)
  .filter(([denom]) => !exact || denom !== denomination)
  .map(([denom, { conversion }]) => [denom, conversion])
  .sort(([, a], [, b]) => priority === "high" ? a - b : b - a);
```

arbron:
> Not sure if this PR is intended to fix all or just some of the issues in #6627. It doesn't handle the the case of deducting 1gp from an actor with 1pp (`Math.floor(1 * 0.1) = 0`). We would need to generate change for that, and I'm not sure if this method is intending to do that.

This fix is focused on fixing the behavior that broke at some point in the past. I added another issue (https://github.com/foundryvtt/dnd5e/issues/6708) to add a `makeChange` option that would allow the system to go one step further.

PR #6709, opened 2026-02-09
[#6661] More generalized and configurable rest management

### LINKED ISSUE #6661: Rest buttons do not infer from CONFIG
As a developer, I expected adjusting `CONFIG.DND5E.restTypes` to propagate to the sheet header. However, the buttons are hard-coded, meaning that adjustments to the labels, icons, and adding any new rests does not have an effect.

Aioros:
An attempt to generalize some rest logic to allow developers to define new types of rest that are then directly usable from actor sheets.
Adds a few property to RestTypeConfiguration:
- `dialog` (already present in RestConfiguration);
- `dialogClass` (also added to RestConfiguration), to allow a fully custom configuration dialog;
- `chat`, which controls the result chat message.

Common blocks of the existing rest dialogs have been extracted into partials, which should make it a little easier to integrate into any custom dialog.

One part that I don't feel great about is the generic rest hooks, which use the rest `type` to compose their actual name. If that's too questionable, we might want to switch to a fixed generic name plus specific ones for types `short` and `long`.

Closes #6661

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -1,17 +1,17 @@
+import LongRestDialog from "./applications/actor/rest/long-rest-dialog.mjs";
+import ShortRestDialog from "./applications/actor/rest/short-rest-dialog.mjs";
 import CalenderHUD from "./applications/calendar/calendar-hud.mjs";
 import MapLocationControlIcon from "./canvas/map-location-control-icon.mjs";
 import { ConsumptionTargetData } from "./data/activity/fields/consumption-targets-field.mjs";
+import VehicleData from "./data/actor/vehicle.mjs";
 import { CalendarGreyhawk, CALENDAR_OF_GREYHAWK } from "./data/calendar/calendar-of-greyhawk.mjs";
 import { CalendarHarptos, CALENDAR_OF_HARPTOS } from "./data/calendar/calendar-of-harptos.mjs";
 import { CalendarKhorvaire, CALENDAR_OF_KHORVAIRE } from "./data/calendar/calendar-of-khorvaire.mjs";
+import MappingField from "./data/fields/mapping-field.mjs";
 import * as activities from "./documents/activity/_module.mjs";
 import Actor5e from "./documents/actor/actor.mjs";
 import * as advancement from "./documents/advancement/_module.mjs";
 import { preLocalize } from "./utils.mjs";
-import MappingField from "./data/fields/mapping-field.mjs";
-import VehicleData from "./data/actor/vehicle.mjs";
-import ShortRestDialog from "./applications/actor/rest/short-rest-dialog.mjs";
-import LongRestDialog from "./applications/actor/rest/long-rest-dialog.mjs";
 
 /**
  * @import {
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -2083,8 +2083,8 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /**
    * Initiate a rest, spending or recovering hit dice, resources, item uses and spell slots according to configuration.
-   * @param {Partial<RestConfiguration>} [config]  Configuration options for the rest.
-   * @returns {Promise<RestResult>}                A Promise which resolves once the rest workflow has completed.
+   * @param {Partial<RestConfiguration>} config  Configuration options for the rest. Must include the rest type.
+   * @returns {Promise<RestResult>}              A Promise which resolves once the rest workflow has completed.
    */
   async initiateRest(config={}) {
     if ( this.system.isVehicle ) return;
@@ -2096,7 +2096,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const clone = this.clone();
     const restConfig = CONFIG.DND5E.restTypes[config.type];
     config = foundry.utils.mergeObject({
-      type: restConfig.type, dialog: true, chat: restConfig.chat !== false,
+      dialog: true, chat: restConfig.chat !== false,
       duration: restConfig.duration[game.settings.get("dnd5e", "restVariant")],
       newDay: restConfig.newDay === true,
       advanceBastionTurn: restConfig.advanceBastionTurn === true, advanceTime: restConfig.advanceTime === true,
@@ -2104,9 +2104,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       recoverTemp: restConfig.recoverTemp, recoverTempMax: restConfig.recoverTempMax,
       exhaustionDelta: restConfig.exhaustionDelta
     }, config);
-    if ( !Object.hasOwn(config, "dialogClass") ) {
-      config.dialogClass = restConfig.dialogClass ?? BaseRestDialog;
-    }
+    config.dialogClass ??= restConfig.dialogClass ?? BaseRestDialog;
 
     /**
      * A hook event that fires before a rest is started. The actual name of the hook will depend on the rest type
--- a/templates/actors/rest/base-rest.hbs
+++ b/templates/actors/rest/base-rest.hbs
@@ -1,5 +1,5 @@
 <section class="flexcol">
-    {{> "dnd5e.formlist" formSections}}
+    {{> "dnd5e.formlist" formSections }}
 
     {{#if request}}
     {{> "systems/dnd5e/templates/actors/rest/parts/rest-request.hbs" }}
--- a/templates/actors/rest/long-rest.hbs
+++ b/templates/actors/rest/long-rest.hbs
@@ -3,7 +3,7 @@
         {{ localize (ifThen isGroup "DND5E.REST.Long.Hint.Group" "DND5E.REST.Long.Hint.Normal") }}
     </div>
 
-    {{> "dnd5e.formlist" formSections}}
+    {{> "dnd5e.formlist" formSections }}
 
     {{#if request}}
     {{> "systems/dnd5e/templates/actors/rest/parts/rest-request.hbs" }}
--- a/templates/actors/rest/short-rest.hbs
+++ b/templates/actors/rest/short-rest.hbs
@@ -3,7 +3,7 @@
         {{ localize (ifThen isGroup "DND5E.REST.Short.Hint.Group" "DND5E.REST.Short.Hint.Normal") }}
     </div>
 
-    {{> "dnd5e.formlist" formSections}}
+    {{> "dnd5e.formlist" formSections }}
 
     {{#if hitDice}}
     {{#if hitDice.canRoll}}


module/documents/actor/actor.mjs
```diff
@@ -2082,36 +2083,33 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
       return;
     }
 
     const clone = this.clone();
-    const restConfig = CONFIG.DND5E.restTypes.short;
+    const restConfig = CONFIG.DND5E.restTypes[type];
     config = foundry.utils.mergeObject({
-      type: "short", dialog: true, chat: true, newDay: false, advanceTime: false, autoHD: false, autoHDThreshold: 3,
-      duration: CONFIG.DND5E.restTypes.short.duration[game.settings.get("dnd5e", "restVariant")],
+      type, dialog: true, dialogClass: restConfig.dialogClass ?? BaseRestDialog, chat: !!restConfig.chat,
```
krbz999: I think a `mergeObject` here might mutate the `dialogClass`. It might need to be moved outside.

Aioros: Thanks Zhell, but I'll have to admit that I'm not sure what you meant (and take the chance to actually learn something here). What are we trying to avoid?

module/documents/actor/actor.mjs
```diff
@@ -2082,36 +2083,36 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
   /*  Resting                                     */
   /* -------------------------------------------- */
 
-  /**
-   * Take a short rest, possibly spending hit dice and recovering resources, item uses, and relevant spell slots.
-   * @param {Partial<RestConfiguration>} [config]  Configuration options for a short rest.
-   * @returns {Promise<RestResult>}                A Promise which resolves once the short rest workflow has completed.
-   */
-  async shortRest(config={}) {
+  async initiateRest(type, config={}) {
```
arbron: Lost the JSDoc here. Also, `RestConfiguration` includes a `type` parameter, so rather than passing in the type separately you can just pass it in with the configuration.

module/documents/actor/actor.mjs
```diff
@@ -2082,36 +2083,36 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    if ( !Object.hasOwn(config, "dialogClass") ) {
+      config.dialogClass = restConfig.dialogClass ?? BaseRestDialog;
+    }
 
     /**
-     * A hook event that fires before a short rest is started.
-     * @function dnd5e.preShortRest
+     * A hook event that fires before a rest is started.
+     * The actual name of the hook will depend on the rest type.
+     * @function dnd5e.preGenericRest
```
arbron: For variable hooks like this, we usually we just don't include the part of the name that is replaced:

```suggestion
     * @function dnd5e.preRest
```

module/documents/actor/actor.mjs
```diff
@@ -2082,36 +2083,36 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
     }, config);
+    if ( !Object.hasOwn(config, "dialogClass") ) {
+      config.dialogClass = restConfig.dialogClass ?? BaseRestDialog;
+    }
 
     /**
-     * A hook event that fires before a short rest is started.
-     * @function dnd5e.preShortRest
+     * A hook event that fires before a rest is started.
+     * The actual name of the hook will depend on the rest type.
```
```suggestion
     * A hook event that fires before a rest is started. The actual name of the hook will depend on the rest type
     * (e.g. `dnd5e.preShortRest` or `dnd5e.preLongRest`).
```

module/documents/actor/actor.mjs
```diff
@@ -2120,19 +2121,19 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-        Object.assign(config, await ShortRestDialog.configure(this, config));
+        Object.assign(config, await config.dialogClass.configure(this, config));
       } catch(err) { return; }
     }
 
     /**
-     * A hook event that fires after a short rest has started, after the configuration is complete.
-     * @function dnd5e.shortRest
+     * A hook event that fires after a rest has started, after the configuration is complete.
+     * @function dnd5e.genericRest
```
```suggestion
     * @function dnd5e.rest
```

module/documents/actor/actor.mjs
```diff
@@ -2143,6 +2144,15 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     return this._rest(config, { clone, dhd, dhp });
   }
 
+  /**
```
arbron, line 2152:
```suggestion
  /* -------------------------------------------- */

  /**
```

module/documents/actor/actor.mjs
```diff
@@ -2143,6 +2144,15 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+   * Take a short rest, possibly spending hit dice and recovering resources, item uses, and relevant spell slots.
+   * @param {Partial<RestConfiguration>} [config]  Configuration options for a short rest.
+   * @returns {Promise<RestResult>}                A Promise which resolves once the short rest workflow has completed.
+   */
+  async shortRest(config={}) {
+    return this.initiateRest("short", { dialogClass: ShortRestDialog, chat: true, autoHD: false, autoHDThreshold: 3, ...config });
```
arbron: Rather than passing the dialog class in here, I'd say add it to the `CONFIG.DND5E.restType` so it can be customized ahead of time.


```suggestion
    return this.initiateRest("short", { autoHD: false, autoHDThreshold: 3, ...config });
```

module/documents/actor/actor.mjs
```diff
@@ -2082,36 +2083,36 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
       return;
     }
 
     const clone = this.clone();
-    const restConfig = CONFIG.DND5E.restTypes.short;
+    const restConfig = CONFIG.DND5E.restTypes[type];
     config = foundry.utils.mergeObject({
-      type: "short", dialog: true, chat: true, newDay: false, advanceTime: false, autoHD: false, autoHDThreshold: 3,
-      duration: CONFIG.DND5E.restTypes.short.duration[game.settings.get("dnd5e", "restVariant")],
+      type, dialog: true, chat: !!restConfig.chat,
```
arbron: Rather than passing this in as `true` for both `shortRest` and `longRest`, it seems cleaner to treat a missing value as `true` and then only pull from the config if there is an explicit value set.

```suggestion
      type, dialog: true, chat: restConfig.chat !== false,
```

module/documents/actor/actor.mjs
```diff
@@ -2205,7 +2174,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
    * @returns {Promise<ChatMessage5e|null>}
    */
   static async handleRestRequest(actor, request, config) {
-    const result = await actor[config.type === "short" ? "shortRest" : "longRest"]({
+    const result = await actor.initiateRest(config.type, {
```
arbron: Because we are modifying the config in the `shortRest` and `longRest` methods, calling `initializeRest("short")` will be different than calling `shortRest()`. Those methods should probably be simplified to something like:

```javascript
  async shortRest(config={}) {
    return this.initiateRest({ type: "short", ...config });
  }
```

Everything else should be handled within `initializeRest`.

module/documents/actor/actor.mjs
```diff
@@ -2319,10 +2288,13 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
     // Determine the chat message to display
     let message;
-    if ( diceRestored && healthRestored ) message = `DND5E.REST.${length}.Result.Full`;
-    else if ( longRest && !diceRestored && healthRestored ) message = "DND5E.REST.Long.Result.HitPoints";
-    else if ( longRest && diceRestored && !healthRestored ) message = "DND5E.REST.Long.Result.HitDice";
-    else message = `DND5E.REST.${length}.Result.Short`;
+    if (typeof typeConfig.chat === "string") message = typeConfig.chat;
+    if (!message) {
```
```suggestion
    if ( typeof typeConfig.chat === "string" ) message = typeConfig.chat;
    if ( !message ) {
```

module/documents/_types.mjs
```diff
@@ -86,21 +86,23 @@
...
- * @property {number} [exhaustionDelta]      A delta exhaustion to apply to creatures undergoing this rest.
- * @property {ChatMessage5e} [request]       Rest request chat message for which this rest was performed.
+ * @property {string} type                          Type of rest to perform.
+ * @property {boolean} dialog                       Present a dialog window which allows for rolling hit dice
+ *                                                  as part of the rest and selecting whether a new day has occurred.
+ * @property {boolean} chat                         Should a chat message be created to summarize the results
+ *                                                  of the rest?
+ * @property {number} duration                      Amount of time passed during the rest in minutes.
+ * @property {boolean} newDay                       Does this rest carry over to a new day?
+ * @property {typeof BaseRestDialog} [dialogClass]  A class for the dialog window.
```
arbron:
```suggestion
 * @property {typeof BaseRestDialog} [dialogClass]  A class for the dialog window.
 * @property {boolean} chat                         Should a chat message be created to summarize the results
 *                                                  of the rest?
 * @property {number} duration                      Amount of time passed during the rest in minutes.
 * @property {boolean} newDay                       Does this rest carry over to a new day?
```

module/_types.mjs
```diff
@@ -397,6 +397,11 @@
  * @property {string} label                         Localized label for the rest type.
  * @property {string} icon                          Icon representing this rest type. Can be either a set of FontAwesome
  *                                                  classes or an image path.
+ * @property {boolean} dialog                       Present a dialog window which allows for rolling hit dice as part of
+ *                                                  the rest and selecting whether a new day has occurred.
+ * @property {typeof BaseRestDialog} [dialogClass]  A class for the dialog window.
+ * @property {string|boolean} [chat]                A chat message to be created to summarize the results of the rest,
+ *                                                  or a boolean indicating whether to create it or not.
```
arbron:
```suggestion
 * @property {string|boolean} [chat=true]           A localization string for the chat text created to summarize the
 *                                                  results of the rest, or a boolean indicating whether to create it
 *                                                  or not.
```

module/_types.mjs
```diff
@@ -397,6 +397,11 @@
  * @property {string} label                         Localized label for the rest type.
  * @property {string} icon                          Icon representing this rest type. Can be either a set of FontAwesome
  *                                                  classes or an image path.
+ * @property {boolean} dialog                       Present a dialog window which allows for rolling hit dice as part of
+ *                                                  the rest and selecting whether a new day has occurred.
```
arbron: We don't usually control default dialog and message creation options in type configuration like this, instead defaulting to `true` and leaving it up to the caller or hooks to handle modifying this if desired. My assumption is that all rests would want to present a confirmation dialog and suppressing that is an outlier best handled using those other methods.

templates/actors/rest/parts/hit-dice.hbs
```diff
@@ -0,0 +1,19 @@
...
+            <i class="fa-solid fa-dice-d20" inert></i>
+        </button>
+    </div>
+    {{ formField autoRoll name="autoHD" value=config.autoHD input=inputs.createCheckboxInput rootId=partId }}
+</fieldset>
+{{else}}
+<div class="note warn">
+    {{ localize "DND5E.REST.HitDice.None" }}
+</div>
+{{/if}}
```
arbron:
```suggestion
{{/if}}

```

templates/actors/npc-header.hbs
```diff
@@ -251,16 +251,12 @@
...
-                        <button type="button" class="gold-button" data-action="rest" data-type="long"
-                                data-tooltip="DND5E.REST.Long.Label"
-                                aria-label="{{ localize 'DND5E.REST.Long.Label' }}">
-                            <i class="fas fa-campground" inert></i>
-                        </button>
+                        {{#each CONFIG.restTypes}}
+                            <button type="button" class="gold-button" data-action="rest" data-type="{{ @key }}"
+                                data-tooltip="{{ label }}"
+                                aria-label="{{ localize label }}">
+                            <i class="{{ icon }}" inert></i>
```
arbron:
```suggestion
                            <button type="button" class="gold-button" data-action="rest" data-type="{{ @key }}"
                                    data-tooltip aria-label="{{ localize label }}">
                                <i class="{{ icon }}" inert></i>
                            </button>
```

templates/actors/character-header.hbs
```diff
@@ -38,14 +38,11 @@
...
-                    <i class="fas fa-utensils" inert></i>
-                </button>
-                <button type="button" class="long-rest gold-button" data-action="rest" data-type="long"
-                        data-tooltip="DND5E.REST.Long.Label" aria-label="{{ localize 'DND5E.REST.Long.Label' }}">
-                    <i class="fas fa-campground" inert></i>
-                </button>
+                {{#each CONFIG.restTypes}}
+                <button type="button" class="{{ @key }}-rest gold-button" data-action="rest" data-type="{{ @key }}"
+                    data-tooltip="{{ label }}" aria-label="{{ localize label }}">
+                <i class="{{ icon }}" inert></i>
```
arbron:
```suggestion
                <button type="button" class="{{ @key }}-rest gold-button" data-action="rest" data-type="{{ @key }}"
                        data-tooltip aria-label="{{ localize label }}">
                    <i class="{{ icon }}" inert></i>
                </button>
```

templates/actors/rest/base-rest.hbs
```diff
@@ -0,0 +1,13 @@
...
+    {{> "systems/dnd5e/templates/actors/rest/parts/fields.hbs" }}
+    {{/if}}
+
+    {{#if hitPoints.length}}
+    {{> "systems/dnd5e/templates/actors/rest/parts/hit-points.hbs" }}
+    {{/if}}
+
+    {{#if request}}
+    {{> "systems/dnd5e/templates/actors/rest/parts/rest-request.hbs" }}
+    {{/if}}
```
arbron: Rather than adding several more of these very simple template that just include a legend and field list, we should probably use "templates/shared/fields/formlist.hbs" instead. This requires preparing an array of the individual field sets in the app instead.

```suggestion
    {{> "dnd5e.formlist" formSections}}

    {{#if request}}
    {{> "systems/dnd5e/templates/actors/rest/parts/rest-request.hbs" }}
    {{/if}}
```

Then in the rest application, you'd define something like:

```javascript
context.formSections = [
  { legend: "DND5E.REST.Configuration", fields: […] },
  { legend: "DND5E.HitPoints", fields: […] }
];
```

We'll have to keep hit dice separate because we can't handle the rolling button in this way. That also means keeping request separate so they appear in the correct order.

templates/actors/group/header.hbs
```diff
@@ -6,14 +6,11 @@
...
-        </button>
-        <button type="button" class="long-rest gold-button" data-action="rest" data-type="long" data-tooltip
-                aria-label="{{ localize 'DND5E.REST.Long.Label' }}">
-            <i class="fas fa-campground" inert></i>
-        </button>
+        {{#each CONFIG.restTypes}}
+        <button type="button" class="{{ @key }}-rest gold-button" data-action="rest" data-type="{{ @key }}" data-tooltip
+                aria-label="{{ localize label }}">
+            <i class="{{ icon }}" inert></i>
+        {{/each}}
```
arbron, line 14:
```suggestion
            <i class="{{ icon }}" inert></i>
        </button>
        {{/each}}
```

module/documents/actor/actor.mjs
Aioros: Quick question on this one while I work on the rest of the changes: I think this would require adding to `RestTypeConfiguration` some more properties that are currently only on `RestConfiguration` (specifically `newDay`, `autoHD`, and `autoHDThreshold` - and at that point no reason to exclude `advanceTime` and `advanceBastionTurn`, I suppose); just checking if that sounds ok.

EDIT: to clarify what I mean a little bit: if `initiateRest` needs to be able to infer stuff like `newDay` and `autoHD` for "short" and "long", either those properties are directly in the RestTypeConfiguration or the method has to do things like `type === "long"` (which I assume is not great)

arbron: Auto HD threshold can probably just be a constant, because it doesn't have an effect if `autoHD` isn't set and I don't imagine new rest types will need to set it. We can just let people override that in a hook if needed.

For the rest then yeah they should probably go into rest configuration.

module/config.mjs
```diff
@@ -10,6 +10,8 @@ import * as advancement from "./documents/advancement/_module.mjs";
 import { preLocalize } from "./utils.mjs";
 import MappingField from "./data/fields/mapping-field.mjs";
 import VehicleData from "./data/actor/vehicle.mjs";
+import ShortRestDialog from "./applications/actor/rest/short-rest-dialog.mjs";
+import LongRestDialog from "./applications/actor/rest/long-rest-dialog.mjs";
```
arbron: These two application imports should be moved up to the top and reversed (the existing data imports here are also in the wrong place, the `MappingField` import should be below the calendar imports and the `VehicleData` import should be above it).

We try to keep our imports organized by depth first and then alphabetically by path.

templates/actors/rest/long-rest.hbs
```diff
@@ -3,21 +3,9 @@
...
-    </fieldset>
-    {{/if}}
-
-    {{#if hitPoints.length}}
-    <fieldset>
-        <legend>{{ localize "DND5E.HitPoints" }}</legend>
-        {{> "dnd5e.fieldlist" hitPoints }}
-    </fieldset>
-    {{/if}}
+    {{> "dnd5e.formlist" formSections}}
```
arbron:
```suggestion
    {{> "dnd5e.formlist" formSections }}
```

templates/actors/rest/base-rest.hbs
```diff
@@ -0,0 +1,7 @@
+<section class="flexcol">
+    {{> "dnd5e.formlist" formSections}}
```
arbron:
```suggestion
    {{> "dnd5e.formlist" formSections }}
```

templates/actors/rest/short-rest.hbs
```diff
@@ -3,19 +3,7 @@
...
-    </fieldset>
-    {{/if}}
-
-    {{#if hitPoints.length}}
-    <fieldset>
-        <legend>{{ localize "DND5E.HitPoints" }}</legend>
-        {{> "dnd5e.fieldlist" hitPoints }}
-    </fieldset>
-    {{/if}}
+    {{> "dnd5e.formlist" formSections}}
```
arbron:
```suggestion
    {{> "dnd5e.formlist" formSections }}
```

Aioros:
Thanks, I'll wait for Kim's notes before cleaning up these last few things (I can't believe I keep missing bracket spacings, by the way).

module/documents/actor/actor.mjs
```diff
@@ -2083,35 +2082,42 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+      duration: restConfig.duration[game.settings.get("dnd5e", "restVariant")],
+      newDay: restConfig.newDay === true,
+      advanceBastionTurn: restConfig.advanceBastionTurn === true, advanceTime: restConfig.advanceTime === true,
+      autoHD: restConfig.autoHD === true, autoHDThreshold: 3,
       recoverTemp: restConfig.recoverTemp, recoverTempMax: restConfig.recoverTempMax,
       exhaustionDelta: restConfig.exhaustionDelta
     }, config);
+    if ( !Object.hasOwn(config, "dialogClass") ) {
+      config.dialogClass = restConfig.dialogClass ?? BaseRestDialog;
+    }
```
```suggestion
    config.dialogClass ??= restConfig.dialogClass ?? BaseRestDialog;
```

module/documents/actor/actor.mjs
```diff
@@ -2083,35 +2082,42 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
       return;
     }
 
     const clone = this.clone();
-    const restConfig = CONFIG.DND5E.restTypes.short;
+    const restConfig = CONFIG.DND5E.restTypes[config.type];
     config = foundry.utils.mergeObject({
-      type: "short", dialog: true, chat: true, newDay: false, advanceTime: false, autoHD: false, autoHDThreshold: 3,
-      duration: CONFIG.DND5E.restTypes.short.duration[game.settings.get("dnd5e", "restVariant")],
+      type: restConfig.type, dialog: true, chat: restConfig.chat !== false,
```
Fyorl: `RestTypeConfiguration` doesn't have a `type` property (maybe it should?). In practice this is always overridden by `config.type`, but `config` is technically optional.


```suggestion
      dialog: true, chat: restConfig.chat !== false,
```

module/documents/actor/actor.mjs
```diff
@@ -2083,35 +2082,42 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
   /* -------------------------------------------- */
 
   /**
-   * Take a short rest, possibly spending hit dice and recovering resources, item uses, and relevant spell slots.
-   * @param {Partial<RestConfiguration>} [config]  Configuration options for a short rest.
-   * @returns {Promise<RestResult>}                A Promise which resolves once the short rest workflow has completed.
+   * Initiate a rest, spending or recovering hit dice, resources, item uses and spell slots according to configuration.
+   * @param {Partial<RestConfiguration>} [config]  Configuration options for the rest.
```
Fyorl: I think `config` is not optional anymore. At least `config.type` must always be provided it looks like.


```suggestion
   * @param {Partial<RestConfiguration>} config  Configuration options for the rest. Must include the rest type.
```

PR #6713, opened 2026-02-11
[#6528] Add separate gear data, gear configuration app

### LINKED ISSUE #6528: Add method to more precisely control what appears in Gear list/Feature tab on NPC sheet
Currently weapons on NPCs can be excluded from the gear list by setting their type to "Natural", but that doesn't work for any other physical item types. It is also impossible to add an item to an NPC that is in the gear list but doesn't appear on the Features tab or in the stat block (for example, a NPC might have a special "Poisoned Dagger" attack they use, but their gear lists several mundane daggers).

We could perhaps implement a flag that allows for more precise control over whether items appear in just the Gear list, just on the Features, or in both locations.

arbron:
Rather than trying to infer NPCs' gear from their items, this makes the gear list an explicitly defined data structure to allow authors more control. The gear list includes a absolute or relative UUID to the available gear and an optional quantity.

Gear Configuration

A migration has been added that takes the existing logic for showing the gear list in stat block embeds and uses that to pre-populate the list of gear.

Minor breaking change in the signature of `getGear`, which is now an asynchronous method because it needs to fetch compendium items.

### Todo
- [x] Automatically populate gear list when certain compendium items are dropped onto the NPC sheet
- [x] Add context option for adding/removing item from gear list
- [x] Make gear list entries draggable from NPC sheet

Closes #6528


module/applications/actor/npc-sheet.mjs
```diff
@@ -581,4 +596,37 @@ export default class NPCActorSheet extends BaseActorSheet {
...
+    return super._onDragStart(event);
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _onDropSingleItem(event, itemData) {
+    const result = await super._onDropSingleItem(event, itemData);
+    if ( !result ) return result;
+    if ( ("quantity" in (itemData.system ?? {})) && itemData._stats?.compendiumSource ) {
```
Fyorl: Do we want to add every 'physical' item that is dropped onto an NPC sheet to their gear? We implement a lot of monster attacks as natural weapons which have quantity. Also noticed that if you drop gear onto the sheet it came from, it will stack another batch of that item into the NPC's inventory.

module/applications/components/inventory.mjs
```diff
@@ -421,6 +424,13 @@ export default class InventoryElement extends (foundry.applications.elements.Ado
       condition: () => ("favorites" in this.actor.system) && item.isOwner && !compendiumLocked,
       callback: li => this._onAction(li, "toggleFavorite"),
       group: "state"
+    }, {
+      name: inGear ? "DND5E.Gear.Action.Remove" : "DND5E.Gear.Action.Add",
+      icon: '<i class="fa-solid fa-person-hiking fa-fw"></i>',
+      condition: () => !!this.actor.system.details?.treasure?.gear && item.isOwner && !compendiumLocked
+        && !!CONFIG.Item.dataModels[item.type]?.schema.fields.quantity,
```
Fyorl:
```suggestion
        && CONFIG.Item.dataModels[item.type]?.schema.has("quantity"),
```

module/data/actor/npc.mjs
```diff
@@ -230,6 +234,30 @@ export default class NPCData extends CreatureTemplate {
 
   /* -------------------------------------------- */
 
+  /**
+   * Migrate gear list inferred from items to explicitly-defined list.
+   * @param {object} source  The item's candidate source data to migrate.
+   */
+  static _migrateGear(source) {
+    if ( source.system?.details?.treasure?.gear !== undefined ) return;
+    const gear = source.items?.values()
```
Fyorl: I'm not sure if there's a way to make this migration resilient to partial updates. It's rare but possible to do something like `actor.update({ items: [{ /* Item A source */ }, { /* Item B source /* }] })`. In that case, `gear` would be `undefined`, so we would migrate `source.items` and overwrite `gear` with a list generated from only the two items in `source`.

I think this migration can only be a world migration. For Actors in packs that have not been migrated by the pack author yet, I think we can check `_stats.systemVersion` in `getGear` and fall back to the old method of determining gear.

templates/actors/npc-sidebar.hbs
```diff
@@ -19,30 +19,8 @@
...
-                   data-action="roll" data-type="skill">
-                    <span class="label">{{ label }}</span>
-                    <span class="separator">&vert;</span>
-                    <span class="value">{{ dnd5e-numberFormat total signDisplay="always" }}</span>
-                </a>
-            </li>
-            {{/each}}
-        </ul>
-    </div>
+    {{> "dnd5e.actor-trait-pills" values=skills key="skills" icon="fas fa-briefcase"
```
Fyorl, line 22: Nice

Fyorl (changes requested):
Looks like a reasonable solution to me, but the quantities might be a problem. If you set up an Actor with, say, Javelin (5) in its gear entry, and in its inventory, then throughout the course of an encounter, it might throw several of those javelins. If the DM grabs the Javelin entry from the gear section in order to distribute loot, they will take the full 5 javelins, even if most of them had otherwise been thrown off a cliff.

I would propose that we instead add `gear` as a property to all physical items, keep the context menu toggle, keep the gear section in the sidebar. Instead of an explicit gear list, we instead derive it by filtering the list of items that are marked 'gear'. When dragging-and-dropping from an NPC, we check the gear property, and instead drag a compendium copy, using the same logic currently in the PR.

It has some advantages in that we don't need any `_clearGear` logic and extra round-trips, and we don't need a bespoke gear configuration app.

For Ember compatibility we would need a way to specify a subset of properties that are maintained on the item and are not replaced by the compendium source (we need to retain quantity anyway).

module/data/actor/npc.mjs
```diff
@@ -500,10 +512,14 @@ export default class NPCData extends CreatureTemplate {
...
    */
-  getGear() {
-    return this.parent.items
-      .filter(i => i.system.quantity && (i.system.type?.value !== "natural"))
-      .sort((lhs, rhs) => lhs.name.localeCompare(rhs.name, game.i18n.lang));
+  async getGear() {
+    const actorOld = foundry.utils.isNewerVersion("5.3", this.parent._stats.systemVersion);
+    return (await Promise.all(this.parent.items
+      .filter(i => i.system.quantity && (actorOld && foundry.utils.isNewerVersion("5.3", i._stats.systemVersion)
+        ? (i.system.type?.value !== "natural") : i.system.properties?.has("gear")))
```
arbron: Running into a problem with this version checking at the moment. It works great for non-migrated actors, but when the migration happens the `_stats.systemVersion` isn't getting updated for the new version. Anything we could tweak in the migrations to get that stat to update properly?

Fyorl: The Actor's `systemVersion` isn't being updated? I will take a look

module/data/item/consumable.mjs
```diff
@@ -283,6 +283,14 @@ export default class ConsumableData extends ItemDataModel.mixin(
   /*  Socket Event Handlers                       */
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _preCreate(data, options, user) {
+    if ( (await super._preCreate(data, options, user)) === false ) return false;
+    await this.preCreateGear(data, options, user);
```
Fyorl:
```suggestion
    this.preCreateGear(data, options, user);
```

module/data/item/container.mjs
```diff
@@ -343,6 +343,14 @@ export default class ContainerData extends ItemDataModel.mixin(
   /*  Socket Event Handlers                       */
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _preCreate(data, options, user) {
+    if ( (await super._preCreate(data, options, user)) === false ) return false;
+    await this.preCreateGear(data, options, user);
```
Fyorl:
```suggestion
    this.preCreateGear(data, options, user);
```

module/data/item/equipment.mjs
```diff
@@ -342,6 +343,7 @@ export default class EquipmentData extends ItemDataModel.mixin(
   async _preCreate(data, options, user) {
     if ( (await super._preCreate(data, options, user)) === false ) return false;
     await this.preCreateEquipped(data, options, user);
+    await this.preCreateGear(data, options, user);
```
Fyorl:
```suggestion
    this.preCreateGear(data, options, user);
```

module/data/item/loot.mjs
```diff
@@ -145,6 +145,14 @@ export default class LootData extends ItemDataModel.mixin(
   /*  Socket Event Handlers                       */
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _preCreate(data, options, user) {
+    if ( (await super._preCreate(data, options, user)) === false ) return false;
+    await this.preCreateGear(data, options, user);
```
Fyorl:
```suggestion
    this.preCreateGear(data, options, user);
```

module/data/item/tool.mjs
```diff
@@ -230,6 +230,7 @@ export default class ToolData extends ItemDataModel.mixin(
   /** @inheritDoc */
   async _preCreate(data, options, user) {
     if ( (await super._preCreate(data, options, user)) === false ) return false;
+    await this.preCreateGear(data, options, user);
```
Fyorl:
```suggestion
    this.preCreateGear(data, options, user);
```

module/data/item/weapon.mjs
```diff
@@ -572,6 +573,7 @@ export default class WeaponData extends ItemDataModel.mixin(
   async _preCreate(data, options, user) {
     if ( (await super._preCreate(data, options, user)) === false ) return false;
     await this.preCreateEquipped(data, options, user);
+    await this.preCreateGear(data, options, user);
```
Fyorl:
```suggestion
    this.preCreateGear(data, options, user);
```

module/migration.mjs
```diff
@@ -567,6 +569,17 @@ export function migrateItemData(item, itemData, migrationData, flags={}) {
     updateData["flags.dnd5e.-=migratedProperties"] = null;
   }
 
+  // Migrate gear property
+  if ( (flags.actorData?.type === "npc") && item.system.quantity && (item.system.type?.value !== "natural")
+    && (!["equipment", "weapon"].includes(item.type) || item._stats.compendiumSource)
+    && !item._stats.compendiumSource?.startsWith("Compendium.dnd-monster-manual.features.")
```
Fyorl, line 580: There is the MM, but also SRD 5.1 & 5.2 have monster feature compendia, no? Probably other content as well. If we didn't have this PR, how would this content behave? I assume our current heuristic would still flag these things as gear regardless?

module/migration.mjs
```diff
@@ -567,6 +569,17 @@ export function migrateItemData(item, itemData, migrationData, flags={}) {
...
+    && (flags.bypassVersionCheck || foundry.utils.isNewerVersion("5.3.0", item._stats.systemVersion)) ) {
+    if ( !("system.properties" in updateData) ) {
+      updateData["system.properties"] = foundry.utils.getProperty(itemData, "system.properties") ?? [];
+    }
+    updateData["system.properties"].push("gear");
```
Fyorl, line 585: We should probably make sure `system.properties` doesn't already have `gear` in order to be idempotent.


```suggestion
    if ( !updateData["system.properties"].includes("gear") ) updateData["system.properties"].push("gear");
```

module/applications/actor/npc-sheet.mjs
```diff
@@ -322,6 +322,20 @@ export default class NPCActorSheet extends BaseActorSheet {
...
+    // Gear
+    const gear = await this.actor.system.getGear();
+    if ( gear.length ) context.gear = gear.map(item => ({
+      draggable: true,
+      label: item.name,
+      link: {
+        action: "showDocument",
+        itemId: foundry.utils.parseUuid(item.getFlag("dnd5e", "gearSource"))?.id,
+        quantity: item.system.quantity,
+        uuid: item.uuid
```
Fyorl, line 334: I think this would mean you get the fresh compendium item when clicking the gear item. Is that what we want? I ask because we are taking some pains to get the `gearSource` item ID above.

module/data/actor/npc.mjs
Fyorl: It seems we have logic in place that will only bump `systemVersion` if 'system-related' data was changed, e.g. a `system` key is provided with the data. This is to prevent issues where a system has yet to complete its migration, but a module does something like update module flags, inadvertently bumping `systemVersion` and potentially preventing some legitimate system migration from running. This won't be a problem when we have a system migration framework in place, but for now, we can bump the `systemVersion` manually when we know that's what we want. I've pushed a change to that effect.

The `systemVersion` on the items themselves will also not be bumped, but that's because of this long-standing issue https://github.com/foundryvtt/foundryvtt/issues/5241. We could similarly manually bump this, but the parent is generally considered the source of truth for things like version numbers so we should really only check the parent Actor's `systemVersion`, not the items'.

Fyorl (changes requested):

module/migration.mjs
arbron: This is mainly to fix certain items in the Monster Manual/SRD 5.2 content that point to weapons in the Monster Features compendium. These items are currently incorrectly included in the gear list (see Pact Blade on Cultist Fanatic), so if we leave this out then they will still appear in the list (which may be more correct since it would leave the actor behaving the same way as pre-migration).

module/applications/actor/npc-sheet.mjs
arbron: The intention was to have clicking these open the compendium version rather than the actor version because that is the canonical "gear" version of the item. I still needed to know which local item it was associated with in order to set the quantity correctly when you drag it.

module/data/actor/npc.mjs
arbron: Hmm, testing your change it appears to work fine in V14, but in V13 the `systemVersion` isn't changed.

Perhaps we should just remove the version checking from `getGear`. That would mean non-migrated compendiums wouldn't list any gear, but it would prevent gear showing up that shouldn't after migration.

Alternatively we could recommend using `refreshCompendium` rather than `migrateCompendium` because that does update `systemVersion` because it completely re-writes the whole document.

Fyorl: I think `refreshCompendium` won't help for World actors in v13, right? I'm ok with removing the version check though. Typically we don't go to such lengths with migrations and instead require that pack authors migrate their packs.

module/migration.mjs
Fyorl: Makes sense. I'll leave it up to you.

module/applications/actor/npc-sheet.mjs
Fyorl: Makes sense, just checking.

arbron:
@Fyorl Added the hook into `asGear` and handling of auto-converting gear when dropping onto actors

I also added a `preserverGear` flag allowing whole items to be preserved even if they have a `_stats.compendiumSource` set. I think the extra work to handle preserving only specific enchantments is a bit large and would probably be best handled in a followup PR rather than bloating this one further.

Fyorl (approved):

TheGiddyLimit:
If there a way to mark an incoming item as _not_ gear, without first adding the item and then editing it afterwards?
From a quick test/skim of e.g. https://github.com/foundryvtt/dnd5e/pull/6713/changes#diff-36b59bb67b8c9af4fd5c684aee18793aa7e15157371426935f652a5c52f1a361R239 I'm not seeing it

PR #6748, opened 2026-02-20
[#6747] Prepare spellcasting on NPCs based on methods set on spells

### LINKED ISSUE #6747: Prepare spellcasting on NPCs based on the methods selected for spells on actor
Currently NPCs will only automatically prepare spells slots using the default "spellcasting" method when their spellcaster level is set. If you want to implement a pact casting NPC (or one using any custom spellcasting type), then you first set their spellcaster level and then manually open the slot config app to give them the right number of slots in the correct spellcasting type and explicitly set the slots for normal spellcasting to zero.

It would be better if the actor could determine what type of spellcasting to prepare based on the spells present on their sheet, so if they only have spells set to the pact casting method, then the NPC will only get pact slots prepared, while if they have both pact and normal spells, then both sets of slots will be prepared.

Related: #6742

arbron:
When preparing spell slots for NPCs the system will now check which spellcasting methods are represented by spells on the actor. It will automatically set the progression using the spellcaster level for any spellcasting methods present on spells, and if no spells are present it will fall back to the default spellcasting type.

Closes #6747

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -543,7 +543,14 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     if ( this.system.isNPC && ("spell" in (this.system.attributes ?? {})) ) {
       const level = Object.values(progression).find(_ => _);
       if ( level ) this.system.attributes.spell.level = level;
-      else progression.spell = this.system.attributes.spell.level ?? 0;
+      else if ( this.system.attributes.spell.level > 0 ) {
+        const methods = this.itemTypes.spell.reduce((m, s) => {
+          if ( s.system.level && CONFIG.DND5E.spellcasting[s.system.method]?.slots ) m.add(s.system.method);
+          return m;
+        }, new Set());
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -543,7 +543,14 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     if ( this.system.isNPC && ("spell" in (this.system.attributes ?? {})) ) {
       const level = Object.values(progression).find(_ => _);
       if ( level ) this.system.attributes.spell.level = level;
-      else progression.spell = this.system.attributes.spell.level ?? 0;
+      else if ( this.system.attributes.spell.level > 0 ) {
```
Fyorl, line 546: Slight change in behaviour in that, previously, `progression.spell` would always be at least `0`. Not sure if it's now `undefined`, or maybe data model cleaning sets it to `0` anyway. Unsure if that will cause any problems.

arbron: Earlier in the method we are already pre-populating progression for each spellcasting method at `0`, so it will remain there and avoid the loop over the spells if not necessary.

PR #6749, opened 2026-02-20
Tooltips for skills and tools are not displayed.

ArtemBogdanov:
Fixed the bug where tooltips for skills and tools are not displayed on the character sheet

Fixed adding references for skills and tools during the preparation of the rendering context for the details tab.

Root cause: `key` in `for..of` statement actually was an index, not a real `key`. 

If `for..of` is preferable, you can ignore this PR and use 
```js
for (const entry of context.skills.concat(context.tools)) { 
// the same code as in PR
}
```

### DIFF
--- a/module/applications/actor/character-sheet.mjs
+++ b/module/applications/actor/character-sheet.mjs
@@ -361,12 +361,12 @@ export default class CharacterActorSheet extends BaseActorSheet {
     // Skills & Tools
     context.skills = this._prepareSkillsTools(context, "skills");
     context.tools = this._prepareSkillsTools(context, "tools");
-    context.skills.concat(context.tools).forEach(entry => {
+    for ( const entry of context.skills.concat(context.tools) ) {
       const key = entry.key;
       entry.class = this.constructor.PROFICIENCY_CLASSES[context.editable ? entry.baseValue : entry.value];
       if ( key in CONFIG.DND5E.skills ) entry.reference = CONFIG.DND5E.skills[key].reference;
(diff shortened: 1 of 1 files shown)


module/applications/actor/character-sheet.mjs
```diff
@@ -361,11 +361,12 @@ export default class CharacterActorSheet extends BaseActorSheet {
     // Skills & Tools
     context.skills = this._prepareSkillsTools(context, "skills");
     context.tools = this._prepareSkillsTools(context, "tools");
-    for ( const [key, entry] of Object.entries(context.skills).concat(Object.entries(context.tools)) ) {
+    context.skills.concat(context.tools).forEach(entry => {
```
arbron:
```suggestion
    for ( const entry of context.skills.concat(context.tools) ) {
```

module/applications/actor/character-sheet.mjs
```diff
@@ -361,11 +361,12 @@ export default class CharacterActorSheet extends BaseActorSheet {
...
+      const key = entry.key;
       entry.class = this.constructor.PROFICIENCY_CLASSES[context.editable ? entry.baseValue : entry.value];
       if ( key in CONFIG.DND5E.skills ) entry.reference = CONFIG.DND5E.skills[key].reference;
       else if ( key in CONFIG.DND5E.tools ) entry.reference = Trait.getBaseItemUUID(CONFIG.DND5E.tools[key].id);
-    }
+    })
```
arbron:
```suggestion
    }
```

Issue #5559, opened 2025-04-29
[Feature Request] Aura Effects


kueppe:
With the 2024 rules integration of so many emanation and aura effects (paladin boost saves, spirit guardians etc.) I can't even imagine playing without the active aura module. I think such a functionality should find its way in the d&d core system.

arbron:
Should be possible with token-attached regions in V14 (see https://github.com/foundryvtt/foundryvtt/issues/13586).

PR #6761, opened 2026-02-24
[#5559, #6722] Add Activity Behaviors that attach to templates

### LINKED ISSUE #6722: Add method for attaching behaviors to measured template regions created from activities
The system needs a method for defining behaviors that are automatically attached to area of effect regions that are created by activities. There should be a configuration that allows for defining what behaviors are available and establishing what is configurable on those behaviors.

So for example, a spell like "Entangle" can include a "Difficult Terrain" region behavior that has the "Plants" type defined. The user configuring the spell would have the option of setting the difficult terrain types, but other parts of the behavior would be inferred such as whether it is considered magical or not.

Because players cannot create regions themselves, the actual region creation will need to be handled by the active GM whenever the region placement is completed by the player. When the template is created then a method would be called for each behavior to be created that would allow it to define the region behavior creation data.

arbron:
Adds a new system for registering Activity Behaviors. These will be automatically attached to any measured templates created from that activity. An Activity Behavior represents a simplified data model for configuring a Region Behavior, allowing for certain parts of the Region Behavior's configuration to be automatically populated based on activity settings (for example, the Difficult Terrain behavior gets its "Magical" property set automatically based on whether the item has the magical property set).

Activity Template Behaviors

Since players cannot create region behaviors themselves, the creation of these behaviors is handled by the active GM user after the player places the template in the scene.

Closes #6722


lang/en.json
```diff
@@ -283,6 +295,24 @@
...
+          },
+          "min": {
+            "label": "Minimum Level"
+          }
+        },
+        "name": {
+          "label": "Name"
+        }
+      },
+      "label": "Applied Behaviors"
```
Fyorl: I think 'behavior' is a bit of an opaque term. We probably have to use it still because what we are adding *are* Region Behaviours. But I wonder if we could call this fieldset something like 'Area Effects', to make it clear that it is related to the 'Area' fieldset in the targeting tab?

module/applications/activity/activity-sheet.mjs
```diff
@@ -266,6 +286,26 @@ export default class ActivitySheet extends PseudoDocumentSheet {
...
+            data, source,
+            additionalFields: instance.generateFields(source.config, { prefix: `behaviors.${data._index}.config.` }),
+            additionalSettings: "systems/dnd5e/templates/activity/parts/activity-behavior-settings.hbs",
+            collapsed: this.expandedSections.get(`behavior.${data._id}`) ? "" : "collapsed",
+            config: CONFIG.DND5E.activityBehaviorTypes[data.type],
+            fields: this.activity.schema.fields.behaviors.element.fields,
+            prefix: `behaviors.${data._index}.`
+          };
+          return this._prepareAppliedBehaviorContext(context, ctx);
+        });
```
Fyorl: Drop nulls

```suggestion
        })
        .filter(_ => _);
```

module/applications/activity/activity-sheet.mjs
```diff
@@ -266,6 +286,26 @@ export default class ActivitySheet extends PseudoDocumentSheet {
...
+        .filter(d => d.type in CONFIG.DND5E.activityBehaviorTypes)
+        .map(data => {
+          const source = context.source.behaviors[data._index];
+          if ( !source ) return null;
+          const instance = new CONFIG.DND5E.activityBehaviorTypes[source.type].model(source.config);
+          const ctx = {
+            data, source,
+            additionalFields: instance.generateFields(source.config, { prefix: `behaviors.${data._index}.config.` }),
+            additionalSettings: "systems/dnd5e/templates/activity/parts/activity-behavior-settings.hbs",
+            collapsed: this.expandedSections.get(`behavior.${data._id}`) ? "" : "collapsed",
```
Fyorl:
```suggestion
            collapsed: this.expandedSections.get(`behaviors.${data._id}`) ? "" : "collapsed",
```

module/data/activity/fields/applied-behavior-field.mjs
```diff
@@ -0,0 +1,22 @@
+const { DocumentIdField, NumberField, ObjectField, SchemaField, StringField } = foundry.data.fields;
+
+/**
+ * Field for storing an region behaviors applied by an activity to its template.
+ */
+export default class AppliedBehaviorField extends SchemaField {
+  constructor(fields={}, options={}) {
+    fields = {
+      _id: new DocumentIdField({ initial: () => foundry.utils.randomID() }),
+      config: new ObjectField(),
```
Fyorl: Did we consider the `AdvancementDataField` pattern? Maybe to verbose/over-engineered? It seems perhaps more idiomatic of our codebase though.

module/data/activity/_types.mjs
```diff
@@ -224,3 +247,42 @@
...
+ * @property {string} _id  ID of the effect to apply.
+ * @property {object} level
+ * @property {number} level.min  Minimum level at which this effect can be applied.
+ * @property {number} level.max  Maximum level at which this effect can be applied.
+ */
+
+/* -------------------------------------------- */
+
+/**
+ * @typedef ConsumptionTargetDataData
```
```suggestion
 * @typedef ConsumptionTargetData
```

module/data/activity/fields/applied-behavior-field.mjs
```diff
@@ -0,0 +1,22 @@
+const { DocumentIdField, NumberField, ObjectField, SchemaField, StringField } = foundry.data.fields;
+
+/**
+ * Field for storing an region behaviors applied by an activity to its template.
```
Fyorl:
```suggestion
 * Field for storing region behaviors applied by an activity to its template.
```

templates/activity/parts/activity-behaviors.hbs
```diff
@@ -0,0 +1,48 @@
+{{#if appliedBehaviors}}
```
Fyorl: `context.appliedBehaviors` is an `Iterator`.
```suggestion
{{#if activity.behaviors}}
```

templates/activity/parts/activity-behaviors.hbs
```diff
@@ -0,0 +1,48 @@
...
+        <span>{{ localize "DND5E.ACTIVITY.FIELDS.behaviors.label" }}</span>
+        <button type="button" class="unbutton control-button" data-action="addBehavior"
+                data-tooltip aria-label="{{ localize 'DND5E.BEHAVIOR.Action.Create' }}">
+            <i class="fa-solid fa-plus" inert></i>
+        </button>
+    </legend>
+    <p class="hint">{{ localize "DND5E.BEHAVIOR.Hint" }}</p>
+    <ul class="separated-list behaviors-list flexcol unlist">
+        {{#each appliedBehaviors}}
+        <li data-index="{{ data.index }}" data-behavior-id="{{ data._id }}" data-expand-id="behaviors.{{ data._id }}">
```
Fyorl: Don't think `data.index` is defined.
```suggestion
        <li data-index="{{ @index }}" data-behavior-id="{{ data._id }}" data-expand-id="behaviors.{{ data._id }}">
```

Fyorl (changes requested):

lang/en.json
arbron: "Area of Effect Behaviors"?

module/data/activity/fields/applied-behavior-field.mjs
arbron: I've taken a look at this and I'm not sure how feasible it is. We can't follow `AdvancementDataField`'s model directly because that relies on knowing the type statically when the data field is created which we can only rely on for advancement because we initialize each advancement with a different class before it creates its schema.

I thought about doing it like activities, where each entry in the `behaviors` array is a different type, but that sacrifices the ability for different activity types to define new properties on `AppliedBehaviorField` (similar to how the `SaveActivity` defines a new property on `AppliedEffectField`).

I took a look at how core's `TypeDataField` does it for system data, but I had trouble getting the `type` data from the parent into the child during initialization, cleaning, and migration as necessary. If you have any thoughts on how to reliable communicate the type then we could go that way.

lang/en.json
Fyorl: Sure, I guess that works, it is a 'behavior' of the area, in the natural language sense of the word.

module/data/activity/fields/applied-behavior-field.mjs
Fyorl: I see how my initial suggestion was a bit naïve. It would be pretty straightforward to handle these the same way as `ActivitiesField`, if we're happy to sacrifice polymorphism over the parent activity type as you pointed out.

If that's required though, then I had a go at your paragraph 3, and I think it is possible now that we have robust `_state` parameter.

```js
export default class AppliedBehaviorField extends SchemaField {
    fields = {
      /* ... */
      config: new BehaviorConfigField()
    };
    /* [snip] */
  }

  _cleanType(data, options, _state={}) {
    const type = data?.type ?? _state.source?.type;
    return super._cleanType(data, options, type ? { ..._state, behaviorType: type } : _state);
  }

  initialize(value, model, options={}) {
    return super.initialize(value, model, value?.type ? { ...options, behaviorType: value.type } : options);
  }
}

export class BehaviorConfigField extends foundry.data.fields.ObjectField {
  static recursive = true;

  getModel(type) {
    return CONFIG.DND5E.activityBehaviorTypes[type]?.model ?? null;
  }

  _cleanType(value, options, _state) {
    const cls = this.getModel(_state?.behaviorType);
    if ( cls ) return cls.cleanData(typeof value === "object" ? value : {}, options, _state);
    return value;
  }

  initialize(value, model, options={}) {
    const cls = this.getModel(options.behaviorType);
    if ( cls ) return new cls(value, { parent: model, ...options });
    return foundry.utils.deepClone(value);
  }

  _migrate(value, options, _state) {
    // _state here is from _cleanType since migration is done as part of cleaning now
    const cls = this.getModel(_state?.behaviorType);
    if ( cls ) cls.migrateDataSafe(value);
    return value;
  }
}
```

Then something like `BaseSaveActivityData` could have `behaviors: new ArrayField(new AppliedBehaviorField({ onSave: xyz }))`, and now we have polymorphism over both the activity and the behaviour type, just that the activity-specific data is sibling to `config` rather than inside it. The places we currently wrap in behaviour type model become a lot neater:

```diff
- const model = new CONFIG.DND5E.activityBehaviorTypes[behavior.type].model(behavior.config);
- const data = model.createBehaviorData(activity);
+ const data = behavior.config.createBehaviorData(activity);
```

```diff
- const instance = new CONFIG.DND5E.activityBehaviorTypes[source.type].model(source.config);
-   additionalFields: instance.generateFields(source.config, { ... }),
+   additionalFields: data.config.generateFields(source.config, { ... }),
```

module/data/region-behavior/base-activity-behavior.mjs
```diff
@@ -0,0 +1,106 @@
...
+   * @param {object} [options={}]
+   * @param {string} [options.prefix=""]  Prefix added before each field's name.
+   * @yields object
+   */
+  *generateFields(source, { prefix="", ...options }={}) {
+    for ( const field of Object.values(options[SCHEMA] ?? this.constructor.schema.fields) ) {
+      const path = `${options[PATH] ?? ""}${field.name}`;
+      if ( field.constructor.hasFormSupport ) {
+        const data = { field, name: `${prefix}${path}`, value: foundry.utils.getProperty(source, path) };
+        if ( field instanceof foundry.data.fields.BooleanField ) data.input = createCheckboxInput;
```
Fyorl, line 101: Need to import `createCheckboxInput`.

Fyorl (changes requested):

module/data/activity/fields/applied-behavior-field.mjs
Fyorl: Just for clarity: If we were to model them like `ActivitiesField`, they would look like this?

```js
export default class BaseActivityBehavior extends foundry.abstract.DataModel {
  static defineSchema() {
    // _id, type, name, level
  }
}

export class DifficultTerrainActivityBehavior extends BaseActivityBehavior {
  static defineSchema() { return { ...super.defineSchema(), types: /* ... */ }; }
}
```

I think that's probably the cleanest solution, if we're happy to make that sacrifice. But if it's required then the previous comment is a way we can implement it without making any sacrifices.

arbron: I've gone with changing just the `config` data model for now. I've refactored out a bunch of code shared by `ActivityField`, `AdvancementField`, and `AdvancementDataField` into a new `TypeDataField5e` so that we can make use of that here so the actual code added is very small.

module/data/activity/fields/applied-behavior-field.mjs
```diff
@@ -0,0 +1,41 @@
+import TypeDataField5e from "../../fields/type-data-field.mjs";
+
+const { DocumentIdField, NumberField, ObjectField, SchemaField, StringField } = foundry.data.fields;
```
Fyorl:
```suggestion
const { DocumentIdField, NumberField, SchemaField, StringField } = foundry.data.fields;
```

Fyorl (approved):
`TypeDataField5e` is nice consolidation, this looks good

PR #6762, opened 2026-02-25
[#6708] Add `makeChange` option to `getActorCurrencyUpdates`

### LINKED ISSUE #6708: Add `makeChange` option to `getActorCurrencyUpdates`
Right now `CurrencyManager#getActorCurrencyUpdates` only works if you have enough currency of some denomination to fulfill the request, but it isn't capable of making change in order to create the necessary denominations if required. For example, if you have a character with 100 gold and no copper and ask to deduct 20 cp then it will not work. Ideally the system would be able to see that the player doesn't have enough copper to complete the transaction and would be able to convert one of the gold pieces into 100 cp before doing the deduction.

Sayshal:
- Break higher-denomination coins when the target denomination is unavailable
- Respects existing `priority` parameter for coin selection order

Closes #6708

---
```js
// Test 1: GP-only actor, deduct CP**
// Actor: { pp: 0, gp: 100, ep: 0, sp: 0, cp: 0 }

const result = dnd5e.applications.CurrencyManager.getActorCurrencyUpdates(actor, 20, "cp", { makeChange: true });
// Result: remainder=0, gp=99, cp=80

// Test 2: PP-only actor, deduct SP
// Actor: { pp: 5, gp: 0, ep: 0, sp: 0, cp: 0 }

const result = dnd5e.applications.CurrencyManager.getActorCurrencyUpdates(actor, 30, "sp", { makeChange: true });
// Result: remainder=0, pp=4, sp=70

// Test 3: Without makeChange (unchanged behavior)
// Actor: { pp: 0, gp: 100, ep: 0, sp: 0, cp: 0 }

const result = dnd5e.applications.CurrencyManager.getActorCurrencyUpdates(actor, 20, "cp");
// Result: remainder=20, no coins changed
```


module/applications/currency-manager.mjs
```diff
@@ -269,9 +269,11 @@ export default class CurrencyManager extends Application5e {
    * @param {boolean} [options.recursive=false]      Deduct currency from containers as well as the base Actor. TODO
    * @param {"high"|"low"} [options.priority="low"]  Prioritize higher denominations before lower, or vice-versa.
    * @param {boolean} [options.exact=true]           Prioritize deducting the requested denomination first.
+   * @param {boolean} [options.makeChange=false]     Convert higher denominations to fulfill the request if needed.
```
arbron: This should also be added to `deductActorCurrency`, but rather than duplicate all of these options in both places we should probably convert this to a `@typedef`:

This can go into `module/applications/_types.mjs` below `AwardOptions`:

```javascript
  /**
   * @typedef CurrencyUpdateOptions
   * @property {boolean} [exact=true]           Prioritize deducting the requested denomination first.
   * @property {boolean} [makeChange=true]      Convert higher denominations to fulfill the request if needed.
   * @property {"high"|"low"} [priority="low"]  Prioritize higher denominations before lower, or vice-versa.
   * @property {boolean} [recursive=false]      Deduct currency from containers as well as the base Actor. TODO
   */

```

Then at the top of `currency-manager.mjs` below the module imports we can add:

```javascript
/**
 * @import { CurrencyUpdateOptions } from "./_types.mjs";
 */
```

And then for `deductActorCurrency` and `getActorCurrencyUpdates` the type of options can be replaced  with `CurrencyUpdateOptions` and all of the breakdown of options after that can be removed.

module/applications/currency-manager.mjs
```diff
@@ -269,9 +269,11 @@ export default class CurrencyManager extends Application5e {
...
    * @returns {{ item: object[], remainder: number, [p: string]: any }}
    */
-  static getActorCurrencyUpdates(actor, amount, denomination, { recursive=false, priority="low", exact=true }={}) {
+  static getActorCurrencyUpdates(actor, amount, denomination, { recursive=false, priority="low", exact=true,
+    makeChange=false }={}) {
```
arbron: I imaging most situations would want to make change, so I'd say make that `true` by default:

```suggestion
  static getActorCurrencyUpdates(
    actor, amount, denomination, { recursive=false, priority="low", exact=true, makeChange=true }={}
  ) {
```

Fyorl: Makes this a breaking change though

module/applications/_types.mjs
Fyorl, line 1: The diff is showing up as a big chunk of changes. Did the line endings get changed or something?

module/applications/currency-manager.mjs
```diff
@@ -288,9 +288,17 @@ export default class CurrencyManager extends Application5e {
...
+        if ( !deduct && makeChange && (conversion < baseConversion) && updates.system.currency[denom] ) {
+          updates.system.currency[denom] -= 1;
+          updates.system.currency[denomination] += Math.floor(baseConversion / conversion);
+          deduct = Math.min(updates.system.currency[denomination], updates.remainder);
+          updates.remainder -= deduct;
+          updates.system.currency[denomination] -= deduct;
+        } else {
+          updates.remainder -= deduct / multiplier;
+          updates.system.currency[denom] -= deduct;
+        }
```
Fyorl: I think this will only ever break 1 higher-denomination coin per call to `getActorCurrencyUpdates` because we only make change if `!deduct`, and then never revisit that denomination again in subsequent passes. For example, if the actor has 2 gp and you try to deduct 150 cp, you expect to get `{ gp: 0, cp: 50, remainder: 0 }`, but you actually get `{ gp: 1, remainder: 50 }`. If we handle multiple breaks in one pass, I think that resolves it, e.g.:

```suggestion
        const deduct = Math.min(updates.system.currency[denom], Math.floor(updates.remainder * multiplier));
        // Handle normal deduction first.
        updates.remainder -= deduct / multiplier;
        updates.system.currency[denom] -= deduct;
        // If there's still a remainder, break the denomination into change.
        if ( updates.remainder && makeChange && (conversion < baseConversion) && updates.system.currency[denom] ) {
          const rate = Math.floor(baseConversion / conversion);
          const breaks = Math.min(updates.system.currency[denom], Math.ceil(updates.remainder / rate));
          updates.system.currency[denom] -= breaks;
          updates.system.currency[denomination] += breaks * rate;
          const change = Math.min(updates.system.currency[denomination], updates.remainder);
          updates.remainder -= change;
          updates.system.currency[denomination] -= change;
        }
```

module/applications/_types.mjs
Sayshal: Yep my editor did some silliness, fixed now.

Fyorl (approved):
Don't feel strongly about whether this should be breaking or not. Means we don't need any code change for bastion deduction at least.

PR #6769, opened 2026-02-25
[#4550] Prevent linebreak in damage enricher parentheses

### LINKED ISSUE #4550: Prevent linebreaks in parentheses in damage enrichers when using `average`
Currently, the result can look like
```md
[...............] This deals 5 (
<roll-icon> 2d4) fire damage [...]
```
It would be good if it was enforced to be eg
```md
[...............] This deals 5
(<roll-icon> 2d4) fire damage [...]
```

Sayshal:
- Wrap parenthesized formula in nowrap span when using average format
- Move parens from localization string into JS-generated HTML
- Add .formula-group CSS class to prevent line breaks within (formula)

Closes #4550

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -869,9 +869,8 @@ export async function enrichDamage(configs, label, options) {
     const types = type?.split("|")
       .map(t => CONFIG.DND5E.damageTypes[t]?.label ?? CONFIG.DND5E.healingTypes[t]?.label)
       .filter(_ => _);
-    const rollLink = createRollLink(formula, {}, { tag: "span" }).outerHTML;
     const localizationData = {
-      formula: rollLink,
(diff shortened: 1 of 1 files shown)


lang/en.json
```diff
@@ -5621,7 +5621,7 @@
   "DamageExtended": "<em>Hit:</em> {damage} damage",
   "DamageDouble": "{first} plus {second}",
   "DamageShort": "{formula} {type}",
-  "DamageLong": "{average} ({formula}) {type}",
+  "DamageLong": "{average} {formula} {type}",
```
arbron: I'd probably prefer to add the span to the localization formulation rather than moving the parenthesis out so that languages are able to localize them:


```suggestion
  "DamageLong": "{average} <span class=\"formula-group\">({formula})</span> {type}",
```

PR #6773, opened 2026-02-26
[#5755] Add description to legacy Monk Unarmed Strike

### LINKED ISSUE #5755: [Feature]: Consider adding a tooltip for the Monk version of Unarmed Strike
There is no description provided for the 2024 PHB Unarmed Strike for Monk which seems odd. Perhaps consider adding some text,  a link to the Martial Arts feature, and/or a link to the Unarmed Strike rules like the non-monk versions of Unarmed strike do.

Reported while using D&D 5E 5.0.3 Foundry VTT 13.345.

Sayshal:
- Add tooltip/description linking to Martial Arts feature
- Matches pattern used by 2024 Monk Unarmed Strike
- Add descriptive text and Martial Arts reference link to both legacy and 2024 Monk Unarmed Strike

Closes #5755

### DIFF
--- a/packs/_source/classes24/monk/class-features/unarmed-strike.yml
+++ b/packs/_source/classes24/monk/class-features/unarmed-strike.yml
@@ -124,10 +124,9 @@ system:
     max: ''
   description:
     value: >-
-      <p>A punch, kick, head-butt, or similar forceful blow (none of which
-      count as weapons). You can use Dexterity instead of Strength for the
-      attack and damage rolls of your unarmed strikes. You can roll your
-      Martial Arts die in place of the normal damage.</p>
+      <p>You can use Dexterity instead of Strength for the attack and damage
+      rolls of your &amp;Reference[Unarmed Strike]{unarmed strikes}. You can
+      roll your Martial Arts die in place of the normal damage.</p>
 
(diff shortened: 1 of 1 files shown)


packs/_source/classes24/monk/class-features/unarmed-strike.yml
```diff
@@ -124,6 +124,11 @@ system:
     max: ''
   description:
     value: >-
+      <p>A punch, kick, head-butt, or similar forceful blow (none of which
+      count as weapons). You can use Dexterity instead of Strength for the
+      attack and damage rolls of your unarmed strikes. You can roll your
+      Martial Arts die in place of the normal damage.</p>
```
Fyorl: This '(none of which count as weapons)' part, while still correct, is legacy wording which is absent from the 2024 rules glossary on Unarmed Strike. Since there is now a rules glossary in the SRD 5.2, I think we would prefer to link directly to that entry. Maybe something like this:

> You can use Dexterity instead of Strength for the attack and damage rolls of your &Reference[Unarmed Strike]{unarmed strikes}. You can roll your Martial Arts die in place of the normal damage.

Sayshal:
Ready for review.

Issue #5444, opened 2025-03-30
CB: Provide option to entirely lock down the filters


krbz999:
Use case:

Prompt a user to pick a huge dragon, but I don't want them to be able to adjust the filters to toggle "beast" for example.

Currently one has to iterate through all the options and set them to `-1`. It would be nice to be able to just blanket-lock them all.

```js
await dnd5e.applications.CompendiumBrowser.select({
    tab: "actors",
    mode: dnd5e.applications.CompendiumBrowser.MODES.ADVANCED,
    selection: { min: 1, max: 3 },
    filters: {
        locked: {
            types: new Set(["npc"]),
            additional: {
                cr: { min: 1, max: 30 },
                movement: { walk: 1 },
                size: { huge: 1 },
                type: { dragon: 1 },
            },
        },
    },
});
```

Sayshal:
I've attempted an implementation of this request in #6776.

PR #6776, opened 2026-02-26
[#5444] Add `locked.exclusive` option to `CompendiumBrowser` filters

Sayshal:
## Summary

Adds a `locked.exclusive` option to `CompendiumBrowserFilters` that locks all choices within filter categories that have explicit values set, preventing additional selections. Categories without values remain unlocked.

Right now, locking filters requires specifying each one individually in `locked.additional`. This adds an `exclusive: true` flag that automatically locks down any category where a value is explicitly provided — preventing users from adding invalid combinations — while leaving unset categories fully interactive.

## Usage

```js
new dnd5e.applications.CompendiumBrowser({
  filters: {
    locked: {
      exclusive: true,
      documentClass: "Actor",
      types: new Set(["npc"]),
      additional: {
        size: { huge: 1 },
        "type.value": { dragon: 1 }
      }
    }
  }
});
```

Setting `locked.exclusive: true` locks Size and Creature Type (which have explicit values) while leaving Habitat, Source, and other filters open for the user to refine.

## Implementation

- `_prepareSidebarContext`: Reads `lockExclusive` from locked filter options
- `generateLocked`: Accepts the filter definition as a second parameter. When `exclusive` is set and the filter has a value, produces the correct locked shape per type (`{ min, max }` for range, all keys for set, `true` for boolean)
- `_renderSourceFilters`: Locks every available source key when `exclusive` is set and source has a locked value
- `currentFilters` getter: Strips `exclusive` from the merged filters so it doesn't leak into actual filter logic

### DIFF
--- a/module/applications/_types.mjs
+++ b/module/applications/_types.mjs
@@ -39,7 +39,7 @@
 
 /**
  * @typedef CompendiumBrowserFilters
- * @property {boolean} [exclusive]    When used in locked filters, locks all choices within filter categories that have
+ * @property {boolean} [exclusive]     When used in locked filters, locks all choices within filter categories that have
  *                                     explicit values, preventing additional selections. Categories without values
  *                                     remain unlocked.
  * @property {string} [documentClass]  Document type to fetch (e.g. Actor or Item).
--- a/module/applications/compendium-browser.mjs
+++ b/module/applications/compendium-browser.mjs
@@ -715,7 +715,7 @@ export default class CompendiumBrowser extends Application5e {
     if ( !sources.length ) return;
     const lockExclusive = this.options.filters?.locked?.exclusive === true;
     const lockedSource = this.options.filters?.locked?.additional?.source;
-    const locked = (lockExclusive && lockedSource !== undefined)
+    const locked = lockExclusive && (lockedSource !== undefined)
       ? Object.fromEntries(Object.keys(this.#sources).map(k => [k, true]))
       : Object.entries(lockedSource ?? {}).reduce((obj, [k, v]) => {
         obj[k.slugify({ strict: true })] = v;
@@ -725,7 +725,7 @@ export default class CompendiumBrowser extends Application5e {
       "systems/dnd5e/templates/compendium/browser-sidebar-filter-set.hbs",
       {
         locked,
-        value: (lockExclusive && lockedSource !== undefined) ? {} : locked,
+        value: lockExclusive && (lockedSource !== undefined) ? {} : locked,
         key: "source",
         expandId: "source",
         label: "DND5E.SOURCE.FIELDS.source.label",


Sayshal:
> I think we need to adjust the behavior of this a bit. Based on the original request and my general feeling of what would be useful this should only lock filters where something is explicitly set.
> 
> So in Zhell's original issue, the filters are set to Size: Huge, Creature Type: Dragon, Movement: Walk. It would make sense to lock down all other options in those three categories to prevent adding say Beast to creature type to get additional invalid options, but you don't necessary want to lock down habitat in case you only want to see Desert dragons or source in case you only want to see dragons from Fizban's.
> 
> This new option would probably need a different name, something like `locked.exclusive` or something.
> 
> We could potentially keep the `locked.all` option, but it seems like the less useful of the two.

Makes sense to me, I've made the necessary changes to more align with the original issues concerns.

module/applications/_types.mjs
```diff
@@ -39,6 +39,9 @@
 
 /**
  * @typedef CompendiumBrowserFilters
+ * @property {boolean} [exclusive]    When used in locked filters, locks all choices within filter categories that have
```
arbron:
```suggestion
 * @property {boolean} [exclusive]     When used in locked filters, locks all choices within filter categories that have
```

module/applications/compendium-browser.mjs
```diff
@@ -704,15 +713,19 @@ export default class CompendiumBrowser extends Application5e {
     const filters = this.element.querySelector('[data-application-part="filters"]');
     filters.querySelector('[data-filter-id="source"]')?.remove();
     if ( !sources.length ) return;
-    const locked = Object.entries(this.options.filters?.locked?.additional?.source ?? {}).reduce((obj, [k, v]) => {
-      obj[k.slugify({ strict: true })] = v;
-      return obj;
-    }, {});
+    const lockExclusive = this.options.filters?.locked?.exclusive === true;
+    const lockedSource = this.options.filters?.locked?.additional?.source;
+    const locked = (lockExclusive && lockedSource !== undefined)
```
arbron:
```suggestion
    const locked = lockExclusive && (lockedSource !== undefined)
```

module/applications/compendium-browser.mjs
```diff
@@ -704,15 +713,19 @@ export default class CompendiumBrowser extends Application5e {
...
+      : Object.entries(lockedSource ?? {}).reduce((obj, [k, v]) => {
+        obj[k.slugify({ strict: true })] = v;
+        return obj;
+      }, {});
     const filter = await foundry.applications.handlebars.renderTemplate(
       "systems/dnd5e/templates/compendium/browser-sidebar-filter-set.hbs",
       {
         locked,
-        value: locked,
+        value: (lockExclusive && lockedSource !== undefined) ? {} : locked,
```
arbron:
```suggestion
        value: lockExclusive && (lockedSource !== undefined) ? {} : locked,
```

PR #6778, opened 2026-02-26
[#6763] Implement `TemplatePlacement` to replace `AbilityTemplate`

### LINKED ISSUE #6763: Re-write `AbilityTemplate` as new `TemplatePlacement` mirroring `TokenPlacement`
Currently activity templates are placed using a subclass of `MeasuredTemplate` with a number of methods used to handle placing the template on the scene. With the deprecation of `MeasuredTemplate` in V14 in favor of regions this will need to be reworked to place regions directly. This will also be necessary to support features like the emanation shape or token-attached regions that aren't supported through the shims on `MeasuredTemplate`.

It would be good to refactor this to use an API similar to `TokenPlacement`, possibly seeing if parts like throttling and mouse events can be shared between the two implementations.

arbron:
Adds a new `dnd5e.canvas.TemplatePlacement` API to handle placing shapes within a scene. This base API deals with basic shape data but the `fromActivity` factory method behaves similar to the old method on `AbilityTemplate` to create a full measured template region.

The created regions are set up to work as similar as possible to the old measured templates, using `coverage` highlight mode and the user's color. When the area of effect count is larger than one, all of the shapes are combined into a single measured template.

`TokenPlacement` has been refactored and much of its code has been shifted into a new `BasePlacement` class that is shared between the two canvas placement systems.

Closes #6763


dev7355608 (commented):
What is the reason that we don't use the new `RegionLayer#placeRegion` API?

arbron:
> What is the reason that we don't use the new `RegionLayer#placeRegion` API?

We need to support placing more than one shape at a time would be the big one. I also have plans for adding additional UI that is shared with our token placement system.

dev7355608:
Could you write up a feature request for the limitations you currently facing with `RegionLayer#placeRegion`. It would be undesirable for us if dnd5e and other systems couldn't use `RegionLayer#placeRegion` and needed to reimplement most of the placement workflow.

arbron:
> Could you write up a feature request for the limitations you currently facing with `RegionLayer#placeRegion`. It would be undesirable for us if dnd5e and other systems couldn't use `RegionLayer#placeRegion` and needed to reimplement most of the placement workflow.

These cover much of it:

- https://github.com/foundryvtt/foundryvtt/issues/13892
- https://github.com/foundryvtt/foundryvtt/issues/13900
- https://github.com/foundryvtt/foundryvtt/issues/13901

But I will say our implementation also shares a bunch of the internals with our `TokenPlacement` API which currently has no equivalent in core, so sharing those internals rather than having one base based on `placeRegion` probably makes more sense for us.

There are also other features for placement restriction/guidance that might not make sense in core:
- Range limits/display from caster
- Line of sight placement restriction
- Limit placement to grid intersections
- Locking template origins to the casting token (e.g. lines start at edge of caster's token and should only go out)
- UI displaying number of templates or tokens placed and number still to go

krbz999:
At least one of those are mentioned in https://github.com/foundryvtt/foundryvtt/issues/9840

dev7355608:
> There are also other features for placement restriction/guidance that might not make sense in core:
> 
> * Range limits/display from caster
> * Line of sight placement restriction
> * Limit placement to grid intersections
> * Locking template origins to the casting token (e.g. lines start at edge of caster's token and should only go out)
> * UI displaying number of templates or tokens placed and number still to go

The options `preConfirm`, `onMove`, and `onRotate` of `RegionLayer#placeRegion` should make this possible to implement such behaviors.

module/canvas/api/base-placement.mjs
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Perform the placement, asking player guidance when necessary.
+   * @param {PlacementConfiguration} config
+   * @returns {Promise<PlacementData[]>}
+   */
+  static place(config) {
+    const placement = new this(config);
+    return placement.place();
+  }
```
Fyorl, line 41: missing separators

module/canvas/template-placement.mjs
```diff
@@ -0,0 +1,169 @@
+import { convertLength } from "../utils.mjs";
+import BasePlacement from "./api/base-placement.mjs";
+
+/**
+ * @import {
+ *   TemplatePlacementConfiguration, TemplatePlacementData, TemplatePlacementShapeConfiguration
+ * } from "./types.mjs";
```
Fyorl:
```suggestion
 * } from "./_types.mjs";
```

module/canvas/template-placement.mjs
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * A hook event that fires before player is prompted for template placement.
+     * @function dnd5e.preCreateMeasuredTemplate
+     * @memberof hookEvents
+     * @param {Activity} activity                               Activity for which the template is being placed.
+     * @param {TemplatePlacementConfiguration} placementConfig  Configuration that drives template placement.
+     * @returns {boolean}                                       Return `false` to prevent template placement.
+     */
+    if ( Hooks.call("dnd5e.preCreateMeasuredTemplate", activity, config) === false ) return null;
```
Fyorl, line 109: `createActivityTemplate` was the hook we were using before. Did we mean to change it?

module/canvas/template-placement.mjs
```diff
@@ -0,0 +1,169 @@
...
+     * A hook event that fires after templates have been placed by the player but before they have been created.
+     * @function dnd5e.createMeasuredTemplate
+     * @memberof hookEvents
+     * @param {Activity} activity      Activity for which the template is being placed.
+     * @param {object[]} templateData  Data for the regions to be created.
+     * @returns {boolean}              Return `false` to prevent template creation.
+     */
+    if ( Hooks.call("dnd5e.createMeasuredTemplate", activity, regionData) === false ) return null;
+
+    const created = canvas.scene.createEmbeddedDocuments("Region", regionData);
```
Fyorl:
```suggestion
    const created = await canvas.scene.createEmbeddedDocuments("Region", regionData);
```

Fyorl (changes requested):
Good simplification. Seem to have lost the handling for `gridAlignedSquareTemplates`.

module/canvas/template-placement.mjs
arbron: I changed it because the `config` data here is very different than the `templateData` data we used in the original hook. Same with the post hooks which have region data as their second argument rather than measured templates.

module/canvas/token-placement.mjs
```diff
@@ -1,253 +1,36 @@
+import BasePlacement from "./api/base-placement.mjs";
+
 /**
  * @import { TokenPlacementConfiguration, TokenPlacementData } from "./types.mjs";
```
Fyorl:
```suggestion
 * @import { TokenPlacementConfiguration, TokenPlacementData } from "./_types.mjs";
```

Fyorl (approved):

PR #6783, opened 2026-02-27
[#5559, #6436] Create proper emanations, attach to tokens

### LINKED ISSUE #6436: Switch to using emanation shape rather than circle shape for emanations
V14 adds a new emanation shape that can be used for measured templates. The system should switch to using that when the emanation area of effect type is selected.

Blocked by #6763

arbron:
Swaps the Emanation area of effect type to use `emanation` shapes under the hood, adapting the behavior for when tokens are moused over during placement to associate a token with the emanation. This leads to the base size of the emanation being adjusted to match the size of the token.

When an emanation is not marked as stationary and placed over a token it will be automatically attached to the token it is placed over, creating an aura that moves with that token.

https://github.com/user-attachments/assets/98d3c3a0-0d2b-4285-8b24-1f3b185c5147

Closes #6436

### DIFF
--- a/module/canvas/template-placement.mjs
+++ b/module/canvas/template-placement.mjs
@@ -58,7 +58,10 @@ export default class TemplatePlacement extends BasePlacement {
       case "circle": return { ...data, radius: size };
       case "cone": return { ...data, angle: CONFIG.MeasuredTemplate.defaults.angle, radius: size };
       case "emanation": return {
-        base: { ...data, width: 1, height: 1, shape: 4, type: "token" },
+        base: {
+          ...data, width: 1, height: 1, type: "token",
+          shape: canvas.grid.isHexagonal ? CONST.TOKEN_SHAPES.ELLIPSE_1 : CONST.TOKEN_SHAPES.RECTANGLE_1
+        },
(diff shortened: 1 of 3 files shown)


module/canvas/template-placement.mjs
```diff
@@ -54,7 +54,10 @@ export default class TemplatePlacement extends BasePlacement {
     switch ( type ) {
       case "circle": return { ...data, radius: size };
       case "cone": return { ...data, angle: CONFIG.MeasuredTemplate.defaults.angle, radius: size };
-      case "emanation": return { base: { ...data }, radius: size }; // TODO: Make this work properly
+      case "emanation": return {
+        base: { ...data, width: 1, height: 1, shape: 4, type: "token" },
```
Fyorl: Nicer preview on hex grids:


```suggestion
        base: { ...data, width: 1, height: 1, shape: canvas.grid.isHexagonal ? CONST.TOKEN_SHAPES.ELLIPSE_1 : CONST.TOKEN_SHAPES.RECTANGLE_1, type: "token" },
```

Fyorl (approved):

PR #6784, opened 2026-02-27
[#6438] Add ring wall type to area of effect options

### LINKED ISSUE #6438: Add ring wall area of effect type
Some spells allow creating a circular wall as opposed to a straight wall. With the new shape options available in V14 we should add "Ring" as an option that has a radius, thickness, and height and creates a ring template on the scene (see https://github.com/foundryvtt/foundryvtt/issues/13374).

arbron:
Adds a new "Wall (ring)" area of effect type and renames the existing wall type to "Wall (straight)". The new wall type uses the `ring` shape using `outerWidth`, so the wall's thickness starts at the radius and goes outward.

Ring Wall Template

Closes #6438

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -4887,7 +4887,13 @@
       }
     },
     "Ring": {
-      "Label": "Wall (ring)"
+      "Label": "Ring",
+      "Counted": {
+        "one": "{number} ring",
+        "other": "{number} rings",
+        "oneSized": "{sizes} Ring",
+        "otherSized": "{number} {sizes} Rings"
+      }
     },
     "Self": {
       "Label": "Self"
@@ -4941,7 +4947,7 @@
       }
     },
     "Wall": {
-      "Label": "Wall (straight)",
+      "Label": "Wall",
       "Counted": {
         "one": "{number} wall",
         "other": "{number} walls",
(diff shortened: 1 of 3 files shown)


lang/en.json
```diff
@@ -4640,6 +4640,9 @@
         "otherSized": "{number} {sizes} Radii"
       }
     },
+    "Ring": {
+      "Label": "Wall (ring)"
```
Fyorl: Maybe just call this 'Ring'? I don't know if the 'Wall' qualifier really adds anything.

arbron: I did want to reflect how official spells word this "You can make the wall up to 60 feet long, 20 feet high, and 1 foot thick, or a ringed wall up to 20 feet in diameter, 20 feet high, and 1 foot thick". Maybe "Ringed Wall" better?

Fyorl: That's fair. I could only find references to this AoE type in wall spells. I think I still prefer just 'Ring' so that we can stick with our single plain noun that we have for all our AoE types. But 'Ring', 'Ringed Wall', and 'Wall (ring)' are fine, so I'll leave it up to you.

Fyorl (approved):
Approved though note that this needs a rebase before merge.

Issue #5433, opened 2025-03-28
Allow for dropping an item directly onto the container's icon on the character sheet


krbz999:
Would be a neat way to move items quickly, removing the need to open the container item's sheet first.

![Image](https://github.com/user-attachments/assets/eb78fcf7-5f1e-4b6b-baea-c2318a6c5a45)

kaelad02:
If we do that, it'd help if the top section of the inventory sheet (the search bar and above) didn't scroll. If not, then you couldn't do this with items at the bottom of the sheet since the container icon could have scrolled off the page.

Sayshal:
Commenting for assignment.

Sayshal:
> If we do that, it'd help if the top section of the inventory sheet (the search bar and above) didn't scroll. If not, then you couldn't do this with items at the bottom of the sheet since the container icon could have scrolled off the page.

With the main issue address I'll likely review this for a seperate PR.

PR #6790, opened 2026-03-01
[#5433] Allow dropping items onto container icons on actor sheets

Sayshal:
- Detect drops on .container[data-item-id] elements in _onDropItem
- Add _onDropItemContainer method to handle container drop logic
- Support same-actor moves (update system.container) and cross-actor copies
- Prevent recursive container drops with allContainers() validation
- Handle NPC gear transformation when dropping onto container icons
- Stack consumables against correct container scope

Closes [#5433](https://github.com/Sayshal/dnd5e/issues/5433)

### DIFF
--- a/module/applications/actor/character-sheet.mjs
+++ b/module/applications/actor/character-sheet.mjs
@@ -1254,7 +1254,7 @@ export default class CharacterActorSheet extends BaseActorSheet {
   /* -------------------------------------------- */
 
   /** @override */
-  async _onDropSingleItem(event, itemData) {
+  async _onDropSingleItem(event, itemData, options={}) {
     // Increment the number of class levels a character instead of creating a new item
     if ( itemData.type === "class" ) {
       const charLevel = this.actor.system.details.level;
@@ -1296,7 +1296,7 @@ export default class CharacterActorSheet extends BaseActorSheet {
       }
     }
 
-    return super._onDropSingleItem(event, itemData);
+    return super._onDropSingleItem(event, itemData, options);
   }
 
   /* -------------------------------------------- */


module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -1733,6 +1756,51 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
+
+    // Transform physical items from NPCs to their gear versions
+    if ( item.actor?.system.isNPC && (item.actor !== this.actor) && item.system.asGear ) {
+      item = await item.system.asGear();
+    }
+
+    // Otherwise, create a new copy inside the container
+    const toCreate = await Item5e.createWithContents([item], {
+      container,
+      transformFirst: itemData => {
```
arbron, line 1790: So this block will never be called when calling `createWithContents` directly into a container. To fix that you'll need to modify `Item5e#createWithContents`:

1) Rename `depth` outside the `createItemData` method to `initialDepth`
2) Adjust the depth check in the `transformCall` to check against `initialDepth` rather than `0`

```javascript
// Lines 1220–1223
    let initialDepth = 0;
    if ( container ) {
      initialDepth = 1 + (await container.system.allContainers()).length;
      if ( initialDepth > PhysicalItemTemplate.MAX_DEPTH ) {

// Line 1232
      if ( transformFirst && (depth === initialDepth) ) newItemData = await transformFirst(newItemData, o);

// Line 1249
    for ( const item of items ) await createItemData(item, container?.id, initialDepth);
```

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -1733,6 +1756,51 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
+        if ( itemData instanceof foundry.abstract.Document ) itemData = itemData.toObject();
+        const stacked = this._onDropStackConsumables(event, itemData, { container: container.id });
+        if ( stacked ) return false;
+        return this._onDropSingleItem(event, itemData);
```
arbron: We'll need to enhance `_onDropSingleItem` here to avoid the need to call `_onDropStackConsumables`. As it is if you drop onto the container but there is a copy of the consumable outside the container it will stack with the item outside the container.

```suggestion
        return this._onDropSingleItem(event, itemData, { container: container.id );
```

Then modify `_onDropSingleItem` to take the container ID and pass it into `_onDropStackConsumables`. Update the JSDoc as necessary of course.

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -1857,13 +1855,15 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
+   * @param {DragEvent} event                  The concluding DragEvent which provided the drop data.
+   * @param {object} itemData                  The item data to create.
+   * @param {object} [options={}]
+   * @param {string} [options.container=null]  ID of the container into which this item is being dropped.
+   * @returns {Promise<object|boolean>}        The item data to create after processing, or false if the item should
+   *                                           not be created or creation has been otherwise handled.
    * @protected
    */
-  async _onDropSingleItem(event, itemData) {
+  async _onDropSingleItem(event, itemData, { container=null }={}) {
```
arbron, line 1866: `CharacterActorSheet` modifies this method, so you'll need to pass this option through that method also.

PR #6798, opened 2026-03-03
[#5559] Add activity behavior for Apply Active Effect behavior

arbron:
Adds a new activity behavior that creates core's Apply Active Effect Region Behavior.


module/data/region-behavior/apply-active-effect.mjs
```diff
@@ -0,0 +1,42 @@
...
+
+  /* -------------------------------------------- */
+  /*  Methods                                     */
+  /* -------------------------------------------- */
+
+  /** @override */
+  createBehaviorData(activity, options={}) {
+    return {
+      system: {
+        effects: this.effects
```
Fyorl: `DifficultTerrainActivityBehavior` uses `Array.from(this.types)` on its own `SetField`. Should they be consistent? Or are they only cosmetically similar?

module/config.mjs
```diff
@@ -1101,6 +1101,11 @@ preLocalize("activityActivationTypes", { key: "label" });
  * @enum {ActivityBehaviorConfiguration}
  */
 DND5E.activityBehaviorTypes = {
+  appylActiveEffect: {
```
```suggestion
  applyActiveEffect: {
```

module/data/region-behavior/apply-active-effect.mjs
```diff
@@ -0,0 +1,42 @@
+import BaseActivityBehavior from "./base-activity-behavior.mjs";
+
+const { DocumentUUIDField, SetField } = foundry.data.fields;
+
+/**
+ * @import { ApplyActiveEffectActivityBehaviorData } from "./_types.mjs";
+ */
+
+/**
+ * Data model representing the difficult terrain activity behavior configuration.
```
Fyorl:
```suggestion
 * Data model representing the active effect activity behavior configuration.
```

module/data/region-behavior/apply-active-effect.mjs
```diff
@@ -0,0 +1,42 @@
...
+
+  /** @override */
+  static LOCALIZATION_PREFIXES = ["DND5E.REGIONBEHAVIORS.APPLYACTIVEEFFECT"];
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    return {
+      effects: new SetField(new DocumentUUIDField())
```
Fyorl:
```suggestion
      effects: new SetField(new DocumentUUIDField({ type: "ActiveEffect", nullable: false }))
```

Fyorl (changes requested):

module/data/region-behavior/apply-active-effect.mjs
arbron: That can probably be removed from the difficult terrain one, just an old habit from before you could use a `Set` to initialize a data model.

arbron:
@Fyorl Rebased this so it can use the system's `dnd5e.applyActiveEffect` behavior rather than core's.

module/applications/create-document-dialog.mjs
```diff
@@ -144,10 +144,10 @@ export default class CreateDocumentDialog extends Dialog5e {
   _onChangeForm(formConfig, event) {
     super._onChangeForm(formConfig, event);
 
-    if ( event.target.name === "type" ) {
+    if ( (event.target.name === "type") && this.documentType.defaultName ) {
       const name = this.element.querySelector('[name="name"]');
       const { pack, parent } = this.options.createOptions;
-      name.placeholder = this.documentType.defaultName({ type: event.target.value, pack, parent });
+      name.placeholder = this.documentType.defaultName?.({ type: event.target.value, pack, parent });
```
Fyorl:
```suggestion
      name.placeholder = this.documentType.defaultName({ type: event.target.value, pack, parent });
```

Fyorl (changes requested):
Thanks for rebase; still got two original suggestions unresolved though.

arbron:
> Thanks for rebase; still got two original suggestions unresolved though.

Whoops, stashed them to do the rebase and then forgot to unstash them :O

Fyorl (approved):

PR #6799, opened 2026-03-03
[#6641] Switch usage messages to store relative UUIDs for effects

arbron:
Changes the newly created usage chat messages to store their effects as relative UUIDs rather than simple IDs to prepare the way for supporting compendium effects in `6.0`.

This contains a small shim in `UsageMessageData#_prepareContext` to handle any messages that were migrated during the testing releae of `5.3.0` which can probably just be removed when `6.0` is released without having to worry about an extra migration.

### DIFF
--- a/module/data/chat-message/usage-message-data.mjs
+++ b/module/data/chat-message/usage-message-data.mjs
@@ -1,7 +1,7 @@
 import ChatMessageDataModel from "../abstract/chat-message-data-model.mjs";
 import { ActorDeltasField } from "./fields/deltas-field.mjs";
 
-const { ArrayField, DocumentIdField, NumberField, ObjectField, StringField } = foundry.data.fields;
+const { ArrayField, DocumentIdField, NumberField, StringField } = foundry.data.fields;
 
 /**
  * @import { UsageMessageSystemData } from "./_types.mjs";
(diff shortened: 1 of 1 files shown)


module/data/chat-message/usage-message-data.mjs
```diff
@@ -21,10 +21,10 @@ export default class UsageMessageData extends ChatMessageDataModel {
   /** @override */
   static defineSchema() {
     return {
-      cause: new StringField(), // TODO: Replace with DocumentUUIDField with `relative: true` in DnD5e 6.0
```
Fyorl, line 24: Do we not still want to do this?

module/data/chat-message/usage-message-data.mjs
```diff
@@ -21,10 +21,10 @@ export default class UsageMessageData extends ChatMessageDataModel {
...
+      cause: new StringField(),
       concentration: new DocumentIdField({ required: false }),
       deltas: new ActorDeltasField({}, { initial: null, nullable: true }),
-      effects: new ArrayField(new DocumentIdField()),
+      effects: new ArrayField(new StringField()),
```
Fyorl: Could also maybe be a `DocumentUUIDField` in v14?


```suggestion
      effects: new ArrayField(new StringField({ blank: false })),
```

PR #6800, opened 2026-03-03
[#6641] Allow standalone AE documents to be linked to activities

### LINKED ISSUE #6641: Allow standalone Active Effect documents to be linked to an activity

arbron:
Adds a new `uuid` field to applied effects data on activities which allows the effect to be retrieved from a compendium rather than from the item.

This is breaking change for `Activity#applicableEffects`, which now returns the effect profiles rather than the effects themselves. A new method `getApplicableEffects` has been added that retrieves the actual effect documents and is async so it can get them from the compendium.

Closes #6641


module/applications/activity/activity-sheet.mjs
```diff
@@ -273,19 +273,20 @@ export default class ActivitySheet extends PseudoDocumentSheet {
         .map(effect => ({
           value: effect.id, label: effect.name, selected: appliedEffects.has(effect.id)
         }));
-      context.appliedEffects = context.activity.effects.reduce((arr, data) => {
-        if ( !data.effect ) return arr;
+      context.appliedEffects = await context.activity.effects.reduce(async (arr, data) => {
```
Fyorl: `Promise.all(map.filter)` is a bit more parallel (and easier to read imo)

module/applications/components/effect-application.mjs
```diff
@@ -70,26 +70,25 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
...
     this.chatMessage = game.messages.get(messageId);
     if ( !this.chatMessage ) return;
 
     // Build the frame HTML only once
     if ( !this.effectsList || !this.targetList ) {
-      if ( !this.effects.length ) {
-        const item = this.chatMessage.getAssociatedItem();
-        this.effects = Array.from(this.querySelectorAll("option")).map(o => item?.effects.get(o.value)).filter(_ => _);
-      }
+      if ( !this.effects.length ) this.effects = (await Promise.all(
```
Fyorl: If the element is moved as part of a render or something, you can often get multiple `connectedCallback` calls in quick succession. Since this is async now, it's possible that we yield on this line and another `connectedCallback` fires in the interim, which will also attempt to construct the `effectsList`.

We need something like `this.#building = this.#buildTray().catch();` and check `if ( !this.#building && (!this.effectsList || !this.targetList )`.

module/applications/components/effect-application.mjs
```diff
@@ -181,9 +180,10 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
    */
   async _applyEffectToActor(effect, actor) {
     const concentration = this.chatMessage.getAssociatedActor()?.effects.get(this.chatMessage.system.concentration);
-    const origin = concentration ?? effect;
+    const item = this.chatMessage.getAssociatedItem() ?? {};
```
Fyorl: If `item` can always be `Item5e|object` then the check `&& item` below is always true.


```suggestion
    const item = this.chatMessage.getAssociatedItem();
```

module/data/activity/fields/applied-effect-field.mjs
```diff
@@ -25,7 +26,21 @@ export default class AppliedEffectField extends SchemaField {
     const item = model.item;
 
     Object.defineProperty(obj, "effect", {
-      get() { return item?.effects.get(this._id); },
+      get() {
+        foundry.utils.logCompatibilityWarning(
```
Fyorl, line 37: When I tested I got some of these warnings fired by `enchant-data.mjs` and `enchant-usage-dialog.mjs`, so they need to switch to `getEffect`.

module/applications/activity/activity-sheet.mjs
```diff
@@ -637,14 +637,19 @@ export default class ActivitySheet extends PseudoDocumentSheet {
...
+    if ( foundry.utils.hasProperty(submitData, "appliedLocalEffects")
+      || foundry.utils.hasProperty(submitData, "appliedRemoteEffects") ) {
+      submitData.effects ??= this.activity.toObject().effects;
+      for ( const _id of submitData.appliedLocalEffects ?? [] ) {
         if ( submitData.effects.find(e => e._id === _id) ) continue;
         submitData.effects.push({ _id });
       }
-      delete submitData.appliedEffects;
+      for ( const uuid of submitData.appliedRemoteEffects ?? [] ) {
+        if ( submitData.effects.find(e => e.uuid === uuid) ) continue;
```
Fyorl: Might as well eagerly filter out enchantments. Or do we want to allow only `base`?


```suggestion
        if ( submitData.effects.find(e => e.uuid === uuid) || fromUuidSync(uuid, { strict: false })?.type === "enchantment" ) continue;
```

module/data/activity/_types.mjs
```diff
@@ -224,3 +224,29 @@
...
  */
+
+/* -------------------------------------------- */
+/*  Fields                                      */
+/* -------------------------------------------- */
+
+/**
+ * @typedef EffectApplicationData
+ * @property {string} _id        ID of configuration, matches local effect ID if UUID isn't specified.
+ * @property {string} uuid       UUID of remove effect to apply.
```
```suggestion
 * @property {string} uuid       UUID of remote effect to apply.
```

module/applications/components/effect-application.mjs
```diff
@@ -196,34 +196,29 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
...
-      if ( !foundry.utils.isEmpty(effectDuration) ) durationOverride = { duration: effectDuration };
-    }
-
     // Enable an existing effect on the target if it originated from this effect
     const existingEffect = actor.effects.find(e => e.origin === origin.uuid);
     if ( existingEffect ) {
       return existingEffect.update(foundry.utils.mergeObject({
-        ...effect.constructor.getInitialDuration(),
-        ...durationOverride,
+        ...effect.constructor.getEffectStart(),
```
Fyorl:
```suggestion
        start: effect.constructor.getEffectStart(),
```

Fyorl (changes requested):

module/applications/components/effect-application.mjs
```diff
@@ -196,34 +198,29 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
...
 
-    // Inherit the activity's duration when the applied effect has no explicit duration of its own
-    let durationOverride = {};
-    if ( !Number.isFinite(effect.duration.value) ) {
-      const effectDuration = this.chatMessage.system.activity?.duration.getEffectData();
-      if ( !foundry.utils.isEmpty(effectDuration) ) durationOverride = { duration: effectDuration };
-    }
-
     // Enable an existing effect on the target if it originated from this effect
     const existingEffect = actor.effects.find(e => e.origin === origin.uuid);
```
Fyorl: Since the origin for remote effects is the item, I think multiple remote effects on one item would collide here.

module/applications/activity/activity-sheet.mjs
```diff
@@ -71,6 +71,14 @@ export default class ActivitySheet extends PseudoDocumentSheet {
 
   /* -------------------------------------------- */
 
+  /**
+   * Type of active effects that can be added to applied effects.
+   * @type {string}
+   */
+  static SUPPORTED_EFFECT_TYPE = "base";
```
Fyorl: This is going to bite us when we inevitably want to allow for multiple AE types.

module/data/activity/enchant-data.mjs
```diff
@@ -71,9 +64,7 @@ export default class BaseEnchantActivityData extends BaseActivityData {
    * @type {EnchantEffectApplicationData[]}
    */
   get availableEnchantments() {
-    const level = this.relevantLevel;
-    return this.effects
-      .filter(e => e.effect && ((e.level.min ?? -Infinity) <= level) && (level <= (e.level.max ?? Infinity)));
+    return this.applicableEffects;
```
Fyorl, line 67: We don't filter for existing effects in `applicableEffects` anymore so we need to check internal uses of `availableEnchantments` and do the filtering there now I suppose.

Fyorl (changes requested):
`enchant-enchantment.hsb` now orphaned I think.

module/data/activity/enchant-data.mjs
arbron: The only place that should be necessary seems to be the usage dialog, and this has already added that filtering when switching over to `getEffect()`.

module/applications/components/effect-application.mjs
arbron: I've moved to checking for the `compendiumSource` for remote effects.

Fyorl: Ah, good solution. I wasn't actually sure what to suggest

module/applications/activity/enchant-sheet.mjs
```diff
@@ -18,14 +18,21 @@ export default class EnchantSheet extends ActivitySheet {
...
     }
   };
 
   /* -------------------------------------------- */
 
+  /** @override */
+  static SUPPORTED_EFFECT_TYPE = "enchantment";
+
+  /* -------------------------------------------- */
+
```
Fyorl:
```suggestion
```

Fyorl (approved):

PR #6806, opened 2026-03-04
[#6805] Fix advancement data being lost if modified after migration

### LINKED ISSUE #6805: [5.3] Modifying advancement on older items causes all advancement data to be lost
Modifying any advancement data (including creating or deleting an advancement) on an actor that has been migrated from the old array-based data structure causes all advancement data on the item to be lost. The advancement data keeps working fine until the world is refreshed and it tries to reinitialize the data.

This happens because the partial changes to the new advancement object don't include any of the data from the old array, so the new advancement data only includes the data that was updated.

arbron:
Adds two fixes to resolve the lost advancement data after the in-memory migration from an array. First, inside the CRUD methods on items, the system checks to see if the item was last updated before `5.3.0` and if so it does a full replacement of the advancement data object.

Second the system does this same change in `migrateItemData` to ensure advancement data is fully commited as an object.

Closes #6805

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -122,6 +122,8 @@ export default class Item5e extends SystemDocumentMixin(Item) {
       }), options);
     }
 
+    Object.defineProperty(this, "_needsAdvancementMigration", { value: Array.isArray(data.system?.advancement) });
+
     return super._initializeSource(data, options);
   }
 
@@ -980,7 +982,7 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     if ( advancement._preCreate(createData) === false ) return;
 
     let update = { [`system.advancement.${advancement.id}`]: advancement.toObject() };
-    if ( !source && foundry.utils.isNewerVersion("5.3.0", this._stats.systemVersion) ) update = {
+    if ( !source && this._needsAdvancementMigration ) update = {
       "system.==advancement": foundry.utils.mergeObject(
         this.system.toObject().advancement, { [advancement.id]: advancement.toObject() }
       )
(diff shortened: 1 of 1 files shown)


module/documents/item.mjs
```diff
@@ -979,7 +979,12 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     const advancement = new cls(data, { parent: this });
     if ( advancement._preCreate(createData) === false ) return;
 
-    const update = { [`system.advancement.${advancement.id}`]: advancement.toObject() };
+    let update = { [`system.advancement.${advancement.id}`]: advancement.toObject() };
+    if ( !source && foundry.utils.isNewerVersion("5.3.0", this._stats.systemVersion) ) update = {
```
Fyorl: I think the embedded children aren't guaranteed to have their `systemVersion`s updated if the parent is updated, particularly via bulk update. So we need to check `(this.parent ?? this)._stats.systemVersion`.

module/documents/item.mjs
```diff
@@ -1002,7 +1007,10 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     if ( !this.system.advancement.has(id) ) throw new Error(`Advancement of ID ${id} could not be found to update`);
 
     const advancement = this.system.advancement.get(id);
-    const update = { [`system.advancement.${id}`]: updates };
+    let update = { [`system.advancement.${id}`]: updates };
+    if ( !source && foundry.utils.isNewerVersion("5.3.0", this._stats.systemVersion) ) update = {
+      "system.==advancement": foundry.utils.mergeObject(this.system.toObject().advancement, { [id]: updates })
```
Fyorl: These updates can contain deletion keys, which we want to apply to the replaced object I think.

```suggestion
      "system.==advancement": foundry.utils.mergeObject(this.system.toObject().advancement, { [id]: updates }, { performDeletions: true })
```

module/documents/item.mjs
arbron: I've realized that using `systemVersion` won't work anyway because any other changes to system data before the advancement data change would lead to the migration not being run. I've swapped over to inspecting the type of the advancement data during initialization and saving that information on the item document itself.

Fyorl (approved):

Issue #6727, opened 2026-02-14
5.1 SRD: Spell Update List


Zanderaf:
Ah shit, here we go again:

# General Issues
- There are rarely any reference enrichers in these spells.
- Most spells do not grant Active Effects.
- I would reckon that having icons and maybe "Cast Spell" named activities might help out the spells to "look" and "feel" better to use.
- Spells could use more diverse icons. Frankly, they're kind of boring and repetitive atm.

## Cantrips

- [ ] Chill Touch could use an applicable AE or two.
- [ ] Dancing Lights could use a 2nd activity with a Bonus Action activation time to move the lights.
- [ ] Druidcraft could maybe use some extra "cosmetic" activities. (Personal preference, spell does a few things so maybe one activity for each effect),
- [ ] Guidance could use an applicable AE.
- [ ] Light could maybe do with an applicable AE and possibly an enchant activity to enchant an item with v14's new active effects affecting token data stuff. Applying a rider AE that grants light might be an idea.
- [ ] Mage Hand could probably use a seperate activity to "control" the mage hand.
- [ ] Prestidigitation; similar to Druidcraft (might be a lot of activities, up for debate IMO)
- [ ] Ray of Frost could use an applicable AE.
- [ ] Resistance could use an applicable AE.
- [ ] Shocking Grasp could use an applicable AE.
- [ ] Thaumaturgy; similar to Druidcraft (might be a lot of activities, up for debate IMO)
- [ ] True Strike could use an applicable AE.
- [ ] Vicious Mockery could use an applicable AE.

## 1st Level Spells

- [ ] Animal Friendship could use an applicable AE.
- [ ] Color Spray could use an applicable AE.
- [ ] Command could use an applicable AE.
- [ ] Comprehend Languages could use an applicable AE.
- [ ] Create or Destroy Water could use several activities.
- [ ] Detect Evil and Good could use an applicable AE.
- [ ] Detect Magic could use an applicable AE.
- [ ] Detect Poison and Disease could use an applicable AE.
- [ ] Disguise Self should be a `transform` activity instead of a `use` activity.
- [ ] Divine Favor could use an applicable AE.
- [ ] Expeditious Retreat could use an applicable AE.
- [ ] Faerie Fire could use an applicable AE.
- [ ] False Life could use an applicable AE to track Temp HP duration.
- [ ] Feather Fall could use an applicable AE.
[...]

PR #6808, opened 2026-03-04
5.1 SRD Update: Cantrips

Zanderaf:
Part of the work needed for https://github.com/foundryvtt/dnd5e/issues/6727. Would appreciate a heads-up on if these changes are acceptable.
- Redid the icons on all of the cantrips to make each one stand out a bit more. (Including assigning icons to the activities)
- Most spells default activity names have been renamed from `Use` `Save` or `Summon` into `Cast Spell` in order to pretty them up a bit.
- Activities such as Chill Touch, Guidance, Ray of Frost, & Resistance now have applicable Active Effects.
- Reference enrichers were added for rules like &reference[brightlight] and &reference[stable].
- Added Description to Shillelagh's enchantment (basically just the spell's description slightly modified).
- Mage Hand now has a 2nd activity to "control" the mage hand, doesn't do anything other than a reminder that controlling it is an action.
- Minor Illusion has a `Check` activity that can be used to make an Investigation check against the caster's Spell Save DC.

### DIFF
--- a/packs/_source/spells/cantrip/ray-of-frost.yml
+++ b/packs/_source/spells/cantrip/ray-of-frost.yml
@@ -141,23 +141,7 @@ effects:
     type: base
     system: {}
     changes:
-      - key: system.attributes.movement.walk
-        mode: 2
-        value: '-10'
-        priority: null
-      - key: system.attributes.movement.fly
-        mode: 2
-        value: '-10'
-        priority: null
-      - key: system.attributes.movement.swim
-        mode: 2
-        value: '-10'
-        priority: null
-      - key: system.attributes.movement.burrow
-        mode: 2
-        value: '-10'
-        priority: null
-      - key: system.attributes.movement.climb
+      - key: system.attributes.movement.bonus
         mode: 2
         value: '-10'
         priority: null


packs/_source/spells/cantrip/acid-splash.yml
```diff
@@ -5,7 +5,7 @@ ownership:
 type: spell
 system:
   description:
-    value: "<p>You hurl a bubble of acid. Choose one creature you can see within range, or choose two creatures you can see within range that are within 5 feet of each other. A target must succeed on a Dexterity saving throw or take\_1d6acid damage.</p><p>This spell's damage increases by\_1d6\_when you reach 5th level (2d6), 11th level (3d6), and 17th level (4d6).</p>"
+    value: "<p>You hurl a bubble of acid. Choose one creature you can see within range, or choose two creatures you can see within range that are within 5 feet of each other. A target must succeed on a Dexterity saving throw or take\_1d6 acid damage.</p><p>This spell's damage increases by\_1d6\_when you reach 5th level (2d6), 11th level (3d6), and 17th level (4d6).</p>"
```
krbz999:
```suggestion
    value: "<p>You hurl a bubble of acid. Choose one creature you can see within range, or choose two creatures you can see within range that are within 5 feet of each other. A target must succeed on a Dexterity saving throw or take 1d6 acid damage.</p><p>This spell's damage increases by 1d6 when you reach 5th level (2d6), 11th level (3d6), and 17th level (4d6).</p>"
```

packs/_source/spells/cantrip/dancing-lights.yml
```diff
@@ -10,10 +10,10 @@ system:
       appear as torches, lanterns, or glowing orbs that hover in the air for the
       duration. You can also combine the four lights into one glowing vaguely
       humanoid form of Medium size. Whichever form you choose, each light sheds
-      dim light in a 10-foot radius.</p><p>As a bonus action on your turn, you
-      can move the lights up to 60 feet to a new spot within range. A light must
-      be within 20 feet of another light created by this spell, and a light
-      winks out if it exceeds the spell's range.</p>
+      &amp;reference[dimlight] in a 10-foot radius.</p><p>As a bonus action on
```
arbron: Since spaces work in reference enrichers, it would be nice to keep these as written so they read a bit better:

```suggestion
      &amp;Reference[dim light] in a 10-foot radius.</p><p>As a bonus action on
```

packs/_source/spells/cantrip/acid-splash.yml
```diff
@@ -102,20 +105,27 @@ system:
       uses:
         spent: 0
         recovery: []
+        max: ''
       sort: 0
+      name: Cast Spell
+      img: icons/magic/acid/projectile-glowing-bubbles.webp
```
arbron: I'd really hate to set the precedent that activities are expected to have custom icons (it's already enough work finding a unique icon for each item).

I've had a thought to add more generic "Verb" SVG icons into the system to aid with these: https://github.com/foundryvtt/dnd5e/issues/6810

packs/_source/spells/cantrip/dancing-lights.yml
Zanderaf: Got it, I wasn't aware of that myself, must of done it wrong the first time I was using it for a personal project and stuck with using them with no spaces.

packs/_source/spells/cantrip/acid-splash.yml
Zanderaf: I'm fine with removing the activity icons if that's not something the system wants to include on its own.

Zanderaf: > I'd really hate to set the precedent that activities are expected to have custom icons (it's already enough work finding a unique icon for each item).
> 
> I've had a thought to add more generic "Verb" SVG icons into the system to aid with these: #6810

Any opinion on the name changes to the activities?

arbron:
Other than those comments, looks like the right direction.

Zanderaf:
> Other than those comments, looks like the right direction.

Removed the Activity icons and added the spaces to the bright/dim light references. Tiny bit sad that we can't get icons for the SRD activities but understandable!

Light also had 2 save activities instead of a Use & Save, should be fixed as well.

packs/_source/spells/cantrip/acid-splash.yml
arbron: In the PHB we used just "Cast" rather than "Cast Spell", but the idea is good.

Zanderaf: Would you like me to swap them to just "Cast" then?

arbron: Yeah, let's match.

Zanderaf:
Made the change from Cast Spell to just Cast.

On a side not, Acid Splash seems to keep adding the /_1d6 acid_/ whenever I make changes, the thing Zhell mentioned when I first made the PR. Whack. Fixed it this time 'round too.

krbz999:
Yeah I can confirm it has that; deleting a words words and typing them out again probably gets rid of it ('remove formatting' does not).

packs/_source/spells/cantrip/ray-of-frost.yml
```diff
@@ -113,20 +120,85 @@ system:
...
+        value: '-10'
+        priority: null
+      - key: system.attributes.movement.burrow
+        mode: 2
+        value: '-10'
+        priority: null
+      - key: system.attributes.movement.climb
+        mode: 2
+        value: '-10'
+        priority: null
```
arbron: These should use `system.attributes.movement.bonus` instead of targeting individual movement types, otherwise you risk double-reducing a certain movement type if it is based on another (e.g. if you set your fly speed equal to your walk speed, and then subtract 10 from both, it will be walk of 30 minus 10 = 20, fly equal to walk minus 10 = 10).


```suggestion
      - key: system.attributes.movement.bonus
        mode: 2
        value: '-10'
        priority: null
```

PR #6811, opened 2026-03-05
[#6809] Apply healing modifications to HP restored via hit dice.

### LINKED ISSUE #6809: Apply healing damage modification to HP restored via hit dice.

Fyorl:
- Closes #6809

### DIFF
--- a/module/documents/_types.mjs
+++ b/module/documents/_types.mjs
@@ -31,7 +31,7 @@
  * @typedef DamageDescription
  * @property {number} value                          Amount of damage.
  * @property {string} type                           Type of damage.
- * @property {Set<string>} properties                Physical properties that affect damage application.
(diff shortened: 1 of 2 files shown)


arbron (approved):
This does make things consistent with other sources of healing, but it feels like we should be excluding healing types from the "All Damage" category. Right now if you are resistant to all damage types then it also makes you are resistant to healing, which doesn't seem right.

Fyorl:
> This does make things consistent with other sources of healing, but it feels like we should be excluding healing types from the "All Damage" category. Right now if you are resistant to all damage types then it also makes you are resistant to healing, which doesn't seem right.

A good point; I've amended our damage application code to exclude healing types from interacting with `ALL`. Let me know if it looks correct to you.

arbron (approved):
Looks good to my eyes and to my testing.

PR #6815, opened 2026-03-06
Improve detached window UX in v14.

Fyorl:
Raising this now so that I don't forget about it, but you won't be able to test it until the next v14 release is out, so no need to review until then.


arbron:
- Opening activity from detached item opens in its own detached window, but active effects & advancement open in base window
- Detached activities are the incorrect size, get smaller every time the tab is changed
- Expected "Add Background" etc buttons on Character sheet to open up detached compendium browser
- Expected rolling dialogs (like when clicking on a skill) to open up in actor's detached window
- Expected delete dialogs to open in actor's detached window
- Expected "Create Document" dialog when invoked on actor sheet to open in same detached window
- Expected active effect sheet to open in actor's detached window when opened from actor

arbron (changes requested):
Noted above...

Fyorl:
> * Opening activity from detached item opens in its own detached window, but active effects & advancement open in base window
> * Detached activities are the incorrect size, get smaller every time the tab is changed
> * Expected "Add Background" etc buttons on Character sheet to open up detached compendium browser
> * Expected rolling dialogs (like when clicking on a skill) to open up in actor's detached window
> * Expected delete dialogs to open in actor's detached window
> * Expected "Create Document" dialog when invoked on actor sheet to open in same detached window
> * Expected active effect sheet to open in actor's detached window when opened from actor

Thanks, these are good catches. I've addressed most of them. The activity window sizing is looking maybe like a core bug. It also affects item sheets if you detach them and change tabs, for example. I need to investigate further.

Fyorl:
> The activity window sizing is looking maybe like a core bug. It also affects item sheets if you detach them and change tabs, for example. I need to investigate further.

I got to the bottom of this. Not a core bug. Should be good to go now.

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -1381,20 +1383,23 @@ export default class BaseActorSheet extends PrimarySheetMixin(
   static #roll(event, target) {
     if ( !target.classList.contains("rollable") ) return;
     if ( this._roll(event, target) === false ) return;
+    const dialog = { sheet: this };
```
arbron: I'm wondering if in this situation since we have the `PointerEvent` that initiated the roll, if we could instead determine the sheet automatically from that. That would allow rolls from enrichers to automatically attach to the appropriate window.

arbron (changes requested):
One comment about how to handle rolls, and one other note. Otherwise it is looking good. 
 
- Activities used from actor sheet should open in same window

module/enrichers.mjs
```diff
@@ -243,10 +243,13 @@ export async function enrichAttack(config, label, options) {
  */
 async function rollAttack(config, event) {
   const { activityUuid, attackMode, formula, scaling } = config;
+  let sheet;
+  const app = foundry.applications.instances.get(event?.target?.closest(".application")?.id);
```
arbron: I was thinking of handling this in `BasicRoll.buildConfigure` before the dialog is created with something like:

```javascript
dialog.sheet ??= foundry.applications.instances.get(config.event?.target?.closest(".application")?.id);
```

If not, then you'll need to also pass the sheet to the methods in `_triggerSubsequentActions` on the attack, damage, and heal activities.

arbron (changes requested):

module/enrichers.mjs
Fyorl: Of course, much nicer.

arbron (approved):

PR #6820, opened 2026-03-09
[#6813] Add expanded section showing effect changes & description

### LINKED ISSUE #6813: Add expand button to active effects on actor/item sheets that displays changes & description
Like items can be expanded to see their description and activities, being able to expand active effects entries to display their enriched description as well as any changes they have would be helpful. If #6362 is implemented, this would make a place where activities attached to effects can be displayed. Also, this will be very useful when #6671 is implemented to give a place to see if individual changes have been applied or skipped based on their conditions.

Related #2906

arbron:
Adds the ability to expand active effects on actor and items sheets to display a list of their changes as well as the effect's description and tags.

Effect Changes Breakout - Actor

The change rows display the key (with a human-readable name if available), the change type indicated by a symbol, and the value (displayed as tooltip instead if too long).

Effect Changes Breakout - Enchantment w: Riders

This includes improvements to `getHumanReadableAttributeLabel` to support displaying as many active effect change keys as possible and to improve how certain keys are displayed:
- Support `activities[attack]...` style keys
- Support generic item keys that search through all item schemas
- Improved labels within `MappingField` to include entry name

Closes #6813

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2020,6 +2020,9 @@
           "label": "Base Damage Types"
         }
       },
+      "bonus": {
+        "label": "Damage Bonus"
+      },
       "label": "Ammunition Damage",
       "replace": {
         "label": "Replace Weapon Damage",
@@ -2545,12 +2548,12 @@
 "DND5E.Dusk": "Dusk",
 
 "DND5E.Effect": "Effect",
-
 "DND5E.EFFECT": {
   "Action": {
-    "Create": "Create Effect",
-    "Delete": "Delete Effect",
-    "Dissociate": "Dissociate Effect",
+    "CreateEffect": "Create Effect",
+    "DeleteEffect": "Delete Effect",
+    "DissociateEffect": "Dissociate Effect",
+    "EditEffect": "Edit Effect",
     "Search": "Search effects"
   },
   "Application": {
@@ -2571,9 +2574,16 @@
       }
     }
   },
+  "Category": {
+    "Temporary": "Temporary Effects",
+    "Passive": "Passive Effects",
+    "Inactive": "Inactive Effects",
+    "Unavailable": "Unavailable Effects"
+  },
   "DropHint": "Drop an Active Effect here",
   "Empty": "No associated effects. Use the <i class=\"fas fa-plus\"></i> button above to create one, or select an existing effect from the drop-down.",
   "Label": "Available Effects",
+  "New": "New Effect",
   "RIDER": {
     "Activity": "Enchantment Rider Activity",
     "Effect": "Enchantment Rider Effect",
@@ -2586,9 +2596,17 @@
       }
     }
   },
+  "Status": {
+    "Inactive": "Inactive",
+    "Passive": "Passive",
+    "Temporary": "Temporary",
+    "Unavailable": "Unavailable"
+  },
+  "Suppressed": {
+    "Hint": "Source item must be equipped or attuned to activate these"
+  },
   "Tab": "Effects"
 },
-
 
 "DND5E.ENCHANT": {
   "FIELDS": {
@@ -2732,6 +2750,44 @@
   }
 },
 
+"DND5E.Encumbrance": "Encumbrance",
+"DND5E.ENCUMBRANCE": {
+  "FIELDS": {
+    "attributes": {
+      "encumbrance": {
+        "bonuses": {
+          "encumbered": {
+            "label": "Bonus to Encumbered Threshold"
+          },
+          "heavilyEncumbered": {
+            "label": "Bonus to Heavily Encumbered Threshold"
+          },
+          "maximum": {
+            "label": "Bonus to Maximum Carrying Capacity"
+          },
+          "overall": {
+            "label": "Bonus to All Encumberance Thresholds"
+          }
+        },
+        "multipliers": {
+          "encumbered": {
+            "label": "Encumbered Threshold Multiplier"
+          },
+          "heavilyEncumbered": {
+            "label": "Heavily Encumbered Threshold Multiplier"
+          },
+          "maximum": {
+            "label": "Maximum Carrying Capacity Multiplier"
+          },
+          "overall": {
+            "label": "Multiplier for All Encumberance Thresholds"
+          }
+        }
+      }
+    }
+  }
+},
+
 "DND5E.Environment": "Environment",
 
 "DND5E.EQUIPMENT": {
@@ -2790,6 +2846,7 @@
 },
 
 "DND5E.Expertise": "Expertise",
+"DND5E.Eyes": "Eyes",
 
 "DND5E.FACILITY": {
   "Action": {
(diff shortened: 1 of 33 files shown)


thatlonelybugbear:
There are modules which elect to push flags on the common status effects, like blinded etc.

Would the system allow for those to also be viewable?

PS: I love that this showcases one small typo in the Tactial Master 😅 

image

arbron:
> There are modules which elect to push flags on the common status effects, like blinded etc.
> 
> Would the system allow for those to also be viewable?

How do you mean? Are these flags added to the effect created by the status condition?

thatlonelybugbear:
Yes, via preCreate hooks, and updateSource for example.

arbron:
> Yes, via preCreate hooks, and updateSource for example.

~~Well, unless they are creating new actual changes, then they won't display here. If they do create changes then they will :).~~

Oh, I guess if you mean will the status effects at the bottom that normally aren't displayed as their own effects, will those list changes if present. Not currently.

thatlonelybugbear:
Yes the toggleable _Conditions_ in the Actor's effects tab

module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -249,26 +253,27 @@ export default class BaseActorSheet extends PrimarySheetMixin(
...
     for ( const category of Object.values(context.effects) ) {
+      category.columns = columns;
       category.effects = await category.effects.reduce(async (arr, effect) => {
-        effect.updateDuration();
         if ( conditionIds.has(effect.id) && !effect.duration.remaining ) return arr;
-        const { id, name, img, disabled, duration } = effect;
         const toggleable = !this._concentration?.effects.has(effect);
-        let source = await effect.getSource();
+        const isExpanded = this.expandedSections.get(`effects.${effect.id}`) === true;
+        const context = {
```
Fyorl: Bit safer to avoid shadowing outer `context`.
```suggestion
        const ctx = {
```

module/applications/components/effects.mjs
```diff
@@ -274,6 +337,44 @@ export default class EffectsElement extends (foundry.applications.elements.Adopt
...
+    if ( !effect ) return;
+
+    const expanded = this.app.expandedSections.get(`effects.${effect.id}`);
+    if ( expanded ) {
+      summary.parentElement.addEventListener("transitionend", () => {
+        if ( row.classList.contains("collapsed") ) summary.querySelector(".item-summary")?.remove();
+      }, { once: true });
+      this.app.expandedSections.set(`effects.${effect.id}`, false);
+    } else {
+      const context = await effect.getPreviewContext();
```
Fyorl:
```suggestion
      const context = await effect.getPreviewContext({ secrets: effect.isOwner });
```

module/data/actor/templates/common.mjs
```diff
@@ -32,23 +32,30 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
...
         }),
         proficient: new NumberField({
-          required: true, integer: true, min: 0, max: 1, initial: 0, label: "DND5E.ProficiencyLevel"
+          required: true, integer: true, min: 0, max: 1, initial: 0, label: "DND5E.ProficiencyLevel",
+          labelFormatter: "DND5E.ABILITY.Formatter.SaveProficiency"
         }),
         max: new NumberField({
-          required: true, integer: true, nullable: true, min: 0, initial: null, label: "DND5E.AbilityScoreMax"
+          required: true, integer: true, nullable: true, min: 0, initial: null, label: "DND5E.AbilityScoreMax",
+          labelFormatter: "DND5E.ABILITY.Formatter.Maxmimum"
```
Fyorl:
```suggestion
          labelFormatter: "DND5E.ABILITY.Formatter.Maximum"
```

lang/en.json
```diff
@@ -94,6 +94,19 @@
     "SaveLabel": "{ability} Save",
     "Title": "Configure {ability}"
   },
+  "Formatter": {
+    "CheckAdvantageMode": "{name} Check Advantage Mode",
```
Fyorl: We are generally moving away from these sorts of concatenated keys so I would probably expect this to be

```json
"Check": {
  "AdvantageMode": "{name} Check Advantage Mode"
  ...
}
```

And then we adjust the `labelFormatterPrefix` to `DND5E.ABILITY.Formatter.Check.`. I guess there's a reason we can't do this easily?

module/documents/active-effect.mjs
```diff
@@ -973,30 +973,81 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
-      content: await foundry.applications.handlebars.renderTemplate(
-        "systems/dnd5e/templates/effects/parts/effect-tooltip.hbs", {
-          effect: this,
-          description: await TextEditor.enrichHTML(this.description ?? "", { relativeTo: this, ...enrichmentOptions }),
-          durationParts: this.duration.remaining ? this.duration.label.split(", ") : [],
-          showDuration: Number.isFinite(this.duration.value),
-          properties
+      properties,
+      description: await TextEditor.enrichHTML(this.description ?? "", {
+        relativeTo: this
```
Fyorl, line 995:
```suggestion
        ...enrichmentOptions,
        relativeTo: this
```

module/documents/active-effect.mjs
```diff
@@ -973,30 +973,81 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Prepare the context used to display an effect on an actor or item sheet.
+   * @param {object} [options={}]
+   * @param {number} [options.maxKeyLength=35]  Maximum length of the attribute key displayed.
+   * @returns {object}  An object of chat data to render.
+   */
+  async getSheetContext({ maxKeyLength=35 }={}) {
```
Fyorl, line 1011: These full attribute keys look very noisy on the sheet when the effects are expanded. I am going to do a style pass that relegates them to tooltips similar to the ones we have on the change value column, but consider removing them entirely.

module/documents/active-effect.mjs
```diff
@@ -973,30 +973,81 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+          displayKey = displayKey.replace(/^systems.|^flags.|^activities\[/, "");
+          if ( displayKey.length > maxKeyLength ) displayKey = displayKey.slice(1 - maxKeyLength);
+          displayKey = `…${displayKey}`;
         }
+        return {
+          ...change, displayKey,
+          name: getHumanReadableAttributeLabel(change.key, { ...attributeCtx, prefixItemName: false })
+        };
+      }),
+      durationParts: Number.isFinite(duration.remaining) ? duration.label.split(", ") : [],
```
Fyorl, line 1032: We check falsiness in `richTooltip` so we should align the two one way or the other, though I think with #6837 we have a bespoke method for duration parts regardless.

module/utils.mjs
```diff
@@ -1327,52 +1332,71 @@ export function getHumanReadableAttributeLabel(attr, { actor, item }={}) {
...
+    label = mapping.getFieldLabel(key, keyPath.toReversed());
   }
 
   // Senses
   else if ( attr.startsWith("attributes.senses.ranges.") ) {
     const key = attr.split(".")[3];
     label = CONFIG.DND5E.senses[key]?.label;
   }
+  // Senses
+  else if ( attr.startsWith("attributes.senses.ranges.") ) label = CONFIG.DND5E.senses[attr.split(".")[3]]?.label;
```
Fyorl, line 1393: Already handling this case above. Maybe rebase artefact?

Fyorl (changes requested):

Fyorl (approved):

PR #6821, opened 2026-03-10
[#6672] Implement `FiltersField`, `FiltersEditor`, and input

### LINKED ISSUE #6672: Create `FiltersField` for storing filters data
Add a new data field for storing data structures for `Filters` to allow them to be stored. This will be used for AE conditions (https://github.com/foundryvtt/dnd5e/issues/6671) and Rule effects (https://github.com/foundryvtt/dnd5e/issues/6315).

Because `Filters` are a recursive structure and data models don't support recursion, this data will need to be either stored as raw json (using a `JSONField`) or flattened out when saved and then re-nested when accessed.

arbron:
Adds a new `FiltersField` that extends `JSONField` for handling filters. When this field is initialized it creates a new `Filter` object that makes it easy to check the filter and caches the result.

Filter Input

When placed into a form it displays a list of the filters and a button that opens the filter editor. The editor is currently just a `<code-mirror>` editor which will be replaced with a proper UI in the future.

Closes #6672

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3082,7 +3082,7 @@
   "Invalid": "Invalid Filter",
   "Operator": {
     "And": "and",
-    "Not": "Not",
+    "Not": "not",
     "Or": "or"
   }
 },
--- a/lang/en.json
+++ b/lang/en.json
@@ -3073,6 +3073,20 @@
 "DND5E.FilterGroupOrigin": "Group by Origin",
 "DND5E.FilterGroupAction": "Group by Action",
 "DND5E.FilterNoSpells": "No spells found for this set of filters.",
+
+"DND5E.FILTER": {
+  "Action": {
+    "Edit": "Edit Filters"
+  },
+  "Empty": "No Filters",
+  "Invalid": "Invalid Filter",
+  "Operator": {
+    "And": "and",
+    "Not": "Not",
+    "Or": "or"
+  }
+},
+
 "DND5E.NoSpellLevels": "This character has no spellcaster levels, but you may add spells manually.",
 
 "DND5E.FLAGS": {
--- a/less/elements.less
+++ b/less/elements.less
@@ -96,6 +96,51 @@ filter-state {
 }
 
 /* ---------------------------------- */
+/*  Filters Input                     */
+/* ---------------------------------- */
+
+filters-input {
+  &, filter-comparison, filter-operator {
+    display: flex;
+    align-items: center;
+    gap: 0.25rem;
+  }
+  filters-breakdown {
+    flex: 1;
+    display: flex;
+    flex-wrap: wrap;
+    font-family: var(--dnd5e-font-roboto);
+    font-size: var(--font-size-11);
+  }
+  filter-comparison {
+    border: var(--pill-border);
+    border-radius: 3px;
+    padding: 0.1875rem 0.375rem;
+
+    filter-key { display: contents; }
+    &[operator="exact"] filter-key::after { content: "="; }
+    &[operator="gt"] filter-key::after { content: ">"; }
+    &[operator="gte"] filter-key::after { content: "≥"; }
+    &[operator="lt"] filter-key::after { content: "<"; }
+    &[operator="lte"] filter-key::after { content: "≤"; }
+    &[operator$="startswith"] filter-key::after { content: "⋞"; }
+    &[operator$="endswith"] filter-key::after { content: "⋟"; }
+    &[operator$="contains"] filter-key::after { content: "⊂"; }
+    &[operator^="has"] filter-key::after { content: "∩"; }
+    &[operator="in"] filter-key::after { content: "⊂"; }
+  }
+  filter-operator {
+    flex-wrap: wrap;
+
+    &::before, &::after { font-size: var(--font-size-18); }
+    &:not([operator="NOT"]) {
+      &::before { content: "⟨"; }
+      &::after { content: "⟩"; }
+    }
+  }
+}
+
+/* ---------------------------------- */
 /*  Identifier Input                  */
 /* ---------------------------------- */
 
--- a/less/elements.less
+++ b/less/elements.less
@@ -118,16 +118,16 @@ filters-input {
     padding: 0.1875rem 0.375rem;
 
     filter-key { display: contents; }
-    &[operator="exact"] filter-key::after { content: "="; }
-    &[operator="gt"] filter-key::after { content: ">"; }
-    &[operator="gte"] filter-key::after { content: "≥"; }
-    &[operator="lt"] filter-key::after { content: "<"; }
-    &[operator="lte"] filter-key::after { content: "≤"; }
-    &[operator$="startswith"] filter-key::after { content: "⋞"; }
-    &[operator$="endswith"] filter-key::after { content: "⋟"; }
-    &[operator$="contains"] filter-key::after { content: "⊂"; }
-    &[operator^="has"] filter-key::after { content: "∩"; }
-    &[operator="in"] filter-key::after { content: "⊂"; }
+    &[operator="exact"] > filter-key::after { content: "="; }
+    &[operator="gt"] > filter-key::after { content: ">"; }
+    &[operator="gte"] > filter-key::after { content: "≥"; }
+    &[operator="lt"] > filter-key::after { content: "<"; }
+    &[operator="lte"] > filter-key::after { content: "≤"; }
+    &[operator$="startswith"] > filter-key::after { content: "⋞"; }
+    &[operator$="endswith"] > filter-key::after { content: "⋟"; }
+    &[operator$="contains"] > filter-key::after { content: "⊇"; }
+    &[operator^="has"] > filter-key::after { content: "∋"; }
+    &[operator="in"] > filter-key::after { content: "∈"; }
   }
   filter-operator {
     flex-wrap: wrap;
--- a/module/applications/components/filters-input.mjs
+++ b/module/applications/components/filters-input.mjs
@@ -1,0 +1,209 @@
+import { COMPARISON_FUNCTIONS, OPERATOR_FUNCTIONS } from "../../filter.mjs";
+import { getHumanReadableAttributeLabel } from "../../utils.mjs";
+import FiltersEditor from "../filters-editor.mjs";
+
+/**
+ * Element for displaying and editing filters.
+ * @extends {AbstractFormInputElement}
+ */
+export default class FiltersInputElement extends foundry.applications.elements.AbstractFormInputElement {
+
+  /** @override */
+  static tagName = "filters-input";
+
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Element containing the filters breakdown.
+   * @type {HTMLElement}
+   */
+  #breakdown;
+
+  /* -------------------------------------------- */
+
+  /**
+   * Element for opening the editor.
+   * @type {HTMLButtonElement}
+   */
+  #button;
+
+  /* -------------------------------------------- */
+
+  /**
+   * Editor instance opened by this input.
+   * @type {FiltersConfig}
+   */
+  #editor;
+
+  /* -------------------------------------------- */
+
+  /** @override */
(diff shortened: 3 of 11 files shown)


lang/en.json
```diff
@@ -2879,16 +2881,26 @@
...
+
+"DND5E.FILTER": {
+  "Action": {
+    "Edit": "Edit Filters"
+  },
+  "Empty": "No Filters",
+  "Invalid": "Invalid Filter",
+  "Operator": {
+    "And": "and",
+    "Not": "Not",
```
```suggestion
    "Not": "not",
```

less/elements.less
```diff
@@ -78,6 +78,51 @@ filter-state {
...
+    filter-key { display: contents; }
+    &[operator="exact"] filter-key::after { content: "="; }
+    &[operator="gt"] filter-key::after { content: ">"; }
+    &[operator="gte"] filter-key::after { content: "≥"; }
+    &[operator="lt"] filter-key::after { content: "<"; }
+    &[operator="lte"] filter-key::after { content: "≤"; }
+    &[operator$="startswith"] filter-key::after { content: "⋞"; }
+    &[operator$="endswith"] filter-key::after { content: "⋟"; }
+    &[operator$="contains"] filter-key::after { content: "⊂"; }
+    &[operator^="has"] filter-key::after { content: "∩"; }
```
Fyorl: ∋ is the symbol for 'contains as an element', e.g. `{a,b,c} ∋ b`

```suggestion
    &[operator^="has"] filter-key::after { content: "∋"; }
```

less/elements.less
```diff
@@ -78,6 +78,51 @@ filter-state {
...
+    &[operator="in"] filter-key::after { content: "⊂"; }
```
Fyorl: And the inverse:


```suggestion
    &[operator="in"] filter-key::after { content: "∈"; }
```

less/elements.less
```diff
@@ -78,6 +78,51 @@ filter-state {
...
+
+    filter-key { display: contents; }
+    &[operator="exact"] filter-key::after { content: "="; }
+    &[operator="gt"] filter-key::after { content: ">"; }
+    &[operator="gte"] filter-key::after { content: "≥"; }
+    &[operator="lt"] filter-key::after { content: "<"; }
+    &[operator="lte"] filter-key::after { content: "≤"; }
+    &[operator$="startswith"] filter-key::after { content: "⋞"; }
+    &[operator$="endswith"] filter-key::after { content: "⋟"; }
+    &[operator$="contains"] filter-key::after { content: "⊂"; }
```
Fyorl: Think this is the wrong way around. Also 'superset of or equal to' is probably more precise:

```suggestion
    &[operator$="contains"] filter-key::after { content: "⊇"; }
```

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+  #refreshBreakdown() {
+    let filters;
+    try {
+      filters = JSON.parse(this.value);
+    } catch(err) {
+      this.#breakdown.innerText = game.i18n.localize("DND5E.FILTER.Invalid");
+      return;
+    }
+
+    if ( foundry.utils.isEmpty(filters) ) {
```
Fyorl: Can we add a class so that styling the empty state is easier:


```suggestion
    const isEmpty = foundry.utils.isEmpty(filters);
    this.#breakdown.classList.toggle("empty", isEmpty);
    if ( isEmpty ) {
```

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+      this.#breakdown.innerText = game.i18n.localize("DND5E.FILTER.Empty");
+      return;
+    }
+
+    if ( !Array.isArray(filters) ) filters = [filters];
+    this.#breakdown.replaceChildren(...filters.map(f => this.#renderBreakdownNode(f)).filter(_ => _));
```
Fyorl: This doesn't match the semantics of `perfomCheck`, which ANDs top-level elements.
```suggestion
    this.#breakdown.replaceChildren(this.#renderBreakdownNode(
      Array.isArray(filters) ? { o: "AND", v: filters } : filters
    ));
```

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+  /* -------------------------------------------- */
+
+  /**
+   * Turn a single filter node into an breakdown element.
+   * @param {FilterDescription} filter
+   * @returns {HTMLElement|void}
+   */
+  #renderBreakdownNode(filter) {
+    let { k: key, o: operator="exact", v: value } = filter;
+    if ( !value ) return;
```
Fyorl: Presumably values like `0` or `false` are valid filter values.

```suggestion
    if ( value == null ) return;
```

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+   * Element for opening the editor.
+   * @type {HTMLButtonElement}
+   */
+  #button;
+
+  /* -------------------------------------------- */
+
+  /**
+   * Editor instance opened by this input.
+   * @type {FiltersConfig}
```
Fyorl:
```suggestion
   * @type {FiltersEditor}
```

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+
+  /**
+   * Handle opening the filters config application for this input.
+   * @param {PointerEvent} event  Triggering click event.
+   */
+  #onOpenEditor(event) {
+    event.preventDefault();
+    const edit = new Event("edit", { bubbles: true, cancelable: true });
+    this.dispatchEvent(edit);
+    if ( edit.defaultPrevent ) return;
```
Fyorl:
```suggestion
    if ( edit.defaultPrevented ) return;
```

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+
+    this.#editor = new FiltersEditor({ value: this.value });
```
Fyorl, line 161: Make sure any existing editor is closed first.


```suggestion
    this.#editor?.close();
    this.#editor = new FiltersEditor({ value: this.value });
```

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    // If in COMPARISON_FUNCTIONS, create entry
+    if ( operator in COMPARISON_FUNCTIONS ) {
+      const node = document.createElement("filter-comparison");
+      node.setAttribute("operator", operator);
+      const keyElement = document.createElement("filter-key");
+      keyElement.innerText = getHumanReadableAttributeLabel(key) ?? key;
+      const valueElement = document.createElement("filter-value");
+      valueElement.innerText = Array.isArray(value) ? value.join(", ") : value;
```
Fyorl: `has` supports sub-filter values, so we need to recurse into `value` here if it is a plain object.

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  connectedCallback() {
+    const operator = this.getAttribute("operator");
+    const symbol = operator?.endsWith("AND") ? game.i18n.localize("DND5E.FILTER.Operator.And")
+      : operator?.endsWith("OR") ? game.i18n.localize("DND5E.FILTER.Operator.Or") : null;
+    if ( symbol ) this.querySelectorAll(":is(filter-comparison, filter-operator):not(:last-child)").forEach(e =>
+      e.insertAdjacentHTML("afterend", `<span class="seperator">${symbol}</span>`)
```
Fyorl:
```suggestion
      e.insertAdjacentHTML("afterend", `<span class="separator">${symbol}</span>`)
```

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+    );
+    if ( operator?.startsWith("N") ) this.insertAdjacentHTML(
+      "afterbegin", `<span class="seperator">${game.i18n.localize("DND5E.FILTER.Operator.Not")}`
```
Fyorl:
```suggestion
      "afterbegin", `<span class="separator">${game.i18n.localize("DND5E.FILTER.Operator.Not")}</span>`
```

module/applications/components/filters-input.mjs
```diff
@@ -0,0 +1,209 @@
...
+    );
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  disconnectedCallback() {
+    this.querySelectorAll(".seperator").forEach(e => e.remove());
```
Fyorl:
```suggestion
    this.querySelectorAll(".separator").forEach(e => e.remove());
```

less/elements.less
```diff
@@ -78,6 +78,51 @@ filter-state {
...
+    font-family: var(--dnd5e-font-roboto);
+    font-size: var(--font-size-11);
+  }
+  filter-comparison {
+    border: var(--pill-border);
+    border-radius: 3px;
+    padding: 0.1875rem 0.375rem;
+
+    filter-key { display: contents; }
+    &[operator="exact"] filter-key::after { content: "="; }
```
Fyorl: Seems we allow for `o: "_"` to mean 'exact' as well.


```suggestion
    &[operator="exact"], &[operator="_"] filter-key::after { content: "="; }
```

module/data/fields/filters-field.mjs
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A field for storing filters configuration that initializes to a `Filter` instance.
+ */
+export default class FiltersField extends foundry.data.fields.JSONField {
+
+  /** @override */
+  static get _defaults() {
+    return Object.assign(super._defaults, {
+      initial: "{}"
+    });
```
Fyorl:
```suggestion
    return Object.assign(super._defaults, { initial: "{}" });
```

module/applications/filters-editor.mjs
```diff
@@ -0,0 +1,107 @@
...
+    }
+  };
+
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * The active CodeMirror editor instance.
+   * @type {EditorView}
```
Fyorl:
```suggestion
   * @type {HTMLCodeMirrorEditor}
```

module/filter.mjs
```diff
@@ -2,6 +2,72 @@
...
+
+  /* -------------------------------------------- */
+  /*  Methods                                     */
+  /* -------------------------------------------- */
+
+  /**
+   * Check some data against the filter to determine if it matches. Result will be cached between multiple calls.
+   * @param {object} data  Arbitrary data object to check.
+   * @returns {boolean}
+   * @throws
```
Fyorl: We catch the error and don't re-throw, so annotation is not accurate.


```suggestion
```

module/filter.mjs
```diff
@@ -2,6 +2,72 @@
...
+    }
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Clear the cached result of the filter check and perform the check again.
+   * @param {object} data  Arbitrary data object to check.
+   * @returns {boolean}
+   * @throws
```
Fyorl:
```suggestion
```

module/filter.mjs
```diff
@@ -2,6 +2,72 @@
...
+   */
+  check(data) {
+    if ( this.#result !== undefined ) return this.#result;
+    try {
+      if ( foundry.utils.isEmpty(this.#definition) ) this.#result = true;
+      else this.#result = performCheck(data, this.#definition);
+      return this.#result;
+    } catch(err) {
+      console.error(err.message);
```
Fyorl:
```suggestion
      console.error(err);
```

Fyorl (changes requested):

less/elements.less
arbron: That's just how it handles empty operators internally in filters, but no one should be entering `{ o: "_" }` in their filters so I don't think we'll need to handle that (in fact I should probably just change `_check` to use `exact` rather than `_` as an initial value for operator).

Fyorl (approved):

PR #6822, opened 2026-03-10
[#6154] Add new Region Behavior type for condition AE application

### LINKED ISSUE #6154: [Feature] Active Effects on Regions
Would be incredible if we could add an Active Effect into a Region, to affect actors based on certain configurations (like type of creature, disposition, size, etc).

arbron:
Adds a new Region Behavior that behaves the same as core's Apply Active Effect behavior but with some extra conditions that are evaluated before applying the effect. Tokens can be filtered by disposition, size, and creature type.

Apply Effect Behavior

Closes #6154

### DIFF
--- a/module/data/region-behavior/apply-active-effect.mjs
+++ b/module/data/region-behavior/apply-active-effect.mjs
@@ -1,0 +1,112 @@
+const { DocumentUUIDField, NumberField, SetField, StringField } = foundry.data.fields;
+
+/**
+ * @import { ApplyActiveEffectRegionBehaviorSystemData } from "./_types.mjs";
+ */
+
+/**
+ * The data model for a region behavior that applies active effects to certain tokens.
+ * @extends {foundry.data.regionBehaviors.RegionBehaviorType<ApplyActiveEffectRegionBehaviorSystemData>}
+ * @mixes ApplyActiveEffectRegionBehaviorSystemData
+ */
+export default class ApplyActiveEffect5eRegionBehaviorType extends foundry.data.regionBehaviors.RegionBehaviorType {
+
+  /** @override */
+  static LOCALIZATION_PREFIXES = ["DND5E.REGIONBEHAVIORS.APPLYACTIVEEFFECT"];
+
+  /* ---------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    const dispositions = { ...foundry.applications.sheets.TokenConfig.TOKEN_DISPOSITIONS };
+    delete dispositions[CONST.TOKEN_DISPOSITIONS.SECRET];
+    return {
+      effects: new SetField(new DocumentUUIDField({ type: "ActiveEffect", nullable: false })),
+      dispositions: new SetField(new NumberField({ choices: dispositions })),
+      sizes: new SetField(new StringField({ choices: () => CONFIG.DND5E.actorSizes })),
+      types: new SetField(new StringField({ choices: () => CONFIG.DND5E.creatureTypes }))
+      // TODO: Add FiltersField for arbitrary conditions once https://github.com/foundryvtt/dnd5e/issues/6672 is done
+    };
+  }
+
+  /* ---------------------------------------- */
+  /*  Methods                                 */
+  /* ---------------------------------------- */
+
+  /**
+   * Check the conditions to decide if effects should be added to this token.
+   * @param {TokenDocument5e} token  The token to which to add the effects.
+   * @returns {boolean}
+   */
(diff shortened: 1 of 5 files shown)


Zanderaf:
Any chance for more nuanced filtering as well? Such as by creature subtype (Wood-Elf, Orc, Goblin, etc)?

roth-michael:
> Any chance for more nuanced filtering as well? Such as by creature subtype (Wood-Elf, Orc, Goblin, etc)?

Based on
```js
// TODO: Add FiltersField for arbitrary conditions once https://github.com/foundryvtt/dnd5e/issues/6672 is done
```
in the code: yeah, eventually.

guldelox:
Would this be able to handle something like Sorcerer's Careful Spell metamagic, that lets them designate specific creatures to be unaffected by an AoE?

module/data/region-behavior/apply-active-effect.mjs
```diff
@@ -0,0 +1,112 @@
...
+  static async #onTokenEnter(event) {
+    if ( !event.user.isSelf ) return;
+    const { token, movement } = event.data;
+    const actor = token.actor;
+    if ( !actor || !this.#evaluateConditions(token) ) return;
+    const resumeMovement = movement ? token.pauseMovement() : undefined;
+    const effects = await Promise.all(this.effects.map(fromUuid));
+    const toCreate = [];
+    for ( const effect of effects ) {
+      const data = effect.toObject();
```
Fyorl: `fromUuid` 3 lines up might return `null`.


```suggestion
      if ( !effect ) continue;
      const data = effect.toObject();
```

module/data/region-behavior/apply-active-effect.mjs
```diff
@@ -0,0 +1,112 @@
...
+  static async #onTokenExit(event) {
+    if ( !event.user.isSelf ) return;
+    const { token, movement } = event.data;
+    const actor = token.actor;
+    if ( !actor ) return;
+    const toDelete = [];
+    for ( const effect of actor.effects ) {
+      if ( effect.origin === this.parent.uuid ) {
+        toDelete.push(effect.id);
+      }
```
Fyorl:
```suggestion
      if ( effect.origin === this.behavior.uuid ) toDelete.push(effect.id);
```

module/data/region-behavior/apply-active-effect.mjs
```diff
@@ -0,0 +1,112 @@
...
+      toCreate.push(data);
+    }
+    if ( toCreate.length ) await actor.createEmbeddedDocuments("ActiveEffect", toCreate);
+    await resumeMovement?.();
+  }
+
+  /* ---------------------------------------- */
+
+  /**
+   * Un-apply the Active Effects when the Token exists the Region.
```
Fyorl:
```suggestion
   * Un-apply the Active Effects when the Token exits the Region.
```

module/data/region-behavior/apply-active-effect.mjs
```diff
@@ -0,0 +1,112 @@
...
+      delete data._id;
+      if ( effect.compendium ) {
+        data._stats.duplicateSource = null;
+        data._stats.compendiumSource = effect.uuid;
+      } else {
+        data._stats.duplicateSource = effect.uuid;
+        data._stats.compendiumSource = null;
+      }
+      data._stats.exportSource = null;
+      data.origin = this.parent.uuid;
```
Fyorl: Having the region be the origin makes some sense to me, but is it possible that one behaviour ends up deleting another (sibling) behaviour's AEs sometimes?


```suggestion
      data.origin = this.behavior.uuid;
```

Fyorl (changes requested):
Do we need an analogue to the core `ApplyActiveEffectRegionBehaviorType`'s `_onUpdate`? This fires if the behaviour is updated such that tokens already inside the region need to be re-evaluated as to whether the region's criteria still apply to them. 

We would factor out the effect creation and effect deletion marshalling from their respective methods so they can be called from `_onUpdate`.

module/data/region-behavior/apply-active-effect.mjs
arbron: (may want to fix the same typo in core :P)

module/data/region-behavior/apply-active-effect.mjs
arbron: Since we are within the system data `this.parent === this.behavior`, but I'll swap it anyway to make the intent clear.

Fyorl (approved):

PR #6834, opened 2026-03-16
[#6261] Add modifiers to `DamageData`

### LINKED ISSUE #6261: Add `modifiers` field to `DamageData`
Add the ability to specify modifiers on damage data that will be applied to both automatic and manual formulas. This would be useful to allow a modifier to be added without having to switch to a completely custom formula, which would make it easier to use modifiers on versatile weapons or that are applied via enchantment.

This would also open up the possibility of more globally applied modifiers, such as from a feat like Elemental Adept.

arbron:
Adds a new modifiers field for damage that is applied automatically to the formula. The `DamageData` model now automatically adds these modifiers to the formula, attaching to the first roll (if one can be found when using a custom formula).

The `DamageData#scaledFormula` supports passing in additional modifiers that are also appended. This is to support a future Rule active effect that adds modifiers to certain rolls at evaluation time (such as Elemental Adept, which would add `min2` to all damage rolls from spells that deal a certain damage type).

Closes #6261

### DIFF
--- a/module/data/shared/damage-field.mjs
+++ b/module/data/shared/damage-field.mjs
@@ -90,7 +90,7 @@ export class DamageData extends foundry.abstract.DataModel {
   _manualFormula({ modifiers }={}) {
     if ( !this.custom.formula ) return "";
     modifiers = Array.from(this.modifiers).concat(modifiers ?? []).join("");
-    return this.custom.formula.replace(/(?:\d|\))d(?:\d+\w*|\(.+\)\d*\w*)/, `$&${modifiers}`);
+    return this.custom.formula.replace(/(?:\d|\))d(?:\d+\w*|\([^)]+\)\d*\w*)/, `$&${modifiers}`);
   }
 
   /* -------------------------------------------- */
@@ -110,7 +110,8 @@ export class DamageData extends foundry.abstract.DataModel {
       case "half": increase = Math.floor(increase * .5); break;
       default: increase = 0; break;
     }
-    if ( !increase ) return this.custom.enabled ? this._manualFormula() : this._automaticFormula(0, { modifiers });
+    if ( !increase ) return this.custom.enabled
+      ? this._manualFormula({ modifiers }) : this._automaticFormula(0, { modifiers });
     let formula;
 
     // If dice count scaling, increase the count on the first die rolled
--- a/module/data/shared/damage-field.mjs
+++ b/module/data/shared/damage-field.mjs
@@ -1,4 +1,3 @@
-import simplifyRollFormula from "../../dice/simplify-roll-formula.mjs";
 import Scaling from "../../documents/scaling.mjs";
 import FormulaField from "../fields/formula-field.mjs";
 
@@ -77,7 +76,7 @@ export class DamageData extends foundry.abstract.DataModel {
     const number = (this.number ?? 0) + increase;
     if ( number && this.denomination ) {
       formula = `${number}d${this.denomination}${modifiers !== false
-        ? Array.from(this.modifiers).concat(modifiers ?? []).join("") : ""}`;
+        ? Array.from(this.modifiers).concat(...(modifiers ?? [])).join("") : ""}`;
     }
     if ( this.bonus ) formula = formula ? `${formula} + ${this.bonus}` : this.bonus;
     return formula ?? "";
@@ -94,7 +93,7 @@ export class DamageData extends foundry.abstract.DataModel {
   _manualFormula({ modifiers }={}) {
     if ( !this.custom.formula ) return "";
     if ( modifiers === false ) return this.custom.formula;
-    modifiers = Array.from(this.modifiers).concat(modifiers ?? []).join("");
+    modifiers = Array.from(this.modifiers).concat(...(modifiers ?? [])).join("");
     return this.custom.formula.replace(/(?:\d|\))?d(?:\d+\w*|\([^)]+\)\d*\w*)/, `$&${modifiers}`);
   }
 
--- a/module/data/shared/damage-field.mjs
+++ b/module/data/shared/damage-field.mjs
@@ -1,7 +1,12 @@
+import simplifyRollFormula from "../../dice/simplify-roll-formula.mjs";
 import Scaling from "../../documents/scaling.mjs";
 import FormulaField from "../fields/formula-field.mjs";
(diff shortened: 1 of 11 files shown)


module/data/shared/damage-field.mjs
```diff
@@ -60,44 +61,65 @@ export class DamageData extends foundry.abstract.DataModel {
...
+  scaledFormula(increase, { modifiers }={}) {
     if ( increase instanceof Scaling ) increase = increase.increase;
 
     switch ( this.scaling.mode ) {
       case "whole": break;
       case "half": increase = Math.floor(increase * .5); break;
       default: increase = 0; break;
     }
-    if ( !increase ) return this.formula;
+    if ( !increase ) return this.custom.enabled ? this._manualFormula() : this._automaticFormula(0, { modifiers });
```
Fyorl:
```suggestion
    if ( !increase ) return this.custom.enabled ? this._manualFormula({ modifiers }) : this._automaticFormula(0, { modifiers });
```

module/data/shared/damage-field.mjs
```diff
@@ -60,44 +61,65 @@ export class DamageData extends foundry.abstract.DataModel {
...
+   * Produce the manual formula from the `custom.formula` and `modifiers` (if possible).
+   * @param {object} [options={}]
+   * @param {Set<string>} [options.modifiers]  Additional modifiers to apply to the formula, if possible.
+   * @returns {string}
+   * @protected
+   */
+  _manualFormula({ modifiers }={}) {
+    if ( !this.custom.formula ) return "";
+    modifiers = Array.from(this.modifiers).concat(modifiers ?? []).join("");
+    return this.custom.formula.replace(/(?:\d|\))d(?:\d+\w*|\(.+\)\d*\w*)/, `$&${modifiers}`);
```
Fyorl: I don't think this will correctly match a formula like `d6 + @scale.foo`. Also, `.+` is greedy, so for something like `2d(3)kh + 1d(4)`, it will match the full expression. Need to use `[^)]+` instead.

More to the point though, we have a roll parser, so we should probably parse rolls with that parser rather than with regex. It's a bit awkward because we need to preserver data terms. We have to do something like this:
```js
const tree = foundry.dice.RollGrammar.parse(this.custom.formula);
const ast = CONFIG.Dice.parser.toAST(tree);
const terms = Roll.defaultImplementation.instantiateAST(ast);
// Find the first DiceTerm
const die = terms.find(t => t instanceof foundry.dice.terms.DiceTerm);
if ( die ) die.modifiers.push(...modifiers);
return Roll.defaultImplementation.getFormula(terms);
```

A lot slower than the regex though, so I'm inclined to stick with the regex until such a time as it becomes evident that there are some gnarly edge cases that we need the full parser to untangle.

Fyorl (changes requested):

module/data/shared/damage-field.mjs
Fyorl: This this updated regex still doesn't match something like `d6`, does it?

module/data/shared/damage-field.mjs
```diff
@@ -51,7 +52,7 @@ export class DamageData extends foundry.abstract.DataModel {
    * @type {string}
    */
   get formula() {
-    if ( this.custom.enabled ) return this.custom.formula ?? "";
+    if ( this.custom.enabled ) return this._manualFormula();
     return this._automaticFormula();
```
Fyorl, line 60: Also something I missed from my original round of feedback, was that we use `formula` for display on sheets as well. This isn't a new problem because AEs will also cause this, but it would be nice if we just had the 'pristine' formula on display, and any modifiers omitted.

module/data/shared/damage-field.mjs
arbron: Nope, missed the first sentence there :O

module/data/shared/damage-field.mjs
arbron: Added the ability to pass `modifiers: false` into the formula methods and tied that into the label generation for the damage parts. It doesn't strip out existing modifiers from custom formulas, but it does avoid adding new modifiers.

Passing through `DamageFormulaOptions` is also going to be necessary in the future for rule AEs that add modifiers so it is good to get it piped in now.

module/data/shared/damage-field.mjs
```diff
@@ -60,44 +66,65 @@ export class DamageData extends foundry.abstract.DataModel {
...
    * @protected
    */
-  _automaticFormula(increase=0) {
+  _automaticFormula(increase=0, { modifiers }={}) {
     let formula;
     const number = (this.number ?? 0) + increase;
-    if ( number && this.denomination ) formula = `${number}d${this.denomination}`;
+    if ( number && this.denomination ) {
+      formula = `${number}d${this.denomination}${modifiers !== false
+        ? Array.from(this.modifiers).concat(modifiers ?? []).join("") : ""}`;
```
Fyorl: `DamageFormulaOptions` has `modifiers` as a `Set`, so this will concat the whole set into the array I think.


```suggestion
        ? Array.from(this.modifiers).concat(...(modifiers ?? [])).join("") : ""}`;
```

module/data/shared/damage-field.mjs
```diff
@@ -60,44 +66,65 @@ export class DamageData extends foundry.abstract.DataModel {
...
+  /**
+   * Produce the manual formula from the `custom.formula` and `modifiers` (if possible).
+   * @param {DamageFormulaOptions} [options={}]  Options to configure the formula.
+   * @returns {string}
+   * @protected
+   */
+  _manualFormula({ modifiers }={}) {
+    if ( !this.custom.formula ) return "";
+    if ( modifiers === false ) return this.custom.formula;
+    modifiers = Array.from(this.modifiers).concat(modifiers ?? []).join("");
```
Fyorl:
```suggestion
    modifiers = Array.from(this.modifiers).concat(...(modifiers ?? [])).join("");
```

module/data/shared/damage-field.mjs
```diff
@@ -1,8 +1,13 @@
+import simplifyRollFormula from "../../dice/simplify-roll-formula.mjs";
```
Fyorl: Not used

Fyorl (approved):

Issue #5019, opened 2025-01-20
Feature Request: Durations on active effects measured in minutes / hours.


Zanderaf:
Probably not a big thing, but when making active effects that last, say 8 hours- it becomes a bit of a pain to constantly convert that into seconds. Not sure if a system or core software issue, but posting it anyway.

JPMeehan:
The duration field is core, but having some UI to auto convert could also be core if they're improving calendar support

Zanderaf:
Fair point. It would geniunly speed up work with making active effects with long durations, not that I dont already have the second to minute/hour conversions memorized already, just for QOL tbh.

arbron:
Going to be handled by core in V14: https://github.com/foundryvtt/foundryvtt/issues/13332

Will leave open until we determine if some system-level changes are required.

arbron:
Doesn't look like there is anything the system will have to do here.

PR #6837, opened 2026-03-20
[#5067] Automatically expire Active Effects as per their duration

### LINKED ISSUE #5067: Add Active Effect Duration Automation
As a user, I would like expired effects to stop applying to their targets.

---

_Technical Proposal_: The existing `isSuppressed` property (which is overridden from a readonly getter and replaced with a property?) is an opportunity to - without unneeded data loss - prevent expired effects from applying. There would need UI support to communicate that an effect is disabled because its duration is expired, but the actual check could be simple - some variation of the following would do plenty to make sure effects whose duration runs out fall off.

```js
if (Number.isNumeric(this.duration.remaining) && this.duration.remaining <= 0) return true;
```

roth-michael:
Closes #5067 
Reliant (obviously) on V14. Marking as a draft for now, as other efforts (like #5941) have changes that should be built on in this PR.

### Summary of behavior:
- Any effects (enchantment or base) which are expired by Core while the actor is not in combat (either the combat associated with the effect's initial application, or the currently active combat) are deleted immediately.
- Any effects (enchantment or base) which are expired by Core while the actor is in combat (either the combat associated with the effect's initial application, or the currently active combat) are marked as expired, become inactive, and are deleted upon exiting combat.
- Four "pseudo-expiries" are added to the expiry dropdown, one each for start/end of the source/target's _next turn_. When one of these is selected, duration is locked in at 1 Round, and the effect will expire according to the selected option.
- Any effects with the one of the above "pseudo-expiries" which are still active (not expired) at the time of an actor leaving combat will be deleted.
- Any _rounds-duration_ effect without a specified expiry will be given the core "Turn Start" expiry, meaning that it will expire after `n` rounds (according to the duration) at the start of the active-at-the-time-of-application Actor's turn.

### Non-user-facing changes:
- Expired Enchantments go into the "inactive" section.
- Applied Enchantments get their `start` populated the same way actor-targeting effects do.
- In order to accommodate "start/end of target's next turn" expiring in the same round for a target yet to have gone, but in the next round for a target who has already taken their turn, the 1-round duration is reduced to 0 rounds in `_preCreate` for a yet-to-go combatant.

### TODOs:
- Currently storing `flags.dnd5e.specialDuration` to track whether a still-active effect should be deleted when leaving combat. Ideally would be a value in the data model, I think. (Requires that we implement an extension of the base AE data model).
- Currently modifying the DOM directly on rendering AE Config sheet. Ideally we would extend it and have our own logic to add the pseudo-expiries & dynamically disable duration value/rounds when one is selected.
- Need to address our labeling of remaining duration in the shared active effects template. If nothing else, we should account for when `duration.remaining` is `0` (0 rounds, can still infer number of turns), but could perhaps have some `[Start|End] of [Your|the source actor's name] next turn`, if desired.


Sayshal:
> Currently modifying the DOM directly on rendering AE Config sheet. Ideally we would extend it and have our own logic to add the pseudo-expiries & dynamically disable duration value/rounds when one is selected.

Working on #5019 and ran into the same thing. The cleanest fix is replacing the `duration.seconds` form-group with a value+unit pair so you can type/select "8 hours" instead of 28800, but doing that from `onRenderActiveEffectConfig` gets ugly fast.

If a real `ActiveEffectConfig5e` extension is on the table I'd rather wait and put the duration UI inside it than ship another hook shim that just has to be migrated later alongside the pseudo-expiries.

You planning to take that on as part of this, or should I just throw up the hook version for now?

Sayshal:
I was attempting to resolve #7002 (bloodied vanishing on round change) and landed in this same area, so flagging it for this PR.

The bloodied effect is flagged `dnd5e.isTemporary: true` but has no real duration (value/expiry null). Because that flag makes isTemporary return true, core's `_prepareDuration` skips the `if ( !this.isTemporary ) duration.expired = false` short-circuit, so once combat advances it computes expired -> isSuppressed -> bloodied drops to "Unavailable" and its icon disappears (never deleted).

Whatever auto-expiry lands here should account for this.

module/documents/active-effect.mjs
```diff
@@ -569,8 +594,36 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
-      this.updateSource({ disabled: false });
+      const start = this.constructor.getEffectStart();
+      for ( const key of Object.keys(start) ) {
+        if ( data.start?.[key] !== undefined ) delete start[key]; // Prefer user-defined duration data
+      }
+      this.updateSource({ start, disabled: false });
     }
+    const actor = this.isAppliedEnchantment ? this.parent.parent : this.parent;
+    if ( !(actor instanceof Actor) || !this.start?.combat?.started ) return;
+    const {units, value, expiry} = this.duration;
```
```suggestion
    const { units, value, expiry } = this.duration;
```

module/documents/active-effect.mjs
```diff
@@ -569,8 +594,36 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+    if ( units !== "rounds" ) return;
+
+    // Default combat-duration expiry to turnStart to avoid effect expiry at round turnover
+    if ( !expiry ) {
+      this.updateSource({"duration.expiry": "turnStart"});
```
```suggestion
      this.updateSource({ "duration.expiry": "turnStart" });
```

module/documents/active-effect.mjs
```diff
@@ -569,8 +594,36 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+    }
+
+    // Convert start/end of source/target next turn expiries to start/end expiries with the proper combatant
+    // (default for source, combatant of actor it's applied to for target) and proper duration
+    if ( !this.constructor.PSEUDO_EXPIRIES.has(expiry) ) return;
+    // TODO: Use data model instead of flag
+    const effectUpdate = {"flags.dnd5e.specialDuration": expiry, duration: {}};
+    const relevantActor = expiry.startsWith("target") ? actor : fromUuidSync(this.origin).actor;
+    const combatant = this.start.combat.getCombatantsByActor(relevantActor)[0];
+    if ( combatant && (combatant.turnNumber !== null) ) effectUpdate.start = {combatant: combatant.id};
```
```suggestion
    if ( combatant && (combatant.turnNumber !== null) ) effectUpdate.start = { combatant: combatant.id };
```

module/documents/active-effect.mjs
```diff
@@ -569,8 +594,36 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+    const decreaseDuration = combatant.turnNumber > this.start.combat.turn;
+    if ( decreaseDuration ) effectUpdate.duration.value = value - 1;
```
```suggestion
    if ( combatant ) {
      const decreaseDuration = combatant.turnNumber > this.start.combat.turn;
      if ( decreaseDuration ) effectUpdate.duration.value = value - 1;
    }
```

module/documents/active-effect.mjs
```diff
@@ -569,8 +594,36 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+      this.updateSource({"duration.expiry": "turnStart"});
+      return;
+    }
+
+    // Convert start/end of source/target next turn expiries to start/end expiries with the proper combatant
+    // (default for source, combatant of actor it's applied to for target) and proper duration
+    if ( !this.constructor.PSEUDO_EXPIRIES.has(expiry) ) return;
+    // TODO: Use data model instead of flag
+    const effectUpdate = {"flags.dnd5e.specialDuration": expiry, duration: {}};
+    const relevantActor = expiry.startsWith("target") ? actor : fromUuidSync(this.origin).actor;
```
Fyorl: `this.origin` could be a lot of document types, or simply not resolve via `fromUuidSync` at all, so this line needs more consideration.

module/documents/active-effect.mjs
```diff
@@ -617,6 +668,13 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     const newEncumbrance = data.statuses?.[0];
     const name = this.name;
 
+    // If out of combat & effect expires, delete it
+    if ( game.user.isActiveGM && data.duration?.expired ) {
+      const actor = this.isAppliedEnchantment ? this.parent.parent : this.parent;
+      const combat = this.start.combat ?? game.combat;
```
Fyorl: Not sure if it's guaranteed that pre-6.0 AEs in the world will have `this.start`.


```suggestion
      const combat = this.start?.combat ?? game.combat;
```

module/documents/active-effect.mjs
```diff
@@ -617,6 +668,13 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+      if ( !combat?.getCombatantsByActor(actor).length ) return this.delete()
```
```suggestion
      if ( !combat?.getCombatantsByActor(actor).length ) return this.delete();
```

module/documents/combat.mjs
```diff
@@ -149,6 +130,40 @@ export default class Combat5e extends Combat {
...
+        }
+      }
+      if ( toDelete.length ) batchDelete.push({
+        action: "delete",
+        documentName: "ActiveEffect",
+        ids: toDelete,
+        parent: item
+      });
+    }
+    foundry.documents.modifyBatch(batchDelete);
```
Fyorl: Return this so we don't race with `_clearMovementHistoryOnExit`.

```suggestion
    return foundry.documents.modifyBatch(batchDelete);
```

Fyorl (commented):
I think this can move ahead now that #5941 is merged.

Translation from the 'pseudo-expiries' into core data is a good place to start, and the code seems correct to me. There are a lot of gnarly edge-cases though around things like inserting/removing combatants or re-rolling initiative. I think it's possible in some of those cases to end up with orphaned effects that never expire.

So to solve those we may need to register real expiry events and/or make use of the special duration fields which will move into the system model now.

Let me know if you have capacity for this feedback, otherwise I can continue work on the PR.

module/documents/active-effect.mjs
roth-michael: Addressed this by implementing `ActiveEffect5e#getSourceActor` in the latest commit.

module/documents/active-effect.mjs
```diff
@@ -160,6 +188,34 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+   */
+  getSourceActor() {
+    let sourceActor = fromUuidSync(this.origin);
+    if ( (sourceActor instanceof dnd5e.dataModels.activity.BaseActivityData)
+      || (sourceActor instanceof dnd5e.documents.ActiveEffect5e )
+      || (sourceActor instanceof dnd5e.documents.Item5e)
+    ) {
+      sourceActor = sourceActor.actor;
+    }
+    return (sourceActor instanceof dnd5e.documents.Actor5e) ? sourceActor : null;
```
```suggestion
    const origin = fromUuidSync(this.origin);
    return (origin instanceof dnd5e.documents.Actor5e) ? origin : (origin.actor || null);
```

module/documents/active-effect.mjs
```diff
@@ -425,6 +481,22 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
 
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  _prepareDuration(duration, context) {
+    super._prepareDuration();
```
```suggestion
    super._prepareDuration(duration, context);
```

module/documents/active-effect.mjs
```diff
@@ -617,6 +710,21 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
 
+  /** @inheritDoc */
+  async _preUpdate(changed, options, user) {
+    if ( await super._preUpdate(changed, options, user) === false ) return false;
+    const newExpiry = foundry.utils.getProperty(changed, "duration.expiry");
+    if ( this.constructor.PSEUDO_EXPIRIES.has(newExpiry) ) {
+      foundry.utils.mergeObject(changed, {
+        duration: {
+          value: null
+        }
```
```suggestion
        duration: { value: null },
        "system.specialDuration": newExpiry
```

Not clear if we need to do the full conversion from `_preCreate` or if the above is enough.

module/documents/active-effect.mjs
```diff
@@ -584,6 +656,27 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+      return;
+    }
+
+    // Convert start/end of source/target next turn expiries to start/end expiries with the proper combatant
+    // (default for source, combatant of actor it's applied to for target) and null duration
+    if ( !this.constructor.PSEUDO_EXPIRIES.has(expiry) ) return;
+    const effectUpdate = {"system.specialDuration": expiry, duration: {value: null}};
+    const relevantActor = expiry.startsWith("target") ? actor : this.getSourceActor();
+    const combatant = this.start.combat.getCombatantsByActor(relevantActor)[0];
+    if ( combatant && (combatant.turnNumber !== null) ) effectUpdate.start = { combatant: combatant.id };
```
Fyorl: We probably don't care but I suppose if an effect is applied to a combatant from a non-combatant, then the source becomes whoever the current combatant is because we fall through to core's logic.


```suggestion
    const combatant = this.start.combat.getCombatantsByActor(relevantActor)[0];
    if ( combatant && (combatant.turnNumber !== null) ) effectUpdate.start = { combatant: combatant.id };
    else if ( !combatant ) effectUpdate.start = { combatant: null };
```

module/documents/combat.mjs
```diff
@@ -149,6 +149,40 @@ export default class Combat5e extends Combat {
...
   }
 
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _onExit(combatant) {
+    await super._onExit(combatant);
+    const actor = combatant.actor;
+    if ( !actor ) return;
+    // TODO: Use data model instead of flag
```
Fyorl:
```suggestion
```

module/documents/combat.mjs
```diff
@@ -149,6 +149,40 @@ export default class Combat5e extends Combat {
...
+    const batchDelete = [{
+      action: "delete",
+      documentName: "ActiveEffect",
+      ids: [],
+      parent: actor
+    }];
+    for ( const effect of actor.effects ) {
+      if ( effect.duration.expired || effect.system.specialDuration ) batchDelete[0].ids.push(effect.id);
```
Fyorl:
```suggestion
      if ( effect.specialDuration ) batchDelete[0].ids.push(effect.id);
```

module/documents/combat.mjs
```diff
@@ -149,6 +149,40 @@ export default class Combat5e extends Combat {
...
+    }
+    for ( const item of actor.items ) {
+      const toDelete = [];
+      for ( const effect of item.effects ) {
+        if ( effect.isAppliedEnchantment && (effect.duration.expired || effect.system.specialDuration) ) {
```
Fyorl:
```suggestion
        if ( effect.isAppliedEnchantment && effect.specialDuration ) {
```

module/documents/active-effect.mjs
```diff
@@ -675,6 +790,29 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+  /* -------------------------------------------- */
+  /*  Expiration                                  */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  isExpiryEvent(event, context) {
+    const shouldExpire = super.isExpiryEvent(event, context);
+
+    // Catch false negatives for source-based expiries where the source is no longer in combat
+    if ( !shouldExpire && this.system.specialDuration?.startsWith("source") ) {
```
```suggestion
    if ( !shouldExpire && this.specialDuration?.startsWith("source") ) {
```

module/documents/active-effect.mjs
```diff
@@ -675,6 +790,29 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+      const combat = context.combat ?? game.combat;
+      const hasCombatant = (combat === this.start.combat) && combat.combatants.has(this.start.combatant);
+      if ( !hasCombatant ) return (this.start?.round ?? 0) < context.round;
+    }
+
+    // Otherwise if not going to expire or not "end of turn" special expiry, defer to core logic
+    if ( !shouldExpire || !this.system.specialDuration || (event !== "turnEnd") ) return shouldExpire;
```
```suggestion
    if ( !shouldExpire || !this.specialDuration || (event !== "turnEnd") ) return shouldExpire;
```

module/documents/active-effect.mjs
```diff
@@ -675,6 +790,29 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+
+  /** @inheritDoc */
+  isExpiryEvent(event, context) {
+    const shouldExpire = super.isExpiryEvent(event, context);
+
+    // Catch false negatives for source-based expiries where the source is no longer in combat
+    if ( !shouldExpire && this.system.specialDuration?.startsWith("source") ) {
+      const combat = context.combat ?? game.combat;
+      const hasCombatant = (combat === this.start.combat) && combat.combatants.has(this.start.combatant);
+      if ( !hasCombatant ) return (this.start?.round ?? 0) < context.round;
```
Fyorl: Pretty sure it's impossible to hit `isExpiryEvent` if `this.start` doesn't exist.

```suggestion
      const combat = context.combat ?? game.combat;
      const hasCombatant = (combat === this.start.combat) && combat.combatants.has(this.start.combatant);
      if ( !hasCombatant ) return this.start.round < context.round;
```

module/documents/active-effect.mjs
```diff
@@ -675,6 +790,29 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+    }
+
+    // Otherwise if not going to expire or not "end of turn" special expiry, defer to core logic
+    if ( !shouldExpire || !this.system.specialDuration || (event !== "turnEnd") ) return shouldExpire;
+
+    // If "End of Source/Target turn"-expiry effect was created on the currently-ending turn, do not yet expire
+    if ( (this.start?.round === context.round) && (this.start?.turn === context.turn) ) return false;
```
```suggestion
    if ( (this.start.round === context.round) && (this.start.turn === context.turn) ) return false;
```

module/documents/active-effect.mjs
```diff
@@ -675,6 +790,29 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     super._displayScrollingStatus(enabled);
   }
 
+  /* -------------------------------------------- */
+  /*  Expiration                                  */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  isExpiryEvent(event, context) {
```
Fyorl: `context` annotated as optional, though I think it's actually always passed. We can avoid optional chaining everywhere by using a default param.


```suggestion
  isExpiryEvent(event, context={}) {
```

Fyorl (changes requested):
Just wanted to demonstrate what fully dynamic `isExpiryEvent` handling would look like since I perhaps did not do a good job of explaining myself.

```js

_preCreate(data, options, user) {
  if ( options.keepOrigin === false ) this.updateSource({ origin: this.parent.uuid });

  const actor = this.isAppliedEnchantment ? this.parent.parent : this.parent;
  if ( !(actor instanceof Actor) || !this.start?.combat?.started ) return;
  const { expiry, units } = this.duration;

  // Default combat duration expiry to turnStart to avoid effect expiry at round change.
  if ( !expiry && (units === "rounds) ) this.updateSource({ "duration.expiry": "turnStart" });

  // Special expiries are evaluated live in isExpiryEvent so we set duration to `null` so it is always triggered.
  else if ( this.constructor.PSEUDO_EXPIRIES.has(expiry) ) this.updateSource({ "duration.value": null });
}

_preUpdate(changed, options, user) {
  if ( await super._preUpdate(changed, options, user) === false ) return false;
  const newExpiry = foundry.utils.getProperty(changed, "duration.expiry");
  if ( this.constructor.PSEUDO_EXPIRIES.has(newExpiry) ) foundry.utils.setProperty(changed, "duration.value", null);
}

isExpiryEvent(event, context={}) {
  const special = this.specialDuration;
  if ( !special ) return super.isExpiryEvent(event, context);

  // Out of combat, any time advancement expires the effect.
  if ( event === "updateWorldTime" ) return !this.actor?.inCombat;

  // Skip irrelevant events.
  const isStart = special.endsWith("Start");
  if ( event !== (isStart ? "turnStart" : "turnEnd") ) return false;

  // These expiries are only driven by the combat they were created in.
  const combat = context.combat ?? game.combat;
  if ( combat !== this.start.combat ) return false;

  // Re-derive the origin combatant.
  // If they have left combat, expire once we are past the creation round.
  const origin = special.startsWith("target") ? this.actor : this.getSourceActor();
  const [combatant] = combat.getCombatantsByActor(origin);
  if ( !combatant ) return this.start.round < context.round;

  // Only the origin's own turn edge triggers expiry.
  const originTurn = isStart ? (combat.combatant === combatant) : (combat.previous.combatantId === combatant.id);
  if ( !originTurn ) return false;

  // Wait for the origin's next turn. Skip the turn the effect was applied on.
  return (this.start.round !== context.round) || (this.start.turn !== context.turn);
}
```

It's about the same amount of code since we shift the complexity from `_preCreate` into `isExpiryEvent`. We do cover all of the edge cases I could think of, and it obviates the need for some of my comments on this PR. We don't really care about those edge cases though because you generally have to do silly things to induce them, and we're talking about 1-round duration effects here. So I don't really count it as a plus.

It does mean we actually don't need `system.specialDuration` at all. We can use `duration.expiry` directly since that's allowed to be any arbitrary system-defined value. That feels a bit neater to me.

Main downside is that we have to duplicate some core logic that we would have gotten for free otherwise (that your PR takes some pains to make sure we can use), and we have to resolve the origin Actor every time, dynamically. It should be fast since it's a `fromUuidSync` but it's not as good as baking it once in `_preCreate` aside from some of the edge cases it allows you to handle.

This fits a little bit more what I was expecting, but I'm not sure it's necessarily better, I just didn't think at the time we would be able to manage this with core durations and assumed we would always have to do the logic ourselves.

module/documents/combat.mjs
roth-michael: For this (and the similar one for enchantments) - the idea is that _any_ active effects (not just special expiry ones) that expire during combat aren't deleted mid-combat, and are purged from the actor on combat exit. Therefore we need the `effect.duration.expired` check to ensure they're removed

module/data/abstract/active-effect-data-model.mjs
```diff
@@ -3,7 +3,7 @@
  * @abstract
  */
 export default class ActiveEffectDataModel extends foundry.data.ActiveEffectTypeDataModel {
-  /* -------------------------------------------- */
+/* -------------------------------------------- */
```
Fyorl:
```suggestion
  /* -------------------------------------------- */
```

module/documents/active-effect.mjs
```diff
@@ -68,6 +68,15 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
 
   /* -------------------------------------------- */
 
+  /**
+   * System-specific "expiry" choices which do not require registration or custom expiry events, and instead
+   * are transformed into standard expiries upon application.
```
```suggestion
   * System-specific "expiry" choices which do not require registration or custom expiry events, and instead
   * are handled dynamically.
```

module/documents/active-effect.mjs
```diff
@@ -675,6 +765,40 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+  isExpiryEvent(event, context={}) {
+    const special = this.specialDuration;
+    if ( !special ) return super.isExpiryEvent(event, context);
+
+    // Out of combat, any time advancement expires the effect
+    if ( (event === "updateWorldTime") && !this.actor?.inCombat ) return true;
+
+    // Skip irrelevant events
+    const isStart = special.endsWith("Start");
+    if ( event !== (isStart ? "turnStart" : "turnEnd") ) return;
```
Fyorl: My mistake I think

```suggestion
    if ( event !== (isStart ? "turnStart" : "turnEnd") ) return false;
```

module/documents/active-effect.mjs
```diff
@@ -675,6 +765,40 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+
+    // These expiries are only driven by the combat they were created in
+    const combat = context.combat ?? game.combat;
+    if ( combat !== this.start.combat ) return false;
+
+    // Re-derive the origin combatant
```
Fyorl: I kept calling this the 'origin' but it's not really the origin. It's more like 'actor to whom the event's expiry is relative' but I don't have a good word for that.

module/documents/combat.mjs
```diff
@@ -149,6 +149,47 @@ export default class Combat5e extends Combat {
...
+      }
+    }
+    for ( const item of actor.items ) {
+      const toDelete = [];
+      for ( const effect of item.effects ) {
+        if ( effect.isAppliedEnchantment && (effect.duration.expired || effect.specialDuration) ) {
+          const parentEffect = fromUuidSync(effect.getFlag("dnd5e", "dependentOn"));
+          const applied = parentEffect?.modifiesActor || parentEffect?.isAppliedEnchantment;
+          const expired = parentEffect?.duration.expired || parentEffect?.specialDuration;
+          if ( !parentEffect || !applied || !expired ) toDelete.push(effect.id);
```
Fyorl: Maybe add a `const shouldSweep = effect =>` closure at the top of this method to factor this out. Could also be a helper I suppose but didn't feel important enough to be its own method to me.

Fyorl (approved):
Good stuff. We can leave this pending until the AE sheet is merged in if you prefer.

module/applications/components/effect-application.mjs
```diff
@@ -190,22 +188,22 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
...
   async _applyEffectToActor(effect, actor) {
     const concentration = this.chatMessage.getAssociatedActor()?.effects.get(this.chatMessage.system.concentration);
-    const item = this.chatMessage.getAssociatedItem();
-    const origin = concentration ?? (effect.inCompendium && item ? item : effect);
+    const origin = concentration ?? effect;
     if ( !game.user.isGM && !actor.isOwner ) {
       throw new Error(_loc("DND5E.EFFECT.Application.Warning.Ownership"));
     }
 
     const effectFlags = {
```
arbron, line 198: You can just do `dependentOn: concentration?.uuid` and avoid the extra lines, since an update that sets an `undefined` value is just ignored.

Fyorl (approved):

PR #6842, opened 2026-03-24
POC: General combat data (unarmed strikes), forward activity targeting other items

krbz999:
Proof of concept pull request.

Adding an embedded model for Unarmed Strike data on Creatures means that any item can reference what the current unarmed strike will do with just `@combat.unarmed + @mod` as the formula. Effects would modify this, e.g., the Monk class would would apply effects targeting `system.combat.unarmed.faces/.number`.

In addition, adds a `activity.item` string field to the Forward activity to allow it to use a different item's activity. There are some more things here that aren't accounted for, like consumption and such, but as a proof of concept it works OK. (No UI changes in this PR.)


module/data/actor/templates/creature.mjs
```diff
@@ -336,3 +344,65 @@ function makeAttackBonuses(schemaOptions={}) {
     damage: new FormulaField({required: true})
   }, schemaOptions);
 }
+
+/* -------------------------------------------- */
+
+class UnarmedStrikeDamageData extends foundry.abstract.DataModel {
```
arbron: Any reason to not use `DamageData` here? It would allow for spells like Alter Self that change your Unarmed Strike damage type.

krbz999: Forgot about it at the time.

I pushed a change extending `DamageData`, though it required making at least some of its fields `required` (or `initial: null`, whichever is the preference), otherwise the `"upgrade"` change mode will not work.

Thinking further though, it may be nice to be able to entirely ignore a damage part in either the base weapon or in an Attack/Damage/Save activity in favor of the actor's unarmed data. That'd streamline everything from the formula to the damage types and properties in all locations in one go.

PR #6848, opened 2026-03-27
Add migration progress bars for packs

arbron:
Adds progress bars for the compendium migration & refresh methods to give a better idea of the progress. Also added console logs in the refresh method to make it clear what document is currently being refreshed in case of errors.

Also adds an option to `refreshAllCompendiums` to specify a package ID and only migrate the compendiums from that package.

### DIFF
--- a/module/migration.mjs
+++ b/module/migration.mjs
@@ -286,7 +286,11 @@ export async function migrateCompendium(pack, { bypassVersionCheck=false, increm
         err.message = `Failed dnd5e system migration for document ${doc.name} in pack ${pack.collection}: ${err.message}`;
         console.error(err);
         hasErrors = true;
-        if ( strict ) throw err;
+        if ( strict ) {
+          progress.element?.classList.add("error");
+          progress.update({ format, message: "MIGRATION.DND5E.Compendium.Failed" });
+          throw err;
+        }
       }
 
       finally {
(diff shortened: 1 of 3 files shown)


module/migration.mjs
```diff
@@ -9,7 +9,7 @@ import { formatIdentifier, log } from "./utils.mjs";
  */
 export async function migrateWorld({ bypassVersionCheck=false }={}) {
   const version = game.system.version;
-  const progress = ui.notifications.info("MIGRATION.5eBegin", {
+  const progress = ui.notifications.info("MIGRATION.DND5E.World.Begin", {
```
Fyorl, line 12: Thank you, as always, for migrating these keys to the improved structure

module/migration.mjs
```diff
@@ -278,20 +285,25 @@ export async function migrateCompendium(pack, { bypassVersionCheck=false, increm
...
     log(`Migrated all ${documentName} documents from Compendium ${pack.collection}`);
   } finally {
     // Apply the original locked status for the pack
-    await pack.configure({locked: wasLocked});
+    await pack.configure({ locked: wasLocked });
     game.compendiumArt.enabled = true;
   }
+
+  progress.element?.classList.add(hasErrors ? "warning" : "success");
+  progress.update({ format, message: "MIGRATION.DND5E.Compendium.Migrate", pct: 1 });
```
Fyorl, line 310: Do we want to do anything in particular with these progress bars if we throw in strict mode? Dismiss them? Set them to error class? At the moment I believe they will just end up stuck on the screen.

module/migration.mjs
```diff
@@ -340,9 +352,11 @@ export function reparentCompendiums(from, to) {
  * @param {boolean} [options.bypassVersionCheck=false]  Bypass certain migration restrictions gated behind system
  *                                                      version stored in item stats.
  * @param {boolean} [options.migrate=true]  Also perform a system migration before refreshing.
+ * @param {string} [options.package]        Only update compendiums belonging to this package.
  */
 export async function refreshAllCompendiums(options) {
```
Fyorl: We try to access `options.package` below without optional chaining, so this will throw if no options are passed.


```suggestion
export async function refreshAllCompendiums(options={}) {
```

Fyorl (approved):

PR #6849, opened 2026-03-27
[#6846] Add fractional currency support on actor sheets, conversion

### LINKED ISSUE #6846: Allow fractional currency values on actor sheet, possibly linked to config
Currently all currencies on actor sheets are limited to integers, but there are some systems that use fractions of a denomination (such as the credits in Dark Matter from Mage Hand Press). The system should have support for this. The simplest way would be to just remove `integer: true` from the `CurrencyTemplate` data module, but it might be helpful to also allow specific denominations to specify whether they support decimals or not. Then Dark Matter could allow decimal credits while standard DnD5e would still require whole-number gold pieces.

arbron:
Allows specific currency denominations to specify a number of fractional digits they support. If set, this will allow those denominations to use decimals on the actor sheets and all rounding during currency conversions and adjustments will take those values into account.

Currently doesn't enforce this within item data because they allow fractional values already and enforcing this value now might lead to existing items having the incorrect prices.

Closes #6846

### DIFF
--- a/module/_types.mjs
+++ b/module/_types.mjs
@@ -227,6 +227,8 @@
  * @property {string} label         Localized label for the currency.
  * @property {string} abbreviation  Localized abbreviation for the currency.
  * @property {number} conversion    Number by which this currency should be multiplied to arrive at a standard value.
+ * @property {integer} [fractionalDigits=0]  Number of digits to round currency values of this denomination to. Set to
+ *                                           Infinity to prevent any rounding of the value.
  * @property {string} icon          Icon representing the currency in the interface.
  */
 
--- a/module/_types.mjs
+++ b/module/_types.mjs
@@ -227,8 +227,8 @@
  * @property {string} label         Localized label for the currency.
  * @property {string} abbreviation  Localized abbreviation for the currency.
  * @property {number} conversion    Number by which this currency should be multiplied to arrive at a standard value.
- * @property {integer} [fractionalDigits=0]  Number of digits to round currency values of this denomination to. Set to
- *                                           Infinity to prevent any rounding of the value.
+ * @property {number} [fractionalDigits=0]  Number of digits to round currency values of this denomination to. Set to
+ *                                          Infinity to prevent any rounding of the value.
  * @property {string} icon          Icon representing the currency in the interface.
  */
 
--- a/module/applications/currency-manager.mjs
+++ b/module/applications/currency-manager.mjs
@@ -1,4 +1,4 @@
-import { filteredKeys } from "../utils.mjs";
+import { filteredKeys, roundCurrency } from "../utils.mjs";
 import Award from "./award.mjs";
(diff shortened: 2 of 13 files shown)


module/applications/currency-manager.mjs
```diff
@@ -300,7 +300,9 @@ export default class CurrencyManager extends Application5e {
           updates.system.currency[denomination] += breaks * rate;
           const change = Math.min(updates.system.currency[denomination], updates.remainder);
           updates.remainder -= change;
-          updates.system.currency[denomination] -= change;
+          updates.system.currency[denomination] = roundCurrency(
+            updates.system.currency[denomination] - change, denomination
+          );
         }
         if ( !updates.remainder ) return updates;
```
Fyorl: We can't rely on exact equals comparison here when dealing with potentially fractional values now.


```suggestion
        if ( updates.remainder.almostEqual(0) ) return updates;
```

module/applications/currency-manager.mjs
Fyorl: Probably should use `roundCurrency` here too.

module/applications/currency-manager.mjs
```diff
@@ -288,10 +288,10 @@ export default class CurrencyManager extends Application5e {
...
       for ( const [denom, conversion] of currencies ) {
         const multiplier = conversion / baseConversion;
-        const deduct = Math.min(updates.system.currency[denom], Math.floor(updates.remainder * multiplier));
+        const deduct = Math.min(updates.system.currency[denom], roundCurrency(updates.remainder * multiplier, denom));
         // Handle normal deduction first.
         updates.remainder -= deduct / multiplier;
-        updates.system.currency[denom] -= deduct;
+        updates.system.currency[denom] = roundCurrency(updates.system.currency[denom] - deduct, denom);
         // If there's still a remainder, break the denomination into change.
         if ( updates.remainder && makeChange && (conversion < baseConversion) && updates.system.currency[denom] ) {
```
Fyorl:
```suggestion
        if ( !updates.remainder.almostEqual(0) && makeChange && (conversion < baseConversion) && updates.system.currency[denom] ) {
```

module/_types.mjs
```diff
@@ -227,6 +227,8 @@
  * @property {string} label         Localized label for the currency.
  * @property {string} abbreviation  Localized abbreviation for the currency.
  * @property {number} conversion    Number by which this currency should be multiplied to arrive at a standard value.
+ * @property {integer} [fractionalDigits=0]  Number of digits to round currency values of this denomination to. Set to
```
Fyorl: `Infinity` is not an integer.


```suggestion
 * @property {number} [fractionalDigits=0]  Number of digits to round currency values of this denomination to. Set to
```

Fyorl (changes requested):

Fyorl (approved):

Issue #6853, opened 2026-03-29
[BUG] Roll mode changes from Active Effects/statuses are ignored in V14 with 5.3.x


Kapuzenjoe:
I know that Foundry V14 and dnd5e 5.3.0 are still unreleased, but this also looks like a regression worth addressing before release.

On Foundry V14 with dnd5e 5.3.x, roll mode changes coming from Active Effects or status-driven system logic are not applied correctly during rolls.

On the same dnd5e 5.3.x system source, the behavior works as expected on Foundry V13.

### Steps to Reproduce

1. Create or open an actor with no manual roll mode configured.
2. Apply either:
   - a status such as `Restrained`, or
   - an Active Effect that changes a `*.roll.mode` field, for example `system.abilities.dex.save.roll.mode`
3. Roll the affected check/save/other roll.

### Expected Result

The roll should respect the roll mode change coming from the Active Effect or status.

### Actual Result

On V14, the roll mode change from the Active Effect/status is ignored.

An odd side effect is that if a manual roll mode is already configured on the actor, interactions can sometimes appear partially correct, but the AE/status-driven roll mode handling itself still seems broken.

### Possible Cause

This looks like a V14 compatibility regression in dnd5e.

`dnd5e` still appears to programmatically set roll modes through a legacy numeric `change.mode` path in `AdvantageModeField.setMode()`, while Foundry V14 field change handling now resolves changes through string-based `change.type`.

Because of that, these programmatic roll mode updates do not seem to reach the intended `add/upgrade/downgrade` handlers anymore and can fall back incorrectly, effectively leaving or resetting the final roll mode to `Normal`.

Tested without any modules on Foundry V13 351 and V14 358, using the latest current 5.3.x branch state from GitHub.

Fyorl:
> I know that Foundry V14 and dnd5e 5.3.0 are still unreleased, but this also looks like a regression worth addressing before release.

If the system is in pre-release it means we are actively soliciting bug reports like this, so thank you for the report.

PR #6854, opened 2026-03-29
[#6853] fix AdvantageModeField.setMode() for V14 AE change 

Kapuzenjoe:
Fix `AdvantageModeField.setMode()` for V14 AE change handling by setting `type` alongside legacy `mode` to address #6853.


Fyorl (approved):
Had some trouble reproducing this actually because saving throw rolls do not actually ever invoke `setMode`. However, Stealth Disadvantage does use `setMode` so I was able to verify a bug.

Issue #6857, opened 2026-03-30
Bestow Curse (2024)'s "Curse Ability" activity does not include the potential effects


MikeCook9994:
The effects that impose disadvantage on ability checks and saves for each of the particular abilities exist on the spell but aren't associated with the activity. I'm going to open a PR for this in a moment.

MikeCook9994:
#6858

PR #6858, opened 2026-03-30
Add effects and appliedEffects to bestow-curse.yml

MikeCook9994:
It seems like `appliedEffects` might be legacy? My cli doesn't produce it when writing the file to source from my compendium to verify.


MikeCook9994:
#6857

Issue #6675, opened 2026-01-30
Add missing conditionEffects for heavily encumbered and dodging


kaelad02:
The 5.1.0 release added some new `CONFIG.DND5E.conditionEffects` to handle advantage/disadvantage to ability checks and saving throws. It seems that the `heavilyEncumbered` and `dodging` statuses were not included in that update to give advantage/disadvantage with their respective rolls.

Reading the variant rules for Encumbrance, when you're heavily encumbered,

>... you have disadvantage on ability checks, attack rolls, and saving throws that use Strength, Dexterity, or Constitution.

Dodging also has something we could add:

>... you make Dexterity saving throws with Advantage. You lose these benefits if you have the Incapacitated condition or if your Speed is 0.

I know that last clause for Dodging might complicate things but maybe it's possible to handle that in the `Actor#hasConditionEffect` function. It already has a check for the `rulesVersion` setting for Exhaustion, so maybe it could do something similar for Dodging.

Kapuzenjoe:
Similar #6020

Rabulias:
There is also Squeezing (2014 PHB p.192)
A creature that is Squeezing through a smaller space:
- Must spend 1 extra foot for every foot it moves.
- Has disadvantage on attack rolls
- Has disadvantage on Dexterity saving throws
- Attack rolls against the creature have advantage while it's in the smaller space

kaelad02:
Squeezing isn't one of the status effects the system has (i.e. it's not in `CONFIG.statusEffects`)

Rabulias:
I was suggesting adding it. Should that be a separate issue?

PR #6859, opened 2026-03-30
[#6675] add advantage mode for dodging and heavilyEncumbered conditions effects

Kapuzenjoe:
A direct `movement.speed === 0` check for the actor is not reliable in `hasConditionEffect()`, because movement has not been fully resolved yet during ability roll-mode preparation. Supporting that properly would require a broader change to preparation order or moving the rule to a later stage.

Because of that, this PR uses the existing `noMovement` condition path instead. That keeps the logic consistent with `CONFIG.DND5E.conditionEffects` and avoids a more invasive change for this issue.

For the `heavilyEncumbered` part, I used the name `strengthDexterityConstitutionCheckDisadvantage`. It is explicit, though I think something shorter like `physicalCheckDisadvantage` could also be a reasonable option.

The attack disadvantage/advantage part is not addressed in this PR.


module/data/actor/templates/common.mjs
```diff
@@ -166,10 +166,16 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
       if ( !Number.isFinite(abl.max) ) abl.max = CONFIG.DND5E.maxAbilityScore;
 
       // Adjust rolling mode
-      if ( this.parent.hasConditionEffect("abilityCheckDisadvantage") ) {
+      const isPhysicalAbility = ["str", "dex", "con"].includes(id);
```
Fyorl: Try to avoid throwaway arrays in data prep:


```suggestion
      const isPhysicalAbility = CONFIG.DND5E.abilities[id]?.type === "physical";
```

module/config.mjs
```diff
@@ -3789,9 +3789,13 @@ DND5E.conditionEffects = {
   dehydrated: new Set(["dehydration"]),
   malnourished: new Set(["malnutrition"]),
   abilityCheckDisadvantage: new Set(["poisoned", "exhaustion-1"]),
+  strengthDexterityConstitutionCheckDisadvantage: new Set(["heavilyEncumbered"]),
   abilitySaveDisadvantage: new Set(["exhaustion-3"]),
+  strengthDexterityConstitutionSaveDisadvantage: new Set(["heavilyEncumbered"]),
+  strengthDexterityConstitutionAttackDisadvantage: new Set(["heavilyEncumbered"]),
```
Fyorl: Yeah let's go with your suggestion of these being `physicalCheckDisadvantage`, `physicalSaveDisadvantage`, etc. We already distinguish between mental and physical abilities scores in `CONFIG.DND5E.abilities`.

module/documents/actor/actor.mjs
```diff
@@ -535,9 +535,15 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const applyExhaustion = (level !== null) && !imms.has("exhaustion")
       && (dnd5e.settings.rulesVersion === "legacy");
     const statuses = this.statuses;
-    return props.some(k => {
+    const isActiveSource = k => {
       const l = Number(k.split("-").pop());
       return (statuses.has(k) && !imms.has(k)) || (applyExhaustion && Number.isInteger(l) && (level >= l));
+    };
+    const dodgingInactive = statuses.has("incapacitated")
+      || (CONFIG.DND5E.conditionEffects.noMovement?.some(isActiveSource) ?? false);
```
Fyorl: Prefer to frame this in the same way as exhaustion, i.e. `applyDodging`

```suggestion
    const applyDodging = !statuses.has("incapacitated")
      && !(CONFIG.DND5E.conditionEffects.noMovement?.some(isActiveSource) ?? false);
```

module/documents/actor/actor.mjs
```diff
@@ -535,9 +535,15 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    return props.some(k => {
+      if ( (key === "dexteritySaveAdvantage") && (k === "dodging") && dodgingInactive ) return false;
```
Fyorl: We don't need to check for the specific key. You lose all benefits of dodging if incapacitated/speed 0.


```suggestion
      if ( (k === "dodging") && !applyDodging ) return false;
```

Fyorl (changes requested):

Fyorl (approved):

PR #6861, opened 2026-03-30
[#6860] Add syntax for specifying limit when modifying number field

### LINKED ISSUE #6860: Add ability to specify min and max for active effects adding and subtracting on number fields
There are a number of effects in 5e that add to a value but are capped by a certain maximum. These include items such as the Ioun Stone of Strength, adding a +2 to Strength up to a maximum of 20, or the Manual of Gainful Exercise, adding a +2 to Strength up to a maximum of 30. Currently there is no way to properly implement these two features, and the SRD in fact implements them both differently:

- **Ioun Stone of Strength**: Contains a second change of `system.abilities.str.value | DOWNGRADE | 20`. This results in your strength score being lowered to `20` if you equipment the item and your strength is already higher.
- **Manual of Gainful Exercise**: Contains a second change of `system.abilities.str.max | UPGRADE | 30`. This allows your strength score to be increased up to `30` for normal ASI feats that should still be capped at `20`, but doesn't actually limit the strength bonus granted by the item to the appropriate maximum.

To fix this, we should add a special active effect value format that allows you to specify a maximum (with add) or minimum (with subtract) value for the result after the change. The two items above would end up only having a single change that handles both parts and the system would ensure that the value remains within the limit:

- `system.abilities.str.value | ADD | 2 <= 20`
- `flags.dnd5e.weaponCriticalThreshold | SUBTRACT | 1 >= 16`

arbron:
Adds a new syntax to the change value for `NumberField` that allows for specifying a maximum or minimum when adding or subtracting:

- `system.abilities.str.value | ADD | 2 <= 20`
- `flags.dnd5e.weaponCriticalThreshold | SUBTRACT | 1 >= 16`

Also supports formulas within the limit:

- `system.abilities.str.value | ADD | 2 <= 18 + @prof`

Closes #6860

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -247,24 +247,19 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     if ( (field instanceof NumberField)
       && (((change.type === "add") && change.value.includes?.("<="))
       || ((change.type === "subtract") && change.value.includes?.(">="))) ) {
-      let [value, limit] = change.value.split(/<=|>=/);
+      let [delta, limit] = change.value.split(/<=|>=/);
       try {
+        delta = simplifyBonus(field._replaceDataRefs(delta, replacementData), {}, { strict: true });
         limit = simplifyBonus(field._replaceDataRefs(limit, replacementData), {}, { strict: true });
       } catch(err) {
         const warningHeader = change.effect ? `Active Effect (${change.effect.uuid}) | ` : "";
         console.warn(`${warningHeader} "${change.type}" change to ${change.key} failed to resolve: ${err.message}`);
         return current;
       }
-      const newValue = super.applyChangeField(
-        model, { ...change, value }, { field, ...options, modifyTarget: false }
-      );
-      if ( (change.type === "add") && (newValue > limit) ) {
-        change.value = `max(0, ${value} - ${newValue - limit})`;
-      } else if ( (change.type === "subtract") && (newValue < limit) ) {
-        change.value = `min(0, ${value} + ${limit - newValue})`;
-      } else {
-        change.value = value;
-      }
+      const result = change.type === "add"
+        ? Math.max(current, Math.min(current + delta, limit))
+        : Math.min(current, Math.max(current - delta, limit));
+      return super.applyChangeField(model, { ...change, type: "override", value: result }, options);
     }
 
     // If current value is `null`, UPGRADE & DOWNGRADE should always just set the value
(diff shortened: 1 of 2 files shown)


Fyorl (commented):
Given our strategy for conditional AE application, perhaps a special syntax just for number fields is unnecessary, and this would be subsumed by the conditional AE work.

module/documents/active-effect.mjs
```diff
@@ -236,13 +236,37 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+        console.warn(`${warningHeader} "${change.type}" change to ${change.key} failed to resolve: ${err.message}`);
+        return current;
+      }
+      const newValue = super.applyChangeField(
+        model, { ...change, value }, { field, ...options, modifyTarget: false }
+      );
+      if ( (change.type === "add") && (newValue > limit) ) {
+        change.value = `max(0, ${value} - ${newValue - limit})`;
+      } else if ( (change.type === "subtract") && (newValue < limit) ) {
+        change.value = `min(0, ${value} + ${limit - newValue})`;
```
Fyorl: If `newValue < limit` then `limit - newValue > 0` so `min(0, value + (limit - newValue))` will always evaluate to `0`.

```suggestion
        change.value = `max(0, ${value} + ${limit - newValue})`;
```

module/documents/active-effect.mjs
```diff
@@ -236,13 +236,37 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
-    const { field } = options;
+    const { field, replacementData } = options;
 
     // Replace value when using string interpolation syntax
     if ( (field instanceof StringField) && (change.type === "override") && change.value?.includes?.("{}") ) {
       change.value = change.value.replace("{}", current ?? "");
     }
 
+    // Handle `<=` when adding and `>=` when subtracting from number fields
+    if ( (field instanceof NumberField)
```
Fyorl, line 247: Since this is guaranteed to be a `NumberField`, we know that `super.applyChangeField` is just going to add the delta so I don't think we need to do two passes through the entire machinery (once calling `super` here and then again at the end of the method). Something like this might be simpler:

```js
if ( (field instanceof NumberField) /* && is add/subtract && has <= / >= */ ) {
  let [delta, limit] = change.value.split(/<=|>=/);
  try {
    delta = simplifyBonus(field._replaceDataRefs(delta, replacementData), {}, { strict: true });
    limit = simplifyBonus(field._replaceDataRefs(limit, replacementData), {}, { strict: true });
  } catch ( err ) { /* warn */ return current; }
  const result = change.type === "add"
    ? Math.max(current, Math.min(current + delta, limit))
    : Math.min(current, Math.max(current - delta, limit));
  return super.applyChangeField(model, { ...change, type: "override", value: result }, options);
}
```

Fyorl (changes requested):

Fyorl (approved):

PR #6864, opened 2026-04-01
[#4698] Auto-sync actor senses to prototype token vision

### LINKED ISSUE #4698: Update prototype token with the appropriate vision mode when darkvision or other special vision traits are set on the character sheet
The prototype token should be updated with the appropriate vision mode when a feature is added to the character.
The example below is specific to darkvision, but a fix should apply to all special vision modes added to the character sheet.

Character sheet vision options:
Screenshot 2024-11-11 at 11 45 50 AM

Prototype Token (current):
Screenshot 2024-11-11 at 11 38 41 AM

Prototype Token (desired):
Screenshot 2024-11-11 at 11 42 15 AM

Sayshal:
Updates the prototype token's sight range, vision mode, and detection modes when an actor's senses change. Works for direct edits on the actor sheet, senses inherited from species items, and Active Effects that modify sense ranges.

`CONFIG.DND5E.senses` has been expanded from string labels to objects so each sense can declare a `detectionMode`, whether it `grantsSight`, and a `visionMode`:

```javascript
DND5E.senses = {
  blindsight: { label: "DND5E.SenseBlindsight", detectionMode: "blindsight" },
  darkvision: { label: "DND5E.SenseDarkvision", grantsSight: true, visionMode: "darkvision" },
  tremorsense: { label: "DND5E.SenseTremorsense", detectionMode: "feelTremor" },
  truesight: { label: "DND5E.SenseTruesight", detectionMode: "seeAll", grantsSight: true, visionMode: "darkvision" }
};
```

Modules can extend this config to add custom senses with their own detection/vision mode mappings.

Vision range is the max of all senses with `grantsSight`. Senses without it only add detection modes. So an actor with darkvision 10, truesight 30, and tremorsense 60 gets a sight range of 30 with `feelTremor` and `seeAll` detection modes.

The prototype token is kept in sync through CRUD handlers (`preCreateSenses`, `preUpdateSenses` on `AttributesFields`, and `_syncTokenSenses` on `Actor5e` called from descendant hooks). Placed tokens on the canvas also get overrides applied during `TokenDocument5e.prepareBaseData`, before V14 AE token changes run.

A few notes:
- Controlled by a new `senseVisionSync` world setting (default `true`)
- Compendium imports and explicit prototype token edits are left alone
- Non-sense detection modes the user has configured manually are preserved
- Includes a migration for existing actors (`systemVersion < 6.0.0`)

Closes #4698

### DIFF
--- a/less/elements.less
+++ b/less/elements.less
@@ -247,13 +247,3 @@ section.secret > button.reveal {
 prose-mirror {
   button.toggle:disabled { display: none; }
 }
-
-/* -------------------------------------------- */
-/*  Sense Sync Notice                           */
-/* -------------------------------------------- */
-
-.token-config .sense-sync-notice a {
-  text-decoration: underline currentcolor;
-  text-underline-offset: 2px;
-  cursor: var(--cursor-pointer);
-}
--- a/less/elements.less
+++ b/less/elements.less
@@ -247,3 +247,13 @@ section.secret > button.reveal {
 prose-mirror {
   button.toggle:disabled { display: none; }
 }
+
+/* -------------------------------------------- */
+/*  Sense Sync Notice                           */
+/* -------------------------------------------- */
+
+.token-config .sense-sync-notice a {
+  text-decoration: underline currentcolor;
+  text-underline-offset: 2px;
+  cursor: var(--cursor-pointer);
+}
--- a/module/applications/token-config.mjs
+++ b/module/applications/token-config.mjs
@@ -1,4 +1,5 @@
 import { getHumanReadableAttributeLabel } from "../utils.mjs";
+import MovementSensesConfig from "./shared/movement-senses-config.mjs";
 
 /**
  * Custom token configuration application for handling dynamic rings & resource labels.
@@ -10,6 +11,7 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
     await super._onRender(context, options);
     if ( !this.rendered ) return;
     this._prepareResourceLabels(this.element);
+    this._applySenseSyncUX(this.element);
   }
 
   /* -------------------------------------------- */
@@ -76,6 +78,51 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
       });
     }
   }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Lock the Vision tab fields that are derived from the actor's senses and surface a sync notice.
+   * @param {HTMLElement} html  The rendered markup.
+   * @protected
+   */
+  _applySenseSyncUX(html) {
+    if ( !game.settings.get("dnd5e", "senseVisionSync") ) return;
+    const actor = this.actor ?? this.object?.actor;
+    const senses = actor?.system?.attributes?.senses;
+    if ( !senses ) return;
+
+    const { sight, detectionModes } = CONFIG.Token.documentClass.computeSenseOverrides(senses);
+    if ( !sight.enabled && foundry.utils.isEmpty(detectionModes) ) return;
+
+    // Lock sight to the derived values; these bind to source, so they would otherwise show stale, editable data.
+    if ( sight.enabled ) {
+      const range = html.querySelector('[name="sight.range"]');
+      const mode = html.querySelector('[name="sight.visionMode"]');
+      if ( range ) Object.assign(range, { value: sight.range, disabled: true });
+      if ( mode ) Object.assign(mode, { value: sight.visionMode, disabled: true });
+    }
+
+    // Hide the override control on sense-derived rows; core already renders their derived range disabled.
+    const ids = sight.enabled ? [...Object.keys(detectionModes), "basicSight"] : Object.keys(detectionModes);
+    for ( const id of ids ) {
+      const pencil = html.querySelector(`[data-id="${id}"] [data-action="overrideDetectionMode"]`);
+      if ( pencil ) pencil.hidden = true;
+    }
+
+    // Surface a notice atop the Vision tab linking to the senses config.
+    const tab = html.querySelector('[data-application-part="vision"]');
+    if ( !tab || tab.querySelector(".sense-sync-notice") ) return;
+    const link = `<a data-action="editSenses">${_loc("SETTINGS.DND5E.AUTOMATION.SenseVision.Senses")}</a>`;
+    const notice = document.createElement("p");
+    notice.className = "hint sense-sync-notice";
+    notice.innerHTML = `<i class="fa-solid fa-circle-info"></i> ${
+      _loc("SETTINGS.DND5E.AUTOMATION.SenseVision.Notice", { senses: link })}`;
+    notice.querySelector("[data-action=editSenses]")?.addEventListener("click", () => {
+      new MovementSensesConfig({ document: actor, type: "senses" }).render({ force: true });
+    });
+    tab.prepend(notice);
+  }
 }
 
 /**
@@ -87,6 +134,7 @@ export class PrototypeTokenConfig5e extends foundry.applications.sheets.Prototyp
     await super._onRender(context, options);
     if ( !this.rendered ) return;
     TokenConfig5e.prototype._prepareResourceLabels.call(this, this.element);
+    TokenConfig5e.prototype._applySenseSyncUX.call(this, this.element);
   }
 
   /* -------------------------------------------- */
--- a/module/applications/token-config.mjs
+++ b/module/applications/token-config.mjs
@@ -11,7 +11,7 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
     await super._onRender(context, options);
     if ( !this.rendered ) return;
     this._prepareResourceLabels(this.element);
-    this._applySenseSyncUX(this.element);
+    this._applySenseSyncNotice(this.element);
   }
 
   /* -------------------------------------------- */
@@ -86,7 +86,7 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
    * @param {HTMLElement} html  The rendered markup.
    * @protected
    */
-  _applySenseSyncUX(html) {
+  _applySenseSyncNotice(html) {
     if ( !game.settings.get("dnd5e", "senseVisionSync") ) return;
     const actor = this.actor ?? this.object?.actor;
     const senses = actor?.system?.attributes?.senses;
@@ -134,7 +134,7 @@ export class PrototypeTokenConfig5e extends foundry.applications.sheets.Prototyp
     await super._onRender(context, options);
     if ( !this.rendered ) return;
     TokenConfig5e.prototype._prepareResourceLabels.call(this, this.element);
-    TokenConfig5e.prototype._applySenseSyncUX.call(this, this.element);
+    TokenConfig5e.prototype._applySenseSyncNotice.call(this, this.element);
   }
 
   /* -------------------------------------------- */
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2939,16 +2939,41 @@ preLocalize("restTypes", { key: "label" });
 /* -------------------------------------------- */
 
 /**
+ * Configuration data for an actor sense type.
+ *
+ * @typedef {object} SenseConfiguration
+ * @property {string} label              Localized label for the sense.
+ * @property {string} [detectionMode]    Detection mode ID to add to the token (e.g. "blindsight", "feelTremor").
+ * @property {boolean} [grantsSight]     Whether this sense grants token vision (sight.enabled & sight.range).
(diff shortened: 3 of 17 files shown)


thatlonelybugbear:
Is there a need for this to be gated behind a setting?

Sayshal:
> Is there a need for this to be gated behind a setting?

I think so, it's automation that writes to tokens, and the system already gates similar behavior behind toggles (`disableConcentration`, `disableAdvancements`, `movementAutomation`). It also gates the migration, so existing worlds with manually configured token vision don't get overwritten on upgrade.

The `DND5E.senses` config restructuring itself isn't gated — the `detectionMode`, `grantsSight`, `visionMode` metadata is always available. The setting only controls whether the system acts on it (writing to prototype tokens and applying runtime overrides on placed tokens).

Defaults to `true` so it just works unless someone explicitly turns it off. That said, I'm not strongly attached to it, so if the consensus is it's unnecessary overhead I'm fine to drop it.

roth-michael:
A high-level consideration: We might want to, instead of manually setting values on the token during base data prep, leverage `Actor#tokenActiveEffectChanges` for handling the downstream effects. If we did that, I don't think we'd need to touch the Token code at all.

Sayshal:
> A high-level consideration: We might want to, instead of manually setting values on the token during base data prep, leverage `Actor#tokenActiveEffectChanges` for handling the downstream effects. If we did that, I don't think we'd need to touch the Token code at all.

I tried following through with this in the PR. I got things functioning, but two things came up that I want to ask about before updating the PR.

First, the synthetic `changes` need a stand-in effect. `ActiveEffect5e.applyChange` opens with `change = change.effect._applyChangeShim(change)`, so a change I push into `tokenActiveEffectChanges` with no `.effect` throws on scene load. I worked around it with this:
```js
const effect = { _applyChangeShim: change => change};
changes.push({ effect, key: "sight.range", value: sight.range, type: "upgrade", priority: 0 });
```
This works, but its clearly abusing `_applyChangeShim`.

Second, and the bigger one: even with that, the change can't move what actually governs darkvision. `_prepareDetectionModes` runs in `prepareBaseData` and seeds `basicSight.range` from `sight.range`, but my injection lands in the final phase, after that. So I can set the `sight.range` field, but `basicSight` is already baked from the source value. The only reason vision looked correct in testing is that the prototype-token write had set `sight.range` on the source early enough for `basicSight` to pick it up. The `tokenActiveEffectChanges` pass changed nothing.

Am I missing a cleaner way to give these changes a real effect, or to get the range set before detection-mode prep?
Otherwise I think the prototype write is the way to go here.

Fyorl (commented):
Thank you for working on this. I understand the reported issue focuses on the prototype token, but I suspect this implementation could be a lot leaner if we went the fully-derived route.

We are already using `prepareBaseData` to somewhat derive this functionality for placed tokens. But the CRUD prototypeToken handlers actually get in the way there. If you edit an Actor and adjust its senses, its prototypeToken is modified, so when you place a token, those vision values get baked into the token's `_source`, such that removing the sense from an unlinked token has no effect - the vision has become desynced from the senses.

Similarly, because the prototypeToken data is real, persisted data, we have a problem with the migration. The new setting defaults to `true`, so as soon as a GM loads into their world, all prototypeTokens will be destructively synced to their Actor's senses, with no way to opt-out. You can't adjust the setting, because it's too late by that point.

If we rip out the prototypeToken syncing, we no longer have this migration issue, and GMs can just turn off the setting and get their pre-configured vision values back. And we no longer have the unlinked token desync issue.

I would recommend that we override `TokenDocument5e#_prepareDetectionModes`, which is probably the most appropriate place to do this, rather than `prepareBaseData`. And then we need a corresponding `_onRelatedUpdate` override to call `canvas.perception.update({ initializeVision: true })` whenever we detect a senses change on the related Actor.

The main downside is that the prototype token config is now a bit confusing because adjusting the vision settings in there while this setting is enabled will do nothing. But we have options there. I prefer that we get a clean implementation first and we can tackle the UX in a follow-up.

Sayshal:
> Thank you for working on this. I understand the reported issue focuses on the prototype token, but I suspect this implementation could be a lot leaner if we went the fully-derived route.
> 
> We are already using `prepareBaseData` to somewhat derive this functionality for placed tokens. But the CRUD prototypeToken handlers actually get in the way there. If you edit an Actor and adjust its senses, its prototypeToken is modified, so when you place a token, those vision values get baked into the token's `_source`, such that removing the sense from an unlinked token has no effect - the vision has become desynced from the senses.
> 
> Similarly, because the prototypeToken data is real, persisted data, we have a problem with the migration. The new setting defaults to `true`, so as soon as a GM loads into their world, all prototypeTokens will be destructively synced to their Actor's senses, with no way to opt-out. You can't adjust the setting, because it's too late by that point.
> 
> If we rip out the prototypeToken syncing, we no longer have this migration issue, and GMs can just turn off the setting and get their pre-configured vision values back. And we no longer have the unlinked token desync issue.
> 
> I would recommend that we override `TokenDocument5e#_prepareDetectionModes`, which is probably the most appropriate place to do this, rather than `prepareBaseData`. And then we need a corresponding `_onRelatedUpdate` override to call `canvas.perception.update({ initializeVision: true })` whenever we detect a senses change on the related Actor.
> 
> The main downside is that the prototype token config is now a bit confusing because adjusting the vision settings in there while this setting is enabled will do nothing. But we have options there. I prefer that we get a clean implementation first and we can tackle the UX in a follow-up.

While we contemplate the UX tackling, would we be fine with a small hint in the `Detection Mode` fieldset so users understand, when the setting is enabled, that these values are locked to the senses assigned to the actor?

Fyorl:
> While we contemplate the UX tackling, would we be fine with a small hint in the Detection Mode fieldset so users understand, when the setting is enabled, that these values are locked to the senses assigned to the actor?

If you want to address that in this PR rather than a follow-up, that's no problem. There are a few options, including disabling these fields entirely with a '?' button that opens the relevant settings for example. Some initial ideas I had anyway.

Sayshal:
msedge_sRJYXy8OAi

This seems clean to me.

less/elements.less
```diff
@@ -247,3 +247,13 @@ section.secret > button.reveal {
 prose-mirror {
   button.toggle:disabled { display: none; }
 }
+
+/* -------------------------------------------- */
+/*  Sense Sync Notice                           */
+/* -------------------------------------------- */
+
+.token-config .sense-sync-notice a {
```
Fyorl: Should be in `v2/apps.less`. `elements.less` is for custom elements.

module/applications/token-config.mjs
```diff
@@ -10,6 +11,7 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
     await super._onRender(context, options);
     if ( !this.rendered ) return;
     this._prepareResourceLabels(this.element);
+    this._applySenseSyncUX(this.element);
```
Fyorl: Maybe something like `_applySenseSyncNotice` to match the CSS class name.

module/config.mjs
```diff
@@ -2935,17 +2947,43 @@ preLocalize("restTypes", { key: "label" });
...
+/**
+ * Configuration data for an actor sense type.
+ *
+ * @typedef {object} SenseConfiguration
+ * @property {string} label              Localized label for the sense.
+ * @property {string} [detectionMode]    Detection mode ID to add to the token (e.g. "blindsight", "feelTremor").
+ * @property {boolean} [grantsSight]     Whether this sense grants token vision (sight.enabled & sight.range).
+ * @property {string} [visionMode]       Vision mode ID to set on the token when this sense provides sight.
+ */
+
```
arbron: Should be moved into `module/_types.mjs`

arbron (changes requested):

module/documents/token.mjs
```diff
@@ -38,6 +44,89 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+    let maxSightRange = 0;
+    let sightVisionMode = null;
+
+    for ( const [key, config] of Object.entries(CONFIG.DND5E.senses) ) {
+      const range = senses.ranges?.[key];
+      if ( !range ) continue;
+
+      if ( config.detectionMode ) {
+        detectionModes[config.detectionMode] = Math.max(detectionModes[config.detectionMode] ?? 0, range);
+      }
```
Fyorl: The `maxSightRange` logic seems correct to me. We have to set the token's vision range to highest of its (sight-granting) senses, otherwise its vision is truncated compared to what its senses allow.

However, I'm not sure about this detection mode logic. The design in this PR is that the actor's senses are the authoritative source on these detection modes. I would think the only time it would have a higher detection range is if the token already had some data before the setting was turned on. But in that case we are not honouring the concept of syncing detection modes to actor senses. And in fact, the only way to actually get it to sync would be to disable the setting, delete the detection mode override, and then turn the setting back on.

I would suggest that we directly override rather than take the max, but I wondered if maybe there was some consideration I was missing.

```suggestion
      if ( config.detectionMode ) detectionModes[config.detectionMode] = range;
```

module/documents/token.mjs
```diff
@@ -38,6 +44,89 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+      } else {
+        target.detectionModes[id] = { enabled: true, range };
+      }
+    }
+
+    if ( sight.enabled ) {
+      target.sight.enabled = true;
+      target.sight.range = sight.range;
+      target.sight.visionMode = sight.visionMode;
+    }
```
Fyorl, line 117:
```suggestion
    if ( sight.enabled ) Object.assign(target.sight, { enabled: true, range: sight.range, visionMode: sight.visionMode });
```

module/documents/token.mjs
```diff
@@ -38,6 +44,89 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+   * @param {Record<string, number>} senses.ranges  Mapping of sense keys to their range values.
+   * @param {object} target                         Target with `sight` and `detectionModes` properties.
+   */
+  static applySenseOverrides(senses, target) {
+    const { sight, detectionModes } = TokenDocument5e.computeSenseOverrides(senses);
+
+    for ( const [id, range] of Object.entries(detectionModes) ) {
+      const existing = target.detectionModes[id];
+      if ( existing ) {
+        existing.range = Math.max(existing.range ?? 0, range);
```
Fyorl: Same comment as above: I think we should override rather than take the max.

module/documents/token.mjs
```diff
@@ -232,6 +321,24 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+    const senses = this.actor?.system?.attributes?.senses;
+    if ( !senses ) return;
+
+    // Re-derive vision whenever sense-granting data changes, covering direct edits and item/effect-granted senses.
+    const overrides = TokenDocument5e.computeSenseOverrides(senses);
+    if ( foundry.utils.equals(overrides, this.#senseOverrides) ) return;
+    this.#senseOverrides = overrides;
+    if ( !this.parent?.isView ) return;
+    this.reset();
+    this.object?.initializeVisionSource();
```
Fyorl, line 331: I discovered in testing that my guidance to use `canvas.perception.update` was not strong enough. We need to fully re-initialise the vision source in order to get instant feedback when something like an AE flips a sense. I updated this to use `initializeVisionSource` instead.

module/config.mjs
```diff
@@ -2935,17 +2947,43 @@ preLocalize("restTypes", { key: "label" });
...
+  truesight: {
+    label: "DND5E.SenseTruesight",
+    detectionMode: "seeAll",
+    grantsSight: true,
+    visionMode: "darkvision"
+  }
 };
-preLocalize("senses", { sort: true });
+preLocalize("senses", { key: "label", sort: true });
+patchConfig("senses", "label", { since: "DnD5e 6.0", until: "DnD5e 6.2" });
```
Fyorl, line 2976: Added back-compat shim.

module/settings.mjs
```diff
@@ -95,6 +95,16 @@ export function registerSystemSettings() {
...
   });
 
+  // Sense-to-token vision sync
+  game.settings.register("dnd5e", "senseVisionSync", {
+    name: "SETTINGS.DND5E.AUTOMATION.SenseVision.Name",
+    hint: "SETTINGS.DND5E.AUTOMATION.SenseVision.Hint",
+    scope: "world",
+    config: true,
+    default: true,
+    type: Boolean
```
Fyorl:
```suggestion
    type: Boolean,
    onChange: () => {
      if ( canvas?.ready ) canvas.draw();
    }
```

Fyorl (changes requested):

Sayshal:
All feedback/changes done.

Fyorl (approved):

PR #6870, opened 2026-04-02
[#2667] Allow active effect durations to be set from activity duration when applying

### LINKED ISSUE #2667: Add Active Effects "per level" duration
Some DnD effects have a level based duration (1 round / level, for example) but i am unable to find a way to use a formula (@details.level or @classes.CLASS.levels) in the duration form of Active Effects.

Sayshal:
Adds an "Apply Activity Duration" option to each applied effect's Additional Settings in the activity configuration. When enabled, applying the effect from a chat card overrides the effect's stored duration with the activity's resolved duration.

This lets effects inherit formula-based durations like `@details.level` rounds from their parent activity, which is common for spells whose effect durations scale with caster level (e.g. 1 round/level).

The option is a new `applyDuration` boolean on the base `AppliedEffectField`, so it's available on all activity types that use the standard effects template (Attack, Check, Damage, Heal, Save, Utility). Activity types with their own effect handling (Enchant, Summon, etc.) are unaffected.

At application time (`EffectApplicationElement._applyEffectToActor`), the activity's `duration.getEffectData()` is called to convert the resolved formula value into Foundry's native effect duration format (`turns`, `rounds`, or `seconds`). This applies to both newly created effects and re-enabled existing effects. If the activity's duration is instantaneous or the formula didn't resolve, the override is skipped and the effect keeps its own duration.

Closes #2667

### DIFF
--- a/module/applications/components/effect-application.mjs
+++ b/module/applications/components/effect-application.mjs
@@ -196,11 +196,21 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
       }
     };
 
+    // Override effect duration with the activity's duration if configured
+    const activity = this.chatMessage.system.activity;
+    const appliedConfig = activity?.effects?.find(e => e._id === effect.id);
+    let durationOverride = {};
+    if ( appliedConfig?.applyDuration && activity.duration?.getEffectData ) {
+      const effectDuration = activity.duration.getEffectData();
+      if ( !foundry.utils.isEmpty(effectDuration) ) durationOverride = { duration: effectDuration };
+    }
+
     // Enable an existing effect on the target if it originated from this effect
     const existingEffect = actor.effects.find(e => e.origin === origin.uuid);
     if ( existingEffect ) {
       return existingEffect.update(foundry.utils.mergeObject({
         ...effect.constructor.getInitialDuration(),
+        ...durationOverride,
         disabled: false
       }, effectFlags));
     }
@@ -212,6 +222,7 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
     // Otherwise, create a new effect on the target
     const effectData = foundry.utils.mergeObject({
       ...effect.toObject(),
+      ...durationOverride,
       disabled: false,
       transfer: false,
       origin: origin.uuid
--- a/module/applications/components/effect-application.mjs
+++ b/module/applications/components/effect-application.mjs
@@ -198,8 +198,7 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
 
     // Inherit the activity's duration when the applied effect has no explicit duration of its own
     let durationOverride = {};
-    const { seconds, rounds, turns } = effect.duration;
-    if ( !seconds && !rounds && !turns ) {
+    if ( !Number.isFinite(effect.duration.value) ) {
       const effectDuration = this.chatMessage.system.activity?.duration.getEffectData();
       if ( !foundry.utils.isEmpty(effectDuration) ) durationOverride = { duration: effectDuration };
     }
--- a/module/applications/components/effect-application.mjs
+++ b/module/applications/components/effect-application.mjs
@@ -196,11 +196,20 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
       }
     };
 
+    // Inherit the activity's duration when the applied effect has no explicit duration of its own
+    let durationOverride = {};
+    const { seconds, rounds, turns } = effect.duration;
+    if ( !seconds && !rounds && !turns ) {
+      const effectDuration = this.chatMessage.system.activity?.duration.getEffectData();
+      if ( !foundry.utils.isEmpty(effectDuration) ) durationOverride = { duration: effectDuration };
+    }
+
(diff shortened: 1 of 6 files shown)


Fyorl (commented):
I feel like it was always intended that the activity duration be used as the AE duration in cases where the AE has no explicit duration.

Maybe something for @arbron to comment on, but seems like we could save ourselves a field in an already very dense dialog by just assuming that any AE attached to an activity that doesn't already have an explicit duration inherits the activity's duration.

arbron:
> Maybe something for @arbron to comment on, but seems like we could save ourselves a field in an already very dense dialog by just assuming that any AE attached to an activity that doesn't already have an explicit duration inherits the activity's duration.

Unless someone can think of a situation where you have an activity that has a duration applying an effect that shouldn't have a duration we could probably skip the extra field.

The other use of the option is to make it explicit that the duration for an effect that already has a duration should be overridden by that of the activity (perhaps a special casting that extends the duration of the effect, this way you wouldn't have to create a second effect just to give it a different duration if the original effect already has one).

Fyorl:
> The other use of the option is to make it explicit that the duration for an effect that already has a duration should be overridden by that of the activity (perhaps a special casting that extends the duration of the effect, this way you wouldn't have to create a second effect just to give it a different duration if the original effect already has one).

I know there are some features that, for example, allow you to cast a Summon spell for free, which modifies its duration. But those we don't model with AE anyway. So maybe there are some features with that form, but maybe we can go with the automatic inherit option for now since it makes the PR even smaller and makes content authoring more intuitive imo. You just set the duration in your activity and you naturally expect the AE to inherit it, rather than having to remember to also check another box.

SagaTympana:
> Unless someone can think of a situation where you have an activity that has a duration applying an effect that shouldn't have a duration we could probably skip the extra field.
one).

What about the other way around? The durations involved are extreme, but [Befuddlement](https://www.dndbeyond.com/spells/2618832-befuddlement) has an instantaneous duration but applies an effect that lasts indefinitely, with a saving throw every 30 days.

arbron:
> What about the other way around? The durations involved are extreme, but [Befuddlement](https://www.dndbeyond.com/spells/2618832-befuddlement) has an instantaneous duration but applies an effect that lasts indefinitely, with a saving throw every 30 days.

In that case the activity duration is instantaneous, so it wouldn't perform any change to the effect (which would just be set to never expire).

JPMeehan:
Wall of Light has a 10 minute duration on itself but the blinded condition it applies has a duration of 1 minute. I'm sure there are other persistent-AOEs that apply conditions with shorter durations (Hunger of Hadar does blinded while within the area, Sleet Storm applies prone)

Fyorl:
> Wall of Light has a 10 minute duration on itself but the blinded condition it applies has a duration of 1 minute. I'm sure there are other persistent-AOEs that apply conditions with shorter durations (Hunger of Hadar does blinded while within the area, Sleet Storm applies prone)

The proposal is that effects with no duration would inherit the duration of their activity. So for the cases you listed you would put an explicit duration on the effect.

Zanderaf:
Would 2014's Bestow Curse maybe be an edge case here? Or would that be a different thing entirely.
`At Higher Levels: If you cast this spell using a spell slot of 4th level or higher, the duration is concentration, up to 10 minutes. If you use a spell slot of 5th level or higher, the duration is 8 hours. If you use a spell slot of 7th level or higher, the duration is 24 hours. If you use a 9th level spell slot, the spell lasts until it is dispelled. Using a spell slot of 5th level or higher grants a duration that doesn't require concentration.`

arbron:
> Would 2014's Bestow Curse maybe be an edge case here? Or would that be a different thing entirely. `At Higher Levels: If you cast this spell using a spell slot of 4th level or higher, the duration is concentration, up to 10 minutes. If you use a spell slot of 5th level or higher, the duration is 8 hours. If you use a spell slot of 7th level or higher, the duration is 24 hours. If you use a 9th level spell slot, the spell lasts until it is dispelled. Using a spell slot of 5th level or higher grants a duration that doesn't require concentration.`

Hmm, the way I would expect that to be implemented now would be 4 different AEs with level limits in the applied effects config. That would cause issues with 5th level or higher though, because that AE wouldn't have a duration causing the activity's duration to be applied which is undesired.

For that we would really need a way to have non-formula-based duration scaling.

Sayshal:
I agree with the general consensus, ya'll hashed it out while I was out for the day. I'll rework it to auto-inherit.

module/applications/components/effect-application.mjs
```diff
@@ -196,11 +196,20 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
       }
     };
 
+    // Inherit the activity's duration when the applied effect has no explicit duration of its own
+    let durationOverride = {};
+    const { seconds, rounds, turns } = effect.duration;
+    if ( !seconds && !rounds && !turns ) {
```
Fyorl: These are deprecated properties.


```suggestion
    if ( !Number.isFinite(effect.duration.value) ) {
```

Fyorl (approved):

PR #6884, opened 2026-04-02
[#6868] Fix error thrown when updating unparented Advancements.

### LINKED ISSUE #6868: TraitAdvancement and similar can no longer be updated without parent objects (V14)
In v14 TraitAdvancements and similar cannot be updated without providing a parent object.

In 5.3.0 in v13 the following works:

```js
const advancement = new game.dnd5e.documents.advancement.TraitAdvancement();
advancement.updateSource({
      title: "Languages",
      configuration: {
        allowReplacements: true,
      },
      level: 1,
    });
```

In v14 it throws the following error:

```
foundry.mjs:13836 Uncaught TypeError: Cannot read properties of null (reading 'parent')
    at get item (pseudo-document.mjs:113:26)
    at get uuid (pseudo-document.mjs:103:22)
    at TraitAdvancement._updateDiff (foundry.mjs:13899:24)
    at TraitAdvancement.updateSource (foundry.mjs:13833:19)
    at TraitAdvancement.updateSource (pseudo-document.mjs:206:13)
    at <anonymous>:1:13
```

Fyorl:
- Closes foundryvtt/dnd5e#6868

Not sure how meaningful it is to have unparented advancements really, but it's a relatively trivial change to resolve.

### DIFF
--- a/module/documents/mixins/pseudo-document.mjs
+++ b/module/documents/mixins/pseudo-document.mjs
@@ -97,10 +97,12 @@ export default function PseudoDocumentMixin(Base) {
 
     /**
      * Globally unique identifier for this PseudoDocument.
-     * @type {string}
+     * @type {string|null}
      */
     get uuid() {
-      return `${this.item.uuid}.${this.documentName}.${this.id ?? this._source._id}`;
+      const id = this.id ?? this._source._id;
+      if ( !this.item || !id ) return null;
+      return `${this.item.uuid}.${this.documentName}.${id}`;
     }
 
     /* -------------------------------------------- */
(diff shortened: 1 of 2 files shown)


module/documents/mixins/pseudo-document.mjs
```diff
@@ -97,10 +97,12 @@ export default function PseudoDocumentMixin(Base) {
 
     /**
      * Globally unique identifier for this PseudoDocument.
-     * @type {string}
+     * @type {string|null}
```
Fyorl, line 100: Matches core behaviour when attempting to retrieve the UUID of some ephemeral doc

arbron (approved):

PR #6888, opened 2026-04-03
Remove deprecated systems, V13/14 cross-compatibility shims, fixes for resolved core issues

arbron:


Fyorl (approved):

module/data/fields/embedded-data-field.mjs
```diff
@@ -1,12 +1,12 @@
 /**
  * Version of embedded data field that properly initializes data models added via active effects.
- * TODO: Remove when we can fully rely on https://github.com/foundryvtt/foundryvtt/issues/12528
  */
 export default class EmbeddedDataField5e extends foundry.data.fields.EmbeddedDataField {
```
Fyorl, line 4: Doesn't look like this was ever exported, so we don't need to shim it. Let me know if I've missed something. Otherwise I will tentatively delete in a follow-up.

PR #6891, opened 2026-04-03
[#6839] Add placed encounter members to combat

### LINKED ISSUE #6839: Feature Request: Added functionality for placing tokens from Encounter Token
So a nice addition for the Encounter Token would be a option (or extra button) to not only place the tokens in the Encounter on to the canvas, but also add them to the Initiative Tracker, cutting down some time requiring the DM to then select the tokens and then add them manually.

This could also be extended further with also automatically rolling their initiative when added.

Aioros:
While creating combatants and rolling initiative for placed members is fairly straightforward, I think the main point of contention on the feature would be how to present the option and where it is actually stored. I opted for a client-setting-backed little dialog, shown right before the actual placement of the encounter members.

Closes #6839

### DIFF
--- a/module/data/actor/encounter.mjs
+++ b/module/data/actor/encounter.mjs
@@ -4,6 +4,7 @@ import FormulaField from "../fields/formula-field.mjs";
 const { ArrayField, DocumentUUIDField, NumberField, SchemaField } = foundry.data.fields;
 
 /**
+ * @import { EncounterPlacementSettingData } from "../settings/_types.mjs";
  * @import { EncounterActorSystemData } from "./_types.mjs";
  */
 
@@ -157,6 +158,26 @@ export default class EncounterData extends GroupTemplate {
       }
       return member;
     }))).filter(m => m.quantity.value);
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Place all members in the encounter on the current scene and return the associated token documents.
+   * @param {object} [config]                                        Configuration options for the encounter placement.
+   * @param {EncounterPlacementSettingData} [config.combatBehavior]  Combat tracker options for the placed tokens.
+   * @returns {Promise<TokenDocument[]>}
+   */
+  async placeMembers(config={}) {
+    const tokenDocuments = await super.placeMembers();
+    config.combatBehavior ??= game.settings.get("dnd5e", "encounterPlacementBehavior");
+    if ( tokenDocuments.length && config.combatBehavior !== "none" ) {
+      const combatants = await TokenDocument.implementation.createCombatants(tokenDocuments);
+      if ( config.combatBehavior === "rollInitiative" ) {
+        await game.combats.viewed.rollInitiative(combatants.map(c => c.id));
+      }
+    }
+    return tokenDocuments;
   }
 
   /* -------------------------------------------- */
--- a/module/data/actor/encounter.mjs
+++ b/module/data/actor/encounter.mjs
@@ -171,7 +171,7 @@ export default class EncounterData extends GroupTemplate {
   async placeMembers(config={}) {
     const tokenDocuments = await super.placeMembers();
     config.combatBehavior ??= game.settings.get("dnd5e", "encounterPlacementBehavior");
-    if ( tokenDocuments.length && config.combatBehavior !== "none" ) {
+    if ( tokenDocuments.length && (config.combatBehavior !== "none") ) {
       const combatants = await TokenDocument.implementation.createCombatants(tokenDocuments);
       if ( config.combatBehavior === "rollInitiative" ) {
         await game.combats.viewed.rollInitiative(combatants.map(c => c.id));
--- a/module/data/actor/templates/group.mjs
+++ b/module/data/actor/templates/group.mjs
@@ -60,7 +60,8 @@ export default class GroupTemplate extends ActorDataModel.mixin(CurrencyTemplate
   /* -------------------------------------------- */
 
   /**
-   * Place all members in the group on the current scene.
+   * Place all members in the group on the current scene and return the associated token documents.
+   * @returns {Promise<TokenDocument[]>}
    */
   async placeMembers() {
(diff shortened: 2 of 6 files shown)


module/applications/actor/encounter-placement-dialog.mjs
```diff
@@ -0,0 +1,152 @@
...
+   * @returns {Promise<EncounterPlacementSetting|null>}  Placement settings to apply.
+   */
+  static async prompt(options={}) {
+    return new Promise(resolve => {
+      const { windowId, ...dialogOptions } = options;
+      const dialog = new this(dialogOptions);
+      dialog.addEventListener("close", event =>
+        resolve(dialog.shouldPlace ? dialog.config : null)
+      , { once: true });
+      dialog.render({ force: true, ...(windowId ? { window: { windowId, detached: true } } : {}) });
```
krbz999: Using `Application#renderChild` works better cus it has the parent window carry the child with it during attachment.

Aioros: Looking into it, thanks

module/data/actor/encounter.mjs
```diff
@@ -161,6 +162,24 @@ export default class EncounterData extends GroupTemplate {
 
   /* -------------------------------------------- */
 
+  /**
+   * Place all members in the encounter on the current scene and return the associated token documents.
+   * @param {EncounterPlacementSettingData} [config]  Encounter placement configuration options.
+   * @returns {Promise<TokenDocument[]>}
+   */
+  async placeEncounterMembers(config={}) {
```
arbron: Any reason to not just wrap `placeMembers` rather than creating a new method? We could then avoid having to register a new action on the sheet. A config object could still be used to modify the default combat behavior for anyone who wants to get the default behavior of `placeMembers`.

module/data/settings/encounter-placement-setting.mjs
```diff
@@ -0,0 +1,24 @@
...
+export default class EncounterPlacementSetting extends foundry.abstract.DataModel {
+  /** @override */
+  static defineSchema() {
+    return {
+      createCombatants: new BooleanField({
+        required: true, initial: true, label: "DND5E.Group.PlaceMembersSetting.CreateCombatants"
+      }),
+      rollInitiative: new BooleanField({
+        required: true, initial: false, label: "DND5E.Group.PlaceMembersSetting.RollInitiative"
+      })
```
arbron: Since `rollInitiative: true` & `createCombatants: false` isn't a valid state, we should store the setting in such a way to make that invalid state non-representable.

arbron (changes requested):
I'd probably try to avoid the dialog solution if possible, instead having a setting in the Combat Settings config for what to do when placing actors from an encounter: "Add to Combat", "Add to Combat & Roll Initiative", "Do Nothing".

If there isn't already an existing combat, then nothing would happen.

Aioros:
The dialog was basically a personal preference coming from the way I imagined myself using the feature. I reasoned that if I had multiple encounters, and only wanted the combatants to be created for some of them, switching the setting back and forth would have felt more cumbersome, while this allows me to decide on the fly, and remember the last choice. I do trust your opinion more though, so let me prepare a different version.

> If there isn't already an existing combat, then nothing would happen.

Not sure about this part, I thought creating the combat if it doesn't exist already was kind of a given. Any reason why that would be undesirable?

Aioros:
The updated branch uses a setting as suggested (and it's quite simpler as a result). It still creates the combat if it doesn't already exist, though.

module/data/actor/templates/group.mjs
Fyorl: These early-returns need to be amended if we are changing the signature.


```suggestion
if ( !game.user.isGM || !canvas.scene ) return [];
const members = await this.getPlaceableMembers();
if ( !members.length ) return [];
```

module/data/actor/templates/group.mjs
```diff
@@ -94,6 +95,6 @@ export default class GroupTemplate extends ActorDataModel.mixin(CurrencyTemplate
       if ( minimized ) this.parent.sheet.maximize();
     }
 
-    await canvas.scene.createEmbeddedDocuments("Token", tokensData);
+    return await canvas.scene.createEmbeddedDocuments("Token", tokensData);
```
Fyorl: Already `async` so no need to `return await`.


```suggestion
    return canvas.scene.createEmbeddedDocuments("Token", tokensData);
```

module/data/actor/encounter.mjs
```diff
@@ -161,6 +162,26 @@ export default class EncounterData extends GroupTemplate {
...
+  /**
+   * Place all members in the encounter on the current scene and return the associated token documents.
+   * @param {object} [config]                                        Configuration options for the encounter placement.
+   * @param {EncounterPlacementSettingData} [config.combatBehavior]  Combat tracker options for the placed tokens.
+   * @returns {Promise<TokenDocument[]>}
+   */
+  async placeMembers(config={}) {
+    const tokenDocuments = await super.placeMembers();
+    config.combatBehavior ??= game.settings.get("dnd5e", "encounterPlacementBehavior");
+    if ( tokenDocuments.length && config.combatBehavior !== "none" ) {
```
Fyorl:
```suggestion
    if ( tokenDocuments.length && (config.combatBehavior !== "none") ) {
```

Fyorl (approved):
I think there's a very broad range of what people might want/expect from this feature:
- A dialog that prompts them every time they place members
- A dialog that only prompts on placement if there is a combat already active
- No dialog, but only places members in combat if there is a combat already active
- No dialog, automatically creates a combat if one is not already active (this PR)

I *suspect* that most people are only placing encounter members when the combat is ready to go. But equally they might be using them as a store of pre-baked encounters that they reuse when they create new scenes.

I don't have a good feel for it; I don't use encounter actors that much in my own games. Happy to go with what we have for now and iterate further.

arbron:
@Aioros Throw up a comment on the original issue for assignment

Issue #6796, opened 2026-03-02
Add UI when placing tokens and templates


arbron:
When using the token or template placement system there are several things that can be improved to help with the experience. I am thinking the system could add a small UI to the screen to indicate what placement is currently happening and the progress within it (so if you are placing 10 NPC tokens from an encounter you know how many have been placed or skipped so far).

It would also be useful for new players to have hints indicating that controls are used to place, cancel, and rotate the placement.

PR #6904, opened 2026-04-06
[#4060] Add teleport activity

### LINKED ISSUE #4060: Activity Type: Teleport
Activity type for spells like Misty Step that allows teleporting one or more actors to a certain range, bypassing normal movement (and avoiding triggering intervening regions).

Kapuzenjoe:
I used the opportunity to turn the new Foundry API from [foundryvtt/foundryvtt#13935](https://github.com/foundryvtt/foundryvtt/issues/13935), specifically `token.planMovement`, into a dedicated activity type for #4060 

Functionally, activating the activity lets the user drag the token to the desired destination. After the move is completed, the token returns to its default movement mode.

teleport

https://github.com/user-attachments/assets/c10f2824-a252-4311-a063-b7c6df293d6b

Feel free to rate the implementation. I tried to stick as closely as possible to the standard system.


module/documents/activity/teleport.mjs
```diff
@@ -0,0 +1,153 @@
...
+    const token = this.getUsageToken();
+    if ( !token?.object ) {
+      ui.notifications.warn("DND5E.ActionWarningNoToken", { localize: true });
+      return null;
+    }
+
+    if ( (typeof token.object.planMovement !== "function") || (typeof token.startMovement !== "function") ) {
+      ui.notifications.error("DND5E.TELEPORT.Warning.Unsupported", { localize: true });
+      return null;
+    }
```
arbron: Since this would go in 6.0 at the earliest which will be V14-only, no reason to protect against V13.

module/documents/activity/teleport.mjs
```diff
@@ -0,0 +1,153 @@
...
+
+    const maxDistance = this.#getSceneMaxDistance();
+    if ( !(maxDistance > 0) ) {
```
arbron:
```suggestion
    if ( maxDistance <= 0 ) {
```

module/documents/activity/teleport.mjs
```diff
@@ -0,0 +1,153 @@
...
+  /* -------------------------------------------- */
+  /*  Activation                                  */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _usageChatButtons(message) {
+    if ( !this.canPlanTeleport ) return super._usageChatButtons(message);
+    return [{
+      label: game.i18n.localize("DND5E.TELEPORT.Action.Teleport"),
+      icon: '<i class="fa-solid fa-person-from-portal" inert></i>',
```
arbron: Unfortunately not a free icon from Font Awesome, so we can't use it in the system :(. Perhaps we can use `person-walking-dashed-line-arrow-right` which is part of the free icon set.

module/documents/activity/teleport.mjs
```diff
@@ -0,0 +1,153 @@
...
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  prepareData() {
+    super.prepareData();
+    const source = this.toObject();
+    this._source.teleport = foundry.utils.mergeObject(source.teleport, this._source.teleport ?? {}, { inplace: false });
+    for ( const field of ["activation", "duration", "range", "target", "uses"] ) {
+      this._source[field] ??= source[field];
+    }
```
arbron: What's all this about?

module/data/activity/teleport-data.mjs
```diff
@@ -0,0 +1,42 @@
...
+ */
+export default class BaseTeleportActivityData extends BaseActivityData {
+  /** @inheritDoc */
+  static defineSchema() {
+    return {
+      ...super.defineSchema(),
+      teleport: new SchemaField({
+        unlimited: new BooleanField(),
+        value: new FormulaField({ deterministic: true }),
+        units: new StringField({ required: true, blank: false, initial: "ft" })
```
arbron: Should be using `defaultUnit` for localization:

```suggestion
        units: new StringField({ required: true, blank: false, initial: () => defaultUnits("length") })
```

lang/en.json
```diff
@@ -4711,6 +4711,34 @@
...
+    }
+  },
+  "Hint": "Teleport a token to a chosen destination.",
+  "SECTIONS": {
+    "Teleport": "Teleport"
+  },
+  "Title": "Teleport",
+  "Warning": {
+    "InvalidDistance": "Teleport can't be used in scenes without supported movement units.",
+    "Unsupported": "Teleport requires a compatible core version."
```
```suggestion
    "InvalidDistance": "Teleport can't be used in scenes without supported movement units."
```

icons/svg/activity/teleport.svg
arbron, line 1: You'll need to add an entry to `icons/LICENSE` to indicate the source of this artwork.

module/applications/activity/teleport-sheet.mjs
```diff
@@ -0,0 +1,49 @@
...
+      ]
+    }
+  };
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _prepareEffectContext(context, options) {
+    context = await super._prepareEffectContext(context, options);
+    context.source.teleport = this.activity._source.teleport ?? context.source.teleport;
```
arbron: Same question about this as with the similar code in `TeleportActivity#prepareData`.

arbron (changes requested):

module/documents/activity/teleport.mjs
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async planTeleport() {
+    if ( !this.canPlanTeleport ) return null;
+
+    const token = this.getUsageToken();
```
arbron: Works for spells that only teleport that token casting the spell, but can't be used for spells that teleport anyone else. Should probably swap to using `getSceneTargets` instead, perhaps with a loop in cases where multiple tokens are selected.

arbron (changes requested):

module/applications/activity/teleport-sheet.mjs
Kapuzenjoe: old unused code from the first prototype

module/documents/activity/teleport.mjs
Kapuzenjoe: old unused code from the first prototype

module/documents/activity/teleport.mjs
Kapuzenjoe: Do you mean something like `const tokens = canvas.tokens?.controlled.length ? getSceneTargets() : [];` and then a loop over all selected tokens?

The problem with that approach is that, in most cases, a player would also end up teleporting themselves, even if the teleport is only meant to affect other targets. Because right now, the function is implemented so that affected tokens immediately get `planMovement`, which means they either have to move right away or stay stuck in `planMovement`.

I could change this so that `planMovement` is not triggered immediately after the activity is used, but only after the chat button is clicked, similar to the transform activity (removing `async _triggerSubsequentActions()`).

In addition, tokens that trigger the teleport are currently centered, because that is currently my only visual indicator that the token is about to start `planMovement`. Do you think that behavior is fine, or should that be adjusted as well?

arbron: We could either remove the call to `planMovement` in `triggerSubsequentActions` or add a check so it is only called when the spell specifically targets "Self". I'd say we remove it for now and add a TODO indicating that we should revisit that behavior later, perhaps in a more holistic approach across how we handle "Self" targets across activities.

Centering seems like it is fine for now.

Kapuzenjoe: Okay, I've commented out `_triggerSubsequentActions` and added a TODO note.
That should take care of all your comments.

module/applications/activity/teleport-sheet.mjs
```diff
@@ -0,0 +1,48 @@
...
+    return context;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Prepare the tab information for the sheet.
+   * @returns {Record<string, Partial<ApplicationTab>>}
+   * @protected
+   */
```
arbron:
```suggestion
  /** @inheritDoc */
```

module/data/activity/teleport-data.mjs
```diff
@@ -0,0 +1,42 @@
+import { prepareFormulaValue } from "../../utils.mjs";
```
arbron:
```suggestion
import { defaultUnits, prepareFormulaValue } from "../../utils.mjs";
```

arbron (changes requested):

module/applications/activity/teleport-sheet.mjs
```diff
@@ -0,0 +1,44 @@
...
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _prepareEffectContext(context, options) {
+    context = await super._prepareEffectContext(context, options);
+    return context;
+  }
+
+  /* -------------------------------------------- */
+
```
arbron:
```suggestion
```

module/applications/activity/teleport-sheet.mjs
```diff
@@ -0,0 +1,44 @@
...
+  /** @inheritDoc */
+  static PARTS = {
+    ...super.PARTS,
+    effect: {
+      template: "systems/dnd5e/templates/activity/teleport-effect.hbs",
+      templates: [
+        ...super.PARTS.effect.templates,
```
arbron: Not including an active effects on this tab, so the other templates aren't necessary.

```suggestion
```

arbron (approved):
Just a bit of clean up for the sheet but otherwise its looking good to me.

A few thoughts for future improvement to make things clearer when planning the teleport:
- [ ] Display icon underneath or outline around token to be teleported
- [ ] Add some way to cancel planned movement
- [ ] Add UI to indicate movement being planned like discussed in https://github.com/foundryvtt/dnd5e/issues/6796 for placements

Kapuzenjoe:
> Just a bit of clean up for the sheet but otherwise its looking good to me.
> 
> A few thoughts for future improvement to make things clearer when planning the teleport:
> 
> * [ ]  Display icon underneath or outline around token to be teleported
> * [ ]  Add some way to cancel planned movement
> * [ ]  Add UI to indicate movement being planned like discussed in #6796 for placements

A cancel option should be fairly easy to add. Maybe with an active listener on `canvas.app.view.oncontextmenu` so a right-click cancels the currently planned teleport (similar for placing Tokens).

What kind of icon/outline did you have in mind? Maybe something like in Crucible where the selection frame rotates around the token while it's marked for teleportation?

Fyorl (approved):
Nice feature. I will get this merged into 6.0 and we can tackle any follow-up later.

> A few thoughts for future improvement to make things clearer when planning the teleport:
> 
> * [ ]  Display icon underneath or outline around token to be teleported
> * [ ]  Add some way to cancel planned movement
> * [ ]  Add UI to indicate movement being planned like discussed in #6796 for placements

For (1) we should hook into the turn marker machinery. We can pick some suitable asset to use here.

For (2) I observed that simply deselecting the controlled token is enough to cancel the teleport, which is good enough for now I think.

I also added the `useRange` property as I have to imagine that the targeting range will be the same as the teleport range most of the time (notable exceptions like Misty Step notwithstanding).

PR #6910, opened 2026-04-08
[#5192, #5489] Allow configuration of initiative grouping

### LINKED ISSUE #5192: [Suggestion] Group Initiative Rolls
So, we have the initiative grouping now, which is awesome, but I'd really like to see an option to "Roll NPCs of the same type as one". The ideal function would be simple:

Assume we have a group of 5 wolves. You add all five to the encounter and click "Roll NPCs". It rolls the first wolf, then, upon detecting subsequent wolves, it simply applies that value to them. This would make it possible to roll for wolf initiative and group all wolves together.

This should be easily implementable with a temporary key:value pair in the `Roll NPCs` and `Roll All` flows. Simply use the UUID as the key and say `If UUID exists, apply value, else continue to roll` or something similar.

### LINKED ISSUE #5489: Combat Tracker Groups Cannot Be Disabled
I'm filing this as an issue because the combat tracker groups present several problems for combat management and break other UI features.

One such feature is that usually when you hover over a token, it highlights the hovered token in the initiative tracker:

Image
(Expected behavior: the hovered spider is on initiative 10)

However, with the initiative groups, if the hovered token is inside a group, nothing is highlighted. This makes it difficult to find a token from the map in the initiative tracker:

Image

Image
(Errant behavior: the hovered tokens' initiative is not displayed)

Another issue is that expanding the group box does not expand the tracker, leading to some actors being pushed off the bottom of the tracker until you move the tracker around to force it to redraw. This is with all modules disabled:

Image

Image

Finally, the UI pattern of highlighting the group and also the active token create further visual confusion when hovering over a token with the group open:

Image
(Which token is being hovered over?)

The token groups thereby render the highlighting UI element essentially unusable for tokens within groups and in some cases also for tokens which are not in groups. In addition the groups add friction because of the need to expand the groups to access the tokens within and then redraw the tracker window to access tokens below the group which were pushed out of the display.

A setting to disable actor grouping in combat would solve this issue.

Aioros:
This change adds two settings to the Initiative section of the combat configuration:
1) Roll Once per Creature: actors of the same type (roll formula - disposition - base actor) share a single initiative roll;
2) Group Combatants: actors of the same type with the same rolled initiative are grouped in the tracker (the current default functionality, but it can now be disabled).

Closes #5192 
Closes #5489

### DIFF
--- a/module/documents/combat.mjs
+++ b/module/documents/combat.mjs
@@ -61,7 +61,25 @@ export default class Combat5e extends Combat {
 
   /** @inheritDoc */
   async rollInitiative(ids, options={}) {
-    await super.rollInitiative(ids, options);
+    const combatantsInfo = ids.reduce((info, id) => {
+      const rollGroupingKey = this.combatants.get(id).getInitiativeGroupingKey() ?? id;
+      if ( info.toBeRolled[rollGroupingKey] ) {
+        info.toBeDerived[id] = info.toBeRolled[rollGroupingKey][0];
+      } else {
+        info.toBeRolled[rollGroupingKey] ??= [];
+        info.toBeRolled[rollGroupingKey].push(id);
+      }
+      return info;
+    }, { toBeRolled: {}, toBeDerived: {} });
+
+    await super.rollInitiative(Object.values(combatantsInfo.toBeRolled).flat(), options);
+
+    const updates = Object.keys(combatantsInfo.toBeDerived).map(id => ({
+      _id: id, initiative: this.combatants.get(combatantsInfo.toBeDerived[id]).initiative
+    }));
+    if ( !updates.length ) return this;
+    await this.updateEmbeddedDocuments("Combatant", updates, { turnEvents: false });
+
     for ( const id of ids ) await this._recoverUses({ initiative: this.combatants.get(id) });
     return this;
   }
--- a/module/documents/combat.mjs
+++ b/module/documents/combat.mjs
@@ -63,16 +63,19 @@ export default class Combat5e extends Combat {
   async rollInitiative(ids, options={}) {
     const combatantsInfo = ids.reduce((info, id) => {
       const rollGroupingKey = this.combatants.get(id).getInitiativeGroupingKey() ?? id;
-      if ( info.toBeRolled[rollGroupingKey] ) {
-        info.toBeDerived[id] = info.toBeRolled[rollGroupingKey][0];
-      } else {
-        info.toBeRolled[rollGroupingKey] ??= [];
-        info.toBeRolled[rollGroupingKey].push(id);
+      let deriveFrom = null;
+      if ( dnd5e.settings.initiativeGroupRoll && !this.started ) {
+        deriveFrom = this.combatants.find(c =>
+          (c.getInitiativeGroupingKey() === rollGroupingKey) && (Number.isFinite(c.initiative))
+        )?.id ?? null;
       }
+      deriveFrom ??= info.toBeRolled[rollGroupingKey];
+      if ( deriveFrom ) info.toBeDerived[id] = deriveFrom;
+      else info.toBeRolled[rollGroupingKey] = id;
       return info;
     }, { toBeRolled: {}, toBeDerived: {} });
 
-    await super.rollInitiative(Object.values(combatantsInfo.toBeRolled).flat(), options);
+    await super.rollInitiative(Object.values(combatantsInfo.toBeRolled), options);
 
     const updates = Object.keys(combatantsInfo.toBeDerived).map(id => ({
       _id: id, initiative: this.combatants.get(combatantsInfo.toBeDerived[id]).initiative
--- a/module/documents/combatant.mjs
+++ b/module/documents/combatant.mjs
@@ -59,13 +59,27 @@ export default class Combatant5e extends Combatant {
   /* -------------------------------------------- */
 
   /**
-   * Key for the group to which this combatant should belong, or `null` if it can't be grouped.
+   * Key for the group to which this combatant should belong in the encounter tracker, or `null` if it can't be grouped.
    * @returns {string|null}
    */
   getGroupingKey() {
     if ( this.group ) return this.group.id;
-    if ( this.token?.actorLink || !this.token?.baseActor || (this.initiative === null) ) return null;
+    if ( this.token?.actorLink || !this.token?.baseActor || (this.initiative === null)
+      || !game.settings.get("dnd5e", "initiativeGroupCombatants") ) return null;
     return `${Math.floor(this.initiative).paddedString(4)}:${this.token.disposition}:${this.token.baseActor.id}`;
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Key for the group to which this combatant should belong when rolling initiative, or `null` if it can't be grouped.
+   * @returns {string|null}
+   */
+  getInitiativeGroupingKey() {
+    if ( this.group ) return this.group.id;
+    if ( this.token?.actorLink || !this.token?.baseActor
+      || !game.settings.get("dnd5e", "initiativeGroupRoll") ) return null;
+    return `${this.getInitiativeRoll().formula}:${this.token.disposition}:${this.token.baseActor.id}`;
   }
 
   /* -------------------------------------------- */
--- a/module/documents/combatant.mjs
+++ b/module/documents/combatant.mjs
@@ -65,7 +65,7 @@ export default class Combatant5e extends Combatant {
   getGroupingKey() {
     if ( this.group ) return this.group.id;
     if ( this.token?.actorLink || !this.token?.baseActor || (this.initiative === null)
-      || !game.settings.get("dnd5e", "initiativeGroupCombatants") ) return null;
+      || !dnd5e.settings.initiativeGroupCombatants ) return null;
     return `${Math.floor(this.initiative).paddedString(4)}:${this.token.disposition}:${this.token.baseActor.id}`;
   }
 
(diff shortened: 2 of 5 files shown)


module/documents/combatant.mjs
```diff
@@ -59,17 +59,31 @@ export default class Combatant5e extends Combatant {
...
   /* -------------------------------------------- */
 
+  /**
+   * Key for the group to which this combatant should belong when rolling initiative, or `null` if it can't be grouped.
+   * @returns {string|null}
+   */
+  getInitiativeGroupingKey() {
+    if ( this.group ) return this.group.id;
+    if ( this.token?.actorLink || !this.token?.baseActor
+      || !game.settings.get("dnd5e", "initiativeGroupRoll") ) return null;
```
Fyorl: Small optimisation to use `dnd5e.settings.initiativeGroupRoll` to avoid creating throwaway `Setting`s.


```suggestion
      || !dnd5e.settings.initiativeGroupRoll ) return null;
```

module/settings.mjs
```diff
@@ -408,6 +408,24 @@ export function registerSystemSettings() {
     type: Boolean
   });
 
+  game.settings.register("dnd5e", "initiativeGroupCombatants", {
```
Fyorl, line 411: I wonder if this could be a single setting with a drop-down of options? 'Never Group', 'Group - Roll Each', 'Group - Roll Once' (names TBD).

I might also suggest we set the default to 'Group - Roll Once' since that is what the PHB mandates.

module/documents/combat.mjs
```diff
@@ -61,7 +61,25 @@ export default class Combat5e extends Combat {
 
   /** @inheritDoc */
   async rollInitiative(ids, options={}) {
-    await super.rollInitiative(ids, options);
+    const combatantsInfo = ids.reduce((info, id) => {
+      const rollGroupingKey = this.combatants.get(id).getInitiativeGroupingKey() ?? id;
+      if ( info.toBeRolled[rollGroupingKey] ) {
+        info.toBeDerived[id] = info.toBeRolled[rollGroupingKey][0];
+      } else {
+        info.toBeRolled[rollGroupingKey] ??= [];
```
Fyorl: We only enter this `else` branch if `toBeRolled[rollGroupingKey]` is falsy, so this line always assigns and the `??=` is dead.


```suggestion
        info.toBeRolled[rollGroupingKey] = [];
```

module/documents/combat.mjs
```diff
@@ -61,7 +61,25 @@ export default class Combat5e extends Combat {
 
   /** @inheritDoc */
   async rollInitiative(ids, options={}) {
-    await super.rollInitiative(ids, options);
+    const combatantsInfo = ids.reduce((info, id) => {
+      const rollGroupingKey = this.combatants.get(id).getInitiativeGroupingKey() ?? id;
+      if ( info.toBeRolled[rollGroupingKey] ) {
+        info.toBeDerived[id] = info.toBeRolled[rollGroupingKey][0];
+      } else {
```
Fyorl:
```suggestion
      if ( info.toBeRolled[rollGroupingKey] ) info.toBeDerived[id] = info.toBeRolled[rollGroupingKey][0];
      else {
```

Fyorl (changes requested):
Implementation looks correct. One additional requirement that I would propose is that when you roll initiative separately for a creature, with the 'roll once' setting enabled, that we check the existing combatants for a `getGroupingKey` match, and use that initiative so it gets added to the existing group automatically.

I suppose we would only do this if combat hasn't started or if it's the first round of combat.

module/settings.mjs
Aioros: Agreed on the default change.
It could be a single dropdown, but there's also the fourth (weird) option, Roll once and don't group. While I don't personally see a lot of value in it, I have the feeling that quite a few users might want the option to follow the "roll once per creature" approach without actually grouping in the tracker, so they felt like two independent settings to me. I'll happily await a final decision and update accordingly.

Fyorl: Yeah, I considered this case as well. It would mean user would have just one entry and no enumeration of all the tokens in the group in the tracker? They would just have to keep track manually? Do I understand correctly?

I would say we should collapse the options then and leave such niche preferences to modules. Our settings space is already at a premium.

Aioros: No, "roll once and don't group" would mean that, if you have a group of goblins, only one actually rolls. The other three would take the initiative value from that result, and be listed individually (with the same initiative value) in the tracker instead of being grouped together.

jarrodlombardo: That fourth case is one I could definitely see myself using in some cases. My most likely use is "Never Group, Roll Each", but for some encounters, "Never Group, Roll Once" could work better.

Fyorl: > No, "roll once and don't group" would mean that, if you have a group of goblins, only one actually rolls. The other three would take the initiative value from that result, and be listed individually (with the same initiative value) in the tracker instead of being grouped together.

Ah got it, yes that's valid then. I guess there's nothing stopping us enumerating the full 2x2 matrix in the drop-down but by that point the 2 checkboxes are neater.

module/documents/combat.mjs
```diff
@@ -61,7 +61,28 @@ export default class Combat5e extends Combat {
...
+      return info;
+    }, { toBeRolled: {}, toBeDerived: {} });
+
+    await super.rollInitiative(Object.values(combatantsInfo.toBeRolled), options);
+
+    const updates = Object.keys(combatantsInfo.toBeDerived).map(id => ({
+      _id: id, initiative: this.combatants.get(combatantsInfo.toBeDerived[id]).initiative
+    }));
+    if ( !updates.length ) return this;
+    await this.updateEmbeddedDocuments("Combatant", updates, { turnEvents: false });
```
Fyorl, line 84: Still want to `_recoverUses` I think

```suggestion
    if ( updates.length ) await this.updateEmbeddedDocuments("Combatant", updates, { turnEvents: false });
```

Can even remove the guard entirely since `updateEmbeddedDocuments` will no-op on an empty array.

module/documents/combatant.mjs
```diff
@@ -59,17 +59,31 @@ export default class Combatant5e extends Combatant {
...
 
+  /**
+   * Key for the group to which this combatant should belong when rolling initiative, or `null` if it can't be grouped.
+   * @returns {string|null}
+   */
+  getInitiativeGroupingKey() {
+    if ( this.group ) return this.group.id;
+    if ( this.token?.actorLink || !this.token?.baseActor
+      || !dnd5e.settings.initiativeGroupRoll ) return null;
+    return `${this.getInitiativeRoll().formula}:${this.token.disposition}:${this.token.baseActor.id}`;
```
Fyorl: I guess the desire is to match `getGroupingKey`'s use of `this.initiative`, but that is solely for keeping combatants grouped in the tracker so long as they share initiative.

For this method we just want to detect 'is this creature identical to some other creature that has already rolled initiative', where 'identical' is from the wording in the PHB:
> For a group of identical creatures, the DM makes a single roll, so each member of the group has the same Initiative.

Let me know if there's some other reason we are using this, otherwise I think it's safe to remove.

Fyorl (changes requested):

module/documents/combatant.mjs
arbron: Could we use something like this to share some more of the logic:

```javascript
  /**
   * Key for the group to which this combatant should belong, or `null` if it can't be grouped.
   * @returns {string|null}
   */
  getGroupingKey() {
    if ( this.group ) return this.group.id;
    const uniqueKey = this.getUniqueKey();
    return (uniqueKey === null) || (this.initiative === null) ? null : `${Math.floor(this.initiative)}:${uniqueKey}`;
  }

  /**
   * Key used to group unique sets of actors in the combat, or `null` if they shouldn't be grouped.
   * @returns {string|null}
   */
  getUniqueKey() {
    if ( this.token?.actorLink || !this.token?.baseActor ) return null;
    return `${this.token.disposition}:${this.token.baseActor.id}`;
  }
```

Aioros: I opted to use the initiative roll formula in the grouping key because it seemed like a reasonable interpretation of the rule. I might have a group of identical enemies where half of them are surprised at the start of combat, so I would expect two groups to roll (one normal, one with disadvantage); or I dragged four unlinked Mage tokens, changed Dex to -2 for one of them and called it Clumsy Mage, and similarly expect it to roll separately.

Let me know if that seems reasonable. To me, it seems that "having the same kind of initiative" has to be a prerequisite part of "identical creatures" in this context.

Fyorl (approved):

module/documents/combatant.mjs
Fyorl: It's a reasonable interpretation. The rules are very much underspecified in this regard. I don't feel strongly, so I have approved the PR. If there's anything to be done with regards to sharing code like Jeff's suggestion, feel free to make those changes. Otherwise I will merge in due course.

arbron: Since it is a method, I suppose we can just pass in an `initiative` parameter that falls back to `this.initiative` if not present.

Aioros: Yeah, I think it can go back to being a single method (except the parameter will need to tell if it's for rolling or just for final grouping, since we need to check the appropriate setting). I'll do that and the other final fixes (probably on Monday though)

Fyorl: I added a canonical `getUniqueKey`, but I think we still need separate methods since they check different settings. If I've done something silly, feel free to just amend it.

PR #6918, opened 2026-04-10
[#6900] Improve hit point flow validation.

### LINKED ISSUE #6900: Re-implement hit point advancement error if field isn't filled in
Greetings,

I found something strange on Hit points calculation

I put 17 Constitution (equal a +3 modifier) on a test actor and it seem the Hit point does not compute properly, it take the **Con modifier** only one time
instead of 3 time for a level 3
so the toal should be 19 Hit points and not 13 (with the fighter base 10 HP + i did not roll more HP by leveling up)

I put a screenshot for better understanding

Image

is it normal, or an already known bug ?
Thank you in advance for any remarks or other.

Regards.

Fyorl:
- Closes #6900

image

### DIFF
--- a/module/applications/advancement/hit-points-flow.mjs
+++ b/module/applications/advancement/hit-points-flow.mjs
@@ -1,5 +1,6 @@
-import AdvancementFlow from "./advancement-flow-v2.mjs";
+import Advancement from "../../documents/advancement/advancement.mjs";
 import { simplifyBonus } from "../../utils.mjs";
+import AdvancementFlow from "./advancement-flow-v2.mjs";
 
 /**
  * Inline application that presents hit points selection upon level up.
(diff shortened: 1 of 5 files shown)


module/applications/advancement/hit-points-flow.mjs
```diff
@@ -91,14 +92,25 @@ export default class HitPointsFlow extends AdvancementFlow {
...
     } else if ( event.target?.name === "value" ) {
       newValue = Number.isInteger(event.target.valueAsNumber) ? event.target.valueAsNumber : null;
-    } else return;
+    } else {
+      // If neither the value input nor the useAverage checkbox is present, this is the first-class-level case where
+      // max HP is shown statically and no user input is required.
+      if ( form.querySelector("[name=value], [name=useAverage]") ) {
+        const { useAverage, value } = formData.object;
+        if ( !useAverage && !Number.isInteger(value) ) {
+          const errorType = value === undefined ? "Empty" : "Invalid";
```
arbron: In my testing `value` is `null` when blank, so we are only getting the "Invalid" warning and never the "Empty" warning.

```suggestion
          const errorType = value === null ? "Empty" : "Invalid";
```

arbron (changes requested):

arbron (approved):

PR #6927, opened 2026-04-12
Refresh bars when updating limited item uses.

DJ4ddi:
When using item uses/charges as token attributes, the bar currently doesn't refresh when the item's uses are modified. This PR aims to resolve the issue for updates from any source.


DJ4ddi:
Should have installed the linter before opening the PR :|
Formatting is compliant now.

DJ4ddi:
Reproduction for the described issue (since I was too lazy to create one to link):
- Create an actor with a limited use item (e.g. a potion).
- Create a linked token from the actor.
- Set the token's bar attribute to the item.
- Change the item's uses (e.g. with the token HUD). Observe how the bar does not refresh.
- Repeat previous steps for a linked token actor. Observe how the bar only updates once (when the token actor refreshes), then never again.

Fyorl (approved):

PR #6932, opened 2026-04-13
[#6925] Fix tile & region shape rotation in V14

### LINKED ISSUE #6925: in v14 the Rotate Area Region Behavior is bugged for Tiles and animations
Note this is for v14 only:

1. When using the Rotate Area behavior a tile gets rotated wrongly because the x and y of a TileDocument is the center of the tile. Some math in the behavior is therefore no longer needed for Tiles.
2. Currently for "updating" the placeable doc inside `static #animateFrame` it uses `Object.assign(doc, updates);`. This appears to be not accepted by v14 and requires it to be `doc.updateSource(updates);` then the animation happens smoothly again for all placeables. 
3. Apparently not necessary but maybe advisable but in the same `static #animateFrame`, is to also change the wall's assignment of `c` instead of doing `doc.c = [array]` also doing `doc.updateSource({c: [array]});` for consistency only.

EDIT: updateSource is not the solution it locks the Token document in place afterwards... investigating further

arbron:
- Ignored size when rotating tiles because they are now positioned using their center points rather than their top-left corners
- Added handling for all of the new shape types supported by regions except `GridShapeData`, because there doesn't seem to be a way to rotate those without potentially losing data
- Fixed some calls to deprecated wall edge methods in V14
- Fixed animating tile movement in V14

Closes #6925

### DIFF
--- a/module/data/region-behavior/rotate-area.mjs
+++ b/module/data/region-behavior/rotate-area.mjs
@@ -151,9 +151,7 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
       tiles: docs.tiles.map(t => calculateRotationUpdate(t)),
       tokens: docs.tokens.map(t => calculateRotationUpdate(t)),
       lights: docs.lights.map(l => calculateRotationUpdate(l)),
-      regions: docs.regions.map(r =>
-        RotateAreaRegionBehaviorType.#rotateRegionShapes(r, angle, radians, pivot, this.scene.grid)
-      ),
+      regions: docs.regions.map(r => RotateAreaRegionBehaviorType.#rotateRegionShapes(r, angle, radians, pivot)),
       sounds: docs.sounds.map(s => calculateRotationUpdate(s)),
       walls: docs.walls.map(wall => {
         const first = RotateAreaRegionBehaviorType.#calculatePosition(radians, pivot, {
@@ -182,16 +180,24 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
       this.scene.updateEmbeddedDocuments("AmbientSound", updates.sounds),
       this.scene.updateEmbeddedDocuments("Region", updates.regions),
       this.scene.updateEmbeddedDocuments("Tile", updates.tiles),
-      this.scene.updateEmbeddedDocuments("Token", updates.tokens, {
-        animate: false,
-        movement: updates.tokens.reduce((obj, { _id }) => {
-          obj[_id] = {
-            constrainOptions: { ignoreWalls: true, ignoreCost: true },
-            showRuler: false
-          };
+      game.release.generation >= 14
+        ? this.scene.moveTokens(updates.tokens.reduce((obj, { _id: id, ...destination }) => {
+          obj[id] = { destination };
           return obj;
-        }, {})
-      }),
+        }, {}), {
+          constrainOptions: { ignoreWalls: true, ignoreCost: true },
+          animate: false
+        })
+        : this.scene.updateEmbeddedDocuments("Token", updates.tokens, {
+          animate: false,
+          movement: updates.tokens.reduce((obj, { _id }) => {
+            obj[_id] = {
+              constrainOptions: { ignoreWalls: true, ignoreCost: true },
+              showRuler: false
+            };
+            return obj;
+          }, {})
+        }),
       this.scene.updateEmbeddedDocuments("Wall", updates.walls)
     ]);
   }
@@ -291,23 +297,30 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
    * @param {Degrees} angle    Rotation amount in degrees.
    * @param {Radians} radians  Rotation amount in radians.
    * @param {Point} pivot      Center point for the rotation.
-   * @param {Grid} [grid]      Grid for the current scene.
    * @returns {object}         Update data for the region.
    */
-  static #rotateRegionShapes(region, angle, radians, pivot, grid=Scene.defaultGrid) {
+  static #rotateRegionShapes(region, angle, radians, pivot) {
+    if ( game.release.generation >= 14 ) return {
+      _id: region.id,
+      shapes: region.shapes.map(shape => {
+        const clone = shape.clone();
+        clone.rotate(angle, { pivot });
+        return clone.toObject();
+      })
+    };
+
     const shapes = region.toObject().shapes;
-    for ( let shape of shapes ) {
-      if ( shape.type === foundry.data.EmanationShapeData?.TYPE ) shape = shape.base;
-      let { x, y, width, height } = shape;
+    for ( const shape of shapes ) {
+      const { x, y, width, height } = shape;
       switch ( shape.type ) {
         case foundry.data.RectangleShapeData.TYPE:
-          if ( game.release.generation < 14 ) {
-            Object.assign(shape, this.#calculatePosition(
-              radians, pivot, { x: x + (width / 2), y: y + (height / 2) }, { width, height }
-            ));
-          } else {
-            Object.assign(shape, this.#calculatePosition(radians, pivot, { x, y }));
-          }
+          Object.assign(shape, this.#calculatePosition(
+            radians, pivot, { x: x + (width / 2), y: y + (height / 2) }, { width, height }
+          ));
+          break;
+        case foundry.data.CircleShapeData.TYPE:
+        case foundry.data.EllipseShapeData.TYPE:
+          Object.assign(shape, this.#calculatePosition(radians, pivot, { x, y }));
           break;
         case foundry.data.PolygonShapeData.TYPE:
           const iterator = Iterator.from(shape.points);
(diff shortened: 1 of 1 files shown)


module/data/region-behavior/rotate-area.mjs
```diff
@@ -312,8 +318,26 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
...
+          }
+          width *= grid.sizeX;
+          height *= grid.sizeY;
+          Object.assign(shape, this.#calculatePosition(
+            radians, pivot, { x: x + (width / 2), y: y + (height / 2) }, { width, height }
+          ));
+          break;
+        case foundry.data.GridShapeData?.TYPE:
+          // Not sure it makes sense to rotate these, may lead to situations where rotation's isn't fully reversible
+          return;
```
Fyorl:
```suggestion
          continue;
```

module/data/region-behavior/rotate-area.mjs
```diff
@@ -312,8 +318,26 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
...
+            else height = (0.75 * Math.floor(height)) + (0.5 * (height % 1)) + 0.25;
+          }
+          width *= grid.sizeX;
+          height *= grid.sizeY;
+          Object.assign(shape, this.#calculatePosition(
+            radians, pivot, { x: x + (width / 2), y: y + (height / 2) }, { width, height }
+          ));
+          break;
+        case foundry.data.GridShapeData?.TYPE:
+          // Not sure it makes sense to rotate these, may lead to situations where rotation's isn't fully reversible
```
Fyorl:
```suggestion
          // Not sure it makes sense to rotate these, may lead to situations where rotation isn't fully reversible
```

module/data/region-behavior/rotate-area.mjs
```diff
@@ -151,7 +151,9 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
       tiles: docs.tiles.map(t => calculateRotationUpdate(t)),
       tokens: docs.tokens.map(t => calculateRotationUpdate(t)),
       lights: docs.lights.map(l => calculateRotationUpdate(l)),
-      regions: docs.regions.map(r => RotateAreaRegionBehaviorType.#rotateRegionShapes(r, angle, radians, pivot)),
+      regions: docs.regions.map(r =>
+        RotateAreaRegionBehaviorType.#rotateRegionShapes(r, angle, radians, pivot, this.scene.grid)
+      ),
```
dev7355608:
```suggestion
      regions: docs.regions.map(r => ({ _id: r.id, shapes: r.shapes.map(
        s => s.clone().rotate(angle, {pivot}).toObject()) })),
```

V14 only.

module/data/region-behavior/rotate-area.mjs
dev7355608, line 357:
```suggestion
  static #shapeCenter(shape) {
    if ( game.release.generation >= 14 ) return shape.origin;
```

dev7355608:
```suggestion
      game.release.generation >= 14
        ? this.scene.moveTokens(updates.tokens.reduce((obj, { _id: id, ...destination }) => {
          obj[id] = {destination};
          return obj;
        }, {}), {
          constrainOptions: { ignoreWalls: true, ignoreCost: true },
          showRuler: false,
          animate: false
        })
        : this.scene.updateEmbeddedDocuments("Token", updates.tokens, {
          animate: false,
          movement: updates.tokens.reduce((obj, { _id }) => {
            obj[_id] = {
              constrainOptions: { ignoreWalls: true, ignoreCost: true },
              showRuler: false
            };
            return obj;
          }, {})
        }),
```

The have a better way of moving multiple tokens in V14.

`showRuler: false` should be the default for `"api"` `method` movement and therefore unnecessary to pass.

module/data/region-behavior/rotate-area.mjs
```diff
@@ -274,7 +276,9 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
    */
   static #placeableSize(doc) {
     if ( !(doc instanceof foundry.abstract.Document) ) doc = doc.document;
-    if ( doc instanceof TileDocument ) return { width: doc.width, height: doc.height };
+    if ( (doc instanceof TileDocument) && (game.release.generation < 14) ) {
+      return { width: doc.width, height: doc.height };
+    }
```
dev7355608, line 297: Is this supposed to return `{ width: 0, height: 0 }` for tiles in V14?

module/data/region-behavior/rotate-area.mjs
dev7355608: This is illegal. Interesting that this didn't cause any issues considering `_onAnimationUpdate` takes a second required parameter: `context: TokenAnimationContext`. We should be using `Token#animate`.

module/data/region-behavior/rotate-area.mjs
```diff
@@ -429,15 +453,18 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
...
           doc.object.renderFlags.set({ refreshLine: true });
           if ( game.settings.get("core", "visionAnimation") ) {
-            doc.object.initializeEdge();
+            if ( game.release.generation < 14 ) doc.object.initializeEdge();
+            else doc.initializeEdge();
             canvas.perception.update({
-              refreshEdges: true, initializeLighting: true, initializeVision: true, initializeSounds: true
+              refreshEdges: game.release.generation < 14 ? true : undefined,
+              initializeLighting: true, initializeVision: true, initializeSounds: true
             });
```
dev7355608: This should not be needed in V14, because `WallDocument#initializeEdge` should take care of that, and token animation should set the necessary perception flags.

dev7355608 (changes requested):
This behavior should probably be its own module, I think. It doesn't seem like it essential to play dnd5e or dnd5e specific to me.

module/data/region-behavior/rotate-area.mjs
arbron: Yes, we only use that to offset the rotation center for placeables that use the top-left corner for position rather than the center. Since the origin of tiles changed in V14 we no longer need to make that adjustment.

module/data/region-behavior/rotate-area.mjs
arbron: Probably working because `_onAnimationUpdate` only checks the context if no transformation occurred, so it never falls into that branch.

~~I've written a version that calls `Token#animate` for each token, it works fine except there is a single frame at the end of the animation where the token jumps back to its original position before returning to where it should be:~~

Ignore that, was using the wrong position for the `to`.

Freeze023:
just tested this, works great except for lights that don't have their emission angle set to 360, those lights get moved by an offset.

module/data/region-behavior/rotate-area.mjs
```diff
@@ -180,16 +180,24 @@ export default class RotateAreaRegionBehaviorType extends foundry.data.regionBeh
...
+          movement: updates.tokens.reduce((obj, { _id }) => {
+            obj[_id] = {
+              constrainOptions: { ignoreWalls: true, ignoreCost: true },
+              showRuler: false
+            };
+            return obj;
+          }, {})
+        }),
       this.scene.updateEmbeddedDocuments("Wall", updates.walls)
     ]);
```
dev7355608, line 194: I noticed that after a token is rotated with the behavior, the token is not necessarily in the region anymore. The changes below fixed that (we should make updates like these batches anyway), but the token exits and immediately re-enters, so that might cause some issues with other behaviors on the region potentially. But there's nothing the system can do to fix this.

I also moved the token updates before the region updates in the V13 case, but that may not actually help. We could await the token updates before we await the reset of the updates, but that might cause issues visually. Not sure if it's worth attempting to fix the V13 case.

```js
    // Update all rotated documents
    if ( game.release.generation >= 14 ) {
      await foundry.documents.modifyBatch([
        { action: "update", documentName: "RegionBehavior",
          updates: [{ _id: this.behavior.id, "system.status.rotating": false }], parent: this.region },
        // Tokens must be updated before the region so they re-enter the rotated region after they briefly leave it
        // when moved to the new position while the region shapes have not been updated yet
        { action: "update", documentName: "Token", updates: updates.tokens, parent: this.scene,
          animate: false, constrainOptions: { ignoreWalls: true, ignoreCost: true } },
        { action: "update", documentName: "AmbientLight", updates: updates.lights, parent: this.scene },
        { action: "update", documentName: "AmbientSound", updates: updates.sounds, parent: this.scene },
        { action: "update", documentName: "Region", updates: updates.regions, parent: this.scene },
        { action: "update", documentName: "Tile", updates: updates.tiles, parent: this.scene },
        { action: "update", documentName: "Wall", updates: updates.walls, parent: this.scene }
      ]);
    } else {
      await Promise.all([
        this.parent.update({ "system.status.rotating": false }),
        this.scene.updateEmbeddedDocuments("Token", updates.tokens, {
          animate: false,
          movement: updates.tokens.reduce((obj, { _id }) => {
            obj[_id] = {
              constrainOptions: { ignoreWalls: true, ignoreCost: true },
              showRuler: false
            };
            return obj;
          }, {})
        }),
        this.scene.updateEmbeddedDocuments("AmbientLight", updates.lights),
        this.scene.updateEmbeddedDocuments("AmbientSound", updates.sounds),
        this.scene.updateEmbeddedDocuments("Region", updates.regions),
        this.scene.updateEmbeddedDocuments("Tile", updates.tiles),
        this.scene.updateEmbeddedDocuments("Wall", updates.walls)
      ]);
    }
```

dev7355608 (changes requested):

dev7355608 (approved):

Issue #6923, opened 2026-04-10
Add system-specific adventure importer with standard pre/post actions


arbron:
Our first party adventures include a bunch of code for customizing the import process and offering customizable options for players. By moving these customizations into the system it would reduce a large amount of code in these premium modules and make these features available to developers of third-party adventures.

<img width="936" height="1040" alt="Image" src="https://github.com/user-attachments/assets/5234e79c-22c8-48a3-ae9b-9e441b33b728" />

This also includes integration with the new quick start system in V14, allowing for customization during the first launch of the world.

<img width="436" height="247" alt="Image" src="https://github.com/user-attachments/assets/966ef0d9-1865-415b-91f5-a1c2fc7ccd1c" />

### Actions
- **Activate scene**: Used to open a specific scene after the adventure is imported
- **Display journal**: Opens the first journal in the adventure after import
- **Convert to 2024 Monsters**: Convert monsters in a 2014 adventure into their 2024 equivalents if available. If no 2024 version is available will try to convert the monster's equipment to the appropriate 2024 version.
- **Convert to 2024 Spells**: Convert all spells on actors in the adventure to their 2024 equivalents from the SRD or PHB.
- **Customize Join**: Automatically change the background and description of the world to one provided by the module (not necessary for quick start because that does the same work).

PR #6946, opened 2026-04-16
[#6923] Add system-specific adventure importer dialog with actions

arbron:
Adds a new Adventure Importer sheet that follows the dnd5e design style. Following the example of core with its V2 importer, this sheet is not made default and instead must be opted into on a per-adventure basis.

Adventure Importer 5e

The new sheet includes support for pre- and post-import actions which are presented as checkboxes on the importer sheet to give control to users. There are three actions initially provided by the system: "Activate Scene", "Display Journal", and "Customize World", but modules can add their own actions.

Each adventure can use flags to define which of the actions is run when that adventure is imported and customize their behavior. There is also a new `AdventureRegistry` that allows defining actions for a single adventure or for all adventures in a module. Actions added through the registry can support custom handlers without having to add to the global configuraiton.

```javascript
// These actions will be run across all adventures in the module
dnd5e.registry.adventures.setModule("dnd-adventures-faerun", {
  importActions: ["activateScene", "displayJournal"]
});

// This specific adventure has a different set of actions
dnd5e.registry.adventures.setAdventure("Compendium.dnd-adventures-faerun.adventures.Adventure.OJuc9WnDJDAETIFr", {
  importActions: [{ id: "displayJournal", journalId: "x0TaNijLJkanmufc" }]
});
```
Adventure Quickstart Dialog

The system also contains a dialog that presents actions available after using quickstart to load the adventure. By adding a separate quickstart handler actions can opt in to support on quickstarted adventures. These actions are passed all of the adventures that have that action and their configs so they can optimize if more than one adventure is quickstarted.

### DIFF
--- a/module/_types.mjs
+++ b/module/_types.mjs
@@ -129,6 +129,56 @@
 /* -------------------------------------------- */
 
 /**
+ * @typedef AdventureConfiguration
+ * @param {AdventureImportAction[]} importActions  Actions performed when the adventure is imported.
+ */
+
+/* -------------------------------------------- */
+
+/**
+ * @typedef AdventureImportAction
+ * @property {string} id                   Unique ID for this import action.
+ * @property {string} label                Localized label for the option.
+ * @property {AdventureImportPreHandler|AdventureImportPostHandler} handler  Handler function to call.
+ * @property {"pre"|"post"} lifecycle      Should this handler be called before or after importing.
+ * @property {boolean} [default=false]     Should this option be checked by default on the adventure importer?
+ * @property {AdventureImportQuickstartHandler} [quickstartHandler]  Handler called when after quickstarting a module.
+ *                                                                   Must be set to use an action during quickstart.
+ */
+
+/* -------------------------------------------- */
+
+/**
+ * @callback AdventureImportPostHandler
+ * @this {Adventure}
+ * @param {AdventureImportAction} config
+ * @param {AdventureImportResult} importResult
+ * @param {AdventureImportOptions} importOptions
+ * @returns {Promise<*>}
+ */
+
+/* -------------------------------------------- */
+
+/**
+ * @callback AdventureImportPreHandler
+ * @this {Adventure}
+ * @param {AdventureImportAction} config
+ * @param {AdventureImportData} importData
+ * @param {AdventureImportOptions} importOptions
+ * @returns {Promise<*>}
+ */
+
+/* -------------------------------------------- */
+
+/**
+ * @callback AdventureImportQuickstartHandler
+ * @param {{ adventure: Adventure, config: AdventureImportAction }[]} adventures
+ * @returns {Promise<*>}
+ */
+
+/* -------------------------------------------- */
+
+/**
  * Information needed to represent different area of effect target types.
  *
  * @typedef AreaTargetDefinition
--- a/module/_types.mjs
+++ b/module/_types.mjs
@@ -137,6 +137,56 @@
 /* -------------------------------------------- */
 
 /**
+ * @typedef AdventureConfiguration
+ * @param {AdventureImportAction[]} importActions  Actions performed when the adventure is imported.
+ */
+
+/* -------------------------------------------- */
+
+/**
+ * @typedef AdventureImportAction
+ * @property {string} id                   Unique ID for this import action.
+ * @property {string} label                Localized label for the option.
+ * @property {AdventureImportPreHandler|AdventureImportPostHandler} handler  Handler function to call.
+ * @property {"pre"|"post"} lifecycle      Should this handler be called before or after importing.
+ * @property {boolean} [default=false]     Should this option be checked by default on the adventure importer?
+ * @property {AdventureImportQuickstartHandler} [quickstartHandler]  Handler called when after quickstarting a module.
+ *                                                                   Must be set to use an action during quickstart.
+ */
+
+/* -------------------------------------------- */
+
+/**
+ * @callback AdventureImportPostHandler
+ * @this {Adventure}
+ * @param {AdventureImportAction} config
+ * @param {AdventureImportResult} importResult
+ * @param {AdventureImportOptions} importOptions
+ * @returns {Promise<*>}
+ */
+
+/* -------------------------------------------- */
+
+/**
+ * @callback AdventureImportPreHandler
+ * @this {Adventure}
+ * @param {AdventureImportAction} config
+ * @param {AdventureImportData} importData
+ * @param {AdventureImportOptions} importOptions
+ * @returns {Promise<*>}
+ */
+
+/* -------------------------------------------- */
+
+/**
+ * @callback AdventureImportQuickstartHandler
+ * @param {{ adventure: Adventure, config: AdventureImportAction }[]} adventures
+ * @returns {Promise<*>}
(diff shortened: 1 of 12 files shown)


module/applications/adventure/importer.mjs
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  /* -------------------------------------------- */
+  /*  Event Listeners and Handlers                */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  _onChangeForm(formConfig, event) {
+    if ( (event.target.name === "importFields") && (event.target.getAttribute("value") === "all") ) {
```
Fyorl: Seems to be doing the same thing that `super._onChangeForm` already does. `dnd5e-checkbox` `value` property should return `"all"`, no?

module/documents/adventure.mjs
```diff
@@ -1,8 +1,50 @@
...
+
+  /**
+   * The quickstart configuration data for the module in which this adventure appears. If `adventures` is defined in
+   * the quickstart configuration, will only return a result if this adventure is defined for this system.
+   * @type {QuickstartManifestData|null}
+   */
+  get quickstartConfig() {
+    if ( this.compendium?.metadata.packageType !== "module" ) return null;
+    const manifest = game.modules.get(this.compendium.metadata.packageName);
+    if ( !manifest.quickstart?.adventures?.[game.system.uuid]
```
Fyorl:
```suggestion
    if ( !manifest.quickstart?.adventures?.[game.system.id]
```

Fyorl (changes requested):
Good consolidation. Some parts of the design I am not convinced of yet.

So far our use of registries in the system are *passive*, they collect information at load. We don't have modules actively writing into them. For that pattern we use `CONFIG.DND5E`. I would prefer we continue with that, so I would propose either namespacing `CONFIG.DND5E.adventures.importActions`, e.g. `CONFIG.DND5E.adventures.importActions.dnd-tomb-annihilation = { myHandler: ... }`, with shared actions going under the `dnd5e` namespace; or keeping the shared actions at the top-level and requiring custom actions be prefixed, e.g. `CONFIG.DND5E.adventures.importActions["dnd-tomb-annihilation.myHandler"] = ...`

I am also not a fan of the flags and having multiple ways to define handler payloads. I would prefer we just had one way, also via `CONFIG.DND5E`, e.g.

```js
CONFIG.DND5E.adventure.imports["Compendium.dnd-tomb-annihilation.adventures.Adventure..."] = [
  { id: "activateScene", initialScene: "..." },
  { id: "displayJournal", initialJournal: "..." }
];
```

arbron:
> Good consolidation. Some parts of the design I am not convinced of yet.
> 
> So far our use of registries in the system are _passive_, they collect information at load. We don't have modules actively writing into them. For that pattern we use `CONFIG.DND5E`. I would prefer we continue with that, so I would propose either namespacing `CONFIG.DND5E.adventures.importActions`, e.g. `CONFIG.DND5E.adventures.importActions.dnd-tomb-annihilation = { myHandler: ... }`, with shared actions going under the `dnd5e` namespace; or keeping the shared actions at the top-level and requiring custom actions be prefixed, e.g. `CONFIG.DND5E.adventures.importActions["dnd-tomb-annihilation.myHandler"] = ...`
> 
> I am also not a fan of the flags and having multiple ways to define handler payloads. I would prefer we just had one way, also via `CONFIG.DND5E`, e.g.
> 
> ```js
> CONFIG.DND5E.adventure.imports["Compendium.dnd-tomb-annihilation.adventures.Adventure..."] = [
>   { id: "activateScene", initialScene: "..." },
>   { id: "displayJournal", initialJournal: "..." }
> ];
> ```

Switching over to config sounds good.

My main motivation to also having flags as an option is it allows simple setups to be done without requiring adventure modules to dig into code. A creator can just set a few flags on the adventure itself and not have to add a javascript file and handle extra configuration. I'll put up my updated version to see what you think, but I can still pull out the flag stuff.

Fyorl:
> My main motivation to also having flags as an option is it allows simple setups to be done without requiring adventure modules to dig into code. A creator can just set a few flags on the adventure itself and not have to add a javascript file and handle extra configuration.

Is the idea that they would add these flags directly to their unpacked adventure YAML? Because adding flags in-app also requires some macro/coding, right? Or were you thinking of adding another app like with the journal flags?

arbron:
> > My main motivation to also having flags as an option is it allows simple setups to be done without requiring adventure modules to dig into code. A creator can just set a few flags on the adventure itself and not have to add a javascript file and handle extra configuration.
> 
> Is the idea that they would add these flags directly to their unpacked adventure YAML? Because adding flags in-app also requires some macro/coding, right? Or were you thinking of adding another app like with the journal flags?

Wasn't thinking of another config app, but a macro that just does `adventure.setFlag("dnd5e", "importActions", ["openScene"]);` is pretty simple.

module/documents/adventure.mjs
```diff
@@ -1,8 +1,59 @@
...
+  /**
+   * List of configuration data for each import action. Options can either be defined using the `dnd5e.importActions`
+   * flag on the adventure to use pre-defined actions or fetched from `CONFIG.DND5E.adventure.config` by adventure
+   * UUID or package name.
+   * @type {AdventureImportAction[]}
+   */
+  get importActions() {
+    let actions = this.getFlag("dnd5e", "importActions")
+      ?? CONFIG.DND5E.adventure.config[this.uuid]?.importActions
+      ?? CONFIG.DND5E.adventure.config[this.compendium.metadata.packageName]?.importActions
```
Fyorl: We guard this in `quickstartConfig`. Maybe it's possible to be empty if this is a shallow `Adventure` constructed from the index.


```suggestion
      ?? CONFIG.DND5E.adventure.config[this.compendium?.metadata.packageName]?.importActions
```

module/_types.mjs
```diff
@@ -128,6 +128,56 @@
 
 /* -------------------------------------------- */
 
+/**
+ * @typedef AdventureConfiguration
+ * @param {AdventureImportAction[]} importActions  Actions performed when the adventure is imported.
```
Fyorl:
```suggestion
 * @property {AdventureImportAction[]} importActions  Actions performed when the adventure is imported.
```

module/applications/adventure/quickstart-dialog.mjs
```diff
@@ -0,0 +1,133 @@
...
+      format: { adventure: this.options.adventures[0].name, number: this.options.adventures.length }
+    });
+  }
+
+  /* -------------------------------------------- */
+  /*  Event Listeners and Handlers                */
+  /* -------------------------------------------- */
+
+  /**
+   * Run during the setup hook to present post-quickstart dialogs for each quickstarted adventure.
```
Fyorl:
```suggestion
   * Run during the ready hook to present post-quickstart dialogs for each quickstarted adventure.
```

Fyorl (approved):
> Wasn't thinking of another config app, but a macro that just does `adventure.setFlag("dnd5e", "importActions", ["openScene"]);` is pretty simple.

Well a macro is code, but I suppose it's less code. Seems like it's basically free to support flags though so I don't really have any objection. This all looks good to me.

PR #6949, opened 2026-04-16
[#6941] Fix bloodied visibility in v14.

### LINKED ISSUE #6941: [BUG] Bloodied status does not respect visibility settings
When the visibility setting for the Bloodied status is set to `Only Display for Allies`, player characters can still see the status on enemy tokens.

**Steps to Reproduce**
1. Log in as both a GM and a player.
2. Assign the player ownership of a friendly actor.
3. Set the Bloodied status visibility to `Only Display for Allies`.
4. Reduce a hostile token's HP to below 50%.
5. Check the hostile token from the player's perspective.

The player can still see the Bloodied status on the enemy token, even though the visibility is set to `Only Display for Allies`.

**Tested without modules:**

- System: DnD5e 5.3.1
- Core: Foundry VTT V14 Build 360 (nodeJS)
- Browser: Vivaldi (Chrome)

Fyorl:
- Closes #6941

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -3543,10 +3543,11 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
     return ActiveEffect.implementation.create({
       _id: ActiveEffect5e.ID.BLOODIED,
-      name: game.i18n.localize(CONFIG.DND5E.bloodied.name),
       img: CONFIG.DND5E.bloodied.img,
+      flags: { dnd5e: { isTemporary: true } },
+      name: game.i18n.localize(CONFIG.DND5E.bloodied.name),
       statuses: ["bloodied"],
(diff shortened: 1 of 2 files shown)


module/documents/actor/actor.mjs
```diff
@@ -3543,10 +3543,11 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
     return ActiveEffect.implementation.create({
       _id: ActiveEffect5e.ID.BLOODIED,
-      name: game.i18n.localize(CONFIG.DND5E.bloodied.name),
       img: CONFIG.DND5E.bloodied.img,
+      flags: { dnd5e: { isTemporary: true } },
```
Fyorl, line 3547: Just flagging this as something we might want to include in the system data model in 6.0

arbron (approved):

PR #6951, opened 2026-04-17
[#6866] Split advantage modifier evaluation into two phases. Eagerly expand dice terms to present to RollResolver.

### LINKED ISSUE #6866: [5.3] - Manual rolls with adv/dis do not offer resolution in one go
I would expect the manual roll resolver to behave closer to pre 5.3

Image

Image

Fyorl:
- Closes #6866 

Bit of a tricky one. There's quite a lot required to support this modifier, and I think there are still some bugs that I didn't address. Having to copy most of `_evaluateAsync` is a particular wart. Some other options available to us:

1. Revert to using `kh/dl` for d20 rolls and leave `adv` just for some special damage roll mechanics. They will still have the RollResolver issue but they're much rarer than d20 rolls.
2. We could make some extensions to the core API. Something that allows `DiceTerm`s to advertise some additional `number` would clean this up almost entirely.

In order to advocate (2) though I would need to understand why `adv` is preferable to just using `kh` and I'm not sure I quite understand it enough to defend it yet.

### DIFF
--- a/module/dice/basic-die.mjs
+++ b/module/dice/basic-die.mjs
@@ -16,31 +16,23 @@ export default class BasicDie extends Die {
   /* -------------------------------------------- */
 
   /**
-   * Handle rolling advantage and disadvantage for a die.
-   * @param {string} modifier        The matched modifier query.
-   * @returns {Promise<false|void>}  False if modifier was unmatched.
+   * Handle rolling advantage and disadvantage for a die. The extra rolls have already been performed by the main roll
+   * loop (pre-expansion inflated the dice count before the resolver opened), so this handler just partitions results
+   * into count + 1 sets of size, keeps the set with the best (adv) or worst (dis) total, and discards the others.
+   * @param {string} modifier  The matched modifier query.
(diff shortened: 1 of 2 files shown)


krbz999:
> We could make some extensions to the core API. Something that allows DiceTerms to advertise some additional number would clean this up almost entirely.

Wouldn't hate if this became something of a core feature since I have a [similar need](https://github.com/krbz999/ryuutama/issues/62) in Ryuu.

arbron (approved):
Seems to work as expected.

The main reason this is an improvement over `kh` is how it handles partitioning:

```
// 8d6kh4
2 [4] [3] 3 3 [5] 1 [6] = 18 total
```

```
// 4d6adv
2 4 3 3 [3] [5] [1] [6] = 15 total
```

With keep highest there is no way to represent the second format.

For the system it also allows us to use something like the proposed `damage.modifiers` field to easily add advantage to a damage roll, no need to also adjust the die number when applying that modifier.

PR #6963, opened 2026-04-23
Fix chat description visibility

AndrewBurian:
The tiniest bug fix ever.

Chat descriptions are supposed to be visible, but the reference here to `description.chat` is undefined by the time it's being read so it never triggers.

### DIFF
--- a/module/data/abstract/item-data-model.mjs
+++ b/module/data/abstract/item-data-model.mjs
@@ -198,7 +198,7 @@ export default class ItemDataModel extends SystemDataModel {
     const rollData = (activity ?? this.parent).getRollData();
     const isIdentified = identified !== false;
     const chat = isIdentified ? description.chat || description.value : unidentified?.description;
-    description = game.user.isGM || isIdentified ? description.value : unidentified?.description;
+    const desc = game.user.isGM || isIdentified ? description.value : unidentified?.description;
     uses = this.hasLimitedUses && (game.user.isGM || identified) ? uses : null;
     price = game.user.isGM || identified ? price : null;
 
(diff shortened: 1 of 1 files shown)


module/data/abstract/item-data-model.mjs
```diff
@@ -216,7 +216,7 @@ export default class ItemDataModel extends SystemDataModel {
         chat: await TextEditor.enrichHTML(chat ?? "", {
           rollData, relativeTo: this.parent, ...enrichmentOptions
         }),
-        concealed: game.user.isGM && game.settings.get("dnd5e", "concealItemDescriptions") && !description.chat
+        concealed: game.user.isGM && game.settings.get("dnd5e", "concealItemDescriptions") && !chat
```
arbron: This won't work, because if the chat description isn't defined then it `chat` falls back to the normal description, so this will always result in a description being displayed.

Instead you should keep this line as it was and adjust the line above that is replacing `description` and rename that something different (like just `desc`, replacing its usage in `enrichHTML` also) so that `description.chat` is still available here.

arbron (changes requested):

module/data/abstract/item-data-model.mjs
arbron:
```suggestion
        concealed: game.user.isGM && game.settings.get("dnd5e", "concealItemDescriptions") && !description.chat
```

module/data/abstract/item-data-model.mjs
arbron:
```suggestion
        value: await TextEditor.enrichHTML(desc ?? "", {
```

arbron:
```suggestion
    const desc = game.user.isGM || isIdentified ? description.value : unidentified?.description;
```

arbron (changes requested):

arbron (approved):

Issue #1581, opened 2022-06-13
Anime 5e hitdice


aaclayton:
Originally in GitLab by @salimhakima22

I started running a Anime 5e game and noticed a few classes in it use a D4 hit dice. I am unable to select that has a hit dice option on the character sheets. Can you add that option to the character sheets?

PR #7010, opened 2026-05-20
[#1581] fix: Barbarian Rages scale value increases to 6 at level 17, not 16

vfez1:
Description: Corrects the 2024 Barbarian class advancement table where the Rages ScaleValue incorrectly granted 6 uses at level 16. Per the 2024 PHB, this upgrade should occur at level 17.

D&D Beyond reference: https://www.dndbeyond.com/classes/2190875-barbarian

### DIFF
--- a/packs/_source/classes24/barbarian/barbarian.yml
+++ b/packs/_source/classes24/barbarian/barbarian.yml
@@ -513,7 +513,7 @@ system:
             value: 4
           '12':
             value: 5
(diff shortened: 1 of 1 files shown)


Fyorl (approved):

Fyorl:
Note that changes to the PHB are made separately to the SRD 5.2 included in the game system, so I have removed the 'closes' from the description.

PR #7025, opened 2026-05-31
Prevent exhaustion toggle on auxclick in TokenHUD

### LINKED ISSUE #7020: Right clicking on Exhaustion from the Token's Assign Status Effects HUD, always reduces it to Exhaustion I.
Testing on Foundry v14.363 and 5e 5.3.3.

For example Exhaustion IV drops down to Exhaustion I with one right click.

thatlonelybugbear:
Maybe closes #7020

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -732,6 +732,7 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     Hooks.on("renderTokenHUD", this.onTokenHUDRender);
     document.addEventListener("click", this.onClickTokenHUD.bind(this), { capture: true });
     document.addEventListener("contextmenu", this.onClickTokenHUD.bind(this), { capture: true });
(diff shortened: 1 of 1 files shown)


Fyorl:
Looks like this got broken when we moved to `auxclick` as the core status effect listener. And there are indeed differences in behaviour between OSes as I found out. I won't bore you with the details, but you can see MDN does allude to them [here](https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event):

> Due to timing differences between operating systems, this too is not a preventable default behavior of `auxclick`. Instead, this can be done by preventing the default behavior of the [`contextmenu`](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event) event.

The fix then it to just switch to `auxclick` ourselves to catch the same event as core and `stopPropagation` to prevent default core behaviour. I have updated the PR accordingly.

Fyorl (approved):

thatlonelybugbear:
Cheers! Makes sense 👍

PR #7027, opened 2026-06-01
[#7011] Use the encounter actor folder when importing member actors from a compendium

### LINKED ISSUE #7011: Place automatically imported Actors from Encounter placement within same folder as Encounter
When using the "Place Members" functionality of Encounter type Actors, any automatically imported Actors from compendiums are placed in the root of the actor sidebar list.

It would be helpful, for multiple adventure modules especially, if these automatically imported Actors were placed within the same folder (if any) as that of the Encounter's (base) actor.

There may be an extension of this request to Summoning activities as they use similar (if not shared) code paths, but I do not have a direct use case for such a change.

Aioros:
A small solution using `Actor5e.fetchExisting`'s existing `options` object.

Closes #7011

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -419,6 +419,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       // A linked world actor was found. Create a copy to avoid affecting the original.
       return actor.clone({
         "flags.dnd5e.isAutoImported": true,
+        folder: game.folders.get(folderId) ?? null,
         "_stats.compendiumSource": actor._stats.compendiumSource,
         "_stats.duplicateSource": actor.uuid
       }, { save: true });
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -373,6 +373,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
    * version has already been imported before importing it again.
    * @param {string} uuid                  The Actor's UUID.
    * @param {object} [options]
+   * @param {object} [options.folderId]    Folder ID for importing the Actor from a compendium.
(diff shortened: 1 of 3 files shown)


module/documents/actor/actor.mjs
```diff
@@ -411,7 +412,8 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     if ( actor.pack ) {
       // Template actor resides only in a compendium, import the actor into the world.
       return game.actors.importFromCompendium(game.packs.get(actor.pack), actor.id, {
```
Fyorl, line 414: Does the branch below need this as well?

Aioros: Good point. Looking at it now I want to say yes, but I have to admit I'm not sure exactly which situation leads to that branch.

arbron: That's for when the UUID in the summons config points to a world actor rather than a compendium actor.

Fyorl: Yeah it's pretty rare. I think if we are creating a clone we probably also want it tidied away in the same place.

Aioros: Got it. I'll add the same logic to the summons side then: a summoned, linked, world actor gets cloned in the same folder as the summoner's.

module/documents/actor/actor.mjs
```diff
@@ -361,6 +373,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
    * version has already been imported before importing it again.
    * @param {string} uuid                  The Actor's UUID.
    * @param {object} [options]
+   * @param {object} [options.folderId]    Folder ID for importing the Actor from a compendium.
```
```suggestion
   * @param {string} [options.folderId]    Folder ID for importing the Actor from a compendium.
```

Fyorl (approved):

PR #7033, opened 2026-06-03
[#6676] Add token-dimension size changes to `Actor#tokenActiveEffectChanges`

### LINKED ISSUE #6676: Add capacity to sync token size/scale with actor size
With V14's `tokenOverrides`, this should be something we can accomplish. Probably gated behind a setting, in case somebody wants to use arbitrary sizes for their tokens. Whether that's a world setting or some actor-bound field/flag I'm unsure of.

arbron:
Automatically resizes and scales tokens when the actor's size is changed.

Closes #6676

### DIFF
--- a/module/applications/token-config.mjs
+++ b/module/applications/token-config.mjs
@@ -1,6 +1,8 @@
 import { getHumanReadableAttributeLabel } from "../utils.mjs";
 import MovementSensesConfig from "./shared/movement-senses-config.mjs";
 
+const { BooleanField } = foundry.data.fields;
+
 /**
  * Custom token configuration application for handling dynamic rings & resource labels.
  */
@@ -10,8 +12,9 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
   async _onRender(context, options) {
     await super._onRender(context, options);
     if ( !this.rendered ) return;
-    this._prepareResourceLabels(this.element);
     this._applySenseSyncNotice(this.element);
+    this._applyTokenScaleLock(this.element);
+    this._prepareResourceLabels(this.element);
   }
 
   /* -------------------------------------------- */
@@ -33,10 +36,34 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
   }
 
   /* -------------------------------------------- */
+  /*  Appearance                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Add control to link token scale to actor scale.
+   * @param {HTMLElement} html  The rendered markup.
+   * @protected
+   */
+  _applyTokenScaleLock(html) {
+    const group = new BooleanField().toFormGroup({
+      hint: _loc("DND5E.TOKEN.LockScale.Hint"),
+      label: _loc("DND5E.TOKEN.LockScale.Label")
+    }, {
+      name: "flags.dnd5e.lockScale",
+      value: this.token._source.flags.dnd5e?.lockScale
+    });
+
+    const lockRotation = html.querySelector('.form-group:has([name="lockRotation"])');
+    lockRotation.after(group);
+  }
+
+  /* -------------------------------------------- */
+  /*  Resources                                   */
+  /* -------------------------------------------- */
 
   /**
    * Adds charge based items as attributes for the current token.
-   * @param {object} attributes The attribute groups to add the item entries to.
+   * @param {object} attributes  The attribute groups to add the item entries to.
    * @protected
    */
   _addItemAttributes(attributes) {
@@ -79,6 +106,8 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
     }
   }
 
+  /* -------------------------------------------- */
+  /*  Vision                                      */
   /* -------------------------------------------- */
 
   /**
@@ -133,8 +162,9 @@ export class PrototypeTokenConfig5e extends foundry.applications.sheets.Prototyp
   async _onRender(context, options) {
     await super._onRender(context, options);
     if ( !this.rendered ) return;
-    TokenConfig5e.prototype._prepareResourceLabels.call(this, this.element);
     TokenConfig5e.prototype._applySenseSyncNotice.call(this, this.element);
+    TokenConfig5e.prototype._applyTokenScaleLock.call(this, this.element);
+    TokenConfig5e.prototype._prepareResourceLabels.call(this, this.element);
   }
 
   /* -------------------------------------------- */
--- a/module/applications/token-config.mjs
+++ b/module/applications/token-config.mjs
@@ -45,6 +45,8 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
    * @protected
    */
   _applyTokenScaleLock(html) {
+    if ( html.querySelector('[name="flags.dnd5e.lockScale"]') ) return;
+
     const group = new BooleanField().toFormGroup({
       hint: _loc("DND5E.TOKEN.LockScale.Hint"),
       label: _loc("DND5E.TOKEN.LockScale.Label")
--- a/module/applications/token-config.mjs
+++ b/module/applications/token-config.mjs
@@ -47,7 +47,14 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
   _applyTokenScaleLock(html) {
     if ( html.querySelector('[name="flags.dnd5e.lockScale"]') ) return;
 
-    const group = new BooleanField().toFormGroup({
+    const lockSize = new BooleanField().toFormGroup({
+      hint: _loc("DND5E.TOKEN.LockSize.Hint"),
+      label: _loc("DND5E.TOKEN.LockSize.Label")
+    }, {
+      name: "flags.dnd5e.lockSize",
+      value: this.token._source.flags.dnd5e?.lockSize
+    });
+    const lockScale = new BooleanField().toFormGroup({
       hint: _loc("DND5E.TOKEN.LockScale.Hint"),
       label: _loc("DND5E.TOKEN.LockScale.Label")
     }, {
@@ -56,7 +63,7 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
     });
 
     const lockRotation = html.querySelector('.form-group:has([name="lockRotation"])');
-    lockRotation.after(group);
+    lockRotation.after(lockSize, lockScale);
   }
 
   /* -------------------------------------------- */
--- a/module/applications/token-config.mjs
+++ b/module/applications/token-config.mjs
@@ -8,6 +8,15 @@ const { BooleanField } = foundry.data.fields;
  */
 export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
 
+  /** @override */
+  static DEFAULT_OPTIONS = {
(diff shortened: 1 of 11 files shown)


module/documents/token.mjs
```diff
@@ -232,6 +232,14 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
     }
   }
 
+  /* -------------------------------------------- */
+
+  /** @override */
+  async _onOverrideSize(changes) {
+    if ( this.persisted ) this.update(changes);
+    else this.updateSource(changes);
```
arbron: This doesn't properly resize the token when the token config is open, only the texture scale changes work properly.

stwlam: I think a source update should also be performed if `this.object?.isPreview`

module/documents/actor/actor.mjs
```diff
@@ -365,7 +365,28 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    const sizeData = CONFIG.DND5E.actorSizes[this.system.traits?.size];
+    if ( !sizeData ) return;
+    const tokenSize = sizeData.token ?? 1;
+    this.tokenActiveEffectChanges[phase].push(...["width", "height", "depth"].map(key => ({
+      key, phase,
+      type: "override",
+      priority: 0,
+      value: tokenSize
+    })));
+    const tokenScale = sizeData.dynamicTokenScale ?? 1;
```
Fyorl, line 412: We apply this scaling to dynamic token ring tokens again in `TokenDocument5e#prepareData`. I think it's made redundant by this code.

Though that does raise an issue: We run this code regardless of whether a token actually has an AE affecting it. So all non-dynamic-ring tokens of Small actors will be reduced to 80% size suddenly when they update to 6.0.

module/documents/actor/actor.mjs
```diff
@@ -365,7 +365,28 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
     }
-    return super.applyActiveEffects(phase);
+    super.applyActiveEffects(phase);
+    if ( phase !== "initial" ) return;
+
+    // Translate this Actor's size category into Token changes
+    const sizeData = CONFIG.DND5E.actorSizes[this.system.traits?.size];
+    if ( !sizeData ) return;
+    const tokenSize = sizeData.token ?? 1;
+    this.tokenActiveEffectChanges[phase].push(...["width", "height", "depth"].map(key => ({
```
Fyorl, line 405: We have some SRD tokens that are really huge and exceed the gargantuan 4x4 (ancient red dragon, roc). I can't see any good way to automatically determine whether a token should be auto-sized or whether the author has made specific edits. We probably need to do the same as PF2e then with an explicit opt-in/out.

Image

Fyorl (changes requested):

module/applications/token-config.mjs
```diff
@@ -32,11 +35,35 @@ export class TokenConfig5e extends foundry.applications.sheets.TokenConfig {
...
+    const group = new BooleanField().toFormGroup({
+      hint: _loc("DND5E.TOKEN.LockScale.Hint"),
+      label: _loc("DND5E.TOKEN.LockScale.Label")
+    }, {
+      name: "flags.dnd5e.lockScale",
+      value: this.token._source.flags.dnd5e?.lockScale
+    });
+
+    const lockRotation = html.querySelector('.form-group:has([name="lockRotation"])');
+    lockRotation.after(group);
```
Fyorl: The `TokenConfig` can re-render (toggling 'Is Darkness Source' is an easy one if you want to repro it), so we should only append if our `lockScale` element isn't already in the sheet.

module/documents/token.mjs
Fyorl: As shark mentioned, this needs to be:
```js
if ( this.persisted && !this.object?.isPreview ) this.update(changes);
else this.updateSource(changes);
```
Same-ID clones get `persisted: true`, and with this code you could end up persisting un-committed token config changes in some very rare circumstances.

module/documents/actor/actor.mjs
```diff
@@ -372,7 +373,29 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       }
     }
 
-    return super.applyActiveEffects(phase);
+    super.applyActiveEffects(phase);
+    if ( phase !== "initial" ) return;
```
Fyorl: Maybe a `this.system.isCreature` check here too. We don't want to auto-size vehicles.

module/documents/active-effect.mjs
```diff
@@ -278,10 +278,10 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
     // Apply shims to moved fields
-    change = change.effect._applyChangeShim(change);
+    change = change.effect?._applyChangeShim(change) ?? change;
 
     if ( (model instanceof foundry.abstract.Document)
-      && !change.effect._checkCondition(change, options.replacementData) ) return {};
+      && !this._checkCondition(change, options.replacementData) ) return {};
 
     // Handle special actor flags
     if ( change.key.startsWith("flags.dnd5e.") ) change = change.effect._prepareFlagChange(model, change);
```
```suggestion
    if ( change.key.startsWith("flags.dnd5e.") ) change = change.effect?._prepareFlagChange(model, change);
```

module/documents/token.mjs
Fyorl: Another thought, actually: Should probably guard the non-source update on `game.users.isActiveGM`; I think this method fires on every client.

module/documents/actor/actor.mjs
Fyorl: Since `lockScale` is off by default, this is still an issue. Users/GMs would have to go through each of their size Small tokens and check this box or re-adjust scale. We probably need to migrate all size Small tokens based on `_stats.systemVersion` to have this checked on by default.

I would foresee that we then get some reports about an updated Enlarge/Reduce not working on size Small tokens because they were migrated to have this flag and the user doesn't realise. The only way to avoid that would be to split size from scale, which is a bit of a pain to do just because of this. So I'm less sure about that one. Maybe leave it for now.

Fyorl (changes requested):

module/documents/actor/actor.mjs
```diff
@@ -372,7 +373,29 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       }
     }
 
-    return super.applyActiveEffects(phase);
+    super.applyActiveEffects(phase);
+    if ( (phase !== "initial") && this.system.isCreature ) return;
```
```suggestion
    if ( (phase !== "initial") || !this.system.isCreature ) return;
```

module/documents/active-effect.mjs
```diff
@@ -278,13 +278,13 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
-    change = change.effect._applyChangeShim(change);
+    change = change.effect?._applyChangeShim(change) ?? change;
 
     if ( (model instanceof foundry.abstract.Document)
-      && !change.effect._checkCondition(change, options.replacementData) ) return {};
+      && !this._checkCondition(change, options.replacementData) ) return {};
 
     // Handle special actor flags
-    if ( change.key.startsWith("flags.dnd5e.") ) change = change.effect._prepareFlagChange(model, change);
+    if ( change.key.startsWith("flags.dnd5e.") ) change = change.effect?._prepareFlagChange(model, change);
```
```suggestion
    if ( change.key.startsWith("flags.dnd5e.") ) change = change.effect?._prepareFlagChange(model, change) ?? change;
```

module/migration.mjs
```diff
@@ -502,6 +502,10 @@ export async function migrateSettings() {
 export function migrateActorData(actor, actorData, migrationData, flags={}, { actorUuid }={}) {
   const updateData = {};
   _migrateTokenImage(actorData, updateData);
+  if ( (flags.bypassVersionCheck || foundry.utils.isNewerVersion("6.0.0", actorData._stats?.systemVersion))
+    && (actor.system.traits?.size === "sm") && (actor.prototypeToken.texture.scaleX !== 0.8) ) {
```
Fyorl: We don't want to lock dynamic ring tokens as they were being scaled automatically. Also this function is called on synthetic actors so we can skip those.

```suggestion
  if ( (flags.bypassVersionCheck || foundry.utils.isNewerVersion("6.0.0", actorData._stats?.systemVersion))
    && !actor.isToken && (actor.system.traits?.size === "sm") && (Math.abs(actorData.prototypeToken.texture.scaleX) !== 0.8) && !actorData.prototypeToken.ring.enabled ) {
```

module/migration.mjs
```diff
@@ -840,15 +844,21 @@ export function migrateRollTableData(table, migrationData) {
...
  * @returns {object}                The updateData to apply
  */
-export function migrateSceneData(scene, migrationData) {
+export function migrateSceneData(scene, migrationData, { bypassVersionCheck }={}) {
   const tokens = scene.tokens.reduce((arr, token) => {
     const t = token instanceof foundry.abstract.DataModel ? token.toObject() : token;
     const update = {};
     _migrateTokenImage(t, update);
+    const size = t.delta?.system?.traits?.size ?? game.actors.get(t.actorId)?.system?.traits?.size;
+    if ( (bypassVersionCheck || foundry.utils.isNewerVersion("6.0.0", token._stats?.systemVersion))
```
Fyorl: `TokenDocument` doesn't actually have `_stats`. You need to check the Scene's `_stats` instead.

module/documents/token.mjs
```diff
@@ -315,6 +308,14 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
 
   /* -------------------------------------------- */
 
+  /** @override */
+  async _onOverrideSize(changes) {
+    if ( this.persisted && !this.object?.isPreview && game.user.isActiveGM ) this.update(changes);
+    else this.updateSource(changes);
```
Fyorl:
```suggestion
    if ( !this.persisted || this.object?.isPreview ) this.updateSource(changes);
    else if ( game.user.isActiveGM ) this.update(changes);
```

module/migration.mjs
```diff
@@ -840,15 +844,21 @@ export function migrateRollTableData(table, migrationData) {
...
+      && (size === "sm") && (t.texture.scaleX !== 0.8) ) update["flags.dnd5e.lockScale"] = true;
```
Fyorl:
```suggestion
      && (size === "sm") && (Math.abs(t.texture.scaleX) !== 0.8) && !t.ring.enabled ) update["flags.dnd5e.lockScale"] = true;
```

module/documents/actor/actor.mjs
Fyorl: Sorry to go back and forth on this, but after working through the migration, I realise that of course people will want to change the scale on their tokens without also locking them out of size-change automation. So I think we do need separate toggled regardless. I feel bad though so if you want me to take this PR off your hands I can take care of that.

Fyorl (changes requested):

module/migration.mjs
arbron: We should also probably not do the scale check since the token mode is using `multiply`, so even if the token is manually set to `1.0` scale then it will just get scaled to `0.64`.

Fyorl (approved):

PR #7035, opened 2026-06-03
[#6344] Add field to make actor sizes support AE changes better

### LINKED ISSUE #6344: Change the way a creature's `Size` is done to allow for modifying it via active effects.

arbron:
Adds a custom field to handle modifying actor sizes using active effects. This adds support for `add`, `subtract`, `upgrade`, and `downgrade` active effect modes.

This also adds support for using full size names in active effects rather than the short versions (e.g. `large` instead of `lg`).

Closes #6344

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -1201,6 +1201,7 @@ DND5E.actorSizes = {
   sm: {
     label: "DND5E.SizeSmall",
     abbreviation: "DND5E.SizeSmallAbbr",
+    fullKey: "small",
     hitDie: 6,
     dynamicTokenScale: 0.8,
     numerical: 1
@@ -1208,12 +1209,14 @@ DND5E.actorSizes = {
   med: {
     label: "DND5E.SizeMedium",
     abbreviation: "DND5E.SizeMediumAbbr",
+    fullKey: "medium",
     hitDie: 8,
     numerical: 2
   },
   lg: {
     label: "DND5E.SizeLarge",
     abbreviation: "DND5E.SizeLargeAbbr",
+    fullKey: "large",
     hitDie: 10,
     token: 2,
     capacityMultiplier: 2,
@@ -1222,6 +1225,7 @@ DND5E.actorSizes = {
   huge: {
     label: "DND5E.SizeHuge",
     abbreviation: "DND5E.SizeHugeAbbr",
+    fullKey: "huge",
     hitDie: 12,
     token: 3,
     capacityMultiplier: 4,
@@ -1230,6 +1234,7 @@ DND5E.actorSizes = {
   grg: {
     label: "DND5E.SizeGargantuan",
     abbreviation: "DND5E.SizeGargantuanAbbr",
+    fullKey: "gargantuan",
     hitDie: 20,
     token: 4,
     capacityMultiplier: 8,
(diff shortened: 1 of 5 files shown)


module/data/actor/fields/actor-size-field.mjs
```diff
@@ -0,0 +1,61 @@
...
+      Math.clamp(idx + delta, 0, CONFIG.DND5E.actorSizes.orderedKeys.length - 1)
+    ];
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeSubtract(value, delta, model, change) {
+    if ( !Number.isNumeric(delta) ) return value;
+    return this._applyChangeAdd(value, delta * -1, model, change);;
```
Fyorl:
```suggestion
    return this._applyChangeAdd(value, delta * -1, model, change);
```

module/data/actor/fields/actor-size-field.mjs
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  _applyChangeUpgrade(value, delta, model, change) {
+    const valueN = CONFIG.DND5E.actorSizes[value]?.numerical ?? -Infinity;
```
Fyorl, line 48: No support for numeric deltas here or in `DOWNGRADE`. Also we have no `_applyChangeOverride` implementation so numeric handling there either.

module/config.mjs
```diff
@@ -1238,6 +1243,29 @@ DND5E.actorSizes = {
 };
 preLocalize("actorSizes", { keys: ["label", "abbreviation"] });
 
+Object.defineProperty(DND5E.actorSizes, "fullKeys", {
+  get() {
+    const value = Object.entries(this).reduce((obj, [key, config]) => {
+      obj[config.fullKey ?? key] = key;
+      return obj;
+    }, {});
+    Object.defineProperty(DND5E.actorSizes, "fullKeys", { value });
```
Fyorl, line 1252: How soon do these get memoized? Seems like it might be as soon as world load actor prep. Do modules have a chance to add/remove sizes before we cache them forever?

Fyorl (changes requested):

module/data/actor/fields/actor-size-field.mjs
arbron: I'm not sure what use that would be. You'd do `size | UPGRADE | 4` to get huge rather than doing `size | UPGRADE | huge`?

My general thinking is that we don't want people to actually deal with the numeric values anywhere, that they are only an internal method for relating one size to another.

module/config.mjs
arbron: Happens just before the `setup` hook when world documents are prepared, so modules can insert them during their `init` hook.

module/data/actor/fields/actor-size-field.mjs
Fyorl: Of course, I wasn't considering relative vs. absolute size changes.

Fyorl (approved):

PR #7036, opened 2026-06-03
[#4934] Use roller's spellcasting ability from check activity

### LINKED ISSUE #4934: "Spellcasting" for Check ability in `CheckActivity` should use roller's spellcasting

arbron:
When the check ability for a Check activity is set to "Spellcasting Ability" this will now use the roller's spellcasting ability rather than that of the actor who used the activity.

Adds support for `spellcasting` as an ability in all of the rolling methods on actor.

Also allows using `spellcasting` as an ability option in check and save enrichers, adapting the work from @krbz999's PR here: https://github.com/foundryvtt/dnd5e/pull/4883

Closes #4934

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -5037,6 +5037,8 @@ Object.defineProperty(DND5E, "enrichmentLookup", {
       addFullKeys("spellSchools");
       addFullKeys("tools");
       Object.entries(DND5E.vehicleTypes).forEach(([k, label]) => _enrichmentLookup.tools[k] = { label, key: k });
+
+      _enrichmentLookup.abilities.spellcasting = { label: _loc("DND5E.Spellcasting") };
     }
     return _enrichmentLookup;
   },
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -5038,7 +5038,11 @@ Object.defineProperty(DND5E, "enrichmentLookup", {
       addFullKeys("tools");
       Object.entries(DND5E.vehicleTypes).forEach(([k, label]) => _enrichmentLookup.tools[k] = { label, key: k });
 
-      _enrichmentLookup.abilities.spellcasting = { label: _loc("DND5E.Spellcasting") };
+      _enrichmentLookup.abilities.spellcasting = {
+        abbreviation: _loc("DND5E.Spellcasting"),
+        key: "spellcasting",
+        label: _loc("DND5E.Spellcasting")
+      };
     }
     return _enrichmentLookup;
   },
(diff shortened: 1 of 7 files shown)


module/config.mjs
```diff
@@ -4954,6 +4954,8 @@ Object.defineProperty(DND5E, "enrichmentLookup", {
       addFullKeys("spellSchools");
       addFullKeys("tools");
       Object.entries(DND5E.vehicleTypes).forEach(([k, label]) => _enrichmentLookup.tools[k] = { label, key: k });
+
+      _enrichmentLookup.abilities.spellcasting = { label: _loc("DND5E.Spellcasting") };
```
```suggestion
      _enrichmentLookup.abilities.spellcasting = { abbreviation: _loc("..."), key: "spellcasting", label: _loc("DND5E.Spellcasting") };
```

module/documents/actor/actor.mjs
```diff
@@ -1503,6 +1507,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
    * @returns {Promise<D20Roll[]|null>}                        A Promise which resolves to the created Roll instances.
    */
   async rollSavingThrow(config={}, dialog={}, message={}) {
+    if ( config.ability === "spellcasting" ) config.ability = this.spellcastingAbility;
```
Fyorl: Are you sure that mutating the `config` is safe? I think if you have multiple targets selected then you end up writing, e.g. `config.ability = "int"` and then the next target doesn't see `spellcasting` anymore and skips this line.

```suggestion
    const ability = config.ability === "spellcasting" ? this.spellcastingAbility : config.ability;
```

module/enrichers.mjs
Fyorl:
```suggestion
          ability: CONFIG.DND5E.enrichmentLookup.abilities[ability].label,
```

Fyorl:
```suggestion
  let abilityConfig = CONFIG.DND5E.enrichmentLookup.abilities[config.ability];
```

Fyorl (changes requested):
The caveats from the original PR haven't gone away. Spellcasting ability is not a property of a creature, it's a property of a class. So anything that asks you to roll with your 'spellcasting ability' must be doing so in the context of a class.

The only exception I can find in official content is spells cast from magic items, which let you choose your spellcasting ability (i.e. picking the highest like we do is correct).

So this will be wrong in general for any character that is multi-classed into two spellcasting classes where those spellcasting classes do not have the same spellcasting ability. I imagine that's fairly rare but still.

This creature-level spellcasting ability is a wart that I wish we could get rid of.

Fyorl (approved):
`GroupData#rollSkill` and `GroupData#rollSavingThrow` also look up `ability` from `CONFIG.DND5E.abilities` so probably also need some special-casing for spellcasting. Otherwise this looks good.

PR #7043, opened 2026-06-05
[#6975] Localize healing types in damage resistance traits

### LINKED ISSUE #6975: Healing types and damage traits
Since `system.traits.<dr|di|dv>.value` works with `healing` and `temphp` now, it would be nice if these were localized on the character sheet. Currently they just display "healing" or "temphp".

Also should be added to the wiki.

Probably need to be added to the trait configs, too.

Sayshal:
- Added an `additionalConfig` trait option that merges a secondary config into trait resolution (label, icon, and choice lookups)
- Changed `di`, `dr`, and `dv` to set `additionalConfig: "healingTypes"` so healing, temphp, and maximum localize on sheets and are selectable in the damage config dialog
- Sorted the merged choice list by label via `sortObjectEntries` so healing types interleave alphabetically rather than trailing
- Changed the damage config icon lookup to fall back to `CONFIG.DND5E.healingTypes`

Closes #6975


module/applications/actor/config/damages-config.mjs
```diff
@@ -66,7 +66,7 @@ export default class DamagesConfig extends TraitsConfig {
   /** @inheritDoc */
   _processChoice(data, key, choice, categoryChosen=false) {
     super._processChoice(data, key, choice, categoryChosen);
-    const config = CONFIG.DND5E.damageTypes[key];
+    const config = CONFIG.DND5E.damageTypes[key] ?? CONFIG.DND5E.healingTypes[key];
```
arbron, line 69: If we are going to present these in the config application, they should be split into a separate "Healing Types" card.

There is also a question as to how we handle the maximum HP type. We probably don't want to have a single immunity for cases where an actor gets a feature that makes them immune to any effect that would reduce their maximum hit points, we wouldn't also want to make them immune to the Aid spell.

module/documents/actor/trait.mjs
```diff
@@ -1,4 +1,5 @@
 import SelectChoices from "./select-choices.mjs";
+import { sortObjectEntries } from "../../utils.mjs";
```
arbron, line 2:
```suggestion
import { sortObjectEntries } from "../../utils.mjs";
import SelectChoices from "./select-choices.mjs";
```

module/documents/actor/trait.mjs
```diff
@@ -121,7 +137,10 @@ export function changeKeyPath(key, trait) {
  */
 export async function categories(trait) {
   const traitConfig = CONFIG.DND5E.traits[trait];
-  const config = foundry.utils.deepClone(CONFIG.DND5E[traitConfig.configKey ?? trait]);
+  let config = foundry.utils.deepClone(_traitConfigData(traitConfig, trait));
+
+  // Re-sort by label so merged additional config (e.g. healing types) interleaves rather than trailing.
+  if ( traitConfig.additionalConfig ) config = sortObjectEntries(config, "label");
```
arbron, line 143: Don't like the way these additional configs show up in advancement. If you have an advancement adding "Damage Immunity" and select "All Damage", the interface suggests that you are also no longer able to receive healing.

Image

arbron (changes requested):
I think this whole approach is going to need a rethink before it goes into the system. We may even want to split healing resistances/immunities/etc. off into a whole separate config object.

For now hold off any additional work on this until we have more design clarity.

Sayshal:
I'll leave this PR open for re-consideration post-6.0.

PR #7045, opened 2026-06-05
[#4612] Add `unidentifiedContents` property to containers.

### LINKED ISSUE #4612: Additional container property to not reveal contents if unidentified
Some items, like Bag of Holding, you do not need to identify to see the contents (you can turn it inside-out), but other items you do.

Sayshal:
- Add `unidentifiedContents` container property hiding contents from non-GMs while the container is unidentified
- Gate the container sheet contents list on a new `canViewContents` getter
- Report a zeroed value from `computeCapacity` when contents are concealed, so the capacity bar, thumbnail, and favorites chip show 0/X

**Not covered**: contents weight still counts toward the actor's encumbrance, and the container delete dialog still shows its item count. GMs always see contents and real capacity.

Closes #4612

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -3855,6 +3855,7 @@
     "Thrown": "Thrown",
     "Trait": "Passive Trait",
     "TwoHanded": "Two-Handed",
+    "UnidentifiedContents": "Hidden Until Identified",
     "Versatile": "Versatile",
     "Verbal": "Verbal",
     "WeightlessContents": "Weightless Contents"
--- a/lang/en.json
+++ b/lang/en.json
@@ -1724,6 +1724,10 @@
     "properties": {
       "label": "Container Properties"
     }
+  },
+  "UnidentifiedContentsWarning": {
+    "Title": "Unidentified Container",
+    "Message": "This container is unidentified. Anything you add to it will be hidden until it's identified. Continue?"
   }
 },
 
@@ -3855,7 +3859,7 @@
     "Thrown": "Thrown",
     "Trait": "Passive Trait",
     "TwoHanded": "Two-Handed",
-    "UnidentifiedContents": "Hidden Until Identified",
+    "UnidentifiedContents": "Hide Unidentified Contents",
     "Versatile": "Versatile",
     "Verbal": "Verbal",
     "WeightlessContents": "Weightless Contents"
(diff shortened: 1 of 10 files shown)


module/applications/item/container-sheet.mjs
```diff
@@ -105,7 +105,8 @@ export default class ContainerSheet extends ItemSheet5e {
 
     // Contents
     const Inventory = customElements.get(this.options.elements.inventory);
-    for ( const item of await this.item.system.contents ) {
+    const contents = this.item.system.canViewContents ? await this.item.system.contents : [];
```
arbron: Rather than just showing an empty container, this should instead hide the entire "Contents" tab using the same method used in `ItemSheet5e`. That would also ensure the currency row is properly hidden.

lang/en.json
```diff
@@ -3275,6 +3265,7 @@
     "Thrown": "Thrown",
     "Trait": "Passive Trait",
     "TwoHanded": "Two-Handed",
+    "UnidentifiedContents": "Hidden Until Identified",
```
arbron: Don't think this label is super clear, but I struggle to find one that fits on one line. May just have to use "Hide Unidentified Contents" and accept that the property will flow into a second line.

module/data/item/container.mjs
```diff
@@ -242,7 +253,7 @@ export default class ContainerData extends ItemDataModel.mixin(
       context.max = this.capacity.weight.value;
       context.units = CONFIG.DND5E.weightUnits[this.capacity.weight.units]?.label ?? "";
     }
-    context.value = context.value.toNearest(0.1);
+    context.value = this.canViewContents ? context.value.toNearest(0.1) : 0;
```
arbron: Rather than setting the value to zero, I'd suggest returning something like a `hidden` parameter. Then you can update the template and hide the bar itself. Showing the bar but having it pinned at `0` seems a bit deceptive.

arbron (changes requested):
One other thing worth considering is how to handle dropping items into an unidentified container. If you can't see the contents, can you add to them. If so, any item you add becomes inaccessible until you identify the container.

Sayshal:
> One other thing worth considering is how to handle dropping items into an unidentified container. If you can't see the contents, can you add to them. If so, any item you add becomes inaccessible until you identify the container.

Updated. Contents tab now just doesn't render at all. Capacity bar now hides itself via a flag instead of faking a 0/X. Relabeled the property too. Also added a confirm dialog for players dropping items into a container they can't see into.

arbron (approved):

Fyorl (approved):

PR #7046, opened 2026-06-05
[#7034] Keep current token placement when reverting a transformation

### LINKED ISSUE #7034: When reverting a transformation back to the original token, the elevation is retained but not the level of the transformed token
Whenever the user transforms into anything (e.g. druid into a beast) the level and elevation are retained from the original token into the new transformation.  However, when the token is reverted back to their original token, the transformed token's level isn't retained and is reverted to whatever default level exists.  The elevation is maintained from the transformed beast back to the original token, though.

EDIT: The above was incorrect about elevation.  The elevation was always setting to 10, which happened to be correct at time of original testing, but later when I had 3 levels (0 to 10, 10 to 20, 20 to 30) and moved to the 3rd level (20 to 30) from the 2nd level (10 to 20) while transformed and then transformed back the previous elevation of 10 was maintained instead of 20.  This appears to be due to getting the correct elevation from the transformed token document but then overwriting it with the previousToken (the non-transformed token from before transforming) elevation from the call to `foundry.utils.mergeObject()` that happens afterwards

Source of the issue appears to be here where no level is set (instead it's taken from the deepClone() and never updated).
https://github.com/foundryvtt/dnd5e/blob/0fe006403860eae5a0479a742be2d05f549cb8d0/module/documents/actor/actor.mjs#L3187

Example Video:

https://github.com/user-attachments/assets/582f5ad6-b742-4d91-a0df-fa768d25c910

Sayshal:
- Apply the live token's elevation, level, hidden, and rotation after restoring the pre-transform data, so reverting no longer resets them
- Set the reverted token's level from the live token (canvas path), which previously fell back to the scene default

Closes #7034

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -2807,9 +2807,8 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     // Token appearance updates
     const tokenPropsFromSource = ["width", "height", "alpha", "lockRotation", "ring"];
     const tokenTexturePropsFromSource = ["offsetX", "offsetY", "scaleX", "scaleY", "src", "tint"];
-    const tokenPropsFromSelf = [
-      "bar1", "bar2", "displayBars", "displayName", "actorLink", "disposition", "rotation", "elevation", "hidden"
-    ];
+    const tokenPropsFromSelf = ["bar1", "bar2", "displayBars", "displayName", "actorLink", "disposition"];
+    const tokenPropsPreserved = ["rotation", "elevation", "hidden", "level"];
 
     for ( const k of tokenPropsFromSource ) {
       d.prototypeToken[k] = sourceData.prototypeToken[k];
@@ -2817,7 +2816,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     for ( const k of tokenTexturePropsFromSource ) {
       d.prototypeToken.texture[k] = sourceData.prototypeToken.texture[k];
     }
-    for ( const k of tokenPropsFromSelf ) {
+    for ( const k of [...tokenPropsFromSelf, ...tokenPropsPreserved] ) {
       d.prototypeToken[k] = o.prototypeToken[k];
     }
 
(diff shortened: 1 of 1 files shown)


module/documents/actor/actor.mjs
```diff
@@ -3173,6 +3173,9 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       tokenUpdate.delta = actorData;
 
       foundry.utils.mergeObject(tokenUpdate, this.token.getFlag("dnd5e", "previousTokenData"));
+      tokenUpdate.elevation = this.token.elevation;
+      tokenUpdate.hidden = this.token.hidden;
+      tokenUpdate.rotation = this.token.rotation;
```
Fyorl: I see we take some pains to only store a limited subset of fields in `previousTokenData` (at the end of `transformInto`). So perhaps it would make more sense to simply omit `elevation`, `hidden`, and `rotation` from that set instead.

Fyorl (changes requested):

module/documents/actor/actor.mjs
arbron: We should probably move `rotation`, `elevation`, and `hidden` from `tokenPropsFromSelf` and create a new config containing those and `level` that is shared between `transformInto` and `revertOriginalForm` to indicate what should always be copied over from the token being changed to its new form. That would make it easy to omit from `previousTokenData` when transforming.

Fyorl: It's a given they should be removed from `tokenPropsFromSelf` so that we do not store them in `previousTokenData`, but I do not understand what you mean by creating a config. This set of fields is not configurable, right? `rotation`, `elevation`, `hidden`, and `level` should always be maintained, no?

arbron: I just meant another constant rather than something configurable, but yeah if they aren't saved in `previousTokenData` there is no reason to even reference that constant inside `revertOriginalForm`.

Fyorl (approved):

PR #7049, opened 2026-06-05
[#4505] Allow `0` as a numeric scale value

### LINKED ISSUE #4505: Allow for 0 in numeric scale values
As an example, we have a feature that has a cost of 1 from levels 3 to 16, then has a cost of 0 from level 17 onwards. However, a numeric scale value does not allow for inputting '0'.

We would love to avoid having to write variations of `sign(floor( 16 / @classes.bob.levels ))`. :)

Sayshal:
- Keep numeric scale value entries of `0` on save instead of stripping them as empty
- Inherit a `0` to later levels as a placeholder, like any non-zero value
- Show `0` in the legacy scale value title display
- Skip consumption when an activity's resolved cost is `0` (no dialog prompt, nothing consumed)

Closes #4505

### DIFF
--- a/module/applications/advancement/scale-value-config.mjs
+++ b/module/applications/advancement/scale-value-config.mjs
@@ -132,7 +132,7 @@ export default class ScaleValueConfig extends AdvancementConfig {
     let lastValue = null;
     for ( const [lvl, value] of Object.entries(configuration.scale) ) {
       if ( this.advancement.testEquality(lastValue, value) ) configuration.scale[lvl] = null;
-      else if ( Object.keys(value ?? {}).some(k => value[k]) ) {
+      else if ( this.constructor._hasValue(value) ) {
         this._mergeScaleValues(value, lastValue);
         lastValue = value;
       }
@@ -165,10 +165,22 @@ export default class ScaleValueConfig extends AdvancementConfig {
 
   /* -------------------------------------------- */
 
+  /**
+   * Is a scale value entry non-empty?
+   * @param {object} value  Scale value entry.
+   * @returns {boolean}
+   * @protected
+   */
+  static _hasValue(value) {
+    return Object.keys(value ?? {}).some(k => (value[k] != null) && (value[k] !== ""));
+  }
+
+  /* -------------------------------------------- */
+
   /** @override */
   static _cleanedObject(object) {
     return Object.entries(object).reduce((obj, [key, value]) => {
-      if ( Object.keys(value ?? {}).some(k => value[k]) ) obj[key] = value;
+      if ( this._hasValue(value) ) obj[key] = value;
       else obj[key] = _del;
       return obj;
     }, {});
--- a/module/applications/advancement/scale-value-config.mjs
+++ b/module/applications/advancement/scale-value-config.mjs
@@ -165,6 +165,17 @@ export default class ScaleValueConfig extends AdvancementConfig {
 
   /* -------------------------------------------- */
 
+  /** @override */
+  static _cleanedObject(object) {
+    return Object.entries(object).reduce((obj, [key, value]) => {
+      if ( this._hasValue(value) ) obj[key] = value;
+      else obj[key] = _del;
+      return obj;
+    }, {});
+  }
+
+  /* -------------------------------------------- */
+
   /**
    * Is a scale value entry non-empty?
    * @param {object} value  Scale value entry.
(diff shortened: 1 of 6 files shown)


module/data/advancement/scale-value.mjs
```diff
@@ -204,7 +204,9 @@ export class ScaleValueType extends foundry.abstract.DataModel {
    * @returns {string}
    */
   static getPlaceholder(name, lastValue) {
-    return lastValue?.[name] ?? "";
+    const value = lastValue?.[name];
+    // Coerce to a string so a numeric 0 placeholder is rendered.
+    return value == null ? "" : String(value);
```
arbron:
```suggestion
    return value === null ? "" : String(value);
```

module/data/advancement/scale-value.mjs
arbron: Noticed that when converting from "Anything" to "Numeric" type it will create new `0` entries at the start and after each entry with a value. You'll need to fix `ScaleValueTypeNumber#convertFrom` to specifically ignore initial values of `null` or empty strings.

```suggestion
    if ( (original.formula === null) || (original.formula === "") || Number.isNaN(value) ) return null;
```

module/data/activity/fields/consumption-targets-field.mjs
```diff
@@ -711,6 +711,19 @@ export class ConsumptionTargetData extends foundry.abstract.DataModel {
 
   /* -------------------------------------------- */
 
+  /**
+   * Does this target's cost resolve to a fixed zero, meaning nothing is consumed? Non-deterministic
+   * costs (those containing dice) are never treated as zero.
+   * @param {ActivityUseConfiguration} [config={}]  Usage configuration.
+   * @returns {boolean}
+   */
+  hasZeroCost(config={}) {
```
arbron: Shift this method above `resolveCost` (alphabetically nice).

module/documents/activity/mixin.mjs
```diff
@@ -568,6 +568,7 @@ export default function ActivityMixin(Base) {
           ? this.consumption.targets.keys() : config.consume.resources;
         for ( const index of indexes ) {
           const target = this.consumption.targets[index];
+          if ( target.hasZeroCost(config) ) continue;
```
arbron, line 571: This prevents the cost from being spent, but the system still believes a cost is present when creating the dialog and displaying the chat card.

Image

Image

You'll need to modify `_prepareUsageConfig` so when it determines `hasResourceConsumption` it checks each consumption profile to see if any of them have a zero cost. Something like this should work:

```javascript
const hasResourceConsumption = !!this.consumption.targets.find(c => !c.hasZeroCost(config));
```

module/applications/advancement/scale-value-config.mjs
```diff
@@ -165,10 +165,22 @@ export default class ScaleValueConfig extends AdvancementConfig {
 
   /* -------------------------------------------- */
 
+  /**
+   * Is a scale value entry non-empty?
+   * @param {object} value  Scale value entry.
+   * @returns {boolean}
+   * @protected
+   */
+  static _hasValue(value) {
```
arbron, line 185: Move this below `_cleanedObject`.

arbron (changes requested):

module/documents/advancement/scale-value.mjs
```diff
@@ -71,7 +71,7 @@ export default class ScaleValueAdvancement extends Advancement {
   /** @inheritDoc */
   titleForLevel(level, { configMode=false, legacyDisplay=false }={}) {
     const value = this.valueForLevel(level)?.display;
-    if ( !value || !legacyDisplay ) return this.title;
+    if ( (value == null) || (value === "") || !legacyDisplay ) return this.title;
```
arbron:
```suggestion
    if ( (value === null) || (value === "") || !legacyDisplay ) return this.title;
```

arbron (changes requested):

module/data/advancement/scale-value.mjs
Fyorl: Not strictly incorrect. `value == null` catches both `undefined` and `null`. Maybe:
```js
return value ? String(value) : "";
```

arbron: That would return blank for `0`, while then intention is to return an actual value there.

Sayshal: Something like `return String(lastValue?.[name] ?? "")` would work?

module/documents/advancement/scale-value.mjs
Fyorl: This is also not correct. Now we no longer check for `undefined`, which seems like it's a possible value given the optional-chaining above.

module/documents/advancement/scale-value.mjs
```diff
@@ -71,7 +71,7 @@ export default class ScaleValueAdvancement extends Advancement {
   /** @inheritDoc */
   titleForLevel(level, { configMode=false, legacyDisplay=false }={}) {
     const value = this.valueForLevel(level)?.display;
-    if ( !value || !legacyDisplay ) return this.title;
+    if ( (value === null) || (value === "") || !legacyDisplay ) return this.title;
```
arbron:
```suggestion
    if ( (!value && (value !== 0)) || !legacyDisplay ) return this.title;
```

arbron (changes requested):

arbron (approved):

PR #7053, opened 2026-06-05
[#5926] Rename the Cast activity spellbook toggle to "Display in Spells Tab"

### LINKED ISSUE #5926: Change "Display in Spellbook" to "Display in Spells Tab"
In the Identity tab of a Cast activity, there is a "Display in Spellbook" setting. This shouldn't use the word "Spellbook", as that is a specific term of art that does not apply to every type of spell collection; this should say "Display in Spells Tab".

Image

The description text for this setting should probably also say something about the _requirements_ for this being displayed in the Spells tab (item must have the `Magical` property if it's not a Feature; item must be attuned if it has Attunement required; item must be equipped if it's equippable).

Sayshal:
- Rename the label from "Display in Spellbook" to "Display in Spells Tab"
- Note the display requirements (Magical property, attunement) in the hint

Closes #5926

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -1399,7 +1399,7 @@
       },
       "spellbook": {
         "label": "Display in Spells Tab",
-        "hint": "Display spell in the Spells tab of the character sheet. Equippable items must have the Magical property and be attuned if they require attunement."
+        "hint": "Display spell in the Spells tab of the character sheet. Items must have the Magical property and be attuned if they require attunement."
       },
       "uuid": {
         "label": "Spell to Cast"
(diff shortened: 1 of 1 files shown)


lang/en.json
```diff
@@ -1398,8 +1398,8 @@
         "hint": "Spell components & tags to ignore while casting."
       },
       "spellbook": {
-        "label": "Display in Spellbook",
-        "hint": "Display spell in the Spells tab of the character sheet."
+        "label": "Display in Spells Tab",
+        "hint": "Display spell in the Spells tab of the character sheet. Equippable items must have the Magical property and be attuned if they require attunement."
```
```suggestion
        "hint": "Display spell in the Spells tab of the character sheet. Items must have the Magical property and be attuned if they require attunement."
```

arbron (changes requested):

arbron (approved):

PR #7054, opened 2026-06-05
[#4397] Hide Delete and Expand/Collapse from the favorites context menu

### LINKED ISSUE #4397: Feature Request: Remove "Delete" from Context Menu in Favorites
I keep selecting "Delete" trying to "delete a favorite", which of course ends up deleting that item instead.

image

It does display a dialog requesting confirmation to delete the item itself, but in my head I respond "of course I want to delete this favorite". Which is obviously wrong :). 

I have to imagine that most people clicking delete from this context menu actually mean to click on "Remove Favorite".

It would make sense to remove "Delete" from the Favorites context menus. "Expand"/"Collapse" here also doesn't make sense (or work).

Sayshal:
- Suppress the Delete and Expand/Collapse entries when the context menu opens from the favorites list, leaving Remove Favorite
- Scope to .favorites so inventory, feature, and facility menus are unaffected

Closes #4397

### DIFF
--- a/module/applications/components/inventory.mjs
+++ b/module/applications/components/inventory.mjs
@@ -323,6 +323,7 @@ export default class InventoryElement extends (foundry.applications.elements.Ado
    */
   _getContextOptions(item, element) {
     const compendiumLocked = game.packs.get(item.pack)?.locked;
+    const inFavorites = !!element?.closest?.(".favorites");
 
     // Standard options.
     const options = [{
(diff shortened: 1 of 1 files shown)


module/applications/components/inventory.mjs
```diff
@@ -323,6 +323,7 @@ export default class InventoryElement extends (foundry.applications.elements.Ado
    */
   _getContextOptions(item, element) {
     const compendiumLocked = game.packs.get(item.pack)?.locked;
+    const inFavorites = !!element?.closest?.(".favorites");
```
arbron: Element should always be present and always be an `HTMLElement`, so optional chaining shouldn't be necessary (if you find a case where that isn't true, that would be a bug):

```suggestion
    const inFavorites = !!element.closest(".favorites");
```

module/applications/components/inventory.mjs
arbron: I'd say we should hide "Duplicate" also.

```suggestion
      visible: () => !inFavorites && item.canDuplicate && item.isOwner && !compendiumLocked,
```

arbron (changes requested):

arbron (approved):

PR #7055, opened 2026-06-05
[#3993] Hide the attunement icon on unidentified items from players

### LINKED ISSUE #3993: [UI] You can see attunement symbol on an un-identified item
Hey guys, thank you for your hard work! Had a problem last night. An item that was not identified and wasn't obvious that is magical showed the player the attunement icon on their character sheets. Luckily they didn't notice. Could you please not show this icon to players when the item is not attuned to? Thank you

SCR-20240817-lfgl

Sayshal:
- Gate the attunement control on ctx.concealDetails so it no longer reveals the item's magical status

Closes #3993

### DIFF
--- a/module/applications/actor/character-sheet.mjs
+++ b/module/applications/actor/character-sheet.mjs
@@ -876,7 +876,7 @@ export default class CharacterActorSheet extends BaseActorSheet {
     ctx.concealDetails = !game.user.isGM && (item.system.identified === false);
     ctx.isStack = Number.isNumeric(item.system.quantity) && (item.system.quantity !== 1);
 
-    if ( item.system.attunement ) ctx.attunement = item.system.attuned ? {
+    if ( item.system.attunement && !ctx.concealDetails ) ctx.attunement = item.system.attuned ? {
       icon: "fa-sun",
       cls: "attuned",
       title: "DND5E.AttunementAttuned"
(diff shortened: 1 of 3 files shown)


arbron:
Not sure this is a change we want to make. There is nothing in the rules stating you cannot attune to an item you haven't identified, in fact there is Sage Advice stating the exact opposite.

If we did want to make this change, we would also want to change the context menu because otherwise the player could just click there and see the "Attune" option.

Sayshal:
> Not sure this is a change we want to make. There is nothing in the rules stating you cannot attune to an item you haven't identified, in fact there is Sage Advice stating the exact opposite.
> 
> If we did want to make this change, we would also want to change the context menu because otherwise the player could just click there and see the "Attune" option.

Yeah, this one definitely rides the line of RAW,  but I also see the merit in hiding that an item is anything more than a stick. A setting may be overkill but could end up being part of a more larger sweep of QoL settings (i.e identifying being user-blocked, etc.)

I'll update the PR to hide the contextmenu  for now while it simmers.

Fyorl:
The rules seem reasonably clear to me at least on this point.

> Handling a magic item is enough to give you a sense that it is extraordinary, but learning a magic item’s properties isn’t automatic.

We take 'give you a sense that it is extraordinary' to mean that characters should know when an item is magical or not, but it seems quite explicit that the item's properties are not revealed. Maybe you disagree but I would say that whether an item requires attunement is a property of that item.

Per the rules you are not prevented from attuning to an item you haven't identified, but similarly you are not prevented from activating say, a Wand of Magic Missiles, but we hide the Cast activity regardless.

Sayshal:
> The rules seem reasonably clear to me at least on this point.

Then hiding the attunement icon at least would be consistent with other behavior. I could see a time where we have a visibility settings menu to handle things like this, cast activity, identifying items, and similar 'RAW-lite' settings.

Sayshal:
Rebased.

Fyorl (approved):

PR #7056, opened 2026-06-05
[#3285] Hide death saves when the character's max HP is 0

### LINKED ISSUE #3285: Don't roll out the death saves on the character sheet if the character's max hp is 0
Just an oddity that happens if you remove a class and suddenly have no hit points. Yeah you're at 0 hp, but if the max is 0, too, then death saves are irrelevant.

Sayshal:
- Gate the death-save tray on a max-HP check so it no longer shows for a 0/0 character (e.g. after removing all classes)
- Guard _toggleDeathTray against the tray being absent, since it is auto-opened on a 0-HP update

Closes #3285

### DIFF
--- a/module/applications/actor/character-sheet.mjs
+++ b/module/applications/actor/character-sheet.mjs
@@ -921,12 +921,7 @@ export default class CharacterActorSheet extends BaseActorSheet {
       this._renderSpellbook(context, options);
     }
 
-    // Show death tray at 0 HP
-    const renderContext = options.renderContext ?? options.action;
-    const renderData = options.renderData ?? options.data;
-    const isUpdate = (renderContext === "update") || (renderContext === "updateActor");
-    const hp = foundry.utils.getProperty(renderData ?? {}, "system.attributes.hp.value");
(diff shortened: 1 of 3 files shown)


arbron (changes requested):
I'm reluctant to hide the whole tray, because it means that the death saves config application is completely inaccessible when a character has `0` max HP. Instead we should prevent the death tray from opening automatically unless the character has some max HP.

Sayshal:
> I'm reluctant to hide the whole tray, because it means that the death saves config application is completely inaccessible when a character has `0` max HP. Instead we should prevent the death tray from opening automatically unless the character has some max HP.

Updated! The tray's always accessible now, only the auto-open is gated on max HP. I moved it to the actor's `_onUpdate` instead of the sheet, since deleting a class fires the HP update and the item deletion concurrently and the sheet could catch it mid-race with stale HP max. Using the existing `isAdvancement` flag to skip auto-open there instead, so it's not timing-dependent anymore.

arbron (approved):

Fyorl (approved):

PR #7058, opened 2026-06-06
[#5891] Add tooltips to active effects in the chat effect tray

### LINKED ISSUE #5891: Feature Request: Tooltips to Active Effects in the Active Effects tray in the chat log.
Would be nice to be able to hover over an Active Effect in the chat log's application tray similar to how they show on actor sheets in order to remind yourself what an active effect does in cases where a feature has more than one active effect as a quick reference.

Image

Sayshal:
- Attach a rich effect tooltip to each effect row in the chat tray, matching actor sheets

Closes #5891

### DIFF
--- a/module/applications/components/effect-application.mjs
+++ b/module/applications/components/effect-application.mjs
@@ -120,7 +120,14 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
       effect.updateDuration();
       const li = document.createElement("li");
       li.classList.add("effect");
-      li.dataset.id = effect.id;
+      Object.assign(li.dataset, {
+        id: effect.id,
+        tooltip: `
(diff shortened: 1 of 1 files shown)


module/applications/components/effect-application.mjs
arbron: Let's keep all of the `dataset` changes together up here.

```suggestion
      Object.assign(li.dataset, {
        id: effect.id,
        tooltip: `
          <section class="loading" data-uuid="${effect.uuid}"><i class="fas fa-spinner fa-spin-pulse"></i></section>
        `,
        tooltipClass: "dnd5e2 dnd5e-tooltip item-tooltip themed theme-light",
        tooltipDirection: "LEFT"
      });
```

arbron (changes requested):

arbron (approved):

PR #7059, opened 2026-06-06
[#6960] Show the parent class in subclass tooltips

### LINKED ISSUE #6960: Add the class a subclass belongs to when showing a subclass when viewing its tooltip.
Would be nice if we could see a `Fighter Subclass` in this case when viewing a subclass in a tooltip.

Image

Sayshal:
- Override tooltipSubtitle to show "<Class> Subclass" via a new DND5E.SubclassOf key

Closes #6960

### DIFF
--- a/module/data/item/subclass.mjs
+++ b/module/data/item/subclass.mjs
@@ -65,6 +65,17 @@ export default class SubclassData extends ItemDataModel.mixin(AdvancementTemplat
   }
 
   /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  get tooltipSubtitle() {
+    const cls = dnd5e.registry.classes.choices[this.classIdentifier];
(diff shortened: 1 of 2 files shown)


module/data/item/subclass.mjs
```diff
@@ -64,6 +64,17 @@ export default class SubclassData extends ItemDataModel.mixin(AdvancementTemplat
     ]);
   }
 
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  get tooltipSubtitle() {
+    const cls = dnd5e.registry.classes.choices[this.classIdentifier];
```
arbron: Should probably use `get(this.classIdentifier)?.name` instead, `choices` has to do a bunch of mapping and sorting every time you call it so is more work when all you need is a single class.


```suggestion
    const cls = dnd5e.registry.classes.get(this.classIdentifier)?.name;
```

arbron (changes requested):

arbron (approved):

PR #7060, opened 2026-06-06
[#7009] Prevent duplicate Compendium Browser sidebar buttons

### LINKED ISSUE #7009: detaching/attaching pack sidebar results in multiple "Open Compendium Browser" buttons
can get multiple by popping out the pack sidebar, detaching, re-attaching, etc etc.

Image

Sayshal:
- Skip injecting the button if one already exists, so sidebar re-renders (detach/reattach) don't stack duplicates

Closes #7009


arbron:
Don't think this is an issue anymore after a fix from core.

PR #7064, opened 2026-06-06
[#6598] Widen the inventory uses column to fit up to four digits

### LINKED ISSUE #6598: Limited Uses / Charges field display gets cut off after what looks to be 2 digits.
Not a massive issue, but having an item with a limited uses extending into triple+ digits has the column kinda cut off.

Image

Sayshal:
- Bump the uses/charges column and its value/max fields
- Keep the column registry width in sync with the CSS

Closes #6598

(Note, if wanted, we can drop all these changes except `less/v2/inventory.less:266` if we just want to allow 3 digits nicely. I went with  4 for redundancy)

### DIFF
--- a/less/v2/inventory.less
+++ b/less/v2/inventory.less
@@ -262,7 +262,7 @@
       .item-uses, .item-charges {
         gap: unset;
         &.item-detail { font-weight: bold; }
-        .max, input { width: 20px; }
+        .max, input { width: 26px; }
         .separator { color: var(--dnd5e-color-gold); }
         .max {
           color: var(--color-text-secondary);
--- a/less/v2/inventory.less
+++ b/less/v2/inventory.less
@@ -262,7 +262,7 @@
       .item-uses, .item-charges {
         gap: unset;
         &.item-detail { font-weight: bold; }
-        .max, input { width: 26px; }
+        .max, input { width: 24px; }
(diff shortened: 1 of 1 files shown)


Fyorl:
Some of the sheet designs are very sensitive to column width. Increasing column width here will potentially cause other columns to be hidden at default sheet sizes, which would be undesirable. So unfortunately it's not as simple as just increasing the column width.

Looking at the screenshot, it seems the third digit is only barely clipped so we would need to look at things like checking there is no excess padding in the input, or checking if the column's full width is actually being used. Then we might consider adjusting font size.

I would say 4 digits is right out.

Sayshal:
> Some of the sheet designs are very sensitive to column width. Increasing column width here will potentially cause other columns to be hidden at default sheet sizes, which would be undesirable. So unfortunately it's not as simple as just increasing the column width.
> 
> Looking at the screenshot, it seems the third digit is only barely clipped so we would need to look at things like checking there is no excess padding in the input, or checking if the column's full width is actually being used. Then we might consider adjusting font size.
> 
> I would say 4 digits is right out.

Yeah I can see the fragility there. I'm going to revert to what I was geting nicely with 3 digits, which is just a small bump from `20px` -> `26px` which doesn't cause the column to grow.

Fyorl (approved):

PR #7066, opened 2026-06-06
[#6831] Hide the DC on save buttons when it is zero

### LINKED ISSUE #6831: [Minor Issue] Saving throw chat button
Saving throw buttons in chat cards that have no DC will show "DC 0 {ability} saving throw"

Since saving throws that don't have a DC set or have it set to 0 don't actually check against the DC it would be nice if the DC text wasn't included on the button in this case. 
Had a couple of players get confused thinking a foundry had broken when there simply wasn't a DC.

Sayshal:
Closes #6831

### DIFF
--- a/module/documents/activity/save.mjs
+++ b/module/documents/activity/save.mjs
@@ -44,10 +44,11 @@ export default class SaveActivity extends ActivityMixin(BaseSaveActivityData) {
 
     for ( const abilityId of this.save.ability ) {
       const ability = CONFIG.DND5E.abilities[abilityId]?.label ?? "";
+      const promptTitle = _loc("DND5E.SavePromptTitle", { ability });
       buttons.push({
         label: `
-          <span class="visible-dc">${_loc("DND5E.SavingThrowDC", { dc, ability })}</span>
-          <span class="hidden-dc">${_loc("DND5E.SavePromptTitle", { ability })}</span>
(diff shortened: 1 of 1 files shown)


module/documents/activity/save.mjs
```diff
@@ -44,10 +44,11 @@ export default class SaveActivity extends ActivityMixin(BaseSaveActivityData) {
...
     for ( const abilityId of this.save.ability ) {
       const ability = CONFIG.DND5E.abilities[abilityId]?.label ?? "";
+      const promptTitle = _loc("DND5E.SavePromptTitle", { ability });
       buttons.push({
         label: `
-          <span class="visible-dc">${_loc("DND5E.SavingThrowDC", { dc, ability })}</span>
-          <span class="hidden-dc">${_loc("DND5E.SavePromptTitle", { ability })}</span>
+          <span class="visible-dc">${dc ? _loc("DND5E.SavingThrowDC", { dc, ability }) : promptTitle}</span>
+          <span class="hidden-dc">${promptTitle}</span>
         `,
```
arbron: Let's match how it is written to `CheckActivity` where we just leave out the two spans if no DC is present:

```javascript
        label: dc ? `
          <span class="visible-dc">${_loc("DND5E.SavingThrowDC", { dc, ability })}</span>
          <span class="hidden-dc">${promptTitle}</span>
        ` : promptTitle,
```

arbron (changes requested):

arbron (approved):

PR #7069, opened 2026-06-06
[#5900]  Expose damage calculation context to hooks

### LINKED ISSUE #5900: [Feature] - Pass the calculated damages to the `preApplyDamage` and `applyDamage` hook calls.
This will create an easier way for modules to grab the full damage data.

thatlonelybugbear:
Closes #5900


thatlonelybugbear:
A module could use something like the following to alter on the fly damage calculations, without altering Actor data as it sees fit:
```js
const randomDMCache = new Map();

Hooks.on('dnd5e.preCalculateDamage', (actor, damages, options, context) => {

	if ( !actor.appliedEffects.some(e => e.name === "The Module Forces Changes on this Roll only") ) return;
         
	const rollData = actor.getRollData();

	const messageId = options.originatingMessage?.id ?? 'no-message';
	const key = `${messageId}.${actor.uuid}`;

	// Add a cached random flat damage modification to fire damage.
	let randomDM = randomDMCache.get(key);
	if (randomDM === undefined) {
		randomDM = Math.floor(Math.random() * 4) + 1;
		randomDMCache.set(key, randomDM);
	}

	context.modifications.fire = (context.modifications.fire ?? 0) - randomDM;

	// Add resistance to cold damage.
	if (!context.traits.dr.value.has('cold')) context.traits.dr.value.add('cold');

	// Upgrade fire resistance to fire immunity.
	if (context.traits.dr.value.has('fire')) {
		context.traits.dr.value.delete('fire');
		context.traits.di.value.add('fire');
	}

	// Downgrade lightning immunity to lightning resistance for this calculation.
	if (context.traits.di.value.has('lightning')) {
		context.traits.di.value.delete('lightning');
		context.traits.dr.value.add('lightning');
	}

	// Add a deterministic system-style damage modification formula.
	context.modifications.acid = (context.modifications.acid ?? 0) + dnd5e.utils.simplifyBonus('@abilities.con.mod', rollData);
});

  ```

thatlonelybugbear:
This PR takes a different approach than the original issue title suggested.

Instead of changing the `preApplyDamage` flow, it exposes a shared calculation context to `dnd5e.preCalculateDamage` and `dnd5e.calculateDamage`. That gives modules a supported place to inspect and adjust the cloned damage calculation state before it is applied, including resolved damage modifications, traits, multiplier, and whether maximum HP changes are handled as damage or healing.

The main reason for targeting the calculation hooks is that damage is calculated repeatedly for both preview and final application. Keeping these changes in `calculateDamage` lets the damage tray, target previews, and final applied damage use the same logic. It also avoids adding late recalculation behavior to `preApplyDamage`, where HP updates have already been derived.

The PR also adds `context.recalculate` for `dnd5e.calculateDamage` hooks that mutate damage descriptions, so totals can be rebuilt from the changed entries before the result is returned.

PR #7071, opened 2026-06-06
Swap to using `modifyBatch` in places where bulk updates are used

arbron:
Start using `foundry.documents.modifyBatch` in several places in the system where mulitple update operations are handled including:

- Advancement application
- Enchantment rider creation
- Activity consumption/recovery
- Actor rests
- Linked summoning changes
- Combatant recovery

In order to simplify the common pattern of updating actors as well as their embedded items and to resolve issues with those updates on synthetic actors, a new `performBulkUpdate` method was added that takes `ActorUpdatesDescription` and transforms it into the needed `DatbaseOperation`s.

One place that this doesn't modify is `Actor#transformInto` on unlinked actors since there is currently some special handling to deal with tombstoning items. If that extra work is no longer required due to core changes we should be able to make that change.

### DIFF
--- a/module/applications/advancement/advancement-manager.mjs
+++ b/module/applications/advancement/advancement-manager.mjs
@@ -910,7 +910,7 @@ export default class AdvancementManager extends Application5e {
     }
 
     // Apply changes from clone to original actor
-    this.actor.performBulkUpdate(
+    await this.actor.performBulkUpdate(
       { actor: updates, create: toCreate, delete: toDelete, item: toUpdate },
       { isAdvancement: true, updateOptions: { diff: false, recursive: false } }
     );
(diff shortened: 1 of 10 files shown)


module/applications/advancement/advancement-manager.mjs
```diff
@@ -910,12 +910,10 @@ export default class AdvancementManager extends Application5e {
...
-    await Promise.all([
-      this.actor.update(updates, { isAdvancement: true }),
-      this.actor.createEmbeddedDocuments("Item", toCreate, { keepId: true, isAdvancement: true }),
-      this.actor.updateEmbeddedDocuments("Item", toUpdate, { diff: false, recursive: false, isAdvancement: true }),
-      this.actor.deleteEmbeddedDocuments("Item", toDelete, { isAdvancement: true })
-    ]);
+    this.actor.performBulkUpdate(
+      { actor: updates, create: toCreate, delete: toDelete, item: toUpdate },
+      { isAdvancement: true, updateOptions: { diff: false, recursive: false } }
+    );
```
krbz999: Arrivederci, I won't miss these 5 lines of code.

Fyorl (approved):

PR #7073, opened 2026-06-06
[#5832] Lead actor sheet titles with the actor name

### LINKED ISSUE #5832: Make a minimized statblock start with the name of the actual token
Currently when you minimize a statblock it shows "Non playable character:" or "Player character:" and then the actual name of the token

Image

This makes it kind of impossible to see what statblock you have minimized and especially confusing if you have a few different statblocks minimized

Sayshal:
Closes #5832

### DIFF
--- a/module/applications/actor/api/base-actor-sheet.mjs
+++ b/module/applications/actor/api/base-actor-sheet.mjs
@@ -132,6 +132,13 @@ export default class BaseActorSheet extends PrimarySheetMixin(
 
   /* -------------------------------------------- */
 
+  /** @override */
+  get title() {
+    return this.actor.name;
+  }
+
+  /* -------------------------------------------- */
+
   /**
    * The cached concentration information for the character.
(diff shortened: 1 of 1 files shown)


module/applications/actor/api/base-actor-sheet.mjs
```diff
@@ -132,6 +132,13 @@ export default class BaseActorSheet extends PrimarySheetMixin(
 
   /* -------------------------------------------- */
 
+  /** @override */
+  get title() {
+    return this.actor.name;
```
krbz999, line 137: I think there's still value in displaying the subtype when not minimized.

Sayshal: I might just be missing something obvious - why did/would the subtype display when not minimized?

krbz999: it wouldnt, my mistake. Forgot we entirely hide them.

Fyorl (approved):

PR #7077, opened 2026-06-07
[#7022] Clear expired duration on reapply

### LINKED ISSUE #7022: Reapplying an expired active effect reuses stale expiration state data
Foundry: 14.363
dnd5e: 5.3.3

Reproduction steps:
- Create an empty world using the dnd5e system
- Enable no modules
- Create a blank scene
- Create a Player Character actor and give it an ability that creates an active effect with a limited duration (I used See Invisibility from the SRD, but I believe any will work)
- Place the actor into the scene
- Invoke the ability, and apply the effect to the actor
- Advance the game clock by an amount of time greater than the duration of the effect
    - `await game.time.advance(2 * 60 * 60);` or similar
- Note that the effect immediately moves from Temporary to Unavailable
- Use the ability again and apply the effect again, the effect moves back to Temporary
- Advance time again
Expected: The effect would once again move to unavailable
Actual: The effect remains in Temporary until you refresh the canvas or invoke some other change to the actor (such as editing the hp value)

The likely code path: _applyEffectToActor reuses existingEffect by origin and updates initial duration/disabled, but may not clear duration.expired or is leaving some other expiration state stale causing the game clock hook to ignore the effect.

thatlonelybugbear:
Should be closing #7022


roth-michael:
I suspect this may get a slightly different fix with AE changes for v14 (for instance, `getInitialDuration` is deprecated, so some additional change will be necessary here regardless).

thatlonelybugbear:
> I suspect this may get a slightly different fix with AE changes for v14 (for instance, `getInitialDuration` is deprecated, so some additional change will be necessary here regardless).

Isn't that just a wrapper around
```js 
return { start: this.getEffectStart() };
```
The resetting should still be the correct surface I think 🤔 

In general probably we would end up with something like
```js
return existingEffect.update(
	foundry.utils.mergeObject(
		{
			start: effect.constructor.getEffectStart(),
			...durationOverride,
			'duration.expired': false,
			disabled: false,
		},
		effectFlags,
	),
);
```

Or are there more changes incoming/pending?

roth-michael:
That's probably about right. Will have to check and see how that would interact with a specific `start.combatant` though in retrospect, since iirc most of my proposed logic for handling start/end of my/your turn stuff happens in `_preCreate`

arbron:
This change has already been made in more recent AE changes, but doesn't fix the underlying problem.

PR #6145, opened 2025-09-05
[#5176] Add extended roll configuration to actor data models

arbron:
Adds a new data structure to actor data models for storing more information on rolls. This structure replaces the existing bonuses data and adds roll mode support for setting roll modes, minimums, and maximums. It also allows for more options to apply changes in both more specific and more general ways:

```
// Bonus added to all ability checks
system.rolls.ability.check.bonus | ADD | 1d4
// Disadvantage on all attack rolls
system.rolls.attack.mode | ADD | -1
// Minimum for ranged spell attacks
system.rolls.attack.rsak.min | OVERRIDE | 10
// Bonus to melee weapon attacks
system.rolls.damage.mwak.bonus | ADD | 1d6[radiant]
```

Closes #5176

Issue #4914, opened 2024-12-20
Extend the set of default abilities used for attacks


krbz999:
If the actor type is `character`, then the default ability used for melee attacks should be Strength, and Dexterity for ranged attacks. If `npc`, then the default ability should be the highest of Strength or Dexterity (for both melee and ranged) - this last part is currently missing (see #2735).

*However*, it would be good if this could be extended with additional abilities, perhaps with a flag in special traits; this would allow for artificer subclasses to more easily make use of Intelligence for melee attacks.

E.g., ignoring npcs and Finesse for a moment:
- `flags.dnd5e.meleeAbility: "int"` - would cause the `character` to prefer the highest of INT and STR for melee attacks.
- `flags.dnd5e.rangedAbility: "int"` - would cause the `character` to prefer the highest of INT and DEX for ranged attacks.

PR #7080, opened 2026-06-07
[#7079] Allow selecting attack ability in roll dialog

### LINKED ISSUE #7079: Turn the `Attack Ability` part of Attack Activities in a Multi Select field akin to damage types to allow more than one ability to be chosen (alongside the default)
Currently, something like `You can use Intelligence in place of Strength and/or Dexterity when you attack with this weapon` isn't cleanly possible in terms of all the rules that go along it. Spells like `Shillelagh` add in a new attack activity that just allows you to attack with that one instead of the base attack activity located on the weapon.

This sometimes causes problems when you are enchanting weapons (such as a 2014 Hexblade Warlock making a magic weapon their Hexblade) and that weapon applies an Active Effect on Hit (Sword of Wounding for instance, once per turn when you hit a creature, apply a wound to them). Enchanting a weapon to allow the warlock in this case to `use your Charisma modifier, instead of Strength or Dexterity, for the attack and damage rolls`  means that:
- (a) The just modify the base attack activity and swap it from `default` to `charisma` would make it not ideal for characters with a lower spellcasting stat than their Strength or Dexterity, because it would just default to Charisma in this case and end up not letting you choose Charisma, or the default ability scores that the weapon uses, either Strength or Strength/Dexterity for Finesse weapons.
- (b) Going the new Attack Activity route that a user can use the new `Attack (Charisma)` activity, any effect riders on the attack are missing as described previously in the Sword of Wounding example above.

Proposal: Allow setting multiple choses (similar to damage types for damage rolls), allowing for something like `Default (Strength / Dexterity), Charisma` and have a dropdown in the attack dialog that defaults to the highest value amongst the current attack roll calculations. Would probably break pre-existing content like Shillelagh, but would also allow that enchantment to just be a single line adding `Wisdom` to that field.

Image

thatlonelybugbear:
Closes #7079 

Commits:
- https://github.com/foundryvtt/dnd5e/pull/7080/commits/985922b4c32d234ec2e93f31490c13c14e142c41
  - [x] Adds attack ability options to the attack roll configuration so attacks with multiple valid abilities can expose a selector in the roll dialog.
  - [x] Stores ability options on the roll config before pre-roll hooks run, so modules can extend or replace the available choices during roll configuration.
  - [x] Carries the selected ability into attack roll construction (would trigger Elven Accuracy checks), and follow-up damage rolls, allowing the chosen ability modifier to be used consistently across the attack workflow.
- https://github.com/foundryvtt/dnd5e/pull/7080/commits/ede41e0c2633ba6bb8d62550caedbf1f816b6775
  - [x] Changes attack activity ability configuration to a multi-select field. 
  - [x] Adds migration handling for existing attack activity ability strings.
- https://github.com/foundryvtt/dnd5e/pull/7080/commits/5ead3abba980415a86d5486e38504f84461f9520
  - [x] Preserves legacy `system.ability` enchantment handling for the new multi-select attack ability field.
  - [x] Supports enchantments targeting `activities[attack].attack.ability` and `system.activities.<activityID>.attack.ability`.
  - [x] Supports `Add`, `Override`, and `Subtract` enchantment changes for attack ability multi-select fields, including default attack abilities such as finesse weapon Strength/Dexterity defaults.
  

Stacked work:
- [ ] If this PR is merged, I can rebase the follow-up from here: <https://github.com/thatlonelybugbear/dnd5e/pull/1>
  
Not handled yet:
- [ ] Array-valued enchantment changes such as `["wis", "cha"]` are not supported yet. That should probably be handled separately as part of a broader enchantment refactor for Set-backed ability fields.

```js
Hooks.on('dnd5e.preRollAttack', (config) => {
	config.abilityOptions ??= [];

	for (const ability of ['wis', 'int']) {
		if (config.abilityOptions.some((option) => option.value === ability)) continue;
		config.abilityOptions.push({
			value: ability,
			label: CONFIG.DND5E.abilities[ability].label,
		});
	}
});
```
image

### DIFF
--- a/module/data/activity/_types.mjs
+++ b/module/data/activity/_types.mjs
@@ -51,7 +51,7 @@
 /**
  * @typedef {ActivityData} AttackActivityData
  * @property {object} attack
- * @property {Set<string>} attack.ability         Abilities used to make the attack and determine damage.
+ * @property {string} attack.ability              Ability used to make the attack and determine damage.
  * @property {string} attack.bonus                Arbitrary bonus added to the attack.
  * @property {object} attack.critical
  * @property {number} attack.critical.threshold   Minimum value on the D20 needed to roll a critical hit.
--- a/module/data/activity/_types.mjs
+++ b/module/data/activity/_types.mjs
@@ -51,7 +51,7 @@
 /**
  * @typedef {ActivityData} AttackActivityData
  * @property {object} attack
- * @property {string} attack.ability              Ability used to make the attack and determine damage.
+ * @property {Set<string>} attack.ability         Abilities used to make the attack and determine damage.
  * @property {string} attack.bonus                Arbitrary bonus added to the attack.
  * @property {object} attack.critical
  * @property {number} attack.critical.threshold   Minimum value on the D20 needed to roll a critical hit.
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -278,9 +278,7 @@ export default class BaseAttackActivityData extends BaseActivityData {
    */
   getAttackData({ ammunition, ability, attackMode, situational }={}) {
     const rollData = this.getRollData({ ability, roll: { attackMode } });
-    if ( ability && (ability in CONFIG.DND5E.abilities) ) {
-      rollData.roll.ability = ability;
-    }
+    if ( ability && (ability in CONFIG.DND5E.abilities) ) rollData.roll.ability = ability;
     if ( this.attack.flat ) return CONFIG.Dice.BasicRoll.constructParts({ toHit: this.attack.bonus }, rollData);
 
     const weapon = this.item.system;
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -163,7 +163,7 @@ export default class BaseAttackActivityData extends BaseActivityData {
     return this.item.system.validAttackTypes ?? new Set();
   }
 
-/* -------------------------------------------- */
+  /* -------------------------------------------- */
 
   /** @override */
   static transformTypeData(source, activityData, options) {
@@ -206,6 +206,7 @@ export default class BaseAttackActivityData extends BaseActivityData {
 
   /** @inheritDoc */
   prepareData() {
+    this.#abilities = undefined;
     super.prepareData();
     this.attack.type.value ||= this.item.system.attackType ?? "melee";
     this.attack.type.classification ||= this.item.system.attackClassification ?? "weapon";
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -5,7 +5,7 @@ import FormulaField from "../fields/formula-field.mjs";
 import DamageField from "../shared/damage-field.mjs";
 import BaseActivityData from "./base-activity.mjs";
 
-const { ArrayField, BooleanField, NumberField, SchemaField, StringField } = foundry.data.fields;
+const { ArrayField, BooleanField, NumberField, SchemaField, SetField, StringField } = foundry.data.fields;
 
 /**
  * @import { AttackDamageRollProcessConfiguration } from "../../dice/_types.mjs";
@@ -23,7 +23,7 @@ export default class BaseAttackActivityData extends BaseActivityData {
     return {
       ...super.defineSchema(),
       attack: new SchemaField({
-        ability: new StringField(),
+        ability: new SetField(new StringField()),
         bonus: new FormulaField(),
         critical: new SchemaField({
           threshold: new NumberField({ integer: true, positive: true })
@@ -50,17 +50,37 @@ export default class BaseAttackActivityData extends BaseActivityData {
 
   /** @override */
   get ability() {
-    if ( this.attack.ability === "none" ) return null;
-    if ( this.attack.ability === "spellcasting" ) return this.spellcastingAbility;
-    if ( this.attack.ability in CONFIG.DND5E.abilities ) return this.attack.ability;
+    const configuredAbilities = this.abilities;
+    if ( (configuredAbilities.size === 1) && configuredAbilities.has("none") ) return null;
+    if ( configuredAbilities.size === 1 ) return configuredAbilities.first();
 
-    const availableAbilities = this.availableAbilities;
+    const availableAbilities = configuredAbilities.size ? configuredAbilities : new Set(this.availableAbilities);
+    availableAbilities.delete("none");
     if ( !availableAbilities?.size ) return null;
     if ( availableAbilities?.size === 1 ) return availableAbilities.first();
     const abilities = this.actor?.system.abilities ?? {};
     return availableAbilities.reduce((largest, ability) =>
       (abilities[ability]?.mod ?? -Infinity) > (abilities[largest]?.mod ?? -Infinity) ? ability : largest
     , availableAbilities.first());
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Configured abilities that can be used with this attack.
+   * @type {Set<string>}
+   */
+  get abilities() {
+    const values = foundry.utils.getType(this.attack.ability) === "string"
+      ? [this.attack.ability]
+      : this.attack.ability;
+    const abilities = new Set(values);
+    abilities.delete("");
+    if ( abilities.delete("default") ) {
+      for ( const ability of this.availableAbilities ) abilities.add(ability);
+    }
+    if ( abilities.delete("spellcasting") && this.spellcastingAbility ) abilities.add(this.spellcastingAbility);
+    return abilities;
   }
 
   /* -------------------------------------------- */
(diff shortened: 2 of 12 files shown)


thatlonelybugbear:
I think the `flags.dnd5e.meleeAbility: "int"` from  https://github.com/foundryvtt/dnd5e/issues/4914 could be a good next step

thatlonelybugbear:
Is the Attack ability multi select proposal something that the system wants to support?

If yes, I can take a stab at implementing that as well.

thatlonelybugbear:
Converted to draft, so that I can properly address the Enchantments targeting those fields. Will be ready soon

thatlonelybugbear:
https://github.com/foundryvtt/dnd5e/pull/7080/commits/5ead3abba980415a86d5486e38504f84461f9520

 Examples of supported Enchantments

  - `system.ability` /  `activities[attack].attack.ability  | Add | cha`
    - Adds Charisma to all attack activities on the enchanted item.
    - If the attack uses implicit defaults, such as a finesse weapon, this preserves the default options and adds Charisma.

  - `system.ability` / `activities[attack].attack.ability | Override | cha`
    - Replaces each attack activity's configured abilities with Charisma.

  - `system.ability` / `activities[attack].attack.ability | Subtract | dex` (same with ` | Add | -dex`)
    - Removes Dexterity from each attack activity.
    - If the attack uses implicit defaults, such as Strength/Dexterity from finesse, this expands those defaults first and leaves Strength.

  - `system.activities.<activityID>.attack.ability | Add | cha`
    - Adds Charisma only to the specific attack activity.

  - `system.activities.<activityID>.attack.ability | Override | cha`
    - Replaces the configured abilities only for the specific attack activity.

  - `system.activities.<activityID>.attack.ability | Subtract | dex`
    - Removes Dexterity only from the specific attack activity.

  Not yet supported:

  - `system.ability | Add | ["cha", "wis"]`
  - `activities[attack].attack.ability | Override | ["cha", "wis"]`
  - `system.activities.<activityID>.attack.ability | Add | ["cha", "wis"]`

  Those array-valued changes need broader Set-backed enchantment handling.

thatlonelybugbear:
This can be followed by: https://github.com/thatlonelybugbear/dnd5e/pull/1

> Adds ability-level attack roll configuration alongside existing check and save roll configuration.
>
> Changes:
> - Adds `system.abilities.*.attack.roll` for attack roll mode/min/max.
> - Adds `system.abilities.*.bonuses.attack`.
> - Applies ability attack roll advantage/disadvantage and min/max to attack rolls.
> - Preserves ability attack bonus as `system.abilities.*.attack.value`.
> - Adds ability attack condition effect hooks.
> - Adds the attack roll controls and attack bonus field to ability configuration.
> 

image

module/data/activity/_types.mjs
```diff
@@ -51,7 +51,7 @@
 /**
  * @typedef {ActivityData} AttackActivityData
  * @property {object} attack
- * @property {string} attack.ability              Ability used to make the attack and determine damage.
+ * @property {Set<string>} attack.ability         Abilities used to make the attack and determine damage.
```
arbron: Currently we are thinking of keeping the ability in `AttackActivity` to just a single option. The activity would then create its own `abilities` set internally that can be targeted by enchantments to add new abilities to the list but while keeping the user-facing interface simple.

module/dice/_types.mjs
```diff
@@ -76,6 +76,8 @@
  * @typedef {D20RollProcessConfiguration} AttackRollProcessConfiguration
  * @property {Item5e|boolean} [ammunition]    Specific ammunition to consume, or `false` to prevent any ammo usage.
  * @property {WeaponAttackMode} [attackMode]  Mode to use for making the attack and rolling damage.
+ * @property {string} [ability]               ID of the ability to roll as found in `CONFIG.DND5E.abilities`.
+ * @property {FormSelectOption[]} [abilityOptions]  Ability options that can be selected for this attack.
```
arbron: These should instead be in `AttackRollConfigurationDialogOptions` along with `ammunitionOptions` and the others.

module/documents/activity/attack.mjs
```diff
@@ -95,15 +95,20 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
...
       target: targets.length === 1 ? targets[0].ac : undefined
     }, config);
 
     const ammunitionOptions = this.item.system.ammunitionOptions ?? [];
+    rollConfig.abilityOptions = this._prepareAbilityOptions();
+    if ( rollConfig.abilityOptions.length && !rollConfig.abilityOptions.some(a => a.value === rollConfig.ability) ) {
+      rollConfig.ability = rollConfig.abilityOptions[0]?.value;
+    }
+    rollConfig.elvenAccuracy = this.actor?.getFlag("dnd5e", "elvenAccuracy")
+      && CONFIG.DND5E.characterFlags.elvenAccuracy.abilities.includes(rollConfig.ability);
```
arbron: Since we are calculating this in `_buildAttackConfig` it shouldn't be necessary to do it here, since that method is called even for fast-forwarded rolls.

module/documents/activity/attack.mjs
```diff
@@ -177,6 +183,7 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
       if ( !rolls[0].options[key] ) continue;
       foundry.utils.setProperty(messageConfig.data, `flags.dnd5e.roll.${key}`, rolls[0].options[key]);
     }
+    if ( rolls[0].options.ability ) foundry.utils.setProperty(messageConfig.data, "flags.dnd5e.roll.ability", rolls[0].options.ability);
```
arbron:
```suggestion
for ( const key of ["ability", "ammunition", "attackMode", "mastery"] ) {
if ( !rolls[0].options[key] ) continue;
foundry.utils.setProperty(messageConfig.data, `flags.dnd5e.roll.${key}`, rolls[0].options[key]);
}
```

module/data/activity/attack-data.mjs
```diff
@@ -253,18 +285,23 @@ export default class BaseAttackActivityData extends BaseActivityData {
...
    * @param {string} [config.situational]
    * @returns {{ data: object, parts: string[] }}
    */
-  getAttackData({ ammunition, attackMode, situational }={}) {
+  getAttackData({ ammunition, ability, attackMode, situational }={}) {
     const rollData = this.getRollData({ roll: { attackMode } });
+    if ( ability && (ability in CONFIG.DND5E.abilities) ) {
+      rollData.roll.ability = ability;
+      rollData.mod = this.actor?.system.abilities?.[ability]?.mod ?? 0;
+    }
```
arbron: Shouldn't need to set `rollData.mod` here because that is handled below in `constructParts`.

```suggestion
    if ( ability && (ability in CONFIG.DND5E.abilities) ) rollData.roll.ability = ability;
```

module/data/activity/attack-data.mjs
```diff
@@ -282,8 +319,22 @@ export default class BaseAttackActivityData extends BaseActivityData {
...
+   */
   getDamageConfig(config={}, options={}) {
+    const ability = config.ability;
+    let { rollData } = options;
+    rollData ??= ability !== undefined ? this.getRollData({ roll: { attackMode: config.attackMode } }) : null;
+    if ( rollData ) {
+      rollData.roll ??= {};
+      rollData.roll.ability = ability;
+      rollData.mod = ability in CONFIG.DND5E.abilities ? this.actor?.system.abilities?.[ability]?.mod ?? 0 : null;
+    }
```
arbron: Rather than processing this here, we should probably add a new option to `ActivityRollDataOptions` to provide the ability. `BaseActivity#getDamageConfig` could then check for `ability` in the `DamageRollProcessConfiguration` and pass that in to `getRollData` if it is set.

We'd then change `BaseActivity#getRollData` to use that option if it exists when getting the roll data:

```javascript
rollData.mod = this.actor?.system.abilities?.[options.ability ?? this.ability]?.mod ?? 0;
```

module/documents/activity/attack.mjs
```diff
@@ -255,10 +262,12 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
    */
   _buildAttackConfig(process, config, formData, index) {
     const ammunition = formData?.get("ammunition") ?? process.ammunition;
+    const ability = formData?.get("ability") ?? process.ability;
     const attackMode = formData?.get("attackMode") ?? process.attackMode;
     const mastery = formData?.get("mastery") ?? process.mastery;
 
-    let { parts, data } = this.getAttackData({ ammunition, attackMode });
+    process.ability = ability;
+    let { parts, data } = this.getAttackData({ ammunition, ability, attackMode });
```
arbron:
```suggestion
    let { parts, data } = this.getAttackData({ ability, ammunition, attackMode });
```

arbron (changes requested):

arbron:
@thatlonelybugbear Reworked this a bit to use a non-persisted field which allows for simplifying the active effect stuff a fair bit. Let me know if you have any thoughts on this approach.

thatlonelybugbear:
The latest conflicts should be resolved now too

arbron (approved):

Fyorl (approved):
Thanks for taking care of this. I think we are interested in https://github.com/thatlonelybugbear/dnd5e/pull/1 if it still makes sense after #6145 (I have not re-evaluated)

thatlonelybugbear:
> Thanks for taking care of this. I think we are interested in [thatlonelybugbear#1](https://github.com/thatlonelybugbear/dnd5e/pull/1) if it still makes sense after #6145 (I have not re-evaluated)

I will rebase and take a look into what might be missing

Issue #1246, opened 2021-08-18
Add visual indicator when certain skill or save has advantage or disadvantage


aaclayton:
Originally in GitLab by @MaxPat931

Related to ongoing work at https://gitlab.com/foundrynet/dnd5e/-/merge_requests/323 / https://gitlab.com/foundrynet/dnd5e/-/issues/1176

**Add the option to flag a skill/save/check as having Advantage or Disadvantage**, could be used to create a visual reminder on the character sheet, and fast-forward rolls (or allow rolling modules to use the field for that purpose)

Use Cases - Heavy armor grants disadvantage to Stealth checks, the Rune Knight's Cloud Rune's passive abilities give advantage on DEX (Sleight of Hand) checks and CHA (Deception) checks.

![image](https://gitlab.com/foundrynet/dnd5e/uploads/ae817e1309da9c3234507007e1b0b074/image.png)

**Add the option for a Chat flavor text field, similar to items**

Use Cases - Reminders for situational bonuses e.g. War Caster's "You have advantage on Constitution saving throws that you make to maintain your concentration on a spell when you take damage."

![image](https://gitlab.com/foundrynet/dnd5e/uploads/7bacb319d82ea517272090a7511a7f1f/image.png)

Rabulias:
Possibly related request: Ticket `[1154](https://github.com/foundryvtt/dnd5e/issues/1154)`

revilowaldow:
Now that #809 is complete this should be much easier to implement?

MaxPat931:
Advantage tracking is now implemented via `roll.mode` for concentration, death saves, initiative and skills. 
It would be nice to add `roll` data to abilities for checks and saves, as well as attacks.
Perhaps expanding the `bonuses` property would work if the existing paths are migrated to a new `.value` and adding a `.roll`?
```
system.bonuses.abilities.check
			      .value
			      .roll.min/max/mode
system.bonuses.abilities.save
			     .value
			     .roll.min/max/mode
system.bonuses.abilities.skill
			      .value
			      .roll.min/max/mode
system.bonuses.XXak.attack
			  .value
			  .roll.min/max/mode
```
Not sure how it would best be structured in the `abilities` since it needs to be split for checks and saves and putting it into the `bonuses` here conflicts with how `skills` are structured so that could be confusing, maybe split the `roll` property?
```
system.abilities.XXX.roll
			 .check
			       .min/max/mode
			 .save
			      .min/max/mode
```

[...]

PR #7081, opened 2026-06-08
[#1246] Show roll advantage indicators on PC character sheets

thatlonelybugbear:
A blast from the past! Should be closing #1246 

image


thatlonelybugbear:
The CSS is a bit fiddly...

thatlonelybugbear:
The Initiative advantage is not an issue. I had changed the ability to Strength to test abilities combining advantageModes😅 

image

Kapuzenjoe:
The icon probably still needs a bit of CSS polishing (see screenshot). I think it shouldn't stack vertically but should try to stay in a single row whenever possible.

That said, the more important part would be being able to identify the source of the Advantage Mode via the tooltip (similar to the issue #5175). It would also be helpful to surface that information again later in the roll dialog as an additional hint, although that's a separate feature request.


image

thatlonelybugbear:
If the system introduces some sort of registry with metadata for each source of advantageMode then the expansion for the tooltips holding more info would be doable. 

I tried to make the indicators follow the general layout of their surroundings and blend in with a minimum footprint.  Personally, I didn't like the result when I tested positioning them on top of the modifiers. Felt a bit out of place, but I will keep tweaking based on feedback!

Kapuzenjoe:
Sure, the source metadata/registry would definitely need to exist first.

I actually experimented with something similar in a private module recently. Based on the `AdvantageModeField` class, I put together a patch that extends the roll mode counting logic with source tracking, so each contribution can carry information about where it came from.

I could clean that up and open a real PR later today. It might even provide a possible solution for #5175.

Issue #1379, opened 2021-12-23
New text popup feature needs configuration options


aaclayton:
Originally in GitLab by @BlackMoons

Hi. The new text popup feature for damage/status effects really needs configuration options.

Like how far up it scrolls (Animation options?), how long it stays on screen, font size, if it scales based on damage. If it shows damage, or status effects, or both. 

Personally I think the animation right now is very poor for RPG damage numbers. I much prefer the 'pop up and fall a small distance then stop' sorta animation that RPG's often use.

I tried asking for this feature in #V9-feedback on discord and they said to raise the issue here since the API apparently supports configuration to some degree, and foundry devs do not want to create a core configuration menu.

Issue #1302, opened 2021-09-21
Allow more fine-tuned control of the v9 Scrolling Damage Numbers


aaclayton:
Originally in GitLab by @akrigline

As a follow up to !387, we should add some way to turn off or control the visibility of this specific implementation of scrolling status text.

Similar to how token healthbars are not always visible for all players, the GM might want to hide the scrolling damage numbers from their players.

I propose the following system settings:

> [ ] Display Scrolling Damage
> 
> Allow Players to see Scrolling Damage?
> - All Actors
> - Characters Only
> - Never

This would allow a GM to turn off only 5e's implementation of Scrolling Text (but leave anything a module does in-tact for instance). It would also allow a GM to only show the PCs their characters' bars, or turn it off entirely for Players.

Another option might be "Allies" as determined by token disposition (this would allow the numbers for summons for instance).

I think an even more ideal solution would be to have this tied to resource bar visibility, as that would allow even more fine-tuned control on a per-monster basis, but that might be a bit too far.

Rabulias:
Possibly related suggestion: Ticket [1379](https://github.com/foundryvtt/dnd5e/issues/1379)

Rabulias:
Another related issue: Ticket [2509](https://github.com/foundryvtt/dnd5e/issues/2509)

guldelox:
It would also be nice if we could obscure certain types of health modification from showing up as scrolling status text. Like if I decide I need to do some in-flight tuning to a monster and stealthily raise or lower its max HP, that shows up as damage or healing in the scrolling text for players. As a result I just permanently have this otherwise pretty useful feature disabled.

PR #7084, opened 2026-06-08
[#2509] Scale scrolling combat text to the grid size

### LINKED ISSUE #2509: Floating damage numbers can be hard to spot on high grid sizes
Always being roughly 16px, this can make them nearly invisible on scenes with a high grid size. Suggest making them relative instead.

Sayshal:
Closes #2509

Scales the scrolling damage/healing text to the scene's grid size so it stays legible on large-grid scenes.

On the scaling math: linear scaling `(font * grid/100)` keeps the text a constant fraction of the token, but that 4x's it on a 400px grid (up to 192px), which reads as oversized since you're zoomed out on those scenes. Square-root scaling grows the text with the grid but with diminishing returns, so big grids get readable numbers without runaway size. The HP-fraction emphasis is preserved, just multiplied by the grid factor.

Re #1302 and #1379: I'd suggest closing both. They ask for visibility toggles and animation/duration/font configuration for the scrolling text, and that level of purely cosmetic customization feels outside the system's scope (good module territory). Happy to reconsider if there's appetite for this customization in the system.


module/documents/actor/actor.mjs
```diff
@@ -3481,14 +3481,17 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     const pct = Math.clamp(Math.abs(value) / this.system.attributes.hp.max, 0, 1);
     const fill = CONFIG.DND5E.tokenHPColors[key];
 
+    // Adapt the font size relative to the Actor's HP total to emphasize more significant blows, scaled to the grid
+    const gridScale = Math.sqrt((canvas?.dimensions?.size ?? 100) / 100);
+    const fontSize = (16 + (32 * pct)) * gridScale;
```
arbron, line 3486: Seems like we may want to also take `canvas.dimensions.uiScale` into account here.

Sayshal: Honestly, after reconsidering the uiScale, I don't think the original issue is a concern anymore. on the base 6.0.x I tested at 100px, 200px, 400px and the readability seemed total normal to me. This might be a wont-fix?

krbz999: I think there are still some improvements to be made, but the issue was made in 2023 and lots of things have likely changed. Back in '23, the ui scale didn't even exist.

Sayshal: Open to suggestions!

PR #7085, opened 2026-06-08
[#2469] Warn on duplicate scale value identifiers

### LINKED ISSUE #2469: Show a warning on the class item sheet if scale values with identical identifiers exist

Sayshal:
Adds a warning on the item sheet when two scale value advancements share an identifier.

Rather than invent a one-off warning UI for the item sheet, this reuses the actor sheet's existing preparation-warnings mechanism verbatim: warnings are computed on the document during data prep (`item._preparationWarnings`), and the sheet just deep-clones them into context, identical to how `Actor5e` and the actor sheet already work.

Making it share the actor's implementation instead of duplicating it is why the diff spans more than `item-sheet.mjs`:

- **Template rename** (`actors/parts/actor-warnings-dialog.hbs` -> `shared/sheet-warnings-dialog.hbs`)
- **CSS move** (`actors.less` -> `sheets.less`)
- **`item.mjs`**: inits `_preparationWarnings` in `_clearData` and pushes the scale-value warning in `prepareDerivedData` after the advancement collection is built, mirroring `Actor5e.prepareData` and the data-model warning pushes.

Closes #2469


module/applications/item/item-sheet.mjs
```diff
@@ -774,6 +794,49 @@ export default class ItemSheet5e extends PrimarySheetMixin(DocumentSheet5e) {
 
   /* -------------------------------------------- */
 
+  /**
```
arbron: Most of this could be moved to `PrimarySheetMixin` to avoid duplicating between actor & item sheets.

Sayshal:
Rebased and updated the warnings button to use `this.window.subtitle.after(...)` per abb4e8c.

Fyorl (approved):

PR #7088, opened 2026-06-08
[#4289] Add Macro activity type

### LINKED ISSUE #4289: New Activity Type "Macro"
(Moving this to its own issue as suggested by another user.)

The main idea is that executing the activity would execute a macro. While the initial suggestion is for it to be a new activity type, this could also just be an expansion to the "Utility" type of activity if that is easier, more convenient, and/or is more user friendly.

In theory, it should have a text field where you are able to put the macro data in and the option to make it a chat macro or script macro. Then it just runs when the activity is executed.

Sayshal:
A Macro activity attaches a Foundry Macro to an item and runs it through the standard activity usage lifecycle. The macro is referenced by UUID (mirrors the Cast activity's spell reference) so Foundry's permission gating applies.

By default it posts an "Execute Macro" chat button; toggling that off runs the macro automatically on use. Either way the macro receives `actor`, `token`, `item`, `activity`, `targets`, `event`, and the usage `message` in scope (plus core's auto `speaker`/`character`).

msedge_GHibIUInVP

Closes #4289


Sayshal:
Rebased.

PR #7089, opened 2026-06-08
[#4256] Clarify Bloodied Status setting labels

### LINKED ISSUE #4256: Considerations for the `Bloodied Status` setting text
The `Bloodied Status` setting could use a bit of clarification.

CleanShot 2024-09-15 at 11 32 12

It appears to consider "Allies" as any token with the `Friendly` disposition, and "Enemies" as any token with either the `Hostile` or `Neutral` dispositions.

The clearest text here would be to actually say "Display for Friendly, Neutral, and Hostile tokens" and "Only display for Friendly tokens". If we want to use the "Ally/Enemy" language, then there ought to be hint text that elaborates on what this actually means, in terms of token configuration.

Sayshal:
Closes #4256

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -5874,8 +5874,8 @@
   "BLOODIED": {
     "Name": "Bloodied Status",
     "Hint": "Configure whether the bloodied status is tracked automatically and its visibility.",
-    "All": "Display for Allies & Enemies",
-    "Player": "Only Display for Allies",
+    "All": "Display for Friendly, Neutral & Hostile Tokens",
+    "Player": "Only Display for Friendly Tokens",
     "None": "Never Display"
   },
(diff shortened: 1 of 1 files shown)


lang/en.json
```diff
@@ -5874,8 +5874,8 @@
   "BLOODIED": {
     "Name": "Bloodied Status",
     "Hint": "Configure whether the bloodied status is tracked automatically and its visibility.",
-    "All": "Display for Allies & Enemies",
-    "Player": "Only Display for Allies",
+    "All": "Display for Friendly, Neutral & Hostile Tokens",
```
```suggestion
    "All": "Display for Friendly, Neutral, & Hostile Tokens",
```

arbron (changes requested):

arbron (approved):

PR #7092, opened 2026-06-08
[#3991] Remove stray whitespace around the award enricher button

### LINKED ISSUE #3991: Award enricher in journals has incorrect character spacing
When using the "award" enricher in journal pages, the character spacing for the award button is awkward.

For example the following enricher:
image

Produces this result:
image

It seems there is an extra whitespace character added after the "200 XP" text and before the "Award" button, and no whitespace character after the button, before the word "equally".

The extra whitespace after "200 XP" is highlighted here for visibility.
image

Moreover, adding extra whitespaces after the enricher, before the word "equally" do not produce visible whitespaces as a result, they are all trimmer and it looks like the "Award" button is immediately followed by the word "equally" with no whitespace.

Sayshal:
Before:
msedge_mEiw25tP3m

After:
msedge_AznPIsZKss

Closes #3991

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -316,9 +316,9 @@ export async function enrichAward(config, label, options) {
     entries.push(`<span class="award-entry">${amount} <i class="currency ${
       key}" data-tooltip aria-label="${label}"></i></span>`);
   }
-  if ( parsed.xp ) {
-    entries.push(`<span class="award-entry">${formatNumber(parsed.xp)} ${_loc("DND5E.ExperiencePoints.Abbreviation")}</span>`);
-  }
+  if ( parsed.xp ) entries.push(
+    `<span class="award-entry">${formatNumber(parsed.xp)} ${_loc("DND5E.ExperiencePoints.Abbreviation")}</span>`
+  );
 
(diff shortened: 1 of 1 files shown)


module/enrichers.mjs
```diff
@@ -313,27 +313,17 @@ export async function enrichAward(config, label, options) {
   for ( let [key, amount] of Object.entries(parsed.currency) ) {
     const label = CONFIG.DND5E.currencies[key].label;
     amount = Number.isNumeric(amount) ? formatNumber(amount) : amount;
-    entries.push(`
-      <span class="award-entry">
-        ${amount} <i class="currency ${key}" data-tooltip aria-label="${label}"></i>
-      </span>
-    `);
+    entries.push(`<span class="award-entry">${amount} <i class="currency ${key}" data-tooltip aria-label="${label}"></i></span>`);
```
arbron: Line length is a bit long here, you can split inside one of the expressions:


```suggestion
    entries.push(`<span class="award-entry">${amount} <i class="currency ${
      key}" data-tooltip aria-label="${label}"></i></span>`);
```

arbron (changes requested):

arbron (approved):

PR #7095, opened 2026-06-08
[#6840] Include the source document UUID in enricher parse warnings

### LINKED ISSUE #6840: Provide uuid of document when logging a warning about an enricher not being able to be parsed
in `enrichSave` for example, the attempt input is logged but it isnt possible to know what the context is.

the document is in the `options` object and it'd be nice to have that logged as well.

Image

Sayshal:
Logs a warning:
```js
No ability or linked activity found while enriching [[/save]] (JournalEntry.3JLFXDrVakm66woM.JournalEntryPage.cieXHDWWiWeXmaxF).
```

Closes #6840

### DIFF
--- a/module/enrichers.mjs
+++ b/module/enrichers.mjs
@@ -47,7 +47,6 @@ export async function enrichString(match, options) {
   config = parseConfig(config, { multiple: ["damage", "heal", "healing"].includes(type) });
   config._input = match[0];
   config._rules = getRulesVersion(config, options);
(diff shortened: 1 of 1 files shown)


Fyorl (changes requested):
`config._input` is supposed to be the original input string. It shouldn't be used for this purpose of supplementing logging.

arbron (approved):

PR #7097, opened 2026-06-09
[#4894] Let wide class feature tables scroll horizontally

### LINKED ISSUE #4894: Add horizontal scroll to journal tables
## Problem
With default journal width, some content in tables in PHB 2024 is not visible. See example below: 
Screenshot 2024-12-16 at 10 29 52 PM

## Proposed solution
```css
table.features-table {
    /* width: fit-content; <--- this need to be removed */
    display: block;
    overflow-x: auto;
}
```

Sayshal:
- Add `display: block` and `overflow-x: auto` to `.features-table` so wide tables scroll instead of clipping

Closes #4894


less/v2/journal.less
```diff
@@ -318,6 +318,8 @@
   }
 
   .features-table {
+    display: block;
```
arbron, line 322: This causes the table background to appear under the caption.

Image

Image

Looks like we have an explicit background set on all rows, so it should be safe to just unset the overall table background:


```suggestion
    background: none;
    display: block;
```

arbron (changes requested):

PR #7098, opened 2026-06-09
[#4259] Add Remove Enchantment option to the inventory context menu

### LINKED ISSUE #4259: Add 'Remove Enchantment' option to context menu of enchanted items
Having an option in the context menu would be a convenient way to remove an enchantment, as you would otherwise need to pop out the enchantment chat card or edit the enchanted item to remove the effect, as you can no longer view the enchanted items in the original enchanting item.

image

Sayshal:
- Add a state-group context option, visible when the item has an applied enchantment
- Dispatch a removeEnchantment action that deletes all applied enchantments from the item

Closes #4259


module/applications/components/inventory.mjs
```diff
@@ -575,6 +582,19 @@ export default class InventoryElement extends (foundry.applications.elements.Ado
 
   /* -------------------------------------------- */
 
+  /**
+   * Handle removing all applied enchantments from an item.
+   * @param {Item5e} item  The item.
+   * @returns {Promise}
+   * @protected
+   */
+  _onRemoveEnchantment(item) {
```
arbron: Alphabetic order, move this down after `_onOpenContextMenu` :)

module/applications/components/inventory.mjs
```diff
@@ -417,6 +417,12 @@ export default class InventoryElement extends (foundry.applications.elements.Ado
         return canIdentify && canEdit;
       },
       onClick: (event, target) => this._onAction(target, "identify", { event })
+    }, {
+      label: "DND5E.ENCHANTMENT.Action.Remove",
+      icon: '<i class="fa-solid fa-rotate-left fa-fw"></i>',
+      group: "state",
```
arbron: Concerned about the placement of this option since removing enchantments is a destructive action, especially without a confirmation dialog. I can easily see people going into the context menu to add their "+1 Greatsword" to favorites and accidentally turning it into a plain non-magical greatsword.

I'd say we should at the very least move this into a new section between "action" and "state" for destructive actions (along with "Break Concentration"). A confirmation dialog would probably also be nice.

arbron (changes requested):

module/applications/components/inventory.mjs
krbz999: On that note it might be nice if an enchantment could be marked as semi-permanent (perhaps in `system.origin`) specifically for the cases of "template" enchantments used only for item creation, not actual enchantment. And in such a case, there shouldnt be an option to remove it in the context menu.

Zanderaf: +1 on this, sometimes enchantments are useful as a base on an item to make it work in an intended way and removing those semipermanaent/permanent ones would be a disastor

Sayshal: > Concerned about the placement of this option since removing enchantments is a destructive action, especially without a confirmation dialog. I can easily see people going into the context menu to add their "+1 Greatsword" to favorites and accidentally turning it into a plain non-magical greatsword.
> 
> I'd say we should at the very least move this into a new section between "action" and "state" for destructive actions (along with "Break Concentration"). A confirmation dialog would probably also be nice.

Moved it into a new category. I think if we want an AreYouSure dialog we should also do it to other destructive actions, such as the aforementioned `Break Concentration`. I'll open an issue for this as a seperate action to sweep other destructives.

Sayshal: @arbron I have also added a seperate commit that captures what @krbz999 and @Zanderaf have requested.

Sayshal:
Rebased.

PR #7102, opened 2026-06-09
[#7021] Set `keepId: false` when autoimporting a compendium actor

### LINKED ISSUE #7021: Summoning Spells: Multiple players are unable to summon the same token to the board
**Versions:** 
dnd5e: [5.3.3]

When two or more players have the same summoning spell (e.g Mage Hand, Dancing Lights) and no one has cast the spell before, the first player to cast the spell can successfully summon and place the Actor. However, any subsequent player beyond the first cannot summon another Actor, receiving the error message: 
```
foundry.mjs:26732 Error: User Player3 lacks permission to update Actor [phbsplMageHand00]
    at /C:/Program%20Files/…-backend.mjs:1:7981
    at async Promise.all (index 0)
    at async ServerDatabaseBackend._updateDocuments (/C:/Program%20Files/…-backend.mjs:1:7719)
    at async #try (/C:/Program%20Files/…semaphore.mjs:98:17)
```
**Steps to Reproduce:**
1. Create a new world with D&D5e
2. Create two users Player2 and Player3
3. From Compendium > D&D Modern Content > Actors > Premades > Level 1 import Beiro (Bard) and Zanna (Wizard), who both know the spell Mage Hand
4. Give Player2 ownership of Beiro and Player3 ownership of Zanna
5. Allow Players to Create Tokens
6. From the D&D rules, Allow Summoning
7. Log in as Player2
8. From Beiro's character sheet, cast Mage Hand and place the token on the map
9. Log in as Player3
10. From Zanna's character sheet, cast Mage Hand. 
11. Observe the error message above
My rough guess is this is because there is when casting a summoning spell like Mage Hand or Dancing Lights, the backend checks if the Actor exists, and if not then creates it, setting the owner to the token creator. Then when the second user casts the same spell, the backend find the existing Actor, sees the user does not have ownership, and denies the request.

Aioros:
This prevents a conflict when two different players summon the same actor from a compendium.
I'm trying to think of any issue arising from not keeping the original id here and I'm coming up blank. Let me know if I am in fact missing something.

Closes #7021

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -417,7 +417,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       return game.actors.importFromCompendium(game.packs.get(actor.pack), actor.id, {
         "flags.dnd5e.isAutoImported": true,
         folder: game.folders.get(folderId) ?? null
-      });
+      }, { keepId: false });
     } else {
       // A linked world actor was found. Create a copy to avoid affecting the original.
       return actor.clone({


Fyorl (approved):
> I'm trying to think of any issue arising from not keeping the original id here and I'm coming up blank. Let me know if I am in fact missing something.

Not sure off the top of my head.

PR #7127, opened 2026-06-13
[#5360] Add an "Available" spells filter

### LINKED ISSUE #5360: Prepared spells filter shouldn't filter out "Additional spells"
As a wizard, you want to survey your options with the prepared filter active, but currently this removes viable options like spells cast from items.

Sayshal:
- Show spells that can currently be cast without changing the "prepared" filter
- Includes prepared, always-prepared, innate, at-will, and item-granted spells
- Excludes rituals, which keep their own filter

Closes #5360

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -4674,6 +4674,7 @@
   "Unlabeled": "Unlabeled Spellcasting"
 },
 
+"DND5E.SpellAvailable": "Available",
 "DND5E.SpellcastingClass": "{class} Spellcasting",
 "DND5E.SpellComponent": "Spell Component",
 "DND5E.SpellComponents": "Spell Components",
--- a/module/applications/actor/api/base-actor-sheet.mjs
+++ b/module/applications/actor/api/base-actor-sheet.mjs
@@ -412,6 +412,7 @@ export default class BaseActorSheet extends PrimarySheetMixin(
         { key: "reaction", label: "DND5E.Reaction" },
         { key: "concentration", label: "DND5E.Concentration" },
         { key: "ritual", label: "DND5E.Ritual" },
+        { key: "available", label: "DND5E.SpellAvailable" },
         { key: "prepared", label: "DND5E.Prepared" },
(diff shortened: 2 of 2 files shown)


Sayshal:
Rebased and rewrote the `available` filter based on the changes in 6c7b2058b79c22d01162aa34174b86ad9b7c0164, which moved spell filtering to the `passesFilter` include/exclude model.

Fyorl (commented):
I think the concept of a spell that is 'available' to be cast is unfortunately not a simple thing. We capture some of the logic here: innate/at-will, prepared, granted by an item.

But there is other logic that determines whether a spell is available to cast that we can't or don't model:
- Wizards can cast unprepared rituals from their spellbook (all other classes must prepare the ritual).
- If you have no spell slots, a given spell is not available to cast (unless it has some 'free cast' uses).

So the filter would be wrong for a fairly common and defining case of wizards with unprepared rituals.

There was some discussion on the issue about excluding rituals because they can't be cast in combat but that's not really true either. There are spells with a cast time of 1 action that have the ritual property that you could conceivably cast non-ritually in combat just fine.

So I would say that if we want a filter that is called 'available' or 'can cast', it should show spells that are actually available to cast, and I think we shouldn't implement this filter if it's going to hide spells that can actually be cast but that we can't model right now.

If players are really so upset about having to see their unprepared spells I would suggest that a reasonable interim solution would be an 'unprepared' filter option. Now that we have a tri-state filter, you can right-click that option to filter out unprepared spells. Wizards would still filter out their unprepared rituals but the filter is nonetheless completely clear about what it is doing.

Sayshal:
> If players are really so upset about having to see their unprepared spells I would suggest that a reasonable interim solution would be an 'unprepared' filter option. Now that we have a tri-state filter, you can right-click that option to filter out unprepared spells. Wizards would still filter out their unprepared rituals but the filter is nonetheless completely clear about what it is doing.

I think I'll close this PR then, and in the future consider a PR that does this, for the time being.

PR #7130, opened 2026-06-13
5.1 SRD Update: 3rd Level Spells

Zanderaf:
Here we go again

- Spells now have more activities to handle all their effects (Glyph of Warding, Call Lightning, Blink, etc)
- Added AEs for nearly every effect that feels like it would be useful to have.
- Some spell have self-enchants that add in more activities once the spell is active (Call Lightning for example)
- New Spell Icons
- Spell activities have been renamed to have better clarity on what they do (Meld into Stone use to havve 2 activities just called `Damage` for instance)

related #6727


arbron:
Just some initial thoughts that probably applies to all of your spell PRs:
- Avoid changing capitalization, I noticed in Animate Dead that the capitalization of Skeleton and Zombie was changed from the lowercase of the original text to capitalized
- I don't think we want to do any of the self-enchant stuff. We have features planned in the future for activities that are only available when a spell is "active", so we should avoid implementing a stopgap solution and just wait for better system support

Zanderaf:
> Just some initial thoughts that probably applies to all of your spell PRs:
> 
> * Avoid changing capitalization, I noticed in Animate Dead that the capitalization of Skeleton and Zombie was changed from the lowercase of the original text to capitalized
> * I don't think we want to do any of the self-enchant stuff. We have features planned in the future for activities that are only available when a spell is "active", so we should avoid implementing a stopgap solution and just wait for better system support

Got it. I'll go back through the other levels of spells and do those changes as well.

PR #7133, opened 2026-06-13
[#7132] Add options to move embed name to header or inline

### LINKED ISSUE #7132: Add option to move embed caption or citation to header or inline of first paragraph
A common pattern in books is a header followed by an items description (see the spells chapter of any source book) or to list an activity's name bolded inline with the description (usually on things like class features or complex magic items). Currently the first usage generally requires entering a separate header above the embed while the second generally takes a lot of complex CSS trickery that is prone to breaking.

Image

Image

The system should add support for displaying the item's name either as a header or inline using an option directly on the embed enricher:

```
// Existing
<h2>@UUID[Compendium.dnd5e.spells24.Item.phbsplScorchingR]{Scorching Ray}</h2>
<p>@Embed[Compendium.dnd5e.spells24.Item.phbsplScorchingR details caption=false cite=false]</p>

// Proposed
<p>@Embed[Compendium.dnd5e.spells24.Item.phbsplScorchingR details nameStyle=h2 caption=false]</p>
```

If a caption is present and it equals the item name, then that will be moved to the appropriate position. Otherwise if a citation is present it will be moved and its link will remain. If neither is present it will just create a new element. This means that you can use `nameStyle=h2 cite=false` to have a header with a link to the item or `nameStyle=h2 caption=false` to have a header without the link.

arbron:
Adds a new `nameStyle` option to the `@Embed` enricher for actors and items that allows them to place the document's name inline with the first paragraph of text or as a header. This option accepts an HTML header tag (e.g. `h2`) or `inline` as its value.

Embed Name Styles

This change is implemented in a new `EmbeddableDocumentMixin` so that it can be easily be added to Activity and Advancement when those types get embedding support.

### Future Work
- Adding an `nameAction` option for Items and Activities that changes the link to activate the item or activity rather than opening its sheet

Closes #7132

### DIFF
--- a/module/documents/mixins/embeddable.mjs
+++ b/module/documents/mixins/embeddable.mjs
@@ -13,13 +13,15 @@ export default function EmbeddableDocumentMixin(Base) {
       if ( !config.nameStyle ) return element;
 
       // Use citation if present, use caption if it equals the document's name, otherwise create a new element
-      const figcaption = element.querySelector("figcaption:has(cite, .embed-caption)");
-      const citation = figcaption.querySelector(":scope > cite");
-      const caption = figcaption.querySelector(":scope > .embed-caption");
+      const figcaption = element.querySelector(":scope > figure.content-embed > figcaption:has(cite, .embed-caption)");
+      const citation = figcaption?.querySelector(":scope > cite")
+        ?? element.querySelector(":scope > figure.content-embed > cite");
+      const caption = figcaption?.querySelector(":scope > .embed-caption");
       const originalElement = caption && (caption.innerText === this.name) ? caption : citation;
 
       // Headers use tag provided, inline uses <span>, any others are invalid
-      const tagName = config.nameStyle.match(/h\d/i) ? config.nameStyle : config.nameStyle === "inline" ? "span" : null;
+      const nameStyle = String(config.nameStyle);
+      const tagName = /^h[1-6]$/i.test(nameStyle) ? nameStyle : nameStyle === "inline" ? "span" : null;
       if ( !tagName ) return element;
 
       // Create the new name element
@@ -33,11 +35,13 @@ export default function EmbeddableDocumentMixin(Base) {
 
       // Place headers as the first child, place inline names inside first paragraph
       let prepend = element.querySelector("figure.content-embed") ?? element;
-      if ( config.nameStyle === "inline" ) prepend = element.querySelector(".content-embed p") ?? prepend;
+      if ( config.nameStyle === "inline" ) {
+        prepend = element.querySelector(":scope > figure.content-embed p") ?? prepend;
+      }
       prepend.prepend(nameElement);
 
       // Remove original <figcaption> if empty
-      if ( !figcaption.hasChildNodes() ) figcaption.remove();
+      if ( figcaption && !figcaption.hasChildNodes() ) figcaption.remove();
 
       return element;
     }
--- a/module/documents/mixins/embeddable.mjs
+++ b/module/documents/mixins/embeddable.mjs
@@ -43,7 +43,8 @@ export default function EmbeddableDocumentMixin(Base) {
       // Place headers as the first child, place inline names inside first paragraph
       let prepend = element.querySelector("figure.content-embed") ?? element;
       if ( config.nameStyle === "inline" ) {
-        prepend = element.querySelector(":scope > figure.content-embed p") ?? prepend;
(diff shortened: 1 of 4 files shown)


module/documents/mixins/embeddable.mjs
```diff
@@ -0,0 +1,46 @@
...
+ */
+export default function EmbeddableDocumentMixin(Base) {
+  class EmbeddableDocument extends Base {
+    /** @inheritDoc */
+    async _createFigureEmbed(content, config, options) {
+      const element = await super._createFigureEmbed(content, config, options);
+      if ( !config.nameStyle ) return element;
+
+      // Use citation if present, use caption if it equals the document's name, otherwise create a new element
+      const figcaption = element.querySelector("figcaption:has(cite, .embed-caption)");
```
Fyorl: Avoid hitting nested embeds.
```suggestion
      const figcaption = element.querySelector(":scope > figure.content-embed > figcaption:has(cite, .embed-caption)");
```
Also, if this doesn't match (`cite=false caption=false`) then we throw later.

module/documents/mixins/embeddable.mjs
```diff
@@ -0,0 +1,46 @@
...
+      nameElement.classList.add("embed-name");
+      if ( originalElement ) {
+        nameElement.replaceChildren(...originalElement.childNodes);
+        originalElement.remove();
+      }
+      else nameElement.innerText = this.name;
+
+      // Place headers as the first child, place inline names inside first paragraph
+      let prepend = element.querySelector("figure.content-embed") ?? element;
+      if ( config.nameStyle === "inline" ) prepend = element.querySelector(".content-embed p") ?? prepend;
```
Fyorl:
```suggestion
      if ( config.nameStyle === "inline" ) prepend = element.querySelector(":scope > figure.content-embed > p") ?? prepend;
```

module/documents/mixins/embeddable.mjs
```diff
@@ -0,0 +1,46 @@
...
+      if ( !config.nameStyle ) return element;
+
+      // Use citation if present, use caption if it equals the document's name, otherwise create a new element
+      const figcaption = element.querySelector("figcaption:has(cite, .embed-caption)");
+      const citation = figcaption.querySelector(":scope > cite");
+      const caption = figcaption.querySelector(":scope > .embed-caption");
+      const originalElement = caption && (caption.innerText === this.name) ? caption : citation;
+
+      // Headers use tag provided, inline uses <span>, any others are invalid
+      const tagName = config.nameStyle.match(/h\d/i) ? config.nameStyle : config.nameStyle === "inline" ? "span" : null;
```
Fyorl:
```suggestion
      const nameStyle = String(config.nameStyle);
      const tagName = /^h[1-6]$/i.test(nameStyle) ? nameStyle : nameStyle === "inline" ? "span" : null;
```

module/documents/mixins/embeddable.mjs
```diff
@@ -0,0 +1,46 @@
...
+export default function EmbeddableDocumentMixin(Base) {
+  class EmbeddableDocument extends Base {
+    /** @inheritDoc */
+    async _createFigureEmbed(content, config, options) {
+      const element = await super._createFigureEmbed(content, config, options);
+      if ( !config.nameStyle ) return element;
+
+      // Use citation if present, use caption if it equals the document's name, otherwise create a new element
+      const figcaption = element.querySelector("figcaption:has(cite, .embed-caption)");
+      const citation = figcaption.querySelector(":scope > cite");
```
Fyorl: `<cite>` is moved out of the `<figcaption>` if `captionPosition=top`.

Fyorl (changes requested):

module/documents/mixins/embeddable.mjs
arbron: I'm thinking we may not want to find a direct `<p>` child in cases where the inner text is surrounded by a `<div>`.

Fyorl: I think you will hit nested embeds without it but I didn't test. Also for an inline spell, the name ends up inside the `.item-entry-tag`.

```
@Embed[Compendium.dnd5e.spells24.Item.phbsplScorchingR details nameStyle=inline caption=false cite=false]
```

image

Fyorl (approved):
See my comment but this seems good to me, you can merge it when you're happy with it.

PR #7144, opened 2026-06-15
[#5139] Allow checking prerequisites within compendium browser

### LINKED ISSUE #5139: Add support for feature restrictions with user feedback in compendium browser
When selecting a feature such as an Eldritch Invocation through the compendium browser, there should be a method for displaying whether the user qualifies for that feature and if they don't tell them why not. This could be handled by an additional callback passed through the compendium browser that is called for each index/item. This callback would return a status (`approved`, `rejected`, `neutral`) as well as an optional explanation (e.g. "You need the Fighting Style class feature to take this fighting style feat.").

The compendium browser could then display a checkmark or x-mark next to each item to make it clear what can be selected, invalid items can have their checkboxes disabled, and mousing over the icon can display the explanation.

Would help resolve part of https://github.com/foundryvtt/dnd5e/issues/4686

arbron:
Introduces some new options to the compendium browser options that allow passing in a method for each item to determine whether it can be selected or not. This callback returns a map with a `valid` value that can be `true`, `false`, or `null` as well as a label. The compendium browser will display a checkmark if all conditions are valid, x-mark if any condition is invalid, or a question mark if there conditions that couldn't not be checked fully (this is to allow conditions that players should take into account but cannot be fully evaluated in code, such as certain feats that require a specific faction or to be part of a specific campaign).

Compendium Browser Prerequisites

The prerequisite checking on features has been modified with a new method that returns this mapping for each prerequisite defined on the feature item.

Uses can hover over the row in the compendium browser to view more details on the evaluation in the tooltip. Rather than adding a new tooltip, this adds support for `data-tooltip-extras` which adds more information into an existing item tooltip.

In order to produce human-readable labels for the items conditions on features, a new `IdentifierRegistry` has been added. This scans all items in compendiums on world load and creates a mapping to those item's localized names.

```javascript
dnd5e.registry.identifiers.get("pact-of-the-blade"); // Pact of the Blade
dnd5e.registry.identifiers.get("spell:blade-ward"); // Blade Ward
dnd5e.registry.identifiers.get("cleric", { type: "class" }); // Cleric
```

Closes #5139

### DIFF
--- a/module/applications/_types.mjs
+++ b/module/applications/_types.mjs
@@ -12,6 +12,8 @@
  * @property {number|null} xp                        Amount of experience points to award.
  */
 
+/* -------------------------------------------- */
+
 /**
  * @typedef CurrencyUpdateOptions
  * @property {boolean} [exact=true]                    Prioritize deducting the requested denomination first.
@@ -30,7 +32,11 @@
  *                                              Locked filters won't be able to be changed by the user. Initial filters
  *                                              will be set to start but can be changed.
  * @property {string|null} hint                 Hint displayed in the interface.
- * @property {CompendiumBrowserSelectionConfiguration} selection  Configuration used to define document selections.
+ * @property {CompendiumBrowserSelectionConfiguration} selection     Configuration used to define document selections.
+ * @property {object} prerequisites
+ * @property {boolean} prerequisites.enforce        Prevent selection of results that fail validation.
+ * @property {boolean} prerequisites.fullDocuments  Require the full documents to be fetched for evaluation.
+ * @property {ValidatePrerequisitesCallback} prerequisites.validate  Callback used to evaluate prerequisites.
  */
 
 /**
@@ -80,6 +86,25 @@
  *                                   class are assumed.
  * @property {boolean} [advanced]    Is this tab only available in the advanced browsing mode.
  */
+
+/**
+ * @typedef PrerequisiteValidationResult
+ * @param {string} label           Label describing the prerequisite.
+ * @param {boolean} [quiet=false]  Only display in mark compendium browser if invalid.
+ * @param {boolean|null} valid     Whether the prerequisite is valid or not, or `null` if it couldn't be evaluated.
+ */
+
+/**
+ * @typedef {Map<string, PrerequisiteValidationResult>} PrerequisiteValidationResults
+ */
+
+/**
+ * @callback ValidatePrerequisitesCallback
+ * @param {Document|object} doc  Document for which to evaluate the prerequisites.
+ * @returns {PrerequisiteValidationResults}
+ */
+
+/* -------------------------------------------- */
 
 /**
  * Description for a single part of a property attribution.
--- a/module/applications/_types.mjs
+++ b/module/applications/_types.mjs
@@ -89,9 +89,9 @@
 
 /**
  * @typedef PrerequisiteValidationResult
- * @param {string} label           Label describing the prerequisite.
- * @param {boolean} [quiet=false]  Only display in mark compendium browser if invalid.
- * @param {boolean|null} valid     Whether the prerequisite is valid or not, or `null` if it couldn't be evaluated.
+ * @property {string} label           Label describing the prerequisite.
+ * @property {boolean} [quiet=false]  Only display in mark compendium browser if invalid.
+ * @property {boolean|null} valid     Whether the prerequisite is valid or not, or `null` if it couldn't be evaluated.
  */
 
 /**
@@ -101,7 +101,7 @@
 /**
  * @callback ValidatePrerequisitesCallback
  * @param {Document|object} doc  Document for which to evaluate the prerequisites.
- * @returns {PrerequisiteValidationResults}
+ * @returns {PrerequisiteValidationResults|void}
  */
 
 /* -------------------------------------------- */
--- a/module/applications/advancement/item-choice-flow.mjs
+++ b/module/applications/advancement/item-choice-flow.mjs
@@ -349,6 +349,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
 
     try {
       this.advancement._validateItemType(item);
+      item.system.assertPrerequisites?.(this.advancement.actor, { level: this.featureLevel, throwError: true });
     } catch(err) {
       ui.notifications.error(err.message);
       return null;
--- a/module/applications/advancement/item-choice-flow.mjs
+++ b/module/applications/advancement/item-choice-flow.mjs
@@ -233,7 +233,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
       return;
     }
 
-    const filters = { locked: { additional: {}, documentClass: "Item" } };
+    const filters = { locked: { additional: {}, documentClass: "Item", exclusive: true } };
 
     // Apply restrictions based on type
     if ( config.type ) {
@@ -273,7 +273,14 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
     }
 
     const result = await CompendiumBrowser.select({
-      filters, selection: { min: 1, max: max - current }
+      filters,
+      prerequisites: {
+        validate: item => {
+          if ( !item.system.prerequisiteLabels ) return [];
+          return item.system.prerequisiteLabels({ actor: this.advancement.actor, level: this.level });
+        }
+      },
+      selection: { min: 1, max: max - current }
     }, this.manager?._detachOptions());
     if ( !result?.size ) return;
 
--- a/module/applications/advancement/item-choice-flow.mjs
+++ b/module/applications/advancement/item-choice-flow.mjs
@@ -154,7 +154,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
       items: [...this.pool, ...dropped].reduce((arr, item) => {
         const { id, name, img } = item;
         const uuid = item.flags.dnd5e?.sourceId ?? item.uuid;
-        const validFeature = !item.system.validatePrerequisites || (item.system.validatePrerequisites(
+        const validFeature = !item.system.assertPrerequisites || (item.system.assertPrerequisites(
           this.advancement.actor, { added, removed, level: this.featureLevel }
         ) === true);
         const validSpell = !validateSpellLevel
@@ -276,8 +276,8 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
       filters,
       prerequisites: {
         validate: item => {
-          if ( !item.system.prerequisiteLabels ) return;
-          return item.system.prerequisiteLabels({ actor: this.advancement.actor, level: this.featureLevel });
+          if ( !item.system.validatePrerequisites ) return;
+          return item.system.validatePrerequisites({ actor: this.advancement.actor, level: this.featureLevel });
         }
       },
       selection: { min: 1, max: max - current }
--- a/module/applications/advancement/item-choice-flow.mjs
+++ b/module/applications/advancement/item-choice-flow.mjs
@@ -276,8 +276,8 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
       filters,
       prerequisites: {
         validate: item => {
-          if ( !item.system.prerequisiteLabels ) return [];
-          return item.system.prerequisiteLabels({ actor: this.advancement.actor, level: this.level });
+          if ( !item.system.prerequisiteLabels ) return;
+          return item.system.prerequisiteLabels({ actor: this.advancement.actor, level: this.featureLevel });
         }
       },
       selection: { min: 1, max: max - current }
--- a/module/applications/compendium-browser.mjs
+++ b/module/applications/compendium-browser.mjs
@@ -1,6 +1,6 @@
 import * as Filter from "../filter.mjs";
 import SourceField from "../data/shared/source-field.mjs";
-import { getPluralRules, loadingTooltip } from "../utils.mjs";
+import { bulkFromUuid, getPluralRules, loadingTooltip } from "../utils.mjs";
 import Application5e from "./api/application.mjs";
 import CompendiumBrowserSettingsConfig from "./settings/compendium-browser-settings.mjs";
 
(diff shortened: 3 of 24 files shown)


module/applications/compendium-browser.mjs
```diff
@@ -653,9 +670,27 @@ export default class CompendiumBrowser extends Application5e {
...
+    if ( this.options.prerequisites.validate ) {
+      const results = this.options.prerequisites.validate(entry);
+      if ( results.size ) {
+        context.prerequisites = results.values().reduce((r, result) => {
+          if ( result.valid === false) {
+            r.disabled = this.options.prerequisites.enforce;
+            r.invalid = true;
+            r.valid = false;
+          } else if ( result.valid === null ) {
+            r.indeterminate = true;
```
Fyorl, line 686: We never set `valid: false` here. In the template we have `ifThen valid 'check' (ifThen invalid 'xmark' 'question')` so an `indeterminate` result actually displays as green checkmark. I guess we just need to set `valid = false` here? Or in the template we can check `indeterminate` explicitly.

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -269,7 +269,14 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
     }
 
     const result = await CompendiumBrowser.select({
-      filters, selection: { min: 1, max: max - current }
+      filters,
+      prerequisites: {
+        validate: item => {
+          if ( !item.system.prerequisiteLabels ) return [];
+          return item.system.prerequisiteLabels({ actor: this.advancement.actor, level: this.level });
```
```suggestion
          return item.system.prerequisiteLabels({ actor: this.advancement.actor, level: this.featureLevel });
```

module/registry.mjs
```diff
@@ -141,6 +147,81 @@ class EnchantmentRegisty {
...
+      Hooks.once("babele.ready", () => this.initialize());
+      return;
+    } else if ( !game.ready ) {
+      Hooks.once("ready", () => this.initialize());
+      return;
+    }
+    this.#status = STATUS_STATES.LOADING;
+
+    const indexes = await CompendiumBrowser.fetch(Item, {
+      // types: new Set([this.#itemType]),
```
Fyorl: errant comment

module/registry.mjs
```diff
@@ -141,6 +147,81 @@ class EnchantmentRegisty {
...
+      indexFields: new Set(["system.identifier"]),
+      sort: false
+    });
+    for ( const item of indexes ) {
+      const identifier = item.system?.identifier ?? formatIdentifier(item.name);
```
Fyorl: `system.identifier` is `required: true, blank: true` so a missing identifier is `""`. Maybe not since we're reading the index but I wouldn't chance it.
```suggestion
      const identifier = item.system?.identifier || formatIdentifier(item.name);
```

module/applications/_types.mjs
```diff
@@ -79,6 +85,25 @@
  * @property {boolean} [advanced]    Is this tab only available in the advanced browsing mode.
  */
 
+/**
+ * @typedef PrerequisiteValidationResult
+ * @param {string} label           Label describing the prerequisite.
+ * @param {boolean} [quiet=false]  Only display in mark compendium browser if invalid.
+ * @param {boolean|null} valid     Whether the prerequisite is valid or not, or `null` if it couldn't be evaluated.
```
Fyorl:
```suggestion
 * @property {string} label           Label describing the prerequisite.
 * @property {boolean} [quiet=false]  Only display in mark compendium browser if invalid.
 * @property {boolean|null} valid     Whether the prerequisite is valid or not, or `null` if it couldn't be evaluated.
```

module/applications/compendium-browser.mjs
```diff
@@ -653,9 +670,27 @@ export default class CompendiumBrowser extends Application5e {
...
       entry: { img, name, subtitle, uuid, source },
+      displayPrerequisites: this.displayPrerequisites,
       displaySelection: this.displaySelection,
       selected: this.#selected.has(uuid)
     };
+    if ( this.options.prerequisites.validate ) {
+      const results = this.options.prerequisites.validate(entry);
+      if ( results.size ) {
+        context.prerequisites = results.values().reduce((r, result) => {
+          if ( result.valid === false) {
```
Fyorl:
```suggestion
          if ( result.valid === false ) {
```

module/applications/advancement/item-choice-flow.mjs
```diff
@@ -269,7 +269,14 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
     }
 
     const result = await CompendiumBrowser.select({
-      filters, selection: { min: 1, max: max - current }
+      filters,
+      prerequisites: {
+        validate: item => {
+          if ( !item.system.prerequisiteLabels ) return [];
```
Fyorl: The return type of this callback is annotated as `PrerequisiteValidationResults`, which is a `Map`. Maybe we should allow `void` return?

module/applications/compendium-browser.mjs
```diff
@@ -591,8 +606,10 @@ export default class CompendiumBrowser extends Application5e {
     this.#results = CompendiumBrowser.fetch(CONFIG[context.filters.documentClass].documentClass, {
       filters,
       types: context.filters.types,
+      index: !this.options.prerequisites.validate || !this.options.prerequisites.fullDocuments,
```
Fyorl, line 609: For advancements I know this is only a handful of items, but the `index: false` case is not great really because it doesn't collect entries to batch pull them from compendia. Do we actually use `index: false` anywhere aside from here?

What do we actually need to evaluate prereqs? Just `system.prerequisites` and `system.source`?

module/data/item/feat.mjs
```diff
@@ -325,27 +370,21 @@ export default class FeatData extends ItemDataModel.mixin(
...
   }={}) {
+    const results = this.prerequisiteLabels({ actor, added, level });
     const messages = [];
 
-    // If a feature has item pre-requisites, make sure the other items exist on the actor
-    if ( this.prerequisites.items.size
-      && !Array.from(this.prerequisites.items).some(i => actor.identifiedItems.get(i)?.size) ) {
+    if ( results.get("items")?.valid === false ) {
       messages.push(_loc("DND5E.Prerequisites.Warning.MissingItem", {
         items: game.i18n.getListFormatter({ type: "disjunction" }).format(Array.from(this.prerequisites.items))
```
Fyorl: I think we can substitute in the label from `prerequisiteLabels` here to give the actual item names rather than identifiers.

module/data/item/feat.mjs
```diff
@@ -325,27 +370,21 @@ export default class FeatData extends ItemDataModel.mixin(
   validatePrerequisites(actor, {
```
Fyorl: This is a bit unfortunate because the actual validation is happening in `prerequisiteLabels` now, while this method is mainly a glorified wrapper around a notification. They should probably be swapped. I can only find one module that is calling `validatePrerequisites` so I'd be inclined to just make a clean break here and call this something like `assertPrerequisites` and renamed `prerequisiteLabels` to `validatePrerequisites`.

But if we don't want to do that, and we are already considering making `prerequisiteLabels` static so that it can work off index data, then we could call it something like `prerequisiteStatus` or `prerequisiteReport` since it does actually give a bunch of data about the prerequisites, not just validates them.

module/registry.mjs
```diff
@@ -141,6 +147,81 @@ class EnchantmentRegisty {
   }
 }
 
+/* -------------------------------------------- */
+/*  Identifiers Registry                        */
+/* -------------------------------------------- */
+
+class IdentifierRegistry {
```
Fyorl: I don't know if it really makes sense to have two registries that both lookup items via identifier. Can't we unify them? We can have this `#identifiers = new Mapt([["*", new Map()]]);` central backing and then a call to `dnd5e.registry.classes.get` just forwards to `dnd5e.registry.items.get(key, { type: "class" })`. We make some slight adjustments in this PR then to adjust the places we currently call `dnd5e.registry.identifiers.get()` to add `?.name` to the end.

Fyorl (changes requested):

module/registry.mjs
arbron: My main concern was that we the additional data that would require. For the identify registry we just need to store a single name per identifier per type, while the other item registries we also have image path and a collection of UUIDs for every instance of that item. It's not a lot of data, but it seemed like that extra data for every single Item in every single compendium felt like a lot.

module/applications/compendium-browser.mjs
arbron: Whoops, I was very confused only to realize I had made a last-minute change locally that I never pushed up :O. In my version of the template it checks `indeterminate` first `{{ ifThen indeterminate 'question' (ifThen valid 'check' 'xmark') }}` so this isn't an issue.

module/registry.mjs
Fyorl: Yeah, valid. It's either that or we're stuck with 2 very slightly different registries as public API "forever" though. If we add more item registries (we have 4 now I think on 6.0) we converge on that state anyway, and then the identifiers registry becomes completely redundant. Is there a path forward where we have a single backing registry and we augment it with the extra data on top for only the subset that we use the item registry for currently?

module/registry.mjs
Fyorl: I guess this problem also exists in `ItemRegistry` and I never noticed before.

module/applications/compendium-browser.mjs
arbron: I've take the batching fetch method we had for encounter actors and turned it into a general utility so now the Compendium Browser results are retrieved in bulk. If there are still performance concerns after that, we could also wait until `_renderResults` to actually retrieve the documents and fetch only the ones we are rendering.

It is potentially possible to work fully off of indexes, but it does mean adding a bunch of additional fields to the index data, so its a trade-off between extra full document retrievals when in the browser versus extra indexed fields potentially across the whole session.

arbron:
Pushed up the `bulkFromUuid` and other small fixes, I'll take a gander at the registry and `prerequisitesLabels` issues tomorrow.

module/applications/compendium-browser.mjs
Fyorl: I think the `bulkFromUuid` is all we need. It turns 50 round-trips into a maximum of 10, usually just 1. So I don't have any concerns after that.

module/utils.mjs
```diff
@@ -49,6 +49,41 @@ export function roundCurrency(value, denomination) {
...
+ * @returns {Map<string, Document>}  Documents mapped to the provided UUID.
+ */
+export async function bulkFromUuid(uuids) {
+  const collections = new Map();
+  const redirected = new Map();
+
+  for ( const uuid of uuids ) {
+    const { collection, id, uuid: redirectedUuid } = foundry.utils.parseUuid(uuid);
+    if ( !collections.has(collection) ) collections.set(collection, []);
+    collections.get(collection).push(id);
```
krbz999, line 68: I think this works on all relevant browsers now:

```js
collections.getOrInsert(collection, []).push(id);
```

module/registry.mjs
arbron: Rewrote it to use a single static `ItemRegistry#items` store with specialized instances. Only items that have instances will retrieve the full data, otherwise they just store the name.

module/data/item/feat.mjs
```diff
@@ -363,6 +360,60 @@ export default class FeatData extends ItemDataModel.mixin(
...
+      foundry.utils.logCompatibilityWarning(
+        "`FeatData#validatePrerequisites` has been renamed `assertPrerequisites`.",
+        { since: "DnD5e 6.0", until: "DnD5e 7.0" }
+      );
+      return this.assertPrerequisites(context, _options);
+    }
+
+    const { actor, added=[], level=actor?.system?.details?.level } = context;
+    const prerequisites = new Map();
+    const legacy = this.source.rules === "2014" || dnd5e.settings.rulesVersion === "legacy" ? "Legacy" : "";
```
Fyorl:
```suggestion
    const legacy = (this.source.rules === "2014") || (dnd5e.settings.rulesVersion === "legacy") ? "Legacy" : "";
```

module/registry.mjs
```diff
@@ -213,72 +217,106 @@ class ItemRegistry {
 
   /* -------------------------------------------- */
 
+  /**
+   * All items formatted for a select input.
+   * @type {Iterator<FormSelectOption>}
+   */
+  get #options() {
+    return ItemRegistry.#items.get(this.#itemType).entries()
```
Fyorl: I think the backing `Map` is no longer populated per type automatically now so we need to guard.
```suggestion
    return (ItemRegistry.#items.get(this.#itemType)?.entries() ?? [].values())
```

module/registry.mjs
```diff
@@ -213,72 +217,106 @@ class ItemRegistry {
...
+    return this.#options.map(o => ({ ...o, group: this.label }))
+      .sort((lhs, rhs) => lhs.label.localeCompare(rhs.label, game.i18n.lang));
   }
 
   /* -------------------------------------------- */
   /*  Methods                                     */
   /* -------------------------------------------- */
 
+  /**
+   * Get the name of an item based on identifier. Accepts optional type as either separate option or using
```
Fyorl:
```suggestion
   * Get an item descriptor based on identifier. Accepts optional type as either separate option or using
```

module/utils.mjs
Fyorl: @arbron I see a `getOrInsert` used elsewhere but not here. Did you mean to add one here as well?

module/utils.mjs
```diff
@@ -49,6 +49,41 @@ export function roundCurrency(value, denomination) {
...
+  for ( const [collection, ids] of collections.entries() ) {
+    if ( collection instanceof foundry.documents.collections.CompendiumCollection ) {
+      fetches.push(collection.getDocuments({ _id__in: ids }));
+    } else {
+      fetches.push(ids.map(id => collection.get(id)));
+    }
+  }
+
+  return (await Promise.all(fetches)).flat().reduce((map, doc) => {
+    map.set(redirected.get(doc.uuid), doc);
```
Fyorl:
```suggestion
    if ( !doc ) return map;
    map.set(redirected.get(doc.uuid), doc);
```

Fyorl (approved):
With the `bulkFromUuid` and the `ItemRegistry` changes, this is looking really solid, nice work

PR #7146, opened 2026-06-16
Added level-gated item choice advancement

### LINKED ISSUE #7145: Add a new advancement type to enable choices of different items depending on character LvL
Since I don't realy like modules like SpellBook and such, and since simplu sing a Choose Item advancement makes available lvl 9 spells at lvl 1, I think a new advancement type would be really helpfull.
I already started working on something similar on my own, but having it integrated with the rest would be much better.

CrazyPascu:
Added LevelGatedItemChoice as a first-class dnd5e advancement type, ported from my standalone module into the system
source structure.

This adds:
- A new advancement document class extending ItemChoiceAdvancement
- A dedicated configuration data model with per-pool-entry minLevel data
- Config and flow applications using native dnd5e advancement V2 patterns
- Templates for level-section pool editing and level-grouped choice flow
- Parent/child pool linking support via poolRole, poolId, and parentPoolId
- Pool merging and pending child-pool discovery during advancement flows
- Level-gated availability checks when applying selected items
- Localization strings for the new advancement type and UI
- LESS styles for the config sheet and advancement flow
- Registration in CONFIG.DND5E.advancementTypes
- Barrel exports for documents, applications, and data models

Closes #7145 

Creation UI
LevelGateAdvancementCreation

Selection UI
LevelGateItemSelection

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -558,6 +558,53 @@
       "SpellList": "Only spells available on the {lists} spell list can be chosen for this advancement."
     }
   },
+  "LevelGatedItemChoice": {
+    "AvailableFromLevel": "Available from Level {level}+",
+    "DropHere": "Drop items here for level {level}+",
+    "Hint": "Present the player with item choices from a pool that expands as their advancement level increases.",
+    "LevelLabel": "Level {level}",
+    "Notification": {
+      "Moved": "Moved {name} to level {level}."
+    },
+    "PoolHeading": "Level-Gated Item Pool",
+    "PoolHint": "Drop items into the level where they first become available.",
+    "PoolRole": {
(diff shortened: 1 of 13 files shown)


Zanderaf:
Do wanna ask, how is this different from just an Item choice advancement that allows a choice of items with configured level and item prerequisites? Is this something only to be used for spells?

CrazyPascu:
I created it as such at first, as I wanted a DnD Beyond like spell selection during character creation.
But it also works for other features or scenarios, like eldritch invocations as a warlock.
For me its a bit simpler than having to edit each single item with level limits.
Essentialy its a merger between the grant items and choose items advancement.

PR #7159, opened 2026-06-18
Add system active effect sheet, separate choice config

arbron:
Creates a new active effect sheet that uses the system's styles. This sheet uses a separate config dialog for editing individual changes rather than doing that inline. This means that the main sheet can submit on change.

Active Effect Sheet V2

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2641,7 +2641,7 @@
   "Change": {
     "Configuration": "Change Configuration",
     "Label": "Change",
-    "Title": "{effect} Change #{number}"
+    "Title": "{effect} Change {number}"
   },
   "DropHint": "Drop an Active Effect here",
   "Empty": "No associated effects. Use the <i class=\"fas fa-plus\"></i> button above to create one, or select an existing effect from the drop-down.",
--- a/lang/en.json
+++ b/lang/en.json
@@ -289,7 +289,7 @@
 },
 
 "DND5E.ACTIVEEFFECT": {
-  "AttributeKeyTooltip": "For a list of common keys, see <a href=\"{url}\">the wiki</a>.",
+  "AttributeKeyTooltip": "For a list of common keys and their accepted values, see <a href=\"{url}\">the wiki</a>.",
   "FIELDS": {
     "statuses": {
       "label": "Status Conditions",
@@ -2580,9 +2580,12 @@
 "DND5E.Effect": "Effect",
 "DND5E.EFFECT": {
   "Action": {
+    "AddChange": "Add Change",
     "CreateEffect": "Create Effect",
     "DeleteEffect": "Delete Effect",
+    "DeleteChange": "Delete Change",
     "DissociateEffect": "Dissociate Effect",
+    "EditChange": "Edit Change",
     "EditEffect": "Edit Effect",
     "RemoveStatus": {
       "all": "All Sources",
@@ -2590,6 +2593,7 @@
     },
     "Search": "Search effects"
   },
+  "AdditionalChanges": "Additional Changes",
   "Application": {
     "Action": {
       "ApplyTokens": "Apply to selected tokens"
@@ -2602,6 +2606,26 @@
   },
   "BASE": {
     "FIELDS": {
+      "changes": {
+        "element": {
+          "key": {
+            "label": "Attribute Key"
+          },
+          "phase": {
+            "hint": "The point in data preparation or another event a change is to be applied.",
+            "label": "Phase"
+          },
+          "priority": {
+            "label": "Priority"
+          },
+          "type": {
+            "label": "Type"
+          },
+          "value": {
+            "label": "Value"
+          }
+        }
+      },
       "magical": {
         "hint": "This effect is from a magical source and should be disabled when magic isn't available.",
         "label": "Magical"
@@ -2613,6 +2637,11 @@
     "Passive": "Passive Effects",
     "Inactive": "Inactive Effects",
     "Unavailable": "Unavailable Effects"
+  },
+  "Change": {
+    "Configuration": "Change Configuration",
+    "Label": "Change",
+    "Title": "{effect} Change #{number}"
   },
   "DropHint": "Drop an Active Effect here",
   "Empty": "No associated effects. Use the <i class=\"fas fa-plus\"></i> button above to create one, or select an existing effect from the drop-down.",
@@ -4396,6 +4425,7 @@
 "DND5E.Shape": "Shape",
 
 "DND5E.SheetClass": {
+  "ActiveEffect": "D&D 5e Active Effect Sheet",
   "AdventureImporter": "D&D 5e Adventure Importer",
   "Character": "D&D 5e Character Sheet",
   "ClassSummary": "D&D 5e Class Summary Sheet",
--- a/lang/en.json
+++ b/lang/en.json
@@ -2612,7 +2612,7 @@
             "label": "Attribute Key"
           },
           "phase": {
-            "hint": "The point in data preparation or another event a change is to be applied.",
+            "hint": "The point in data preparation, or another event, at which a change is to be applied.",
             "label": "Phase"
           },
           "priority": {
@@ -2641,7 +2641,7 @@
   "Change": {
     "Configuration": "Change Configuration",
     "Label": "Change",
-    "Title": "{effect} Change {number}"
+    "Title": "{effect} Change #{number}"
   },
   "DropHint": "Drop an Active Effect here",
   "Empty": "No associated effects. Use the <i class=\"fas fa-plus\"></i> button above to create one, or select an existing effect from the drop-down.",
--- a/less/v2/sheets.less
+++ b/less/v2/sheets.less
@@ -49,13 +49,11 @@
   }
 }
 
-.dnd5e2.sheet:is(.item, .actor, .roll-table-sheet) {
+.dnd5e2.sheet:is(.item, .actor) {
   &:where(:not(.minimized)) > header {
     background: transparent;
     position: relative;
     z-index: 1;
-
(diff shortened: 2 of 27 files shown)


templates/effects/effect-changes.hbs
```diff
@@ -0,0 +1,59 @@
+<section class="tab scrollable changes {{~#if tab.active}} active{{/if}}" data-group="{{ tab.group }}" data-tab="{{ tab.id }}">
+
+    <section class="itmes-list changes-list">
```
thatlonelybugbear:
```diff
- <section class="itmes-list changes-list">
+ <section class="items-list changes-list">
```

less/v2/sheets.less
```diff
@@ -263,21 +274,32 @@ body.detached .dnd5e2.sheet .tab { max-height: unset; }
     line-height: 24px;
   }
 
-  form:is(.tab-inventory, .tab-features, .tab-spells, .tab-effects, .tab-advancement, .tab-activities, .tab-crew) .create-child,
-  &:is(.tab-inventory, .tab-features, .tab-spells, .tab-effects, .tab-advancement, .tab-activities, .tab-crew) .create-child {
+  form:is(.tab-inventory, .tab-features, .tab-spells, .tab-effects, .tab-advancement, .tab-activities, .tab-crew, .tab-changes) .create-child,
+  &:is(.tab-inventory, .tab-features, .tab-spells, .tab-effects, .tab-advancement, .tab-activities, .tab-crew, .tab-changes) .create-child, {
```
thatlonelybugbear:
```diff
- .create-child, {
+ .create-child {
```

thatlonelybugbear:
Shouldn't the Status Conditions be also moved from Details tab to Changes tab?

It would make sense to have all the "changes" one AE might be applying to the actor in one place.

krbz999:
Behaves strangely if you perform deletions of changes while the change editors are open, which ends up having latter changes' editors modifying a different change because the index has changed.

Deleting a change also doesn't close the change's editor, likely cus it's not able to tell that the correct index has been removed.

Luckily there's an easy fix for this: Extend the change data with an `_id` property to use instead of array indices.

(Also the header has css transitions every time a change is made.)

roth-michael:
> Shouldn't the Status Conditions be also moved from Details tab to Changes tab?
> 
> It would make sense to have all the "changes" one AE might be applying to the actor in one place.

On the one hand, this would differ from the expectation set by Core. On the other, we already now have "separate status effects" there, so it does feel like it would be consistent.

arbron:
Updated screenshot:

Active Effect Sheet V2 1

lang/en.json
```diff
@@ -2572,6 +2576,26 @@
   },
   "BASE": {
     "FIELDS": {
+      "changes": {
+        "element": {
+          "key": {
+            "label": "Attribute Key"
+          },
+          "phase": {
+            "hint": "The point in data preparation or another event a change is to be applied.",
```
```suggestion
            "hint": "The point in data preparation, or another event, at which a change is to be applied.",
```

module/applications/active-effect/effect-change-config.mjs
```diff
@@ -0,0 +1,147 @@
...
+  get effect() {
+    return this.options.document;
+  }
+
+  /* -------------------------------------------- */
+
+  /** @override */
+  get title() {
+    return _loc("DND5E.EFFECT.Change.Title", {
+      effect: this.effect.name, number: this.options.changeId
```
Fyorl: `changeId` is a random ID now, rather than a number, so this title looks a bit strange, but maybe it's intentional.

module/applications/active-effect/effect-change-config.mjs
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * Prepare rendering context for the config section.
+   * @param {ApplicationRenderContext} context  Context being prepared.
+   * @param {HandlebarsRenderOptions} options   Options which configure application rendering behavior.
+   * @returns {ApplicationRenderContext}
+   * @protected
+   */
+  async _prepareConfigContext(context, options) {
+    context.source = this.effect.system._source.changes.find(c => c._id === this.options.changeId);
```
Fyorl: Maybe we want a getter like `get change()`?

module/applications/active-effect/active-effect-sheet.mjs
```diff
@@ -0,0 +1,195 @@
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Open the config application for editing a change.
+   * @this {ActiveEffectSheet5e}
+   * @type {ApplicationClickAction}
+   */
+  static async #onEditChange(event, target) {
+    const { changeId } = event.target.closest("[data-change-id]")?.dataset ?? {};
```
Fyorl:
```suggestion
    const { changeId } = target.closest("[data-change-id]")?.dataset ?? {};
```

module/applications/active-effect/active-effect-sheet.mjs
```diff
@@ -0,0 +1,195 @@
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Delete a change from the effect's changes array.
+   * @this {ActiveEffectSheet5e}
+   * @type {ApplicationClickAction}
+   */
+  static async #onDeleteChange(event, target) {
+    const index = Number(event.target.closest("[data-index]")?.dataset.index || 0);
```
Fyorl:
```suggestion
    const index = Number(target.closest("[data-index]")?.dataset.index || 0);
```

module/documents/active-effect.mjs
```diff
@@ -1037,6 +993,24 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
 
   /* -------------------------------------------- */
 
+  /**
+   * Prepare the context for individual changes to display on actor, item, or active effect sheets.
+   * @param {object} change                     Change to prepare.
+   * @returns {object}  An object of chat data to render.
+   */
+  getSheetChangeContext(change) {
```
Fyorl: Sheet context preparation is typically `async` even if we don't need it right now, we might need it later. Also, we could do away with the `this.type` check if this were delegated to subtypes. It seems like we might want subtypes to handle this since a given change type might have different meanings across different AE subtypes. We can have the base AE model handle the common change context preparation.

templates/effects/change-config.hbs
```diff
@@ -0,0 +1,9 @@
+<fieldset>
+    <legend>{{ localize "DND5E.EFFECT.Change.Configuration" }}</legend>
+    <p class="note info">{{{ hintText }}}</p>
+    {{ formField fields.key name=(concat prefix "key") value=source.key rootId=partId }}
+    {{ formField fields.type name=(concat prefix "type") value=source.type options=typeOptions rootId=partId }}
+    {{ formField fields.value name=(concat prefix "value") value=source.value elementType="input" rootId=partId }}
+    {{ formField fields.priority name=(concat prefix "priority") value=source.priority
+                 placeholder=defaultPriority rootId=@root.partId }}
```
Fyorl: We are using `rootId=partId` directly above, not sure which is correct.

module/data/abstract/active-effect-data-model.mjs
```diff
@@ -1,8 +1,23 @@
...
 export default class ActiveEffectDataModel extends foundry.data.ActiveEffectTypeDataModel {
+  /* -------------------------------------------- */
+  /*  Model Configuration                         */
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  static defineSchema() {
+    const schema = super.defineSchema();
+    schema.changes.element.extendFields({
+      _id: new DocumentIdField({ initial: () => foundry.utils.randomID() })
```
Fyorl, line 18: At the moment I see we are just using this to help with some of the UI methods, but if we start relying on this ID, then we will have an issue because legacy changes won't have an ID, and they will get a new one every time they are run through `cleanData`. We would need a World migration to give these legacy changes stable IDs.

Fyorl (changes requested):
Nice work, I don't have much to add, see comments below and one additional note that I see the context menu is a TODO but it would be preferable to have it available sooner rather than later. I think having the option of the control or right-click > edit is important. Might actually be nice to have it on the roll table result rows, which is a design that we are closely resembling here already.

module/applications/active-effect/effect-change-config.mjs
arbron: I was basing this off the roll table sheet, but obviously with our design it is much more prominent. I'll swap to the change number instead.

arbron:
> Nice work, I don't have much to add, see comments below and one additional note that I see the context menu is a TODO but it would be preferable to have it available sooner rather than later. I think having the option of the control or right-click > edit is important. Might actually be nice to have it on the roll table result rows, which is a design that we are closely resembling here already.

I'll implement these context menus in a followup PR.

module/data/abstract/active-effect-data-model.mjs
```diff
@@ -57,4 +74,28 @@ export default class ActiveEffectDataModel extends foundry.data.ActiveEffectType
...
+  /* -------------------------------------------- */
+
+  /**
+   * Prepare the context for individual changes to display on actor, item, or active effect sheets.
+   * @param {object} change  Change to prepare.
+   * @returns {object}       An object of chat data to render.
+   */
+  async getSheetChangeContext(change) {
+    return {
+      name: getHumanReadableAttributeLabel(change.key, { actor: this.actor, prefixItemName: false })
```
Fyorl, line 89: I think I didn't catch this before (or maybe I introduced it?) but if there is no human readable label available we are displaying as blank, when we should probably fall back to the change key. I think we can also check if the change key === the human readable label and, if so, suppress the tooltip.

Fyorl (approved):

Issue #6712, opened 2026-02-11
SRD 5.1: Monster Update List


Zanderaf:
The SRD 5.1 Actors feels like their in a poor state atm. Going in order on some issues I personally feel like should be tackled:
- This is in no way a "hey this **_needs_** to be fixed, just some personal insights of what i've seen.

# General Issues
- NPC features should ideally use `[[lookup @name lowercase]]{creature}` in their feature names when mentioning the monster.
  - `The aboleth can breathe air and water.` -> `The [[lookup @name lowercase]]{creature} can breathe air and water.`
- NPC weapons are all set to `proficient` instead of `automatic`. From what I know, NPC actors are considered proficient in all of their equipment.
- Redundant Legendary Actions + Legendary Resistance features.
- NPCS having `natural armor` as their AC calculation when they ideally should just be `unarmored` (see Acolyte & Archmage as examples).
- NPC features do not have the `Monster Feature` Feature type. All of the abilities I've seen are not actually labeled as a type of feature.
- Monsters typically do not grant applicable AEs to grant to players. This feels like it would be important to eventually have for stuff like poison effects running out automatically with v14's expiry stuff.
- No &reference[rule] enrichers in monster features. I feel as if those would be important to include. Conditions should probably be `&referenc[charmed apply=false]` in cases like `One creature &reference[charmed apply=false] by the [[lookup @name lowercase]]{creature} takes [[/damage average]] damage`.
- I feel as if the "Activation Time" activity field should be utilized a bit more. See below:
<img width="302" height="499" alt="Image" src="https://github.com/user-attachments/assets/a539a664-ed22-48dd-b38c-db493fe5c70d" />

- Some features seem to have weird bolded text and some aren't even enriched properly.

<img width="495" height="880" alt="Image" src="https://github.com/user-attachments/assets/a15bbac9-17af-480d-acc8-00ad47061157" />

<img width="643" height="588" alt="Image" src="https://github.com/user-attachments/assets/d6f7f186-b2d6-4464-854d-95bacf402974" />

# Specific Issues
Going by creature type in order:

## Aberrations
**Aboleth:**
- Mucous Cloud does not have an applicable AE
- Tentacles does not have an applicable AE.
- Enslave does not have an applicable AE.
- Detect has two activities, one `check` and one `use`. The `use` activity feels redundant here.

**Chuul:**
- Sense Magic could use a link to Detect Magic (2014).
- Tentacles does not have an applicable AE.

**Cloaker:**
[...]

PR #7160, opened 2026-06-19
5.1 SRD Update: Monsters - Fey

Zanderaf:
Updated the 5.1 SRD Fey creatures to current system standards. Included AEs, enrichers, and lookups where I could.

Does not currently touch spells or spellcasting traits, will handle that in future PR when spells are updated.

Related #6712

### DIFF
--- a/packs/_source/monsters/fey/dryad.yml
+++ b/packs/_source/monsters/fey/dryad.yml
@@ -1661,12 +1661,8 @@ items:
         type: base
         system:
           changes:
-            - key: system.attributes.ac.formula
+            - key: system.attributes.ac.min
               value: 16
-              priority: null
-              type: override
-            - key: system.attributes.ac.calc
-              value: custom
               priority: null
               type: override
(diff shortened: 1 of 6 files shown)


packs/_source/monsters/fey/dryad.yml
```diff
@@ -1624,8 +1659,20 @@ items:
...
+        system:
+          changes:
+            - key: system.attributes.ac.formula
+              value: 16
+              priority: null
+              type: override
+            - key: system.attributes.ac.calc
+              value: custom
+              priority: null
+              type: override
```
arbron: Should be `ac.min` rather than a custom formula. This is correct in the Barkskin spell, so we probably need to re-sync spells after they have been updated.

```suggestion
            - key: system.attributes.ac.min
              value: 16
              priority: null
              type: override
```

arbron (changes requested):

arbron (approved):

PR #7168, opened 2026-06-19
Separate Concentration Checks per Effect

observer1236:
Adds a world setting that enables separate concentration checks for each effect a character is maintaining.
QQ_1781887301388

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -6131,6 +6131,10 @@
       "Hint": "Use the same initiative roll for identical creatures."
     }
   },
+  "ConcentrationPerEffect": {
+    "Name": "Separate Concentration Checks per Effect",
+    "Hint": "When enabled, a character maintaining multiple concentration effects will roll a separate concentration check for each effect when damaged."
+  },
   "CRITICAL": {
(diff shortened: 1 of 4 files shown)


Sayshal:
I don't believe this is RAW for anything first party and any third party content I am familiar with. Is there a use case for this behavior that a module couldn't offer as a homebrew alternative to RAW?

observer1236:
> I don't believe this is RAW for anything first party and any third party content I am familiar with. Is there a use case for this behavior that a module couldn't offer as a homebrew alternative to RAW?

Since the official D&D 5e rules do not provide any resource that would allow a creature to maintain multiple concentrations simultaneously (even though the system permits this feature), the only application scenario I can recall is the Dual-Focused feat from the Tal'dorei Campaign Setting (first edition, not the reborn version). The functionality offered by this pull request may indeed be unnecessary, and I will reconsider and decide whether to close it.

JPMeehan:
There's at least one monster, Niv-Mizzet from GGR, that has multi-concentration and he just does it as one concentration.

PR #7172, opened 2026-06-20
Add item selection to Forward Activity

### LINKED ISSUE #6841: Consider allowing Forward activity to forward to a different item
Use case:
- several monk features that want to roll Unarmed Strike.
- several Beastheart features (MCDM premium module) that want to use a monster's 'primary attack'.

It would allow for not having to implement these attack/damage activities over and over for each item.

observer1236:
Enable Forward Activity to select and trigger any non-forward activity from another item belonging to the same actor. 
This should close #6841 
QQ_1781968457489


module/data/activity/forward-data.mjs
```diff
@@ -21,8 +21,17 @@ export default class BaseForwardActivityData extends BaseActivityData {
     delete schema.target;
     return {
       ...schema,
-      activity: new SchemaField({
-        id: new DocumentIdField()
+      targetItem: new DocumentIdField({
```
pgervais70: This should be similar to a consumption target, to allow setting an type and identifier when the item is not embedded, otherwise it will not be very useable.

Zanderaf: +1 on that, most of the time that Forward is used when setting up a feature its within a compendium or world and not currently on the actor.

observer1236: Fair enough, I'll find time to improve it.

Issue #4351, opened 2024-09-19
Feature Request: Template Targeting and Template Linking


Kapuzenjoe:
Hello,
First of all, if these features have already been requested, I apologize.
If you use AoE spells such as Dragons Breath, you get a measured template with which you can define the targets. However, the actors are not directly targeted, but you have to manually add them to the target one by one afterwards. I find this process very time-consuming, especially in fights with many AoE effects. The combat system in DnD is already very slow.
Is there a possibility that if I activate an AoE spell with a measured template, the actors in the template are directly in target?

It would also be nice if measured templates were linked to an effect (or activity?) and removed from the map after the effect ends (like concentration effects). As a DM, I always have to "clean up" after my players at the moment.

I know that these features are currently possible via mods, but these mods often offer significantly more features and usually have additional dependencies where a system implementation would be much more convenient.

best regards

thatlonelybugbear:
This is something I am working on for a module of mine.

I can instead take a stab at creating a PR for this, if you want.

General direction:
- [ ] Highlight targets before region placement and auto-target on placement
  - [ ] Per-item auto-target setting (+ activity override)
  - [ ] System-level default for auto-target behaviour
  - [ ] If an item creates a persistent region (probably based on the duration or checkbox?), update targets as tokens enter and leave the region
- [ ] 3D templates:
  - [ ] Elevation support
  - [ ] Multi-level support

thatlonelybugbear:
<img width="504" height="931" alt="Image" src="https://github.com/user-attachments/assets/6e733aaf-c2f8-4c6d-a04f-764121c60a1c" />

<img width="644" height="549" alt="Image" src="https://github.com/user-attachments/assets/9d28fb50-de2a-4bbc-83d1-1a25eee449e0" />

<img width="590" height="433" alt="Image" src="https://github.com/user-attachments/assets/fc7f2eda-2007-4622-a20b-3c91b4f8eb55" />

Kapuzenjoe:
I've been experimenting with something similar in my [save-tray-5e](https://github.com/Kapuzenjoe/save-tray-5e) module and ran into a few things worth considering:

- Using the *wall restriction* of Regions for targeting feels like a natural fit.
- `activity.target.affects?.type` against the token's disposition could be a clean way to handle target filtering (e.g. enemies only, allies only). At least in many cases where it is clear.
- Certain creature states like dead or ethereal tokens should probably be excluded from targeting early on?

[...]

PR #7181, opened 2026-06-23
[#4351] Region Template auto targeting

thatlonelybugbear:
What this PR does. Probably too many things...

Relevant to #4351 for the Template Targeting

If I know how far the system is willing to go, I can adjust accordingly

- [ ] Highlight targets before region placement and auto-target on placement
  - [x] Per-item auto-target setting
  - [x] Per-item walls behaviour setting
  - [x] System-level default for auto-target behaviour
  - [x] System-level setting for overriding targeting method, from matching targets to highlighted grid spaces only to using occupied-space overlap, working around the Foundry's grid-space center-point rule.
    - [ ] Match that logic to different region grid highlighting.
- [ ] If an item creates a persistent region (probably based on the duration or checkbox?), update targets as tokens enter and leave the region.
- [x] 3D templates:
  - [x] Elevation support
    - [x] The region matches initial elevation to the source token elevation.
    - [x] During template preview, the default keys handling elevation (Q and E) can be used to increase/decrease template elevation (the base elevation is shown alongside the template length).
      - Spheres and cones extend upwards and downwards to their radius.
        - For example, Caster at 20ft. Casts fireball, which has bottom at 0ft and top at 40ft (non inclusive). 
      - Cubes upwards
  - [x] Multi-level support
- [ ] Walls interaction
  - [x] Walls block: Templates targeting matching core's region/wall interactions
  - [x] Spread around corners: For items like Fireball, if there is a way around walls or though gaps, the full template targets tokens
    - [ ] Match that logic to actual region grid highlighting
  - [x] Ignore walls
- [x] Filter targets by
  - [x] Self. Ally, Enemy
  - [x] Object and Space do not target tokens.
  - [x] Hidden targets are filtered out
  - [x] Ethereal targets are only targeted if the source has also Ethereal status
  - [x] No "Defeated" creatures handling, as that might have undesirable side-effects
  - [ ] Filter per creature type
 - [ ] Respect options for walls behaviour and grid aware_ness_ into created regions


observer1236:
Is there any feasibility in allowing evaluation of the `system.target.affects.special` field before auto selecting targets, or adding a new field to evaluate its contents for target filtering? In official rules, features like Turn Undead only target specific types of creatures.

thatlonelybugbear:
> Is there any feasibility in allowing evaluation of the `system.target.affects.special` field before auto selecting targets, or adding a new field to evaluate its contents for target filtering? In official rules, features like Turn Undead only target specific types of creatures.

I have this working in my local version but not added in this PR yet (all of this was included in a module I was working on).

PR #7183, opened 2026-06-23
[#6671] Add conditions to active effects and changes

### LINKED ISSUE #6671: Add conditional for active effects and changes
Add conditions to active effects that are evaluated against the actor to allow effects to only be applied if the specified conditions are met. These conditions should be available for the effect as a whole as well as for each individual change.

Blocked by #6672

arbron:
Adds the ability to define conditions to both active effects and individual changes. These changes are specified using the filters system which is evaluated before the change is appplied. These conditions don't cause the whole effect to become suppressed, but instead each individual change is skipped which will be marked on the sheet once the AE breakdown is integrated.

Active Effect Conditions

This re-implements the `Actor#applyActiveEffects` method from core completely to add two new pieces of functionality. First, applying status effects to the actor is done in its own pass just before the `initial` phase. This makes conditions based on status effects work reliably and not order dependant. This also calls a new method `ActiveEffect5e#shouldApplyChange` which is where the conditions are evaluated. This is done here rather than in `applyChange` so that an `applied` boolean can be added to changes to be used on the sheet (before the change is cloned).

Closes #6671

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -226,6 +226,8 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     // Apply shims to moved fields
     change = change.effect._applyChangeShim(change);
 
+    if ( !change.effect.shouldApplyChange(change, options.replacementData) ) return {};
+
     // Handle special actor flags
     if ( change.key.startsWith("flags.dnd5e.") ) change = change.effect._prepareFlagChange(model, change);
 
@@ -388,21 +390,6 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     return super._applyChangeUpgrade(actor, change, current, delta, changes);
   }
 
-  /* -------------------------------------------- */
-
-  /**
-   * Evaluate the conditions to see if the effect or change should be applied.
-   * @param {Filter} [conditions]  Filter definition to evaluate.
-   * @returns {boolean}
-   */
-  evaluateConditions(conditions) {
-    if ( !conditions ) return true;
-    const rollData = (this.system.applicableType === "Actor") && this.actor
-      ? this.actor.getRollData({ deterministic: true })
-      : this.item.getRollData({ deterministic: true });
-    return conditions.check(rollData);
-  }
-
   /* --------------------------------------------- */
 
   /**
@@ -439,22 +426,18 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
   /**
    * Determine whether a specific change should be applied during this phase, setting `applied` if approved.
    * @param {object} change           Change that might be applied.
-   * @param {object} [options={}]
-   * @param {string} [options.phase]  Active effect phase being checked.
+   * @param {object} [conditionData]  Data used to evaluate conditions.
    * @returns {boolean}
    */
-  shouldApplyChange(change, { phase }={}) {
-    // TODO: Adapt when core implements https://github.com/foundryvtt/foundryvtt/issues/13931, if it does
-    if ( change.phase !== phase ) return false;
-
-    const targetDoc = this.applicableType === "Actor" ? this.actor : this.applicableType === "Item" ? this.item : null;
-    const conditionData = targetDoc?.getRollData?.();
+  shouldApplyChange(change, conditionData) {
     if ( conditionData ) {
-      if ( change.effect.system.conditions?.check(conditionData) === false ) return false;
+      if ( this.system.conditions?.check(conditionData) === false ) return false;
       if ( change.conditions?.check(conditionData) === false ) return false;
     }
 
-    change.applied = true;
+    const originalChange = this.system.changes.find(c => c._id === change._id);
+    if ( originalChange ) originalChange.applied = true;
+
     return true;
   }
 
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -226,7 +226,8 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     // Apply shims to moved fields
     change = change.effect._applyChangeShim(change);
 
-    if ( !change.effect.shouldApplyChange(change, options.replacementData) ) return {};
+    if ( (model instanceof foundry.abstract.Document)
+      && !change.effect._checkCondition(change, options.replacementData) ) return {};
 
     // Handle special actor flags
     if ( change.key.startsWith("flags.dnd5e.") ) change = change.effect._prepareFlagChange(model, change);
@@ -428,8 +429,9 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
    * @param {object} change           Change that might be applied.
    * @param {object} [conditionData]  Data used to evaluate conditions.
    * @returns {boolean}
+   * @internal
    */
-  shouldApplyChange(change, conditionData) {
+  _checkCondition(change, conditionData) {
     if ( conditionData ) {
       if ( this.system.conditions?.check(conditionData) === false ) return false;
       if ( change.conditions?.check(conditionData) === false ) return false;
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -129,7 +129,7 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
 
   /** @inheritDoc */
   get isSuppressed() {
-    if ( super.isSuppressed ) return true;
+    if ( super.isSuppressed || this.system.isSuppressed ) return true;
     if ( this.system.magical && this.actor?.statuses.has("antimagic") ) return true;
     if ( this.type === "enchantment" ) return false;
     if ( this.type === "condition" ) return false;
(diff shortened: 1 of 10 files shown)


module/documents/item.mjs
```diff
@@ -473,12 +473,12 @@ export default class Item5e extends SystemDocumentMixin(Item) {
...
     for ( const effect of this.allApplicableEffects() ) {
       if ( !effect.active ) continue;
-      changes.push(...effect.changes.map(change => {
-        const c = foundry.utils.deepClone(change);
-        c.effect = effect;
-        c.priority ??= c.mode * 10;
-        return c;
-      }));
+      for ( const change of effect.system.changes ) {
+        if ( (change.key === "") || !effect.shouldApplyChange(change) ) continue;
```
Fyorl:
```suggestion
        if ( (change.key === "") || !effect.shouldApplyChange(change, { phase: change.phase }) ) continue;
```

module/documents/item.mjs
```diff
@@ -487,9 +487,9 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     const overrides = {};
     const replacementData = this.getRollData();
     for ( const change of changes ) {
-      if ( !change.key ) continue;
-      const changes = change.effect.constructor.applyChange(this, change, { replacementData });
-      Object.assign(overrides, changes);
+      const result = ActiveEffect.CHANGE_TYPES[change.type].handler?.(this, change)
```
Fyorl:
```suggestion
      const result = ActiveEffect.CHANGE_TYPES[change.type]?.handler?.(this, change)
```

module/documents/active-effect.mjs
```diff
@@ -388,6 +388,21 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
   }
 
+  /* -------------------------------------------- */
+
+  /**
+   * Evaluate the conditions to see if the effect or change should be applied.
+   * @param {Filter} [conditions]  Filter definition to evaluate.
+   * @returns {boolean}
+   */
+  evaluateConditions(conditions) {
```
Fyorl: Not used

module/documents/active-effect.mjs
```diff
@@ -419,6 +434,30 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+   * @returns {boolean}
+   */
+  shouldApplyChange(change, { phase }={}) {
+    // TODO: Adapt when core implements https://github.com/foundryvtt/foundryvtt/issues/13931, if it does
+    if ( change.phase !== phase ) return false;
+
+    const targetDoc = this.applicableType === "Actor" ? this.actor : this.applicableType === "Item" ? this.item : null;
+    const conditionData = targetDoc?.getRollData?.();
+    if ( conditionData ) {
+      if ( change.effect.system.conditions?.check(conditionData) === false ) return false;
```
Fyorl: `change.effect === this`
```suggestion
      if ( this.system.conditions?.check(conditionData) === false ) return false;
```

module/documents/active-effect.mjs
```diff
@@ -419,6 +434,30 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+      if ( change.conditions?.check(conditionData) === false ) return false;
```
Fyorl, line 437: We never call `recheck` to bust cache, so we evaluate against the same filter each phase, which is probably not correct. I can't see a good place to call `recheck`, so I might suggest just removing the cache for now.

module/documents/active-effect.mjs
```diff
@@ -419,6 +434,30 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+   * @param {object} [options={}]
+   * @param {string} [options.phase]  Active effect phase being checked.
+   * @returns {boolean}
+   */
+  shouldApplyChange(change, { phase }={}) {
+    // TODO: Adapt when core implements https://github.com/foundryvtt/foundryvtt/issues/13931, if it does
+    if ( change.phase !== phase ) return false;
+
+    const targetDoc = this.applicableType === "Actor" ? this.actor : this.applicableType === "Item" ? this.item : null;
+    const conditionData = targetDoc?.getRollData?.();
```
Fyorl: Seems wasteful to call this for every change. Since `shouldApplyChange` is called before any AEs actually apply, it's not possible for `rollData` to be different between two changes. We control `Actor5e#applyActiveEffects` so a simple solution is to hoist the `getRollData` call we already make there (for `replacementData`) up above the first loop over effects. Then we pass it into this method and we also pass it as the `replacementData`.

Similar changes for `Item5e#applyActiveEffects`.

module/documents/actor/actor.mjs
```diff
@@ -365,7 +365,65 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     if ( (this.system?.prepareEmbeddedData instanceof Function) && (phase === "initial") ) {
       this.system.prepareEmbeddedData();
     }
-    return super.applyActiveEffects(phase);
+
+    const ActiveEffect = foundry.documents.ActiveEffect.implementation;
```
Fyorl: I am going to make a proposal for foundryvtt/foundryvtt#13931 but there is no guarantee it gets merged. Since changes have `_id`s now though, I don't think we need to clone the entire method here. We can look up the live change from `change.effect`. So then the implementation becomes something like this:

```js
applyActiveEffects(phase) {
  if ( (typeof this.system?.prepareEmbeddedData === "function") && (phase === "initial") ) {
    this.system.prepareEmbeddedData();
  }

  if ( phase === "initial" ) {
    for ( const effect of this.allApplicableEffects() ) {
      if ( effect.active ) for ( const s of effect.statuses ) this.statuses.add(s);
    }
  }

  return super.applyActiveEffects(phase);
}
```

```js
static applyChange(model, change, options={}) {
  change = change.effect._applyChangeShim(change);
  if ( !change.effect.shouldApplyChange(change, model, options.replacementData) ) return {};
  // existing override
}

shouldApplyChange(change, model, rollData) {
  const target = this.applicableType === "Actor" ? this.actor : this.applicableType === "Item" ? this.item : null;
  const conditionData = ((model === target) ? rollData : null) ?? target?.getRollData();
  if ( conditionData ) {
    if ( this.system.conditions?.check(conditionData) === false ) return false;
    if ( change.conditions?.check(conditionData) === false ) return false;
  }
  const live = this.system.changes.find(c => c._id === change._id);
  if ( live ) live.applied = true;
  return true;
}
```

Fyorl (changes requested):

module/documents/active-effect.mjs
```diff
@@ -226,6 +226,8 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     // Apply shims to moved fields
     change = change.effect._applyChangeShim(change);
 
+    if ( !change.effect.shouldApplyChange(change, options.replacementData) ) return {};
```
Fyorl: 13931 went in so we will probably just update to that. But in case we forget or there's some other reason, then we should not use `shouldApplyChange`. Use a totally separate method like `_checkCondition` for now, and update to `shouldApplyChange` once 14.365 is out.
```suggestion
    if ( !change.effect._checkCondition(change, options.replacementData) ) return {};
```

module/documents/item.mjs
```diff
@@ -487,9 +487,9 @@ export default class Item5e extends SystemDocumentMixin(Item) {
     const overrides = {};
     const replacementData = this.getRollData();
     for ( const change of changes ) {
-      if ( !change.key ) continue;
-      const changes = change.effect.constructor.applyChange(this, change, { replacementData });
-      Object.assign(overrides, changes);
+      const result = ActiveEffect.CHANGE_TYPES[change.type]?.handler?.(this, change)
+        ?? change.effect.constructor.applyChange(this, change, { replacementData });
```
Fyorl: On review, we don't need to call the handler at all. `applyChange` does that already.


```suggestion
      const result = change.effect.constructor.applyChange(this, change, { replacementData });
```

module/documents/active-effect.mjs
Fyorl: For activities we actually call `shouldApplyChange` twice. Once on the parent `Item` and then again once we found the right Activity to route to, but if `shouldApplyChange` succeeded on the `Item`, then it will succeed on the activity too since we don't change anything about the `change` or roll data.


```suggestion
    if ( (model instanceof foundry.abstract.Document) && !change.effect.shouldApplyChange(change, options.replacementData) ) return {};
```

Fyorl (changes requested):

module/documents/active-effect.mjs
arbron: Just to check, is `replacementData` getting passed into core's `shouldApplyChange`?

Fyorl: Not at the moment, it's another change I have proposed but it is currently tangled with some other PRs. So I would suggest we move to another method for now. We can still use `shouldApplyChange` without `replacementData` passed but we will need to do something like this if we want to avoid re-creating roll data per change:

```js
applyActiveEffects() {
  this._conditionData = this.getRollData();
  super.applyActiveEffects();
  delete this._conditionData;
}
```

arbron: Well I've renamed it for now, we can decide later how we want to swap over to the core method.

Fyorl (approved):

PR #7184, opened 2026-06-24
Unify drop area styles. Add document-tags drop-area styles.

Fyorl:
Thought for our `.hidden-tags` `<document-tags>` we could style them as drop areas since that's what we mainly use them for. We lose the ability to type or copy-paste a UUID manually but I don't know if that was really being used. Thought I would run it by you first anyway.

image


arbron (approved):
This looks good 👍🏽

Issue #6315, opened 2025-10-17
Add "Rule" active effect type that uses conditional application rather than specific target path


arbron:
Many features in the DnD5e rules are not currently supported in the system because they only apply in very specific circumstances. Most of these involve modifying rolls in some way. Currently we have a limited set of fields in `system.bonuses` that allow conditionally adding bonuses based on attack type, but they are limited to a small subset of possible rolling conditions. Adding additional fields to support all of the different potential conditions that could occurring while performing a roll is not feasible, so instead the system should have some sort of conditional modification that can occur. The popularity of the dearly departed Build-a-Bonus is evidence of the need for this (see https://github.com/foundryvtt/dnd5e/issues/4692).

### Proposal

To solve this issue, the system should add a new active effect type that takes advantage of the conditional DSL (hopefully) coming in V14 (see https://github.com/foundryvtt/foundryvtt/issues/13465). In this new AE type, the `key` and `mode` values in the changes array would be modified to support its new purpose.

#### Change Mode
The options in the change mode field would be replaced with a custom set of options specified by the system. The mode selected would allow the system to determine what modes to use for what purpose when performing a roll (or anywhere else we find a use for these rules). Some potential modes include:
- **Roll Bonus**: Add a bonus to the roll, same as `system.bonuses`
- **Roll Modifier**: Add or remove a modifier from the primary die in a roll
- **Roll Mode**: Give a D200 roll advantage or disadvantage, or make a damage roll a Critical

#### Change Key
The `key` field would take the conditional, indicating the situations in which the modification would occur. This would allow general filtering using a simple key ("all attack rolls") or something very specific using multiple keys and comparisons ("melee damage rolls made while wielding a two-handed weapon with a single hand and holding a shield"). The syntax of this will hopefully be able to take advantage of the conditional logic in V14, but if that doesn't make it in then perhaps would be based on the existing filtering system used by the system.

#### Change Value
The `value` field would be largely used for the same purpose as current effects, but its meaning would change based on the selected change mode. Due to the more limited domain compared to generic active effects, this maybe a good opportunity to add customized UI that changes based on the selected change mode to make it easier to configure these rules.

Zanderaf:
I just wanna say: Holy shit im excited

krbz999:
> "melee damage rolls made while wielding a two-handed weapon with a single hand and holding a shield"

Needs a UI.

Zanderaf:
[...]

PR #7187, opened 2026-06-24
[#6315] Add base system for applied rules, attack bonus rule

arbron:
Adds a system for collecting rules on actors and items applied by custom active effect change types. When these changes are applied they add to the `Actor5e#appliedRules` or `Item5e#appliedRules` objects grouped by type and key. This allows for quick retrieval of specific rules using a description like `attack:bonus`.

The `AppliedRules` map adds a helper method `collect` that takes a rule description as well as an actor and optional item and returns an iterator for all applicable rules. The iterator has some custom methods to aid handling the rules:

```javascript
// Collects all attack bonuses, checks their conditions, and creates a final formula
AppliedRules
  .collect("attack:bonus", this.actor, this.item)
  .filterWith(rollData)
  .toFormula();
```

This implements a single new change type of `bonus` and a single accepted key of `attack`, allowing attack bonuses to be defined. Additional change types and keys will be added once the basic design is approved.

### DIFF
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -258,7 +258,7 @@ export default class BaseAttackActivityData extends BaseActivityData {
    * @returns {{ data: object, parts: string[] }}
    */
   getAttackData({ ammunition, attackMode, situational }={}) {
-    const rollData = Object.assign(this.getRollData(), { roll: { attackMode } });
+    const rollData = this.getRollData({ data: { roll: { attack: { mode: attackMode } } } });
     if ( this.attack.flat ) return CONFIG.Dice.BasicRoll.constructParts({ toHit: this.attack.bonus }, rollData);
 
     const weapon = this.item.system;
@@ -365,6 +365,20 @@ export default class BaseAttackActivityData extends BaseActivityData {
     }
 
     return game.i18n.getListFormatter({ type: "disjunction" }).format(parts.filter(_ => _));
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  getRollData(options={}) {
+    const rollData = super.getRollData(options);
+    if ( rollData.roll ) {
+      rollData.roll.attack ??= {};
+      rollData.roll.attack.classification = this.attack.type.classification;
+      rollData.roll.attack.type = rollData.roll.attack.mode?.includes("thrown") ? "ranged" : this.attack.type.value;
+      rollData.roll.type = "attack";
+    }
+    return rollData;
   }
 
   /* -------------------------------------------- */
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -258,7 +258,7 @@ export default class BaseAttackActivityData extends BaseActivityData {
    * @returns {{ data: object, parts: string[] }}
    */
   getAttackData({ ammunition, attackMode, situational }={}) {
-    const rollData = this.getRollData({ roll: { attackMode } });
+    const rollData = Object.assign(this.getRollData(), { roll: { attackMode } });
     if ( this.attack.flat ) return CONFIG.Dice.BasicRoll.constructParts({ toHit: this.attack.bonus }, rollData);
 
     const weapon = this.item.system;
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -1,4 +1,5 @@
 import simplifyRollFormula from "../../dice/simplify-roll-formula.mjs";
+import AppliedRules from "../../documents/applied-rules.mjs";
 import { convertLength, formatLength, formatNumber, simplifyBonus } from "../../utils.mjs";
 import FormulaField from "../fields/formula-field.mjs";
 import DamageField from "../shared/damage-field.mjs";
@@ -257,7 +258,7 @@ export default class BaseAttackActivityData extends BaseActivityData {
    * @returns {{ data: object, parts: string[] }}
    */
   getAttackData({ ammunition, attackMode, situational }={}) {
-    const rollData = this.getRollData();
+    const rollData = this.getRollData({ roll: { attackMode } });
     if ( this.attack.flat ) return CONFIG.Dice.BasicRoll.constructParts({ toHit: this.attack.bonus }, rollData);
 
     const weapon = this.item.system;
@@ -269,6 +270,7 @@ export default class BaseAttackActivityData extends BaseActivityData {
       weaponMagic: weapon.magicAvailable ? weapon.magicalBonus : null,
       ammoMagic: ammo?.magicAvailable ? ammo.magicalBonus : null,
       actorBonus: this.actor?.system.bonuses?.[this.getActionType(attackMode)]?.attack,
+      ruleBonus: AppliedRules.collect("attack:bonus", this.actor, this.item).filterWith(rollData).toFormula(),
       situational
     }, rollData);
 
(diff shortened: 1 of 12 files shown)


module/documents/activity/mixin.mjs
```diff
@@ -1217,14 +1217,15 @@ export default function ActivityMixin(Base) {
 
     /**
      * Prepare a data object which defines the data schema used by dice roll commands against this Activity.
-     * @param {RollDataOptions} [options]
+     * @param {ActivityRollDataOptions} [options]
      * @returns {ActivityRollData}
      */
-    getRollData(options) {
+    getRollData({ roll, ...options }={}) {
```
Fyorl: We could consider supporting a generic `data` option that merges directly into `rollData`. So `getAttackData` would instead call `this.getRollData({ data: { roll: { attackMode } } });`

module/documents/applied-rules.mjs
```diff
@@ -0,0 +1,153 @@
...
+    return this.get(key)?.map(c => c.value) ?? [];
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  set(key, value) {
+    if ( !key ) return this;
+    if ( key.type && key.key ) {
+      const type = key.type.split(".")[1];
```
Fyorl: Will module-provided change types potentially collide? Should we only omit the first part if it's `dnd5e`?

module/documents/applied-rules.mjs
```diff
@@ -0,0 +1,153 @@
...
+  getValues(key) {
+    return this.get(key)?.map(c => c.value) ?? [];
+  }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  set(key, value) {
+    if ( !key ) return this;
+    if ( key.type && key.key ) {
```
Fyorl: Maybe a separate `add` method for this branch? It's not *wrong*, but it's a little bit of an abuse of the `Map` API and reads weirdly that we `push(key)`.

module/documents/applied-rules.mjs
```diff
@@ -0,0 +1,153 @@
...
+  /* -------------------------------------------- */
+
+  /**
+   * Filter rules based on effect & change conditions.
+   * @param {RollData} rollData
+   * @returns {RulesIterator}
+   */
+  filterWith(rollData) {
+    return new RulesIterator(this.filter(r => {
+      if ( r.effect?.system.conditions?.recheck(rollData) === false ) return false;
```
Fyorl:
```suggestion
      if ( r.effect?.system.conditions?.check(rollData) === false ) return false;
```

module/documents/applied-rules.mjs
```diff
@@ -0,0 +1,153 @@
...
+      if ( r.conditions?.recheck(rollData) === false ) return false;
```
Fyorl:
```suggestion
      if ( r.conditions?.check(rollData) === false ) return false;
```

module/documents/applied-rules.mjs
```diff
@@ -0,0 +1,153 @@
...
+   * @returns {string}
+   */
+  toFormula() {
+    return this.values(String).toArray().join(" + ");
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Find the highest value among all of the provided rules, or `-Infinity` of no rules are available.
```
Fyorl:
```suggestion
   * Find the highest value among all of the provided rules, or `-Infinity` if no rules are available.
```

But maybe it would be neater if the empty case returned `null`. That way you can easily do a fallback: `toLargest() ?? 0`. Hard to say because we don't use it anywhere yet. I don't feel strongly.

module/documents/applied-rules.mjs
```diff
@@ -0,0 +1,153 @@
...
+   * @returns {number}
+   */
+  toLargest() {
+    return this.values(Number).reduce((max, value) => value > max ? value : max, -Infinity);
```
Fyorl: Do we accept purely numerics or are we potentially ok with deterministic roll expressions also?

module/documents/applied-rules.mjs
```diff
@@ -0,0 +1,153 @@
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Transform each rule element into its underlying value.
+   * @param {Number|String} [type]  Transform value into specific primitive type.
+   * @returns {RulesIterator}
+   */
+  values(type) {
+    return new RulesIterator(this.map(r => type ? type(r.value ?? r) : r.value ?? r));
```
Fyorl: `change.value` is `initial: ""`. So having one blank value somewhere can cause `toFormula` to produce an invalid formula like `1d4 + +`. It can potentially cause problems for `toLargest` and `toSmallest` too.

```suggestion
    return new RulesIterator(this.map(r => r.value ?? r).filter(_ => _).map(v => type ? type(v) : v));
```

module/data/activity/attack-data.mjs
```diff
@@ -257,7 +258,7 @@ export default class BaseAttackActivityData extends BaseActivityData {
    * @returns {{ data: object, parts: string[] }}
    */
   getAttackData({ ammunition, attackMode, situational }={}) {
-    const rollData = this.getRollData();
+    const rollData = this.getRollData({ roll: { attackMode } });
```
Fyorl: I'm trying to think how you would write a condition for something like 'when you make a ranged attack', but it seems very messy. You need to check `activity.attack.type.value` for `ranged`, but also you need to remember to check `{ k: 'roll.attackMode', o: 'in', v: ['thrown', 'thrown-offhand'] }`.

Might be nicer to have correct `type` + `classification` available, something like:
```suggestion
    const rollData = this.getRollData({ roll: {
      attackMode,
      type: {
        value: this.getActionType(attackMode).startsWith("r") ? "ranged" : "melee",
        classification: this.attack.type.classification
      }
    } });
```

I guess we would maybe want to factor out some `getActionType` logic so we don't need to do the ugly `startsWith("r")`

Fyorl (changes requested):
Solid design, see my comments.

module/documents/applied-rules.mjs
arbron: Since this is mainly intended to pass through to `minimum` and `maximum` which only take numeric values we'll need to resolve to numbers before this. I image we would add a `resolve(rollData)` step that resolves all of the values before it gets to this point (so you'd do `AppliedRules.collect(...).filterWith(rollData).resolve(rollData).toLargest()`).

I'll clarify this in a followup PR when I address the min & max rules.

module/documents/applied-rules.mjs
arbron: I was planning this for the `minimum` and `maximum` options in rolls, which accept `Infinite` values. I'll just pull these methods out of this PR and hold them back for that followup PR so we can examine them in context.

module/documents/activity/mixin.mjs
arbron: ~~Revisiting this, I think I'll just do `Object.assign(this.getRollData(), { roll: { attackMode } })` and see how that works.~~

Nevermind, changed to support attack type.

module/data/activity/attack-data.mjs
arbron: I had this in another branch, but I've pulled it back into this one. The attack activity populates `roll.attack.type` in the roll data dynamically using the `attackMode`, so you would just do `{ k: "roll.attack.type", v: "ranged" }`.

Fyorl (approved):

PR #7188, opened 2026-06-24
[#6619] Add AE expiry events for short and long rests

### LINKED ISSUE #6619: Allow Active Effects to expire after certain time periods like Short/Long Rests
There are probably countless active effects that go along the lines of `"Until you finish a short/long rest, x is true"`. It would be nice if we could denote that an AE expires on a short/long rest in order for it to automatically end after that specific time period. Not to name all of them, but I think
- Spell Effects
- Class/Subclass Features
- Monster Features
- Trap & Environmental Effects
Could all benefit from this. Would be certainly nice to have once the v14 AE upgrades are in full effect.

arbron:
Adds "On a Long Rest" and "On a Short Rest" expiry events.

Closes #6619

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2675,8 +2675,8 @@
   "DropHint": "Drop an Active Effect here",
   "Empty": "No associated effects. Use the <i class=\"fas fa-plus\"></i> button above to create one, or select an existing effect from the drop-down.",
   "Expiry": {
-    "LongRest": "On a Long Rest",
-    "ShortRest": "On a Short Rest"
+    "LongRest": "End of Long Rest",
+    "ShortRest": "End of Short Rest"
   },
   "Label": "Available Effects",
   "New": "New Effect",
--- a/lang/en.json
+++ b/lang/en.json
@@ -2674,6 +2674,10 @@
   },
   "DropHint": "Drop an Active Effect here",
   "Empty": "No associated effects. Use the <i class=\"fas fa-plus\"></i> button above to create one, or select an existing effect from the drop-down.",
+  "Expiry": {
+    "LongRest": "On a Long Rest",
+    "ShortRest": "On a Short Rest"
+  },
   "Label": "Available Effects",
   "New": "New Effect",
   "RIDER": {
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -2949,6 +2949,7 @@ DND5E.restTypes = {
     icon: "fa-solid fa-utensils",
     dialogClass: ShortRestDialog,
     activationPeriods: ["shortRest"],
+    expiryEvents: ["shortRest"],
     recoverPeriods: ["sr"],
     recoverSpellSlotTypes: new Set(["pact"])
   },
@@ -2964,6 +2965,7 @@ DND5E.restTypes = {
     dialogClass: LongRestDialog,
     newDay: true,
     activationPeriods: ["longRest"],
+    expiryEvents: ["longRest", "shortRest"],
     recoverHitDice: true,
     recoverHitPoints: true,
     recoverPeriods: ["lr", "sr"],
@@ -4680,6 +4682,19 @@ DND5E.calendarDeltasRecoveryMapping = new Map([
   ["sunrises", "dawn"],
   ["sunsets", "dusk"]
 ]);
+
+/* -------------------------------------------- */
+/*  Expiry Events                               */
+/* -------------------------------------------- */
+
+/**
+ * Active effect expiry events provided by the system.
+ * @enum {string}
+ */
+DND5E.expiryEvents = Object.seal({
+  longRest: "DND5E.EFFECT.Expiry.LongRest",
+  shortRest: "DND5E.EFFECT.Expiry.ShortRest"
+});
 
 /* -------------------------------------------- */
 /*  Requests                                    */
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -2427,7 +2427,7 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       for ( const item of this.items ) {
         for ( const effect of item.effects ) {
           if ( effect.getFlag("dnd5e", "dependentOn") ) continue;
-          if ( effect.isAppliedEnchantment || effect.transfer ) expireEffect(effect);
+          if ( effect.isAppliedEnchantment ) expireEffect(effect);
         }
       }
       const operations = expired.entries()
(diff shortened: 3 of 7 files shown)


lang/en.json
```diff
@@ -2451,6 +2451,10 @@
   },
   "DropHint": "Drop an Active Effect here",
   "Empty": "No associated effects. Use the <i class=\"fas fa-plus\"></i> button above to create one, or select an existing effect from the drop-down.",
+  "Expiry": {
```
roth-michael, line 2677: In #6837 I put the pseudo-expiry i18n under `DND5E.ACTIVEEFFECT.EXPIRIES` and in lowerCamelCase - I'm happy to move mine to align with this but raising it to verify which is preferred.

It may actually be the case that they should be intentionally different, since these are _actual_ expiries, whereas the ones I added in my PR are "fake" ones.

arbron: I'm fine putting them together. In fact we should probably move the few things that are in `DND5E.ACTIVEEFFECT` into `DND5E.EFFECT`.

Match my naming (singular category `Expiry` and UpperCamelCase) because that matches the pattern of these types through the rest of the localization.

Fyorl (commented):
Could you rebase this when you get a chance? I would like to see how it integrates with the expiries we already added.

lang/en.json
```diff
@@ -2674,6 +2674,10 @@
...
+    "LongRest": "On a Long Rest",
```
Fyorl: Maybe 'End of Long Rest', 'End of Short Rest'? Most rules text I've seen usually says 'at the end of a short or long rest'. It also fits better with the other expiries in the drop-down.

module/config.mjs
```diff
@@ -4681,6 +4683,19 @@ DND5E.calendarDeltasRecoveryMapping = new Map([
...
 
+/* -------------------------------------------- */
+/*  Expiry Events                               */
+/* -------------------------------------------- */
+
+/**
+ * Active effect expiry events provided by the system.
+ * @enum {string}
+ */
+DND5E.expiryEvents = Object.seal({
```
Fyorl, line 4694: These expiries don't hide the duration fields like the pseudo-expiries do, and they don't auto-null duration either in `_preCreate`/`_preUpdate`. If a user does enter a duration then the effect won't actually expire at the end of the rest unless the requisite time has also passed. I don't think there are any D&D features that work that way, so we should probably make the same changes as with the pseudo-expiries so users do not fall into that trap.

Fyorl (changes requested):

module/documents/actor/actor.mjs
```diff
@@ -2410,6 +2410,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     await this.deleteEmbeddedDocuments("Item", result.deleteItems, { isRest: true });
     await this.updateEmbeddedDocuments("Item", result.updateItems, { isRest: true });
 
+    // Expire active effects
+    const restConfig = CONFIG.DND5E.restTypes[config.type ?? "long"];
+    for ( const event of restConfig.expiryEvents ?? [] ) {
+      await ActiveEffect.registry.refresh(event, { actors: new Set([this]) });
```
Fyorl: I should have flagged this earlier. But if a user is invoking `_rest` themselves, either via rest request or just clicking the button on their sheet, this `refresh` call won't dispatch to connected clients, so it won't reach the `activeGM` and effects don't expire.

We could check for a `rest` message, which gives us the rest type. There we could check `activeGM` and call `ActiveEffect.registry.refresh`.

Or, since we know the user performing the rest must own the actor and its effects, we could handle the expiry ourselves. Loop over effects, check `isExpiryEvent` on each (no need to check duration), and delete the expired ones.

Fyorl (changes requested):

module/documents/actor/actor.mjs
arbron: Manually expired the events because there is no guarantee that a chat message is created when you rest.

module/documents/actor/actor.mjs
```diff
@@ -2410,6 +2410,22 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     await this.deleteEmbeddedDocuments("Item", result.deleteItems, { isRest: true });
     await this.updateEmbeddedDocuments("Item", result.updateItems, { isRest: true });
 
+    // Expire active effects
+    const expiryEvents = CONFIG.DND5E.restTypes[config.type ?? "long"].expiryEvents ?? [];
```
```suggestion
    const expiryEvents = CONFIG.DND5E.restTypes[config.type ?? "long"]?.expiryEvents ?? [];
```

module/documents/actor/actor.mjs
```diff
@@ -2410,6 +2410,22 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+    if ( expiryEvents.length ) {
+      const expired = new Map();
+      for ( const effect of this.allApplicableEffects() ) {
```
Fyorl: We probably need to match the logic in `Combat#_onExit` (maybe factor out a helper?). If we want to catch enchantments we need something like this:
```js
for ( const effect of actor.effects ) { /* ... */ }
for ( const item of actor.items ) { for ( const effect of item.effects ) { /* ... */ } }
```
And mirror at least the `dependenOn` skip in `shouldDelete`. We could de-dupe in this actor, but we can't de-dupe (easily) across multiple actors that are all taking a rest at the same time with inter-dependent effects.

Fyorl (changes requested):

module/documents/actor/actor.mjs
```diff
@@ -2410,6 +2410,31 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
+          expired.get(effect.parent).push(effect.id);
+        }
+      };
+      for ( const effect of this.effects ) {
+        if ( !effect.getFlag("dnd5e", "dependentOn") ) expireEffect(effect);
+      }
+      for ( const item of this.items ) {
+        for ( const effect of item.effects ) {
+          if ( effect.getFlag("dnd5e", "dependentOn") ) continue;
+          if ( effect.isAppliedEnchantment || effect.transfer ) expireEffect(effect);
```
Fyorl: Riders or effects in profiles are usually `transfer: true` so we would end up deleting those. It's possible to have an effect on an item affecting an actor and that effect has a duration but that's not really what the system expects. It expects `transfer: true` effects on items to be passive effects, riders, or profiles, not active effects.
```suggestion
          if ( effect.isAppliedEnchantment ) expireEffect(effect);
```

Fyorl (approved):

PR #7189, opened 2026-06-26
Use the existing `TYPES.RegionBehavior.dnd5e.applyActiveEffect` localization key

thatlonelybugbear:


arbron:
I didn't do that because I saw no reason to include `5e` in the activity behavior. We need it in the normal region behavior name to distinguish from core's implementation, but the activity behavior only has the single option so the distinction isn't necessary.

PR #7201, opened 2026-07-01
[#2759] Move speeds into `movement.speeds` to allow custom speeds

### LINKED ISSUE #2759: Allow for custom speeds using World Scripts
I want to create custom speeds and it would be nice if I could do it via `CONFIG.DND5E.movementTypes`

arbron:
Moves individual movement speeds into `movement.speeds` as a `MappingField` to allow for custom speeds, shim the old location, and add permanent active effect redirections.

Closes #2759

### DIFF
--- a/module/data/shared/movement-field.mjs
+++ b/module/data/shared/movement-field.mjs
@@ -1,9 +1,11 @@
 import FormulaField from "../fields/formula-field.mjs";
+import MappingField from "../fields/mapping-field.mjs";
 
 const { BooleanField, SetField, StringField } = foundry.data.fields;
 
 /**
  * @import { TravelPace5e } from "../actor/fields/_types.mjs";
+ * @import { MovementData } from "./_types.mjs";
  */
 
 /**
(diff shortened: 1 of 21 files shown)


module/data/shared/movement-field.mjs
```diff
@@ -31,4 +31,58 @@ export default class MovementField extends foundry.data.fields.SchemaField {
...
+      delete movement[key];
+    }
+  }
+
+  /* -------------------------------------------- */
+  /*  Data Shims                                  */
+  /* -------------------------------------------- */
+
+  /**
+   * Apply shims to the senses field so old sense locations still work.
```
Fyorl: Check comment

module/utils.mjs
```diff
@@ -1386,13 +1386,18 @@ export function getHumanReadableAttributeLabel(attr, { actor, item, prefixItemNa
...
   }
 
+  // Movement
+  else if ( attr.startsWith("attributes.movement.speeds.") ) {
+    label = CONFIG.DND5E.movementTypes[attr.split(".")[3]]?.label;
+  } else if ( attr.startsWith("attributes.movement.") ) label = CONFIG.DND5E.movementTypes[attr.split(".")[2]]?.label;
+
   // Senses
   else if ( attr.startsWith("attributes.senses.ranges.") ) label = CONFIG.DND5E.senses[attr.split(".")[3]]?.label;
   else if ( attr.startsWith("attributes.senses.") ) label = CONFIG.DND5E.senses[attr.split(".")[2]]?.label;
```
Fyorl: If all we want is the terminal part, better to express that with `attr.split(".").at(-1)` I think. These 4 lines. Very minor though.

Fyorl (approved):
Thanks for taking care of this

PR #7202, opened 2026-07-01
[#6102] Add school restriction within item choice advancement

### LINKED ISSUE #6102: Add spell school restrictions similar to spell list restrictions to item choice advancements.
Feats like Shadow / Fey Touched could use some spell school restrictions for item choice advancements. Could use spell lists custom made for them but any 1st or 3rd party premium content adding say, necromancy spells means they wont show up when selecting spells for Shadow Touched.

arbron:
Allows for restricting spell selection within an item choice advancement to a certain school.

Closes #6102

### DIFF
--- a/module/applications/advancement/item-choice-flow.mjs
+++ b/module/applications/advancement/item-choice-flow.mjs
@@ -233,7 +233,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
       return;
     }
 
-    const filters = { locked: { additional: {}, documentClass: "Item" } };
+    const filters = { locked: { additional: {}, documentClass: "Item", exclusive: true } };
 
     // Apply restrictions based on type
     if ( config.type ) {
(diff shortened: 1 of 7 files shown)


module/applications/advancement/item-choice-flow.mjs
```diff
@@ -233,7 +233,7 @@ export default class ItemChoiceFlow extends ItemGrantFlow {
       return;
     }
 
-    const filters = { locked: { additional: {}, documentClass: "Item" } };
+    const filters = { locked: { additional: {}, documentClass: "Item", exclusive: true } };
```
Fyorl: I understand why we would want this, but it does prevent players from *narrowing* their results. They can't *uncheck* a spell school if maybe they only want to look at spells from one of their available schools, or they can't increase the min level, for example, even if still within the allowed bounds.

Given that we validate the choices anyway once they have been made, it's not possible for a player to circumvent the constraints even if don't have `exclusive: true`. So I might prefer we don't have it at all, or we add another option that allows for a player to adjust filters to narrow results only but not expand them, though I appreciate that's probably non-trivial effort.

Fyorl (approved):

module/applications/advancement/item-choice-flow.mjs
arbron: I'll pull it out for now and we can consider other options.

PR #7203, opened 2026-07-02
[#7196] Allow for subtracting an instance of advantage or disadvantage

### LINKED ISSUE #7196: AdvantageModeField missing implementation of `_applyChangeSubtract`
Currently just does `value - delta` which can't be right, surely?

Should it remove an instance of (dis)advantage?

Should it act like `add` but invert the delta (simply call `_applyChangeAdd` with `delta * -1`) ?

krbz999:
Closes #7196.

`subtract 1` removes an instance of advantage.

`subtract -1` removes an instance of disadvantage.

`subtract <anything else>` does nothing.

Final count (when comparing advantages to disadvantages) uses `Math.max` to prevent the two counts going below 0.

### DIFF
--- a/module/data/fields/advantage-mode-field.mjs
+++ b/module/data/fields/advantage-mode-field.mjs
@@ -40,6 +40,17 @@ export default class AdvantageModeField extends foundry.data.fields.NumberField 
   /* -------------------------------------------- */
 
   /** @override */
+  _applyChangeSubtract(value, delta, model, change) {
+    if ( (delta !== -1) && (delta !== 1) ) return value;
+    const counts = this.constructor.getCounts(model, change.key);
+    if ( delta === 1 ) counts.advantages.count--;
+    else counts.disadvantage.count--;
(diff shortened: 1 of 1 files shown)


module/data/fields/advantage-mode-field.mjs
```diff
@@ -39,6 +39,17 @@ export default class AdvantageModeField extends foundry.data.fields.NumberField
 
   /* -------------------------------------------- */
 
+  /** @override */
+  _applyChangeSubtract(value, delta, model, change) {
+    if ( (delta !== -1) && (delta !== 1) ) return value;
+    const counts = this.constructor.getCounts(model, change.key);
+    if ( delta === 1 ) counts.advantages.count--;
+    else counts.disadvantage.count--;
```
Fyorl:
```suggestion
    else counts.disadvantages.count--;
```

Fyorl (approved):

PR #7207, opened 2026-07-02
Only show Area of Effect Behaviors for activities w/ shape configured

roth-michael:


arbron (changes requested):
We could put this into `activity-behaviors.hbs` instead so it doesn't have to be repeated. Place the fieldset otherwise display `<section class="hidden"></section>`. 

We should also display the behaviors section if it is a rider activity, because we can't know that the activity won't be put onto a spell that has an area of effect defined.

arbron (approved):

PR #7216, opened 2026-07-04
Handle generating dispositions for area of effect behaviors

arbron:
Adds `BaseActivityBehavior#getDispositions` to determine what dispositions should be applied to a region behavior based on the targeting data from an activity.

Also makes difficult terrain regions and apply active effect regions ignore tokens with the secret disposition.

### DIFF
--- a/module/data/region-behavior/apply-active-effect.mjs
+++ b/module/data/region-behavior/apply-active-effect.mjs
@@ -210,9 +210,10 @@ export class ApplyActiveEffectActivityBehavior extends BaseActivityBehavior {
 
   /** @override */
   createBehaviorData(activity, options={}) {
+    const disposition = activity.actor?.token?.disposition ?? activity.actor?.prototypeToken?.disposition;
     return {
       system: {
-        dispositions: this.getDispositions(activity.target),
+        dispositions: this.getDispositions(activity.target, { relativeTo: disposition }),
         effects: this.effects,
         sizes: this.sizes,
         types: this.types
--- a/module/data/region-behavior/apply-active-effect.mjs
+++ b/module/data/region-behavior/apply-active-effect.mjs
@@ -41,6 +41,7 @@ export default class ApplyActiveEffect5eRegionBehaviorType extends foundry.data.
    * @returns {boolean}
    */
   #evaluateConditions(token) {
+    if ( token.disposition === CONST.TOKEN_DISPOSITIONS.SECRET ) return false;
     if ( this.dispositions.size && !this.dispositions.has(token.disposition) ) return false;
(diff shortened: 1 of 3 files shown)


txm3278:
Does this handle spells or features that specify "creatures you choose" as well or only for disposition?

module/data/region-behavior/base-activity-behavior.mjs
```diff
@@ -105,4 +105,20 @@ export default class BaseActivityBehavior extends foundry.abstract.DataModel {
...
   }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Determine what dispositions this region should apply to or ignore based on activity targeting data.
+   * @param {TargetData} target          Target data from the activity.
+   * @param {object} [options={}]
+   * @param {boolean} [options.ignored]  Should this generate an ignored list rather than a targeted list.
+   * @returns {Set<string>}              Set of dispositions to target.
```
Fyorl:
```suggestion
   * @returns {Set<number>}              Set of dispositions to target or ignore.
```

module/data/region-behavior/base-activity-behavior.mjs
```diff
@@ -105,4 +105,20 @@ export default class BaseActivityBehavior extends foundry.abstract.DataModel {
...
+   */
+  getDispositions(target, { ignored }={}) {
+    const { HOSTILE, NEUTRAL, FRIENDLY } = CONST.TOKEN_DISPOSITIONS;
```
Fyorl, line 120: I think these dispositions need to be relative to the caster. If they are casting a spell that affects only allies and their disposition is HOSTILE, `getDispositions` should return `[HOSTILE]`. Since we have the activity, it should be possible to pass a `source, target, options={}`. Or maybe if sometimes there isn't a `source` then it can be an option and we only do relative disposition when it's present.

Fyorl (changes requested):

module/data/region-behavior/apply-active-effect.mjs
```diff
@@ -209,12 +210,13 @@ export class ApplyActiveEffectActivityBehavior extends BaseActivityBehavior {
 
   /** @override */
   createBehaviorData(activity, options={}) {
+    const disposition = activity.actor?.token?.disposition ?? activity.actor?.prototypeToken?.disposition;
```
Fyorl, line 213: Seems fine in 99% of cases but I can definitely start to see how we might want the actual token in a lot of places. A linked actor's proto token might have friendly disposition but their token instance on the canvas might temporarily hostile or something because they're dominated for example.

Not for this PR though. We should be able to add the token uuid to the flags in `TemplatePlacement.fromActivity`, then we read it in the `placeTemplateBehaviors` hook and pass it in here.

Fyorl (approved):

PR #7229, opened 2026-07-06
Add single `attributes.movement.multiplier` field to fix doubling

arbron:
Adds a non-persisted multiplier field to movement data to be targeted by active effects. This field is required to ensure that speed multiplication occurs at the correct point and that speed multiplication isn't done more than once (for example, if a speed uses the formula `@attributes.movement.speed`, then walking speed would be halved first and then this speed would be halved, ending up with only a quarter of the desired speed).

### DIFF
--- a/module/data/shared/movement-field.mjs
+++ b/module/data/shared/movement-field.mjs
@@ -1,7 +1,7 @@
 import FormulaField from "../fields/formula-field.mjs";
 import MappingField from "../fields/mapping-field.mjs";
 
-const { BooleanField, SetField, StringField } = foundry.data.fields;
+const { BooleanField, NumberField, SetField, StringField } = foundry.data.fields;
 
 /**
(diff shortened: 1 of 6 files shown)


module/data/shared/movement-field.mjs
```diff
@@ -15,6 +15,9 @@ export default class MovementField extends foundry.data.fields.SchemaField {
   constructor(fields={}, { initialUnits=null, ...options }={}) {
     fields = {
       bonus: new FormulaField({ deterministic: true, label: "DND5E.MOVEMENT.FIELDS.bonus.label" }),
+      multiplier: new NumberField({
+        min: 0, initial: 1, persisted: false, label: "DND5E.MOVEMENT.FIELDS.multiplier.label"
```
Fyorl, line 19: Don't think this label exists.

Fyorl (approved):

PR #7232, opened 2026-07-07
[#5942, #6628] Filter containers & disabled packs in index lookups

### LINKED ISSUE #5942: Extend `game.documentIndex` to use `dnd5e.packSourceConfiguration` for lookups
Unsure if this class can actually be extended, however a monkeypatch like so was useful to me:
```js
const olookup = game.documentIndex.lookup;
game.documentIndex.lookup = function(query, { filterEntries, ...options } = {}) {
    const packs = Object.entries(game.settings.get("dnd5e", "packSourceConfiguration")).filter(k => k[1] === false).map(k => k[0]);
    filterEntries ??= entry => !packs.includes(entry.pack);
    return olookup.call(this, query, { ...options, filterEntries });
};
game.documentIndex.lookup("fireball");
```

When doing data entry, writing descriptions for items, etc., it's nice to be able to quickly get a link to other documents, especially so for magic items that allow for casting spells. Currently we have to guess which of the 3+ results all named "fireball" are from the PHB (or whichever module you wish to link to).

### LINKED ISSUE #6628: Prevent items in containers from appearing in Content Link Matching list
Currently the Content Link Matching system finds everything in a compendium which includes items in containers, leading to a bunch of duplicate entries for common items that appear in several equipment packs:

Image

This duplication makes it impossible to use content matching to select the correct item. The system should filter out these entries and only show ones that are not in containers.

arbron:
Patches `game.documentIndex#lookup` to filter entries so items in containes aren't displayed and documents in packs that are disabled in the "Configure Sources" window are also hidden.

In order to ensure that the container filtering works properly, this re-indexes all Item compendiums during the ready hook.

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -507,7 +507,34 @@ Hooks.once("setup", function() {
   document.body.append(probe);
   document.documentElement.style.setProperty("--dnd5e-scrollbar-width", `${probe.offsetWidth - probe.clientWidth}px`);
   probe.remove();
+
+  // Patch Document Index
+  _originalLookup = game.documentIndex.lookup;
+  game.documentIndex.lookup = documentIndexLookup;
 });
+
+/* --------------------------------------------- */
+
+let _originalLookup;
+
+/**
+ * Return entries that match the given string prefix.
+ * @param {string} query                      The search prefix or phrase.
+ * @param {object} [options]                  Additional options to configure behaviour.
+ * @param {string[]} [options.ignoreCompendiumFiltering]  Don't filter based on pack source configuration.
+ * @returns {Record<string, WordTreeEntry[]>} A number of entries that have the given prefix, grouped by document
+ *                                            type.
+ */
+function documentIndexLookup(query, { ignoreCompendiumFiltering, ...options }={}) {
+  return _originalLookup.call(this, query, {
+    ...options,
+    filterEntries: entry => {
+      if ( fromUuidSync(entry.uuid, { strict: false })?.system?.container ) return false;
+      if ( !ignoreCompendiumFiltering && (dnd5e.settings.packSourceConfiguration[entry.pack] === false) ) return false;
+      return options.filterEntries?.(entry) ?? true;
+    }
+  });
+}
 
 /* --------------------------------------------- */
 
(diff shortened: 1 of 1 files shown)


dnd5e.mjs
```diff
@@ -590,6 +617,9 @@ Hooks.once("ready", function() {
   // Register items by type
   dnd5e.registry.items.initialize();
 
+  // Re-index any item compendiums to ensure container property is indexed
+  game.packs.filter(p => p.documentName === "Item").forEach(p => p.getIndex());
```
Fyorl: Feel like we must already be re-indexing in the `ItemRegistry#initialize` call above, no?

dnd5e.mjs
```diff
@@ -507,10 +507,37 @@ Hooks.once("setup", function() {
...
+ * @param {object} [options]                  Additional options to configure behaviour.
+ * @param {string[]} [options.ignoreCompendiumFiltering]  Don't filter based on pack source configuration.
+ * @returns {Record<string, WordTreeEntry[]>} A number of entries that have the given prefix, grouped by document
+ *                                            type.
+ */
+function documentIndexLookup(query, { ignoreCompendiumFiltering, ...options }={}) {
+  return _originalLookup.call(this, query, {
+    ...options,
+    filterEntries: entry => {
+      if ( fromUuidSync(entry.uuid, { strict: false })?.system?.container ) return false;
```
Fyorl:
```suggestion
      if ( entry.entry.system?.container ) return false;
```

dnd5e.mjs
```diff
@@ -507,10 +507,37 @@ Hooks.once("setup", function() {
...
 
 /* --------------------------------------------- */
 
+let _originalLookup;
+
+/**
+ * Return entries that match the given string prefix.
+ * @param {string} query                      The search prefix or phrase.
+ * @param {object} [options]                  Additional options to configure behaviour.
+ * @param {string[]} [options.ignoreCompendiumFiltering]  Don't filter based on pack source configuration.
```
Fyorl:
```suggestion
 * @param {boolean} [options.ignoreCompendiumFiltering]  Don't filter based on pack source configuration.
```

Fyorl (approved):
Raised an issue for this at foundryvtt/foundryvtt#14727 but this patching looks pretty reasonable to go ahead with for now.

dnd5e.mjs
arbron: The one issue with that is that the registry uses `CompendiumBrowser#fetch`, which means packs that are currently not included in the source configuration won't be re-indexed. If you then change the pack filtering without reloading items in containers will start appearing in your searches for previously hidden compendiums. Probably not a huge issue in practice.

PR #7243, opened 2026-07-11
[#6315] Add rule bonuses to damage rolls

arbron:
To ensure bonuses are only applied once per damage roll while still allowing different filtering for each damage part, this adds a `consume()` method to `RulesIterator` that filters out any consumed rules and marks the remaining rules as consumed. Done after the filtering step this ensures that each rule can only be used once per damage roll. Calling `reset()` after all of the parts are prepared clears the consumed status.

Reworked the `roll` option passed into `getRollData` so now it is a proper options object that individual `getRollData` methods can respond to rather than data assigned to the roll data (since that can be easily handled after roll data is created).

### DIFF
--- a/module/data/activity/attack-data.mjs
+++ b/module/data/activity/attack-data.mjs
@@ -284,44 +284,41 @@ export default class BaseAttackActivityData extends BaseActivityData {
 
   /** @override */
   getDamageConfig(config={}, options={}) {
-    const rollConfig = super.getDamageConfig(config, options);
-
-    // Handle ammunition
+    // Copy properties from selected ammunition
     const ammo = config.ammunition?.system;
     if ( ammo ) {
       const properties = Array.from(ammo.properties).filter(p => CONFIG.DND5E.itemProperties[p]?.isPhysical);
       if ( this.item.system.properties?.has("mgc") && !properties.includes("mgc") ) properties.push("mgc");
+      config.properties ??= [];
+      for ( const property of properties ) {
+        if ( !config.properties.includes(property) ) config.properties.push(property);
+      }
+    }
 
-      // Add any new physical properties from the ammunition to the damage properties
-      for ( const roll of rollConfig.rolls ) {
-        for ( const property of properties ) {
-          if ( !roll.options.properties.includes(property) ) roll.options.properties.push(property);
-        }
+    options.rollData ??= this.getRollData({ roll: { attackMode: config.attackMode } });
+    const rollConfig = super.getDamageConfig(config, options);
+
+    // Add the ammunition's damage
+    if ( ammo?.damage.base.formula ) {
+      const basePartIndex = rollConfig.rolls.findIndex(i => i.base);
+      const damage = ammo.damage.base.clone(ammo.damage.base);
+
+      // If mode is "replace" and base part is present, replace the base part
+      if ( ammo.damage.replace & (basePartIndex !== -1) ) {
+        damage.base = true;
+        rollConfig.rolls.splice(
+          basePartIndex, 1,
+          this._processDamagePart(damage, config, options.rollData, basePartIndex, options.formulaOptions)
+        );
       }
 
-      // Add the ammunition's damage
-      if ( ammo.damage.base.formula ) {
-        const basePartIndex = rollConfig.rolls.findIndex(i => i.base);
-        const damage = ammo.damage.base.clone(ammo.damage.base);
-        const rollData = this.getRollData();
-
-        // If mode is "replace" and base part is present, replace the base part
-        if ( ammo.damage.replace & (basePartIndex !== -1) ) {
-          damage.base = true;
-          rollConfig.rolls.splice(
-            basePartIndex, 1,
-            this._processDamagePart(damage, config, rollData, basePartIndex, options.formulaOptions)
-          );
-        }
-
-        // Otherwise stick the ammo damage after base part (or as first part)
-        else {
-          damage.ammo = true;
-          rollConfig.rolls.splice(
-            basePartIndex + 1, 0,
-            this._processDamagePart(damage, rollConfig, rollData, basePartIndex + 1, options.formulaOptions)
-          );
-        }
+      // Otherwise stick the ammo damage after base part (or as first part)
+      else {
+        damage.ammo = true;
+        rollConfig.rolls.splice(
+          basePartIndex + 1, 0,
+          this._processDamagePart(damage, rollConfig, options.rollData, basePartIndex + 1, options.formulaOptions)
+        );
       }
     }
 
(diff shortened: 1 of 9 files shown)


module/data/activity/save-data.mjs
```diff
@@ -146,4 +146,13 @@ export default class BaseSaveActivityData extends BaseActivityData {
 
     return rollConfig;
   }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  getRollData(options={}) {
+    const rollData = super.getRollData(options);
+    if ( rollData.roll ) rollData.roll.type = "save";
```
Fyorl: A bit unsure under what circumstances this is set. It also seems to conflict with 7246 which is setting `roll.type: "ability"` when rolling a saving throw.

module/data/activity/base-activity.mjs
```diff
@@ -816,15 +825,16 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
       if ( this.item.system.damage?.bonus ) parts.push(String(this.item.system.damage.bonus));
     }
 
-    const lastType = this.item.getFlag("dnd5e", `last.${this.id}.damageType.${index}`);
+    const ruleBonus = AppliedRules.collect("damage:bonus", this.actor, this.item)
```
Fyorl: I'm not sure about this design. Do we really need to assign this CONSUMED state on the centrally registered parts so that we can call `AppliedRules.collect` in this context-free way here and in `getDamageConfig`? Even if the guarantees that you always get the same set of parts each time are pretty good, I feel like a boring solution where we do something like `collect().toArray()`, pass that array in here, and then splice or filter parts out of the array as they are consumed is probably a bit easier to read and potentially debug/maintain down the line. Maybe I misunderstood the call stack and it's more tangled than that? In which case, this solution would start looking more attractive.

module/data/activity/attack-data.mjs
```diff
@@ -372,9 +372,11 @@ export default class BaseAttackActivityData extends BaseActivityData {
   /** @inheritDoc */
   getRollData(options={}) {
     const rollData = super.getRollData(options);
-    if ( rollData.roll ) {
+    if ( options.roll ) {
+      rollData.roll ??= {};
       rollData.roll.attack ??= {};
       rollData.roll.attack.classification = this.attack.type.classification;
+      rollData.roll.attack.mode = options.roll.attackMode;
       rollData.roll.attack.type = rollData.roll.attack.mode?.includes("thrown") ? "ranged" : this.attack.type.value;
```
Fyorl: Missed this before but checking `thrown` is not enough. The attack mode itself might be 'ranged' for attacks that are 'melee or ranged attack'.

Fyorl (changes requested):
Having a bit of trouble reviewing this trio of PRs that seem to all have different ways of handling the roll data's `roll` object. As far as I can tell, it's just additional data, which doesn't seem like it would warrant its own rollData option.

7243 and 7245 seem to be the same thing, but I guess 7245 is using the 'older' method? 7246 looks like it's intending to pass extra data as part of `options.roll` judging by the typedef amendments, but doesn't actually do so and instead appends it afterwards.

What is the actual intended design here?

module/data/activity/save-data.mjs
arbron: This is used to indicate that damage is from a save, rather from an attack (since when rolling the save itself, it doesn't go through activity roll data).

arbron:
> Having a bit of trouble reviewing this trio of PRs that seem to all have different ways of handling the roll data's `roll` object. As far as I can tell, it's just additional data, which doesn't seem like it would warrant its own rollData option.
> 
> 7243 and 7245 seem to be the same thing, but I guess 7245 is using the 'older' method? 7246 looks like it's intending to pass extra data as part of `options.roll` judging by the typedef amendments, but doesn't actually do so and instead appends it afterwards.
> 
> What is the actual intended design here?

My thinking is that we don't need to pass in all of the potential extra data into `getRollData`, we only need to pass in specific data that `getRollData` might need to respond to (so in this case `attackMode`, because we need that information to populate some other attack info, and in other cases just the presence of `roll` is enough). Any other data necessary for the roll can be added to the roll data after the fact easily since it doesn't actually affect the process of creating the roll data (passing in attack mode is useful because it means we can avoid recreating all of the other associated data in various places).

```javascript
const rollData = this.getRollData({ data: { roll: { attack: { mode: attackMode } } } });

// vs

const rollData = this.getRolldata({ roll: { attackMode } });

// or perhaps (this one is actually feeling a better design in the end)

const rollData = this.getRollData({ attackMode, roll: true });
```

The other two PRs are still set up around the method in the initial rules PR because I didn't want propose the change on three separate branches at the same time.

Fyorl:
> My thinking is that we don't need to pass in all of the potential extra data into `getRollData`, we only need to pass in specific data that `getRollData` might need to respond to (so in this case `attackMode`, because we need that information to populate some other attack info, and in other cases just the presence of `roll` is enough). Any other data necessary for the roll can be added to the roll data after the fact easily since it doesn't actually affect the process of creating the roll data (passing in attack mode is useful because it means we can avoid recreating all of the other associated data in various places).

Makes sense, just the typedef in 7246 I think made me uncertain. My vote would be your 2nd proposal and we have the option typed as `boolean|RollDataRollOptions` so that something without any roll-specific options can still pass `roll: true` but we don't mix these roll options with all other `rollData` options.

module/data/activity/attack-data.mjs
Fyorl: Since these are added after the call to `super`, I don't think they will be available to conditions. Similar issue with the damage part splicing below.

arbron: Yep, I also corrected an issue with the "replace" option for ammunition damage.

module/data/activity/base-activity.mjs
arbron: I've done some reworking so the consumed status of rules is now tracked on a separate external `Set`, rather than internally on each rule. This allows us to retain the `RulesIterator` so we don't have to re-implement the `filterWith` and `toFormula` logic.

module/data/activity/attack-data.mjs
```diff
@@ -284,47 +284,20 @@ export default class BaseAttackActivityData extends BaseActivityData {
...
-            this._processDamagePart(damage, rollConfig, rollData, basePartIndex + 1, options.formulaOptions)
-          );
-        }
+      config.properties ??= [];
+      for ( const property of properties ) {
+        if ( !config.properties.includes(property) ) config.properties.push(property);
       }
     }
 
+    options.rollData ??= this.getRollData({ roll: { attackMode: config.attackMode } });
```
Fyorl: We already do this in `super`.


```suggestion
```

module/data/activity/base-activity.mjs
```diff
@@ -767,8 +768,19 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
     if ( !this.damage?.parts ) return foundry.utils.mergeObject({ rolls: [] }, config);
 
     const rollConfig = foundry.utils.deepClone(config);
-    rollData ??= this.getRollData();
-    rollConfig.rolls = this.damage.parts
+    rollData ??= this.getRollData({ roll: { attackMode: config.attackMode } });
+    rollData.roll ??= {};
+    rollData.rules = {
```
Fyorl: Should this go in `rollData`? Can we send it as part of `_processDamagePart` options`?

```js
.map((d, index) => this._processDamagePart(d, rollConfig, rollData, index, { ...formulaOptions, rules })
```

I don't feel that strongly either way.

Fyorl (approved):

PR #7246, opened 2026-07-11
[#6315] Expand rule bonuses to checks & saves

arbron:
Adds rules bonuses to checks and saves. The `roll` object within roll data has been moved up to actor so it is available everywhere. This gets different useful data depending on the roll type:

- `ability`
- `proficient` (multiplier 1 or higher, not half-proficiency)
- `skill` or `tool`
- `type` (e.g. `ability`, `skill`, `initiative`, etc.)

Also adds the new `d20` key which applied to attack, check, and save rolls.

Screenshot 2026-07-11 at 11 05 48

### DIFF
--- a/module/data/actor/templates/common.mjs
+++ b/module/data/actor/templates/common.mjs
@@ -168,7 +168,7 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
       const checkBonusRules = simplifyBonus(
         AppliedRules.collect("check:bonus", this.parent).filterWith(rollData).toFormula(), rollData
       );
-      abl.checkBonus = checkBonusAbl + checkBonus;
+      abl.checkBonus = checkBonusAbl + checkBonusRules + checkBonus;
 
       const saveBonusAbl = simplifyBonus(abl.bonuses?.save, rollData);
       const cover = id === "dex" ? Math.max(ac?.cover ?? 0, this.parent.coverBonus) : 0;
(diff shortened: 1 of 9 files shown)


module/documents/_types.mjs
```diff
@@ -189,12 +178,27 @@
 
 /**
  * @typedef RollData
+ * @property {RollDescription} [roll]  Data describing a specific roll being performed.
+ */
+
+/**
+ * @typedef RollDescription
```
Fyorl, line 185: As mentioned in #7243 we don't need this typedef.

module/data/actor/templates/common.mjs
```diff
@@ -158,15 +159,25 @@ export default class CommonTemplate extends ActorDataModel.mixin(CurrencyTemplat
...
+      abl.saveProf = abl.merged ? originalAbility.saveProf.clone() : new Proficiency(prof, abl.proficient);
 
-      const cover = id === "dex" ? Math.max(ac?.cover ?? 0, this.parent.coverBonus) : 0;
-      abl.saveBonus = saveBonusAbl + saveBonus + cover;
+      rollData = { ...rollData };
+      rollData.roll = { ability: id, proficient: abl.checkProf.multiplier >= 1, type: "ability" };
 
-      abl.saveProf = abl.merged ? originalAbility.saveProf.clone() : new Proficiency(prof, abl.proficient);
       const checkBonusAbl = simplifyBonus(abl.bonuses?.check, rollData);
+      const checkBonusRules = simplifyBonus(
```
Fyorl, line 168: Not used

Fyorl (approved):

module/documents/_types.mjs
arbron: I would like to have somewhere where we document what can potentially be found in the roll data used for rules filtering, so this felt like a good place because this is information that may end up in `RollData` (just not roll data options).

PR #7249, opened 2026-07-13
Clean up `updateDowned` by using `fromStatusEffect`

roth-michael:
A minor improvement, but a good one I think. Any calls to `updateDowned` were already being `await`ed, so the change to `async` isn't impactful. Using `fromStatusEffect` instead of manually recreating the results of it twice (once for each condition) saves some lines of code, and means less tech debt if there are any future changes to the way we'd want to create such an effect.


thatlonelybugbear:
Forcing `overlay: true` for Dead, seems a bit too invasive.

Maybe a setting for both statuses overlay behaviour would be better received?

PS: Probably not the best spot to discuss that, but oh well, if needed I can split to a new issue?

Fyorl (approved):

PR #7253, opened 2026-07-14
[#6315] Re-apply damage rule bonuses when type is changed

arbron:
Adds `_buildDamageConfig` that is passed to the damage roll dialog so that the rule bonus can be re-calculated when the damage type is changed within the dialog.


module/documents/activity/mixin.mjs
```diff
@@ -937,6 +941,36 @@ export default function ActivityMixin(Base) {
...
+     * @param {FormDataExtended|void} formData          Any data entered into the rolling prompt.
+     * @param {number} index                            Index of the roll within all rolls being prepared.
+     */
+    _buildDamageConfig(process, config, formData, index) {
+      if ( index === 0 ) process.rules = {
+        bonus: AppliedRules.collect(`${this.constructor.damageRuleCategory}:bonus`, this.actor, this.item).toArray(),
+        consumed: new Set()
+      };
+
+      config.data.roll.damageType = config.options.type;
```
Fyorl, line 960: Is it possible parts passed via `config.rolls` won't necessarily have `data.roll`?


```suggestion
      config.data.roll ??= {};
      config.data.roll.damageType = config.options.type;
```

Fyorl (approved):

PR #7254, opened 2026-07-14
[#6315] Implement minimum & maximum rule types

arbron:
Adds a new "Maximum" and "Minimum" rule change types to target D20 rolls. These can be grabbed individually, or combined into one set of rules using the `range` key. The `toRange` method is then used to combine these into a final range.

### Todo
- [ ] Implement roll ranges for non-d20 rolls (such as Hit Dice)

### DIFF
--- a/module/config.mjs
+++ b/module/config.mjs
@@ -3917,6 +3917,18 @@ DND5E.activeEffectChangeTypes = Object.freeze({
     defaultPriority: 100,
     handler: ActiveEffect5e._applyChangeRule,
     skipConditions: true
+  },
+  "dnd5e.maximum": {
+    label: "DND5E.ACTIVEEFFECT.ChangeType.Maximum.Label",
+    defaultPriority: 100,
+    handler: ActiveEffect5e._applyChangeRule,
+    skipConditions: true
+  },
+  "dnd5e.minimum": {
+    label: "DND5E.ACTIVEEFFECT.ChangeType.Minimum.Label",
+    defaultPriority: 100,
+    handler: ActiveEffect5e._applyChangeRule,
+    skipConditions: true
   }
 });
 
--- a/module/documents/activity/attack.mjs
+++ b/module/documents/activity/attack.mjs
@@ -262,9 +262,11 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
     const options = {
       ...(config.options ?? {}),
       maximum: this.actor
-        ? AppliedRules.collect("attack:maximum", this.actor, this.item).filterWith(data).toSmallest() : undefined,
+        ? AppliedRules.collect("attack:maximum", this.actor, this.item).filterWith(data).resolve(data).toSmallest()
+        : undefined,
       minimum: this.actor
-        ? AppliedRules.collect("attack:minimum", this.actor, this.item).filterWith(data).toLargest() : undefined,
+        ? AppliedRules.collect("attack:minimum", this.actor, this.item).filterWith(data).resolve(data).toLargest()
+        : undefined,
     };
     if ( ammunition !== undefined ) options.ammunition = ammunition;
     if ( attackMode !== undefined ) options.attackMode = attackMode;
--- a/module/documents/activity/attack.mjs
+++ b/module/documents/activity/attack.mjs
@@ -123,13 +123,11 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
       this.actor.system, [],
       AppliedRules.collect("attack:advantage", this.actor, this.item).filterWith(rollData).toAdvantageCounts()
     ) : {};
-    const { maximum, minimum } = this.actor
-      ? AppliedRules.collect("attack:range", this.actor, this.item).filterWith(rollData).toRange() : {}
 
     rollConfig.hookNames = [...(config.hookNames ?? []), "attack", "d20Test"];
     rollConfig.rolls = [CONFIG.Dice.D20Roll.mergeConfigs({
       options: {
-        advantage, disadvantage, maximum, minimum,
+        advantage, disadvantage,
         ammunition: rollConfig.ammunition,
         attackMode: rollConfig.attackMode,
         criticalSuccess: this.criticalThreshold,
@@ -261,7 +259,13 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
     const mastery = formData?.get("mastery") ?? process.mastery;
 
     let { parts, data } = this.getAttackData({ ammunition, attackMode });
-    const options = config.options ?? {};
+    const options = {
+      ...(config.options ?? {}),
+      maximum: this.actor
+        ? AppliedRules.collect("attack:maximum", this.actor, this.item).filterWith(data).toSmallest() : undefined,
+      minimum: this.actor
+        ? AppliedRules.collect("attack:minimum", this.actor, this.item).filterWith(data).toLargest() : undefined,
+    };
     if ( ammunition !== undefined ) options.ammunition = ammunition;
     if ( attackMode !== undefined ) options.attackMode = attackMode;
     if ( mastery !== undefined ) options.mastery = mastery;
--- a/module/documents/activity/attack.mjs
+++ b/module/documents/activity/attack.mjs
@@ -259,15 +259,14 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
     const mastery = formData?.get("mastery") ?? process.mastery;
 
     let { parts, data } = this.getAttackData({ ammunition, attackMode });
-    const options = {
-      ...(config.options ?? {}),
+    const options = foundry.utils.mergeObject({
       maximum: this.actor
         ? AppliedRules.collect("attack:maximum", this.actor, this.item).filterWith(data).resolve(data).toSmallest()
         : undefined,
       minimum: this.actor
         ? AppliedRules.collect("attack:minimum", this.actor, this.item).filterWith(data).resolve(data).toLargest()
         : undefined,
-    };
+    }, config.options ?? {});
     if ( ammunition !== undefined ) options.ammunition = ammunition;
     if ( attackMode !== undefined ) options.attackMode = attackMode;
     if ( mastery !== undefined ) options.mastery = mastery;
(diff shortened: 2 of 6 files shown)


module/documents/applied-rules.mjs
```diff
@@ -56,6 +57,11 @@ export default class AppliedRules extends Map {
   static *#collect(rule, actor, item) {
     const [category, key] = rule.split(":");
     if ( D20_TESTS.has(category) ) yield* AppliedRules.#collect(`d20:${key}`, actor, item);
+    if ( key === "range" ) {
```
Fyorl: Needs to go above `D20_TESTS` or otherwise `attack:range` expands to `d20:maximum`+`d20:minimum` on line 59, then enters this block and becomes `attack:maximum`->`d20:maximum`+`attack:minimum`->`d20:minimum`. So the result is duplicated `d20:*`. I guess it still works because they're idempotent? Not sure but don't think it's ideal either way.

module/documents/actor/actor.mjs
```diff
@@ -1570,13 +1570,14 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
     }, rollData);
 
-    const { advantage, disadvantage } = AdvantageModeField.combineFields(this.system, [
-      `abilities.${config.ability}.${type}.roll.mode`
-    ], AppliedRules.collect(`${type}:advantage`, this).filterWith(rollData).toAdvantageCounts());
     const options = {
-      advantage, disadvantage,
-      maximum: ability?.[type]?.roll.max,
-      minimum: ability?.[type]?.roll.min
+      ...AdvantageModeField.combineFields(this.system, [
```
Fyorl: We get some things that aren't `advantage`, `disadvantage` now, like `mode`. Don't know if we care.

module/documents/activity/attack.mjs
```diff
@@ -123,11 +123,13 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
       this.actor.system, [],
       AppliedRules.collect("attack:advantage", this.actor, this.item).filterWith(rollData).toAdvantageCounts()
     ) : {};
+    const { maximum, minimum } = this.actor
+      ? AppliedRules.collect("attack:range", this.actor, this.item).filterWith(rollData).toRange() : {}
```
Fyorl:
```suggestion
      ? AppliedRules.collect("attack:range", this.actor, this.item).filterWith(rollData).toRange() : {};
```

module/documents/applied-rules.mjs
```diff
@@ -175,6 +181,45 @@ class RulesIterator extends Iterator {
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Find the minimum and maximum limit among all of the provided rules.
+   * @param {Partial<RollRange>} [range]  Initial range to fine tune.
+   * @returns {RollRange}
+   */
+  toRange(range) {
+    return this.reduce((range, rule) => {
```
Fyorl: Should this route via `values()` or at least I think we need to check for empty string in the body, otherwise a blank change ends up clamping every roll to 1.

module/documents/applied-rules.mjs
```diff
@@ -175,6 +181,45 @@ class RulesIterator extends Iterator {
 
   /* -------------------------------------------- */
 
+  /**
+   * Find the highest value among all of the provided rules, or `-Infinity` if no rules are available.
+   * @param {number} [initial]  Starting value to compare against.
+   * @returns {number}
+   */
+  toLargest(initial) {
```
Fyorl, line 185: You mentioned a possible `resolve` method that we could use if we wanted to handle more than just plain numerics.

module/documents/applied-rules.mjs
```diff
@@ -175,6 +181,45 @@ class RulesIterator extends Iterator {
...
+  /**
+   * Find the minimum and maximum limit among all of the provided rules.
+   * @param {Partial<RollRange>} [range]  Initial range to fine tune.
+   * @returns {RollRange}
+   */
+  toRange(range) {
+    return this.reduce((range, rule) => {
+      const value = Number(rule.value);
+      if ( !Number.isFinite(value) ) return range;
+      if ( (rule.type === "dnd5e.maximum") && (value < range.maximum) ) range.maximum = value;
```
Fyorl: Have these already had the `dnd5e` removed by this point? Should we be checking just for `maximum` not `dnd5e.maximum`? Or should we be checking for both?

module/config.mjs
```diff
@@ -3917,6 +3917,18 @@ DND5E.activeEffectChangeTypes = Object.freeze({
     defaultPriority: 100,
     handler: ActiveEffect5e._applyChangeRule,
     skipConditions: true
+  },
+  "dnd5e.maximum": {
```
Fyorl, line 3921: What do we do for the 'Roll Result' type rule that probably also wants `maximum`, `minimum`, `average`? Maybe these should be `floor` and `ceil`?

Fyorl (changes requested):
I am perhaps misunderstanding but it seems we are still not using `toLargest` and `toSmallest` when I thought they would be used for this particular rule.

We build an iterator that mixes both `:maximum` and `:minimum` in the same stream in the `if ( key === "range" )` branch of `#collect`, and then we have to split them out again in `toRange`.

Should `toRange` instead be a static helper that uses `toLargest` and `toSmallest`?

```js
static toRange(category, actor, item, rollData, range) {
  return {
    maximum: AppliedRules.collect(`${category}:maximum`, actor, item).filterWith(rollData).toSmallest(range?.maximum),
    minimum: AppliedRules.collect(`${category}:minimum`, actor, item).filterWith(rollData).toLargest(range?.minimum)
  };
}
```

Then the current places we call them are mostly the same except we're using the static helper rather than the method:

```js
const { maximum, minimum } = this.actor ? AppliedRules.toRange("attack", this.actor, this.item, rollData) : {};
```

It means we could remove the branch in `#collect` and all the APIs don't need to worry about mixed streams ever.

I get we want the same shape as everything else with `AppliedRules.collect().filterWith().xyz` but those are all homogenous streams, so it seems to me like maybe `toRange` is conceptually different to them.

module/config.mjs
arbron: I think roll result would just be a single rule type and its values would be `maximize`, `minimize`, and `average`.

module/documents/activity/attack.mjs
```diff
@@ -259,7 +259,15 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
...
-    const options = config.options ?? {};
+    const options = {
+      ...(config.options ?? {}),
+      maximum: this.actor
+        ? AppliedRules.collect("attack:maximum", this.actor, this.item).filterWith(data).resolve(data).toSmallest()
+        : undefined,
+      minimum: this.actor
+        ? AppliedRules.collect("attack:minimum", this.actor, this.item).filterWith(data).resolve(data).toLargest()
+        : undefined,
+    };
```
Fyorl: Should probably merge `config.options` over this, like we do in `getInitiativeRoll`. That way we don't clobber caller-supplied `options.minimum/maximum`.


```suggestion
    const options = foundry.utils.mergeObject({
      maximum: this.actor
        ? AppliedRules.collect("attack:maximum", this.actor, this.item).filterWith(data).resolve(data).toSmallest()
        : undefined,
      minimum: this.actor
        ? AppliedRules.collect("attack:minimum", this.actor, this.item).filterWith(data).resolve(data).toLargest()
        : undefined
    }, config.options ?? {});
```

module/documents/actor/actor.mjs
```diff
@@ -1458,6 +1454,15 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
 
+    config.options = {
+      ...(config.options ?? {}),
+      maximum: AppliedRules.collect("check:maximum", this).filterWith(rollData).resolve(rollData).toSmallest(
+        Math.min(relevant?.roll.max ?? Infinity, ability?.check.roll.max ?? Infinity)
+      ),
+      minimum: AppliedRules.collect("check:minimum", this).filterWith(rollData).resolve(rollData).toLargest(
+        Math.max(relevant?.roll.min ?? -Infinity, ability?.check.roll.min ?? -Infinity)
+      )
+    };
```
```suggestion
    config.options = foundry.utils.mergeObject{
      maximum: AppliedRules.collect("check:maximum", this).filterWith(rollData).resolve(rollData).toSmallest(
        Math.min(relevant?.roll.max ?? Infinity, ability?.check.roll.max ?? Infinity)
      ),
      minimum: AppliedRules.collect("check:minimum", this).filterWith(rollData).resolve(rollData).toLargest(
        Math.max(relevant?.roll.min ?? -Infinity, ability?.check.roll.min ?? -Infinity)
      )
    }, config.options ?? {});
```

module/documents/actor/actor.mjs
Fyorl: unused

module/documents/applied-rules.mjs
```diff
@@ -175,10 +177,43 @@ class RulesIterator extends Iterator {
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Use `simplifyBonus` to deterministically resolve all values.
+   * @param {object} data  Data to use for replacing @ strings.
+   * @returns {RulesIterator<number>}
+   */
+  resolve(data) {
+    return new RulesIterator(this.values(String).map(v => simplifyBonus(v, data)));
```
Fyorl: At least for when this is getting piped into `toSmallest` or `toLargest`, I think we need to drop 0s, otherwise a non-deterministic term or erroneous formula will floor all legitimate values in a `toSmallest` chain. I can't think of a case where 0 would be valid with regards to d20 `minimum` and `maximum` anyway. And for some other stream of bonuses, dropping 0 is harmless.


```suggestion
    return new RulesIterator(this.values(String).map(v => simplifyBonus(v, data)).filter(_ => _));
```

module/documents/actor/actor.mjs
```diff
@@ -1575,8 +1580,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     ], AppliedRules.collect(`${type}:advantage`, this).filterWith(rollData).toAdvantageCounts());
     const options = {
       advantage, disadvantage,
-      maximum: ability?.[type]?.roll.max,
-      minimum: ability?.[type]?.roll.min
+      maximum: AppliedRules.collect(`${type}:maximum`, this).filterWith(rollData).resolve(rollData).toSmallest(
```
Fyorl, line 1581: I guess with #6145 this all gets rolled into `combineFields` and we have a single call that produces `{ advantage, disadvantage, minimum, maximum }`?

Fyorl (changes requested):

module/documents/actor/actor.mjs
arbron: That is the intention.

Fyorl (approved):

PR #7258, opened 2026-07-16
Coerce strings that look like bools or numbers into booleans or numbers when applying unguided AEs

krbz999:
Currently, using `flags.world.foo | override | true` results in `Actor#flags.world.foo === "true"` and similar applies when using numeric-like values.

~~This changes it such that if the change is unguided (not targeting a datafield) but target `flags`, it is attempted to be coerced into a boolean or number.~~

This helps with the new conditionals as currently something like this is not possible:

```
flags.world.foo | Override | true
```
and a different change with
```
k: "flags.world.foo",
o: "exact",
v: true
```

(Note: in conjunction with #7257)

e: The second commit reduces this to simply parsing effect values when submitting changes.

### DIFF
--- a/module/applications/active-effect/effect-change-config.mjs
+++ b/module/applications/active-effect/effect-change-config.mjs
@@ -156,6 +156,9 @@ export default class EffectChangeConfig extends DocumentSheet5e {
     const change = changes.find(c => c._id === this.options.changeId);
     if ( !change ) return;
     foundry.utils.mergeObject(change, foundry.utils.expandObject(formData.object));
+    try {
+      if (typeof change.value === "string") change.value &&= JSON.parse(change.value);
+    } catch{}
     this.effect.update({ "system.changes": changes });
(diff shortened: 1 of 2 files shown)


stwlam:
Note that the underlying issue here is the bypassing of an operation in the core AE sheet:

```js
  /**
   * Process submission data for a single change object.
   * @param {EffectChangeData} change The submitted change object with the value deserialized
   * @param {number} index            The object's index in the submitted array
   * @protected
   */
  _processChangeSubmission(change, index) {
    try {
      if ( typeof change.value === "string" ) change.value &&= JSON.parse(change.value);
    }
    catch{}
  }
```

It's arguably preferable that the value be stored as its intended primitive type rather than it being coerced during every data preparation cycle.

krbz999:
Seems more like a bug then. And probably an easy addition to `EffectChangeConfig`.

PR #7260, opened 2026-07-16
[#7259] Support typed custom flags for Active Effect conditions

### LINKED ISSUE #7259: Support typed custom flags for Active Effect conditions
Based on a recent discord discussion, it would be useful to have a simple way to make user-defined flags work reliably with Active Effect conditions.

AE change values come from form fields, so custom `flags.world.*` values do not have a known type. This can make strict checks awkward when a flag is set with `Actor#setFlag` as a Number or Boolean, but changed through an Active Effect as text.

One option would be a typed dnd5e namespace: 
- `flags.dnd5e.conditional.number.<id>`
- `flags.dnd5e.conditional.boolean.<id>`
- `flags.dnd5e.conditional.string.<id>`

The final part would remain user-defined. The type in the path would tell dnd5e how to handle the Active Effect value.

thatlonelybugbear:
Work in progress!
I was thinking of something along these lines

Closes #7259


krbz999:
If I understand this proposal right, it requires using flags.dnd5e. But then you may as well register a new character flag in which case it is already cast to the right type.

thatlonelybugbear:
> If I understand this proposal right, it requires using flags.dnd5e. But then you may as well register a new character flag in which case it is already cast to the right type.

Yep, that makes sense for flags we already know about.

The only reason for this would be when the final ID is user-defined, so there is nothing concrete to register in characterFlags ahead of time

PR #7262, opened 2026-07-16
Support transient modifiers for damage rolls

thatlonelybugbear:
Adds a `modifiers` option to individual damage-roll configs.

Each entry in `config.rolls` can have its own modifiers.

Modules can set modifiers during `dnd5e.preRollDamage` without rewriting formulas:

```js
config.rolls[index].options.modifiers = ["adv"];
config.rolls[index].options.modifiers = ["dis", "min2"];
```

The modifiers are applied to the roll's dice terms before damage critical handling. Existing formula modifiers are preserved, and duplicate tokens are avoided.

Critical bonus damage also receives the roll-level modifiers, without being critically scaled.

This works independently for modules modifying config.rolls.

#7105 would additionally allow persisted DamageData.options to flow into generated roll configs.


thatlonelybugbear:
This would provide the per-roll mechanism needed for a potential core solution for features like Elemental Adept.

"When you roll damage of the chosen type, treat any 1 on a damage die as a 2."

It would need a system rule that selects matching damage types.

PR #7264, opened 2026-07-17
Implement falling automation

Fyorl:
Forgot I hadn't actually raised this PR. Let me know if you prefer it to be split up, though the commits should be fairly straightforward to follow.

- Adds `dnd5e.rules` helpers for calculating fall damage and showing a fall damage chat card.
- Adds the 'fall' movement action.
- Adds surface detection (check the last commit here since I made some amendments to this at the end).
- Adds `TokenDocument5e#plummet` API for sending a token to its closest resting surface & level and rolling appropriate fall damage.
- Adds interactivity to the falling status effect icon, allowing DMs to plummet tokens.
- Adds landing VFX
- Implements elevation scaling for tokens.

https://github.com/user-attachments/assets/c8f71d73-4c62-4a1b-9f18-3253c05a274b

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -70,6 +70,7 @@ Hooks.once("init", function() {
   CONFIG.Actor.documentClass = documents.Actor5e;
   CONFIG.Adventure.documentClass = documents.Adventure5e;
   CONFIG.Canvas.layers.tokens.layerClass = canvas.layers.TokenLayer5e;
+  CONFIG.Canvas.vfx.enabled = true;
   CONFIG.ChatMessage.documentClass = documents.ChatMessage5e;
   CONFIG.Combat.documentClass = documents.Combat5e;
   CONFIG.Combatant.documentClass = documents.Combatant5e;
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -28,6 +28,7 @@ import * as inserts from "./module/inserts.mjs";
 import * as migrations from "./module/migration.mjs";
 import { registerModuleData, registerModuleRedirects, setupModulePacks } from "./module/module-registration.mjs";
 import { default as registry } from "./module/registry.mjs";
+import * as rules from "./module/rules/_module.mjs";
 import Tooltips5e from "./module/tooltips.mjs";
 import * as utils from "./module/utils.mjs";
 import DragDrop5e from "./module/drag-drop.mjs";
@@ -48,6 +49,7 @@ globalThis.dnd5e = {
   inserts,
   migrations,
   registry,
+  rules,
   ui: {},
   utils
 };
--- a/module/canvas/_module.mjs
+++ b/module/canvas/_module.mjs
@@ -1,7 +1,7 @@
 export {default as BasePlacement} from "./api/base-placement.mjs";
 export * as detectionModes from "./detection-modes/_module.mjs";
 export * as layers from "./layers/_module.mjs";
-export {playLandingVfx} from "./landing-vfx.mjs";
+export * as vfx from "./vfx/_module.mjs";
 export {default as MapLocationControlIcon} from "./map-location-control-icon.mjs";
 export {default as Note5e} from "./note.mjs";
 export {default as TemplatePlacement} from "./template-placement.mjs";
--- a/module/canvas/_module.mjs
+++ b/module/canvas/_module.mjs
@@ -1,6 +1,7 @@
 export {default as BasePlacement} from "./api/base-placement.mjs";
 export * as detectionModes from "./detection-modes/_module.mjs";
 export * as layers from "./layers/_module.mjs";
+export {playLandingVfx} from "./landing-vfx.mjs";
 export {default as MapLocationControlIcon} from "./map-location-control-icon.mjs";
 export {default as Note5e} from "./note.mjs";
 export {default as TemplatePlacement} from "./template-placement.mjs";
--- a/module/canvas/token.mjs
+++ b/module/canvas/token.mjs
@@ -3,6 +3,15 @@
  */
 export default class Token5e extends foundry.canvas.placeables.Token {
 
+  /** @inheritDoc */
+  static RENDER_FLAGS = {
+    ...super.RENDER_FLAGS,
+    // Elevation changes must also rescale the mesh, as tokens are scaled by their distance from the viewed level.
+    refreshElevation: { propagate: ["refreshTooltip", "refreshMesh"] }
+  };
+
+  /* -------------------------------------------- */
+
   /**
    * Update the token ring when this token is targeted.
    * @param {User5e} user         The user whose targeting has changed.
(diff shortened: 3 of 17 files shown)


dnd5e.mjs
```diff
@@ -67,6 +69,7 @@ Hooks.once("init", function() {
   CONFIG.Actor.documentClass = documents.Actor5e;
   CONFIG.Adventure.documentClass = documents.Adventure5e;
   CONFIG.Canvas.layers.tokens.layerClass = canvas.layers.TokenLayer5e;
+  CONFIG.Canvas.vfx.enabled = true;
```
krbz999, line 73: Should the warning about the VFX system be suppressed?

module/documents/token.mjs
```diff
@@ -242,6 +258,157 @@ export default class TokenDocument5e extends SystemFlagsMixin(TokenDocument) {
...
+
+  /**
+   * Determine whether this token should be considered falling after its movement. Movement that ends with a fly or
+   * burrow action does not leave the token falling, regardless of elevation.
+   * @param {TokenMovementOperation} movement  The concluded movement.
+   * @returns {boolean}
+   */
+  #shouldFall(movement) {
+    const { action } = movement.passed.waypoints.at(-1) ?? {};
+    if ( (action === "fly") || (action === "burrow") ) return false;
```
arbron: What about climbing? If someone is climbing up a wall during their turn and only gets halfway up this seems like it would cause them to fall.

Might also want to handle flying creatures that get the Prone condition and don't have Hover.

module/canvas/token.mjs
```diff
@@ -146,6 +155,23 @@ export default class Token5e extends foundry.canvas.placeables.Token {
 
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _drawEffect(src, tint) {
+    const icon = await super._drawEffect(src, tint);
+    if ( !game.user.isGM ) return icon;
+    if ( icon && src && (src === CONFIG.statusEffects.falling?.img) ) {
```
arbron, line 162: It would be good to have some indication that the status effect is interactable. Since the whole token uses the pointer cursor when hovered, there is nothing to indicate that the status effect icon specifically can be clicked on.

module/canvas/_module.mjs
```diff
@@ -1,6 +1,7 @@
 export {default as BasePlacement} from "./api/base-placement.mjs";
 export * as detectionModes from "./detection-modes/_module.mjs";
 export * as layers from "./layers/_module.mjs";
+export {playLandingVfx} from "./landing-vfx.mjs";
```
arbron: Don't know how many of these we will add, but it may be good to stick them under `dnd5e.canvas.vfx` folder from the get go in case we do more.

arbron (changes requested):
Seemed to work, though in building my own test world I found the requirement that lower-level surfaces must also belong to the higher levels in order for the falling to register a bit counter-intuitive.

Found significant bug when testing in the Restored Keep map. In the chapel, when climbing down the ladder from the balcony to the first floor, it appears to hit some sort of issue. The status effect is applied but the icon is briefly rendering giant and the whole game stutters. There are no lasting effects, but this is in the console:

Image

Fyorl:
> Seemed to work, though in building my own test world I found the requirement that lower-level surfaces must also belong to the higher levels in order for the falling to register a bit counter-intuitive.

That's not really related to this PR though. Surfaces that affect a level must be registered to that level, otherwise they are invisible to it. If you have two levels separated by a movement-blocking surface, unless that surface is registered in both levels, it won't block any movement between them (or it will block in only one direction, depending on which of the two it is registered to).

module/documents/token.mjs
Fyorl: Just to be clear: All it will do is apply the falling status. If you're not actually falling, you can ignore it (try to not misclick it though). But extending this to the climb movement type should be no problem.

module/canvas/token.mjs
```diff
@@ -146,6 +155,23 @@ export default class Token5e extends foundry.canvas.placeables.Token {
 
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
+  async _drawEffect(src, tint) {
+    const icon = await super._drawEffect(src, tint);
+    if ( !game.user.isGM ) return icon;
```
roth-michael, line 161: I think there's value in having the icon interaction not be GM-only. You've already got a permissions check inside `plummet`. Could be fun as a player to send your token falling ~~to their death~~.

module/canvas/_module.mjs
roth-michael: Somewhat related: Though this is our only VFX currently, I think we should incorporate a system-level VFX toggle. One may want the fall without the animation (or whatever future automation we add without whatever associated future animations).

module/canvas/token.mjs
Fyorl: Typically players can only change level via the change level region so I'm not sure about circumventing that.

module/canvas/token.mjs
Fyorl: Open to suggestions. In the absence of that I am content with this being something that is learned once.

dnd5e.mjs
Fyorl: Don't think there's a way to suppress it

module/canvas/token.mjs
arbron: Maybe just change colors when the status is hovered over? I suppose since this is canvas we don't have access to the icon's fill variable, but since it is our icon we could tint it reliably.

module/canvas/token.mjs
roth-michael: Imho this is one place where I think it would make sense to deviate from core's handling. Then again I'm also the one who logged https://github.com/foundryvtt/foundryvtt/issues/14276 so I also think it makes sense for _core_ to deviate from core's handling. But enabling player use of plummeting doesn't immediately seem to present any opportunity for shenanigans. The only time it could reveal new information is if the level they're in does not have the level they'd fall to as being visible, which is a situation I can only associate with the "different plane" use case of levels; which actually I've just realize probably needs to be coded against here.

If a candidate level is _not_ in token's current level's `visibility.levels`, it should probably be stricken from the list of candidates; or at least deprioritized compared to another candidate level which _is_ in the current level's `visibility.levels`. In the case where one has, for instance, a 2-story building, in which each store has both a "material" and "ethereal" version represented by different levels, one should not be able to fall from the 2nd story material plane into the 1st story ethereal plane. And I suppose the reverse is true as well; that 2nd story ethereal should not fall into 1st story material. Which, if the material levels are marked visible from the ethereal levels, might mean considering bi-directional visibility (prioritize a level which you can see and which can see you vs a level which you can see but cannot see you).

Went on a bit of a tangent there; relevant to implementation, but not to my main point here which was: I still think it would make sense to allow players to plummet their own tokens.

module/documents/token.mjs
Fyorl: Climbing now exempt

module/canvas/_module.mjs
Fyorl: I have it respecting the client's `prefers-reduced-motion: reduce` setting now

module/documents/token.mjs
Fyorl: Also implemented the flying while prone handling. It made me realise that using the special statuses doesn't make sense. As far as the rules are concerned, so long as you have a fly speed, you can fly. So now we just check for fly speed and check for hover when prone/incap come into play.

arbron (changes requested):
Everything is good except for a race condition with effects. If you apply the Unconscious effect then you get a `The _id [dnd5efalling0000] already exists within the parent collection: Actor [AkraLv0100000000] effects` error because it has the Prone condition as a rider.

module/canvas/token.mjs
Fyorl: Works for me, added a foundry orange tint

Fyorl:
> Everything is good except for a race condition with effects. If you apply the Unconscious effect then you get a `The _id [dnd5efalling0000] already exists within the parent collection: Actor [AkraLv0100000000] effects` error because it has the Prone condition as a rider.

A bit involved but I pulled in your `modifyBatch` branch, and then moved everything to `_onCreateOperation` so that all riders could be resolved in one batch and be awaited, and that should have resolved it now.

arbron (approved):
Looking good to me!

PR #7265, opened 2026-07-17
Provide specific casting token data to regions and use it for disposition resolution.

Fyorl:
As mentioned on the PR, what we have already is probably correct in most cases. I think having this data available on the region is generally useful though, and we have one concrete use here to handle edge cases where linked tokens have a temporarily different disposition to their proto token.

### DIFF
--- a/module/canvas/template-placement.mjs
+++ b/module/canvas/template-placement.mjs
@@ -150,7 +150,7 @@ export default class TemplatePlacement extends BasePlacement {
       highlightMode: "coverage",
       flags: {
         dnd5e: {
-          caster: activity.getUsageToken()?.uuid,
+          activity: activity.uuid,
           dimensions: {
             size: templateData.size,
             width: templateData.width,
@@ -158,7 +158,7 @@ export default class TemplatePlacement extends BasePlacement {
             units: canvas.scene.grid.units
           },
           item: activity.item.uuid,
-          origin: activity.uuid,
+          origin: activity.getUsageToken()?.uuid,
           spellLevel: rollData.item.level
         }
       }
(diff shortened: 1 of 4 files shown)


module/canvas/template-placement.mjs
```diff
@@ -150,6 +150,7 @@ export default class TemplatePlacement extends BasePlacement {
       highlightMode: "coverage",
       flags: {
         dnd5e: {
+          caster: activity.getUsageToken()?.uuid,
```
arbron: I wonder if `token` may be better, since it doesn't imply a spell origin.

module/documents/activity/mixin.mjs
```diff
@@ -1182,9 +1182,10 @@ export default function ActivityMixin(Base) {
       const behaviors = activity?.applicableBehaviors;
       if ( !behaviors?.length ) return;
 
+      const caster = await fromUuid(region.getFlag("dnd5e", "caster"));
```
arbron: Since this is a token it doesn't seem like there is a chance this would be a compendium document so `fromUuidSync` should be safe.

arbron (changes requested):

module/canvas/template-placement.mjs
Fyorl: I didn't want it to be confused with the attached `token`. I am ok with changing it though.

arbron: Valid, `originToken`?

Fyorl: Well, we have `origin` already. Ideally that might be the token and then the activity goes under `activity` to match `item`. Does `origin` need to point to the activity for back-compat or is it something we added in 6.0 with the new template placement?

arbron: It has been around for longer, but since we are changing these from `AbilityTemplate` to `Region`, I'm not sure if it is necessary to keep the flags exactly the same since consumers will have to be swapping out hooks and other handling anyway.

Fyorl: Yeah good point. You're getting an entirely different document now. I'll make the change.

arbron (approved):

PR #7267, opened 2026-07-18
Add expiry events to `DurationData`, apply duration to enchantments

arbron:
Adds `duration.expiry` to activity and spell data. Does not hide the duration field for duration-less expiry events because you may need to keep the duration data (for example, Shield has a duration of 1 Round and the expiry event of "Start of Source's Next Turn").

Changes how duration overrides are handled for applied effects by moving the logic into the activity itself, to give activities a chance to fully modify the effect being applied. In the process this also applies those same changes to applied enchantments.

### DIFF
--- a/module/applications/active-effect/active-effect-sheet.mjs
+++ b/module/applications/active-effect/active-effect-sheet.mjs
@@ -113,18 +113,9 @@ export default class ActiveEffectSheet5e extends ApplicationV2Mixin(ActiveEffect
    * @protected
    */
   async _prepareDurationContext(context, options) {
-    const general = _loc("DND5E.ACTIVEEFFECT.Expiry.GROUPS.General");
-    const specific = _loc("DND5E.ACTIVEEFFECT.Expiry.GROUPS.Specific");
-
-    for ( const [expiry, label] of Object.entries(context.expiryEvents) ) {
-      context.expiryEvents[expiry] = { group: general, label };
-    }
-    for ( const expiry of dnd5e.documents.ActiveEffect5e.PSEUDO_EXPIRIES ) {
-      context.expiryEvents[expiry] = {
-        group: specific,
-        label: _loc(`DND5E.ACTIVEEFFECT.Expiry.${expiry.capitalize()}`)
-      };
-    }
+    context.expiryEvents = Object.fromEntries(
+      ActiveEffect.implementation.expiryOptions.map(({ value, ...data }) => [value, data])
+    );
     return context;
   }
 
--- a/module/applications/components/effect-application.mjs
+++ b/module/applications/components/effect-application.mjs
@@ -214,8 +214,7 @@ export default class EffectApplicationElement extends ChatTrayElement {
     }
 
     // Get any effect changes provided by the activity & add flags
-    const changes = this.chatMessage.getAssociatedActivity({ scaled: true })
-      ?.getAppliedEffectChanges(effect, { chatMessage: this.chatMessage, target: actor }) ?? {};
+    const changes = activity?.getAppliedEffectChanges(effect, { chatMessage: this.chatMessage, target: actor }) ?? {};
     foundry.utils.mergeObject(changes, {
       flags: {
         dnd5e: {
--- a/module/applications/components/effect-application.mjs
+++ b/module/applications/components/effect-application.mjs
@@ -213,7 +213,10 @@ export default class EffectApplicationElement extends ChatTrayElement {
       throw new Error(_loc("DND5E.EFFECT.Application.Warning.Ownership"));
     }
 
-    const effectFlags = {
+    // Get any effect changes provided by the activity & add flags
+    const changes = this.chatMessage.getAssociatedActivity({ scaled: true })
+      ?.getAppliedEffectChanges(effect, { chatMessage: this.chatMessage, target: actor }) ?? {};
+    foundry.utils.mergeObject(changes, {
       flags: {
         dnd5e: {
           dependentOn: concentration?.uuid,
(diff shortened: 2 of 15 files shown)


module/data/shared/duration-field.mjs
```diff
@@ -56,21 +57,23 @@ export default class DurationField extends SchemaField {
...
-      case "day": return { value, units: "days" };
-      case "month": return { value, units: "months" };
-      case "year": return { value, units: "years" };
-      default: return {};
+      case "turn": units = "turns"; break;
+      case "round": units = "rounds"; break;
+      case "second": units = "seconds"; break;
+      case "minute": units = "minutes"; break;
+      case "hour": units = "hours"; break;
+      case "day": units = "days" ; break;
```
Fyorl:
```suggestion
      case "day": units = "days"; break;
```

module/documents/active-effect.mjs
```diff
@@ -120,6 +120,21 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
 
+  /**
+   * Possible expiry options.
+   * @type {FormSelectOption[]}
+   */
+  static get expiryOptions() {
+    const general = _loc("DND5E.EFFECT.Expiry.Group.General");
+    const specific = _loc("DND5E.EFFECT.Expiry.Group.Specific");
+    return [
+      ...Object.entries(ActiveEffect5e.EXPIRY_EVENTS).map(([value, l]) => ({ value, label: _loc(l), group: general })),
```
```suggestion
      ...Object.entries(this.EXPIRY_EVENTS).map(([value, l]) => ({ value, label: _loc(l), group: general })),
```

module/documents/active-effect.mjs
```diff
@@ -1118,13 +1134,24 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+  /* -------------------------------------------- */
+
   /**
    * Determine whether the provided expiry should be able to display duration fields.
    * @param {string} [expiry]  Active effect expiry event to check, defaults to this effect's current expiry.
    * @returns {boolean}
    */
   expirySupportsDuration(expiry=this.duration.expiry) {
-    return !this.constructor.PSEUDO_EXPIRIES.has(expiry) && !this.constructor.DURATIONLESS_EXPIRIES.has(expiry);
+    return ActiveEffect5e.expirySupportsDuration(expiry);
```
```suggestion
    return this.constructor.expirySupportsDuration(expiry);
```

module/applications/active-effect/active-effect-sheet.mjs
```diff
@@ -113,18 +113,9 @@ export default class ActiveEffectSheet5e extends ApplicationV2Mixin(ActiveEffect
...
-      context.expiryEvents[expiry] = { group: general, label };
-    }
-    for ( const expiry of dnd5e.documents.ActiveEffect5e.PSEUDO_EXPIRIES ) {
-      context.expiryEvents[expiry] = {
-        group: specific,
-        label: _loc(`DND5E.ACTIVEEFFECT.Expiry.${expiry.capitalize()}`)
-      };
-    }
+    context.expiryEvents = Object.fromEntries(
+      ActiveEffect.implementation.expiryOptions.map(({ value, ...data }) => [value, data])
```
Fyorl, line 117: Do these need to be alphabetised?

module/config.mjs
```diff
@@ -4723,10 +4723,25 @@ DND5E.calendarDeltasRecoveryMapping = new Map([
...
+  longRest: "DND5E.EFFECT.Expiry.Type.LongRest",
+  shortRest: "DND5E.EFFECT.Expiry.Type.ShortRest"
 });
 
+/* -------------------------------------------- */
+
+/**
+ * System-specific "expiry" choices which do not require registration or custom expiry events, and instead
+ * are handled dynamically in `ActiveEffect#isExpiryEvent`.
+ * @type {Record<string>}
```
```suggestion
 * @type {Record<string, string>}
```

Fyorl (changes requested):
This hint reads a bit weirdly for spell details. Maybe just no hint in this case to match the other fields.

Image

module/applications/active-effect/active-effect-sheet.mjs
arbron: I probably like them better unsorted, because then they follow a sort of logical order (rest, combat/round/turn start, combat/round/turn end, start/end source's turn, start/end target's turn):

Screenshot 2026-09-02 at 14 08 57
Screenshot 2026-09-02 at 14 08 36

We could even do a little tweak to the order of our special expiry events to match the ordering of the combat ones (e.g. both start events first, then both end events):

Screenshot 2026-09-02 at 14 11 36

module/applications/components/effect-application.mjs
```diff
@@ -213,7 +213,10 @@ export default class EffectApplicationElement extends ChatTrayElement {
       throw new Error(_loc("DND5E.EFFECT.Application.Warning.Ownership"));
     }
 
-    const effectFlags = {
+    // Get any effect changes provided by the activity & add flags
+    const changes = this.chatMessage.getAssociatedActivity({ scaled: true })
+      ?.getAppliedEffectChanges(effect, { chatMessage: this.chatMessage, target: actor }) ?? {};
```
Fyorl: Missed this earlier but we already have `this.chatMessage.getAssociatedActivity({ scaled: true })` available.

```suggestion
    const changes = activity?.getAppliedEffectChanges(effect, { chatMessage: this.chatMessage, target: actor }) ?? {};
```

Fyorl (approved):
LGTM. If you would like to take care of applying `turnStart` expiry at point of application too, this would be a good place for that change.

PR #7268, opened 2026-07-20
Convert Hit Points, Hit Die, and Utility roll messages into bespoke system sub-types

Fyorl:
Starting the process of converting our chat cards into sub-types that render from data. I started with the easy ones first, the ones that only render their rolls (for now), and also adjusted our chat card migration to migrate in batches since on worlds with thousands of messages this could be a real bottleneck.

### DIFF
--- a/module/migration.mjs
+++ b/module/migration.mjs
@@ -1,6 +1,14 @@
 import { formatIdentifier, log } from "./utils.mjs";
 
 /**
+ * Number of chat messages to update per batch during world migration.
+ * @type {number}
+ */
+const MESSAGE_MIGRATION_BATCH_SIZE = 64;
+
+/* -------------------------------------------- */
+
+/**
  * Perform a system migration for the entire World, applying migrations for Actors, Items, and Compendium packs.
  * @param {object} [options={}]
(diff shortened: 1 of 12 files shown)


module/migration.mjs
```diff
@@ -106,19 +114,27 @@ export async function migrateWorld({ bypassVersionCheck=false }={}) {
     incrementProgress();
   }
 
-  // Migrate World Messages
+  // Migrate World Messages in batches
+  let messageBatch = [];
+  const flushMessageBatch = async () => {
+    if ( !messageBatch.length ) return;
+    await ChatMessage.implementation.updateDocuments(messageBatch, { enforceTypes: false, render: false });
```
arbron, line 121: Still getting an error when testing this migration:

Image

Fyorl: Blocked on foundryvtt/foundryvtt#14229 it seems

Fyorl: Commented-out these migrations for now. There's no workaround that I can think of and I need to move on with these card type conversions. Since these are very simple cards, the only issue I can find with not migrating them is that some of them will have apply damage/apply healing context menu entries when they shouldn't. But only on old cards of course.

arbron (approved):

PR #7270, opened 2026-07-21
[#7269] Support Total Cover as an explicit attack outcome

### LINKED ISSUE #7269: [Feature Request] Support Total Cover as an explicit attack outcome
Add first-class Total Cover support to D&D5e attack targeting.

Currently, Total Cover is represented by a target with `ac: null`, displayed as infinity AC. This does not represent the D&D 5e rule correctly: normal attacks can be classified as hits through JavaScript coercion, and critical successes always bypass the AC comparison.

thatlonelybugbear:
Closes #7269 

Adds explicit Total Cover handling for attack targets.
- [x] Derived configurable cover levels from active cover statuses.
- [x] Preserves a target’s calculated AC while identifying Total Cover separately.
- [x] Treats Total Cover as an automatic miss, including for expanded critical-success ranges.
- [x] Adds `ignoreTotalCover` for module / macro overrides.
- [ ] Expose `ignoreTotalCover` as a trait or item/activity Booleans

image

```js
Hooks.on("dnd5e.preRollAttack", (config) => {
    config.ignoreTotalCover = true;
});
  
//or
await activity.rollAttack({ ignoreTotalCover: true });
 ```
 
image


Kapuzenjoe:
Curious why the focus here is only on `ignoreTotalCover` rather than also covering Half and Three-Quarters Cover? Even though the system doesn't full automate cover in general yet, this might be a good spot to lay some groundwork for a model that could later be extended (e.g. spells like *Sacred Flame* ignore Half/Three-Quarters Cover but not Total Cover), so it might be worth considering a design where a single boolean isn't the only way to represent that distinction down the line.

https://github.com/issues/created?issue=foundryvtt%7Cdnd5e%7C6369

thatlonelybugbear:
I was mainly waiting on some feedback on how to proceed in general and what kind of automation the system wants to support.

I agree that there should be a more complete solution

PR #7277, opened 2026-07-22
Convert ability check rolls into bespoke system sub-types.

Fyorl:


module/data/chat-message/check-message-data.mjs
```diff
@@ -0,0 +1,35 @@
...
+export default class CheckMessageData extends RollMessageData {
+
+  /* -------------------------------------------- */
+  /*  Model Configuration                         */
+  /* -------------------------------------------- */
+
+  /** @override */
+  static defineSchema() {
+    return {
+      ability: new StringField({ blank: false, required: true }),
```
arbron, line 23: Skill & tool checks previously weren't storing the selected ability in the message data, so this results in a `ability: may not be undefined` error when migrating those messages.

We also don't strongly enforce that any actual ability is provided when you call `rollSkillTool`, so if you provide a tool or skill without a defined ability then the roll process will succeed without a selected ability up until the message creation, which will fail.

I'd probably say we support not including an ability with a check roll.

module/documents/actor/actor.mjs
```diff
@@ -1379,19 +1379,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-              type
-            }
-          }
-        },
         flavor: type === "skill"
           ? _loc("DND5E.SkillPromptTitle", { skill: skillConfig.label, ability: abilityLabel })
           : _loc("DND5E.ToolPromptTitle", { tool: Trait.keyLabel(config.tool, { trait: "tool" }) ?? "" }),
-        speaker: ChatMessage.getSpeaker({ actor: this })
+        speaker: ChatMessage.getSpeaker({ actor: this }),
+        system: { ability: abilityId, [type]: config[type] },
```
arbron: While we weren't including it in flags, we do now support providing both a skill and a tool to a roll to handle the 2024 rules that give you advantage if using a tool and skill together, so we should store both in the message.

arbron (changes requested):

module/data/chat-message/check-message-data.mjs
Fyorl: You can't make an ability check without an ability. I don't think that's something we should support. For the migration we can pick the default configured ability for the given skill or tool.

arbron (approved):

PR #7279, opened 2026-07-22
[#7080] Follow-up: Add ability-specific attack roll configuration

thatlonelybugbear:
Adds attack roll configuration to individual abilities, building on #6145 and the ability selection support from #7080 .

- Adds per-ability attack bonus, advantage mode, and minimum/maximum roll configuration.
- Adds global ability attack roll configuration.

image


thatlonelybugbear:
I also re-ordered the Attack / Check / Saves (https://github.com/foundryvtt/dnd5e/pull/7279/commits/1f60b85674f24f49925db7af74d68c22c41e00b4) because it was triggering me 😅 

image

thatlonelybugbear:
I think we could avoid some of the visual noise added by the repeating "ABILITY" in each section.
Maybe leave the "ABILITY SCORE" intact only?

image

module/data/actor/templates/creature.mjs
```diff
@@ -38,6 +38,7 @@ export default class CreatureTemplate extends CommonTemplate {
       }),
       rolls: new SchemaField({
         ability: new SchemaField({
+          attack: new D20RollModificationField({}, { labelPrefix: "DND5E.ROLL.FIELDS.rolls.ability.attack." }),
```
arbron: We already have `rolls.attack` as a modification field that applies to all attacks, so is there any realistic case where `roll.ability.attack.bonus` is used differently that `roll.attack.bonus`?

module/documents/activity/attack.mjs
```diff
@@ -121,9 +121,14 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
...
 
     const rollData = this.getRollData({ roll: { ability: rollConfig.ability, attackMode: rollConfig.attackMode } });
-    const { advantage, disadvantage } = this.actor ? D20RollModificationField.combineFields(this.actor.system, [
-      "rolls.attack", `rolls.attack.${this.getActionType(rollConfig.attackMode)}`
-    ], { rules: { category: "attack", actor: this.actor, item: this.item, rollData } }) : {};
+    const rollFields = [
+      ...(rollConfig.ability ? [`abilities.${rollConfig.ability}.attack.roll`] : []),
+      "rolls.ability.attack", "rolls.attack", `rolls.attack.${this.getActionType(rollConfig.attackMode)}`
+    ];
+    const { advantage, disadvantage, maximum, minimum } = this.actor
```
arbron: Shouldn't be handling minimum or maximum here, because that will just bake in the ability-based range so it won't change properly when you select a different abiltiy in the dialog.


```suggestion
    const { advantage, disadvantage } = this.actor
```

module/documents/activity/attack.mjs
```diff
@@ -132,7 +137,9 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
         ammunition: rollConfig.ammunition,
         attackMode: rollConfig.attackMode,
         criticalSuccess: this.criticalThreshold,
-        mastery: rollConfig.mastery
+        mastery: rollConfig.mastery,
+        maximum,
+        minimum
```
arbron:
```suggestion
        mastery: rollConfig.mastery
```

module/documents/activity/attack.mjs
```diff
@@ -262,13 +269,21 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
     const attackMode = formData?.get("attackMode") ?? process.attackMode;
     const mastery = formData?.get("mastery") ?? process.mastery;
 
+    process.ability = ability;
```
arbron: Don't think setting ability in the process here does anything. All of the code after the rolling process instead checks `rolls[0].options.ability`, so setting that should be sufficient.


```suggestion
```

module/documents/activity/attack.mjs
```diff
@@ -262,13 +269,21 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
...
     let { parts, data } = this.getAttackData({ ability, ammunition, attackMode });
-    const { maximum, minimum } = this.actor ? D20RollModificationField.combineFields(this.actor.system, [
-      "rolls.attack", `rolls.attack.${this.getActionType(attackMode)}`
-    ], { rules: { category: "attack", actor: this.actor, item: this.item, rollData: data } }) : {};
+    const rollFields = [
+      ...(ability ? [`abilities.${ability}.attack.roll`] : []),
+      "rolls.ability.attack", "rolls.attack", `rolls.attack.${this.getActionType(attackMode)}`
+    ];
+    const { advantage, disadvantage, maximum, minimum } = this.actor
```
arbron: Recalculating advantage and disadvantage here won't have any effect, since `options.advantageMode` is already set on the roll. Since we don't handle advantage changes in the dialog for any roll type at the moment and it would be a substantial amount of work to handle it, just take it out of here for now.


```suggestion
    const { maximum, minimum } = this.actor
```

module/documents/activity/attack.mjs
```diff
@@ -262,13 +269,21 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
...
+      ? D20RollModificationField.combineFields(this.actor.system, rollFields, {
+        rules: { category: "attack", actor: this.actor, item: this.item, rollData: data }
+      }) : {};
     const options = CONFIG.Dice.D20Roll.mergeOptions({
       elvenAccuracy: this.actor?.getFlag("dnd5e", "elvenAccuracy")
         && CONFIG.DND5E.characterFlags.elvenAccuracy.abilities.includes(ability),
+      advantage,
+      disadvantage,
```
arbron:
```suggestion
```

arbron (changes requested):

module/data/actor/templates/creature.mjs
thatlonelybugbear: Yeah I cannot think of a realistic case, so I will remove the Global Attack Ability Bonus field.
https://github.com/foundryvtt/dnd5e/pull/7279/commits/fd7cf1122ffdb006a0a60966c32afde11ace3605

thatlonelybugbear:
I also took the liberty of tightening the UI slightly with https://github.com/foundryvtt/dnd5e/pull/7279/commits/cca8fc1be3add3c7cbd9883b57c05934c9f155b2 (let me know to revert if it's not needed)
  
image

arbron (approved):

Fyorl (approved):

PR #7281, opened 2026-07-24
Convert death save rolls into bespoke system sub-type.

Fyorl:
Since death saves alter the actor, I've used `deltas` here to record the changes (so they can be appropriately reversed when a legendary resistance is used). Also removed the split message cards reporting death save terminal outcomes. Those are included in the main card now, meaning the hook signature has also changed slightly.

### DIFF
--- a/module/data/actor/templates/attributes.mjs
+++ b/module/data/actor/templates/attributes.mjs
@@ -14,7 +14,6 @@ const { NumberField, SchemaField, SetField, StringField } = foundry.data.fields;
 
 /**
  * @import { ActorRollData, DeathSaveOutcome } from "../../../documents/_types.mjs";
- * @import { ActorDeltasData } from "../../chat-message/fields/_types.mjs";
  * @import { ArmorClassData, AttributesCommonData, AttributesCreatureData, HitPointsData } from "./_types.mjs";
  */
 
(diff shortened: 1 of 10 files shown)


module/data/chat-message/save-message-data.mjs
```diff
@@ -64,7 +76,16 @@ export default class SaveMessageData extends RollMessageData {
...
+    const actor = this.parent.getAssociatedActor();
     Object.assign(context, { canResist, resisted });
+    context.death = this.ability === "death";
+    if ( actor ) {
+      context.deltas = ActorDeltasField.processDeltas.call(this.deltas, actor, this.parent.rolls);
+      context.deltasHeading = "DND5E.DeathSave";
+    }
+    if ( this.outcome ) context.outcome = game.i18n.format(`DND5E.DEATH.Outcome.${this.outcome}`, {
+      name: actor?.name ?? ""
+    });
```
krbz999:
```suggestion
    if ( this.outcome ) context.outcome = _loc(`DND5E.DEATH.Outcome.${this.outcome}`, {
      name: actor?.name ?? ""
    });
```

module/data/actor/templates/attributes.mjs
```diff
@@ -713,7 +714,62 @@ export default class AttributesFields {
...
+      const successes = (death.success || 0) + 1;
+
+      // Critical success - revive with 1 hp.
+      if ( isCritical ) {
+        Object.assign(updates, {
+          "system.attributes.death.success": 0,
+          "system.attributes.death.failure": 0,
+          "system.attributes.hp.value": 1
+        });
+        deltas.actor.push({ delta: 1, keyPath: "system.attributes.hp.value" });
```
arbron: We'd probably want to save the successes & failures changes in the deltas to get back to the original values if reversed (since we are saving deltas for successes on a normal success would probably be expected).

arbron (changes requested):

module/data/actor/templates/attributes.mjs
Fyorl: I opted not to do this because you end up with weird results where the card says +1 HP / -2 successes, which is technically what happened to the data, but not really expected. The stable outcome also reads weirdly if you do it like this. For example, you get +1 success, +1 success, and then the third success reads -2 successes.

It's necessary if you want to be able to reverse the result, but there's only one instance we actually want to do this, which is in the case of legendary resistance, and that works because we do not actually clear death saves at all at the death outcome.

We need to track how many failures were gained as part of the roll, and whether a legendary resistance was used. I co-opted the deltas framework for this since it seemed close enough but we could also just save it as system data separately and get rid of the deltas altogether.

arbron: We could filter out deltas that target successes & failures before rendering the card. That would allow us to take advantage of the existing infrastructure without adding data to the cards that isn't particularly helpful to players, instead only displaying the +1 HP on a critical success.

Fyorl: Sure, that can work. I think it's maybe nice to show +2 failures to make it clear the result of a critical failure if players aren't familiar but whether that's worth abusing deltas for and making them irreversible I'm not sure, so I have made the change.

arbron (approved):

PR #7282, opened 2026-07-24
Convert concentration save rolls into bespoke system sub-type.

Fyorl:


module/data/chat-message/save-message-data.mjs
```diff
@@ -23,8 +23,9 @@ export default class SaveMessageData extends RollMessageData {
     return {
       ability: new StringField({ blank: false, required: true }),
       deltas: new ActorDeltasField(),
+      isConcentration: new BooleanField(),
```
arbron: Not a huge preference either way, but how do you feel about using `type` instead (with values of `ability`, `death`, and `concentration`). The we could use `validateJoint` on `ability` so it is required unless the type is `death`.

module/data/chat-message/save-message-data.mjs
arbron, line 75: The actor can still resist after breaking concentration, but that doesn't restore concentration. Should probably disable resisting if concentration has already been broken.

module/migration.mjs
```diff
@@ -845,6 +845,17 @@ export function migrateMessageData(messageData) {
     updateData["flags.dnd5e.roll"] = _del;
   }
 
+  else if ( (rollType === "concentration") && (messageData.type === "base") ) {
```
arbron: Don't think we were setting the roll type before, they were just set to `save`. Not seeing any way to identify old concentration saves beyond checking the flavor for "Constitution Saving Throw".

arbron (changes requested):

module/migration.mjs
Fyorl: Right, we do have `roll.options.isConcentration` but I think on review I'd rather just delete this block and let legacy concentration fall through to the `save` block and become normal saving throws. Doesn't really lose anything unless you happened to be right in the middle of a combat when you update.

module/data/chat-message/save-message-data.mjs
Fyorl: I will see how it looks and if there are any issues with `validateJoint`

Fyorl: Not a lot of effort after all, and a cleaner design.

arbron (approved):

PR #7285, opened 2026-07-28
Convert roll requests into bespoke system sub-type.

Fyorl:
No migration for this one. Roll requests didn't write a `roll.type` flag and I didn't want a heuristic check, so I think these cards will just remain unmigrated and eventually their click handlers will stop being applied but I don't foresee that being a problem for these types of cards.

I also opted for `prompt` over `rollRequest` partly because I am biased against camelCase property names in general, and partly because we had co-opted this message type for the end concentration prompt, which is not a request for a roll. I figured it might be more generally useful to have a card that will accept extensible data for prompting for various game actions.

I also considered whether this should be unified with the `request` card in some way, since they can both request rolls, but the `request` card has specific targets rather being broadcast for anyone to optionally roll, and it is also stateful, unlike this card.

### DIFF
--- a/module/documents/actor/actor.mjs
+++ b/module/documents/actor/actor.mjs
@@ -8,7 +8,6 @@ import ActivationsField from "../../data/chat-message/fields/activations-field.m
 import { ActorDeltasField } from "../../data/chat-message/fields/deltas-field.mjs";
 import D20RollModificationField from "../../data/shared/d20-roll-modification-field.mjs";
 import TransformationSetting from "../../data/settings/transformation-setting.mjs";
-import { createRollLabel } from "../../enrichers.mjs";
 import {
   convertTime, defaultUnits, formatLength, formatNumber, formatTime, simplifyBonus, staticID
 } from "../../utils.mjs";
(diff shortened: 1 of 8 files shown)


module/documents/actor/actor.mjs
```diff
@@ -1166,23 +1148,11 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
...
-            dataset: { action: "endConcentration", actorUuid: this.uuid, visibility: "all" },
-            buttonLabel: label,
-            hiddenLabel: label
-          }]
-        }
-      ),
-      whisper: game.users.filter(user => this.testUserPermission(user, "OWNER")),
-      speaker: ChatMessage.implementation.getSpeaker({ actor: this })
+      speaker: ChatMessage.implementation.getSpeaker({ actor: this }),
+      system: { broadcast: false, buttons: [{ type: "endConcentration" }] },
```
Fyorl, line 1153: We don't need `actorUuid` anymore. We know that if `broadcast: false` then the prompt is for a specific actor and should always roll against that actor (or in this case, end that actor's concentration).

arbron (approved):

PR #7287, opened 2026-07-29
Convert healing & damage rolls to bespoke system sub-types.

Fyorl:


module/applications/components/damage-application.mjs
```diff
@@ -88,6 +90,13 @@ export default class DamageApplicationElement extends TargetedApplicationMixin(C
     this.chatMessage = game.messages.get(messageId);
     if ( !this.chatMessage ) return;
 
+    const rolls = this.chatMessage.rolls.filter(r => r instanceof DamageRoll);
+    this.damages = aggregateDamageRolls(rolls, { respectProperties: true }).map(roll => ({
+      properties: new Set(roll.options.properties ?? []),
+      type: roll.options.type,
+      value: Math.max(0, roll.total)
+    }));
```
Fyorl, line 98: Unchanged, just moved

module/dice/aggregate-damage-rolls.mjs
```diff
@@ -33,6 +43,58 @@ export default function aggregateDamageRolls(rolls, { respectProperties }={}) {
 
 /* -------------------------------------------- */
 
+/**
+ * Collect a roll's dice results and its deterministic remainder into a single breakdown.
+ * @param {Roll} roll  The roll to aggregate.
+ * @returns {DamageBreakdown}
+ */
+export function aggregateDamageTerms(roll) {
```
Fyorl, line 51: Unchanged, just moved

module/rules/falling.mjs
```diff
@@ -51,14 +51,15 @@ export async function postFallDamage(targets, distance) {
     data: {
       flags: {
         dnd5e: {
-          context: { fall: true }, messageType: "roll", roll: { type: "damage" },
+          context: { fall: true },
           targets: getTargetDescriptors(targets)
         }
       },
       flavor: game.i18n.format("DND5E.FALLING.DamageFlavor", {
```
krbz999:
```suggestion
      flavor: _loc("DND5E.FALLING.DamageFlavor", {
```

module/data/chat-message/healing-message-data.mjs
```diff
@@ -0,0 +1,29 @@
...
+ */
+export default class HealingMessageData extends DamageMessageData {
+
+  /* -------------------------------------------- */
+  /*  Properties                                  */
+  /* -------------------------------------------- */
+
+  /** @override */
+  get canApplyDamage() {
+    return false;
```
arbron: Previously we could apply damage from healing cards, we have special handling in `applyChatCardDamage` to handle that case.

module/documents/activity/mixin.mjs
```diff
@@ -898,14 +898,9 @@ export default function ActivityMixin(Base) {
...
           flavor: `${this.item.name} - ${this.damageFlavor}`,
-          flags: {
-            dnd5e: {
-              ...this.messageFlags,
-              messageType: "roll",
-              roll: { type: "damage" }
-            }
-          },
-          speaker: ChatMessage.getSpeaker({ actor: this.actor })
+          flags: { dnd5e: { ...this.messageFlags } },
```
arbron: Since `messageFlags` is created from scratch each time, skipping the spread should be safe.

```suggestion
          flags: { dnd5e: this.messageFlags },
```

templates/chat/damage-card.hbs
```diff
@@ -0,0 +1,31 @@
...
+        <div class="dice-formula">{{#if isPrivate}}???{{else}}{{ formula }}{{/if}}</div>
+        {{#unless isPrivate}}
+        {{> "dnd5e.damage-breakdown" }}
+        {{/unless}}
+        <h4 class="dice-total">{{#if isPrivate}}?{{else}}{{ total }}{{/if}}</h4>
+    </div>
+</div>
+
+{{#if onSave}}
+<p class="supplement"><strong>{{ localize "DND5E.SAVE.OnSave" }}</strong> {{ onSave }}</p>
```
arbron, line 26: Is this worth hiding for private rolls?

arbron (changes requested):

module/data/chat-message/healing-message-data.mjs
Fyorl: Maybe you are thinking this is used to gate the damage tray? The damage tray always appears. This getter is only used in `ChatMessage5e#canApplyDamage` which previously only checked `roll.type === "damage"`, so damage options should not have been available for healing messages which have had `roll.type: "healing"` for some time now. Maybe that was a mistake though.

Fyorl: Well I had a look at `applyChatCardDamage` and it seems like it should work, so I am thinking maybe the guard in `ChatMessage5e#canApplyDamage` was a mistake or it wasn't updated when we made `applyChatCardDamage`? I have removed the override for now.

arbron (approved):

PR #7302, opened 2026-08-03
[#6626] Move active effect origin into system data, split by type

### LINKED ISSUE #6626: Store AE origins on system data
### Initial Proposal
```js
class ActiveEffectData extends foundry.data.ActiveEffectTypeDataModel {
  static defineSchema() {
    return {
      origin: new foundry.data.fields.SchemaField({
        uuid: new foundry.data.fields.StringField({ nullable: true, blank: false })
      })
    };
  }
}
```
This will also allow us to extend the field in the future.

arbron:
Adds a new `system.origin` object that contains UUIDs for activity, actor, effect, and item origins. This allows storing more detailed information on origins and avoids relying on the `originText` flag when activities are the origin.

Closes #6626

### DIFF
--- a/module/applications/components/effect-application.mjs
+++ b/module/applications/components/effect-application.mjs
@@ -315,7 +315,7 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
     const concentration = originActor?.effects.get(this.chatMessage.system.concentration);
     const item = this.chatMessage.getAssociatedItem();
     const activity = this.chatMessage.getAssociatedActivity({ scaled: true });
-    const origin = concentration ?? (effect.inCompendium && item ? item : effect);
+    const sourceKey = effect.inCompendium ? "compendiumSource" : "duplicateSource";
     if ( !game.user.isGM && !actor.isOwner ) {
       throw new Error(_loc("DND5E.EFFECT.Application.Warning.Ownership"));
     }
@@ -345,9 +345,7 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
     }
 
     // Enable an existing effect on the target if it originated from this effect
-    const existingEffect = effect.inCompendium
-      ? actor.effects.find(e => e._stats.compendiumSource === effect.uuid)
-      : actor.effects.find(e => e.origin === origin.uuid);
+    const existingEffect = actor.effects.find(e => e._stats[sourceKey] === effect.uuid);
     if ( existingEffect ) {
       return { action: "update", data: foundry.utils.mergeObject({
         ...durationOverride,
@@ -364,7 +362,7 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
       disabled: false,
       transfer: false,
       _stats: {
-        [effect.inCompendium ? "compendiumSource" : "duplicateSource"]: effect.uuid,
+        [sourceKey]: effect.uuid,
         [effect.inCompendium ? "duplicateSource" : "compendiumSource"]: null
       }
     }, effectFlags);
--- a/module/applications/components/effect-application.mjs
+++ b/module/applications/components/effect-application.mjs
@@ -289,7 +289,7 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
   /**
    * Handle applying an Active Effect to a Token.
    * @param {ActiveEffect5e} effect      The effect to apply.
-   * @param {Actor5e} actor              The actor.
+   * @param {Actor5e} actor              The actor to which the effect should be applied.
    * @returns {Promise<ActiveEffect5e>}  The created effect.
    * @throws {Error}                     If the effect could not be applied.
    * @protected
@@ -327,6 +327,12 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
           scaling: this.chatMessage.system.scaling,
           spellLevel: this.chatMessage.system.level
         }
+      },
+      system: {
+        origin: {
+          [activity ? "activity" : "item"]: activity ? activity.uuid : item?.uuid,
+          effect: concentration?.uuid
+        }
       }
     };
 
@@ -357,7 +363,6 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
       ...durationOverride,
       disabled: false,
       transfer: false,
-      origin: origin.uuid,
       _stats: {
         [effect.inCompendium ? "compendiumSource" : "duplicateSource"]: effect.uuid,
         [effect.inCompendium ? "duplicateSource" : "compendiumSource"]: null
--- a/module/applications/components/effects.mjs
+++ b/module/applications/components/effects.mjs
@@ -297,11 +297,9 @@ export default class EffectsElement extends (foundry.applications.elements.Adopt
 
     const dataset = target.closest("[data-effect-id]")?.dataset;
     const effect = this.getEffect(dataset);
-    if ( (action !== "create") && !effect ) return;
+    if ( !effect ) return;
 
     switch ( action ) {
-      case "create":
-        return this._onCreate(target);
       case "delete":
         await effect.deleteDialog({ sheet: this.#app }, { render: false });
         return this.#app.render();
@@ -391,6 +389,10 @@ export default class EffectsElement extends (foundry.applications.elements.Adopt
    * @returns {Promise<ActiveEffect5e>}
    */
   async _onCreate(target) {
+    foundry.utils.logCompatibilityWarning(
+      "`EffectsElement#_onCreate` method is not longer supported.",
+      { since: "DnD5e 6.0", until: "DnD5e 6.2" }
+    );
     const li = target.closest("li");
     const isActor = this.document instanceof Actor;
     const isEnchantment = li.dataset.effectType.startsWith("enchantment");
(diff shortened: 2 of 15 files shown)


module/documents/active-effect.mjs
roth-michael, line 716: I wonder if this bit should be changed to make use of the new fields, rather than the derived `origin` field.

roth-michael: Don't think we want to be setting `origin` directly any more since it's derived during data prep. Instead would want to set values of the new fields.

roth-michael (commented):
As a more general question, I wonder whether we shouldn't populate as many of the new `origin` subfields as we can. e.g. in `EnchantActivity` we set `origin.activity` and `origin.effect`, but we _could_ also set `origin.item`, which could be useful if attempting to compare origin item to some other item without actually retrieving the activity & going to its parent.

module/applications/components/effect-application.mjs
```diff
@@ -182,12 +182,13 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
...
+   * @param {Actor5e} actor              The actor to which the effect should be applied.
    * @returns {Promise<ActiveEffect5e>}  The created effect.
    * @throws {Error}                     If the effect could not be applied.
    * @protected
    */
   async _applyEffectToActor(effect, actor) {
+    const activity = this.chatMessage.getAssociatedActivity({ scaled: true });
     const concentration = this.chatMessage.getAssociatedActor()?.effects.get(this.chatMessage.system.concentration);
     const item = this.chatMessage.getAssociatedItem();
     const origin = concentration ?? (effect.inCompendium && item ? item : effect);
```
Fyorl: Now that we are not setting the `origin` as `origin.uuid`, we need to find an existing effect differently. Using `duplicateSource` is probably the most reliable.


```suggestion
    const sourceKey = effect.inCompendium ? "compendiumSource" : "duplicateSource";
```

module/applications/components/effect-application.mjs
```diff
@@ -234,7 +241,6 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
       ...durationOverride,
       disabled: false,
       transfer: false,
-      origin: origin.uuid,
       _stats: {
         [effect.inCompendium ? "compendiumSource" : "duplicateSource"]: effect.uuid,
```
Fyorl:
```suggestion
        [sourceKey]: effect.uuid,
```

module/applications/components/effect-application.mjs
Fyorl:
```suggestion
      const existingEffect = actor.effects.find(e => e._stats[sourceKey] === effect.uuid);
```

module/data/item/templates/physical-item.mjs
```diff
@@ -323,7 +323,15 @@ export default class PhysicalItemTemplate extends SystemDataModel {
...
-        effects: [{ ...enchantment.toObject(), disabled: false, origin: enchantment.parent.uuid }]
+        effects: [{
+          ...enchantment.toObject(),
+          disabled: false,
+          system: {
+            origin: {
+              item: enchantment.parent.uuid
+            }
+          }
+        }]
```
Fyorl:
```suggestion
        effects: [foundry.utils.mergeObject(enchantment.toObject(), {
          disabled: false,
          system: {
            origin: {
              item: enchantment.parent.uuid
            }
          }
        })]
```

module/migration.mjs
```diff
@@ -1153,6 +1154,33 @@ function _migrateEffectMagical(effect, parent, updateData) {
...
+ * Migrates effect origin data from core field into system data.
+ * @param {object} effect      Effect data to migrate.
+ * @param {object} parent      The parent of this effect.
+ * @param {object} updateData  Existing update to expand upon.
+ * @returns {object}           The updateData to apply.
+ */
+function _migrateEffectOrigin(effect, parent, updateData) {
+  const origin = effect.flags?.core?.originText ?? effect.origin;
+  const systemDataModel = CONFIG.ActiveEffect.dataModels[effect.type];
+  if ( !origin || !systemDataModel?.schema.fields.origin ) return updateData;
```
Fyorl: Bail if we already migrated this effect.
```suggestion
  if ( !origin || effect.system?.origin || !systemDataModel?.schema.fields.origin ) return updateData;
```

module/applications/components/effects.mjs
```diff
@@ -391,6 +389,10 @@ export default class EffectsElement extends (foundry.applications.elements.Adopt
    * @returns {Promise<ActiveEffect5e>}
    */
   async _onCreate(target) {
+    foundry.utils.logCompatibilityWarning(
+      "`EffectsElement#_onCreate` method is not longer supported.",
```
Fyorl:
```suggestion
      "`EffectsElement#_onCreate` method is no longer supported.",
```

module/applications/item/item-sheet.mjs
Fyorl:
```suggestion
      effectData.system.origin.item ??= effect.parent?.uuid;
```

module/migration.mjs
```diff
@@ -1153,6 +1154,33 @@ function _migrateEffectMagical(effect, parent, updateData) {
...
+
+  const field = origin.includes("Activity") ? "activity"
```
Fyorl, line 1186: Should be able to check `parseUuid(origin).type`, no?

Fyorl (changes requested):

Fyorl:
> As a more general question, I wonder whether we shouldn't populate as many of the new `origin` subfields as we can. e.g. in `EnchantActivity` we set `origin.activity` and `origin.effect`, but we _could_ also set `origin.item`, which could be useful if attempting to compare origin item to some other item without actually retrieving the activity & going to its parent.

Need to be careful with this because of the precedence we have for `origin`. If we set `system.origin.activity` then for example `ItemSheet5e#_onDropActiveEffect` that is comparing against `effect.origin`, expecting it to be an `Item5e` will suddenly stop working. So we need to take care to audit for all those places.

module/migration.mjs
arbron: Origin could contain relative UUIDs and I don't believe we can properly parse those without the relative document. Currently we only pass in the parent's source data, so it would be a pretty big refactor to get the actual document that far down the chain.

module/documents/active-effect.mjs
arbron: Could, but I think we'll still need the fallback for non-migrated origins so I'm not sure it is worth it. Best case it would save a single `instanceof` check, worst case it would lead to multiple `fromUuid` calls.

module/documents/active-effect.mjs
```diff
@@ -747,7 +744,7 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
       if ( effectData ) {
         delete effectData._id;
         delete effectData.flags?.dnd5e?.rider;
-        effectData.origin = this.origin;
+        foundry.utils.setPoperty(effectData, "system.origin", this.system.origin);
```
```suggestion
        foundry.utils.setProperty(effectData, "system.origin", this.system.origin);
```

module/applications/item/item-sheet.mjs
```diff
@@ -965,12 +967,13 @@ export default class ItemSheet5e extends PrimarySheetMixin(DocumentSheet5e) {
     const effect = await ActiveEffect.implementation.fromDropData(data);
     if ( !this.item.isOwner || !effect
       || (this.item.uuid === effect.parent?.uuid)
-      || (this.item.uuid === effect.origin) ) return false;
+      || (this.item.uuid === effect.origin)
+      || (this.item.uuid === effect.system.origin?.uuid) ) return false;
```
Fyorl:
```suggestion
      || (this.item.uuid === effect.system.origin?.item) ) return false;
```

module/documents/activity/enchant.mjs
Fyorl: Re-using a concentration self-enchant was adding a duplicate.
```suggestion
      ? this.item.effects.find(e => e.isAppliedEnchantment && (e.system.origin?.activity === this.uuid)) : undefined;
```

module/documents/active-effect.mjs
```diff
@@ -796,7 +793,16 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
   /** @inheritDoc */
   async _preCreate(data, options, user) {
     if ( await super._preCreate(data, options, user) === false ) return false;
-    if ( options.keepOrigin === false ) this.updateSource({ origin: this.parent.uuid });
+    if ( options.keepOrigin === false ) this.updateSource({
```
Fyorl, line 797: Probably should still clear base `origin` as well. `ConditionData` doesn't have `system.origin` (maybe it should?)

Fyorl (changes requested):
For `ApplyActiveEffectRegionBehaviorType##getEffectsToCreate` we write `data.origin` which I guess is all we can do because we don't have a place in `system.origin` for a behaviour UUID. But then we need to delete `system.origin` too otherwise it takes precedence and points to the wrong thing.

In `Actor5e#transformInto` there are few places we check `e.origin` when we need to check `e.system.origin`.

I wonder if a small helper like this, maybe on the AE model, will make some of these checks a bit more ergonomic. Typically we just want to know if one of the `system.origin` UUIDs matches.

```js
matchesOrigin(other) {
  const o = this.system.origin;
  return (this._source.origin === uuid)
    || (o?.activity === uuid)
    || (o?.actor === uuid)
    || (o?.effect === uuid)
    || (o?.item === uuid)
}
```

module/documents/active-effect.mjs
arbron: Do we have any instances where Condition active effects are not applied directly by the player?

Fyorl (approved):

PR #7303, opened 2026-08-03
Refactor card data pipeline

Fyorl:
This is a larger body of work that I am trying to split up so it's easier to review. So bear in mind that some of this code is temporary wiring to keep things somewhat testable in isolation.

The motivation here is to have `getCardData` return structured data that directly becomes the `usage` message's `system` data. To do that I spun off `getTooltipData` that produces actual tooltip render context from the structured card data, taking into account some differences around displaying identified/unidentified information to GMs (I preserved status quo behaviour here).

As part of this work, I also refactored the pills/property generation to produce structured data. I think these properties (maybe 'tags' is a better word?) are fairly important, and there may be more emphasis on them in future, so I wanted to move away from us shuttling around opaque strings.

This means the various `*Properties` getters are updated to also return structured data, which I think is an improvement. All the various areas in the UI that generate pills have been consolidated to generate them in the same way now.

Summary from the commit is repeated here:
 
 - `getCardData` now returns structured data only, suitable for storing in a schema
 - `getTooltipData` converts `getCardData` into tooltip context suitable for tooltip display
 - Added `getUsageData` on Items & Activities to produce activation usage structured data
 - Adjusted various `*Properties` getters to also produce structured data
 - Added `PropertyField` as a `TypedSchemaField` subclass that wraps an individual pill/property element
 - Added `PropertyField.getLabels` which converts structured data to a human-readable label for pill display
 - Consolidated all sites that generate Item pills to go via `PropertyField.getLabels` and transact structured data
 - Removed `tooltipSubtitle` getter and rolled it into `getTooltipData`

With the return values changing and removal of some getters I've marked this PR as breaking. I suspect that downstream is not really using these methods at all, but if they are, I think we can provide a deprecation path, but that will involve creating new methods/getter names rather than repurposing these ones.

### DIFF
--- a/module/data/abstract/item-data-model.mjs
+++ b/module/data/abstract/item-data-model.mjs
@@ -1,4 +1,5 @@
 import * as Trait from "../../documents/actor/trait.mjs";
+import PropertyField from "../shared/property-field.mjs";
 import SystemDataModel from "./system-data-model.mjs";
 
 const TextEditor = foundry.applications.ux.TextEditor.implementation;
@@ -140,16 +141,6 @@ export default class ItemDataModel extends SystemDataModel {
   }
 
   /* -------------------------------------------- */
-
-  /**
-   * Parts making up the subtitle on the item's tooltip.
-   * @type {string[]}
-   */
-  get tooltipSubtitle() {
-    return [this.type?.label ?? _loc(CONFIG.Item.typeLabels[this.parent.type])];
-  }
-
-  /* -------------------------------------------- */
   /*  Data Preparation                            */
   /* -------------------------------------------- */
 
@@ -188,7 +179,7 @@ export default class ItemDataModel extends SystemDataModel {
   async richTooltip(enrichmentOptions={}) {
     return {
       content: await foundry.applications.handlebars.renderTemplate(
-        this.constructor.ITEM_TOOLTIP_TEMPLATE, await this.getCardData(enrichmentOptions)
+        this.constructor.ITEM_TOOLTIP_TEMPLATE, await this.getTooltipData(enrichmentOptions)
       ),
       classes: ["dnd5e2", "dnd5e-tooltip", "item-tooltip", "themed", "theme-light"]
     };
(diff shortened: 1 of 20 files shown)


module/data/abstract/item-data-model.mjs
```diff
@@ -197,57 +188,92 @@ export default class ItemDataModel extends SystemDataModel {
...
+    const { description: desc, unidentified } = this;
+    const description = (context.identified === false) ? unidentified?.description : desc.value;
+    let { identified, price, uses } = this;
+    uses = this.hasLimitedUses && (game.user.isGM || identified) ? uses : null;
+    price = game.user.isGM || identified ? price : null;
+
+    Object.assign(context, {
+      name, type, img, price, uses, extras,
+      config: CONFIG.DND5E,
+      controlHints: dnd5e.settings.controlHints,
```
Fyorl, line 270: Oops this is actually wrong I remember but the fix was in the next commit.

module/data/shared/property-field.mjs
```diff
@@ -0,0 +1,111 @@
...
+        case "range": return (range.long && (range.long !== range.value))
+          ? `${formatNumber(range.value)}/${formatLength(range.long, range.units)}`
+          : RangeField.getLabels({ range }).simple;
+        case "reach": return _loc("DND5E.RANGE.Formatted.Reach", { reach: formatLength(range.reach, range.units) });
+        case "target": {
+          const labels = TargetField.getLabels({ target });
+          return labels.template.statblock || labels.affects.sheet;
+        }
+        case "text": return p.text;
+        case "weight": return `${p.value} ${_loc("DND5E.AbbreviationLbs")}`;
```
arbron: This seems like a longstanding issue where the tag won't properly reflect the selected weight units. We should probably pass in the unit and then use `formatWeight`.

arbron (changes requested):

arbron (approved):

PR #7304, opened 2026-08-03
Add schema for basic item card and render from it.

Fyorl:
This is the basic card that is generated when invoking `Item5e#displayCard` or using an activity-less items. So it carries the bulk of the schema that will be shared with the activity card, which inherits from this one.

Summary from the commit is repeated here:

 - Added `ItemMessageData` and associated `item` system sub-type.
 - Posting an item card snapshots its state via `getCardData`.
 - Item cards now render from snapshotted state data rather than content blob.
 - `flags.dnd5e.item.*` migrated to `system.item.*`. `getAssociatedItem` reads from `system.item` and legacy flags.
 - Removed `item()` and `activity()` getters which now conflict with system data fields.


module/data/chat-message/_types.mjs
```diff
@@ -100,6 +101,46 @@
...
 
+/**
+ * @typedef ItemMessageSystemData
+ * @property {ActivationData|null} activation    How the item is activated, if at all.
+ * @property {boolean} concealed                 Whether the description is concealed from players.
+ * @property {string} description                Enriched description displayed on the card.
+ * @property {DurationData|null} duration        How long the item's effects last.
+ * @property {boolean|null} identified           Identified state when the card was created, if applicable.
+ * @property {ItemMessageItemData} item          Item the card describes.
+ * @property {number|null} level                 Level at which a spell was cast.
```
Fyorl, line 112: This replaces `spellLevel`. Maybe too generic but I thought it was reasonable.

module/data/chat-message/_types.mjs
```diff
@@ -100,6 +101,46 @@
...
+ * @property {string|null} id    Document ID.
+ * @property {string} img        Document image.
+ * @property {string} name       Document name.
+ * @property {string} type       Document type.
+ * @property {string|null} uuid  Document UUID.
+ */
+
+/**
+ * @typedef {ItemMessageSourceData} ItemMessageItemData
+ * @property {string|null} compendiumSource  UUID of the compendium item this item is based on.
```
Fyorl, line 133: Not used but I thought this might be useful.

module/data/chat-message/item-message-data.mjs
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Fields describing a document a card was created from.
+   * @returns {Record<string, DataField>}
+   * @protected
+   */
+  static _sourceFields() {
```
Fyorl, line 66: This becomes a shared field definition in a later PR.

templates/chat/item-card.hbs
```diff
@@ -1,44 +1 @@
...
-    <ul class="card-footer pills unlist">
-        {{#each data.properties}}
-        <li class="pill transparent">
-            <span class="label">{{ this }}</span>
-        </li>
-        {{/each}}
-    </ul>
-    {{/if}}
-</div>
+{{> "dnd5e.card-face" }}
```
Fyorl, line 1: This partial is shared by the activity card later. We could remove `item-card.hbs` entirely but I thought maybe we keep it around in case we want to add some item-specific wrapping or something.

system.json
```diff
@@ -56,6 +56,7 @@
       "healing": {},
       "hitDie": {},
       "hitPoints": {},
+      "item": {},
```
arbron:
```suggestion
      "item": {
        "filePathFields": {
          "item.img": ["IMAGE"]
        },
        "htmlFields": ["description"]
      },
```

arbron (changes requested):

module/data/chat-message/_types.mjs
arbron: I guess the question is, since `UsageMessageData` already has `spellLevel` in system data are you planning on migrating that. The PR at the moment ends up with both `system.level` and `system.spellLevel` on usage cards.

krbz999: Can it be used for more than spells?

Fyorl: > I guess the question is, since `UsageMessageData` already has `spellLevel` in system data are you planning on migrating that. The PR at the moment ends up with both `system.level` and `system.spellLevel` on usage cards.

I was planning on migrating it, yes.

arbron (approved):

PR #7306, opened 2026-08-03
Convert usage buttons to descriptors.

Fyorl:


module/documents/activity/mixin.mjs
```diff
@@ -796,6 +790,8 @@ export default function ActivityMixin(Base) {
         title: `${this.item.name} - ${this.name}`,
         type: messageType
       };
+      const buttons = this._usageChatButtons(message);
```
arbron, line 793: We are still calling this in `_usageChatContext` also, which should be safe to remove.

arbron (approved):

module/documents/activity/mixin.mjs
Fyorl: Thank you, yeah, I had cleanup at the end which I am just splitting out now.

PR #7307, opened 2026-08-03
Convert activity use to new chat system data. Adjust activity chat handlers to transact button descriptors rather than DOM nodes.

Fyorl:
`UsageMessageData` now inherits from `ItemMessageData`. It gets an `activity` source field that replaces `flags.dnd5e.activity` or `flags.dnd5e.use.activity`. `shouldHideChatButton` now gets the stored button descriptor rather than raw strings from the DOM node's `dataset`, which I think is an improvement. Again this will be breaking though if downstream are implementing this method, and again there is potentially a deprecation path available if we want to consider it.

### DIFF
--- a/module/data/chat-message/usage-message-data.mjs
+++ b/module/data/chat-message/usage-message-data.mjs
@@ -1,26 +1,32 @@
-import ChatMessageDataModel from "../abstract/chat-message-data-model.mjs";
+import ItemMessageData from "./item-message-data.mjs";
 import { ActorDeltasField } from "./fields/deltas-field.mjs";
 
 const { ArrayField, DocumentIdField, NumberField, ObjectField, SchemaField, StringField } = foundry.data.fields;
 
 /**
+ * @import { ActivityUsageChatButton } from "../../documents/activity/_types.mjs";
  * @import { UsageMessageSystemData } from "./_types.mjs";
  */
 
 /**
  * Data stored in a usage chat message.
- * @extends {ChatMessageDataModel<UsageMessageSystemData>}
+ * @extends {ItemMessageData<UsageMessageSystemData>}
  * @mixes UsageMessageSystemData
  */
-export default class UsageMessageData extends ChatMessageDataModel {
+export default class UsageMessageData extends ItemMessageData {
 
   /* -------------------------------------------- */
   /*  Model Configuration                         */
   /* -------------------------------------------- */
 
-  /** @override */
+  /** @inheritDoc */
   static defineSchema() {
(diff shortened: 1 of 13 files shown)


module/data/chat-message/usage-message-data.mjs
```diff
@@ -1,26 +1,32 @@
...
   /*  Model Configuration                         */
   /* -------------------------------------------- */
 
-  /** @override */
+  /** @inheritDoc */
   static defineSchema() {
     return {
+      ...super.defineSchema(),
+      activity: new SchemaField({
+        ...this._sourceFields(),
```
arbron, line 27: Getting an error whenever attempting to create a chat card because `DocumentUUIDField` doesn't like activity UUIDs.

Image

arbron (changes requested):

module/data/chat-message/usage-message-data.mjs
```diff
@@ -1,26 +1,33 @@
...
+        chatFlavor: new StringField(),
+        uuid: new StringField({ blank: false, nullable: false, required: true })
```
arbron: Probably want this to be nullable to match `system.item.uuid`, as-is it causes an error for existing chat cards.


```suggestion
        uuid: new StringField({ blank: false, nullable: true, required: true })
```

arbron (changes requested):

module/data/chat-message/usage-message-data.mjs
Fyorl: I did have a migration for this, it's split off here: https://github.com/foundryvtt/dnd5e/pull/7310

So existing chat cards shouldn't error. Unless you are saying it's possible that they don't have `flags.dnd5e.activity.uuid`. I didn't think it was possible to have a usage card without one, but maybe the really old `flags.dnd5e.use` cards?

arbron: Tested the migration in #7310 with data from 5.3 and none of the usage chat cards migrated since by the time the world migration runs they are already considered invalid documents due to the missing UUIDs. We could handle that by adding a migration into `ChatMessage#migrateData` to handle just the activity UUIDs.

Fyorl: Hmm, seems really old cards with `flags.dnd5e.use` would get migrated to `usage` type and then also fail validation because they have no activity UUID to migrate. So the only way is with `nullable: true` after all to allow the world migration to work.

arbron (approved):

PR #7308, opened 2026-08-03
Cleanup some unused legacy code paths after usage message refactor.

Fyorl:
- Delete old usage message render path.
- Remove various unused handlers from the base chat message class.
- Refactor `label` + `labelHidden` -> `label: { hidden, value }`


module/documents/chat-message.mjs
```diff
@@ -44,14 +42,6 @@ export default class ChatMessage5e extends ChatMessage {
 
   /* -------------------------------------------- */
 
-  /** @inheritDoc */
-  get isRoll() {
```
Fyorl, line 48: Unrelated to the usage message but I think this got missed when the rest card got its own sub-type.

module/documents/activity/mixin.mjs
```diff
@@ -676,40 +674,6 @@ export default function ActivityMixin(Base) {
 
     /* -------------------------------------------- */
 
-    /**
-     * Prepare the context used to render the usage chat card.
-     * @param {ActivityMessageConfiguration} message  Configuration info for the created message.
-     * @returns {object}
-     * @protected
-     */
-    async _usageChatContext(message) {
```
arbron, line 685: This removal looks like it might break 3rd-party modules that are shipping their own activities.

module/documents/activity/_types.mjs
```diff
@@ -75,10 +74,10 @@
  * @typedef ActivityUsageChatButton
  * @property {string} action                      Action performed when the button is clicked.
  * @property {object} [dataset]                   Parameters passed to the action that handles the button.
- * @property {string} [hiddenLabel]               Label displayed in place of label when challenge details are
- *                                                concealed from the viewer.
  * @property {string} icon                        FontAwesome classes or a path to an icon image.
- * @property {string} label                       Localization key or label text.
+ * @property {object} label                       Labels displayed on the button.
+ * @property {string} [label.hidden]              Label displayed when challenge details are concealed from the viewer.
+ * @property {string} label.value                 Localization key or label text.
```
arbron, line 80: Should probably have a migration for any modules that are providing their own buttons in their own activities. Just a check if `label` is a String or if `hiddenLabel` exists and migrating that to the new form before creating the card should be sufficient.

module/documents/activity/_types.mjs
```diff
@@ -16,7 +16,6 @@
  * @property {typeof ActivitySheet} sheetClass          Sheet class used to configure this activity.
  * @property {object} usage
  * @property {Record<string, Function>} usage.actions   Actions that can be triggered from the chat card.
- * @property {string|null} usage.chatCard               Template used to render the chat card.
```
arbron, line 19: Another place a migration would probably be good to handle 3rd-party activities.

arbron (changes requested):

module/documents/activity/mixin.mjs
Fyorl: I'm not sure how we offer a deprecation shim that isn't just leaving the entire template render path alive. Which we could do with a deprecation message, but I wasn't sure if that's what you had in mind. Did you mean we should just keep the method so that it doesn't throw?

Migration path downstream is reasonably straightforward since we still honour `content`, but they would basically have to copy `_usageChatContext` into their own module.

Fyorl: I added the deprecations in https://github.com/foundryvtt/dnd5e/pull/7311 and it was roughly what I expected. We have to keep the old `activity-card.hbs` behind and most of the rendering code. But it's not worse than what we've had to do before.

module/documents/activity/_types.mjs
Fyorl: Modules shouldn't be providing `hiddenLabel` as that was added as part of this stack (and I change it to `label.hidden` here). But I have added a shim to handle legacy button formats.

arbron (approved):

PR #7309, opened 2026-08-03
Refactor getTargetDescriptors to provide the token UUID. Migrate targets to system data.

### LINKED ISSUE #4720: [Question - 4.1.1] Why does getTargetDescriptors() return Actor specific data, instead of targeted Token?
I would propose for this to return:
- the `token.document.uuid` instead of `token.actor.uuid` so that it is easier to get the "right" targeted token in case multiple placeables from a linked actor have been placed (debatable usecase, but still it happens).

- the `token.document.texture.src` instead of the `token.actor.img`, so that it can pick up the "proper" targeted token's image mainly for unlinked actors .
   - that could be gated behind a general system setting (actor/tokenDocument img priority)
   - Might need some extra sauce for animated tokens 
```js
//for example
if (VideoHelper.hasVideoExtension(img)) {
   img = await game.video.createThumbnail(img, { width: 100, height: 100 });
}
```

That information is readily available and I think that removing the specificity of the target is not ideal.

Now, I am not sure yet as to what that might be affecting downstream, but could take a look if need be.

If the issue is the need to make sure that the targeted token is not an actorless one, I guess the already implemented check for `if ( uuid )` can still be present and then just pass along 
```js
const { name, document: { texture: { src: tokenImg }, uuid: tokenDocumentUuid } } = token;
const { img: actorImg, system, uuid: actorUuid, statuses } = token.actor ?? {};
if ( actorUuid ) {
      const ac = statuses.has("coverTotal") ? null : system.attributes?.ac?.value;
      const img = "some settings test" ? actorImg : tokenImg;
      targets.set(uuid, { name, img, uuid: tokenDocumentUuid, ac: ac ?? null });
}
```

Relevant code snippet.
```js
/**
 * Important information on a targeted token.
 *
 * @typedef {object} TargetDescriptor5e
 * @property {string} uuid  The UUID of the target.
 * @property {string} img   The target's image.
 * @property {string} name  The target's name.
 * @property {number} ac    The target's armor class, if applicable.
 */

/**
 * Grab the targeted tokens and return relevant information on them.
 * @returns {TargetDescriptor[]}
 */
function getTargetDescriptors() {
  const targets = new Map();
  for ( const token of game.user.targets ) {
    const { name } = token;
    const { img, system, uuid, statuses } = token.actor ?? {};
    if ( uuid ) {
      const ac = statuses.has("coverTotal") ? null : system.attributes?.ac?.value;
      targets.set(uuid, { name, img, uuid, ac: ac ?? null });
    }
  }
  return Array.from(targets.values());
}
```

Suggested change 
```js
/**
 * Important information on a targeted token.
 *
 * @typedef {object} TargetDescriptor5e
 * @property {string} uuid  The UUID of the target.
 * @property {string} img   The target's image.
 * @property {string} name  The target's name.
 * @property {number} ac    The target's armor class, if applicable.
 */

/**
 * Grab the targeted tokens and return relevant information on them.
 * @returns {TargetDescriptor[]}
 */
function getTargetDescriptors() {
  const targets = new Map();
  for ( const token of game.user.targets ) {
    const { name } = token;
    const { img, system, uuid, statuses } = token.actor ?? {};
    if ( uuid ) {
      const ac = statuses.has("coverTotal") ? null : system.attributes?.ac?.value;
      targets.set(uuid, { name, img, uuid, ac: ac ?? null });
    }
  }
  return Array.from(targets.values());
}
```

Fyorl:
- Closes #4720 

We should generally be able to losslessly retrieve the exact targeted token now. Also storing token image here rather than actor image. But with the planned new chat card designs I think I won't use it, so maybe it should be removed, and then we don't need such lengthy sanitisation schema.

### DIFF
--- a/module/documents/chat-message.mjs
+++ b/module/documents/chat-message.mjs
@@ -1,4 +1,5 @@
 import aggregateDamageRolls from "../dice/aggregate-damage-rolls.mjs";
+import { resolveTargetDescriptor } from "../utils.mjs";
 
 export default class ChatMessage5e extends ChatMessage {
 
@@ -161,6 +162,7 @@ export default class ChatMessage5e extends ChatMessage {
     html.classList.add("dnd5e2");
 
     // Header matter
+    const token = this.getAssociatedToken();
     const actor = this.getAssociatedActor();
     const avatar = document.createElement("a");
     avatar.classList.add("avatar");
@@ -187,7 +189,8 @@ export default class ChatMessage5e extends ChatMessage {
     }
     img ??= CONST.DEFAULT_TOKEN;
 
-    if ( actor ) avatar.dataset.uuid = actor.uuid;
+    if ( actor ) avatar.dataset.actorUuid = actor.uuid;
+    if ( token ) avatar.dataset.tokenUuid = token.uuid;
     Object.assign(avatarImg, { src: img, alt: nameText });
     avatar.append(avatarImg);
 
@@ -320,10 +323,9 @@ export default class ChatMessage5e extends ChatMessage {
    */
   async _onTargetMouseDown(event) {
     event.stopPropagation();
-    const uuid = event.currentTarget.dataset.uuid;
-    const actor = fromUuidSync(uuid);
-    const token = actor?.token?.object ?? actor?.getActiveTokens()[0];
-    if ( !token || !actor.testUserPermission(game.user, "OBSERVER")) return;
+    const { actorUuid, tokenUuid } = event.currentTarget.dataset;
+    const { actor, token } = resolveTargetDescriptor({ actor: actorUuid, token: tokenUuid });
+    if ( !token || !actor?.testUserPermission(game.user, "OBSERVER")) return;
     const releaseOthers = !event.shiftKey;
     if ( token.controlled ) token.release();
     else {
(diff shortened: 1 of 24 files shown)


module/migration.mjs
```diff
@@ -786,24 +786,32 @@ export function migrateMacroData(macro, migrationData) {
...
-  if ( messageData.type !== "base" ) return {};
-
   const updateData = {};
   const { flags } = messageData;
-  const rollType = foundry.utils.getProperty(flags, "dnd5e.roll.type");
+  const targets = flags?.dnd5e?.targets?.map(({ ac, img, name, uuid }) => ({ ac, img, name, actor: uuid }));
+  if ( targets ) updateData["flags.dnd5e.targets"] = _del;
+
+  if ( messageData.type !== "base" ) {
+    if ( targets && CONFIG.ChatMessage.dataModels[messageData.type]?.schema.has("targets") ) {
```
Fyorl, line 795: I see this got left in, but I ditch the schema check in a later commit.

module/documents/chat-message.mjs
```diff
@@ -320,10 +323,9 @@ export default class ChatMessage5e extends ChatMessage {
    */
   async _onTargetMouseDown(event) {
     event.stopPropagation();
-    const uuid = event.currentTarget.dataset.uuid;
-    const actor = fromUuidSync(uuid);
-    const token = actor?.token?.object ?? actor?.getActiveTokens()[0];
-    if ( !token || !actor.testUserPermission(game.user, "OBSERVER")) return;
+    const { actorUuid, tokenUuid } = event.currentTarget.dataset;
+    const { actor, token } = TargetsField.resolve({ actor: actorUuid, token: tokenUuid });
+    if ( !token || !actor?.testUserPermission(game.user, "OBSERVER")) return;
```
arbron:
```suggestion
    if ( !token || !actor?.testUserPermission(game.user, "OBSERVER") ) return;
```

arbron (approved):

thatlonelybugbear:
However this is closed, it will be closing #4720 too

PR #7310, opened 2026-08-03
Migrate item, activity, and originatingMessage flags into system data

Fyorl:


module/data/chat-message/roll-message-data.mjs
```diff
@@ -26,6 +29,11 @@ export default class RollMessageData extends ChatMessageDataModel {
   /** @override */
   static defineSchema() {
     return {
+      activity: new SourceReferenceField({}, { initial: null, nullable: true }),
```
arbron: This will need the UUID field replacement to match `UsageMessageData` to avoid the activity UUID issue.

arbron (changes requested):

arbron (approved):

PR #7312, opened 2026-08-04
Migrate spellLevel -> level on UsageMessageData

Fyorl:


module/documents/activity/mixin.mjs
```diff
@@ -512,10 +512,7 @@ export default function ActivityMixin(Base) {
         usageConfig.scaling = Math.max(0, levelingFlag.value - levelingFlag.base);
       } else if ( this.isSpell ) {
         const level = this.actor.system.spells?.[usageConfig.spell?.slot]?.level;
-        if ( level ) {
-          usageConfig.scaling = level - item.system.level;
-          foundry.utils.setProperty(messageConfig, "data.system.spellLevel", level);
```
Fyorl, line 517: So long as we set scaling, `rollData.item.level` becomes the correct upcast level, which then get written to `system.level` in `getCardData`, so we don't need to write it here anymore.

arbron (approved):

PR #7314, opened 2026-08-04
CSS cleanup

Fyorl:
- Normalises scrollbar styles across all browser/OS combos (that I could test). Firefox on Linux has 1px overlay scrollbars which means several sheets/tabs have no padding on the right. MacOS has an OS-level setting apparently which controls overlay scrollbars that most browsers seem to honour. And Chrome's 'thin' non-overlay scrollbars are the fattest of the lot.
- Use doughnut scoping to implement theming rather than layers.
- Remove vestigial system theme machinery.

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -477,6 +477,14 @@ Hooks.once("setup", function() {
     }
   `;
   document.head.append(style);
+
+  // Measure the space a scrollbar takes up so that sheets can center theirs within their padding.
+  const probe = document.createElement("div");
+  probe.style.cssText = "position: absolute; visibility: hidden; overflow-y: auto; scrollbar-gutter: stable; "
+    + "scrollbar-width: thin; inline-size: 100px; block-size: 100px";
+  document.body.append(probe);
+  document.documentElement.style.setProperty("--dnd5e-scrollbar-width", `${probe.offsetWidth - probe.clientWidth}px`);
+  probe.remove();
 });
 
 /* --------------------------------------------- */
--- a/less/v2/high-contrast/character.less
+++ b/less/v2/high-contrast/character.less
@@ -38,24 +38,26 @@
 /*  Light Mode                        */
 /* ---------------------------------- */
 
-.dnd5e-flag-high-contrast:is(.dnd5e-theme-light, .theme-dark) .dnd5e2.sheet.actor.character,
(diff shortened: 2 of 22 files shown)


dnd5e.mjs
```diff
@@ -477,6 +477,14 @@ Hooks.once("setup", function() {
     }
   `;
   document.head.append(style);
+
+  // Measure the space a scrollbar takes up so that sheets can center theirs within their padding.
```
Fyorl, line 481: No way to do this without javascript as far as I could tell.

less/v2/high-contrast/character.less
```diff
@@ -38,24 +38,26 @@
...
-.dnd5e2.sheet.actor.character.dnd5e-flag-high-contrast.dnd5e-theme-light {
-  .sheet-header { background: var(--dnd5e-color-maroon); }
-  .sidebar .favorites > h3 { color: var(--dnd5e-color-dark); }
-  .tab.features .classes .class .level > span { color: var(--dnd5e-color-dark); }
+@scope (.dnd5e-flag-high-contrast.theme-light, .dnd5e-flag-high-contrast .theme-light) to (.themed) {
+  .dnd5e2.sheet.actor.character, :where(:scope).dnd5e2.sheet.actor.character {
+    .sheet-header { background: var(--dnd5e-color-maroon); }
+    .sidebar .favorites > h3 { color: var(--dnd5e-color-dark); }
+    .tab.features .classes .class .level > span { color: var(--dnd5e-color-dark); }
+  }
```
arbron, line 46: Some of the colors for dark mode aren't getting applied correctly. Looks like its because we are defining colors above outside of the light/dark mode blocks.

Image

Moving these two definitions from above into the light mode section fixes the issue:

```suggestion
    .ability-scores .ability-score {
      .label { color: var(--dnd5e-color-dark); }
    }
    .skills, .tools {
      .ability, .passive { color: var(--color-text-dark-primary); }
    }
  }
```

arbron (changes requested):

less/v2/high-contrast/character.less
Fyorl: Still getting the hang of how `@scope` works. I think moving some of these variables inside the scopes, and using already-themed variables elsewhere should have fixed this. Let me know if you spot anything else.

arbron (approved):

PR #7319, opened 2026-08-07
Expose `transfer` field for enchantments, use to determine `isApplied`

### LINKED ISSUE #7318: Consider a way of specifying whether an Enchantment should apply to its parent item or not via the UI
Currently `isApplied` determines this based on the effect's origin (it must exist) and its current parent item (it must not match the origin). This means one cannot create an enchantment directly on an item and have it affect that item, or create an enchantment in an AE compendium and drag it onto items to have it affect them. 

One potential approach would be adding some enchantment-specific analog to `ActiveEffect`'s `transfer` property. This value could be used directly for `isApplied`, and could default (and be migrated) based on the current `isApplied` logic.

roth-michael:
Closes #7318 
An enchantment created directly on an item may now be set to apply to said item. Most of the code changes should be fairly self-explanatory. Potential exceptions:
- `EnchantSheet#_processSubmitData` newly overridden. Any newly-added local enchantments on an Enchant activity will be updated to `transfer: false` if they were previously `transfer: true`. While not _necessary_ (this is double checked in `isApplied`) it is cleaner if all enchantments which exist as an enchant activity have a visibly false `transfer` checkbox. In theory this logic could be done in `_onUpdate` somewhere to also account for programmatic changes I suppose. And could block `transfer: true` in `EnchantmentData#_preUpdate` when necessary.
- `EnchantmentData#onRenderActiveEffectConfig` now also replaces label & hint text for the transfer form group (if it exists) and disables & adds a tooltip to the checkbox itself (if it should be disabled).
- Don't _love_ the `_migrateEnchantmentTransfer` origin/id comparison, but I think it's the best we can do since we don't have direct access to the uuid at that point. With the migration as-is, any existing enchantments for which `isApplied` would have returned `true` will receive `transfer: true`, all others will receive `transfer: false`. The exception introduced by the lack of uuid would be if an enchantment had an `origin` of an _Activity_ whose id matched exactly the id of the _Item_ the activity was on. In that absurdly unlikely scenario, `isApplied` would have returned `true` but it would receive `transfer: false`. I'm not worried about it.
- Exposed the transfer checkbox for parentless Enchantments. This is just so that one could configure whether a compendium Enchantment should be "I drag this on to affect this item" or "I drag this on because I'm going to use it on an enchant activity." Would be open to adding a `isEnchantment` property to context rather than doing a string comparison there.

### DIFF
--- a/module/applications/activity/enchant-sheet.mjs
+++ b/module/applications/activity/enchant-sheet.mjs
@@ -130,7 +130,8 @@ export default class EnchantSheet extends ActivitySheet {
       type: "enchantment",
       name: name || this.item.name,
       img: img || this.item.img,
-      disabled: true
+      disabled: true,
+      transfer: false
     };
   }
 }
--- a/module/data/active-effect/enchantment.mjs
+++ b/module/data/active-effect/enchantment.mjs
@@ -67,7 +67,9 @@ export default class EnchantmentData extends ActiveEffectDataModel {
   /** @inheritDoc */
   prepareDerivedData() {
     super.prepareDerivedData();
-    if ( this.isApplied && this.parent.uuid ) dnd5e.registry.enchantments.track(this.parent.origin, this.parent.uuid);
+    if ( this.isApplied && this.parent.uuid && this.parent.origin ) {
+      dnd5e.registry.enchantments.track(this.parent.origin, this.parent.uuid);
+    }
   }
 
   /* -------------------------------------------- */
@@ -172,8 +174,9 @@ export default class EnchantmentData extends ActiveEffectDataModel {
       if ( this.isOnActivity ) {
         const transferCheckbox = transferFormGroup.querySelector('dnd5e-checkbox[name="transfer"]');
         if ( transferCheckbox ) {
-          transferCheckbox.dataset.tooltip = "DND5E.ENCHANTMENT.Transfer.DisabledTooltip";
+          if ( app.isEditable ) transferCheckbox.dataset.tooltip = "DND5E.ENCHANTMENT.Transfer.DisabledTooltip";
           transferCheckbox.disabled = true;
+          transferCheckbox.checked = false;
         }
       }
     }
--- a/module/data/active-effect/enchantment.mjs
+++ b/module/data/active-effect/enchantment.mjs
@@ -43,11 +43,11 @@ export default class EnchantmentData extends ActiveEffectDataModel {
   /* -------------------------------------------- */
 
   /**
-   * Has this enchantment been applied by another item, or was it directly created.
+   * Should this enchantment apply to its parent item?
    * @type {boolean}
    */
   get isApplied() {
-    return !!this.parent.origin && (this.parent.origin !== this.item?.uuid);
+    return this.parent.transfer && !!this.item;
   }
 
   /* -------------------------------------------- */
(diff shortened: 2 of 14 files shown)


module/applications/activity/enchant-sheet.mjs
```diff
@@ -130,7 +130,20 @@ export default class EnchantSheet extends ActivitySheet {
...
-      disabled: true
+      disabled: true,
+      transfer: false
     };
   }
+
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  async _processSubmitData(event, submitData) {
```
Fyorl: Can we use the same technique as #5998 and prepare `transfer: false` if on `isOnEnchantActivity`? I think then this goes away, and we also preserve the source `transfer` value which is maybe useful.

module/data/active-effect/enchantment.mjs
```diff
@@ -43,11 +43,22 @@ export default class EnchantmentData extends ActiveEffectDataModel {
   /* -------------------------------------------- */
 
   /**
-   * Has this enchantment been applied by another item, or was it directly created.
+   * Should this enchantment apply to its parent item?
    * @type {boolean}
    */
   get isApplied() {
-    return !!this.parent.origin && (this.parent.origin !== this.item?.uuid);
+    return this.parent.transfer && this.item && !this.isOnEnchantActivity;
```
Fyorl: If we prepare `transfer: false` then we don't also need to check here.
```suggestion
    return this.parent.transfer && !!this.item;
```

module/data/active-effect/enchantment.mjs
Fyorl: Direct enchantments don't have an `origin`. I guess we don't need to track them at all.

module/data/active-effect/enchantment.mjs
```diff
@@ -43,11 +43,22 @@ export default class EnchantmentData extends ActiveEffectDataModel {
...
+  }
+
+  /* -------------------------------------------- */
+
+  /**
+   * Does this enchantment exist as an enchantment profile on an enchant activity?
+   * @type {boolean}
+   */
+  get isOnEnchantActivity() {
+    return this.item
```
Fyorl: Or change the `@returns` annotation to `{boolean|void}`. Up to you.
```suggestion
    return !!this.item
```

module/migration.mjs
```diff
@@ -756,6 +756,7 @@ export function migrateEffectData(effect, migrationData, { bypassVersionCheck, p
   _migrateDocumentIcon(effect, updateData, { ...migrationData, field: "img" });
   _migrateEffectChanges(effect, updateData, { setIds: migrate600 });
   if ( migrate600 ) _migrateEffectMagical(effect, parent, updateData);
+  if ( effect.type === "enchantment" ) _migrateEnchantmentTransfer(effect, parent, updateData);
```
Fyorl: Needs to also check `migrate600` I believe, since otherwise we migrate legitimate direct enchantments if the migration is re-run.

module/migration.mjs
```diff
@@ -1154,6 +1155,21 @@ function _migrateEffectMagical(effect, parent, updateData) {
...
 
+/**
+ * Migrates Enchantment effect's transfer field according to whether they should be applied.
+ * @param {object} effect      Effect data to migrate.
+ * @param {object} parent      The parent of this effect.
+ * @param {object} updateData  Existing update to expand upon.
+ * @returns {object}           The updateData to apply.
+ */
+function _migrateEnchantmentTransfer(effect, parent, updateData) {
+  const origin = effect.flags?.core?.originText ?? effect.origin;
```
Fyorl, line 1186: Is it possible this has already been migrated to `system.origin` by this point?

Fyorl (changes requested):
In `EnchantActivity#applyEnchantment` we do `effect.clone()` and nothing is setting `transfer: true, disabled: false`.

module/applications/activity/enchant-sheet.mjs
roth-michael: Yeah, I think this'll see a rewrite once the other PR is merged.

module/migration.mjs
roth-michael: Guess it can't hurt. It think my reasoning was that at this point, if an enchant is newer than 6.0 it'll already not have `origin` in its source data. But I could be wrong about that.

module/migration.mjs
roth-michael: Shouldn't be, I put it in right before the `_migrateEffectOrigin` call in `migrateEffectData` for just that reason.

roth-michael:
Apologies for the full forced push rather than incremental. Key differences from the previous review:
- I moved the (previously nonexistent) `prepareBaseData` method and `isOnActivity` getter from `BaseEffectData` to `ActiveEffectDataModel`, since enchantments benefit from this behavior as well
  - Note: Also fixed a bug in that logic: `a.effects` could be undefined for Forward activities
- Because we can now rely on `transfer` being true for applied enchantments, `isApplied` is simplified
- In `EnchantmentData#onRenderActiveEffectConfig`, check `this.isOnActivity` instead of `this.isOnEnchantmentActivity` (that level of specificity seems unnecessary, since enchantments can _only_ be on enchant activities)
- In `EnchantActivity#applyEnchantment`, when cloning the effect, set `disabled: false` and `transfer: true`
- When migrating, only do enchantment transfer migration if `migrate600` (also moved the effect type check to an early return in `_migrateEnchantmentTransfer`, just looked cleaner)

Fyorl (approved):

Fyorl:
Merged manually

Issue #7198, opened 2026-06-29
Update wiki for 6.0 release


arbron:
- [x] Active Effects Guide
  - [x] Replacing formula data (https://github.com/foundryvtt/dnd5e/pull/4907)
  - [x] Changes to `bonuses` (https://github.com/foundryvtt/dnd5e/pull/6145)
  - [x] Multiple AC formula support (https://github.com/foundryvtt/dnd5e/pull/6658)
  - [x] Hiding items using active effects (https://github.com/foundryvtt/dnd5e/pull/6666)
  - [x] Active effects compendium (https://github.com/foundryvtt/dnd5e/pull/6679)
  - [x] Limits when adding or subtracting on numeric fields (https://github.com/foundryvtt/dnd5e/pull/6861)
  - [x] Common token changes, including size (https://github.com/foundryvtt/dnd5e/pull/7033)
  - [x] New active effect sheet (https://github.com/foundryvtt/dnd5e/pull/7159)
  - [x] Conditions on effects & changes (https://github.com/foundryvtt/dnd5e/pull/7183)
  - [x] Rules active effect types (https://github.com/foundryvtt/dnd5e/pull/7187)
  - [x] Ability attack bonuses (https://github.com/foundryvtt/dnd5e/pull/7279)
- [x] Activities
  - [x] Bonus field on check & save activities (https://github.com/foundryvtt/dnd5e/pull/4786)
  - [x] Visibility on check & save activities (https://github.com/foundryvtt/dnd5e/pull/4815)
  - [x] Descriptions on activities (https://github.com/foundryvtt/dnd5e/pull/5379)
  - [x] Area of effect behaviors (https://github.com/foundryvtt/dnd5e/pull/6761)
  - [x] Standalone applied active effects/enchantments (https://github.com/foundryvtt/dnd5e/pull/6800)
  - [x] Modifiers on damage data (https://github.com/foundryvtt/dnd5e/pull/6834)
  - [x] Active effect durations set by activity duration (https://github.com/foundryvtt/dnd5e/pull/6870)
  - [x] Teleport activity (https://github.com/foundryvtt/dnd5e/pull/6904)
  - [x] Ability selection in attack dialogs (https://github.com/foundryvtt/dnd5e/pull/7080)
  - [x] Expiry events in duration (https://github.com/foundryvtt/dnd5e/pull/7267)
- [x] Advancement
[...]

PR #7323, opened 2026-08-10
[#7198] Add documentation for AE conditions & filters

arbron:
Adds a new Active Effects Overview page that provides general info on AEs and links to the other active effect documents. Added a new page describing the Rules change types and how they work with several examples and a Filters page describing how filters are constructed.


wiki/Active-Effect-Rules.md
```diff
@@ -0,0 +1,160 @@
...
+
+This also accepts deterministic formulas, so you can enter a minimum of `@prof` and that will resolve to the proficiency modifier of the person rolling the die.
+
+
+## Conditions
+
+Rules are extremely broad by default. Adding a damage bonus rule means that that bonus is applied to all damage, whether it be from a spell or weapon, melee or ranged, and of any type. To narrow down when a rule is applied, effect & change conditions can be used.
+
+The [active effect conditions page](Active-Effects.md#Conditions) goes over the details on how conditions are defined and what data can be generally filtered on. Of special importance to rules is the rolling description fields which provide additional information that can be used to narrow down when a rule should be applied (e.g. using the `roll.damageType` to only apply a bonus when the character does fire damage, or `roll.attack.mode` to apply a bonus only to thrown weapon attacks).
```
krbz999:
```suggestion
The [active effect conditions page](Active-Effects.md#Conditions) goes over the details on how conditions are defined and what data can be generally filtered on. Of special importance to rules is the rolling description fields which provide additional information that can be used to narrow down when a rule should be applied (e.g. using the `roll.damage.type` to only apply a bonus when the character does fire damage, or `roll.attack.mode` to apply a bonus only to thrown weapon attacks).
```

wiki/Active-Effect-Rules.md
```diff
@@ -0,0 +1,160 @@
+![Up to date as of 6.0.0](https://img.shields.io/static/v1?label=dnd5e&message=6.0.0&color=informational)
+
+Rules are a special type of active effect changes provided by the system. Rather than being applied directly to data on an actor or item, these rules are evaluated at roll time an modify only a specific roll.
```
Fyorl:
```suggestion
Rules are a special type of active effect changes provided by the system. Rather than being applied directly to data on an actor or item, these rules are evaluated at roll time and modify only a specific roll.
```

wiki/Active-Effects.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Rule Change Types
+
+The system provides a set of special change types called rules. Rather than being applied directly to data on an actor or item, these rules are evaluated at roll time an modify only a specific roll.
+
+More information on these change types can be found on the [active effect rules page](Active-Effect-Rules.md).
+
+## Change Replacement
+
+When active effects are applied to an actor or item, their change values are left unmodified by default. The *Replacement* option can be used to modify the values, replacing attribute strings (e.g. `@abilities.wis.mod`) with a final value from either the actor applying the effect (using the "Origin" option) or the actor to which the effect is being applied (using the "Target" option).
```
Fyorl: Not sure of exact wording but probably need to make it clear that leaving them un-replaced means they are dynamic, while using replacement means they are snapshotted at the moment the effect is applied.

wiki/Filters.md
```diff
@@ -0,0 +1,118 @@
...
+      "o": "has",
+      "v": "thr"
+    }
+  ]
+}
+```
+
+## Operators & Comparisons
+
+The `o` property in a filter control how an individual filter is evaluated. They come in two varieties: comparisons and operators. If not operator is provided for a filter, then the `exact` comparison operator is used.
```
Fyorl:
```suggestion
The `o` property in a filter control how an individual filter is evaluated. They come in two varieties: comparisons and operators. If an operator is not provided for a filter, then the `exact` comparison operator is used.
```

Fyorl (approved):

wiki/Filters.md
```diff
@@ -0,0 +1,118 @@
...
+
+When used in an enchantment, this applies to any weapon that is ranged or has the thrown property:
+
+```json
+{
+  "o": "OR",
+  "v": [
+    {
+      "k": "item.type.value",
+      "v": "ranged"
```
Fyorl:
```suggestion
      "k": "item.type.value",
      "o": "in",
      "v": ["simpleR", "martialR"]
```

Issue #7131, opened 2026-06-13
Adding embedding support for activities and advancement pseudo-documents


arbron:
Activities should be able to embed their description with an optional link for using the activity.

Advancement should be able to embed their hints with additional per-type support, such as listing the description of each item in an `ItemChoiceAdvancement` pool.

- [x] Activity embeds
- [ ] Advancement embeds

PR #7324, opened 2026-08-10
[#7131] Add embedding support for activities

arbron:
Adds embedding methods in activities by copying the methods over from `Actor`. Because of citations this also required copying over methods like `toAnchor` add `_createDocumentLink`. This does mean that Advancement and Activity documents can now be linked like all other document types.

These copied methods also required adding `CONFIG.Activity` and `CONFIG.Advancement` so that core could fetch the document class, icon, embed handlers, and type labels.

Activties have gained a `_buildEmbedHTML` method so they will show their descriptions when embedding. No embed support has been done for Advancement documents yet.

### DIFF
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -106,7 +106,7 @@ Hooks.once("init", function() {
     sidebarIcon: "fa-solid fa-person-hiking",
     typeLabels: new Proxy(CONFIG.DND5E.activityTypes, {
       get(target, prop) {
-        return target[prop]?.documentClass?.metadata?.title
+        return target[prop]?.documentClass?.metadata?.title;
       }
     })
   };
@@ -116,7 +116,7 @@ Hooks.once("init", function() {
     sidebarIcon: "fa-solid fa-person-dress-burst",
     typeLabels: new Proxy(CONFIG.DND5E.advancementTypes, {
       get(target, prop) {
-        return target[prop]?.documentClass?.metadata?.title
+        return target[prop]?.documentClass?.metadata?.title;
       }
     })
   };
--- a/dnd5e.mjs
+++ b/dnd5e.mjs
@@ -99,6 +99,28 @@ Hooks.once("init", function() {
   CONFIG.ui.items = applications.item.ItemDirectory5e;
   CONFIG.ux.DragDrop = DragDrop5e;
 
+  // Add configuration sections for activities & advancement
+  CONFIG.Activity = {
+    documentClass: documents.activity.UtilityActivity,
+    embedHandlers: [],
+    sidebarIcon: "fa-solid fa-person-hiking",
+    typeLabels: new Proxy(CONFIG.DND5E.activityTypes, {
+      get(target, prop) {
+        return target[prop]?.documentClass?.metadata?.title
+      }
+    })
+  };
+  CONFIG.Advancement = {
+    documentClass: documents.advancement.Advancement,
+    embedHandlers: [],
+    sidebarIcon: "fa-solid fa-person-dress-burst",
+    typeLabels: new Proxy(CONFIG.DND5E.advancementTypes, {
+      get(target, prop) {
+        return target[prop]?.documentClass?.metadata?.title
+      }
+    })
+  };
+
   // Register System Settings
   registerSystemSettings();
   registerSystemKeybindings();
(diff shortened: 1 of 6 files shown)


dnd5e.mjs
```diff
@@ -99,6 +99,28 @@ Hooks.once("init", function() {
...
   CONFIG.ux.DragDrop = DragDrop5e;
 
+  // Add configuration sections for activities & advancement
+  CONFIG.Activity = {
+    documentClass: documents.activity.UtilityActivity,
+    embedHandlers: [],
+    sidebarIcon: "fa-solid fa-person-hiking",
+    typeLabels: new Proxy(CONFIG.DND5E.activityTypes, {
+      get(target, prop) {
+        return target[prop]?.documentClass?.metadata?.title
```
Fyorl:
```suggestion
        return target[prop]?.documentClass?.metadata?.title;
```

dnd5e.mjs
```diff
@@ -99,6 +99,28 @@ Hooks.once("init", function() {
...
+      }
+    })
+  };
+  CONFIG.Advancement = {
+    documentClass: documents.advancement.Advancement,
+    embedHandlers: [],
+    sidebarIcon: "fa-solid fa-person-dress-burst",
+    typeLabels: new Proxy(CONFIG.DND5E.advancementTypes, {
+      get(target, prop) {
+        return target[prop]?.documentClass?.metadata?.title
```
Fyorl:
```suggestion
        return target[prop]?.documentClass?.metadata?.title;
```

module/documents/activity/mixin.mjs
```diff
@@ -1174,6 +1174,21 @@ export default function ActivityMixin(Base) {
...
 
+    /* -------------------------------------------- */
+    /*  Helpers                                     */
+    /* -------------------------------------------- */
+
+    /** @override */
+    async _buildEmbedHTML(config, options={}) {
+      if ( !this.description?.value ) return null;
+      const enriched = await foundry.applications.ux.TextEditor.implementation.enrichHTML(this.description.value, {
+        rollData: this.item.getRollData(), ...options, relativeTo: this.item
```
Fyorl: Should this be `...options, rollData: this.item.getRollData()`? If you embed an activity from item A into item B will resolve A's description against B's `rollData`.

Fyorl (changes requested):
I don't think that continuing to build out more scaffolding that makes pseudo-documents pretend to be more documents is advisable. Eventually you will hit a wall because pseudo documents cannot behave like real documents at some fundamental levels. I think it's preferable that the wall be encountered sooner rather than later, and that, when working with them, it is clear that you are working with plain objects rather than real documents, and that the developer understands the costs and limitations associated with that.

But the issue is probably that core code is too inflexible here and makes assumptions about Documents that it doesn't need to. `@Embed[]` is a generally useful framework that could probably apply to arbitrary `DataModel`s rather than just `Document`s. I don't really see any alternative without core changes then. We could use custom enrichers like `[[activity]]` or `[[advancement]]` but then we have to copy most of the `@Embed[]` stack, and we don't get drag-drop into ProseMirror without extra effort, etc.

Some changes not on the diff:

1. `ItemSheet5e#_onDropActivity` needs to resolve `fromDropData` rather than reading the ID and source off the data, since that isn't guaranteed to exist anymore.
2. `PseudoDocument` needs a `pack()` getter, otherwise activity embeds inside compendium items throw on drag.

```js
get pack() {
  return this.item?.pack ?? null
}
```

module/documents/mixins/pseudo-document.mjs
```diff
@@ -141,6 +149,26 @@ export default function PseudoDocumentMixin(Base) {
 
     /* -------------------------------------------- */
 
+    /**
+     * Return a string which creates a dynamic link to this PseudoDocument instance.
+     * @returns {string}
+     */
+    get link() {
+      return `@UUID[${this.uuid}]{${foundry.utils.escapeHTML(this.name ?? this.title)}}`;
```
Fyorl: The copied `toAnchor` doesn't account for this title vs. name discrepancy we have in `BaseAdvancementData`. Probably the easiest is to just alias it with a `name` getter on `Advancement`. Then we also don't need `?? this.title` here.

Fyorl (approved):

module/documents/mixins/pseudo-document.mjs
arbron: I'll add the getter for now, but since I've been wanting to migrate `title` to `name` and `icon` to `img` in advancement documents anyway I'll look in to doing that this release.

PR #7327, opened 2026-08-10
Refactor chat cards around compact design. Add dark mode theming.

### LINKED ISSUE #5014: Feature Request: An option for Activity names and Icons to appear in Item Cards instead of the base item's

Fyorl:
Sorry for large diff but 1/3 of it is CSS so hopefully not too arduous to review.

Summary of changes in this PR:
- Enable dark mode chat log.
- Add theming to v2 chat card styles.
- Include activity name, icon, & description in chat cards.
- New compact style for activity usage and item display chat cards.
- Collapses multiple `rollSave` and `rollCheck` into singular 'group' buttons that provide a drop-down of options on click.

image

The following not attempted here but still TODO:
- Restyle effects tray (effects tray just entirely removed in this PR)
- Handling of progressive appending for additional cards spawned by the usage card

### DIFF
--- a/less/v3/chat.less
+++ b/less/v3/chat.less
@@ -1,0 +1,197 @@
+/* ---------------------------------- */
+/*  Chat Cards                        */
+/* ---------------------------------- */
+
+.chat-log, .chat-popout {
+  .message.compact {
+    background: var(--dnd5e-chat-background);
+    border: var(--dnd5e-border-gold);
+    border-block-end-width: 2px;
+    border-radius: 6px;
+    color: var(--color-text-primary);
+    font-family: var(--dnd5e-font-roboto);
+    font-size: var(--font-size-13);
+    padding: 8px;
+
+    &.blind, &.emote, &.whisper { border-block-end-width: 3px; }
+    &.emote { border-block-end-color: #e5af29; }
+    &.whisper { border-block-end-color: #7d7df1; }
+    &.blind { border-block-end-color: #c12fc1; }
+
+    .message-content {
+      color: var(--color-text-primary);
+      overflow: visible;
+    }
+
+    .message-delete { display: none; }
+
+    .message-header {
+      align-items: center;
+      color: var(--color-text-primary);
+      gap: 8px;
+      margin-block-end: 4px;
+    }
+
+    .message-metadata {
+      color: var(--color-text-secondary);
+      flex: none;
+      font-size: var(--font-size-10);
+    }
+
+    .message-sender {
+      flex: 1;
+      font-family: var(--dnd5e-font-roboto-slab);
+      font-size: var(--font-size-14);
+      font-weight: bold;
+      font-variant: small-caps;
+
+      &:hover {
+        cursor: var(--cursor-pointer);
+        text-shadow: 0 0 8px var(--color-shadow-primary);
+      }
+    }
+
+    .chat-card {
+      display: flex;
+      flex-direction: column;
+      gap: 4px;
+    }
+
+    .card-header {
+      --icon-size: 24px;
+      display: flex;
+      gap: 4px;
+
+      &[data-action] { cursor: var(--cursor-pointer); }
+
+      .activity-icon, .item-icon {
+        --size: 24px;
+
+        dnd5e-icon, img {
+          --icon-fill: var(--activity-icon-color);
+          --gold-icon-size: var(--size);
+          max-width: unset;
+        }
+      }
+
+      .activity-icon {
+        block-size: var(--size);
+        inline-size: var(--size);
+
+        &:has(> dnd5e-icon) {
+          border: 1px solid var(--dnd5e-color-gold);
+          border-radius: 4px;
+          padding: 2px;
+        }
+
+        dnd5e-icon { --icon-size: calc(var(--size) - 6px); }
+      }
+
+      .item-icon:hover { box-shadow: 0 0 6px var(--dnd5e-color-gold); }
+
+      .name-stacked {
+        flex: 1;
+        justify-content: space-between;
+        line-height: 1;
+
+        .title { font-size: var(--font-size-12); }
+      }
+
+      .chevron {
+        block-size: 24px;
+        font-size: var(--font-size-11);
+        inline-size: 24px;
+        place-content: center;
+
+        .fa-chevron-down { transition: transform 250ms ease; }
+      }
+
+      &.collapsed .chevron .fa-chevron-down { transform: rotate(-90deg); }
+    }
+
+    .card-flavor {
+      color: var(--color-text-secondary);
+      font-family: var(--dnd5e-font-roboto-condensed);
+      font-size: var(--font-size-11);
+      font-style: italic;
+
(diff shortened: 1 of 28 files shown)


module/documents/chat-message.mjs
```diff
@@ -246,10 +210,77 @@ export default class ChatMessage5e extends ChatMessage {
...
+
+  /* -------------------------------------------- */
+
+  /**
+   * Enrich chat card header matter.
+   * @param {HTMLElement} html  The chat card markup.
+   * @param {ChatMessageEnrichmentOptions} options
+   * @protected
+   */
+  async _enrichHeader(html, options={}) {
```
Fyorl, line 223: This code is mostly the same, just moved, and some slight tweaks to optionally render the avatar.

less/v3/chat.less
```diff
@@ -0,0 +1,197 @@
...
+    border-block-end-width: 2px;
+    border-radius: 6px;
+    color: var(--color-text-primary);
+    font-family: var(--dnd5e-font-roboto);
+    font-size: var(--font-size-13);
+    padding: 8px;
+
+    &.blind, &.emote, &.whisper { border-block-end-width: 3px; }
+    &.emote { border-block-end-color: #e5af29; }
+    &.whisper { border-block-end-color: #7d7df1; }
```
arbron: Don't feel like this reads quite as well as the blind color.

Image

Image

I did some fiddling in OKLCH to get a more visually-similar color and got this:

Image

Image


```suggestion
    &.whisper { border-block-end-color: #6b5cff; }
```

less/v3/chat.less
```diff
@@ -0,0 +1,197 @@
...
+
+    .card-description {
+      background: var(--dnd5e-chat-description-background);
+      border-block: 1px dashed var(--dnd5e-color-blue-gray-3);
+      font-size: var(--font-size-11);
+      margin-inline: -8px;
+      padding-inline: 16px;
+      text-align: justify;
+
+      &.collapsed { border: none; }
```
arbron, line 132: The transition on the border feels sudden. We could add a transition and modify `border-color` to make it smoother.

```suggestion
      transition: border-color 250ms ease;

      &.collapsed { border-color: transparent; }
```

module/documents/activity/attack.mjs
```diff
@@ -50,7 +50,7 @@ export default class AttackActivity extends ActivityMixin(BaseAttackActivityData
   _usageChatButtons(message) {
     const buttons = [{
       action: "rollAttack",
-      icon: "systems/dnd5e/icons/svg/trait-weapon-proficiencies.svg",
+      icon: "fa-solid fa-sword",
```
arbron, line 53: This is a pro icon, so potential license issue if anyone else updates this code :(

less/v3/chat.less
```diff
@@ -0,0 +1,197 @@
...
+          max-width: unset;
+        }
+      }
+
+      .activity-icon {
+        block-size: var(--size);
+        inline-size: var(--size);
+
+        &:has(> dnd5e-icon) {
+          border: 1px solid var(--dnd5e-color-gold);
```
arbron: Might be nice to leave out the border for SVG icons:

Image

That's a bit more in line with how we treat them on sheets.


```suggestion
          border: 1px solid transparent;
```

less/v3/chat.less
```diff
@@ -0,0 +1,197 @@
...
+
+    .message-sender {
+      flex: 1;
+      font-family: var(--dnd5e-font-roboto-slab);
+      font-size: var(--font-size-14);
+      font-weight: bold;
+      font-variant: small-caps;
+
+      &:hover {
+        cursor: var(--cursor-pointer);
```
arbron:
```suggestion
      cursor: var(--cursor-pointer);

      &:hover {
```

templates/chat/parts/card-face.hbs
```diff
@@ -1,55 +1,79 @@
...
-        <li class="pill transparent">
-            <span class="label">{{ this }}</span>
-        </li>
-        {{/each}}
-    </ul>
+    {{!-- Rows --}}
+    {{#each rows}}
+    {{#if entries.length}}
+    <section class="icon-row">
+        <i class="fa-fw {{ icon }}" inert></i>
```
arbron: Rather than making this fully inert, it would probably be good to add an `aria-label` to the rows to communicate their contents the same way the icon does.

arbron (changes requested):
Testing out dark mode in chat cards leads me to believe I definitely want a separate theme selector for them. I'd like my cards to match the light mode styling of my sheets, rather than the dark mode styling of my UI.

less/v3/chat.less
Fyorl: Without `border: none:` we end up with a 2px gap when the description is collapsed (more visible in dark mode):

image

module/documents/activity/attack.mjs
Fyorl: It's not the first pro icon in the system. I am licensed to author them, as is everyone with merge permissions except you maybe. If you want I can see about getting you a seat?

less/v3/chat.less
arbron: Could also transition a small negative margin to compensate for the border's width. Not hugely noticeable I suppose, so we could just skip it to keep things simple.

module/documents/activity/attack.mjs
arbron: I mean I wouldn't object :). If you think it will be okay then that's fine.

arbron (changes requested):
Noticed that in dark mode the additional options control/delete button and the flavor text on roll cards are significantly darker than on the other cards.

Image

Fyorl:
> Testing out dark mode in chat cards leads me to believe I definitely want a separate theme selector for them. I'd like my cards to match the light mode styling of my sheets, rather than the dark mode styling of my UI.

Forced light mode on the chat log has actually lasted longer than it was supposed to. I think it was supposed to be removed in v14 but we forget. I would imagine that when it's removed there might be a question of whether we add a separate theme option for the chat log.

If you want I could implement something in the system, it's not too much trouble. But normally this sort of customisation would go to modules.

Fyorl:
> If you want I could implement something in the system, it's not too much trouble. But normally this sort of customisation would go to modules.

I changed my mind on this. Since we are using system code to adjust the chat log theme I suppose there's some argument that it should be a system setting so I have added one.

arbron (approved):

PR #7329, opened 2026-08-11
Small fixes for statuses

krbz999:
- Better handling for statuses registered with eg `levels: undefined`.
- Fixes an issue where trying to increase Exhaustion from 6 triggered an update and scrolling status text despite no change made.

### DIFF
--- a/module/data/active-effect/condition.mjs
+++ b/module/data/active-effect/condition.mjs
@@ -21,7 +21,7 @@ export default class ConditionData extends foundry.data.ActiveEffectTypeDataMode
    */
   static hasLevels(type) {
     const config = CONFIG.DND5E.conditionTypes[type];
-    return !!config && ("levels" in config);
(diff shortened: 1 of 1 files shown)


module/data/active-effect/condition.mjs
```diff
@@ -21,7 +21,7 @@ export default class ConditionData extends foundry.data.ActiveEffectTypeDataMode
    */
   static hasLevels(type) {
     const config = CONFIG.DND5E.conditionTypes[type];
-    return !!config && ("levels" in config);
+    return Number.isFinite(config?.levels);
```
JPMeehan, line 24: Are `Infinite` levels a problem? Could this be Number.isNumeric?

krbz999: Yes, because no storage device can store an infinite number of assets for an infinite number of levels of a condition.

Fyorl (approved):

PR #7341, opened 2026-08-14
Add compact styling to effects tray.

Fyorl:
Slightly awkward in-between state. I think the damage tray is probably not functional here, but I will hopefully get to that soon-ish.

There are some functionality changes here in addition to style changes:
- Addition of the `TargetPillElement` which encapsulates logic around grouping multiple targets and offering a context menu for individually toggling members within that group.
- Multiple effects can be applied at once now, rather than having to be applied one-at-a-time.

I also added the icon + label compact style for the various special expiry labels to the actor/item sheets where it was currently taking up a little too much space. The AE tooltips retain the full text though.

image

### DIFF
--- a/less/v2/apps.less
+++ b/less/v2/apps.less
@@ -415,37 +415,37 @@
     display: flex;
     flex-wrap: wrap;
     gap: .25rem;
+  }
 
-    .pill {
-      font-family: var(--dnd5e-font-roboto-condensed);
-      font-weight: 400;
-      font-size: var(--font-size-11);
-      line-height: 1;
-      text-transform: none;
-      border: var(--pill-border);
-      border-radius: 3px;
-      background: var(--dnd5e-background-card);
-      padding: 2px 5px;
-      display: flex;
-      gap: .25rem;
-      margin: 0;
+  .pill {
+    font-family: var(--dnd5e-font-roboto-condensed);
+    font-weight: 400;
+    font-size: var(--font-size-11);
+    line-height: 1;
+    text-transform: none;
+    border: var(--pill-border);
+    border-radius: 3px;
+    background: var(--dnd5e-background-card);
+    padding: 2px 5px;
+    display: flex;
+    gap: .25rem;
+    margin: 0;
 
-      &.green, &.maroon {
-        color: var(--pill-colored-text-color);
-        border: var(--pill-colored-border);
-      }
-
-      &.green { background-color: var(--pill-background-green); }
-      &.maroon { background-color: var(--pill-background-maroon); }
-
-      &.transparent {
-        color: var(--pill-transparent-color);
-        border: var(--pill-border-dotted);
-        background-color: transparent;
-      }
-
-      > .separator { color: var(--color-text-tertiary); }
+    &.green, &.maroon {
+      color: var(--pill-colored-text-color);
+      border: var(--pill-colored-border);
     }
+
+    &.green { background-color: var(--pill-background-green); }
+    &.maroon { background-color: var(--pill-background-maroon); }
+
+    &.transparent {
+      color: var(--pill-transparent-color);
+      border: var(--pill-border-dotted);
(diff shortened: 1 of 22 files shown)


less/v2/apps.less
```diff
@@ -415,37 +415,37 @@
...
-      padding: 2px 5px;
-      display: flex;
-      gap: .25rem;
-      margin: 0;
-
-      &.green, &.maroon {
-        color: var(--pill-colored-text-color);
-        border: var(--pill-colored-border);
-      }
+  .pill {
```
Fyorl, line 420: Not changed, just hoisted out of `.pills`.

module/documents/combatant.mjs
```diff
@@ -90,18 +90,6 @@ export default class Combatant5e extends Combatant {
 
   /* -------------------------------------------- */
 
-  /**
-   * Key identifying a unique set of actors in the combat, optionally prefixed by a grouping discriminator.
-   * @param {string} [prefix]  Discriminator placed at the head of the key.
-   * @returns {string|null}
-   */
-  getUniqueKey(prefix) {
```
Fyorl, line 98: This method was just calling `this.token` for everything, so really it belongs on the `TokenDocument`, which also means we can use the same name `getGroupingKey` instead of `getUniqueKey`, because it's kind of the opposite of a unique key.

less/v3/chat.less
```diff
@@ -187,9 +187,125 @@
...
+      }
+    }
+  }
+
+  .target.none {
+    border: 1px solid var(--color-level-error-border);
+    color: var(--color-level-error);
+  }
+
+  .icon-row > .target-source-toggle {
```
arbron: Not applying due to specificity.


```suggestion
  .icon-row > button.target-source-toggle {
```

less/v3/chat.less
```diff
@@ -187,9 +187,125 @@
...
+    block-size: unset;
```
arbron, line 291: This button can shift out from under your mouse if you toggle to a mode with many options, perhaps we could stretch the hit area to the full height to avoid that.


```suggestion
    align-self: stretch;
    block-size: unset;
```

module/applications/components/effect-application.mjs
```diff
@@ -131,52 +169,115 @@ export default class EffectApplicationElement extends TargetedApplicationMixin(C
...
+        <span class="title"></span>
+        <div class="duration pill transparent">
+          <i class="fa-solid ${icon}" inert></i>
+          <span>${label}</span>
         </div>
-        <button class="apply-effect" type="button" data-action="applyEffect"
-                data-tooltip aria-label="${_loc("DND5E.EFFECT.Application.Action.ApplyTokens")}">
-          <i class="fas fa-reply-all fa-flip-horizontal" inert></i>
-        </button>
+        <span class="pip" aria-pressed="${this.effectChecked(effect.uuid)}"></span>
```
arbron: Since we are making the whole row clickable, wondering if would make more sense for `aria-pressed` to be on the row. I also don't think `aria-pressed` does anything for elements that don't have `roll="button"`. Perhaps we should nest the whole row's contents into a `<button>` so we get all the necessary accessibility from that.

module/applications/components/target-menu.mjs
```diff
@@ -0,0 +1,26 @@
+/**
+ * A FilterMenu implementation for listing a group of targets.
+ */
+export default class TargetMenu extends foundry.applications.ux.FilterMenu {
```
arbron, line 4: To be used in the future?

arbron (changes requested):

module/applications/components/target-menu.mjs
Fyorl: It's used in `TargetPillElement`

arbron (approved):

arbron (approved):

PR #7349, opened 2026-08-17
Add tooltips to activities

arbron:
Adds tooltips to activities similar to items and effects. These will display the item icon alongside the activity icon, uses, description, and usage properties.

For cast activites, if no activity description is provided then it will display the spell's description. This also fixes some bugs where cast activites weren't properly getting the activation, duration, range, and target from the spell.

### DIFF
--- a/less/v2/tooltips.less
+++ b/less/v2/tooltips.less
@@ -56,7 +56,7 @@
   /*  Item Tooltips                     */
   /* ---------------------------------- */
 
-  &:is(.activity-tooltip, .effect-tooltip, .item-tooltip) {
+  &.document-tooltip {
     width: 300px;
 
     .loading {
@@ -84,11 +84,6 @@
         .subtitle { font-size: var(--font-size-11); }
       }
 
-      dnd5e-icon {
-        --icon-fill: var(--activity-icon-color);
-        --icon-size: var(--header-image-size);
-      }
-
       img {
         width: var(--header-image-size);
         height: var(--header-image-size);
@@ -103,6 +98,11 @@
         display: flex;
         gap: 8px;
         --header-image-size: 48px;
+
+        dnd5e-icon {
+          --icon-fill: var(--activity-icon-color);
+          --icon-size: var(--header-image-size);
+        }
       }
 
       .bottom {
--- a/less/v2/tooltips.less
+++ b/less/v2/tooltips.less
@@ -56,7 +56,7 @@
   /*  Item Tooltips                     */
   /* ---------------------------------- */
 
-  &:is(.item-tooltip, .effect-tooltip) {
+  &:is(.activity-tooltip, .effect-tooltip, .item-tooltip) {
     width: 300px;
 
     .loading {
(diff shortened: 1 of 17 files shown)


less/v2/tooltips.less
```diff
@@ -83,16 +84,27 @@
...
         box-shadow: 0 0 4px var(--dnd5e-shadow-45);
         border-radius: 8px;
         object-fit: cover;
         background: var(--dnd5e-background-alt-1);
       }
 
+      .images {
+        display: flex;
+        gap: 8px;
+        --header-image-size: 48px;
```
Fyorl, line 100: Move the `dnd5e-icon` styles from above under here, otherwise we hit the spell school icon too.

```suggestion
        --header-image-size: 48px;
        
        dnd5e-icon {
          --icon-fill: var(--activity-icon-color);
          --icon-size: var(--header-image-size);
        }
```

less/v2/tooltips.less
```diff
@@ -56,7 +56,7 @@
   /*  Item Tooltips                     */
   /* ---------------------------------- */
 
-  &:is(.item-tooltip, .effect-tooltip) {
+  &:is(.activity-tooltip, .effect-tooltip, .item-tooltip) {
```
Fyorl: Note we only check for `.item-tooltip` for the overflowing logic in `Tooltips5e#_positionItemTooltip`. Activity content links only get `.activity-tooltip` and not `.item-tooltip` as well. I think `.effect-tooltip` might have the same problem.

module/data/activity/base-activity.mjs
```diff
@@ -939,8 +978,9 @@ export default class BaseActivityData extends foundry.abstract.DataModel {
       configurable: true,
       enumerable: false
     });
-    if ( obj.canOverride && !obj.override && !this.isRider ) {
+    if ( obj.canOverride && !obj.override && !obj.overrideSet && !this.isRider ) {
       foundry.utils.mergeObject(obj, foundry.utils.getProperty(item.system, keyPath));
+      obj.overrideSet = true;
```
Fyorl:
```suggestion
      Object.defineProperty(obj, "overrideSet", { value: true, configurable: true, enumerable: false });
```

module/data/activity/cast-data.mjs
```diff
@@ -85,4 +86,23 @@ export default class BaseCastActivityData extends BaseActivityData {
...
+  async getTooltipData({ extras, ...enrichmentOptions }={}) {
+    const context = await super.getTooltipData({ extras, ...enrichmentOptions });
+
+    if ( !context.description ) {
+      const cachedSpell = this.cachedSpell;
+      context.description = await TextEditor.enrichHTML(
+        cachedSpell?.system.description.value || "",
+        { ...enrichmentOptions, relativeTo: cachedSpell }
+      );
+    }
```
Fyorl:
```suggestion
    const cachedSpell = this.cachedSpell;
    if ( !context.description && cachedSpell ) {
      context.description = await TextEditor.enrichHTML(cachedSpell?.system.description.value || "", {
        ...enrichmentOptions, relativeTo: cachedSpell, rollData: cachedSpell.getRollData()
      });
    }
```

Fyorl (changes requested):

Fyorl (approved):

PR #7352, opened 2026-08-18
Add compact styling for attack roll cards.

Fyorl:
I guess `chat/ui/effects` is a de facto staging branch. I'd prefer to merge them all into a working state rather than leaving the damage tray semi-broken.

image

image

Hopefully reasonably self-explanatory from the screenshots. The roll breakdown is a popover now. Target interaction handled by the target-pills themselves rather than externally.

I did add a small bit of extra functionality: Choosing 'select hit/missed targets' will open the effects tray of the originating message (if it has one) and flip it to select mode. Makes it slightly easier to apply on-hit effects when using separated cards.


thatlonelybugbear:
Judging from those screenshots only, I think that the red "Krusk 24" badge blends too much into the dark card background and is somewhat difficult to distinguish at a glance. Noting this as an accessibility concern.

less/v3/chat.less
```diff
@@ -184,17 +191,127 @@
...
+    /* Rolls */
+    button.dice-roll {
+      align-items: center;
+      block-size: unset;
+      flex: 1;
+      flex-direction: row;
+      padding: 0 8px;
+      position: relative;
+
+      &:hover:not(:disabled) { box-shadow: none; }
```
arbron: The box shadow reappears when you click on the button and then move your mouse away.

```suggestion
      &:hover:not(:disabled), &:focus { box-shadow: none; }
```

less/elements.less
```diff
@@ -365,4 +367,25 @@ prose-mirror {
...
+
+  &[data-hit]::before {
+    color: var(--color-level-success);
+    content: "\f00c";
+  }
+
+  &[data-miss] {
+    border: 1px solid var(--color-level-error-border);
+    color: var(--color-level-error);
+    &::before { content: "\f00d"; }
```
arbron: I agree that the miss pills are hard to read in dark mode. Played around a bit and got this:

Image

Image

```suggestion
    background-color: #591010;
    color: var(--pill-colored-text-color);
    &::before {
      content: "\f00d";
      color: var(--color-level-error);
    }
```

less/v3/chat.less
```diff
@@ -184,17 +191,127 @@
...
+        .icons, .roll { color: var(--dnd5e-color-success); }
+      }
+
+      &.success:not(.fumble) .icons { color: var(--dnd5e-color-success); }
+
+      &.fumble {
+        .icons, .roll { color: var(--dnd5e-color-failure); }
+      }
+
+      &.failure:not(.critical) .icons { color: var(--dnd5e-color-failure); }
```
arbron: Both the success and failure colors here look fine in light mode, but are hard to parse in dark mode.

module/applications/components/target-pill.mjs
```diff
@@ -1,4 +1,5 @@
 import TargetMenu from "./target-menu.mjs";
+import TargetsField from "../../data/chat-message/fields/targets-field.mjs";
```
arbron:
```suggestion
import TargetsField from "../../data/chat-message/fields/targets-field.mjs";
import TargetMenu from "./target-menu.mjs";
```

arbron (changes requested):

arbron (approved):

PR #7354, opened 2026-08-18
Enable recursive roll formula replacement

arbron:
Sets `options.recursive` to `true` by default when replacing formula data on `BasicRoll`.

Also rewrites `dnd5e.utils.replaceFormulaData` to also accept the recursive option which defaults to `true`.

### DIFF
--- a/module/dice/basic-roll.mjs
+++ b/module/dice/basic-roll.mjs
@@ -52,10 +52,7 @@ export default class BasicRoll extends Roll {
     for ( const [key, value] of Object.entries(parts) ) {
       if ( !value && (value !== 0) ) continue;
       finalParts.push(`@${key}`);
-      foundry.utils.setProperty(
-        data, key, foundry.utils.getType(value) === "string"
-          ? BasicRoll.replaceFormulaData(value, data, { missing: 0 }) : value
-      );
+      foundry.utils.setProperty(data, key, value);
     }
     return { parts: finalParts, data };
   }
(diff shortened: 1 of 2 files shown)


module/utils.mjs
```diff
@@ -507,21 +510,35 @@ export function prepareFormulaValue(model, keyPath, label, rollData) {
...
+      case "string": {
+        value = value.trim();
+        if ( recursive && (depth < 3) && value.includes("@") ) {
+          return replaceFormulaData(value, data, { ...options, [DEPTH]: depth + 1, [MISSING]: missingReferences });
+        }
+        return value;
+      }
+      case "boolean": return String(Number(value));
+      case "number": return String(value);
+      default: return String(value).trim();
```
Fyorl: These lines are equivalent.
```suggestion
      default: return String(value).trim();
```

module/dice/basic-roll.mjs
```diff
@@ -329,11 +329,12 @@ export default class BasicRoll extends Roll {
...
 
   /** @inheritDoc */
-  static replaceFormulaData(formula, data, options) {
+  static replaceFormulaData(formula, data, options={}, _r=0) {
+    options.recursive ??= true;
     // This looks for the pattern `$!!$` and replaces it with just the value between the marks (the bang has
     // been added to ensure this is a deliberate shim from the system, not a unintentional usage that should
     // show an error).
-    return super.replaceFormulaData(formula, data, options).replaceAll(/\$"?!(.+?)!"?\$/g, "$1");
+    return super.replaceFormulaData(formula, data, options, _r).replaceAll(/\$"?!(.+?)!"?\$/g, "$1");
```
Fyorl: I would think we don't need this anymore. We removed the shim a while ago, and core `replaceFormulaData` honours `toString` on objects now so there is no need for this escape.
```suggestion
    return super.replaceFormulaData(formula, data, options, _r);
```

Fyorl (approved):

module/dice/basic-roll.mjs
Fyorl: I think we don't need this anymore now that we resolve recursively. Pushing to `finalParts` should be all that's necessary and the nested references will be unrolled at parse time.

```suggestion
```

PR #7356, opened 2026-08-19
Assorted chat card polish

Fyorl:
Working on the compact damage card + tray and wanted to consolidate on using popovers, as well as a bunch of other styling tweaks that I made that I split out here.

- Use popover rather than FilterMenu for target-pill additional options
 - Restore missing pan-to-token functionality on target-pill additional options
 - Use anchor-based positioning for popover rather than JS
 - Additional dark & light mode style tweaks for chat card buttons & dice roll total

image


arbron (approved):

Kapuzenjoe:
Hey, after playing around with it a bit, I wanted to share a couple of thoughts on the button style. First, the switch between "targeted" and "selected" is a bit hard to make out now  (it used to be easier to see at a glance, though that might just be me being used to the old way).
Second, I really like how compact the action buttons like Damage Roll or Saving Throw are, but I keep thinking about less technically inclined players who might lose track of what to click. A button that actually says "Damage" is just easier to spot than a bare icon, and I have a feeling a lot of people will find the icon-only version a bit overwhelming or hard to read (the compact style probably raises the chance of misclicks).

nschoenwald:
> Hey, after playing around with it a bit, I wanted to share a couple of thoughts on the button style. First, the switch between "targeted" and "selected" is a bit hard to make out now (it used to be easier to see at a glance, though that might just be me being used to the old way). Second, I really like how compact the action buttons like Damage Roll or Saving Throw are, but I keep thinking about less technically inclined players who might lose track of what to click. A button that actually says "Damage" is just easier to spot than a bare icon, and I have a feeling a lot of people will find the icon-only version a bit overwhelming or hard to read (the compact style probably raises the chance of misclicks).

Agreed, please retain the button labeling (at least Attack / Damage / Save).

PR #7358, opened 2026-08-19
[#6626, #6660] Fix Select Form transform mode, add profile origin

arbron:
Changes how the "Select Form" transformation mode works to create the effect directly on the actor rather than transfering the effect to work with recent changes to how transfer works with applied effects. This also ensures that duration is set properly when using this transformation mode and it is linked to concentration.

Adds a `system.origin.profile` so each effect can know which profile on the activity was used to apply it. This is populated for transformation and enchantment activities, but not yet for base actvities because the profile data isn't fully communicated to the chat cards.

### DIFF
--- a/module/documents/activity/transform.mjs
+++ b/module/documents/activity/transform.mjs
@@ -157,7 +157,7 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
         (e.system.type === "concentrating") && (e.flags.dnd5e?.activity?.id === this.id)
       )?.uuid,
       scaling: config.scaling ? config.scaling : undefined,
-      spellLevel: this.item.level
+      spellLevel: this.item.system.level
     });
   }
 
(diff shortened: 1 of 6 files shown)


module/documents/activity/transform.mjs
```diff
@@ -199,31 +210,58 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
...
+   * @param {object} [flags={}]   Flags to apply to the created effect.
    */
-  async #transformToForm(profileId) {
+  async #transformToForm(profileId, flags={}) {
     if ( this.transform.mode !== "form" ) return;
     const profile = this.applicableEffects.find(e => e._id === profileId);
     if ( !profile && !this.transform.formless ) return;
 
     let targets = getSceneTargets(this.actor, { checkBaseActor: true }).map(t => t.actor);
     if ( !targets.length ) targets.push(this.actor);
```
Fyorl: Should de-dupe now that we are creating new effects. Or maybe `getSceneTargets` should de-dupe? It's pretty rare to have more than one linked token for a given actor on the scene but I suppose it can happen.

```suggestion
let targets = new Set(getSceneTargets(this.actor, { checkBaseActor: true }).map(t => t.actor));
if ( !targets.size ) targets.add(this.actor);
```

module/documents/activity/transform.mjs
```diff
@@ -199,31 +210,58 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
...
+                activity: foundry.utils.buildRelativeUuid(this.uuid, target.uuid),
+                profile: profileId
+              }
+            }
+          };
+
+          // Effect already exists, reset its duration
+          if ( appliedEffect ) operations.push({
+            action: "update", documentName: "ActiveEffect", updates: [foundry.utils.mergeObject({
+              _id: appliedEffect._id, start: ActiveEffect5e.getEffectStart()
```
Fyorl: In case the user manually disabled it for some reason.
```suggestion
              _id: appliedEffect._id, start: ActiveEffect5e.getEffectStart(), disabled: false
```

module/documents/activity/transform.mjs
```diff
@@ -151,7 +152,13 @@ export default class TransformActivity extends ActivityMixin(BaseTransformActivi
     if ( this.transform.mode !== "form" ) return;
     const profile = results.message?.flags?.dnd5e?.transform?.profile
       ?? results.message?.data?.flags?.dnd5e?.transform?.profile;
-    this.#transformToForm(profile);
+    this.#transformToForm(profile, {
+      dependentOn: results.effects?.find(e =>
+        (e.system.type === "concentrating") && (e.flags.dnd5e?.activity?.id === this.id)
+      )?.uuid,
+      scaling: config.scaling ? config.scaling : undefined,
+      spellLevel: this.item.level
```
Fyorl:
```suggestion
      spellLevel: this.item.system.level
```

Fyorl (changes requested):
Do we need to update `BaseTransformActivitData#currentProfile` to check the actor's effects for `origin.profile`?

Fyorl (approved):

PR #7361, opened 2026-08-20
Add compact damage tray styling. Consolidate with effects tray.

Fyorl:
There's a lot of shuffling methods around or factoring method internals out. So the actual size of new code is not that much.

The underlying damage tray still functions pretty much the same as before, there is just a new UI on top of it. Quite difficult to compact the tray and I had to make some concessions. All the original functionality is still available, but a lot of it is buried behind menus now unfortunately.

It takes a couple of extra clicks per target if they are in a group to change an individual target's multiplier now. I am hoping that with some improvements to the saving throw roll coming next, that it shouldn't be necessary to make adjustments to damage targets frequently. I am intending to have the multipliers update automatically based on the activity's 'on save' setting, and also planning to add 'select succeeded' and 'select failed' context menu options like we do with attack cards.

If that's not enough, then it's possible to fall back to selecting groups and changing the global tray multiplier, applying damage, then selecting the next group, etc. That workflow is supported now we have a global tray multiplier and is potentially faster than doing the same with the old tray in some cases.

image

image

image

One advantage is that we don't have to cram the res/imm/vuln stuff into a tiny icon in the subtitle anymore, which should hopefully make things a bit clearer as to what is going on.


module/applications/components/damage-application.mjs
```diff
@@ -263,69 +395,158 @@ export default class DamageApplicationElement extends TargetedApplicationMixin(C
...
+   */
+  #buildTargetPane(pill, pane) {
+    pane.classList.add("damage-pane");
+    const open = this.#paneState.group === pill;
+    const uuids = open && this.#paneState.targets.length ? this.#paneState.targets : pill.targets;
+    const [uuid] = uuids;
+    const token = fromUuidSync(uuid);
+    let { calc, options } = open ? this.#paneState : {};
+    if ( !calc ) {
+      options = this.getMergedOptions(uuid);
```
arbron, line 494: Maybe not worth it because it is generally a very shortly live state, but when you go into a group and change the multiplier of a single of a single member of that group and then go back to the top level of the group, only the multiplier of the first member is represented.

Image

It would be nice if it could show multiple multipliers selected to indicate the mixed state of the group.

Image

This would also make it a bit clearer that clicking a single multiplier in the list will sync all of the group's members back to that multiplier.

module/applications/components/damage-application.mjs
```diff
@@ -55,33 +56,188 @@ export default class DamageApplicationElement extends TargetedApplicationMixin(C
...
-  /*  Rendering                                   */
+
+  /**
+   * Set imm/res/vuln ignore & downgrade.
+   * @param {DamageApplicationOptions} options            Options which adjust damage calculation.
+   * @param {{ change: string, type: string }} operation  The operation being performed.
+   */
+  #cycleChangeSource(options, { type, change }={}) {
+    if ( change === "immunity" ) {
+      if ( options.ignore?.immunity?.has(type) ) {
```
arbron, line 196: Not quite sure where the problem sits, but changing an immunity to downgrade now causes a new resistance to that same damage type to appear that does nothing when clicked:

Image

Image

Image

Image

If the resistance is in the "Applied" state, then changing the immunity back to applied will cause the resistance to disappear. If the resistance is in the "Ignored" state, then it stays around even after changing the immunity.

templates/chat/damage-card.hbs
```diff
@@ -1,30 +1,57 @@
...
+    {{!-- Supplements --}}
+    {{#if onSave}}
+    <p class="supplement"><strong>{{ localize "DND5E.SAVE.OnSave" }}</strong> {{ onSave }}</p>
+    {{/if}}
+
+    {{!-- Rows --}}
+    {{#each rows}}
+    {{#if entries.length}}
+    <section class="icon-row">
+        <i class="fa-fw {{ icon }}" aria-label="{{ localize label }}"></i>
```
arbron, line 30: Notice a lot of alignment issues in these icons despite the `fa-fw` class.

Image

Probably because that class only sets the `width` which doesn't help in the flex context. Setting `flex-basis` helps:

```css
flex: 0 0 var(--fa-width);
```

module/applications/components/target-pill.mjs
```diff
@@ -322,7 +374,8 @@ export default class TargetPillElement extends foundry.applications.elements.Abs
    * Handle toggling the target menu.
    */
   #onToggleMenu() {
-    if ( this.#wasOpen ) this.list.hidePopover();
-    else this.list.showPopover();
+    if ( !this.isConnected ) return;
```
arbron, line 377: Would be good to dismiss any open context menus when the pane is opened to avoid overlaps:

Image

arbron (changes requested):

templates/chat/damage-card.hbs
Fyorl: Are you sure it's not just the target source button that's out of alignment? That one didn't get an explicit width because it's not a `fa-fw`, so I gave it one already in the other PR https://github.com/foundryvtt/dnd5e/pull/7363. That seems to have fixed any alignment issues that I noticed at a glance. But let me know if you think it's still not right.

module/applications/components/damage-application.mjs
Fyorl: Seems this was a bug that already existed in the damage application tray but we never saw it because we didn't re-render the entry when toggling it.

templates/chat/damage-card.hbs
Fyorl: I also added your line to avoid these icons shrinking. Combined with the change in the following PR it should be all good now.

arbron (approved):

PR #7363, opened 2026-08-20
Move base target logic out of effect-application and into its own element

Fyorl:
Something I need for the work on save cards. It should now be much more obvious when you select a bunch of tokens that you are about to roll saving throws for all of them when you click the button. It will also mean you can pre-target the tokens when you roll, and you should be able to roll saves for them in the same way as if they had been selected, with the usual targeted/selected toggle available. I haven't implemented rolling saves for targets though as part of this work.

image

The majority of this work is again shuffling methods. `TargetedApplicationMixin` was removed as we now delegate to this element rather than mixin the behaviour. Most of `effect-application` internals were also moved to the element as the 'default' target selection behaviour, with `damage-application` overriding it.


module/data/chat-message/usage-message-data.mjs
```diff
@@ -110,12 +110,14 @@ export default class UsageMessageData extends ItemMessageData {
...
       context.activity = this.activity;
       context.buttons = this._prepareButtons();
+      context.showTargets = true;
       this._prepareButtonGroups(context);
       if ( this.activity.name ) context.subtitle = this.activity.name;
     }
 
     const item = this.parent.getAssociatedItem();
     const activity = this.parent.getAssociatedActivity();
+    context.showTargets = this.effects.length || (activity.type === "check") || (activity.type === "save");
```
Fyorl, line 120: I imagine we want some getter or metadata on the activity to control this. We only really want to show the tray for 'request'-type activities. An attack roll you choose targets per attack rather than at the time of using the activity, for example, so we don't show targets there, except if the attack has some effects associated with it, in which case we have to show targets for the effects.

I suppose it might be confusing if you have an attack activity with associated effects and you think the attacks will all be rolled against the pre-targeted tokens rather than the currently-targeted ones. Not sure if that would actually be an issue in practice.

arbron: I can see some future development of activities like teleport and transform which currently rely entirely on selected tokens for who they affect, where in the future we might want to handle that using targeting data.

arbron (approved):

module/data/chat-message/usage-message-data.mjs
Fyorl: I've moved this out to a metadata property in a follow-up PR.

PR #7366, opened 2026-08-21
[#3011] Allow `@item` use in effect values, move custom `_checkCondition` into core `shouldApplyChange`

### LINKED ISSUE #3011: Allow active effects on items to use the `@item` variable in change values
I was going to file this for core Foundry but since it's the system that adds `@item` to the roll data I filed it here.

With the new active effect transferral mode, the active effect exists on the item but is applied to the actor. Could we use the item's `getRollData` so we'd have access to the `@item` variable? There are some people on the Discord server that use one of the 3 resources to track a value that then gets used by an active effect. For example, in Van Richten's Guide to Ravenloft, there's a Stress mechanic that gives a penalty to d20 rolls. With a resource, it'd be easy to reference that value in an active effect but with them being deprecated that makes it hard to track this. If it's an active effect on an item, it could use `- @item.system.uses.value` in the change values.

roth-michael:
Closes #3011 
The changes to `ActiveEffect5e` are self-explanatory. Other piece is that I moved the call of `prepareFinalAttributes` on embedded items & `_prepareSpellcasting` from `Actor5e#prepareData` into the "final" phase in `Actor5e#applyActiveEffects`, so that resulting values can actually be modified by `final`-phase AE changes.

Follow-up improvement of this (specifically, being able to use `@item.uses.value`) in #7368.

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -596,10 +596,7 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
       if ( this.system.conditions?.check(conditionData) === false ) return false;
       if ( change.conditions?.check(conditionData) === false ) return false;
     }
-
-    const originalChange = this.system.changes.find(c => c._id === change._id);
-    if ( originalChange ) originalChange.applied = true;
-
+    change.applied = true;
     return true;
   }
 
@@ -607,15 +604,14 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
 
   /** @inheritDoc */
   getReplacementData(baseData) {
+    if ( !this.item ) return super.getReplacementData(baseData);
     baseData = { ...super.getReplacementData(baseData) };
-    if ( this.item ) {
-      baseData.item = {
-        ...this.item.system,
-        flags: this.item.flags,
-        name: this.item.name
-      };
-      baseData.scaling = new Scaling(this.item.scalingIncrease);
-    }
+    baseData.item = {
+      ...this.item.system,
+      flags: this.item.flags,
+      name: this.item.name
+    };
+    baseData.scaling = new Scaling(this.item.scalingIncrease);
     return baseData;
   }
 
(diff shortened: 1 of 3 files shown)


module/documents/actor/actor.mjs
```diff
@@ -399,6 +388,12 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       }
     }
 
+    // Prepare final attributes & spellcasting so that they can be used/modified in "final" phase
+    if ( phase === "final" && (this.system?.modelProvider === dnd5e) ) {
```
arbron: We should take this chance to remove the model provider check. Moving this into the final phase makes it harder for module-provided actor types to call this reliably at the correct moment and I don't see any reason to exclude them from these calls anyway.

```suggestion
    if ( phase === "final" ) {
```

module/documents/active-effect.mjs
```diff
@@ -609,6 +602,19 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
+  /* -------------------------------------------- */
+
+  /** @inheritDoc */
+  getReplacementData(baseData) {
+    baseData = { ...super.getReplacementData(baseData) };
+    if ( this.item ) baseData.item = {
+      ...this.item.system,
+      flags: this.item.flags,
+      name: this.item.name
+    };
```
arbron: `Item.getRollData` also sets `rollData.scaling`, which I could see being useful (don't forget to import `Scaling`).

```suggestion
    if ( this.item ) {
      baseData.item = {
        ...this.item.system,
        flags: this.item.flags,
        name: this.item.name
      };
      baseData.scaling = new Scaling(this.item.scalingIncrease);
    }
```

arbron (changes requested):

arbron (approved):

arbron:
Added the `breaking` flag because module-provided actor types that previously had to manually call `prepareFinalAttributes` on their items will now be calling them twice.

Fyorl (approved):

PR #7369, opened 2026-08-21
Add compact styling for saving throw cards. Allow save cards to be rendered as a summary of the usage card that spawned them.

Fyorl:
Adds compact styling for standalone saving throw cards, like death saves or concentration saves.

image

If a save card is spawned from a usage card, it will show the standalone card in the notification tray, but will be appended as a summary to the usage card that spawned it.

Screenshot From 2026-08-21 20-43-11

A damage card spawned by the same usage card is also now aware of any outcomes and will update its multipliers accordingly, including if an outcome comes in after the damage has already been rolled.

Some other changes:
- Deleting a usage card deletes all attached summary cards.
- Hovering a summary card will hover the token that rolled it.
- Added context menu options to select all summary outcomes, or just the successes or failures.

Most of these have also been applied to the check card.

### DIFF
--- a/module/data/chat-message/check-message-data.mjs
+++ b/module/data/chat-message/check-message-data.mjs
@@ -39,6 +39,7 @@ export default class CheckMessageData extends RollMessageData {
   /*  Rendering                                   */
   /* -------------------------------------------- */
 
+  /** @inheritDoc */
   async _prepareContext(options) {
     const context = await super._prepareContext(options);
     const actor = this.parent.getAssociatedActor();
--- a/module/data/chat-message/check-message-data.mjs
+++ b/module/data/chat-message/check-message-data.mjs
@@ -31,6 +31,26 @@ export default class CheckMessageData extends RollMessageData {
 
   /** @inheritDoc */
   static metadata = Object.freeze(foundry.utils.mergeObject(super.metadata, {
+    summaryTemplate: "systems/dnd5e/templates/chat/check-summary.hbs",
     template: "systems/dnd5e/templates/chat/check-card.hbs"
   }, { inplace: false }));
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  async _prepareContext(options) {
+    const context = await super._prepareContext(options);
+    const actor = this.parent.getAssociatedActor();
+    const token = this.parent.getAssociatedToken();
+    const item = this.parent.getAssociatedItem();
+    const isPrivate = !this.parent.isContentVisible;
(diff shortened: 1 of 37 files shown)


module/documents/chat-message.mjs
```diff
@@ -351,6 +388,27 @@ export default class ChatMessage5e extends ChatMessage {
         group: "attack",
         visible: canTarget,
         onClick: (_, target) => game.messages.get(target.dataset.messageId)?.selectTargets("miss")
+      },
+      {
+        label: _loc("DND5E.CHATMESSAGE.Action.SelectAll"),
```
krbz999: They should be localized automatically.

module/data/chat-message/check-message-data.mjs
```diff
@@ -31,6 +31,26 @@ export default class CheckMessageData extends RollMessageData {
...
   static metadata = Object.freeze(foundry.utils.mergeObject(super.metadata, {
+    summaryTemplate: "systems/dnd5e/templates/chat/check-summary.hbs",
     template: "systems/dnd5e/templates/chat/check-card.hbs"
   }, { inplace: false }));
+
+  /* -------------------------------------------- */
+  /*  Rendering                                   */
+  /* -------------------------------------------- */
+
+  async _prepareContext(options) {
```
arbron, line 43:
```suggestion
  /** @inheritDoc */
  async _prepareContext(options) {
```

arbron (changes requested):
Check messages when we have multiple checks in a single card don't get grouped properly because their `origin` isn't being set. We set the origin using the event, but for checks the event is inside the dropdown menu rather than the button on the card itself so it cannot find the correct chat message to associate.

Noticed while testing that damage cards from enrichers or straight in chat (`/damage 1d4 fire`) throw an error because they expect `this.item` to exist in `_getDamageProperties`, buts its `null` for those cards.

Few nice-to-haves that don't need to go in here:
- Option for non-summary cards
- Ability to delete individual summarized cards from their parent card

module/settings.mjs
```diff
@@ -203,6 +203,15 @@ export function registerSystemSettings() {
     requiresReload: true
   });
 
+  // Chat card summaries
+  game.settings.register("dnd5e", "chatCardSummary", {
+    default: true,
```
arbron, line 209: Intentionally not configurable? If so, I'm fine with it being a module-only option.

arbron (approved):

module/settings.mjs
Fyorl: Oops, no

PR #7370, opened 2026-08-22
Add compact styles to some remaining messages. Convert initiative rolls to check-type rolls.

Fyorl:
I think only bastion cards left. I plan to do those after I've looked at your bastion time advancement PR.

I think 'spoken' messages, i.e. ones typed in chat should probably stay as the 'jumbo' style. Those seem worth keeping the avatar and other header matter. But everything else I have converted to the compact style. I also noticed that initiative rolls were still type `base` which I hadn't realised was the case. They become `check` rolls now.


module/documents/actor/actor.mjs
```diff
@@ -1876,7 +1877,10 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
       this._cachedInitiativeRoll = new CONFIG.Dice.BasicRoll(String(options.fixed), data, options);
     }
 
-    await this.rollInitiative({ createCombatants: true, initiativeOptions: { messageOptions } });
+    await this.rollInitiative({
+      createCombatants: true,
+      initiativeOptions: { messageMode: CONFIG.Dice.BasicRoll.getMessageMode() }
```
Fyorl: Fixes deprecation warning

module/documents/chat-message.mjs
```diff
@@ -151,7 +151,9 @@ export default class ChatMessage5e extends ChatMessage {
       this._collapseTrays(html);
     }
 
-    if ( this.system?.summaryTemplate && this.system.origin && !options.canClose ) html.hidden = true;
+    if ( this.system?.summaryTemplate && this.system.origin?.system?.rendersSummaries && !options.canClose ) {
```
Fyorl: Noticed that checks rolled as part of a group check request were being hidden because `check` messages have `summaryTemplate`. So I have added the `rendersSummaries` counterpart to check that the origin supports summaries. Though with group checks in particular, the request card kind of acts as a summary already so I wonder if we do actually want to keep these check rolls suppressed, and maybe show the roll breakdown in the group request card on popover.

less/v3/chat.less
```diff
@@ -204,6 +206,12 @@
       &[hidden] { display: none; }
     }
 
+    .card-buttons button {
```
arbron: Looks like we lost the gap between buttons at some point.

Image

module/documents/actor/actor.mjs
arbron: Should grab this from `messageOptions` above, otherwise changes made in the `_preRollInitiative` hook won't stick.

```suggestion
      initiativeOptions: { messageMode: messageOptions.rollMode }
```

arbron (changes requested):

arbron (approved):

PR #7373, opened 2026-08-24
BUG: Guard against effects not existing on cast activities

MrPrimate:
In the v6 codebase moving into regions with effects breaks if the actor has an item with a cast activity. I think this is because the cast activity deletes effects from the schema data. https://github.com/foundryvtt/dnd5e/blob/6db48c81992abed2ebd84833d2ff741807d467fa/module/data/activity/cast-data.mjs#L19

This fix guards against effects not existing. Not sure if you want to do this another way.

### DIFF
--- a/module/data/active-effect/base.mjs
+++ b/module/data/active-effect/base.mjs
@@ -61,7 +61,7 @@ export default class BaseEffectData extends ActiveEffectDataModel {
    */
   get isOnActivity() {
     return this.item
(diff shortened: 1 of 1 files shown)


roth-michael:
Yeah accidentally introduced this.  It'd even yell during data prep if a forward activity existed. It's actually fixed in #7319 with some mostly unrelated changes, but since it's approved it'll likely go in with those.

PR #7376, opened 2026-08-26
Convert bastion cards to compact styles. Mark items as claimed when dragged from the card.

Fyorl:
image

image


module/data/chat-message/bastion-attack-message-data.mjs
```diff
@@ -53,27 +59,46 @@ export default class BastionAttackMessageData extends ChatMessageDataModel {
...
+        label: "DND5E.Bastion.Attack.Automatic"
+      }] };
     }
     if ( this.damaged ) {
       const facility = this.actor?.items.get(this.damaged);
       if ( facility ) context.damaged = _loc("DND5E.Bastion.Attack.Result.Damaged", {
         link: facility.toAnchor().outerHTML
       });
     }
+    if ( !foundry.utils.isEmpty(buttons) )context.buttonGroups = buttons;
```
arbron:
```suggestion
    if ( !foundry.utils.isEmpty(buttons) ) context.buttonGroups = buttons;
```

arbron (approved):
Looking at all the messages together, the completed order message summary being smaller and left-aligned feels out of place with the others. It would be nice to standardize one way or the other across the bastion messages.

Image

Otherwise looks good!

krbz999:
Maybe `balance` the linebreaks as well with `text-wrap-style` so the message doesn't have a single dangling word on a new
line.

Fyorl:
> Looking at all the messages together, the completed order message summary being smaller and left-aligned feels out of place with the others. It would be nice to standardize one way or the other across the bastion messages.

Well this is actually a list of all the orders that completed. But now we track progress properly, it's a bit rarer that you have multiple facility orders completing at the same time, especially for crafting facilities. I can add some special-case styling for the messages that have only one order completion.

PR #7383, opened 2026-08-26
[#5782] Migrate advancement `title` & `icon` to match other docs

### LINKED ISSUE #5782: Migrate advancement `title` and `icon` to be in line with all other documents
The advancement data structure uses `title` and `icon`, which all the core document types use `name` and `img`. It would clarify and simplify some things to swap to using the core names for these properties.

arbron:
Migrate `title` to `name` and `icon` to `img` and shims to handle the old paths. The properties in `metadata` aren't renamed nor are the `_defaultTitle` and `_defaultIcon` methods.

Also reorganized the remaining advancement localization keys to be inside the `DND5E.ADVANCEMENT` object.

Closes #5782


module/data/advancement/base-advancement.mjs
```diff
@@ -57,6 +53,14 @@ export default class BaseAdvancementData extends SparseDataModel {
     super.migrateData(source);
     if ( !source ) return source;
 
+    if ( source.title ) {
```
Fyorl: So we catch cases of `title: ""`
```suggestion
    if ( "title" in source ) {
```

module/data/advancement/base-advancement.mjs
```diff
@@ -57,6 +53,14 @@ export default class BaseAdvancementData extends SparseDataModel {
...
+      source.name ||= source.title;
+      delete source.title;
+    }
+    if ( source.icon ) {
```
Fyorl:
```suggestion
    if ( "icon" in source ) {
```

Fyorl (approved):

PR #7386, opened 2026-08-26
Evaluate effect conditions during initial phase, set suppressed

arbron:
Evaluates effect-wide conditions during the initial phase of application and use the result in `isSuppressed`. This allows conditions to prevent status effects from being applied but limits the data that can be evaluated using effect conditions.

Still evaluate the effect conditions for each phase, allowing individual changes to be disabled if another effect makes that condition invalid even if the whole effect is still enabled.

### DIFF
--- a/module/data/abstract/active-effect-data-model.mjs
+++ b/module/data/abstract/active-effect-data-model.mjs
@@ -144,7 +144,7 @@ export default class ActiveEffectDataModel extends foundry.data.ActiveEffectType
    * @param {RollData} baseData
    */
   evaluateCondition(baseData) {
-    this.#conditionSuppressed = !this.conditions?.check(this.parent.getReplacementData(baseData));
+    this.#conditionSuppressed = this.conditions?.check(this.parent.getReplacementData(baseData)) === false;
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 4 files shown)


module/data/abstract/active-effect-data-model.mjs
```diff
@@ -130,6 +139,16 @@ export default class ActiveEffectDataModel extends foundry.data.ActiveEffectType
   /*  Helpers                                     */
   /* -------------------------------------------- */
 
+  /**
+   * Evaluate the effect's conditions and store the result.
+   * @param {RollData} baseData
+   */
+  evaluateCondition(baseData) {
+    this.#conditionSuppressed = !this.conditions?.check(this.parent.getReplacementData(baseData));
```
Fyorl: Otherwise an AE sub-type without a top-level `conditions` is permanently suppressed.
```suggestion
    this.#conditionSuppressed = !!this.conditions && !this.conditions.check(this.parent.getReplacementData(baseData));
```

module/documents/actor/actor.mjs
```diff
@@ -382,8 +382,10 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
     }
 
     // Pre-collect statuses so they can be reliably used for conditions
+    const replacementData = this.getRollData();
     if ( phase === "initial" ) {
       for ( const effect of this.allApplicableEffects() ) {
+        effect.system.evaluateCondition?.(replacementData);
```
```suggestion
    if ( phase === "initial" ) {
      const replacementData = this.getRollData();
      for ( const effect of this.allApplicableEffects() ) {
        effect.system.evaluateCondition?.(replacementData);
```

Fyorl (approved):

PR #7390, opened 2026-08-28
Add identifiers to Actors. Show identifier field on details tab for all Items.

Fyorl:


module/documents/actor/actor.mjs
```diff
@@ -159,6 +159,17 @@ export default class Actor5e extends SystemDocumentMixin(Actor) {
 
   /* -------------------------------------------- */
 
+  /**
```
arbron: Since this is shared we could probably stick it in `SystemDocumentMixin`.

arbron (approved):

Issue #7367, opened 2026-08-21
Allow customizing number of successes/failures required for death saves


arbron:
Currently these are hard-coded, but an item like "Memento Mori" in Valda's Spire of Secrets increases the number of failures required to 5. It would be good if these could be customized. These probably don't need to be exposed in the Death Saves Config, but having a pair of non-persisted fields (`system.attributes.death.threshold.success` & `system.attributes.death.threshold.failure`) would be great.

Customizing the death save tray would probably be beyond the scope of this issue.

roth-michael:
Commenting to note that auto-dead on last failed death save would also need to account for this change

gubacelar:
Hey! I’d be interested in working on this if nobody has picked it up yet.

One thing I was wondering before starting: would it make sense for the success/failure thresholds to also be targetable by Active Effects?

That would allow features like Memento Mori to modify the failure threshold through an AE instead of permanently changing the actor data. The actor could keep a base value of 3, while the effective threshold could be modified when needed.

I’d also make sure the auto-death logic uses the configured failure threshold.

If that approach fits the system’s conventions, I’d be happy to work on it.

arbron:
> One thing I was wondering before starting: would it make sense for the success/failure thresholds to also be targetable by Active Effects?

That's why I mentioned making them non-persisted fields. I don't see any reason that these should be directly editable on the actor sheet. AE-only modification should work fine.

gubacelar:
Got it, that makes sense. Thanks for clarifying!
I’ll follow that approach and make the thresholds non-persisted and AE-targetable, and I’ll make sure the auto-death logic uses the effective failure threshold too.

gubacelar:
Hey @arbron 
I understand that changing the death save tray itself is outside the scope of this issue. However, while testing this, I noticed that the current behavior can look a little odd and may give users the wrong impression about the actual death save state.

Will there be a separate issue to address the tray behavior/visuals?

<img width="233" height="92" alt="Image" src="https://github.com/user-attachments/assets/438d428c-cd90-42bb-a9a0-54709688dee9" />

<img width="296" height="644" alt="Image" src="https://github.com/user-attachments/assets/c5b2c7e8-3c9d-4eda-b8fb-ec21e18382ac" />

krbz999:
Just like exhaustion, the sheet should be adjusted to display the variable number of fails or successes, but it could be left up to modules to apply some css.

PR #7394, opened 2026-08-31
[#7367] Allow customizable death save thresholds

gubacelar:
Add non-persisted success and failure thresholds that Active Effects can modify, and use them throughout death save and auto-death handling.

```
Key: system.attributes.death.threshold.failure
Mode: Override
Value: 5
```

```
Key: system.attributes.death.threshold.failure
Mode: Add
Value: 2
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "eslint-plugin-jsdoc": "^63.0.10",
         "fancy-log": "^2.0.0",
         "globals": "^17.7.0",
-        "js-yaml": "^5.2.0",
+        "js-yaml": "^5.4.1",
         "less": "^4.6.7",
         "less-watch-compiler": "^1.17.5",
         "rollup": "^4.62.2",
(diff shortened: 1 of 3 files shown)


module/data/actor/character.mjs
```diff
@@ -78,6 +78,10 @@ export default class CharacterData extends CreatureTemplate {
           failure: new NumberField({
             required: true, nullable: false, integer: true, min: 0, initial: 0, label: "DND5E.DeathSaveFailures"
           }),
+          threshold: new SchemaField({
+            success: new NumberField({ integer: true, min: 1, initial: 3 }),
+            failure: new NumberField({ integer: true, min: 1, initial: 3 })
```
krbz999: Might need `nullable: false` too.

gubacelar: nullable already defaults to false for NumberField in V14, but making the invariant explicit here makes sense. I’ll add it to both Character and NPC schemas.

gubacelar: updated :)

gubacelar: should I squash?

krbz999: > nullable already defaults to false for NumberField in V14, but making the invariant explicit here makes sense. I’ll add it to both Character and NPC schemas.

That is definitely not the case. Regardless, good to guard against the `EffectChangeData#value` being an empty string, which would set the field to `null` when using the `override` change type (see `NumberField#clean`).

gubacelar: You're righ... I was looking at the general DataField default rather than NumberField's override. The explicit nullable: false now guards against empty override values being cleaned to null. Thanks for the clarification, that's make sense

PR #7396, opened 2026-09-02
V6 Concentration Effect Fixes

MrPrimate:
Fixes the following:

- Concentration effects no longer show up in effects panel
- When concentration expires the effect remains (not visible), but the spell still picked up it was concetrating (swirly spell icon) in the spell list ui.

This PR:
- Removes the concentration effect when it expires.
- Restores visibility in the UI
- Adds a guard aginst showing in active put still present concentration effects on an actor.

### DIFF
--- a/module/documents/active-effect.mjs
+++ b/module/documents/active-effect.mjs
@@ -889,10 +889,8 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     const newEncumbrance = data.statuses?.[0];
     const name = this.name;
 
-    // Delete expired concentration effects immediately, and other expired effects outside of combat
+    // If out of combat & effect expires, delete it
     if ( game.user.isActiveGM && data.duration?.expired ) {
-      const isConcentration = this.statuses.has(CONFIG.specialStatusEffects.CONCENTRATING);
-      if ( isConcentration ) return this.delete();
       const actor = this.isAppliedEnchantment ? this.parent.parent : this.parent;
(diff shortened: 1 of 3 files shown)


module/documents/active-effect.mjs
```diff
@@ -889,8 +889,10 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     const newEncumbrance = data.statuses?.[0];
     const name = this.name;
 
-    // If out of combat & effect expires, delete it
+    // Delete expired concentration effects immediately, and other expired effects outside of combat
     if ( game.user.isActiveGM && data.duration?.expired ) {
+      const isConcentration = this.statuses.has(CONFIG.specialStatusEffects.CONCENTRATING);
+      if ( isConcentration ) return this.delete();
```
Fyorl: I fixed a bug in https://github.com/foundryvtt/dnd5e/commit/80b3e85e99e85e757d540c38463fddf2c425c810 that was allowing expired effects to continue to apply their changes. With that fixed, the `!effect.active` in `get concentration()` should catch these so it shouldn't be necessary to immediately delete them.

We intentionally do not immediately delete effects during combat because accidentally advancing a turn can cause a cascade of deleted effects which can be arduous to recover.

Fyorl (changes requested):

module/documents/active-effect.mjs
MrPrimate: Gotcha, removed.

Fyorl (approved):

PR #7398, opened 2026-09-02
Evaluate conditions as late as possible, add condition warnings

arbron:
Rather than evaluating all effect conditions before statuses are applied, instead only evaluate conditions for effects that apply their own statuses at that point. Other conditions are evaluated during the earliest phase of any of their changes. Any effect that doesn't have changes is checked during the final phase.

Also introduces warnings to the active effect config sheet when a condition key is used that won't work in an effect condition. This includes using a `statuses.` key while also applying a status effect or using any `roll.` property.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -2787,7 +2787,11 @@
   "Suppressed": {
     "Hint": "Source item must be equipped or attuned to activate these"
   },
-  "Tab": "Effects"
+  "Tab": "Effects",
+  "Warning": {
+    "RollCondition": "Effect conditions cannot reference 'roll.' properties. Move this condition into the individual changes for the effect to be applied.",
+    "StatusCondition": "Effect that grant statuses cannot also be conditional on statuses. Move this condition into the individual changes for the effect to be applied."
+  }
 },
 
 "DND5E.ENCHANT": {
--- a/lang/en.json
+++ b/lang/en.json
@@ -2790,7 +2790,7 @@
   "Tab": "Effects",
   "Warning": {
     "RollCondition": "Effect conditions cannot reference 'roll.' properties. Move this condition into the individual changes for the effect to be applied.",
-    "StatusCondition": "Effect that grant statuses cannot also be conditional on statuses. Move this condition into the individual changes for the effect to be applied."
+    "StatusCondition": "Effects that grant statuses cannot also be conditional on statuses. Move this condition into the individual changes for the effect to be applied."
   }
 },
 
--- a/module/applications/active-effect/active-effect-sheet.mjs
+++ b/module/applications/active-effect/active-effect-sheet.mjs
@@ -100,6 +100,14 @@ export default class ActiveEffectSheet5e extends ApplicationV2Mixin(ActiveEffect
    * @protected
    */
   async _prepareDetailsContext(context, options) {
+    const { statuses, system } = context.document;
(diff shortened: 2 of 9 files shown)


module/documents/active-effect.mjs
```diff
@@ -173,6 +173,7 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
 
   /** @inheritDoc */
   get isSuppressed() {
+    if ( super.isSuppressed ) return true;
```
arbron, line 176: This got removed completely during the recent fixes for expired effects, meaning `system.isSuppressed` was never being called.

Fyorl: Whoops, only meant to delete the `this.system.isSuppressed` half

lang/en.json
```diff
@@ -2794,7 +2787,11 @@
   "Suppressed": {
     "Hint": "Source item must be equipped or attuned to activate these"
   },
-  "Tab": "Effects"
+  "Tab": "Effects",
+  "Warning": {
+    "RollCondition": "Effect conditions cannot reference 'roll.' properties. Move this condition into the individual changes for the effect to be applied.",
+    "StatusCondition": "Effect that grant statuses cannot also be conditional on statuses. Move this condition into the individual changes for the effect to be applied."
```
```suggestion
    "StatusCondition": "Effects that grant statuses cannot also be conditional on statuses. Move this condition into the individual changes for the effect to be applied."
```

module/documents/active-effect.mjs
```diff
@@ -602,13 +603,15 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
...
   /** @inheritDoc */
-  shouldApplyChange(change, options) {
+  shouldApplyChange(change, options={}) {
     if ( !super.shouldApplyChange(change, options) ) return false;
-    const conditionData = options?.replacementData;
-    if ( conditionData && !CONFIG.ActiveEffect.changeTypes[change.type]?.skipConditions ) {
-      if ( this.system.conditions?.check(conditionData) === false ) return false;
-      if ( change.conditions?.check(conditionData) === false ) return false;
+    const conditionData = options.replacementData;
+    if ( this.system.evaluateCondition && this.system.changes.some(c => c.phase === options.phase) ) {
```
Fyorl: Not possible for `this.system.changes.some(c => c.phase === options.phase)` to be `false` because `super.shouldApplyChange` checks `change.phase === options.phase`, so there is always at least one change with `change.phase === options.phase`, this one.


```suggestion
    if ( this.system.evaluateCondition ) {
```

module/applications/active-effect/active-effect-sheet.mjs
```diff
@@ -100,6 +100,14 @@ export default class ActiveEffectSheet5e extends ApplicationV2Mixin(ActiveEffect
    * @protected
    */
   async _prepareDetailsContext(context, options) {
+    const { statuses, system } = context.document;
+    if ( system.conditions ) {
+      if ( statuses.size && system.conditions.some(f => f.k?.startsWith?.("statuses")) ) {
```
Fyorl:
```suggestion
      if ( statuses.size && system.conditions.some(f => f.k?.startsWith?.("statuses.")) ) {
```

Fyorl (approved):

PR #7402, opened 2026-09-03
[#6576] Fix synthetic steps in advancement manager

### LINKED ISSUE #6576: Advancement of an item granted by an Advancement doesn't then do its advancement and grant further items when awarded the item
Foundry: 13.351
D&D: 5.2.3

Reproduction steps

Subclass that awards a level 3 feature .. This feature itself has a advancement of choose 1 of 2 features....   These Features both have advancements.

When the actor gets to level 3 choose the appropriate subclass, get awarded the feature from the Subclass Advancement make a choice of feature, but the advancement from the chosen feature doesn't pop up.

On investigation of the features advancement it shows a triangle warning `not configured` 

Image

rerunning the advancement on the feature works perfectly.

Then leveling up further... the advancement of the feature works perfectly.

Part 2 - if the choice advancement is redone ... then the feature is awarded but the advancement of the feature has all levels marked as `not configured.  

Image

Again manualy running each level with the triangle again resolves the problem... but idealy these advancements should be done when the feature is assigned to the actor

arbron:
- Move `getLevel` so it isn't being defined in the loop bodies
- Track maximum level and create steps up to the maximum level if the end of the steps array is reached
- Clarify the start and end range of advancement steps being created at any point, set `handledLevel` to `rangeEnd`
- Default `handledLevel` to `-Infinity` rather than `0` to ensure "Any Level" steps are always captured

Closes #6576

### DIFF
--- a/module/applications/advancement/advancement-manager.mjs
+++ b/module/applications/advancement/advancement-manager.mjs
@@ -736,22 +736,27 @@ export default class AdvancementManager extends Application5e {
     const addedIds = postIds.difference(preIds).difference(initialIds);
     const deletedIds = preIds.difference(postIds).difference(initialIds);
 
+    const getLevel = (step, advancementClassLinked) => (advancementClassLinked ? undefined : step?.level)
+      ?? step?.flow?.level ?? step?.class?.level ?? step?.level;
+
     for ( const addedId of addedIds ) {
       const item = this.clone.items.get(addedId);
       if ( !item.hasAdvancement ) continue;
 
-      let handledLevel = 0;
+      let maximumLevel = 0;
+      let handledLevel = -Infinity;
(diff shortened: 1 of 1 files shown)


module/applications/advancement/advancement-manager.mjs
Fyorl, line 774: For `forDeletedItem()` we already add all the reversals, followed by a final `delete` step. Now that we don't filter out `initialIds`, that `delete` step is getting caught here and adding another set of reversals. I don't think it breaks anything but seems easy enough to avoid:


```suggestion
      if ( this.step.item?.id === deletedId ) continue;
      let item = this.#preEmbeddedItems.find(i => i.id === deletedId);
```

Fyorl (approved):

PR #7403, opened 2026-09-03
[#6160] Allow items to declare multiple rarities and have them matched in the compendium browser.

### LINKED ISSUE #6160: Allow items to have multiple rarities
Many magic items have different rarity levels, like the Wand of the War Mage, Rod of the Pact Keeper, etc that can be multiple different rarities.

Fyorl:
- Closes #6160

### DIFF
--- a/module/data/item/templates/physical-item.mjs
+++ b/module/data/item/templates/physical-item.mjs
@@ -166,7 +166,7 @@ export default class PhysicalItemTemplate extends SystemDataModel {
    */
   get physicalItemSheetFields() {
     return [{
-      label: Array.from(this.rarities).map(r => CONFIG.DND5E.itemRarity[r]).filterJoin(", "),
+      label: this.rarities.size > 1 ? _loc("DND5E.ITEM.Rarity.Varies") : CONFIG.DND5E.itemRarity[this.rarity],
       value: this._source.rarities,
       requiresIdentification: true,
       field: this.schema.getField("rarities"),
--- a/module/data/item/templates/physical-item.mjs
+++ b/module/data/item/templates/physical-item.mjs
@@ -1,7 +1,7 @@
 import { convertWeight, defaultUnits } from "../../../utils.mjs";
 import SystemDataModel from "../../abstract/system-data-model.mjs";
 
-const { ForeignDocumentField, NumberField, SchemaField, StringField } = foundry.data.fields;
+const { ForeignDocumentField, NumberField, SchemaField, SetField, StringField } = foundry.data.fields;
 
 /**
  * @import { CompendiumBrowserFilterDefinitionEntry } from "../../../applications/compendium-browser.mjs";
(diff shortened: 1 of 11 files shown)


module/documents/active-effect.mjs
```diff
@@ -98,7 +99,8 @@ export default class ActiveEffect5e extends DependentDocumentMixin(ActiveEffect)
     "system.bonuses.rsak.damage": { key: "system.rolls.damage.rsak.bonus" },
     "system.bonuses.abilities.check": { key: "system.rolls.ability.check.bonus" },
     "system.bonuses.abilities.save": { key: "system.rolls.ability.save.bonus" },
-    "system.bonuses.abilities.skill": { key: "system.rolls.ability.skill.bonus" }
+    "system.bonuses.abilities.skill": { key: "system.rolls.ability.skill.bonus" },
+    "system.rarity": { key: "system.rarities" }
```
Fyorl, line 103: Overriding with a plain string will be automatically cleaned to a Set as part of the delta cast.

module/data/item/templates/physical-item.mjs
```diff
@@ -113,12 +166,11 @@ export default class PhysicalItemTemplate extends SystemDataModel {
    */
   get physicalItemSheetFields() {
     return [{
-      label: CONFIG.DND5E.itemRarity[this.rarity],
-      value: this._source.rarity,
+      label: Array.from(this.rarities).map(r => CONFIG.DND5E.itemRarity[r]).filterJoin(", "),
```
arbron:
```suggestion
      label: game.i18n.getListFormatter({ type: "unit" }).format(
        Array.from(this.rarities).map(r => CONFIG.DND5E.itemRarity[r]).filter(_ => _)
      ),
```

I suppose we could also just say "Varies" to match how these items are presented in the text. Its sort of a mixture of some items saying "Rarity Varies", while others list out the specific rarities.

module/filter.mjs
```diff
@@ -316,6 +316,18 @@ export function iendswith(data, value) {
 
 /* -------------------------------------------- */
 
+/**
+ * Check if the data is empty.
+ * @param {*} data
+ * @param {*} value
+ * @returns {boolean}
+ */
+export function empty(data, value=true) {
```
arbron, line 325: Should add symbols to `filter-comparison` styling in case these new methods are used in AE filters.

arbron (changes requested):
Not the greatest layout if you select too many rarities, but this will probably never happen in normal usage:

Image

Did a bit of a prototype of a way to get it looking nicely, but not sure if its 100% possible with CSS:

Image

Fyorl:
> Did a bit of a prototype of a way to get it looking nicely, but not sure if its 100% possible with CSS:
> Image

Turns out it is possible with some trial and error

arbron (approved):
Had to do a little fix to `<filters-input>` when you use the `empty` operator, but otherwise looking great.

PR #7406, opened 2026-09-04
Move rider collection methods into AE system data models

arbron:
Move `collectRiderConditions` and `collectRiderEnchantments` into `system.collectRiders` so system AE types can provide riders themselves.


module/data/active-effect/mixins/rider-provider.mjs
```diff
@@ -0,0 +1,41 @@
...
+ * @mixin
+ */
+export default function RiderProviderMixin(Base) {
+  class RiderProvider extends Base {
+    /**
+     * Handle creating rider documents when this active effect is created.
+     * @param {DatabaseOperation} options            Options passed to the create or update operation.
+     * @returns {Promise<DatabaseWriteOperation[]>}  Batch entries suitable for `foundry.documents.modifyBatch`.
+     */
+    async collectRiders(options) {
```
Fyorl:
```suggestion
    async collectRiders(options={}) {
```

module/data/active-effect/enchantment.mjs
```diff
@@ -249,6 +250,92 @@ export default class EnchantmentData extends ActiveEffectDataModel {
   /*  Helpers                                     */
   /* -------------------------------------------- */
 
+  /** @override */
+  async collectRiders(options) {
```
Fyorl:
```suggestion
  async collectRiders(options={}) {
```

module/data/active-effect/mixins/rider-provider.mjs
```diff
@@ -0,0 +1,41 @@
...
+      if ( !this.parent.actor ) return [];
```
Fyorl: Seems like if I add an effect with the `unconscious` status to an owned item, it immediately applies `prone` to the actor. We used to check `effect.parent instanceof Actor` explicitly, so I think we need to do that here:

```suggestion
      if ( !(this.parent.target instanceof Actor) ) return [];
```

But this method seems very AE-specific rather than something that gets mixed-into any system data model. Shouldn't it be on `ActiveEffectDataModel`? Is the issue that `ConditionData` doesn't extend `ActiveEffectDataModel`? Should it?

Fyorl (changes requested):

module/data/active-effect/mixins/rider-provider.mjs
krbz999: It could and maybe should. It wouldn't hurt. I did not extend `ActiveEffectDataModel` initially cus a lot of the scaffolding in said model seemed irrelevant to status effects.

Fyorl: Then we probably wouldn't need the mixin at all. But having it extend the base `ActiveEffectDataModel` will bear some testing to make sure there's nothing we're missing. Suppression logic in particular is sensitive I think.

arbron: I originally had it on `ActiveEffectDataModel` and swapped to a mixin due to that issue, but I can rework `ConditionData` to inherit from that instead.

Fyorl (approved):

PR #7431, opened 2026-09-10
[#7429] Refactor summoned token creation for non gm owner issue

### LINKED ISSUE #7429: Player Unable to place summoned actor
DND 6.0.0
Foundry: 14.367

When trying to summon an actor as a player  the following error is produced and no actor arrives on the scene:

Image

## Reproduction Steps

1. Level 3 Beastmaster Ranger
2. Use Primal Companion feature, - select any summon....
3. Recieve error

## Additional Notes:

- Everytime the summon happens an Actor is placed in the sidebar, and that player is set as the owner of that actor
- Player permisions are set with Create Actors, Create New Tokens, Use File Browser
- DND Game Settings set to Allow Summoning
- If the player goes and drags the summon from the actor side bar on to the scene it goes on to the scene no problem

The error feels like its coming when the system should be showing the preview of the token for where to place the token... not when actually placing the token but this might just be my perception.

thatlonelybugbear:
Closes #7429

- Should be fixing non-GM summon placement failing. `actor.getTokenDocument()` previously ran only after placement and the preview was built from `actor.prototypeToken` without an `actorId`.
- Moves creation of a "proper" Scene `TokenDocument` before placement.
- Uses that preview document for final creation, which should be keeping any wildcard selections consistent.

### DIFF
--- a/module/canvas/token-placement.mjs
+++ b/module/canvas/token-placement.mjs
@@ -20,7 +20,7 @@ export default class TokenPlacement extends BasePlacement {
     const uniqueTokens = new Map();
     const base = this.config.origin?.elevation ?? canvas.level.elevation.base; // Use the summoner's elevation.
     await canvas.tokens.placeTokens(this.config.tokens.map(t => ({
-      ...t.toObject(), elevation: base, level: canvas.level.id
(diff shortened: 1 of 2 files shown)


Fyorl (approved):
Opted for the minimal fix for now in order to reduce risk for the release, but your original commit remains available.

PR #7438, opened 2026-09-11
[#7436] Validate condition filters and keep their editor accessible for invalid saved filters.

### LINKED ISSUE #7430: Filters input becomes uneditable if filter JSON is a string
If you accidentally enter a string rather than an object or array in the filter editor then the whole `<filters-input>` enters a state where it doesn't finish rendering and the edit button doesn't appear, making it impossible to clear the invalid filter data.

### LINKED ISSUE #7436: Invalid change Filter data can make the Change Conditions editor inaccessible
Currently it is possible to save a malformed JSON Filter and Submit Changes without realising.

Minimal reproduction by saving a Filter like `{"":""}`

We should be disallowing closing/saving the editor while its JSON is invalid and also, in case someone manages to create and save a malformed filter, keep them accessible.

```js
Uncaught TypeError: Cannot read properties of undefined (reading 'startsWith')
[Detected 1 package: system:dnd5e(6.0.0)]
    at getHumanReadableAttributeLabel (utils.mjs:1374:13)
    at #renderBreakdownNode (filters-input.mjs:127:30)
    at #refreshBreakdown (filters-input.mjs:96:48)
    at FiltersInputElement._buildElements (filters-input.mjs:57:27)
    at FiltersInputElement.connectedCallback (foundry.mjs:41308:27)
    at EffectChangeConfig._insertElement (foundry.mjs:30924:24)
    at #render (foundry.mjs:30534:18)
    at async #try (foundry.mjs:7673:17)
```

Image

thatlonelybugbear:
- Closes #7436 by keeping the editor accessible when invalid filters have been saved, allowing them to be corrected.
- Adds condition filters validation before closing the editor, preventing malformed filters from being saved.
- Blank input is normalized to `{}` to avoid requiring users to enter an empty JSON object manually.
- Closes #7430

### DIFF
--- a/module/applications/components/filters-input.mjs
+++ b/module/applications/components/filters-input.mjs
@@ -1,4 +1,4 @@
-import { COMPARISON_FUNCTIONS, OPERATOR_FUNCTIONS } from "../../filter.mjs";
+import { COMPARISON_FUNCTIONS, isValidFilter, OPERATOR_FUNCTIONS } from "../../filter.mjs";
 import { getHumanReadableAttributeLabel } from "../../utils.mjs";
 import FiltersEditor from "../filters-editor.mjs";
 
@@ -90,6 +90,11 @@ export default class FiltersInputElement extends foundry.applications.elements.A
     this.#breakdown.classList.toggle("empty", isEmpty);
     if ( isEmpty ) {
       this.#breakdown.innerText = _loc("DND5E.FILTER.Empty");
+      return;
+    }
+
+    if ( !isValidFilter(filters) ) {
+      this.#breakdown.innerText = _loc("DND5E.FILTER.Invalid");
       return;
     }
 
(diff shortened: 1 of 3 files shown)


thatlonelybugbear:
So it is evident that this PR introduces more validation logic than the minimal fix required for both #7436 and #7430 😅  

If the extra safeguards are not wanted, I can revise to only include the minimal fix for keeping the conditions editor accessible.

arbron:
I pushed up a very simple fix to just fix the un-openable editor issue. I'd be open to more complex validation, but I'd probably want improved diagnostics with it to guide users to what part of the filter isn't valid.

I'm also opposed to blocking the closing of the filter dialog since it is a modal window, so someone could end up with an invalid filter but want to do something else and be unable to do anything else in Foundry until they fixed it.

thatlonelybugbear:
I am dealing with something similar in Rave5e so I can try to add something in the future.
[raveJSON.webm](https://github.com/user-attachments/assets/c1208db9-11c0-49b2-b427-38907391d0d9)


About the blocking, 
I just don't like the idea of saving malformed data, then realising and having to click again to open the editor.

Maybe adding a Cancel/Reset state button, so that one could get out fast would be the best of two worlds?

PR #7451, opened 2026-09-12
[#7450] Use rolled item data in Rule filters and expose the source item

### LINKED ISSUE #7450: Associate `item` consistently to the rolled Item in Filters and Rules formulas
- Filter conditions:
  - Effect parent is Item: `item` refers to the effect's source Item.
  - Effect parent is Actor: `item` refers to the rolled Item.
- Rule formulas:
  - `@item` refers to the rolled Item.

thatlonelybugbear:
Closes #7450 

I added `sourceItem` as a reference to the effect's parent if that is an Item, which could be helpful.

### DIFF
--- a/module/documents/applied-rules.mjs
+++ b/module/documents/applied-rules.mjs
@@ -124,6 +124,11 @@ class RulesIterator extends Iterator {
     return new RulesIterator(this.filter(r => {
       if ( consumed?.has(r) ) return false;
       const conditionData = r.effect?.getReplacementData(rollData) ?? rollData;
+      if ( r.effect?.item ) {
+        conditionData.sourceItem = conditionData.item;
+        conditionData.item = rollData.item;
+        conditionData.scaling = rollData.scaling;
+      }
       if ( r.effect?.system.conditions?.check(conditionData) === false ) return false;
       if ( r.conditions?.check(conditionData) === false ) return false;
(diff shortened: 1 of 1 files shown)


pgervais70:
Would allowing effects on actor also expose sourceItem from system.origin.item if defined, something that makes sense?

thatlonelybugbear:
I was thinking about it but not with this PR.

Fyorl (approved):
I factored out the logic into `ActiveEffect5e#getRuleConditionData` and added a `#getEffectItemData` that is shared with `getReplacementData`.

We should have everything covered now:

Eager AEs (normal ones):
- Item transfer effect
  - `@item` is the SOURCE item (no change)
  - `@sourceItem` also the SOURCE item, added so you don't need to guess what `@item` is
- Actor effect
  - No `@item` or `@sourceItem` available

Deferred AEs ('Rules'):
- Item transfer effect
  - `@item` is the ROLLED item (your change)
  - `@sourceItem` is the SOURCE item (your change)
  - `@roll.item` is the ROLLED item (already existed, similar to `@sourceItem` in that you don't have to guess what `@item` is)
- Actor effect
  - No `@sourceItem` available (no change)
  - `@item` is the ROLLED item (no change)
  - `@roll.item` is the ROLLED item (already existed)

PR #7458, opened 2026-09-13
Add support for dsn to the new summary feature

JiDW:
Hello,

I got a report on the Dice So Nice repo about an issue with DsN and the new summary feature.

https://gitlab.com/riccisi/foundryvtt-dice-so-nice/-/work_items/582

Looking at the code, I found the same issue with the "group roll request" feature.
Here is a fix for both, that is safe to add when if DsN is not installed and regardless of the users DsN settings.

Note that I had to add a Promise.resolve() because both codes run before Dice So Nice createChatMessage listener (from _onCreate for the summary, from the system hook for the request), so the message isn't flagged yet and the wait resolves immediately.

The same helper would be needed everywhere you show a roll result outside of the chatmessage containing the roll object.

Let me know if you have any questions or feedback.


Fyorl:
Module-specific code does not belong in the system.

PR #7468, opened 2026-09-14
[#7462] Fix negative terms not appearing in breakdown. Render separate constants.

### LINKED ISSUE #7462: Roll formula simplification hides negative modifers from chat
If the overall modifier for a roll is negative, it is not shown on the chat card, and just re-states the dice rolled value, despite the final result being lower.

Image

Additionally, all rolls show a simplified total of the modifiers, not each instance as they used to. I found this very useful to understand what modifiers are applying to what, particularly if troubleshooting something. If this simplification is an intentional part of the new UI design, could a an option be added to retain the full roll formula detail?

Fyorl:
- Closes #7462 

image


arbron (changes requested):
Damage tooltips are getting a scrollbar:

Image

I think we should do a bit more to separate out the total from the other constants, this really looks like I could be getting a +6 to the roll not a +3:

Image

arbron (approved):

PR #7469, opened 2026-09-15
Handle attack disadvantage effects

thatlonelybugbear:
Currently the implementation is not wired in it seems


kaelad02:
Can you add the same for `blinded` and `restrained`?

I'm thinking `frightened` should be left off here since it's "while the source of fear is within line of sight." Probably safer for people to manually handle that.

What about `invisible` though? We could argue that since it's conditional since if a creature can see you, you don't have Advantage on attacks and they don't have Disadvantage on attacks against you.

thatlonelybugbear:
Added `blinded` and `restrained`.

`frightened`, `grappled` and `invisible` are situational, so I think they shouldn't add a source of disadvantage by default.

thatlonelybugbear:
One could say `blinded` can be countered by Blindsense or any other relevant detection mode, in which case, we can either deem this situational too, or leave it to the GM to not use the `blinded` status when a creature has other means to "see".

_My personal preference would be to add it in, as I am dealing with this by adding condition immunity to Blinded for any creature that has other means of "seeing"._

PR #7473, opened 2026-09-15
[#7472] Add hit dice recovery percentage property

### LINKED ISSUE #7472: Add ability to set the fraction of hit dice recovered using AE
There are a number of 3rd-party features in 5e (Early Riser and Hardy Constitution in Valda's Spire of Secrets are two examples) that modify the percentage of hit dice recovered during a long rest. This is currently possible using a hook on `dnd5e.preLongRest`, but it would be nice to able to change using an AE.

If we added a non-persisted `attributes.hd.recovery` field that defaults to `0.5` for 5e and `1` for 5.5e, then this could be targeted by active effects to easily modify the amount recovered.

arbron:
Adds a new `attributes.hd.recovery` property to characters & NPCs that allows active effects to modify what fraction of an actor's hit dice are recovered on a long rest. This is a non-persisted field.

Closes #7472


module/data/actor/templates/attributes.mjs
```diff
@@ -64,6 +66,22 @@ export default class AttributesFields {
...
   /* -------------------------------------------- */
+
+  /**
+   * Hit dice fields shared between characters and NPCs.
+   * @type {HitDiceData}
+   */
+  static get hitDice() {
+    return {
+      recovery: new NumberField({
+        min: -1, max: 1, persisted: false, initial: () => dnd5e.settings.rulesVersion === "modern" ? 1 : 0.5,
```
krbz999, line 77: Should this perhaps go into `CONFIG.DND5E`?

PR #7476, opened 2026-09-16
[#7475] Fix advantage mode counts when parent roll data is missing

### LINKED ISSUE #7475: Attack roll mode changes fail when the parent roll data is not initialized
There is a strange issue with the attached actor (discussion thread: https://discord.com/channels/170995199584108546/1549546890180304939)

 `system.rolls.attack.mwak.mode | Add | 1` fails to grant advantage.

```js
await _token.actor.update({
    "system.rolls.attack.mode": 0,
    "system.rolls.attack.mwak.mode": 0,
    "system.rolls.attack.rwak.mode": 0,
    "system.rolls.attack.msak.mode": 0,
    "system.rolls.attack.rsak.mode": 0
  });
``` 
makes it start behaving.

It seems that the `#migrateBonusData()` creates the `system.rolls.attack.*` paths only when migrating legacy **non** empty bonus
```js
static #migrateBonusData(source, paths) {
	for (const [original, updated] of paths ?? CommonTemplate.#BONUS_FIELD_PATHS) {
		if (foundry.utils.hasProperty(source, updated)) continue;

		const value = foundry.utils.getProperty(source, original);

		if (!value) continue; // this skips empty ones

		foundry.utils.setProperty(source, updated, value);
	}
}
```
An actor without any legacy attack bonuses can retain `source.rolls = {}` .

During the effect application the `AdvantageModeField.getCounts()` comes upon the missing path `source.rolls.attack.mwak`.

That means we get a fallback object via `const roll = foundry.utils.getProperty(model, parentKey) ?? {};`

Any advantage count written in `roll.modeCounts` is therefore lost
(`mode: 1`, but `modeCounts.advantages.count` remains 0 in the screenshot)

Image

*The AE is disabled on the actor's export (Adv Test)*

[fvtt-Actor-test-fM9UOhVrWYR1jh6S.json](https://github.com/user-attachments/files/32265998/fvtt-Actor-test-fM9UOhVrWYR1jh6S.json)

thatlonelybugbear:
Closes #7475


Fyorl (approved):
Thank you for your PR. I opted for a more global solution that has these fields prepare themselves to empty objects (this is different to having `initial: {}` as it does not create the empty object in `_source`, only in the prepared data).

Issue #7228, opened 2026-07-06
Allow durations to include roll formulas


Zanderaf:
Currently, you need to manually adjust the durations of effects like the Potion of Growth (1d4 hour duration). The duration field itself understandably does not seem to accept a formula, either.

arbron:
Implementation note: this should roll when the activity is used and then store that result, so any effects applied from the usage can get the same duration.

PR #7497, opened 2026-09-18
[#7495] Add missing SRD 5.1 sample poisons

### LINKED ISSUE #7495: SRD 5.1 missing Sample Poisons
The SRD 5.1 includes the Basic Poison from the equipment section, but is missing the sample poisons presented later in the "Poisons" section:

Image

arbron:
Adds the sample poisons from the "Poisons" section of the GM tools.

Closes #7495


packs/_source/items/poison/malice.yml
```diff
@@ -0,0 +1,182 @@
...
+        message: null
+      conditions: '{}'
+      magical: false
+      rider:
+        statuses: []
+    disabled: false
+    start: null
+    duration:
+      value: 1
+      units: seconds
```
Fyorl:
```suggestion
      units: hours
```

packs/_source/items/poison/torpor.yml
```diff
@@ -0,0 +1,185 @@
...
+        message: null
+      conditions: '{}'
+      magical: false
+      rider:
+        statuses: []
+    disabled: false
+    start: null
+    duration:
+      value: 4
+      units: hours
```
Fyorl: Should we leave the duration unset so it can be input after rolling the 4d6?

packs/_source/items/poison/pale-tincture.yml
```diff
@@ -0,0 +1,217 @@
...
+        message: null
+      conditions: '{}'
+      magical: false
+      rider:
+        statuses: []
+    disabled: false
+    start: null
+    duration:
+      value: 24
+      units: hours
```
Fyorl: This requires 7 successful saves so its duration is potentially indefinite.

packs/_source/items/poison/midnight-tears.yml
```diff
@@ -0,0 +1,160 @@
...
+  activities:
+    dnd5eactivity000:
+      _id: dnd5eactivity000
+      type: save
+      activation:
+        type: action
+        value: 1
+        condition: ''
+        override: false
+      consumption:
```
Fyorl: Missing consumption

packs/_source/rules/chapter-11-dm-tools.yml
Fyorl: While we are here I noticed 5 counts of "Injested". The Truth Serum effect is also labelled as "Truthfull."

Fyorl (changes requested):

packs/_source/items/poison/torpor.yml
arbron: Yeah, we should probably consider implementing https://github.com/foundryvtt/dnd5e/issues/7228 at some point so the activity can roll the duration and then set it on the active effect when it is applied.

Fyorl: Yeah that would be a good one to do

Fyorl (approved):

